PR #8447, opened 2021-01-11
Pin scalafmt version in configuration file

stefanobaghino-da:
Looks like there is no other way for IntelliJ to pick up an arbitrary version.

See: https://www.jetbrains.com/help/idea/work-with-scala-formatter.html#change_scalafmt

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,8 +1,8 @@
+version = 2.7.5
+
 project.git = true
 
 align.preset = none # never align to make the fmt more diff friendly
 maxColumn = 100
 runner.fatalWarnings = true
 trailingCommas = multiple
-
-version = 2.7.5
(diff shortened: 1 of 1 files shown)


.scalafmt.conf
```diff
@@ -4,3 +4,5 @@ align.preset = none # never align to make the fmt more diff friendly
 maxColumn = 100
 runner.fatalWarnings = true
 trailingCommas = multiple
+
+version = 2.7.5
```
cocreature, line 8: bikeshedding but maybe put it at the top given that it determines which of the following options are valid?

cocreature (approved):
Thanks!

.scalafmt.conf
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8447/commits/bb0753c3289d3490969dc83304af1faf72ea5963

PR #8452, opened 2021-01-11
Build and test //ledger-api/... against Scala 2.13

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


ledger-api/perf-testing/src/main/scala/com/digitalasset/ledger/api/perf/util/PerformanceTest.scala
```diff
@@ -49,8 +49,8 @@ abstract class PerformanceTest extends Bench[Double] {
 
   @transient lazy val persistor = Persistor.None
 
-  protected def daConfig: Array[org.scalameter.KeyValue] =
-    Array(
+  protected def daConfig: Seq[org.scalameter.KeyValue] =
+    Seq[org.scalameter.KeyValue](
```
stefanobaghino-da, line 53: Probably not particularly important, but should we use some sort of `IndexedSeq` here instead, since we are replacing an `Array`?

ledger-api/perf-testing/src/main/scala/com/digitalasset/ledger/api/perf/util/reporter/JMeterXmlGenerator.scala
```diff
@@ -91,10 +91,11 @@ private[reporter] object JMeterXmlGenerator {
     }
 
     private def paramSuffix(parameters: Parameters): String = {
-      val filtered: Vector[String] = parameters.axisData
+      val filtered: Vector[String] = parameters.axisData.view
```
stefanobaghino-da, line 94: ❤️

stefanobaghino-da (approved):
I left a couple of comments which are probably not too important, overall LGTM.

ledger-api/perf-testing/src/main/scala/com/digitalasset/ledger/api/perf/util/PerformanceTest.scala
cocreature: The only use is in https://github.com/digital-asset/daml/blob/57217f33e8f5d0fa1944c4c6698c42bd2aab66b3/ledger-api/rs-grpc-akka/src/perf/scala/com/digitalasset/grpc/adapter/ServerStreamingBenchmark.scala#L34 which seemed to fall under the category of “I really don’t care what data structure we use here” which means I generally default to `Seq` but happy to use something else.

ledger-api/perf-testing/BUILD.bazel
```diff
@@ -4,20 +4,30 @@
 load(
     "//bazel_tools:scala.bzl",
     "da_scala_library",
+    "silencer_plugin",
 )
 
 da_scala_library(
     name = "perf-testing",
```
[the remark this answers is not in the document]
cocreature: I think we don’t run them but we compile them :shrug:

ledger-api/perf-testing/src/main/scala/com/digitalasset/ledger/api/perf/util/PerformanceTest.scala
stefanobaghino-da: The `Seq` should actually be better in this case, [IIRC](https://stackoverflow.com/a/54323858/3314107). Thanks for pointing that out.

ledger-api/perf-testing/BUILD.bazel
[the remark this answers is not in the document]
stefanobaghino-da: Given the recent progress with performance tracking, this is probably safe to delete at this point (in a separate PR). @gerolf-da, do you agree?

PR #8457, opened 2021-01-11
Remove redundant transaction decoding in TransactionCommitter

tudor-da:
Removes double decoding of transactions in `TransactionCommitter`

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
@@ -19,6 +19,7 @@ import com.daml.lf.data.Ref.{PackageId, Party}
 import com.daml.lf.data.Time.Timestamp
 import com.daml.lf.engine.{Blinding, Engine, ReplayMismatch}
 import com.daml.lf.language.Ast
+import com.daml.lf.transaction.Transaction.Transaction
 import com.daml.lf.transaction.{
   BlindingInfo,
   GlobalKey,
@@ -54,6 +55,8 @@ private[kvutils] class TransactionCommitter(
     engine: Engine,
     override protected val metrics: Metrics,
     inStaticTimeMode: Boolean,
+    decodeTransaction: TransactionOuterClass.Transaction => Transaction =
+      Conversions.decodeTransaction,
 ) extends Committer[DamlTransactionEntrySummary] {
   override protected val committerName = "transaction"
 
@@ -61,7 +64,7 @@ private[kvutils] class TransactionCommitter(
       commitContext: CommitContext,
       submission: DamlSubmission,
   ): DamlTransactionEntrySummary =
-    DamlTransactionEntrySummary(submission.getTransactionEntry)
+    DamlTransactionEntrySummary(submission.getTransactionEntry, decodeTransaction)
 
   override protected val steps: Iterable[(StepInfo, Step)] = Iterable(
     "authorize_submitter" -> authorizeSubmitters,
@@ -332,7 +335,7 @@ private[kvutils] class TransactionCommitter(
       val blindingInfo = Blinding.blind(transactionEntry.transaction)
       setDedupEntryAndUpdateContractState(
         commitContext,
-        transactionEntry.copy(
+        transactionEntry.copyPreservingDecodedTransaction(
           submission = transactionEntry.submission.toBuilder
             .setBlindingInfo(Conversions.encodeBlindingInfo(blindingInfo))
             .build
@@ -384,7 +387,7 @@ private[kvutils] class TransactionCommitter(
       .setTransaction(newTransaction)
       .build()
 
-    StepContinue(DamlTransactionEntrySummary(newTransactionEntry))
+    StepContinue(DamlTransactionEntrySummary(newTransactionEntry, decodeTransaction))
   }
 
   /** Builds the log entry as the final step.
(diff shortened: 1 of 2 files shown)


fabiotudone-da (approved):
LGTM, do we know how much we gain?

mziolekda (commented):
LGTM, but according to my agreement with Miklos, a member of his team has to approve.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
```diff
@@ -748,15 +748,29 @@ private[kvutils] class TransactionCommitter(
...
     val ledgerEffectiveTime: Timestamp = parseTimestamp(submission.getLedgerEffectiveTime)
     val submitterInfo: DamlSubmitterInfo = submission.getSubmitterInfo
     val commandId: String = submitterInfo.getCommandId
     val submitters: List[Party] =
       submitterInfo.getSubmittersList.asScala.toList.map(Party.assertFromString)
-    lazy val transaction: Tx.Transaction = Conversions.decodeTransaction(submission.getTransaction)
+    lazy val transaction: Tx.Transaction = tx
     val submissionTime: Timestamp = Conversions.parseTimestamp(submission.getSubmissionTime)
     val submissionSeed: crypto.Hash = Conversions.parseHash(submission.getSubmissionSeed)
+    // On copy, avoid decoding the transaction again if not needed
```
SamirTalwar-DA, line 760: Please add a newline here.

```suggestion

    // On copy, avoid decoding the transaction again if not needed
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
```diff
@@ -748,15 +748,29 @@ private[kvutils] class TransactionCommitter(
...
+    def copyPreservingDecodedTransaction(
+        submission: DamlTransactionEntry
+    ): DamlTransactionEntrySummary = {
+      new DamlTransactionEntrySummary(submission, transaction)
```
SamirTalwar-DA, line 764: At this point, you are forcing `transaction` regardless, so why not just do it up front and make `tx` a strict (non-lazy) value?

(And rename it to `transaction`, as you will no longer need the `lazy val` later.)

tudor-da: I've kept the lazy semantics for historical reasons, since I believe it's an optimization (e.g. `authorizeSubmitters` does not use it so if this step emits `StepStop` it won't ever be needed).

tudor-da: On a closer look, `authorizeSubmitters` would cause an abort just in exceptional cases, so I guess there won't be a significant impact if we decode it eagerly.

tudor-da: Done

tudor-da:
> LGTM, do we know how much we gain?

As obeserved previously through profiling, ~1.8 ms per submission.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
```diff
@@ -748,15 +748,29 @@ private[kvutils] class TransactionCommitter(
...
-    lazy val transaction: Tx.Transaction = Conversions.decodeTransaction(submission.getTransaction)
     val submissionTime: Timestamp = Conversions.parseTimestamp(submission.getSubmissionTime)
     val submissionSeed: crypto.Hash = Conversions.parseHash(submission.getSubmissionSeed)
+
+    // On copy, avoid decoding the transaction again if not needed
+    def copyPreservingDecodedTransaction(
+        submission: DamlTransactionEntry
+    ): DamlTransactionEntrySummary = {
+      new DamlTransactionEntrySummary(submission, transaction)
+    }
```
miklos-da, line 765:
```suggestion
    ): DamlTransactionEntrySummary =
      new DamlTransactionEntrySummary(submission, transaction)
```

miklos-da (approved):

SamirTalwar-DA (approved):

SamirTalwar-DA:
Looks like this test is failing, somehow:

```
TransactionCommitterSpec:
com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec *** ABORTED *** (6 milliseconds)
  java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec
```

I have no idea what this means.

miklos-da:
> Looks like this test is failing, somehow:
> 
> ```
> TransactionCommitterSpec:
> com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec *** ABORTED *** (6 milliseconds)
>   java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec
> ```
> 
> I have no idea what this means.

Could it be that the test suite cannot be initialized due to an exception being thrown by initialization of one of the `val`s?

tudor-da:
> > Looks like this test is failing, somehow:
> > ```
> > TransactionCommitterSpec:
> > com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec *** ABORTED *** (6 milliseconds)
> >   java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec
> > ```
> > 
> > 
> > I have no idea what this means.
> 
> Could it be that the test suite cannot be initialized due to an exception being thrown by initialization of one of the `val`s?

Yep. Transaction decoding kicks in in tests eagerly now so it fails. Looking into it

tudor-da:
> > > Looks like this test is failing, somehow:
> > > ```
> > > TransactionCommitterSpec:
> > > com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec *** ABORTED *** (6 milliseconds)
> > >   java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: com.daml.ledger.participant.state.kvutils.committer.TransactionCommitterSpec
> > > ```
> > > 
> > > 
> > > I have no idea what this means.
> > 
> > 
> > Could it be that the test suite cannot be initialized due to an exception being thrown by initialization of one of the `val`s?
> 
> Yep. Transaction decoding kicks in in tests eagerly now so it fails. Looking into it

Needed to _mock_ `decodeTransaction`. CC @SamirTalwar-DA @miklos-da

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
tudor-da: > At this point, you are forcing `transaction` regardless, so why not just do it up front and make `tx` a strict (non-lazy) value?
> 
> (And rename it to `transaction`, as you will no longer need the `lazy val` later.)

@SamirTalwar-DA , I was just about to merge this but I double-checked the need for `lazy`. I missed the fact that the `DamlTransactionEntrySummary` is created again in `trimUnnecessaryNodes`, but the decoded transaction is however not needed in the subsequent step `buildFinalLogEntry`. I'll revert to using the lazy implementation

PR #8463, opened 2021-01-11
Update offset when party/config rejected on a duplicate

mziolekda:
CHANGELOG_BEGIN
Advance ledger end when the indexer encounters an duplicate party or configuration update. This allows indexer to skip-over the unexpected message and continue processing subsequent updates.
CHANGELOG_END

### DIFF
--- a/ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
+++ b/ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
@@ -103,7 +103,7 @@ private[dao] trait JdbcLedgerDaoPartiesSpec {
     for {
       response <- storePartyEntry(fred, nextOffset())
       _ = response should be(PersistenceResponse.Ok)
-      response <- storePartyEntry(fred, nextOffset(), shouldUpdateLegerEnd = false)
+      response <- storePartyEntry(fred, nextOffset())
     } yield {
       response should be(PersistenceResponse.Duplicate)
     }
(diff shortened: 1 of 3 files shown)


rautenrieth-da (approved):
The fix and the new tests looks good.

I think at some point we should move away from the current approach (unconditional insert + scan the error message) to something that does not rely on SQL errors. For example, add `ON CONFLICT ... DO NOTHING` to the insert statement and check the number of written rows to determine whether to emit a log warning.

tudor-da (approved):

oliverse-da:
Thank you, @mziolekda !

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
```diff
@@ -123,18 +123,47 @@ private[dao] trait JdbcLedgerDaoPartiesSpec {
...
+  it should "store just offset when duplicate party update encountered" in {
+    val party = Ref.Party.assertFromString(s"Alice-${UUID.randomUUID()}")
+    val aliceDetails = PartyDetails(
+      party,
+      displayName = Some("Alice Arkwright"),
+      isLocal = true,
+    )
+    val aliceAgain = PartyDetails(
+      party,
+      displayName = Some("Alice Carthwright"),
```
miklos-da, line 135: Does the `displayName` matter from the test case's perspective? If not, you could simply assign `aliceAgain` to `aliceDetails`.

rautenrieth-da: It does, the test asserts that the second `storePartyEntry` call did not alter the already existing entry (see `parties should contain allOf (aliceDetails, ...)`. We could have decided to update existing entries instead.

PR #8465, opened 2021-01-11
LF: get ride of user-defined patterns in ValueTranslation.

remyhaemmerle-da:
We prune the code of ValueTransalation from the user-defined patterns
in the code of Speedy's ValueTranslation. In particular we remove the usage of
com.daml.lf.language.Util.TTyConApp which:
- induces bad complexity,
- make difficult optimization.

CHANGELOG_BEGIN
CHANGELOG_END 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
@@ -22,7 +22,7 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages) 
   // this is not tail recursive, but it doesn't really matter, since types are bounded
   // by what's in the source, which should be short enough...
   @throws[PreprocessorException]
-  private[this] def replaceParameters(params: Traversable[(TypeVarName, Type)], typ0: Type): Type =
+  private[this] def replaceParameters(params: Iterable[(TypeVarName, Type)], typ0: Type): Type =
     if (params.isEmpty) { // optimization
       typ0
     } else {
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
@@ -79,12 +79,11 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages) 
     compiledPackages.getSignature(pkgId).getOrElse(throw PreprocessorMissingPackage(pkgId))
 
   @tailrec
-  private[this] def destructApp(typ: Type, args: List[Type] = List.empty):  (Type,  List[Type]) =
+  private[this] def destructApp(typ: Type, args: List[Type] = List.empty): (Type, List[Type]) =
     typ match {
-      case TApp(tyfun, arg) => destructApp(tyfun, arg:: args)
+      case TApp(tyfun, arg) => destructApp(tyfun, arg :: args)
       case otherwise => otherwise -> args
     }
-
 
   // For efficient reason we do not produce here the monad Result[SValue] but rather throw
   // exception in case of error or package missing.
@@ -154,25 +153,25 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages) 
                     } else {
                       SValue.SGenMap(
                         isTextMap = true,
-                        entries = entries.iterator.map {
-                          case (k, v) =>
-                            SValue.SText(k) -> go(typeArg0, v, newNesting)
-                        }
+                        entries = entries.iterator.map { case (k, v) =>
+                          SValue.SText(k) -> go(typeArg0, v, newNesting)
+                        },
                       )
                     }
                   case _ =>
                     typeMismatch
                 }
-              case typeArg0 :: typeArg1 :: Nil if bt == BTGenMap && value.isInstanceOf[ValueGenMap[_]] =>
+              case typeArg0 :: typeArg1 :: Nil
+                  if bt == BTGenMap && value.isInstanceOf[ValueGenMap[_]] =>
                 val ValueGenMap(entries) = value
                 if (entries.isEmpty) {
                   SValue.SValue.EmptyTextMap
                 } else {
                   SValue.SGenMap(
                     isTextMap = true,
-                    entries = entries.iterator.map {
-                      case (k, v) => go(typeArg0, k, newNesting) -> go(typeArg1, v, newNesting)
-                    }
+                    entries = entries.iterator.map { case (k, v) =>
+                      go(typeArg0, k, newNesting) -> go(typeArg1, v, newNesting)
+                    },
                   )
                 }
               case _ =>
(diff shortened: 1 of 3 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -23,7 +22,7 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
   // this is not tail recursive, but it doesn't really matter, since types are bounded
   // by what's in the source, which should be short enough...
   @throws[PreprocessorException]
-  private def replaceParameters(params: ImmArray[(TypeVarName, Type)], typ0: Type): Type =
+  private[this] def replaceParameters(params: Traversable[(TypeVarName, Type)], typ0: Type): Type =
```
cocreature, line 25:
```suggestion
  private[this] def replaceParameters(params: Iterable[(TypeVarName, Type)], typ0: Type): Type =
```
scala 2.13 killed traversable

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +96,175 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
-
-          // genMap
-          case (TGenMap(keyType, valueType), ValueGenMap(entries)) =>
-            SValue.SGenMap(
-              isTextMap = false,
-              entries = entries.iterator.map { case (k, v) =>
-                go(keyType, k, newNesting) -> go(valueType, v, newNesting)
-              },
-            )
+        def newNesting = nesting + 1
```
cocreature, line 102: why did you change this to a def instead of a val?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +95,184 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
+                  case (BTText, ValueText(t)) =>
+                    SValue.SText(t)
+                  case (BTParty, ValueParty(p)) =>
+                    SValue.SParty(p)
+                  case _ =>
+                    typeMismatch
+                }
+              case typeArg0 :: Nil =>
+                (bt, value) match {
+                  case (BTNumeric, ValueNumeric(d)) if typeArg0.isInstanceOf[TNat] =>
```
cocreature, line 129: I’m not that fond of `isInstanceOf`, can we pattern match instead?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +95,184 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
+                        entries = entries.iterator.map { case (k, v) =>
+                          SValue.SText(k) -> go(typeArg0, v, newNesting)
+                        },
+                      )
+                    }
+                  case _ =>
+                    typeMismatch
+                }
+              case typeArg0 :: typeArg1 :: Nil
+                  if bt == BTGenMap && value.isInstanceOf[ValueGenMap[_]] =>
```
cocreature, line 165: more `isinstanceOf` :disappointed:

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +95,184 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
+                val ValueGenMap(entries) = value
+                if (entries.isEmpty) {
+                  SValue.SValue.EmptyTextMap
+                } else {
+                  SValue.SGenMap(
+                    isTextMap = true,
```
cocreature, line 171: I think this should be false.

cocreature (commented):
Looks good but I’d like to get rid of the `asInstanceOf` before approving and merging.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: reverted

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: replaced.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: replaced as well.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: good catch !. fixed.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +97,190 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
+                    } else {
+                      SValue.SList(ls.map(go(typeArg0, _, newNesting)))
+                    }
+                  // textMap
+                  case (BTTextMap, ValueTextMap(entries)) =>
+                    if (entries.isEmpty) {
+                      SValue.SValue.EmptyTextMap
+                    } else {
+                      SValue.SGenMap(
+                        isTextMap = false,
```
cocreature, line 160: I think you replaced the wrong one? you are matching on `TextMap` here so this should be true. The one below where you’re matching on `GenMap` on the other hand should be false.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -89,156 +97,190 @@ private[engine] final class ValueTranslator(compiledPackages: CompiledPackages)
...
-                fail(
-                  s"Mismatching record id, the type tells us $typeRecordId, but the value tells us $id"
+                variantDef.constructorRank.get(constructorName) match {
+                  case None =>
+                    fail(
+                      s"Couldn't find provided variant constructor $constructorName in variant $tyCon"
+                    )
+                  case Some(rank) =>
+                    val (_, argTyp) = variantDef.variants(rank)
+                    val replacedTyp =
```
cocreature, line 209: is `toSeq.view` actually more efficient than just `map`?

cocreature (approved):
LGTM apart from the `isTextMap` fields which I think are now both wrong.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: I need another coffee.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: In this case, I think it is.

First I need the result to be `Iterable` (to be able to use `.toMap`). 

1 - Since `dataTypParams` is an ImmArray I need a `.toSeq` somewhere. 
2 - `dataTypParams.toSeq.map(_._1).zip(tyArgs)` will allocate 2 intermediate `ImmArraySeq`
3 - In general Array builder allocate several intermediate arrays (as they do not know the final size of the array they are building).

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
remyhaemmerle-da: I had a test.

cocreature: who needs coffee if you have tests instead :slightly_smiling_face:

PR #8466, opened 2021-01-11
Upgrade scopt to 4.0.0

cocreature:
Scopt 3.x has some issues with Scala 2.13 because it expects an
immutable Seq on 2.13 meaning you cannot just pass in an Array. Rather
than fixing our callsites to convert to an immutable Seq everywhere,
this PR bumps to Scopt 4.0 which goes back to collection.Seq.

and leaving that aside, I’m a fan of upgrading dependencies anyway :)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


extractor/src/main/scala/com/digitalasset/extractor/config/ConfigParser.scala
```diff
@@ -50,7 +50,7 @@ object ConfigParser {
   private val configParser: OptionParser[CliParams] =
     new scopt.OptionParser[CliParams]("extractor") {
 
-      override def showUsageOnError: Boolean = true
+      override def showUsageOnError: Option[Boolean] = Some(true)
```
stefanobaghino-da, line 53: As a minor side note, this can be a `val`. But more importantly, do you know what this API change means? An `Option[Boolean]` sounds a bit weird.

stefanobaghino-da (approved):

extractor/src/main/scala/com/digitalasset/extractor/config/ConfigParser.scala
cocreature: Very good question, I admit I hadn’t looked at it.

Previously the behavior was: If `showUsageOnError` is true, show usage on error, else call `showTryHelp`. Now it is the following:
If `showUsageOnError` is true show usage on error, if it is false call `showTryHelp`. If it is nothing treat it as true iff there is no help option, otherwise as false. So setting it to `Some(b)` appears to be the equivalent of `b` before.

PR #8472, opened 2021-01-12
Build DAML Script library for all LF versions

cocreature:
The hardcoded list is obviously garbage and didn’t include 1.11 so
this PR changes it to use the one we already have so they are always
in sync.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -21,6 +21,10 @@ LF_VERSIONS = [
     "dev",
 ]
 
+# The subset of LF versions accepted by the compiler in the syntax
+# expected by the --target option.
+COMPILER_LF_VERSIONS = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS if ver != "1.6"]
+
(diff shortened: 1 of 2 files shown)


cocreature:
Not sure how to write a useful test for this. Whatever I can come up just relies on the list of versions in the test being in sync with the one here which doesn’t seem super useful.

aherrmann-da (approved):

remyhaemmerle-da (commented):
:+1:

bame-da:
@cocreature Maybe a useful test as part of our compatibility suite would be to `daml start` the skeleton project with `--target=X` for all supported LF versions?

cocreature:
> @cocreature Maybe a useful test as part of our compatibility suite would be to `daml start` the skeleton project with `--target=X` for all supported LF versions?

The problem is that "all suported LF versions" needs to come from somewhere and keeping the list in the test in sync with the list of available versions is basically the same problem as keeping the list of versions we build daml script against in sync with the list of available versions.

PR #8473, opened 2021-01-12
only update NOTICES from main daily compats

garyverhaegen-da:
See #8468 for failure mode of current config.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #8475, opened 2021-01-12
support Eq, Ord, Show for all tuples

garyverhaegen-da:
If the language allows for tuples up to 20 elements, it should also fully support them. I think. Unless that breaks something?

```
CHANGELOG_BEGIN

- [Daml] Tuples up to size 20 now fully support the base type classes
  Eq, Ord and Show. Previously, Eq and Ord were only supported up to 15
  elements, while Show was only supported up to 5 elements.

  The maximum number of elements in a tuple remains 20.

  Note that this should not be seen as encouraging the use of large
  tuples; in almost all cases where you end up having a tuple with 4 or
  more elements, you would likely be better served by introducing names
  for your elements, e.g. by using a record instead.

CHANGELOG_END
```


garyverhaegen-da:
Note: I've tested this locally with `daml-sdk-head` and the following script:
```
module Main where

import Daml.Script

setup : Script ()
setup = do
  let tuple20 = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20)
  let zeroes = (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
  assert $ tuple20 == tuple20
  assert $ tuple20 > zeroes
  debug $ show tuple20
  return ()
```

Happy to add that to a test suite somewhere if you point me to the right place.

cocreature:
I think we considered that before and decided against it. Haskell also omits those instances.

It increases DAR size, compile times (hopefully only on our CI but the instances in scope can and do slow down typechecking) and tbh I think nobody should be using tuples that large anyway.

garyverhaegen-da:
> I think nobody should be using tuples that large anyway.

I would agree, but then I'd argue the language should just not have tuples above 5 elements. I really don't like having them but slightly broken.

I've updated the PR to instead document the decision. Hopefully that'll save the next well-meaning dev some trouble. Feel free to close instead if you prefer not changing those files at all.

cocreature (approved):

cocreature:
> I would agree, but then I'd argue the language should just not have tuples above 5 elements. I really don't like having them but slightly broken.

Yeah that’s a fair point. I’m not actually sure why we do have them. I don’t think GHC forces us but I could be wrong.

garyverhaegen-da:
I suppose at this point that ship has sailed. Removing them would require a long deprecation cycle and I assuming non-trivial tooling in the compiler to produce deprecation warnings for people currently using them. Oh well. 🤷

cocreature:
Yeah removing them now isn’t an option imho. We also don’t gain that much by removing them.

PR #8480, opened 2021-01-12
fix race condition in DB update when ACS has later contracts than the ledger-end

S11001001:
Demonstrated in https://github.com/digital-asset/daml/pull/8226#issuecomment-756446537. Extracted from #8226, which tests for this by virtue of a websocket test that is highly likely to exercise this race condition.

```rst
CHANGELOG_BEGIN
- [JSON API] Under rare conditions, a multi-template query backed by Postgres
  could have mismatched snapshots of the ACS for different templates.  These
  conditions are now checked and accounted for.
  See `issue #8226 <https://github.com/digital-asset/daml/pull/8226#issuecomment-756446537>`__.
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
```diff
@@ -73,25 +73,62 @@ private class ContractsFetch(
...
+          // re-traverse any that != the max returned bookmark (overriding lastOffset)
+          // fetch cannot go "too far" the second time
+          templateIds.traverse(fetchAndPersist(jwt, parties, None, absEnd, _)).flatMap {
+            actualAbsEnds =>
+              val newAbsEndTarget = {
+                import scalaz.std.list._, scalaz.syntax.foldable._,
+                domain.Offset.{ordering => `Offset ordering`}
+                // it's fine if all yielded LedgerBegin, so we don't want to conflate the "fallback"
+                // with genuine results
+                actualAbsEnds.maximum getOrElse AbsoluteBookmark(absEnd.toDomain)
```
cocreature, line 95: We only reach the fallback if the list of template ids is empty right?

cocreature (approved):
Nice, thank you! I think some comments that describethe conditions  when we hit this would be nice and if possible of course a test although I’m not clear on how to test this reliably unfortunately.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
S11001001: Right.

PR #8481, opened 2021-01-12
Add grpc-reverse-proxy with optional server interceptors

stefanobaghino-da:
- https://github.com/digital-asset/daml/commit/f1578df699770ac738416bc443ea322b7c22a618: factor `withServices` out of `ServerReflectionClientSpec`
- https://github.com/digital-asset/daml/commit/efc0d714ca74e8282617e4413ed759b42a7278f0: add `grpc-reverse-proxy` with optional server interceptors

### Pull Request Checklist

trigger the build.

### DIFF
--- a/libs-scala/grpc-reverse-proxy/BUILD.bazel
+++ b/libs-scala/grpc-reverse-proxy/BUILD.bazel
@@ -15,7 +15,7 @@ da_scala_library(
     ],
     deps = [
         "//libs-scala/grpc-server-reflection-client",
-        "@maven//:com_google_protobuf_protobuf_java",
+        "@maven//:com_google_guava_guava",
         "@maven//:io_grpc_grpc_api",
         "@maven//:io_grpc_grpc_services",
         "@maven//:io_grpc_grpc_stub",
--- a/libs-scala/grpc-reverse-proxy/BUILD.bazel
+++ b/libs-scala/grpc-reverse-proxy/BUILD.bazel
@@ -26,6 +26,9 @@ da_scala_library(
 da_scala_test_suite(
     name = "test",
     srcs = glob(["src/test/scala/**/*.scala"]),
+    versioned_scala_deps = {
+        "2.12": ["@maven//:org_scala_lang_modules_scala_collection_compat"],
+    },
     deps = [
         ":grpc-reverse-proxy",
         "//libs-scala/grpc-test-utils",
--- a/libs-scala/grpc-reverse-proxy/BUILD.bazel
+++ b/libs-scala/grpc-reverse-proxy/BUILD.bazel
@@ -1,0 +1,37 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+load(
+    "//bazel_tools:scala.bzl",
+    "da_scala_library",
+    "da_scala_test_suite",
+)
+
+da_scala_library(
+    name = "grpc-reverse-proxy",
+    srcs = glob(["src/main/scala/**/*.scala"]),
+    visibility = [
+        "//:__subpackages__",
+    ],
+    deps = [
+        "//libs-scala/grpc-server-reflection-client",
+        "@maven//:com_google_protobuf_protobuf_java",
+        "@maven//:io_grpc_grpc_api",
+        "@maven//:io_grpc_grpc_services",
+        "@maven//:io_grpc_grpc_stub",
+    ],
+)
+
+da_scala_test_suite(
+    name = "test",
+    srcs = glob(["src/test/scala/**/*.scala"]),
+    deps = [
+        ":grpc-reverse-proxy",
+        "//libs-scala/grpc-test-utils",
+        "@maven//:com_google_protobuf_protobuf_java",
+        "@maven//:io_grpc_grpc_api",
+        "@maven//:io_grpc_grpc_core",
+        "@maven//:io_grpc_grpc_services",
+        "@maven//:io_grpc_grpc_stub",
+    ],
+)
--- a/libs-scala/grpc-reverse-proxy/BUILD.bazel
+++ b/libs-scala/grpc-reverse-proxy/BUILD.bazel
@@ -10,6 +10,7 @@ load(
 da_scala_library(
     name = "grpc-reverse-proxy",
     srcs = glob(["src/main/scala/**/*.scala"]),
+    tags = ["maven_coordinates=com.daml:grpc-reverse-proxy:__VERSION__"],
     visibility = [
         "//:__subpackages__",
     ],
--- a/libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
+++ b/libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
 package com.daml.grpc
--- a/libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
+++ b/libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
@@ -3,24 +3,24 @@
 
 package com.daml.grpc
 
-import java.io.InputStream
+import java.io.{ByteArrayInputStream, InputStream}
 
 import com.daml.grpc.reflection.ServiceDescriptorInfo
-import com.google.protobuf.ByteString
+import com.google.common.io.ByteStreams
 import io.grpc.{CallOptions, Channel, MethodDescriptor, ServerServiceDefinition}
 
 private[grpc] object ForwardService {
 
   def apply(backend: Channel, service: ServiceDescriptorInfo): ServerServiceDefinition =
     service.methods
-      .map(_.toMethodDescriptor(NoopMarshaller, NoopMarshaller))
+      .map(_.toMethodDescriptor(ByteArrayMarshaller, ByteArrayMarshaller))
       .map(ForwardCall(_, backend, CallOptions.DEFAULT))
       .foldLeft(ServerServiceDefinition.builder(service.fullServiceName))(_ addMethod _)
       .build()
 
-  private object NoopMarshaller extends MethodDescriptor.Marshaller[ByteString] {
-    override def parse(input: InputStream): ByteString = ByteString.readFrom(input)
-    override def stream(bytes: ByteString): InputStream = bytes.newInput()
+  private object ByteArrayMarshaller extends MethodDescriptor.Marshaller[Array[Byte]] {
+    override def parse(input: InputStream): Array[Byte] = ByteStreams.toByteArray(input)
+    override def stream(bytes: Array[Byte]): InputStream = new ByteArrayInputStream(bytes)
   }
 
 }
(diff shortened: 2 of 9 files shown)


libs-scala/grpc-reverse-proxy/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+load(
+    "//bazel_tools:scala.bzl",
+    "da_scala_library",
+    "da_scala_test_suite",
+)
+
+da_scala_library(
```
cocreature, line 10: Needs more Scala 2.13 :slightly_smiling_face:

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
```diff
@@ -3,35 +3,28 @@
...
   import ServerReflectionClientSpec._
+  import Services._
 
   behavior of "getAllServices"
 
   it should "fail if reflection is not supported" in withServices(health) { channel =>
     val stub = ServerReflectionGrpc.newStub(channel)
     val client = new ServerReflectionClient(stub)
-    client.getAllServices().failed.futureValue shouldBe a[StatusRuntimeException]
+    client.getAllServices().failed.map(_ shouldBe a[StatusRuntimeException])
```
cocreature, line 23: doesn’t this mean that now you’ll succeed if it doesn’t fail?

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def apply(backend: Channel, service: ServiceDescriptorInfo): ServerServiceDefinition =
+    service.methods
+      .map(_.toMethodDescriptor(ByteArrayMarshaller, ByteArrayMarshaller))
+      .map(ForwardCall(_, backend, CallOptions.DEFAULT))
+      .foldLeft(ServerServiceDefinition.builder(service.fullServiceName))(_ addMethod _)
+      .build()
+
+  private object ByteArrayMarshaller extends MethodDescriptor.Marshaller[Array[Byte]] {
+    override def parse(input: InputStream): Array[Byte] = ByteStreams.toByteArray(input)
```
cocreature, line 22: I realize you might have copied that from my hacked up prototype and I know that I did it this way because I couldn’t figure out another way but let me ask anyway: Can we stream this instead of materializing the input and output fully? It seems like it should work but I couldn’t quite figure out how to plug it together.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ReverseProxy.scala
```diff
@@ -0,0 +1,54 @@
...
+      backend: Channel,
+      serverBuilder: ServerBuilder[_],
+  )(implicit
+      ec: ExecutionContext
+  ): Future[Server] = create(backend, serverBuilder, Map.empty[String, Seq[ServerInterceptor]])
+
+  def create(
+      backend: Channel,
+      serverBuilder: ServerBuilder[_],
+      interceptor: (String, Seq[ServerInterceptor]),
```
cocreature, line 24: Why do we special-case the first one here?

libs-scala/grpc-reverse-proxy/src/test/scala/com/daml/grpc/ReverseProxySpec.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  it should "expose the backend's own services" in withServices(health, reflection) { backend =>
+    val proxyName = InProcessServerBuilder.generateName()
+    val proxyBuilder = InProcessServerBuilder.forName(proxyName)
+    ReverseProxy
+      .create(backend, proxyBuilder)
+      .map { proxyServer =>
+        proxyServer.start()
+        val proxy = InProcessChannelBuilder.forName(proxyName).build()
+        getHealthStatus(backend) shouldEqual getHealthStatus(proxy)
```
cocreature, line 49: Maybe also test reflection itself? You claim to expose the backend’s own services but currently you only test that the health endpoint is proxied through not that reflection is proxied through.

cocreature (approved):
Awesome, thank you!

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
stefanobaghino-da: It's the same as before, I just noticed that using `.futureValue` caused a deadlock in another test when using the serial `ExecutionContext` built-in into `AsyncFlatSpec`, so I preferred removing existing usages.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ReverseProxy.scala
stefanobaghino-da: This "pattern" forces the called to provide at least one interceptor.

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
cocreature: So we suffered from the same issue before or it’s not an issue?

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ReverseProxy.scala
cocreature: it does but then you have another overload that works without supplying an interceptor which seems to make that redundant?

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
stefanobaghino-da: I asked myself the first question, but without copying I had weird errors when running the conformance test suite on the proxied sandbox.

libs-scala/grpc-reverse-proxy/src/test/scala/com/daml/grpc/ReverseProxySpec.scala
stefanobaghino-da: Makes sense. I was just afraid of using the reflection client to not have weird circular dependencies between components, but probably I can find a way to make some form of simplified call just to make sure that the reflection endpoint is exposed.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
cocreature: yeah that sounds similar to what I was seeing. it looked like something might have been too lazy and was leaking resources but I couldn’t quite pin it down or fix it. Totally fine to ignore it for now.

On a related note since you mentioned it: should we have a test that runs the ledger api test tool against the proxied sandbox? Doesn’t have to be in this PR.

libs-scala/grpc-reverse-proxy/BUILD.bazel
stefanobaghino-da: Does it? It passes tests with both versions, doesn't it? 🤔

cocreature: oh good point duh, I got confused by the lack of `scala_deps` but you just don’t have Scala dependencies.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ReverseProxy.scala
stefanobaghino-da: Good point, I'll remove it.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
stefanobaghino-da: I considered it. Does it make sense to add it here? Or does it make more sense (although it's more complicated) to have this as an integration test on the non-repudiation middleware?

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
stefanobaghino-da: I had this issue while creating the tests for the reverse proxy. I can revert it back to what it was before, it's just that I'm not fully aware of how or why `.futureValue` can deadlock and I would prefer to leave it out where possible.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
cocreature: Just adding it to the non-repudiation middleware seems sufficient. I guess if we really want to we could also add it to both but I don’t see too much benefit.

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
cocreature: can we just turn this into an async test suite and wait on the future completing or if you don’t want to do that, `Await.result`? Looking through this again, it seems like the `map` will also executed asynchronously and it isn’t clear to me that the test cannot succeed before the future fails.

libs-scala/grpc-reverse-proxy/BUILD.bazel
stefanobaghino-da: As a side note, the gRPC Java API makes a lot of sense for Java, but probably there are things that can be improved to make it simpler and more typesafe to use via Scala. Not sure though, just a general sense I got while using it.

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ReverseProxy.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8481/commits/b641b9e8e409240cdfb83f701a036de09d8a2b97

libs-scala/grpc-reverse-proxy/src/test/scala/com/daml/grpc/ReverseProxySpec.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8481/commits/a4a7b9021935fa081adcc2f535ca7185dfb0cc12

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
stefanobaghino-da: > can we just turn this into an async test suite

Sorry, I lost you here, this is already an `AsyncFlatSpec`. 🤔 

> and wait on the future completing or if you don’t want to do that, `Await.result`? Looking through this again, it seems like the `map` will also executed asynchronously and it isn’t clear to me that the test cannot succeed before the future fails.

In an `AsyncFlatSpec` you are supposed to return a `Future[Assertion]` in your test blocks and Scalatest will take care of executing it correctly. You can test this by putting a failing assertion in the `map`.

Does this make sense or am I missing something?

libs-scala/grpc-reverse-proxy/src/main/scala/com/daml/grpc/ForwardService.scala
stefanobaghino-da: If the reverse proxy grows in scope it will probably make sense to add more tests. for sure.

libs-scala/grpc-server-reflection-client/src/test/scala/com/daml/grpc/reflection/ServerReflectionClientSpec.scala
cocreature: duh, I missed that this was already an async test suite and that this was the last statement. All my worries are unsubstantiated and this is perfectly fine. Sorry for the confusion

PR #8483, opened 2021-01-12
kvutils: Make migration to the new key and value types easier.

SamirTalwar-DA:
It proved quite painful to move to them without help.

These implicit conversions will eventually be removed.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Raw.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Raw.scala
@@ -5,20 +5,68 @@ package com.daml.ledger.participant.state.kvutils
 
 import com.google.protobuf.ByteString
 
+import scala.language.implicitConversions
+
 object Raw {
 
-  sealed trait Bytes {
+  trait Bytes {
     def bytes: ByteString
 
     final def size: Long = bytes.size.toLong
   }
 
+  object Bytes {
+
+    /** This implicit conversion exists to aid in migration.
+      * It will be deprecated and subsequently removed in the future.
+      */
+    implicit def `ByteString to Raw.Bytes`(bytes: ByteString): Bytes = Unknown(bytes)
+
+    /** This implicit conversion exists to aid in migration.
+      * It will be deprecated and subsequently removed in the future.
+      */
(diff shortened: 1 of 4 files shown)


fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Raw.scala
```diff
@@ -5,20 +5,62 @@ package com.daml.ledger.participant.state.kvutils
...
 
     final def size: Long = bytes.size.toLong
   }
 
+  object Bytes {
+
+    /** This implicit conversion exists to aid in migration.
+      * It will be deprecated and subsequently removed in the future.
+      */
+    implicit def `ByteString to Raw.Bytes`(bytes: ByteString): Bytes = Unknown(bytes)
```
miklos-da, line 23: Wouldn't we want to use @deprecated annotation for these?

SamirTalwar-DA: Yes, but I was thinking of doing that for 1.10, and removing them in 1.11.

SamirTalwar-DA: Do you think we should do this now?

miklos-da: Yes, I'd prefer adding the @deprecated annotations now so that we fix a timeline for actually removing these.

SamirTalwar-DA: I would like to at least finalize the types here. Currently I don't think `Raw.Key` and `Raw.Value` are good enough. I'd like to see (at least) `Raw.LogEntryId`, `Raw.StateKey`, and `Raw.Envelope` instead. There might be more that I'm missing.

Until then, I'd prefer to keep the implicit conversions so that users don't have to migrate twice.

SamirTalwar-DA: Follow-up PR at #8849.

PR #8488, opened 2021-01-13
Port more of //ledger/... to Scala 2.13

cocreature:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
@@ -70,8 +70,8 @@ private[kvutils] case class CommitContext(
     */
   def collectInputs[B, That](
       partialFunction: PartialFunction[(DamlStateKey, Option[DamlStateValue]), B]
-  )(implicit bf: BuildFrom[Map[DamlStateKey, Option[DamlStateValue]], B, That]): That = {
-    val builder = bf.newBuilder(inputs)
+  )(implicit bf: Factory[B, That]): That = {
+    val builder = bf.newBuilder
     builder ++= inputs.view.collect(partialFunction)
     inputs.keys.foreach(accessedInputKeys.add)
     builder.result()
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
@@ -70,8 +70,8 @@ private[kvutils] case class CommitContext(
     */
   def collectInputs[B, That](
       partialFunction: PartialFunction[(DamlStateKey, Option[DamlStateValue]), B]
-  )(implicit bf: Factory[B, That]): That = {
-    val builder = bf.newBuilder
+  )(implicit factory: Factory[B, That]): That = {
+    val builder = factory.newBuilder
     builder ++= inputs.view.collect(partialFunction)
     inputs.keys.foreach(accessedInputKeys.add)
     builder.result()
--- a/ledger/participant-state/kvutils/tools/BUILD.bazel
+++ b/ledger/participant-state/kvutils/tools/BUILD.bazel
@@ -55,7 +55,6 @@ da_scala_test(
         "@maven//:com_typesafe_akka_akka_stream",
         "@maven//:org_mockito_mockito_scala",
         "@maven//:org_scalatest_scalatest",
-        #        "@maven//:org_scala_lang_modules_scala_collection_compat",
     ],
     deps = [
         "//ledger/participant-integration-api",
--- a/ledger/participant-state/kvutils/tools/BUILD.bazel
+++ b/ledger/participant-state/kvutils/tools/BUILD.bazel
@@ -8,8 +8,10 @@ load(
     "da_scala_library",
     "da_scala_test",
     "lf_scalacopts",
+    "silencer_plugin",
 )
 load("@os_info//:os_info.bzl", "is_windows")
+load("@scala_version//:index.bzl", "scala_major_version")
 
 da_scala_library(
     name = "tools",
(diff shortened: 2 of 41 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
```diff
@@ -70,8 +70,8 @@ private[kvutils] case class CommitContext(
     */
   def collectInputs[B, That](
       partialFunction: PartialFunction[(DamlStateKey, Option[DamlStateValue]), B]
-  )(implicit bf: BuildFrom[Map[DamlStateKey, Option[DamlStateValue]], B, That]): That = {
```
cocreature, line 73: There don’t bee to be `BuildFrom` instances for things like `BuildFrom[Map[DamlStateKey, Option[DamlStateValue]], Unit, Seq[Unit]]` on Scala 2.13. I don’t think we gain much from `BuildFrom` here so using `Factory` seemed easier than trying to have two versions one for the cases where `BuildFrom` works and  one where it doesn’t. It does make type inference slightly worse but it’s not too bad.

ledger/participant-state/kvutils/tools/src/test/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityCheckerSpec.scala
```diff
@@ -163,5 +164,5 @@ final class IntegrityCheckerSpec
   }
 
   private def countOccurrences(input: String, pattern: String): Int =
-    input.sliding(pattern.length).count(_ == pattern)
+    new WrappedString(input).sliding(pattern.length).map(_.toString).count(_ == pattern)
```
cocreature, line 167: this is a bit of a mess. On 2.13, `sliding` on string is deprecated. `.toSeq.sliding` works but doesn’t work in 2.12. This convoluted mess works in both.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
SamirTalwar-DA: I'm going to have to take your word (and the tests') that this is legit. Perhaps @miklos-da can weigh in?

I suggest you rename this variable from `bf` to `factory`.

ledger/participant-state/kvutils/tools/BUILD.bazel
```diff
@@ -49,15 +51,18 @@ da_scala_test(
     size = "small",
     srcs = glob(["src/test/scala/**/*.scala"]),
     resources = glob(["src/test/resources/**/*"]),
+    scala_deps = [
+        "@maven//:com_typesafe_akka_akka_stream",
+        "@maven//:org_mockito_mockito_scala",
+        "@maven//:org_scalatest_scalatest",
+        #        "@maven//:org_scala_lang_modules_scala_collection_compat",
```
SamirTalwar-DA, line 58: Delete?

ledger/participant-state/kvutils/tools/src/test/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityCheckerSpec.scala
SamirTalwar-DA: How about `Pattern.quote(pattern).split(input).length - 1`?

Probably marginally less efficient, but this is a test, so :shrug: .

cocreature: That’s not equivalent for overlapping patterns but I guess we don’t care about that, fixed in 2a3c37c59920cf8a2ee7b6dbf4b10093183f91e8

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
cocreature: renamed in 0c1f641ea8935db50f0cf98660f41ad33bd2bd85. I’m happy to go for something else if you have a suggestion.

ledger/participant-state/kvutils/tools/BUILD.bazel
cocreature: :facepalm: f4dd2896451ad126de3f58059e0a89ccf717236f

ledger/participant-state/kvutils/tools/src/test/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityCheckerSpec.scala
cocreature: actually this is a mess. we rely on matches at the end afaict because we match on line endings. While we probably can fix this to make it work with a regex match, at that point it seems worse than this.

SamirTalwar-DA: Well, splitting should handle that (you'll get an empty string at the end of the array), but I'm happy with this solution too if you like.

SamirTalwar-DA (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
miklos-da: I don't have a better suggestion. `Factory` should be preferred in Scala 2.13 if you don't want to rebuild with the best matching type for an existing source collection (but would like to leave that open to the consumer). It makes sense in this instance as the user knows e.g. what `partialFunction` will be producing and how selective it is.

miklos-da (approved):

fabiotudone-da (approved):

PR #8490, opened 2021-01-13
Remove redundant list of LF versions

cocreature:
After #8472, I realized that there must be a list used for daml-stdlib
and daml-prim already and it turns out there is. I’ve removed that one
in favor of the one added in #8472 since I like having all in one
place and the one from #8472 is created by filtering an existing list
instead of creating a completely separate list like we do here.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -23,7 +23,7 @@ LF_VERSIONS = [
 
 # The subset of LF versions accepted by the compiler in the syntax
 # expected by the --target option.
-COMPILER_LF_VERSIONS = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS if ver != "1.6"]
+COMPILER_LF_VERSIONS = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS]
 
 LF_VERSION_PACKAGE_DIGITALASSET = {"1.6": "digitalasset", "1.7": "digitalasset", "1.8": "digitalasset", "1.11": "daml"}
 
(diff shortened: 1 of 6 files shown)


triggers/daml/BUILD.bazel
```diff
@@ -5,12 +5,7 @@
 # daml_compile instead of a genrule.
 
 load("@build_environment//:configuration.bzl", "ghc_version", "sdk_version")
-
-DAML_LF_VERSIONS = [
```
cocreature, line 9: and another bug caused by a redundant list :crying_cat_face:

daml-lf/language/daml-lf.bzl
```diff
@@ -23,7 +23,7 @@ LF_VERSIONS = [
 
 # The subset of LF versions accepted by the compiler in the syntax
 # expected by the --target option.
-COMPILER_LF_VERSIONS = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS if ver != "1.6"]
```
cocreature, line 26: I thought we stopped supporting 1.6 in the compiler but apparently we didn’t. We don’t support it in daml script and triggers since those rely on `Any`.

remyhaemmerle-da (approved):

daml-lf/language/daml-lf.bzl
sofiafaro-da: How do you feel about introducing a `SCRIPT_LF_VERSIONS` and/or `TRIGGER_LF_VERSIONS` below this to filter out "1.6", instead of filtering 1.6 everywhere else?

cocreature: Great idea, I added the filters one after the other so I didn’t fully realize how often I repeated myself. I’ve only introduced SCRIPT_LF_VERSIONS since they use the same and the differentiation didn’t seem super helpful. Triggers depend on script so they can never be more general.

sofiafaro-da: That makes a lot of sense!

PR #8491, opened 2021-01-13
fix out-of-rotation release notifs

garyverhaegen-da:
Only ping people on rotation for automated releases.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #8506, opened 2021-01-14
hide with-dlint option

garyverhaegen-da:
Quoting @cocreature on Slack:

> the fact that --with-dlint is exposed as a public option at all seems broken. I don’t think it was ever intended to be one.

```
CHANGELOG_BEGIN

- [Daml CLI] Three options that were never meant to be public are now
  hidden from the output of various `daml xxx --help` commands. The
  options still have the same effect when present, so this is not (yet)
  a breaking change. Using these options tends to result in very
  unhelpful, when not outright broken, dlint configurations, so we expect
  existing usage to be low to nonexistent. If you do use them please
  share your use-case with us. The options are:
  - `--with-dlint`
  - `--without-dlint`
  - `--allow-overrides`

CHANGELOG_END
```

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Options.hs
+++ b/compiler/damlc/lib/DA/Cli/Options.hs
@@ -216,16 +216,19 @@ dlintEnabledOpt = DlintEnabled
   <$> strOption
   ( long "with-dlint"
     <> metavar "DIR"
+    <> internal
     <> help "Enable linting with 'dlint.yaml' directory"
   )
   <*> switch
   ( long "allow-overrides"
+    <> internal
     <> help "Allow '.dlint.yaml' configuration overrides"
   )
 
 dlintDisabledOpt :: Parser DlintUsage
 dlintDisabledOpt = flag' DlintDisabled
   ( long "without-dlint"
+    <> internal
     <> help "Disable dlint"
   )
 


garyverhaegen-da:
Turns out this does not hide the option.

cocreature:
> Turns out this does not hide the option.

How did you check? this seems reasonable to me from a quick look.

garyverhaegen-da:
> How did you check? this seems reasonable to me from a quick look.

Using `daml-sdk-head` then `daml-head test --help` I still see the option.

cocreature:
Ah right you need `internal` not `hidden`. We probably also want to hide the inverse `without-dlint`.

garyverhaegen-da:
What about `--allow-overrides`? I'm not too clear on what it does, but it's next to the other two.

cocreature:
I’d also hide it. Afaict, it only has an effect in `damlc lint` if you use `with-dlint` which is half broken so it doesn’t seem like something users should ever use.

garyverhaegen-da:
Hidden all three now, and double-checked that `internal` does indeed result in them not showing up in the `--help` output. Note that this seems to affect all `damlc`-based commands (I tested `daml build`, `daml repl` and `daml test`).

cocreature:
The only two commands that have a working dlint config by default are `damlc ide` and `damlc dlint` and even those break if you add the options so I think removing it everywhere is exactly what we want.

cocreature:
Maybe worth mentioning in the changelog that these options mostly resulted in broken dlint configs.

cocreature (approved):
Thanks!

garyverhaegen-da:
> Maybe worth mentioning in the changelog that these options mostly resulted in broken dlint configs.

I've amended the message to mention that.

PR #8512, opened 2021-01-14
LF: fix Transaction Decoder/Encoder

remyhaemmerle-da:
This includes the following changes: 

1. The Decoder was not verifying the value inside a node use the exact
  same version as the node itself.

2. We use the node version directly in the node encoder (instead of
  used `versionedX` node methods).

3. We avoid used of `.asJava` in the encoder.

4. We add test for testing the decode verify 1.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
@@ -60,19 +60,27 @@ object TransactionCoder {
           )
     }
 
-  def encodeValue[Cid](
+  def encodeVersionedValue[Cid](
       cidEncoder: ValueCoder.EncodeCid[Cid],
-      value: VersionedValue[Cid],
+      nodeVersion: TransactionVersion,
+      value: Value[Cid],
   ): Either[EncodeError, ValueOuterClass.VersionedValue] =
-    ValueCoder.encodeVersionedValue(cidEncoder, value)
+    ValueCoder.encodeVersionedValue(cidEncoder, nodeVersion, value)
 
-  def decodeValue[Cid](
+  private[this] def decodeValue[Cid](
       cidDecoder: ValueCoder.DecodeCid[Cid],
+      nodeVersion: TransactionVersion,
       value: ValueOuterClass.VersionedValue,
   ): Either[DecodeError, Value[Cid]] =
-    for {
-      versionedValue <- ValueCoder.decodeVersionedValue(cidDecoder, value)
-    } yield versionedValue.value
+    ValueCoder.decodeVersionedValue(cidDecoder, value).flatMap {
+      case VersionedValue(`nodeVersion`, value) => Right(value)
+      case VersionedValue(version, _) =>
+        Left(
+          DecodeError(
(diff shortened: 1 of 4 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -115,17 +141,17 @@ object TransactionCoder {
...
-      TransactionOuterClass.KeyWithMaintainers
-        .newBuilder()
-        .setKey(_)
-        .addAllMaintainers(key.maintainers.toSet[String].asJava)
-        .build()
-    )
+    encodeVersionedValue(encodeCid, nodeVersion, key.key).map { encodedKeyValue =>
+      val builder = TransactionOuterClass.KeyWithMaintainers.newBuilder()
+      builder.setKey(encodedKeyValue)
+      key.maintainers.foreach(builder.addMaintainers)
```
cocreature, line 152: :+1:

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -340,6 +340,161 @@ class TransactionCoderSpec
       }
     }
 
+    def changeVersion(
+        value: com.daml.lf.value.ValueOuterClass.VersionedValue,
+        version: String,
+    ) =
+      value.toBuilder.setVersion(version).build()
+
+    "fail if try to encode a create node containing value with version different from node" in {
```
cocreature, line 349: This sounds like we fail during encoding but that seems to work fine. What fails is decoding. Should we fail to encode such values already?

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -340,6 +340,161 @@ class TransactionCoderSpec
...
+              ValueCoder.CidDecoder,
+              nodeVersion,
+              encodedNode.toBuilder.setCreate(node).build(),
+            ) shouldBe a[Left[_, _]]
+          )
+        }
+      }
+    }
+
+    "fail if try to encode a fetch node containing value with version different from node" in {
```
cocreature, line 391: same here, we also only fail during decoding

cocreature (approved):
Good catch! My only question is if we should already fail to encode on version mismatches instead of only failing during decoding.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
remyhaemmerle-da: In Scala land the Value in GenNode are not wrapped in VersionedValue any more, so we cannot generate "such" value.

cocreature: ah I see, makes sense.

PR #8513, opened 2021-01-14
Update both hidden.json and snapshots.json

cocreature:
The assistant relies on the latter, our docs cronjob on the former. I
have no idea why we have two but keeping them in sync should be fine.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -136,13 +136,15 @@ fetch_if_missing opts temp v = do
 
 update_s3 :: DocOptions -> FilePath -> Versions -> IO ()
 update_s3 opts temp vs = do
-    putStrLn "Updating versions.json, hidden.json, latest..."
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):

garyverhaegen-da:
For the record: this was meant as a private file for the internal processes of the docs cron. Arguably storing it on the website was not the best choice; perhaps somewhere on a GCS bucket would have been better.

It was created as snapshots.json, then I decided to change the name to hidden.json because it sometimes lists non-snapshot prereleases. Both existed for a while because it was necessary to have some overlap for this script not to be broken. That should have been pretty much just running this script once but then I forgot about it so we had both files for a while.

I'm not sure when the assistant started depending on it, but it really shouldn't. I guess that ship has sailed, though, if we don't want to break existing installations.

PR #8514, opened 2021-01-14
docs cron: update meta files unconditionally

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -463,4 +463,9 @@ main = do
       Docs -> do
           docs sdkDocOpts
           docs damlOnSqlDocOpts
+          -- FIXME: this is meant to run once to create the `latest` file
+          -- without waiting for the next stable release.
+          gh_versions <- fetch_gh_versions (includedVersion sdkDocOpts)
+          IO.withTempDir $ \temp_dir -> do
+              update_s3 sdkDocOpts temp_dir gh_versions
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
Obviously the comment is garbage; I just reverted #8385.

ci/cron/src/Main.hs
```diff
@@ -463,4 +463,9 @@ main = do
       Docs -> do
           docs sdkDocOpts
           docs damlOnSqlDocOpts
+          -- FIXME: this is meant to run once to create the `latest` file
```
cocreature, line 466: the comment sounds wrong, we want to update `snapshots.json`.

cocreature: Also not sure if at this point it isn’t just better to keep this code long term.

cocreature (approved):
Thanks!

ci/cron/src/Main.hs
garyverhaegen-da: I'll merge as is to fix the issue tonight; I'll think a bit more about what to do longer-term.

cocreature: :+1:

garyverhaegen-da:
This does not seem to have worked 🤔

garyverhaegen-da:
Caching was getting in the way, the situation is now resolved.

PR #8515, opened 2021-01-14
ci: clean-up disk cache if it gets too big

garyverhaegen-da:
Hopefully this works around our recent CI disk space issues, while 80GB should be large enough that it only happens once per machine per day, so perf shouldn't be impacted too much.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

garyverhaegen-da:
This won't work on Linux nodes, which are starting to fill up too.

PR #8516, opened 2021-01-14
non-empty newtypes

### LINKED ISSUE #8507: non-empty variant of Set, Map
> Did you consider a `OneAnd` equivalent for maps? I guess it’s probably not worth it given that this is relatively local. _@cocreature in https://github.com/digital-asset/daml/pull/8226#discussion_r557176893_

Set and Map not being covariant in all their data makes OneAnd slightly awkward (albeit not fatally so) for use with them, as OneAnd's operations are best suited to traversable functors. It might be worth trying a newtype on Set/Map that enforces non-emptiness, with appropriately shifted types for the operations, rather than the structural enforcement that OneAnd entails. We could unfortunately not tie MapOps/MapLike knots in them, because they contain emptying operations.

S11001001:
Fixes #8507. Includes some (very small!) changes to `uniqueSets` in the json-api to demonstrate how non-emptiness proof is preserved from one iteration to the next, replacing the assumption of `+:` with the total/exhaustive `+-:` (operators suggested by @SamirTalwar-DA `±:` and `:∓` also included in this PR).

### Pull Request Checklist

trigger the build.


ledger-service/db-backend/src/test/scala/com/digitalasset/http/dbbackend/QueriesSpec.scala
```diff
@@ -6,6 +6,7 @@ package com.daml.http.dbbackend
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
+import scala.collection.immutable.Seq
 
 class QueriesSpec extends AnyWordSpec with Matchers with ScalaCheckDrivenPropertyChecks {
   "uniqueSets" should {
```
SamirTalwar-DA, line 12: I'd really appreciate a couple of example-based tests cases for `uniqueSets` so I can see what it actually does. It puzzles me.

SamirTalwar-DA (commented):
Could you give an example of a bug this would catch?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -354,18 +358,22 @@ object Queries {
...
+      case NonEmpty(m) =>
+        Some {
+          val hd = m transform { (_, abs) =>
+            val (_, b) +-: _ = abs
+            b
+          }
+          val tl = m collect { case (a, _ +-: NonEmpty(tl)) => (a, tl) }
+          (hd, tl)
         }
-        val tl = m collect { case (a, _ +: (tl @ (_ +: _))) => (a, tl) }
```
S11001001, line 366: An example bug for @SamirTalwar-DA: imagine that I had forgotten to filter out the empty `tl`s here; they would then leak into the next iteration and crash where the assumption is made on old line 363. Embedding the `NonEmpty` type into the loop state requires me to do this check and filtering instead of just calling `tail`, so the assumption on new line 370 can be made in perfect safety.

SamirTalwar-DA: Nice, I like it.

libs-scala/scala-utils/src/test/scala/com/daml/scalautil/nonempty/NonEmptySpec.scala
```diff
@@ -0,0 +1,110 @@
...
+      g: Set[imm.Map[Int, NonEmpty[imm.Set[Int]]]]
+    }
+
+    "produce Seqs for Seqs" in {
+      val g = Set(imm.Seq(1) groupBy1 identity)
+      g: Set[imm.Map[Int, NonEmpty[imm.Seq[Int]]]]
+    }
+
+    "reject maybe-mutable structures" in {
+      illTyped(
```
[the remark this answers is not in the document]
S11001001: It has some limitations, because bounds-checking and warnings-as-errors don't apply (it doesn't go far enough in typechecking phases to see those), but is still very valuable.

ledger-service/db-backend/src/test/scala/com/digitalasset/http/dbbackend/QueriesSpec.scala
S11001001: Maybe a quick colloquial would be even faster to understand: groupBy, but all the first values, then all the second values, then all the 3rd values...

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/nonempty/package.scala
```diff
@@ -0,0 +1,39 @@
...
+    * parameters `C[T]`, `toF` will infer a `NonEmptyF[C, T]` for whatever
+    * values of C and T scalac chooses.  For example, if `bar: NonEmpty[Map[K,
+    * V]]`, then `bar.toF: NonEmptyF[Map[K, *], V]`, because that's how scalac
+    * destructures that type.
+    */
+  type NonEmptyF[F[_], A] = NonEmpty.NonEmptyF[F, A]
+
+  // aliases for Samir
+  val ±: : +-:.type = +-:
+  val :∓ : :-+.type = :-+
```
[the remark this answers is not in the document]
S11001001: The singleton type facilities at the heart of Scala's type relation model were defined so many years ago so that one day we could reach this pinnacle: `±:` made trivially easy.

cocreature (approved):
Nice!

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/nonempty/NonEmpty.scala
```diff
@@ -0,0 +1,141 @@
...
+    * immutable type.  So [[scala.collection.Seq]] will not work.
+    *
+    * The type-checker will not permit you to apply this to a value that already
+    * has the [[NonEmpty]] type, so don't worry about redundant checks here.
+    */
+  def unapply[Self](self: Self with imm.Iterable[_]): Option[NonEmpty[Self]] = apply(self)
+}
+
+/** If you ever have to import [[NonEmptyColl]] or anything from it, your Scala
+  * settings are probably wrong.
```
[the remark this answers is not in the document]
S11001001: They are mentioned [elsewhere in the PR](https://github.com/digital-asset/daml/pull/8516/files#diff-9bdf34cca7a920c088625f09fc889e938dff0d8126b8c1fd9ff1efba075adca3R28-R29).

PR #8525, opened 2021-01-15
ci/cron: remove hidden.json

garyverhaegen-da:
I discovered yesterday that the `snapshots.json` (and actually also the `versions.json`) file is no longer purely internal to the docs process, as it was meant to be, but is now depended upon by the assistant. This means the renaming from `snapshots.json` to `hidden.json` cannot happen, and we reverted that yesterday in #8513 (& #8514), though that was done in a bit of a hurry. This PR aims at cleaning up the resulting mess and achieve a better long-term end state.

I will be manually removing the `hidden.json` file as soon as this is merged, so that nothing ends up depending on _that_. There is no occurrence of `hidden.json` outside this docs cron so hopefully this works out.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -136,29 +136,25 @@ fetch_if_missing opts temp v = do
 
 update_s3 :: DocOptions -> FilePath -> Versions -> IO ()
 update_s3 opts temp vs = do
-    putStrLn "Updating versions.json, hidden.json, snapshots.json, latest..."
-    create_versions_json (dropdown vs) (temp </> "versions.json")
-    let hidden = Data.List.sortOn Data.Ord.Down $ Set.toList $ all_versions vs `Set.difference` (Set.fromList $ dropdown vs)
-    let push f = proc_ ["aws", "s3", "cp", temp </> f, s3Path opts f, "--acl", "public-read"]
-    let hiddenVersionFiles = ["hidden.json", "snapshots.json"]
-    Data.Foldable.for_ hiddenVersionFiles $ \f -> do
-        create_versions_json hidden (temp </> f)
-        push f
-    push "versions.json"
-    Control.Monad.Extra.whenJust (top vs) $ \latest -> do
-        writeFile (temp </> "latest") (show latest)
-        push "latest"
+    let displayed = dropdown vs
+    let hidden = Data.List.sortOn Data.Ord.Down $ Set.toList $ all_versions vs `Set.difference` Set.fromList displayed
+    -- The assistant depends on these three files, they are not just internal
+    -- to the docs process.
+    push (versions_json displayed) "versions.json"
+    push (versions_json hidden) "snapshots.json"
+    Control.Monad.Extra.whenJust (top vs) $ \latest -> push (show latest) "latest"
     putStrLn "Done."
     where
-        create_versions_json versions file = do
-            -- Not going through Aeson because it represents JSON objects as
-            -- unordered maps, and here order matters.
-            let versions_json = versions
-                                & map show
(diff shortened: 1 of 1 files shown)


ci/cron/src/Main.hs
```diff
@@ -245,13 +241,11 @@ fetch_gh_versions pred = do
 fetch_s3_versions :: DocOptions -> IO Versions
 fetch_s3_versions opts = do
     -- On the first run, this will fail so treat it like an empty file.
-    -- We could technically remove the catch later.
```
garyverhaegen-da, line 248: Keeping this on the off chance we'd need to reinitialize the docs bucket from scratch some day.

ci/cron/src/Main.hs
```diff
@@ -463,9 +457,4 @@ main = do
       Docs -> do
           docs sdkDocOpts
           docs damlOnSqlDocOpts
-          -- FIXME: this is meant to run once to create the `latest` file
-          -- without waiting for the next stable release.
-          gh_versions <- fetch_gh_versions (includedVersion sdkDocOpts)
```
garyverhaegen-da, line 468: I've decided against keeping this long-term, because the only case in which this code makes sense is when renaming files, and I'm not doing that again.

ci/cron/src/Main.hs
```diff
@@ -245,13 +241,11 @@ fetch_gh_versions pred = do
...
-    -- We could technically remove the catch later.
-    dropdown <- fetch "versions.json" False `catchIO`
-      (\_ -> pure [])
-    hidden <- fetch "hidden.json" True `catchIO`
-      (\_ -> pure [])
+    dropdown <- fetch "versions.json" False `or` []
+    hidden <- fetch "snapshots.json" True `or` []
     return $ versions $ dropdown <> hidden
-    where fetch file prerelease = do
+    where or io def = catchIO io (\_ -> pure def)
```
cocreature, line 247: not a huge fan of hiding names defined in `prelude` but I don’t have a great suggestion for another name. Maybe just keep it inlined?

ci/cron/src/Main.hs
cocreature: Don’t we also need this if we add a new file? Not that I have anything in mind that we might want to add but `latest` has been pretty recent so it seems at least feasible we’ll need this again. Of course, we can always add it once and remove it afterwards in that case but the "add it once and remove it again" stuff has been required quite often now and we’ve messed it up occasionally and had to add it in a separate PR. Is there a downside to keeping this apart from the fact that we push it even if nothing changed (which should be cheap)?

cocreature (approved):
Thanks!

ci/cron/src/Main.hs
```diff
@@ -136,29 +136,25 @@ fetch_if_missing opts temp v = do
 
 update_s3 :: DocOptions -> FilePath -> Versions -> IO ()
 update_s3 opts temp vs = do
-    putStrLn "Updating versions.json, hidden.json, snapshots.json, latest..."
```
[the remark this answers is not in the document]
garyverhaegen-da: Yes, because we already print pretty much the exact same message just before calling the function.

ci/cron/src/Main.hs
garyverhaegen-da: Because the rest of the process relies on reading those files to figure out if it needs to build something, decoupling writing those files from the rest of the process seems dangerous.

Yes, we might want to do this when we add a file, like we did for `latest`, but that's only if we don't want to wait for the next release to update the files. I'll think a bit more about this.

PR #8530, opened 2021-01-15
Update LF spec for exceptions.

sofiafaro-da:
The LF spec was out of date compared to our current design for exceptions (which wasn't written down anywhere, but I'm trying to fix that). This PR tries to bring the spec up to date.

The biggest change from the last version of the spec is that we no longer throw `AnyException`. Instead we throw values of an exception type directly. The `AnyException` type is relegated to wrapping/unwrapping exception values but still plays an important role in defining `try catch` blocks. `AnyException` is no longer serializable (this could be changed in the future if we want). One cannot throw an `AnyException` anymore, one can only throw exception values at known exception types (this could also be changed in the future if we want, by adding a "THROW_ANY_EXCEPTION" built-in).

At first this was a small change, but then I realized we needed the original type of the exception in the semantics of try-catch blocks. This resulted in attaching the type of the exception to the operational semantics of exception throwing & propagation. In practice, the exception type can be inferred from the value, since exception types are very restrictive.

I also added the notion of "incomplete ledger action" and "incomplete transaction" to the LF spec, based on the chat we (Nick, Rémy, & me) had about the transaction tree for exceptions. Incomplete transactions are transactions that may optionally have an incomplete exercise node at the end. They only show up in rollback nodes or other incomplete exercise nodes.


daml-lf/spec/daml-lf-1.rst
```diff
@@ -2728,12 +2742,21 @@ as described by the ledger model::
...
 
+  Incomplete ledger actions
+    iact
+      ::= 'iexercise' v Contract ChKind itr  -- v must be of type 'List' 'Party'
+
+  Incomplete ledger transactions
+    itr
+      ::= tr
+       |  tr · iact
+
```
cocreature, line 2759: I don’t quite understand why this definition is setup the way it is. Rollback makes sense and that’s what I would expect. The part that I don’t understand is incomplete ledger transactions:

The only difference between `itr` and `tr` is that the former special cases incomplete exercises and allows those at the end (however, regular exercises are also still allowed at the end). I assume we get a regular transaction if we executed a choice body and we ran some successful actions and then failed outside of a create or exercise, e.g., a division by zero in the choice body. We get an `iexercise` if a choice called from the choice body fails. But why don’t we have an `icreate` for cases where we call create from the choice body and the precondition fails? It looks like we just treat this like the precondition failed in the choice body outside of the create? That seems somewhat unintuitive to me and I’d like to understand the motivation for this choice. Overall, I think we need some human-readable comments that explain how things are represented here. Doesn’t have to be in the spec but arguably the chances of keeping it updated here are higher than in a google doc.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3102,79 +3134,83 @@ as described by the ledger model::
...
+     'try' @τ e₁ 'catch' x. e₂ ‖ S₀
+       ⇓ᵤ
+     (Ok v₁, tr₁) ‖ S₁
 
-     e₁  ⇓  Err v₁
-     e₂[x ↦ v₁]  ⇓  Ok ('None' @σ)
-   —————————————————————————————————————————————————————————————————————— EvUpdTryCatchErr2e
-     'try' @τ e₁ 'catch' x. e₂ ‖ S₀  ⇓ᵤ  (Err v₁, ε)
+     e₁  ⇓  Err @τ₁ v₁
+     e₂[x ↦ 'to_any_exception' @τ₁ v₁]  ⇓  Ok ('None' @σ)
```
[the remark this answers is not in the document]
sofiafaro-da: Yes!

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: That's right: In the execution of a "try" you can run into exceptions that aren't in the body of an exercise, so that all the ledger actions under a rollback node are "complete", even though they will be rolled back.  *You only get an incomplete ledger action when an exception propagates up from a choice body (`iexercise`).*

We could have an `icreate`. I don't have a strong opinion on this, but there's maybe some reasons not to: it's not needed for validation of rolled back transactions (unlike `exercise`/`iexercise` nodes), and the contract value you could attach to it isn't well formed (otherwise the create would have succeeded). Note that we only generate an `iexercise` node when we are evaluating or executing a choice body. An ill-formed exercise (e.g. exercise on inactive contract, see EvUpdExercInactive) does not result in an `iexercise` node, only in an exception.

The other thing we'll probably want at some point is `fetch` nodes in the spec (... and maybe `ifetch` nodes?) I'm going to talk with Rémy about this soon.

cocreature: I’m still struggling to understand what exactly makes exercise special here. What information is in `iexercise` that we need for validation that doesn’t apply to create? Is it the `contract` because it does not just contain the cid which we could get by reinterpretation but also the contract argument which we need to interpret the body of the choice whereas for a create, we have the contract argument directly in the DAML-LF call so we can interpret the `ensure` clause without having the `create` in the transaction? That would make sense to me but then it seems like `fetch` runs into exactly the same issue even without involving any exceptions so we definitely need to fix this. The current definition suffers from the same issue as the transaction tree you get on the ledger API which is not sufficient for validation exactly because it is missing fetch nodes (they’re not quite equivalent, the one on the ledger API is also missing the contract argument on exercises).

Assuming it is about the contract argument, I have one last question: We cannot omit `iexercise` like we can for `create` since we need the contract argument but what happens if we just use an `exercise`? That loses the information of whether we failed within the body of the choice we called or within the caller but as far as I can tell that information is redundant and can be recovered during validation (reinterpret and see where it blows up). It seems perfectly sensible to encode this separately even if we could recover it, I’m just trying to figure out which info is strictly needed and which is “only” to make the spec clearer.

cocreature: After a brief chat with @remyhaemmerle-da, I think the current specification for transactions is neither very useful nor correct. I would suggest to not spend more time on this for now (we can merge this state or just omit it completely) and instead update the ledger model to include exceptions first to clarify what exactly we want and then we can later come back to the LF spec and see how we can fix the spec here and make it actually useful.

cocreature (approved):
Thank you! As mentioned in the comment, I would suggest to not spend more time on transactions in the LF spec for now and instead update the ledger model.

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: I think the difference between exercise and choice here is that the exercise body needs to be executed with a  different authorization context, so we need to track exercises carefully in the failed transaction, and the difference between exercise and iexercise is that the former returns a value from the exercise but the latter doesn't. Neither of these (authorization contexts, and values in exercises) are being tracked in the current model for the semantics, so it's definitely something to clarify/improve at some point ...

cocreature: Yeah I think I get why we need the exercise and in a full transaction, I get that at the very least the choice result is a difference between a complete and an incomplete exercise (although I’m not quite clear if you couldn’t just recompute that during validation but that’s a separate discussion). What confused me was why we can omit some information (e.g., fetches, choice results, authorization contexts), but not other information. After this discussion, the conclusion seems to be that we probably cannot omit any of this and the current transaction spec is just plain wrong and neither describes the ledger API nor the transaction we use for validation.

PR #8533, opened 2021-01-15
Release 1.8.1-snapshot.20210115.5842.0.59f5d407

fabiotudone-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


gerolf-da (approved):

LATEST
```diff
@@ -1,5 +1,6 @@
 9ed787cb3dadd2d6d79f77a2ee93c576f8b8ac4c 1.9.0-snapshot.20210113.6060.0.9ed787cb
 a8ae8e4aa782b8949d32aec3ae05a3211a2306f7 1.8.0-snapshot.20201214.5841.0.a8ae8e4a
```
cocreature, line 2: Can we replace this 1.8.0 snapshot? I forgot what it was for but having to 1.8.0, 1.8.0 snapshot and 1.8.1 snapshot seems off.

fabiotudone-da: I don' think we can; @gerolf-da?

cocreature: ah right that was the fast but broken version. Alright, then let’s keep it. There is nothing stopping us from deleting that line later once vmbc doesn’t need it anymore.

fabiotudone-da:
Manual tests passed on Windows.

fabiotudone-da:
Manual tests passed on Linux.

PR #8535, opened 2021-01-18
Improve release/RELEASE.md

fabiotudone-da:
## Description

Adding some tips and notes about specific steps that may be obscure for first-timers.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -33,7 +33,8 @@ patches we backport to the 1.0 release branch).
    ```
    Then open a PR _to be merged to `main`_ (even if it's for a maintenance release)
    with the changed `LATEST` file, add the line produced by the `release.sh`
-   invocation in a meaningful position and add the `Standard-Change` label. It
+   invocation in a meaningful position (if you’re not sure, [semver](https://semver.org/) ordering is
+   probably the right thing to do) and add the `Standard-Change` label. It
    is better to add such a label _before confirming the PR's creation_, else
    the associated CI check will fail and merging the PR will require you to
    re-run it after all the other ones have completed successfully.
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -31,7 +31,7 @@ patches we backport to the 1.0 release branch).
    $ ./release.sh snapshot cc880e2 0.1.2
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR _to be merged in `main`_ (even if it's for a maintenance release)
+   Then open a PR _to be merged to `main`_ (even if it's for a maintenance release)
    with the changed `LATEST` file, add the line produced by the `release.sh`
    invocation in a meaningful position and add the `Standard-Change` label. It
    is better to add such a label _before confirming the PR's creation_, else
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -22,7 +22,7 @@ patches we backport to the 1.0 release branch).
    (`#team-daml`) if you're not sure.
 
    If you are manually creating the PR for an out-of-schedule snapshot, start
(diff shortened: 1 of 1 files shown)


release/RELEASE.md
```diff
@@ -31,8 +31,12 @@ patches we backport to the 1.0 release branch).
    $ ./release.sh snapshot cc880e2 0.1.2
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR with the changed `LATEST` file, and add the `Standard-Change`
-   label.
+   Then open a PR _to be merged in `main`_ (even if it's for a maintenance release)
+   with the changed `LATEST` file, add the line produced by the `release.sh`
+   invocation in a meaningful position and add the `Standard-Change` label. It
```
cocreature, line 36: The extra warning here isn’t actually that important. You can rerun only that check and it finishes in less than a minute.

cocreature: still nice to add it initially though so you can have mergify take over, so no objection to add it.

cocreature (approved):
Thank you!

release/RELEASE.md
fabiotudone-da: I'd leave it because AZP forces you to wait for all the other checks to complete before re-running it, which is one more annoying manual-step-after-waiting.

garyverhaegen-da (approved):
Thanks!

release/RELEASE.md
```diff
@@ -31,8 +31,12 @@ patches we backport to the 1.0 release branch).
    $ ./release.sh snapshot cc880e2 0.1.2
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR with the changed `LATEST` file, and add the `Standard-Change`
-   label.
+   Then open a PR _to be merged in `main`_ (even if it's for a maintenance release)
```
miklos-da, line 34:
```suggestion
   Then open a PR _to be merged to `main`_ (even if it's for a maintenance release)
```

release/RELEASE.md
miklos-da, line 36: Please elaborate here what 'meaningful position` means, i.e., you should keep the list of versions sorted from newest to oldest.

miklos-da (approved):
Thanks for improving the release instructions!

release/RELEASE.md
fabiotudone-da: I actually left it vague on purpose (and I am appealing to common sense) because version order and time order can be (and are now already) different and mandating one over the other can be confusing IMO, also considering that code changes are not linear across branches.

@garyverhaegen-da @moritzkiefer-da WDYT?

cocreature: in principle semver ordering would be the right thing to do but that falls apart once we have weird things like the 1.8 snapshot so maybe state something like "if you’re not sure, semver ordering is probably the right thing to do".

garyverhaegen-da: I agree on semver ordering. Everything should order nicely by semver, as that's why I designed the version numbers that way. Unless we somehow gave a random version number to an official release at some point 🤷

garyverhaegen-da: Ah right I see what you mean. `1.8.0-snapshot.20201214.5841.0.a8ae8e4a` is garbage.  Not sure how to handle that; I'd go for renaming it to 1.8.1 and renaming the current breed of new-1.8 1.8.2. I don't think it's a big issue that 1.8.2 is not strictly built on top of 1.8.1. Another option would be to call it something like `1.8.0-garbage`, but I'm not 100% confident we don't have some code somewhere that expects exactly `-snapshot`, so I'd have to check for that first.

PR #8541, opened 2021-01-18
damlc: incremental package db initialization

robin-da:
### Pull Request Checklist

trigger the build.


robin-da:
Where's the proper place to add a test for this?

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -70,120 +72,135 @@ import SdkVersion
...
+
+    -- Read the metadata of an already existing package database and see if wee need to reinitialize.
+    let metadataFileFp = metadataFile projectRoot
+    hasMetadata <- doesFileExist metadataFileFp
+    needsReinitalization <- if hasMetadata
+                            then do
+                              metaData <- readMetadata projectRoot
+                              pure $ hashDependencies metaData /= Just depHash
+                            else pure True
+    when needsReinitalization $ do
```
cocreature, line 96: This is growing quite long, maybe a good idea to factor out the check as well as the body of the `when`, so this just turns into something like
```
createProjectPackageDb … = do
  needsReinitialization <- getNeedsReinitialization
  when needsReinitialization $ initializeDb
```

Fine to leave that for a separate PR.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -70,120 +72,135 @@ import SdkVersion
...
+        foldM
+            (\acc dep -> do
+                 bs <- BSL.readFile dep
+                 pure $ encode $ hash $ bs <> acc)
+            BSL.empty
+            (deps ++ dataDeps)
+
+    -- Read the metadata of an already existing package database and see if wee need to reinitialize.
+    let metadataFileFp = metadataFile projectRoot
+    hasMetadata <- doesFileExist metadataFileFp
```
cocreature, line 90: `doesFileExist` is almost always racy, I recomend reading it and catching the exception instead. Probably not a big issue here but I’d rather go for the safe version everywhere instead of wondering where the exists check is safe.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -70,120 +72,135 @@ import SdkVersion
...
-    writeMetadata projectRoot (PackageDbMetadata (mainUnitIds dependencyInfo) validatedModulePrefixes)
+createProjectPackageDb projectRoot (disableScenarioService -> opts) (PackageSdkVersion thisSdkVer) modulePrefixes deps' dataDeps
+  = do
+    -- compute the hash over all dependencies
+    deps <- expandSdkPackages (optDamlLfVersion opts) (filter (`notElem` basePackages) deps')
+    depHash <-
+        fmap hash $
+        foldM
+            (\acc dep -> do
+                 bs <- BSL.readFile dep
```
cocreature, line 83: Can we use `https://hackage.haskell.org/package/ghc-8.10.2/docs/Fingerprint.html#v:getFileHash` instead? Afaict, that is specifically written for this type of task and is decently fast and constant memory without depending on lazy IO. We use it in other places already (at least indirectly via ghcide).

compiler/damlc/daml-opts/daml-opts-types/DA/Daml/Options/Packaging/Metadata.hs
```diff
@@ -40,6 +38,9 @@ data PackageDbMetadata = PackageDbMetadata
   -- We do not bother differentiating between exposed and unexposed modules
   -- since we already warn on non-exposed modules anyway and this
   -- is intended for data-dependencies where everything is exposed.
+  , hashDependencies :: Maybe Int
```
cocreature, line 41: Why is this a `Maybe`? Afaict, we always set it to a `Just`.

cocreature (commented):
Very nice! This turned out much simpler than I expected. A few minor comments and I’ll take another look once tests are there.

As for tests, I think I’d probably just add a new test suite. In principle, it could go in the packaging tests but those are already really slow and we had to spend some time making them faster recently so I’m hesitant to add it there.

I think there are 4 different tests we should have:

1. Dependency changed and the project depends on that change (e.g., addition of a new function).
2. Dependency added and the project depends on it.
3. Dependency removed as a regression test.
4. Nothing changed, check that modification time hasn’t changed. https://github.com/digital-asset/daml/blob/441f503a95c936832012d9d3c68823b61063a469/compiler/damlc/tests/src/DA/Test/IncrementalBuilds.hs does something similar for `--incremental=yes`.

compiler/damlc/daml-opts/daml-opts-types/DA/Daml/Options/Packaging/Metadata.hs
robin-da: I made it a `Maybe` because otherwise, you'll get an ugly JSON-parsing error if you run the newest SDK version in a project with an already existing package database and metadata file.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: I'll factor it out.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: Sure, I'll use this one instead.

compiler/damlc/daml-opts/daml-opts-types/DA/Daml/Options/Packaging/Metadata.hs
robin-da: But if we catch the exception in `readMetadata`, we can also drop the `Maybe` and reinitialize and rewrite the metadata when it fails reading the existing metadata.

cocreature: This actually brings up a good point. We need to include the SDK version and the daml-lf version in the hash as well. Catching the exceptions and reinitializing sounds good.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: now catching the exception.

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
```diff
@@ -0,0 +1,273 @@
...
+                , unlines
+                      [ "sdk-version: 0.0.0"
+                      , "name: proj"
+                      , "source: daml"
+                      , "version: 0.0.1"
+                      , "dependencies:"
+                      , "  - daml-prim"
+                      , "  - daml-stdlib"
+                      , "  - dependency/dep.dar"
+                      ])
```
cocreature, line 110: maybe reference something from `B` to make sure everything really worked correctly and we didn’t just write the dep again?

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
```diff
@@ -0,0 +1,273 @@
...
+              ]
+              [ ( "daml/B.daml"
+                , unlines
+                      [ "module B where"
+                      , "template T"
+                      , "  with"
+                      , "    p: Party"
+                      , "  where"
+                      , "    signatory p"
+                      , "f : Int -> Int"
```
cocreature, line 80: Might be nice to make `A` depend on the change in `B`, e.g., also modify it to reference a new definition that we add in `B`.

cocreature (approved):
Awesome, thank you! Very excited that this long-standing bug is finally fixed.

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
```diff
@@ -0,0 +1,273 @@
...
+                , unlines
+                      [ "module A where"
+                      , "import B"
+                      , "x : Int"
+                      , "x = f 1"])
+              ]
+              []
+              [ ( "daml.yaml"
+                , unlines
+                      [ "sdk-version: 0.0.0"
```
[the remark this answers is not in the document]
robin-da: right, I keep forgetting about this. it's fixed now.

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
robin-da: added a reference.

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
robin-da: added an additional reference.

compiler/damlc/tests/src/DA/Test/IncrementalPackageDb.hs
```diff
@@ -0,0 +1,273 @@
...
+    defaultMain $ tests damlc
+
+tests :: FilePath -> TestTree
+tests damlc =
+    testGroup
+        "Incremental package db initialization"
+        [ test
+              "Reinitialize when dependency changes"
+              [ ( "daml.yaml"
+                , unlines
```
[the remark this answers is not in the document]
robin-da: factored it out.

PR #8544, opened 2021-01-19
add sudo capability to Docker image

richardkapolnai-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/docker/daml-sdk/Dockerfile
+++ b/ci/docker/daml-sdk/Dockerfile
@@ -5,6 +5,12 @@ ARG VERSION
 # Otherwise they fail to even resolve localhost.
 RUN echo 'hosts: files dns' > /etc/nsswitch.conf
 RUN addgroup --system daml && adduser --system --ingroup daml daml
+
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
I realize this is a draft, so you may already be planning to do that, but for this kind of change it would be nice to include a bit more context on motivation. My first reaction here would be that if people want to run things as root (presumably install stuff), they should instead build their own image with this one as a base (so they get caching on those installs).

Or do you have a use-case for running Daml components themselves as root?

richardkapolnai-da:
> I realize this is a draft, so you may already be planning to do that, but for this kind of change it would be nice to include a bit more context on motivation. My first reaction here would be that if people want to run things as root (presumably install stuff), they should instead build their own image with this one as a base (so they get caching on those installs).
> 
> Or do you have a use-case for running Daml components themselves as root?

You're right, my intention was to be able to install stuff on a CI job to this image.
Building another image makes sense, closing this PR.

PR #8374, opened 2021-01-04
separate CI config for main & prs

garyverhaegen-da:
This is another take on #8276, with the same underlying motivation. However, this approach is mostly duplication-free, which seems better, especially given the already-pretty-sorry state of our CI config.

CHANGELOG_BEGIN
CHANGELOG_END

PR #8549, opened 2021-01-19
Backport ci changes 1.8.x

garyverhaegen-da:
This updates the config to match #8374 & the corresponding new Azure setup.

It's essentially the same as #8548, but on 1.8.x.

Note: we do not have a release branch for 1.7. If we ever need to patch 1.7.0, it will need a corresponding update, but I don't think it's worth creating the branch branches just for this change. (I.e. I'm hoping we never need to patch 1.7.0.)

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,23 +3,11 @@
 
 # Azure Pipelines file, see https://aka.ms/yaml
 
-# Enable builds on all branches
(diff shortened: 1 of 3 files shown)


cocreature (approved):
Thanks for pushing through all of those!

PR #8551, opened 2021-01-19
Moved default timeout definition up in callstack

daravep:
The default timeout of 250ms turned out to be problematic for Canton.
Therefore, I'm moving the hard coded 250ms up such that we can override
it from within the Canton code. Other users shouldn't be affected by the
change. Therefore, this is just an internal code change.

Additionally, I've added two new performance targets for the performance
envelope test.

Related to https://github.com/DACH-NY/canton/issues/4996

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


SamirTalwar-DA (approved):
Looks good to me.

PR #7945, opened 2020-11-11
participant-integration-api: Dedicated execution context for requests.

SamirTalwar-DA:
Found this branch kicking around on my machine during some cleanup and decided to get it working again.

This is similar (but more extensive) to similar changes made in Canton.

My hypothesis is that by creating a dedicated execution context to handle requests (but ideally not the computation associated with them), we have more control over our runtime.

Unfortunately, I don't have a decent way to profile this. The ledger-on-memory conformance tests are about 5% faster on my machine, but this is probably within error margins.

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8555, opened 2021-01-19
Sandbox-classic: revert back the change of default contract ID seeding

remyhaemmerle-da:
We revert the change of default contract ID seeding that was
introduced in #7945.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
+++ b/language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
@@ -48,7 +48,6 @@ object TestUtil {
   )(implicit resourceContext: ResourceContext): Future[Assertion] = {
     val config = sandbox.DefaultConfig.copy(
       port = Port.Dynamic,
-      seeding = None,
       damlPackages = List(testDalf),
       ledgerIdMode = LedgerIdMode.Static(LedgerId(LedgerID)),
       timeProviderType = Some(TimeProviderType.WallClock),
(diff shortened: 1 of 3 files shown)


language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
```diff
@@ -48,7 +48,6 @@ object TestUtil {
   )(implicit resourceContext: ResourceContext): Future[Assertion] = {
     val config = sandbox.DefaultConfig.copy(
       port = Port.Dynamic,
-      seeding = None,
```
cocreature, line 51: this seems independent of this right?

cocreature (approved):
Thanks for fixing this so quickly!

SamirTalwar-DA (approved):

SamirTalwar-DA:
For context: I think I broke this because I thought I was changing test code. I did not intend to change the behavior of Sandbox in production, only in testing.

language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
remyhaemmerle-da: I revert the two last commits of #7945. This is the last one that overrides the `seeding = Some(Seeding.Weak)`  from the culprit commit. I think it is fair to revert this as well.

cocreature: Makes sense :+1:

PR #8556, opened 2021-01-19
ci/unix: better cache size reporting

garyverhaegen-da:
Current reports look like:

```
Disk cache small enough:\n20G/home/vsts/.bazel-cache
```

because `echo` does not convert `\n`. An alternative would be to replace `echo` with `printf`, but I have had enough issues with subshells-in-strings lately that I prefer just avoiding them when possible.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build-unix.yml
+++ b/ci/build-unix.yml
@@ -38,7 +38,8 @@ steps:
           echo "Deleting '$disk_cache'..."
           rm -rf "$disk_cache"
       else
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

SamirTalwar-DA (approved):
Much more pleasant.

FYI, I believe you can use `echo -e`, but I still prefer this solution.

PR #8562, opened 2021-01-19
LF: Release LF 1.11

remyhaemmerle-da:
CHANGELOG_BEGIN

- [Daml-LF] Release LF 1.11
  - add suport for Generic Map
  - add support for choice observer
  - add generic comparison
  - add convert of contract ID to string (available only in off-ledger mode)
  - reduce size of dar using type interning

CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger-api/VERSION
```diff
@@ -1 +1 @@
-1.8.0
+1.10.0
```
cocreature, line 1:
```suggestion
1.9.0
```
This is not tied to the SDK version, we just bump it when we make a change.

ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxServer.scala
```diff
@@ -413,12 +413,10 @@ final class SandboxServer(
         )
       }
       if (config.seeding.isEmpty) {
-        // TODO https://github.com/digital-asset/daml/issues/7139
-        //  rephrase the message once LF 1.11 is released
         logger.withoutContext.warn(
-          s"""|'${Seeding.NoSeedingModeName}' contract IDs seeding mode is not compatible with the LF 1.11 languages or later which will be released soon.
+          s"""|'${Seeding.NoSeedingModeName}' contract IDs seeding mode is not compatible with the LF 1.11 languages or later.
```
cocreature, line 417: My understanding was that @bame-da and you agreed to not have a default once 1.11 is stable and force users to choose. Is that still the plan? Perfectly fine not to include it in this PR.

daml-lf/language/daml-lf.bzl
```diff
@@ -5,11 +5,11 @@
 # If we make a new LF release, we bump latest and once we make it the default
 # we bump stable.
 lf_stable_version = "1.8"
-lf_latest_version = "1.8"
+lf_latest_version = "1.11"
```
cocreature, line 8: have we already discussed one we flip the default in the compiler? one release after it’s stable like we used to or has that changed with the preview scheme?

cocreature (approved):
Thanks!

daml-lf/language/daml-lf.bzl
remyhaemmerle-da: I am not quite sure. 
@sofiafaro-da what do you think ? 
Should we track that in an issue ?

ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxServer.scala
remyhaemmerle-da: I did not understand that. Let's sync with @bame-da and implement a fix (if necessary) in another PR.

daml-lf/language/daml-lf.bzl
sofiafaro-da: I think the preview scheme is that we have one stable release before LF 1.11 becomes the default, and after that it becomes the default?

cocreature: sgtm so it will be the default in 1.11

sofiafaro-da: That's a nice coincidence -- that LF 1.11 becomes the default starting with SDK 1.11. Is there any chance we'd be able to keep new LF versions in sync with the SDK versions in which they become the default (by skipping a few version numbers ahead each time with the LF veresion)? Or is that a long shot?

PR #8564, opened 2021-01-19
Allow TX updates to arrive in differing order

simonmaxen-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
@@ -35,8 +35,12 @@ object Synchronize {
       _ <- alpha.waitForParties(Set(beta), Set(alice, bob))
       _ <- beta.waitForParties(Set(alpha), Set(alice, bob))
       factory <- alpha.create(alice, AgreementFactory(bob, alice))
-      agreement <- eventually { beta.exercise(bob, factory.exerciseCreateAgreement) }
-      _ <- eventually { alpha.transactionTreeById(agreement.transactionId, alice) }
+      agreement <- eventually {
+        beta.exercise(bob, factory.exerciseCreateAgreement)
+      }
+      _ <- eventually {
+        alpha.transactionTreeById(agreement.transactionId, alice)
+      }
     } yield {
       // Nothing to do, by flatmapping over this we know
       // the two participants are synchronized up to the
@@ -45,12 +49,12 @@ object Synchronize {
   }
 
   final def waitForContract[T](
-      alpha: ParticipantTestContext,
+      participant: ParticipantTestContext,
       party: Party,
       contractId: Primitive.ContractId[T],
   )(implicit ec: ExecutionContext): Future[Unit] = {
     Eventually.eventually {
-      alpha.activeContracts(party).map { events =>
+      participant.activeContracts(party).map { events =>
         assert(
           events.exists(_.contractId == contractId.toString),
           s"Could not find contract $contractId",
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
@@ -3,8 +3,10 @@
 
 package com.daml.ledger.api.testtool.infrastructure
 
+import com.daml.ledger.api.refinements.ApiTypes.Party
 import com.daml.ledger.api.testtool.infrastructure.Eventually.eventually
 import com.daml.ledger.api.testtool.infrastructure.participant.ParticipantTestContext
+import com.daml.ledger.client.binding.Primitive
 import com.daml.ledger.test.model.Test.AgreementFactory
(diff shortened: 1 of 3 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/DivulgenceIT.scala
```diff
@@ -216,10 +217,10 @@ final class DivulgenceIT extends LedgerTestSuite {
     for {
       offer <- alpha.create(proposer, Proposal(from = proposer, to = owner))
       asset <- beta.create(owner, Asset(issuer = owner, owner = owner))
+      _ <- synchronize(alpha, beta)
```
SamirTalwar-DA, line 220: Good catch.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/TransactionServiceIT.scala
```diff
@@ -1582,26 +1582,32 @@ object TransactionServiceIT {
...
+    transactions
+      .map(t =>
+        t.copy(
+          commandId = "commandId",
+          offset = "offset",
+          events = t.events.map(_.modifyWitnessParties(_.sorted)),
+        )
       )
-    )
+      .sortBy(_.transactionId)
```
SamirTalwar-DA, line 1595: I _think_ these sorting changes are fine, but I'm not personally aware of how the transaction service is used in practice. It might be that the transactions are expected to be sorted by applications and, for example, triggers, in which case we should probably be making the ledger do it.

Any idea if anyone is reliant on transaction sort order?

gerolf-da: The sorting is fine. If 2 contracts don't have a relation to each other, the order in which they appear on the transaction stream doesn't matter. You could view this as partial ordering.
This ordering change should only be relevant for the test case `TXSingleMultiSameStakeholders` where the a distributed ledger might re-order the transactions that are created on alpha and beta for a specific party.

gerolf-da (approved):
LGTM

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/TransactionServiceIT.scala
SamirTalwar-DA: Awesome, thanks for the clarification!

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
```diff
@@ -41,4 +43,20 @@ object Synchronize {
       // point before invoking this method
     }
   }
+
+  final def waitForContract[T](
+      alpha: ParticipantTestContext,
```
SamirTalwar-DA, line 48: Please rename to `participant` or something like that.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/DivulgenceIT.scala
```diff
@@ -216,10 +216,10 @@ final class DivulgenceIT extends LedgerTestSuite {
     for {
       offer <- alpha.create(proposer, Proposal(from = proposer, to = owner))
       asset <- beta.create(owner, Asset(issuer = owner, owner = owner))
+      _ <- Synchronize.waitForContract(beta, owner, offer)
```
SamirTalwar-DA, line 219: I think it'd be easier to read if you import `waitForContract` directly.

SamirTalwar-DA (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Synchronize.scala
simonmaxen-da: done

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/DivulgenceIT.scala
simonmaxen-da: done

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/TransactionServiceIT.scala
cocreature: fwiw we do have docs on causality which also include ordering guarantees we make or explicitly don’t make https://docs.daml.com/concepts/local-ledger.html

simonmaxen-da: Yes there are ordering guarantees (like the create must come before an exercise).  The tests using these methods are not concerned with ensuring ordering guarantees (they only create contracts with no contractId arguments).

PR #8565, opened 2021-01-19
faster disk cleanup check

garyverhaegen-da:
I can't reproduce the `chmod` failure and have no clue how to fix it, so for now let's just hope it's not too frequent.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build-unix.yml
+++ b/ci/build-unix.yml
@@ -33,27 +33,19 @@ steps:
       # Location of the disk cache for CI servers set in their init files:
       # infra/macos/2-common-box/init.sh:echo "build:darwin --disk_cache=~/.bazel-cache" > ~/.bazelrc
       # infra/vsts_agent_linux_startup.sh:echo "build:linux --disk_cache=~/.bazel-cache" > ~/.bazelrc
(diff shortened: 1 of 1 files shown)


ci/build-unix.yml
```diff
@@ -33,27 +33,19 @@ steps:
...
-              echo "Deleting '$local_cache'..."
+          echo "removed $disk_cache"
+          local_cache="$HOME/.cache/bazel"
+          if [ -d "$local_cache" ]; then
               chmod -R +w "$local_cache"
               rm -rf "$local_cache"
-          else
-              echo "Local cache small enough:"
-              du -hs "$local_cache"
+              echo "removed $disk_cache"
```
cocreature, line 45:
```suggestion
              echo "removed $local_cache"
```

cocreature (approved):
lgtm and at least locally the bash magic seems to do what we want it to do.

ci/build-unix.yml
garyverhaegen-da: Good catch, thanks!

PR #8570, opened 2021-01-20
LF: generalize usage of GlobalKey#build builder

remyhaemmerle-da:
Hence we avoid usage of the standard Global key builder which require
an extra pass to ensure the contract key does not contract contract
IDs.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
@@ -26,13 +26,6 @@ import scala.collection.immutable
 /** An in-memory representation of a ledger for scenarios */
 object ScenarioLedger {
 
-  @inline
-  def assertNoContractId(key: Value[Value.ContractId]): Value[Nothing] =
-    key.ensureNoCid.fold(
-      cid => crash(s"Not expecting to find a contract id here, but found '$cid'"),
-      identity,
-    )
-
   case class TransactionId(index: Int) extends Ordered[TransactionId] {
     def next: TransactionId = TransactionId(index + 1)
     // The resulting LedgerString is at most 11 chars long
@@ -441,11 +434,8 @@ object ScenarioLedger {
                       val mbNewCache2 = nc.key match {
                         case None => Right(newCache1)
                         case Some(keyWithMaintainers) =>
-                          val gk = GlobalKey(
-                            nc.coinst.template,
-                            // FIXME: we probably should never crash here !
-                            assertNoContractId(keyWithMaintainers.key),
-                          )
+                          // FIXME: we probably should never crash here !
+                          val gk = GlobalKey.assertBuild(nc.coinst.template, keyWithMaintainers.key)
                           newCache1.activeKeys.get(gk) match {
                             case None => Right(newCache1.addKey(gk, nc.coid))
                             case Some(_) => Left(UniqueKeyViolation(gk))
@@ -480,11 +470,8 @@ object ScenarioLedger {
                             case None => newCache0_1
                             case Some(keyWithMaintainers) =>
                               newCache0_1.removeKey(
-                                GlobalKey(
-                                  ex.templateId,
-                                  // FIXME: we probably should'nt crash here !
-                                  assertNoContractId(keyWithMaintainers.key),
-                                )
+                                // FIXME: we probably should'nt crash here !
+                                GlobalKey.assertBuild(ex.templateId, keyWithMaintainers.key)
                               )
                           }
                         } else newCache0
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
@@ -434,7 +434,6 @@ object ScenarioLedger {
                       val mbNewCache2 = nc.key match {
                         case None => Right(newCache1)
                         case Some(keyWithMaintainers) =>
-                          // FIXME: we probably should never crash here !
                           val gk = GlobalKey.assertBuild(nc.coinst.template, keyWithMaintainers.key)
                           newCache1.activeKeys.get(gk) match {
                             case None => Right(newCache1.addKey(gk, nc.coid))
@@ -470,7 +469,6 @@ object ScenarioLedger {
                             case None => newCache0_1
                             case Some(keyWithMaintainers) =>
                               newCache0_1.removeKey(
-                                // FIXME: we probably should'nt crash here !
                                 GlobalKey.assertBuild(ex.templateId, keyWithMaintainers.key)
                               )
                           }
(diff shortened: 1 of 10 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/GlobalKey.scala
```diff
@@ -27,13 +27,16 @@ final class GlobalKey private (
...
 object GlobalKey {
+
   def apply(templateId: Ref.TypeConName, key: Value[Nothing]): GlobalKey =
     new GlobalKey(templateId, key, crypto.Hash.safeHashContractKey(templateId, key))
 
   // Will fail if key contains contract ids
   def build(templateId: Ref.TypeConName, key: Value[ContractId]): Either[String, GlobalKey] =
     crypto.Hash.hashContractKey(templateId, key).map(new GlobalKey(templateId, key, _))
 
+  // Like `build` but throws an exception instead of returning e message in case of error.
```
fabiotudone-da, line 38:
```suggestion
  // Like `build` but,  in case of error, throws an exception instead of returning a message.
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
```diff
@@ -441,11 +434,8 @@ object ScenarioLedger {
                       val mbNewCache2 = nc.key match {
                         case None => Right(newCache1)
                         case Some(keyWithMaintainers) =>
-                          val gk = GlobalKey(
-                            nc.coinst.template,
-                            // FIXME: we probably should never crash here !
-                            assertNoContractId(keyWithMaintainers.key),
-                          )
+                          // FIXME: we probably should never crash here !
+                          val gk = GlobalKey.assertBuild(nc.coinst.template, keyWithMaintainers.key)
```
fabiotudone-da, line 438: How about turning these `FIXMEs` into failfast checks?

cocreature: @fabiotudone-da what exactly are you suggesting? A crash sounds like it fails fast :slightly_smiling_face: 

Either way, this hasn’t changed in this PR so I wouldn’t block it on this.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/InputsAndEffects.scala
```diff
@@ -134,17 +134,7 @@ private[kvutils] object InputsAndEffects {
...
-                        create.coinst.template,
-                        keyWithMaintainers.key,
-                      )
-                      .fold(
-                        _ =>
-                          throw Err
-                            .InvalidSubmission("Contract IDs are not supported in contract keys."),
-                        identity,
-                      )
+                    GlobalKey.assertBuild(create.coinst.template, keyWithMaintainers.key)
```
cocreature, line 137: This looks like a change. You no longer throw `InvalidSubmission`, now we throw `IllegalArgumentException`. Does that matter @fabiotudone-da?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -216,12 +216,6 @@ private[state] object Conversions {
       TransactionCoder.encodeContractInstance(ValueCoder.CidEncoder, coinst),
     )
 
-  def forceNoContractIds(v: Value[Value.ContractId]): Value[Nothing] =
-    v.ensureNoCid.fold(
-      coid => throw Err.InternalError(s"Contract identifier '$coid' encountered in contract key"),
```
cocreature, line 221: same point here, you are changing the exception type.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueCommitting.scala
```diff
@@ -269,10 +269,9 @@ object KeyValueCommitting {
     // to allocate a log entry id and we can directly place the output keys into the submission and do not need
     // to compute outputs from serialized transaction.
     val contractKey =
-      GlobalKey(
+      GlobalKey.assertBuild(
         decodeIdentifier(templateId),
-        decodeVersionedValue(key).value.ensureNoCid
-          .getOrElse(throw Err.DecodeError("ContractKey", "Contract key contained contract id")),
```
cocreature, line 275: another changed exception type.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/TransactionCommitter.scala
```diff
@@ -553,16 +553,7 @@ private[kvutils] class TransactionCommitter(
...
         cs.setContractKey(
           Conversions.encodeGlobalKey(
-            GlobalKey
-              .build(
-                createNode.coinst.template,
-                keyWithMaintainers.key,
-              )
-              .fold(
-                _ =>
-                  throw Err.InvalidSubmission("Contract IDs are not supported in contract keys."),
```
cocreature, line 563: and here as well.

cocreature (commented):
I like the change in principle but it changes the exceptions thrown in kvutils in a bunch of places so I’d like someone who understands kvutils to take a look at this to make sure this is fine.

fabiotudone-da:
> @fabiotudone-da what exactly are you suggesting? A crash sounds like it fails fast slightly_smiling_face

Absolutely, so I propose to either removing the `FIXME`s altogether, if the assert message (+stack) is dev friendly enough, or catching and re-throwing with a more specific/contextualized error.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
fabiotudone-da: Absolutely, so I propose to either remove the FIXMEs altogether, if the assert message (+stack) is dev friendly enough, or catching and re-throwing with a more specific/contextualized error.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/InputsAndEffects.scala
fabiotudone-da: No idea, it might, will check ASAP.

SamirTalwar-DA: Yes, I think it matters. We don't really treat `Err` errors differently from other exceptions, but they do have more meaningful messages. At the least we should probably wrap this in an `Err.InvalidSubmission`.

SamirTalwar-DA: (Same applies to all other exception type changes regarding `Err`.)

remyhaemmerle-da (commented):
There is pre-existing issue with the exception we throw at bit randomly  `InvalidSubmission`, `DecodeError` or `InternalError` (we throw different exception in the same file and even in the same function). 

I tried to to tied up a bit. @SamirTalwar-DA or @fabiotudone-da please double check  if what I have made make sens. 

So concerning the change of exception 
 - `InvalidSubmission` should definitively not be change by `IllegalArgumentException` as this is a validation error.
 -  `InternalError` probably could be changed to `IllegalArgumentException` as this corresponds to something very bad happening in the Engine. 
 - `DecodeError` i am not quite sure, when those shoule be raise. 

In anycase I revert to always throw one of the 3 `Err` instead of `IllegalArgumentException`.

fabiotudone-da (approved):
It should be fine like this. Thanks!

PR #8571, opened 2021-01-20
[Participant] Pipelined transaction indexing

tudor-da:
Currently, the participant indexer processes `TransactionAccepted` atomically. More specifically, all the SQL updates to the affected tables are done in one transaction (`participant_events`, `participant_contracts`, `participant_contract_witnesses`, `participant_command_completions` and `parameters` - for ledger end updates). 

With the purpose of increasing the indexer throughput (by higher parallelism to the index DB), this PR breaks the aforementioned DAML transaction into multiple SQL transactions, updating the involved tables in separate async Akka operators:
* events update (`participant_events`)
* state update (`participant_contracts` and `participant_contract_witnesses`)
* transaction completion (`participant_command_completions` and `parameters` - ledger end update)

### Involved changes
* There will be two supported indexer behaviors: `AtomicExecuteUpdate` (legacy behavior - used for DAML-on-SQL and H2 `JdbcIndexer`) and `PipelinedExecuteUpdate` which implements the mentioned optimization.
* `JdbcIndexer` becomes agnostic of the `JdbcLedgerDao` - it delegates the dao updates to the `ExecuteUpdate` flows. 
* In order to support mid-update crashes (and automatic restarts), `participant_events` insertions become idempotent.
* For consistency reasons, `get(Flat)TransactionBy(Event)Id` queries become guarded by the current ledger end.
* Historically, the ledger entry insertion performed atomically was timed with `daml_index_db_store_ledger_entry`. When using the pipelined insertion, the single metric measures  all three stages combined (across Akka stage boundaries). Additionally, each stage is timed independently by `daml_index_db_store_ledger_entry_events`, `daml_index_db_store_ledger_entry_state` and `daml_index_db_store_ledger_entry_completion`.

### Performance improvement
The benchmarked performance improvement is ~20 %, but expected to be higher with other optimizations affecting events insertion latency.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA (approved):
This looks good to me. Please get a review from someone more familiar with the code before merging.

Follow-up comment coming soon with some ideas for how to improve the safety of this going forward.

SamirTalwar-DA:
The tests timed out on macOS.

SamirTalwar-DA:
I worry that we will accidentally introduce writes to the same table across transactions, or as you pointed out, reads that join tables that span two of the transactions.

Perhaps we could make this safer by creating aggregates that gate access to collections of tables that should always be internally consistent. For example, a "completions" aggregate could manage access to the completions and parameters tables. These aggregates would be responsible for creating the database transaction and passing in the references to the tables, which would mean it would be impossible to access, for example, the events table and the completions table at the same time.

This would be a huge change and I'm not implying we should make changes here to support this, but it might be helpful to think along these lines when refactoring further.

tudor-da:
> The tests timed out on macOS.

Apparently it's just due to this macOS instance being very slow. I see all the tests taking 3-5x more than on the Linux build. I'll just restart it and check the timings.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/TransactionServiceResponseValidator.scala
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.validation
+
+import com.daml.ledger.api.domain.{LedgerId, LedgerOffset}
+import com.daml.lf.data.Ref.LedgerString
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class TransactionServiceResponseValidator(
```
stefanobaghino-da, line 11: Was this just moved here or is there some change to be reviewed?

stefanobaghino-da: Wait, I'm not sure I follow. Why do we validate a response?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -25,6 +25,12 @@ sealed trait Update extends Product with Serializable {
   def recordTime: Timestamp
 }
 
+/** An update describing metadata changes.
+  *
+  * Metadata changes regard parties, packages, ledger configurations and command rejections.
+  */
+sealed trait MetadataUpdate extends Update
```
stefanobaghino-da, line 32: What's the purpose of this change?

stefanobaghino-da (approved):
This was very difficult to review. I hope I didn't miss anything. I would recommend you break it down into easily reviewable chunks if possible.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/TransactionServiceResponseValidator.scala
tudor-da: The LedgerDao and Ledger API offer different guarantees now wrt to consistency. In this sense, the API cannot return a transaction that is past the current ledger end, even though the dao can do it.

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
```diff
@@ -0,0 +1,416 @@
...
+      Timed.future(
+        metrics.daml.index.db.storeTransactionEvents,
+        ledgerDao
+          .storeTransactionEvents(preparedInsert)
+          .map(_ => offsetUpdate),
+      )
+    case offsetUpdate => Future.successful(offsetUpdate)
+  }
+
+  private[indexer] val flow: ExecuteUpdateFlow =
```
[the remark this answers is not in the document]
rautenrieth-da: It might be worth adding something like `com.daml.metrics.Timed.timedAndTrackedFuture`, but for Akka streams. 

Another use case is measuring the number of elements being currently processed, or measuring the number of items in a `.buffer()` (cc @nmarton-da).

This is not something for this PR though.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
tudor-da: I have added it for making threading through the pipeline stages more type safe and modular (e.g. in the pipelined update flow, we either act on the `MetadataUpdate` or on the `TransactionAccepted`).

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
```diff
@@ -0,0 +1,416 @@
...
+    Flow[OffsetUpdate]
+      .mapAsync(1)(prepareUpdate)
+      .mapAsync(1)(insertTransactionState)
+      .mapAsync(1)(insertTransactionEvents)
```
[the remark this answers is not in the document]
tudor-da: Indeed, this is a way to go forward.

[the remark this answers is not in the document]
rautenrieth-da: We could batch through a [.conflate()](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/conflate.html) stream operator. 

But I think this would be counter-productive. The point of this PR is to allow all three updates to run asynchronously, so that the next transaction can start writing contracts as soon as the first transaction is done with contracts and starts writing events.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/TransactionServiceResponseValidator.scala
rautenrieth-da: This validator only checks that the returned transaction is not beyond the stored ledger end (which would be an _extremely_ rare case), right?

The corresponding SQL queries already join the events table with the parameters table, in order to check that the transaction was not pruned: 

```
SELECT event_offset, ...
FROM participant_events
JOIN parameters on participant_pruned_up_to_inclusive is null or event_offset > participant_pruned_up_to_inclusive
WHERE transaction_id = $transactionId
```

Could simply modify the SQL query to also check that the event offset is not beyond the stored ledger end?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
rautenrieth-da: This seems to split ledger updates to "transaction accepted" and "all other changes". This is related to a particular indexer optimization, and I'm not sure such a trait belongs to the participant state interface.

Before this PR, we had:
```
OffsetUpdate = PreparedTransactionInsert | OffsetStepUpdatePair
```
where `OffsetStepUpdatePair` could include any update (including transactions). It looks like you added `MetadataUpdate` so that you have:
```
OffsetUpdate = (PreparedTransactionInsert | MetadataUpdateStep) | OffsetStepUpdatePair
```
where `MetadataUpdateStep` is the same as `OffsetStepUpdatePair`, except that it's guaranteed to not include transactions. Do we really need that split? If anything, shouldn't it be:
```
OffsetUpdate = PreparedTransactionInsert | UnpreparedUpdateThatIsAnythingButATransactionInsert
```
?

The main benefit seems to be to avoid duplicating the code that handles transaction updates. Before this PR, we had duplicated code and a logger warning for this: https://github.com/digital-asset/daml/pull/8571/files#diff-63690f91ba83940aae6dc820d83d141caec268ff769438437ee49703f3a9d238L391.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTablePostgresql.scala
```diff
@@ -176,9 +176,11 @@ object EventsTablePostgresql extends EventsTable {
     val exerciseChildEventIds = "exerciseChildEventIds"
   }
 
-  private val insertStmt = {
+  private val conflictActionClause = if (idempotentEventInsertions) "on conflict do nothing" else ""
```
rautenrieth-da, line 179: This ignores all conflicts. Could you add a comment what potential conflicts there are, and why it's ok to ignore all of them? It's hard to look up what constraints and indices we have on the `participant_events` table.

rautenrieth-da (commented):
First batch of review comments.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -456,37 +488,55 @@ private class JdbcLedgerDao(
...
+      offsetStep: OffsetStep,
+  )(implicit connection: Connection): Unit =
+    Timed.value(
+      metrics.daml.index.db.storeTransactionDbMetrics.insertCompletion,
+      submitterInfo
+        .map(prepareCompletionInsert(_, offsetStep.offset, transactionId, recordTime))
+        .foreach(_.execute()),
+    )
+
+  private def updateLedgerEnd(offsetStep: OffsetStep)(implicit connection: Connection): Unit =
```
rautenrieth-da, line 534: FYI: At one point, we had the convention that methods that take an `implicit connection: Connection` use the `Sync` suffix to make it easier to see that they don't open an extra database connection. I see only `stopDeduplicatingCommandSync` survived the many refactorings, so there's no point keeping the convention.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcIdempotentInsertionsSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.{Inside, LoneElement, OptionValues}
+
+trait JdbcIdempotentInsertionsSpec extends Inside with OptionValues with Matchers with LoneElement {
+  self: AsyncFlatSpec with JdbcLedgerDaoSuite =>
+
+  behavior of "JdbcLedgerDao (on PostgreSQL)"
+
+  it should "allow idempotent transaction insertions" in {
```
rautenrieth-da, line 19: Nice to see you are including many tests! If you want to improve it further, you could add the following test(s):

1. Test idempotency of partial inserts. E.g., first insert only the state of a transaction (manually simulating a crash before the events were inserted), then insert only the state + events of the same transaction, then insert the whole transaction.
2. Test that the ledger end guards are working. E.g., first insert only the state + events of a transaction (manually simulating a crash before the ledger end was updated), then verify that this transaction can't be looked up.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/TransactionServiceResponseValidator.scala
tudor-da: Done.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
tudor-da: As it is a more widely-used API, I've removed the change.

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
```diff
@@ -0,0 +1,416 @@
...
+      .mapAsync(1) { case offsetUpdate @ OffsetUpdate(offsetStep, update) =>
+        withEnrichedLoggingContext(loggingContextFor(offsetStep.offset, update)) {
+          implicit loggingContext =>
+            Timed.future(
+              metrics.daml.indexer.stateUpdateProcessing,
+              executeUpdate(offsetUpdate),
+            )
+        }
+      }
```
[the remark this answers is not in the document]
tudor-da: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTablePostgresql.scala
tudor-da: Done. Added a doc specifying the conflict clause and why is needed.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcIdempotentInsertionsSpec.scala
tudor-da: Done. Added the tests in `JdbcPipelinedInsertionsSpec`

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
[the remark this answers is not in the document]
tudor-da: Added back the timer by threading the timer context through the stages. @rautenrieth-da A reason for which I did not want to time the entire flow is its inherent branching: we would be timing both metadata and pipelined transaction updates on the `store_ledger_entry` timer.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTablePostgresql.scala
```diff
@@ -176,9 +176,19 @@ object EventsTablePostgresql extends EventsTable {
     val exerciseChildEventIds = "exerciseChildEventIds"
   }
 
-  private val insertStmt = {
+  /** Conflict detection relies on the `event_id` primary key only.
```
rautenrieth-da, line 179: This sounds a bit misleading.

If the indexer does perform post-commit validation (daml-on-sql), then it does so through [ActiveLedgerStateManager](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala) and explicit SQL queries.

If the indexer does not perform post-commit validation, then we expect all writes to be free of conflicts because the underlying ReadService should not emit conflicting transactions.

All indices and constraints in the index database should be either performance optimizations or safeguards that are not strictly necessary.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private def assertNoResponseFor(
+      offsetTx: (Offset, LedgerEntry.Transaction)
+  ): Future[Assertion] = {
+    val (offset, tx) = offsetTx
+    val maybeSubmitterInfo = submitterInfo(tx)
+    val preparedInsert = prepareInsert(maybeSubmitterInfo, tx, CurrentOffset(offset))
+    for {
+      _ <- ledgerDao.storeTransactionEvents(preparedInsert)
```
rautenrieth-da, line 79: You could even write one more step:
```suggestion
      _ <- ledgerDao.storeTransactionEvents(preparedInsert)
      _ <- ledgerDao.storeTransactionState(preparedInsert)
```

rautenrieth-da (approved):
This was a bit difficult to review due to the size. The important parts look good:
- The underlying SQL queries for fetching transactions are guarded against access beyond the ledger end
- There are tests that prove that partially written transactions can be safely re-written
- There are tests that prove that you can't fetch transactions beyond the serialized ledger end (which would be an extreme edge case)

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
tudor-da: Done

tudor-da: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTablePostgresql.scala
tudor-da: Ok, my aim intent was not clear then. I was referring to table-level insertion conflict (when inserting a row with an already-existing `event_id`) which is not a _update conflict_ for the pipelined indexer and should just be ignored. I've adapted the comment.

PR #8578, opened 2021-01-20
Bump Maven timeouts and retries

cocreature:
Chosen fairly arbitrarily, I don’t have a good idea what sensible
values are but given that we keep running into issues the current ones
are apparently too low.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/release/src/Upload.hs
+++ b/release/src/Upload.hs
@@ -371,9 +371,9 @@ logStatusRetry shouldRetry _ status =
     else
         $logDebug ("Aborting staging repository check after " <> (tshow $ rsCumulativeDelay status) <> "µs.")
 
--- Retry for 10 minutes total, doubling delay starting with 200ms.
+-- Retry for 5 minutes total, doubling delay starting with 20ms.
 uploadRetryPolicy :: RetryPolicy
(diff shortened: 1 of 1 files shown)


release/src/Upload.hs
```diff
@@ -371,8 +371,9 @@ logStatusRetry shouldRetry _ status =
     else
         $logDebug ("Aborting staging repository check after " <> (tshow $ rsCumulativeDelay status) <> "µs.")
 
+-- Retry for 10 minutes total, doubling delay starting with 200ms.
 uploadRetryPolicy :: RetryPolicy
-uploadRetryPolicy = limitRetriesByCumulativeDelay (60 * 1000 * 1000) (exponentialBackoff (200 * 100))
+uploadRetryPolicy = limitRetriesByCumulativeDelay (5 * 60 * 1000 * 1000) (exponentialBackoff (200 * 100))
```
garyverhaegen-da, line 376: I do not understand this in relation to the comment.

Assuming the unit here is microseconds, the start of the backoff seems to be 20ms, not 200, while the total delay seems to be 5 minutes, not 10. What am I missing?

garyverhaegen-da (approved):

release/src/Upload.hs
garyverhaegen-da: 👍

PR #8579, opened 2021-01-20
Add buffer to speed up transaction streaming

rautenrieth-da:
This PR fixes an issue that was causing the transaction service to become almost twice as slow. I still don't understand what exactly is going on, but this small PR restores the original performance. 

## Changelog

```
CHANGELOG_BEGIN
- [Ledger API Server] Fixed an issue that was causing a slowdown of the
  transaction service.
CHANGELOG_END
```

## Measuring the impact

Method:
- Generate an index database with huntsman (many thanks to Martino)
- Build ledger-on-memory from commit 9ed787cb
- Run ledger-on-memory in ledger-api-only mode, connect to the existing index database
- Run spider-mock to measure the speed of the transaction service
- Repeat with ledger-on-memory where the out-of-order fix was reverted

## Results

Reading 17009 transactions took:
```
mergeSubstreams():                 18.602 secs (before the out-of-order fix)
concatSubstreams():                29.007 secs (after the out-of-order fix)
statefulMapConcat():               28.656 secs (first attempt: code without substreams)
statefulMapConcat() + 2x buffer(): 18.767 secs (second attempt: add buffers)
concatSubstreams() + buffer():     18.981 secs (third attempt: smallest change)
```

Code for the `statefulMapConcat` versions: 
```
source
  .map(Some(_))
  .concat(Source.single(None))
  .statefulMapConcat(() => {
    var previousSegmentKey: K = null.asInstanceOf[K]
    val builder = new VectorBuilder[A]
    entryO => {
      entryO match {
        case Some(entry) =>
          val keyForEntry = by(entry)
          if (keyForEntry != previousSegmentKey) {
            previousSegmentKey = keyForEntry
            val result = builder.result()
            builder.clear()
            builder += entry
            List(result)
          } else {
            builder += entry
            List.empty
          }
        case None =>
          List(builder.result())
      }
    }
  })
```

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
@@ -5,6 +5,7 @@ package com.daml.platform.store.dao.events
 
 import java.sql.Connection
 
+import akka.stream.OverflowStrategy
 import akka.stream.scaladsl.Source
 import akka.{Done, NotUsed}
 import com.daml
@@ -49,6 +50,10 @@ private[dao] final class TransactionsReader(
   private val sqlFunctions = SqlFunctions(dbType)
 
   private val logger = ContextualizedLogger.get(this.getClass)
+
+  // TransactionReader adds an Akka stream buffer at the end of all streaming queries.
+  // This significantly improves the performance of the transaction service.
+  private val outputStreamBufferSize = 128
 
   private def offsetFor(response: GetTransactionsResponse): Offset =
     ApiOffset.assertFromString(response.transactions.head.offset)
@@ -115,6 +120,7 @@ private[dao] final class TransactionsReader(
         val response = EventsTable.Entry.toGetTransactionsResponse(events)
         response.map(r => offsetFor(r) -> r)
       }
+      .buffer(outputStreamBufferSize, OverflowStrategy.backpressure)
       .wireTap(_ match {
         case (_, response) =>
           response.transactions.foreach(txn =>
@@ -201,6 +207,7 @@ private[dao] final class TransactionsReader(
         val response = EventsTable.Entry.toGetTransactionTreesResponse(events)
         response.map(r => offsetFor(r) -> r)
       }
+      .buffer(outputStreamBufferSize, OverflowStrategy.backpressure)
       .wireTap(_ match {
         case (_, response) =>
           response.transactions.foreach(txn =>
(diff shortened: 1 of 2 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/dao/events/package.scala
```diff
@@ -74,6 +75,7 @@ package object events {
       .map(_._1)
       .fold(Vector.empty[A])(_ :+ _)
       .concatSubstreams
+      .buffer(128, OverflowStrategy.backpressure)
```
rautenrieth-da, line 78: `.buffer(16)` did not bring much improvement. I did not investigate other buffer sizes.

SamirTalwar-DA: Can you pull this out into a constant somewhere? Or better yet, a parameter?

I worry about the volume of magic numbers inside our codebase. I'd really like to push them to at the very least, a single file with all the magic numbers.

stefanobaghino-da (approved):
What @SamirTalwar said, plus I would recommend also adding the reason for the buffer. There already is an "implementation detail" section.

Thank you so much for the very accurate analysis.

SamirTalwar-DA:
> Thank you so much for the very accurate analysis.

I second this. It was very helpful!

rautenrieth-da:
@stefanobaghino-da @SamirTalwar-DA I have moved the `.buffer()` call to TransactionReader and have created a documented private val for the magic number. I felt the buffer call does not belong to the generic `groupContiguous` method. I have verified that the performance of the transaction stream is slightly better than before the fix:

```
mergeSubstreams():             16.7 secs (before the out-of-order fix)
concatSubstreams():            26.6 secs (after the out-of-order fix)
concatSubstreams() + buffer(): 16.4 secs (after this PR)
```

stefanobaghino-da:
> @stefanobaghino-da @SamirTalwar-DA I have moved the `.buffer()` call to TransactionReader and have created a documented private val for the magic number. I felt the buffer call does not belong to the generic `groupContiguous` method. I have verified that the performance of the transaction stream is slightly better than before the fix:
> 
> ```
> mergeSubstreams():             16.7 secs (before the out-of-order fix)
> concatSubstreams():            26.6 secs (after the out-of-order fix)
> concatSubstreams() + buffer(): 16.4 secs (after this PR)
> ```

Given the hard-coded number it probably makes more sense like this. Do you know what `.mergeSubstreams` does?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
```diff
@@ -50,6 +51,10 @@ private[dao] final class TransactionsReader(
 
   private val logger = ContextualizedLogger.get(this.getClass)
 
+  // TransactionReader adds an Akka stream buffer at the end of all streaming queries.
+  // This significantly improves the performance of the transaction service.
+  private val outputStreamBufferSize = 128
```
SamirTalwar-DA, line 56: I would still prefer if we moved constants like this into a single file, but I'm OK with this for now.

SamirTalwar-DA (approved):

rautenrieth-da:
> Do you know what .mergeSubstreams does?

I don't, unfortunately. All I know is that concatSubstreams() is implemented as mergeSubstreams() with maxParallelism=1.

PR #8583, opened 2021-01-20
Fixed a flaky RecoveringIndexerSpec test case [DPP-197]

kamil-da:
The `"wait until the subscription completes"` test case of `RecoveringIndexerSpec` is flaky.
The failure is:
```
Vector((INFO,Starting Indexer Server), (INFO,Started Indexer Server), (INFO,Successfully finished processing state updates)) did not contain the same elements in the same (iterated) order as List((INFO,Starting Indexer Server), (INFO,Started Indexer Server)) (RecoveringIndexerSpec.scala:114)
```

This means that the indexer may be progressing faster than the test case expects - it's not only able to start the indexer server but also to process the subscription.

I propose the simplest solution for this which (as far as I understand) doesn't weaken the test case logic.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
@@ -111,7 +111,8 @@ final class RecoveringIndexerSpec
       val resource = recoveringIndexer.start(() => testIndexer.subscribe())
       resource.asFuture
         .map { complete =>
-          readLog() should contain theSameElementsInOrderAs Seq(
+          // at this point the read log should at least contain logs informing about a successful indexer server startup
+          readLog().take(2) should contain theSameElementsInOrderAs Seq(
             Level.INFO -> "Starting Indexer Server",
             Level.INFO -> "Started Indexer Server",
           )


ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
```diff
@@ -111,7 +111,8 @@ final class RecoveringIndexerSpec
       val resource = recoveringIndexer.start(() => testIndexer.subscribe())
       resource.asFuture
         .map { complete =>
-          readLog() should contain theSameElementsInOrderAs Seq(
+          // at this point the read log should at least contain logs informing about a successful indexer server startup
+          readLog().take(2) should contain theSameElementsInOrderAs Seq(
```
tudor-da, line 115: if you flatten over `complete`, you can confidently assert the full output with `readLog()` (including the `(INFO,Successfully finished processing state updates)` line that is making it flaky now, right?

SamirTalwar-DA: Yes, but the next section handles that. This could be split into a separate test, I guess, named `should "return control once the indexer has started"` or something like that.

SamirTalwar-DA (approved):

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
kamil-da: Thank you for comments.
Should I add the additional test case in this PR?

SamirTalwar-DA: If you agree, sure. If not, I already approved. :slightly_smiling_face:

kamil-da: Let's leave it in the current form as I saw some intentions to refactor the `RecoveringIndexer`. :)

PR #8590, opened 2021-01-21
Improve browser compatbility of auth middleware

aherrmann-da:
- Includes the authentication challenge in response body
  Some browsers make it difficult to access the `WWW-Authenticate` response header from Javascript. For example, Firefox 84.0.2 (64-bit) on Linux does not expose the `WWW-Authenticate` header in the result of the `fetch` function, independent of the server's access control headers. In that case the header is only accessible through the `XMLHttpRequest` API, which is more cumbersome to use.

  This adds the challenge to the response body in JSON format as well to avoid exposing users to any such browser related issues.
- Make the `Secure` `Set-Cookie` attribute configurable
  Since [Chrome 80][1] `Set-Cookie` with the `Secure` attribute enabled is reject for connections that don't use `https`. This includes `localhost`. Firefox, at least as of version 84.0.2, allows such cookies on `localhost`.

  This adds a command-line flag to the authorization middleware to make the value of the `Secure` attribute configurable. This way it can be disabled for development purposes.

[1]: https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

### Pull Request Checklist

trigger the build.

### DIFF
--- a/triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
+++ b/triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
@@ -84,7 +84,9 @@ object Config {
 
       opt[Boolean]("cookie-secure")
         .action((x, c) => c.copy(cookieSecure = x))
-        .text("Enable the Secure attribute on the cookie that stores the token.")
+        .text(
+          "Enable the Secure attribute on the cookie that stores the token. Defaults to true. Only disable this for testing and development purposes."
+        )
 
       opt[Long]("login-request-timeout")
         .action((x, c) => c.copy(loginTimeout = FiniteDuration(x, duration.SECONDS)))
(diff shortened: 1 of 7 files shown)


triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
```diff
@@ -79,6 +82,10 @@ object Config {
           "Maximum number of simultaneously pending login requests. Requests will be denied when exceeded until earlier requests have been completed or timed out."
         )
 
+      opt[Boolean]("cookie-secure")
+        .action((x, c) => c.copy(cookieSecure = x))
+        .text("Enable the Secure attribute on the cookie that stores the token.")
```
cocreature, line 87: maybe worth adding that this defaults to secure and that false should only be used for development?

cocreature (approved):
Thanks!

triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
aherrmann-da: Good idea, done.

Issue #7709, opened 2020-10-16
Add choice observers


nickchapman-da:
**TODOS** (not exhaustive)

* ~Implement feature in the engine.~ (#7139)
* ~fix transaction documentation~ (#8281)
* ~fix LF specifications~ (#8592)
* ~add test for TransactionCoder~ (#7795)
* ~add tests for archive Decoder~ (#8281)
* ~ledger-api conformance tests~ (@gerolf-da is fine we do not implement this test, because a specific ledger does not influence the computation of the choice observers. This should still be covered by tests of the blinding information computation)
* ~decide if choice observer is optional~  (#8316)
* ~remove the "early access" comments from the documentation~

cocreature:
closing in favor of #8714

PR #8592, opened 2021-01-21
Add choice observers in LF spec.

sofiafaro-da:
Part of #7709. Adds choice observers to the syntax, type system, and
and operational semantics of LF. Adds a test for the operational
semantics of choice observers.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/daml-test-files/SemanticsEvalOrderSince1_11.daml
+++ b/compiler/damlc/tests/daml-test-files/SemanticsEvalOrderSince1_11.daml
@@ -1,0 +1,41 @@
+
+-- | Test operational semantics for 1.11 features.
+--
+-- @SINCE-LF 1.11
+module SemanticsEvalOrderSince1_11 where
+
+-- @ERROR EvUpdExercWithoutActorsErr OK
+template T_EvUpdExercWithoutActorsErr
+    with
+        p : Party
+    where
+        signatory p
+        choice C_EvUpdExercWithoutActorsErr : ()
+          observer error @Party "EvUpdExercWithoutActorsErr failed at observers"
+          controller error @Party "EvUpdExercWithoutActorsErr OK"
+          do error "EvUpdExercWithoutActorsErr failed at body"
+
+evUpdExercWithoutActorsErr = scenario do
+    p <- getParty "Alice"
+    submit p do
+        t <- create (T_EvUpdExercWithoutActorsErr p)
+        exercise t C_EvUpdExercWithoutActorsErr
+        abort "EvUpdExercWithoutActorsErr failed after exercise"
+
+-- @ERROR EvUpdExercObserversErr OK
+template T_EvUpdExercObserversErr
+    with
+        p : Party
+    where
+        signatory p
+        choice C_EvUpdExercObserversErr : ()
+          observer error @Party "EvUpdExercObserversErr OK"
+          controller p
+          do error "EvUpdExercObserversErr failed at body"
+
+evUpdExercObserversErr = scenario do
+    p <- getParty "Alice"
+    submit p do
+        t <- create (T_EvUpdExercObserversErr p)
+        exercise t C_EvUpdExercObserversErr
+        abort "EvUpdExercObserversErr failed after exercise"
--- a/daml-lf/spec/daml-lf-1.rst
+++ b/daml-lf/spec/daml-lf-1.rst
@@ -685,9 +685,15 @@ available for usage::
       ::= 'no_key'
        |  'key' τ eₖ eₘ
 
+  Temeplate choice observers
+    ChObs
+      ::= 'no_observers'
+        | 'observers' eₒ
+
   Template choice definition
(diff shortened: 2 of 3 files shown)


daml-lf/spec/daml-lf-1.rst
```diff
@@ -685,9 +685,15 @@ available for usage::
       ::= 'no_key'
        |  'key' τ eₖ eₘ
 
+  Temeplate choice observers
```
cocreature, line 688:
```suggestion
  Template choice observers
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -2945,23 +2974,50 @@ as described by the ledger model::
...
+
+     'tpl' (x : T)
+         ↦ { 'choices' { …, 'choice' ChKind Ch (y : 'ContractId' Mod:T) (z : τ) : σ 'by' eₚ ChObs ↦ eₐ, … }, … }  ∈  〚Ξ〛Mod
+     cid ∈ dom(st₀)
+     st₀(cid) = (Mod:T, vₜ, 'active')
+     eₚ[x ↦ vₜ, z ↦ v₂]  ⇓  Ok vₚ
+     v₁ =ₛ vₚ
+     ChObs[x ↦ vₜ, z ↦ v₂]  ⇓  Ok vₒ
+     eₐ[x ↦ vₜ, y ↦ cid, z ↦ v₂]  ⇓  Err @τ v
+   —————————————————————————————————————————————————————————————————————— EvUpdExercBodyEvalErr
```
cocreature, line 2996: this looks like it’s just a duplicate? it has the same name as the one below and I cannot spot a difference.

compiler/damlc/tests/daml-test-files/SemanticsEvalOrderSince1_11.daml
```diff
@@ -0,0 +1,41 @@
+
+-- | Test operational semantics for 1.11 features.
+--
+-- @SINCE-LF 1.11
+module SemanticsEvalOrderSince1_11 where
+
+-- @ERROR EvUpdExercWithoutActorsErr OK
```
cocreature, line 7: the order seems a bit surprising. We enforce an order at the parser level where `observer` must come before `controller` but then we evaluate it the other way around. Is there a particular reason for this choice? It seems like evaluating observers first might be more intuitive and I don’t see why it should be harder from an implementation pov.

cocreature (approved):
Thanks!

daml-lf/spec/daml-lf-1.rst
```diff
@@ -685,9 +685,15 @@ available for usage::
...
        |  'key' τ eₖ eₘ
 
+  Temeplate choice observers
+    ChObs
+      ::= 'no_observers'
+        | 'observers' eₒ
+
   Template choice definition
-    ChDef ::= 'choice' ChKind Ch (y : 'ContractId' Mod:T) (z : τ)  : σ 'by' eₚ ↦ e
+    ChDef ::= 'choice' ChKind Ch (y : 'ContractId' Mod:T) (z : τ) : σ 'by' eₚ ChObs ↦ e
```
remyhaemmerle-da, line 694: I would prefer to not speak about this detail in the spec. 
I mean I would be in favour making the choice observer always a list of parties and ignoring the special case where it is not defined and add a small comment in the `Serialization changes since version 1.6` section saying to choice from LF < 1.11 are interpreted as with empty choice observer. 

From the point of view of the specification this should be the same.

sofiafaro-da: Ok, that would be cleaner. I'll change it like that.

compiler/damlc/tests/daml-test-files/SemanticsEvalOrderSince1_11.daml
sofiafaro-da: Not sure there's a good reason, but it's the semantics in LF 1.11.

Here's a bad justification: From the point of view of the spec, in an "exercise without actors" (which this is) we need to determine the exercise actor via the controller field, before we get an "exercise with actors", which is where all the semantics for an exercise (including evaluating observers) is defined.

I think "exercise with actors" is an outdated notion and should probably be removed from the spec, at some point? (cc @remyhaemmerle-da)

The fact that observers have to come before controllers in the syntax is a bit of an accident. IIRC Martin preferred it to be flexible, or to have it fixed the other way, but Nick ran into issues trying to implement it that way in the GHC parser.

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: Oops! Nice catch :)

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: Thanks :) My keyboard likes to add extra 'e' :-/

PR #8335, opened 2020-12-17
KV TransactionCommitter: report correct rejection reason in case of contract keys contention [KVL-760]

kamil-da:
## Description

Adds a key consistency validation stage to kvutils' `TransactionCommitter`, checking that contract keys associated to a transaction's contract instances haven't changed since the participant submitted.

The new stage relies on additional processing by `InputsAndEffects`:

- such contracts keys are also added to the submission inputs
- a new contract key -> contract_id mapping is shipped as part of the submission by the participant to the committer

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8594, opened 2021-01-21
KV TransactionCommitter: report correct rejection reason in case of contract keys contention, second attempt [KVL-760]

fabiotudone-da:
## Description

Re-implements #8335 fully in the committer, i.e. no proto changes needed and no change to `InputsAndEffects`' logic. Also, for the sake of efficiency, a single transaction tree traversal is shared by uniqueness and consistency checks.

Tests have been carried over as they were in #8335.

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+  private def validateTransactionContractKeysUniquenessAndConsistency(
+      transactionCommitter: TransactionCommitter,
+      recordTime: Option[Timestamp],
+      transactionEntry: DamlTransactionEntrySummary,
+      contractKeyDamlStateKeys: Set[DamlStateKey],
+      contractKeysToContractIds: Map[DamlContractKey, RawContractId],
+  ): StepResult[DamlTransactionEntrySummary] = {
+
+    val keysValidationOutcome = transactionEntry.transaction
+      .foldInExecutionOrder[KeyValidationStatus](
```
fabiotudone-da, line 133: It would be nice to be able break out of this fold, is it possible at all?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+      )
+
+    keysValidationOutcome match {
+      case Right(_) =>
+        StepContinue(transactionEntry)
+      case Left(error) =>
+        val message = error match {
+          case Duplicate => "DuplicateKeys: at least one contract key is not unique"
+          case Inconsistent =>
+            "InconsistentKeys: at least one contract key has changed since the submission"
```
fabiotudone-da, line 154: In #8335 we included the keys in the inconsistent case but I'd like to leave that for a future PR because 1. it's more computation and 2. for consistency we should really do for the duplicate case as well.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+import com.daml.ledger.participant.state.v1.RejectionReason
+import com.daml.lf.data.Ref.TypeConName
+import com.daml.lf.data.Time.Timestamp
+import com.daml.lf.transaction.{Node, NodeId}
+import com.daml.lf.value.Value
+import com.daml.lf.value.Value.ContractId
+
+object TransactionContractKeysValidation {
+
+  private type RawContractId = String
```
fabiotudone-da, line 28: Slightly more readable than `String` (as it was before) but not more type safe. I considered using `ContractId` but it's probably overkill in this context.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+
+  private[transaction] def validate(
+      transactionCommitter: TransactionCommitter
+  ): Step[DamlTransactionEntrySummary] =
+    (commitContext, transactionEntry) => {
+      val damlState = commitContext
+        .collectInputs[(DamlStateKey, DamlStateValue), Map[DamlStateKey, DamlStateValue]] {
+          case (key, Some(value)) if key.hasContractKey => key -> value
+        }
```
SamirTalwar-DA, line 37: We used to append `commitContext.getOutputs` to this. Is there a reason we don't do that any more?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+  private def validateNodeContractKey(
+      node: Node.GenNode[NodeId, ContractId],
+      contractKeysToContractIds: Map[DamlContractKey, RawContractId],
+      keyValidationStatus: KeyValidationStatus,
+  ): KeyValidationStatus =
+    validateNodeKeyConsistency(
+      contractKeysToContractIds,
+      node,
+      validateNodeKeyUniqueness(node, keyValidationStatus),
+    )
```
SamirTalwar-DA, line 175: I would rather you extracted the `flatMap` up several levels, to decouple uniqueness checks from consistency checks as much as possible. They don't actually seem to be related apart from aggregating into the same class.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+          RejectionReason.Inconsistent("Causal monotonicity violated"),
+        ),
+      )
+  }
+
+  private sealed abstract class KeyValidationError()
+  private case object Duplicate extends KeyValidationError
+  private case object Inconsistent extends KeyValidationError
+
+  private case class KeyValidationState(
```
SamirTalwar-DA, line 102: If you were to make this class monoidal, where you could merge two of them together rather than always add to an existing one, it would be much easier to separate the distinct validation checks.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+      transactionCommitter.reject(
+        recordTime,
+        transactionCommitter.buildRejectionLogEntry(
+          transactionEntry,
+          RejectionReason.Inconsistent("Causal monotonicity violated"),
+        ),
+      )
+  }
+
+  private sealed abstract class KeyValidationError()
```
SamirTalwar-DA, line 98: Please move these classes to the bottom of the file.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+      val contractKeysToContractIds: Map[DamlContractKey, RawContractId] =
+        contractKeyDamlStateKeysToContractIds.map(m => m._1.getContractKey -> m._2)
+
+      validateTransactionContractKeysUniquenessAndConsistency(
+        transactionCommitter,
+        commitContext.recordTime,
+        transactionEntry,
+        contractKeyDamlStateKeys,
+        contractKeysToContractIds,
+      ) match {
```
SamirTalwar-DA, line 54: I'd find this much easier to read if you implemented `map` and `flatMap` on `StepResult`. I think it would also allow you to flatten out the logic, so that we could more easily follow through the steps.

SamirTalwar-DA (commented):
First pass done. I will try to do a second pass shortly.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: Right, I forgot to add a note about that: we were adding them before they were populated and it also seemed logically wrong to me, as the keys validation is supposed to work on the initial daml state.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
```diff
@@ -0,0 +1,295 @@
...
+}
+import com.daml.ledger.participant.state.kvutils.committer.{StepContinue, StepResult}
+import com.daml.ledger.participant.state.v1.RejectionReason
+import com.daml.lf.data.Ref.TypeConName
+import com.daml.lf.data.Time.Timestamp
+import com.daml.lf.transaction.{Node, NodeId}
+import com.daml.lf.value.Value
+import com.daml.lf.value.Value.ContractId
+
+object TransactionContractKeysValidation {
```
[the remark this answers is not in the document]
fabiotudone-da: This is really a split off from `TransactionCommitter` but I'll try to split it further. Did you have something specific in mind?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: I don't want to traverse the transaction twice though. I'll see what I can do.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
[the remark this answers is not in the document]
SamirTalwar-DA: I think that we could split uniqueness validation, consistency validation, and causal monotonicity validation into their own objects/classes. If it's harder than I think, please disregard this.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
SamirTalwar-DA: Right, makes sense.

SamirTalwar-DA: Can you prove that it's wrong? As in, can you write a test case that fails with the old code?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
SamirTalwar-DA: Makes sense. I was wondering if you could also push the `fold` up, but maybe it won't work.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: Adding outputs at this validation stage didn't really do anything anyway, as they are still empty in the context and set later, at the `blind` stage. The previous code was just convoluted and logically misleading, but it did the exact same thing.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: I am not sure, as the monotonicity check doesn't need to fold, but I'll have a look.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: Done.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: Done.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
fabiotudone-da: IMO forcing the monotonicity check (as any potential future keys check, if any) into the fold is only going to obfuscate it. I propose to have 2 (for now, and monadic) steps: monotonicity check and traversal-bound (aggregated) check. The traversal-bound check, in turn, has 2 (for now, and monadic) sub-steps: the uniqueness and consistency checks.

This is already the current situation; I am also going to turn the traversal-bound aggregated check state into a rather opaque monoid and that indeed makes the code (a bit) shorter and more pleasant to the eye, but that would be all. Would that be OK for you? Or what did you have in mind?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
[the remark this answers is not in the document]
fabiotudone-da: Done.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionContractKeysValidation.scala
SamirTalwar-DA: I think that sounds like a good compromise to me!

fabiotudone-da: Done, and also the split into smaller classes.

SamirTalwar-DA (approved):
If the tests are happy, I'm happy.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/KeyConsistencyValidation.scala
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+  private def checkKeyConsistency(
+      contractKeysToContractIds: Map[DamlContractKey, RawContractId],
+      key: Option[Node.KeyWithMaintainers[Value[ContractId]]],
+      targetContractId: Option[ContractId],
+      templateId: TypeConName,
+      keyValidationState: KeyValidationState,
+  ): KeyValidationStatus =
+    key.fold(Right(keyValidationState): KeyValidationStatus) {
```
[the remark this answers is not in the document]
fabiotudone-da: I'm not sure I understand what's your suggestion, can you expand?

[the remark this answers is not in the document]
SamirTalwar-DA: I'd prefer to read:

```scala
key match {
  case None => Right(keyValueState)
  case Some(submittedKeyWithMaintainers) => ...
}
```

Not a big deal though.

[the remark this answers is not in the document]
fabiotudone-da: Ah right. I agree it's clearer, done.

kamil-da (approved):
👏

PR #8595, opened 2021-01-21
faster cleanup on Linux

garyverhaegen-da:
I have seen the cleanup step taking up to an hour, just to delete the Bazel cache. This is obviously not great.

I've been researching how to do this faster, and it looks like the best approach here is to create a partition for the cache, and just drop it and reformat it when we want to clear it. This is a conceptually simple approach, but it unfortunately requires access to the `mount` command. There are a couple ways to achieve that, but after a bit of research it seems to me that the safest one is to have a setuid-root executable that normal users can call.

So this PR includes a new C program that's essentially a _safer_ version of a bash script that would do the same thing, somehow, because it has to run as root.

Please review carefully.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/vsts_agent_linux_startup.sh
+++ b/infra/vsts_agent_linux_startup.sh
@@ -25,7 +25,8 @@ apt-get install -qy \
   git \
   netcat \
   apt-transport-https \
-  software-properties-common
+  software-properties-common \
+  gcc
 
 # Install dependencies for Chrome (to run Puppeteer tests on the gsg)
 # list taken from: https://github.com/puppeteer/puppeteer/blob/a3d1536a6b6e282a43521bea28aef027a7133df8/docs/troubleshooting.md#chrome-headless-doesnt-launch-on-unix
@@ -94,6 +95,94 @@ useradd \
   vsts
 #add docker group to user
 usermod -aG docker vsts
+
+# The Bazel cache fills up the hard drive in less than a day, and is a real
+# pain to clean up, as it consists of a very large amount of read-only files in
+# a fairly large tree of read-only directories. Therefore, it looks like the
+# best way to manage it is to mount it on a partition and reset the partition.
+
+cat <<REMOUNT_CACHE > /root/remount_cache.c
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int exists(const char* path) {
+	struct stat s;
+	return stat(path, &s) == 0 && S_ISDIR(s.st_mode);
+}
+
+int execute_f(char** args) {
+	int status;
+	char** child_env = { NULL };
+	pid_t pid = fork();
+	if (0 == pid) {
+		// child process
+		if (-1 == execve(args[0], args, child_env)) {
+			printf("Failed to start child process calling %s.", args[0]);
+			perror(0);
+			exit(1);
+		}
+		// unreachable: if execve succeeded, we're off to another executable
+	}
+	// parent
+	if (-1 == pid) {
+		printf("Failed to fork.");
+		perror(0);
+		exit(1);
+	}
+	if (-1 == waitpid(pid, &status, 0)) {
+		printf("Failed to wait for child process '%s'.", args[0]);
+		exit(1);
+	}
+	if (1 != WIFEXITED(status)) {
+		printf("Child not terminated: '%s', aborting.", args[0]);
+		exit(1);
+	}
+	if (0 != WEXITSTATUS(status)) {
+		printf("Child '%s' exited with status %d", args[0], WEXITSTATUS(status));
+		exit(WEXITSTATUS(status));
+	}
+	return 0;
+}
+
+#define execute(...) execute_f((char*[]) { __VA_ARGS__, NULL })
+
+int main() {
+	char* mount_point = "/home/vsts/.cache/bazel";
+	char* lost = "/home/vsts/.cache/bazel/lost+found";
+	char* file = "/root/cache_file";
+
+	setuid(0);
+
+	if (exists(lost)) {
+		printf("Removing existing cache...");
+        // This will fail if there are dangling processes, say Bazel servers,
+        // with open fds on the cache.
+		execute("/bin/umount", mount_point);
+	}
+
+	execute("/bin/rm", "-f", file);
+    // truncate creates a sparse file. In my testing, for this use-case, there
+    // does not seem to be a significant overhead.
+	execute("/usr/bin/truncate", "-s", "200g", file);
+	execute("/sbin/mkfs.ext3", file);
+	execute("/bin/mount", file, mount_point);
+	execute("/bin/chown", "-R", "vagrant:vagrant", mount_point);
+	return 0;
+}
+REMOUNT_CACHE
+
+mkdir -p /home/vsts/.cache/bazel
+chown -R vsts:vsts /home/vsts/.cache
+( cd /root && gcc remount_cache.c -o /usr/local/bin/remount_cache )
+chmod ug+s /usr/local/bin/remount_cache
+/usr/local/bin/remount_cache
+
+# Done with the cache cleanup stuff.
+
 
 su --login vsts <<'AGENT_SETUP'
 set -euo pipefail
--- a/ci/clean-up.yml
+++ b/ci/clean-up.yml
@@ -8,15 +8,28 @@ steps:
     # infra/macos/2-common-box/init.sh:echo "build:darwin --disk_cache=~/.bazel-cache" > ~/.bazelrc
     # infra/vsts_agent_linux_startup.sh:echo "build:linux --disk_cache=~/.bazel-cache" > ~/.bazelrc
     if [ $(df -m . | sed 1d | awk '{print $4}') -lt 30000 ]; then
-        echo "Disk full, cleaning up..."
+        echo "$(date -Is) Disk full, cleaning up..."
         disk_cache="$HOME/.bazel-cache"
         rm -rf "$disk_cache"
-        echo "removed '$disk_cache'"
+        echo "$(date -Is) Removed '$disk_cache'."
         local_cache="$HOME/.cache/bazel"
-        if [ -d "$local_cache" ]; then
-            chmod -R +w "$local_cache"
-            rm -rf "$local_cache"
-            echo "removed '$local_cache'"
+        # This executable does not exist on macOS
+        if [ -f "/usr/local/bin/remount_cache" ]; then
+            echo "$(date -Is) Removing any dangling Bazel server..."
+            for pid in $(lsof "$local_cache" | sed 1d | awk '{print $2}' | sort -u); do
+                kill -s KILL $pid
+            done
+            echo "$(date -Is) Nuking the cache partition..."
+            /usr/local/bin/remount_cache
+            echo "$(date -Is) Done."
+        else
+            # This path does not exist on macOS
+            if [ -d "$local_cache" ]; then
+                echo "$(date -Is) Removing all files under '$local_cache'..."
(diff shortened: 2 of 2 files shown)


infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+	}
+	return 0;
+}
+
+#define execute(...) execute_f((char*[]) { __VA_ARGS__, NULL })
+
+int main() {
+	char* mount_point = "/home/vsts/.cache/bazel";
+	char* lost = "/home/vsts/.cache/bazel/lost+found";
+	char* file = "/root/cache_file";
```
cocreature, line 156:
```suggestion
	const char* mount_point = "/home/vsts/.cache/bazel";
	const char* lost = "/home/vsts/.cache/bazel/lost+found";
	const char* file = "/root/cache_file";
```
while const in C is not that great, it’s still better than no const imho.

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+
+	setuid(0);
+
+	if (exists(lost)) {
+		printf("Removing existing cache...");
```
cocreature, line 161: the formatting here is a bit of a mess, I would suggest to run clang-format on it.

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+        // This will fail if there are dangling processes, say Bazel servers,
+        // with open fds on the cache.
+		execute("/bin/umount", mount_point);
+	}
+
+	execute("/bin/rm", "-f", file);
```
cocreature, line 167: I admit that calling other binaries as root makes me feel very uneasy. Have you considered calling the corresponding syscalls/library functions in process to have a bit more control? I guess you can argue that reimplementing functionality is just as dangerous and those tools are more battleproven but at least you eliminate cases where someone manages to place a binary in `/bin` or `/usr/bin` and while the surface is not quite as battleproven it will be significantly smaller than those binaries.

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+		exit(1);
+	}
+	if (0 != WEXITSTATUS(status)) {
+		printf("Child '%s' exited with status %d", args[0], WEXITSTATUS(status));
+		exit(WEXITSTATUS(status));
+	}
+	return 0;
+}
+
+#define execute(...) execute_f((char*[]) { __VA_ARGS__, NULL })
```
cocreature, line 151: this seems a bit odd, why don’t we just declare `execute_f` as a variadic function?

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+	return stat(path, &s) == 0 && S_ISDIR(s.st_mode);
+}
+
+int execute_f(char** args) {
+	int status;
+	char** child_env = { NULL };
+	pid_t pid = fork();
+	if (0 == pid) {
+		// child process
+		if (-1 == execve(args[0], args, child_env)) {
```
cocreature, line 123: relying on a macro to enforce that this is not an out of bound scares me a bit. I guess varargs aren’t that much safer but at least more transparent than cpp.

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+		}
+		// unreachable: if execve succeeded, we're off to another executable
+	}
+	// parent
+	if (-1 == pid) {
+		printf("Failed to fork.");
+		perror(0);
+		exit(1);
+	}
+	if (-1 == waitpid(pid, &status, 0)) {
```
cocreature, line 136: do we have to worry about orphaning the child?

cocreature (commented):
A few initial comments on the code. It doesn’t seem entirely unreasonable but for obvious reasons I’d be much happier if we can avoid this so I’ll need to spend some time thinking about it to fully convince myself that this is worth doing.

cocreature:
Have you benchmarked some of the approaches in https://unix.stackexchange.com/questions/37329/efficiently-delete-large-directory-containing-thousands-of-files, e.g., the `rsync` one? Are we bottlenecked on the removal or the chmod?

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: Looking through the documentation on C varargs, I have not found any way to get at the number of arguments passed. Like so many other decisions in this PR, I can assure you going with a cpp macro was not my first choice, but it did seem to be the least-worst available.

Actually... All I need here is an executable. It will have to run as root, but perhaps it does not need to be written in C. How would you feel about this if it were written in Haskell?

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: Maybe. I don't expect this to run more than twice on the lifetime of a machine if everything goes well; if anything goew wrong, I don't expect we'll be able to use the machine (as its disk is already full), so one zombie process isn't going to make much of a difference.

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: The variadic function interface in C doesn't allow for argument counting.
```
execute("progname", "arg");
```
seemed safer and more readable than
```
execute_f(2, "progname", "arg1", "arg2"); // undefined behaviour because 2 < 3
```

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: I did consider that, yes, and I did look for those. However, given the interfaces I've seen, it seemed a lot safer (and vastly faster) for me to understand how to safely call a subprocess than for me to understand all the details of each syscall.

As this script does not take any user input, I feel pretty confident about its security. I'm more concerned about it crashing, but even that is not a huge concern as the end result is the same as not having the script (i.e. the machine is unusable and we have to kill it).

I've taken care to not have any user input in calling the child processes (not taking any user input, not going through a shell, not passing through any env var), so AFAIU the only real attack vector here is if someone manages to swap the actual binaries at these hardcoded locations. If someone is already in a position to write as root on the filesystem, they don't really need this script to take over the machine.

cocreature:
if we cannot make the removal fast enough, can we do something with fuse maybe?

garyverhaegen-da:
> Have you benchmarked some of the approaches in https://unix.stackexchange.com/questions/37329/efficiently-delete-large-directory-containing-thousands-of-files, e.g., the `rsync` one? Are we bottlenecked on the removal or the chmod?

Yes, I did come across that thread. On my Linux VM, after a single `bazel build //...`, the `chmod` call takes about 30s, the `rm -rf` one takes about 1 minute, and the `rsync` approach takes about 5 minutes. Interestingly the `rsync` approach doesn't seem to require a separate `chmod` step. The `perl` script only removes files in the current directory; it's not recursive as it is and I don't care to try and make it recursive; it looks already like line noise as it stands.

infra/vsts_agent_linux_startup.sh
cocreature, line 167: do we need to be root for anything apart from the `mount`?

garyverhaegen-da:
> if we cannot make the removal fast enough, can we do something with fuse maybe?

All I found about fuse was around mounting removable media, but happy to be wrong on that one. If you know how to make normal users able to mount files I'm all for it.

An alternative I have perhaps dismissed too early would be to actually have a separate partition. We can make that happen at the Terraform level (i.e. attach a second virtual disk), though I'm still not sure how we'd let a user format & mount it. Maybe in that context fuse could work, if it looks like a removable thing?

cocreature:
just to be clear: I’m sure you’ve consider most if not all the things I’m proposing. I just need some extra convincing here that custom C code running as root is really justified just for cleaning up disk space.

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: `mount` and `umount` need root access out of the box. I've not actually tried mounting a file owned by another user, so perhaps that could work. There may also be a flag I can pass to mount such that the mounter filesystem belongs to `vsts`, to avoid the `chmod` call... 🤔

cocreature:
maybe another option: if the file system is slow, maybe having this on a tmpfs partition makes it significantly faster? there is no value in this being permanent.

infra/vsts_agent_linux_startup.sh
garyverhaegen-da: Also just to be clear, I'm relying on the macro to ensure that the arg list is properly terminated (i.e. has a `NULL` as the last argument) without too much visual clutter in `execute` calls. I'm relying on myself to ensure bounds are not an issue. (I'm not claiming this is reassuring.)

garyverhaegen-da:
> maybe another option: if the file system is slow, maybe having this on a tmpfs partition makes it significantly faster? there is no value in this being permanent.

I did not experiment with filesystems. If you're thinking of RAM partitions, I don't think we can afford 150GB of RAM per build node. I've read some things that recommended turning off journaling on the filesystem, which could be an option to explore too, as we don't really care about loss of data should the machine unexpectedly shut down (we never restart a machine). I expect that would still be much slower than this, but perhaps it would feel safer.

cocreature:
I think I got confused. I thought you could run tmpfs backed only by a swap partition and no actual ram but I think I got that wrong looking at it again.

cocreature:
On fuse, I believe the APIs would allow you to implement a file-backed fuse filesystem but I can’t find an existing one and maintining our own fuse filesystem for faster cache cleanups seems very silly so discard that idea.

infra/vsts_agent_linux_startup.sh
```diff
@@ -95,6 +96,94 @@ useradd \
...
+		execute("/bin/umount", mount_point);
+	}
+
+	execute("/bin/rm", "-f", file);
+    // truncate creates a sparse file. In my testing, for this use-case, there
+    // does not seem to be a significant overhead.
+	execute("/usr/bin/truncate", "-s", "200g", file);
+	execute("/sbin/mkfs.ext3", file);
+	execute("/bin/mount", file, mount_point);
+	execute("/bin/chown", "-R", "vagrant:vagrant", mount_point);
```
cocreature, line 173: why do we chown to vagrant? I thought we just run directly on the vm on linux?

cocreature:
Having thought about this, I think there are two ways that should accomplish the desired performance goals while reducing the amount of setuid code.

1. Keep the approach here but limit the binary to only `mount`, ideally via the library function rather than `execve` (the mount API seems pretty simple). That should hopefully reduce this to a 10 line main function or so which seems a bit easier to justify.
2. Don’t do the mount dance. Instead, move the directory away to a trash directory. Across the same filesystem, moving is more or less instant. We can then have a long-running daemon on CI machines that cleans up this trash directory. There are some open questions here on what you do if the daemon is not fast enough but I _think_ those can be solved.

If we can make 1 work, it seems like a pretty simple solution so I’d try that first.

aherrmann-da:
Re FUSE, IIUC [`fuse2fs`](https://man7.org/linux/man-pages/man1/fuse2fs.1.html) could be useful here. E.g.
```
$ truncate -s SIZE image
$ mkfs.ext2 -E root_owner=$UID:$GID image
$ fuse2fs image mnt
```
Will create an image file of size `SIZE`, format it as ext2 (could use another fs if preferred), and FUSE mounts it under `mnt`.
When the threshold is met we `umount mnt; rm image` and create a fresh image.
Would that work?

cocreature:
oh nice, fuse2fs looks like it should do the job.

garyverhaegen-da:
Closing in favour of #8599.

PR #8611, opened 2021-01-22
Drop early access tag on trigger service and auth middleware

aherrmann-da:
- Drops the early access tag from Daml triggers, the trigger service, and the auth middleware
- Adds the OAuth 2.0 middleware to the SDK tarball and daml-assistant
- Adds a simple daml-assistant integration test for the OAuth 2.0 middleware (similar to the one for the trigger service)
- Adds a `/livez` endpoint to the OAuth 2.0 middleware for the integration test
- Unhides the auth flags on the trigger service.

### Pull Request Checklist

trigger the build.


daml-assistant/daml-sdk/src/main/scala/com/digitalasset/daml/sdk/SdkMain.scala
```diff
@@ -25,6 +26,7 @@ object SdkMain {
       case "extractor" => Extractor.main(rest)
       case "json-api" => JsonApi.main(rest)
       case "trigger-service" => TriggerService.main(rest)
+      case "oauth2-middleware" => Oauth2Middleware.main(rest)
```
cocreature, line 29: we should also publish standalone JARs for the trigger service and the middleware on GH releases, no need to include it in this PR.

cocreature (approved):
:tada:

PR #230, opened 2019-04-04
add more Azure Pipeline agents

zimbatm:
This PR adds our own Azure Pipeline Linux agents to the infrastructure.

PR #8617, opened 2021-01-25
Upgrade linux nodes to 20.04

garyverhaegen-da:
For ease of review, first commit is just a copy of the existing Linux files; further commits will modify those files for the upgrade.

CHANGELOG_BEGIN

- Our Linux binaries are now built on Ubuntu 20.04 instead of 16.04. We do not expect any user-level impact, but please reach out if you do notice any issue that might be caused by this.

CHANGELOG_END

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,7 +1,7 @@
 # Bazel distributed cache, can be temporarily disabled by passing the following
 # flag: --noremote_accept_cached
 build:darwin --remote_http_cache=https://bazel-cache.da-ext.net
-build:linux --remote_http_cache=https://bazel-cache.da-ext.net
+build:linux --remote_http_cache=https://bazel-cache.da-ext.net/ubuntu_20_04
 build --remote_upload_local_results=false
 # Enable the disk cache in addition to the http cache.
 build:linux --disk_cache=.bazel-cache/disk
--- a/dev-env/windows/manifests/bazel.json
+++ b/dev-env/windows/manifests/bazel.json
@@ -5,8 +5,8 @@
   "bin": "bazel.exe",
   "architecture": {
     "64bit": {
-      "url": "https://daml-binaries.da-ext.net/patch_bazel_windows/bazel-e0202b0f2f3c2d8c3b477ddb59bd30ce.zip",
-      "hash": "7177fbc5660626ea68713d6fab026581deced2e20aad0723e2b5ed3e2eb1f2dc"
+      "url": "https://daml-binaries.da-ext.net/patch_bazel_windows/bazel-f6f61bf596011e671cb3b327b0f29686.zip",
+      "hash": "5b9224a94da96403bffd0309e011d63504807386718d806d0aaf769f674626f5"
     }
   },
   "depends": [
--- a/infra/vsts_agent_linux.tf
+++ b/infra/vsts_agent_linux.tf
@@ -1,8 +1,6 @@
 # Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-resource "secret_resource" "vsts-token" {}
-
 data "template_file" "vsts-agent-linux-startup" {
   template = "${file("${path.module}/vsts_agent_linux_startup.sh")}"
 
--- a/infra/vsts_agent_ubuntu_20_04.tf
+++ b/infra/vsts_agent_ubuntu_20_04.tf
@@ -1,28 +1,26 @@
 # Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-resource "secret_resource" "vsts-token" {}
-
-data "template_file" "vsts-agent-linux-startup" {
-  template = "${file("${path.module}/vsts_agent_linux_startup.sh")}"
+data "template_file" "vsts-agent-ubuntu_20_04-startup" {
+  template = "${file("${path.module}/vsts_agent_ubuntu_20_04_startup.sh")}"
 
   vars = {
     vsts_token   = "${secret_resource.vsts-token.value}"
     vsts_account = "digitalasset"
-    vsts_pool    = "linux-pool"
+    vsts_pool    = "ubuntu_20_04"
   }
 }
 
-resource "google_compute_region_instance_group_manager" "vsts-agent-linux" {
+resource "google_compute_region_instance_group_manager" "vsts-agent-ubuntu_20_04" {
   provider           = "google-beta"
-  name               = "vsts-agent-linux"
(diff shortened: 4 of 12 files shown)


.bazelrc
```diff
@@ -1,7 +1,7 @@
 # Bazel distributed cache, can be temporarily disabled by passing the following
 # flag: --noremote_accept_cached
 build:darwin --remote_http_cache=https://bazel-cache.da-ext.net
-build:linux --remote_http_cache=https://bazel-cache.da-ext.net
+build:linux --remote_http_cache=https://bazel-cache.da-ext.net/ubuntu_20_04
```
garyverhaegen-da, line 4: Is this enough to ensure I don't pollute our existing linux cache?

cocreature: not that worried about this tbh. I trust nix much more in this regard than Windows :)

cocreature (approved):
Thanks! I probably wouldn’t isolate caches at all for this but I guess it doesn’t do any harm either.

garyverhaegen-da:
> Thanks! I probably wouldn’t isolate caches at all for this but I guess it doesn’t do any harm either.

On the one hand, I'm not entirely sure I trust Bazel with this; on the other hand, I'd feel better if I could assume everything has been rebuilt from scratch using 20.04, rather than rely on cached items from 16.04.

Either way, first step is going to be to get this thing to boot at all.

aherrmann-da (approved):
Thanks, LGTM. I agree that the separate cache shouldn't be necessary. But, it seems sensible to use it to check that everything builds from a clean slate.

dev-env/windows/manifests/bazel.json
```diff
@@ -5,8 +5,8 @@
   "bin": "bazel.exe",
   "architecture": {
     "64bit": {
-      "url": "https://daml-binaries.da-ext.net/patch_bazel_windows/bazel-e0202b0f2f3c2d8c3b477ddb59bd30ce.zip",
-      "hash": "7177fbc5660626ea68713d6fab026581deced2e20aad0723e2b5ed3e2eb1f2dc"
+      "url": "https://daml-binaries.da-ext.net/patch_bazel_windows/bazel-f6f61bf596011e671cb3b327b0f29686.zip",
+      "hash": "5b9224a94da96403bffd0309e011d63504807386718d806d0aaf769f674626f5"
```
garyverhaegen-da, line 9: No change to the Bazel base commit or the patch we apply, but we're now pinning the Bazel version used to build Bazel. I don't know which version was used before ("whatever Azure provided by default at that time"), but even though the binary is different (AFAIK this build is not reproducible), I don't expect any issue from this.

infra/vsts_agent_linux.tf
```diff
@@ -1,8 +1,6 @@
 # Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-resource "secret_resource" "vsts-token" {}
```
garyverhaegen-da, line 4: Line moved to `infra/main.tf` as this resource is used by other files.

infra/vsts_agent_ubuntu_20_04.tf
```diff
@@ -0,0 +1,80 @@
...
+    vsts_pool    = "ubuntu_20_04"
+  }
+}
+
+resource "google_compute_region_instance_group_manager" "vsts-agent-ubuntu_20_04" {
+  provider           = "google-beta"
+  name               = "vsts-agent-ubuntu-20-04"
+  base_instance_name = "vsts-agent-ubuntu-20-04"
+  region             = "us-east1"
+  target_size        = 1
```
garyverhaegen-da, line 19: Once that gets merged I'll be redeploying the tf file with a 10 here (and make a new PR for that, and the git checkout thing).

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -0,0 +1,198 @@
...
+# around, even when their caches are only warmed up halfway
+su --login vsts <<'CACHE_WARMUP'
+# user-wide bazel disk cache override
+echo "build:linux --disk_cache=~/.bazel-cache" > ~/.bazelrc
+
+# clone and build
+(
+  git clone https://github.com/digital-asset/daml
+  cd daml
+  git checkout upgrade-linux-nodes-to-20.04
```
garyverhaegen-da, line 178: This is not required once this PR is merged into main, so I'll be redeploying the cluster without this line. Note that this only runs when we rebuild a machine, so a bit of a lag here is acceptable.

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -7,7 +7,7 @@ set -euo pipefail
 
 ## Hardening
 
-# Commit harakiri on failure
+# Commit sepukku on failure
```
cocreature, line 10: crucial change :slightly_smiling_face:

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -21,7 +21,7 @@ apt-get update -q
 apt-get install -qy \
   curl sudo \
   bzip2 rsync \
-  jq liblttng-ust0 libcurl3 libkrb5-3 libicu55 zlib1g \
+  jq liblttng-ust0 libcurl4 libkrb5-3 zlib1g \
```
cocreature, line 24: Do you know what we needed icu for before and why we don’t need it anymore? I know GHC uses it but we should get that from nix.

cocreature (approved):
Thank you looks good!

infra/vsts_agent_ubuntu_20_04_startup.sh
garyverhaegen-da: I do not.

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -41,6 +41,7 @@ apt-get install -qy \
     libdbus-1-3 \
     libexpat1 \
     libfontconfig1 \
+    libgbm-dev \
```
garyverhaegen-da, line 44: This was required to get puppeteer to run.

infra/vsts_agent_ubuntu_20_04_startup.sh
garyverhaegen-da: I was not very satisfied with my answer here so I dug a little bit more. Unfortunately that list of dependencies has been added in one chunk with no explanation and nobody asking any question about it, back in #230. Googling for that specific list of dependencies seem to bring up mostly DotNetCore-related hits, so my best guess at this point is that these were the dependencies needed to get the Azure agent working. (It's written in C#.)

Note that I've removed the `libicu55` one because it was causing a problem (not found anymore), i.e. I have not double-checked that the other ones are still necessary.

PR #8619, opened 2021-01-25
damlc: check for inconsistent dependencies

### LINKED ISSUE #8553: damlc should error out if `dependencies` have a different LF version
We already have a check that `dependencies` and `data-dependencies` have a newer LF version. For `dependencies` we should error out if they have a different LF version regardless of whether it is newer or older.

robin-da:
This fixes #8553.

We check that all dependencies have the expected LF version.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
@@ -547,7 +547,8 @@ checkForInconsistentLfVersions lfTarget DependencyInfo{dalfsFromDependencies, ma
     inconsistentLfDeps =
         [ ((LF.dalfPackageId decodedDalfPkg, decodedUnitId), ver)
         | DecodedDalf {..} <- dalfsFromDependencies
-        , decodedUnitId `elem` mainUnitIds
+        , let mainUnitIdsSet = Set.fromList mainUnitIds
+        , decodedUnitId `Set.member` mainUnitIdsSet
         , let ver = LF.packageLfVersion $ LF.extPackagePkg $ LF.dalfPackagePkg decodedDalfPkg
         , ver /= lfTarget
         ]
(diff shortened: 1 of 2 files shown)


compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -522,18 +525,42 @@ data DependencyInfo = DependencyInfo
...
+        , showDeps inconsistentLfDeps
+        ]
+  where
+    inconsistentLfDeps =
+        filter (\(_, ver) -> ver /= lfTarget) $
+            [ ( (LF.dalfPackageId decodedDalfPkg, decodedUnitId)
+              , (LF.packageLfVersion . LF.extPackagePkg . LF.dalfPackagePkg) decodedDalfPkg
+              )
+            | DecodedDalf{..} <- dalfsFromDependencies
+            , decodedUnitId `elem` mainUnitIds
```
cocreature, line 553: Let’s turn this into a `Set`. The linear lookup seems like we’re just asking to end up on accidentallyquadratic :)

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -522,18 +525,42 @@ data DependencyInfo = DependencyInfo
...
+  | null inconsistentLfDeps = Right ()
+  | otherwise = Left $ concat
+        [ "Targeted LF version "
+        , DA.Pretty.renderPretty lfTarget
+        , " but dependencies have different LF versions: "
+        , showDeps inconsistentLfDeps
+        ]
+  where
+    inconsistentLfDeps =
+        filter (\(_, ver) -> ver /= lfTarget) $
```
cocreature, line 548: can we put the filter in the guard of the list comprehension?

cocreature (approved):
Nice, thank you!

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: You want to turn the `mainUnitIds` field into a `Set` or just here for the lookup?

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: done.

PR #8621, opened 2021-01-25
Revert "fix: package locale-archive with binaries"

robin-da:
Reverts digital-asset/daml#8609


cocreature (approved):
Thanks! Sorry for the back and forth, I had no idea the locale archives are so large

robin-da:
np, I wasn't aware either.

PR #8623, opened 2021-01-25
triggers: do not use the type information from SRecord

remyhaemmerle-da:
This type information is in `SRecord` for debugging purpose, one should not
use it.  We use type from `SAny` instead.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/triggers/runner/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Converter.scala
+++ b/triggers/runner/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Converter.scala
@@ -69,10 +69,13 @@ case class TriggerIds(val triggerPackageId: PackageId) {
     )
 }
 
-case class AnyContractId(templateId: Identifier, contractId: ContractId)
-
 object Converter {
   import com.daml.script.converter.Converter._, Implicits._
+
+  private case class AnyContractId(templateId: Identifier, contractId: ContractId)
+  private case class AnyTemplate(ty: Identifier, arg: SValue)
+  private case class AnyChoice(name: ChoiceName, arg: SValue)
+  private case class AnyContractKey(key: SValue)
 
   private def toLedgerRecord(v: SValue): Either[String, value.Record] =
     lfValueToApiRecord(true, v.toValue)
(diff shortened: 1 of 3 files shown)


triggers/runner/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Converter.scala
```diff
@@ -348,50 +339,39 @@ object Converter {
       } yield AnyContractId(templateId, contractId)
     })
 
-  private def toAnyTemplate(v: SValue): Either[String, SValue] = {
+  private[this] def toAnyTemplate(v: SValue): Either[String, (TypeConName, SValue)] = {
```
cocreature, line 342: can we turn this into a case class, something like `TemplateValue` and `ChoiceArg` for the other one? That should remove some the pattern matching weirdness since we can just access the fields with reasonable names.

triggers/runner/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Converter.scala
```diff
@@ -348,50 +339,39 @@ object Converter {
...
   }
 
-  // toAnyChoice and toAnyContractKey are identical right now
-  // but there is no resaon why they have to be, so we
-  // use two different methods.
-  private def toAnyChoice(v: SValue): Either[String, SValue] =
-    v expect ("AnyChoice", { case SRecord(_, _, JavaList(SAny(_, v), _)) =>
-      v
+  private[this] def toAnyChoice(v: SValue): Either[String, (String, SValue)] =
+    v expect ("AnyChoice", { case SRecord(_, _, JavaList(SAny(TTyCon(tmplId), value))) =>
```
cocreature, line 351: This is the choice argument type not the template id and what you return is the choice name (we exploit the fact that damlc enforces that choice arguments are record types and the name of those records is identical to the choice name). It would be nice to just return a `ChoiceName` from this to make it a bit clearer what the string is.

cocreature (approved):
nice, thank you!

triggers/runner/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Converter.scala
cocreature: We already have a shared library in `//daml-script/converter`, feel free to move more things there either in this PR or separately.

PR #8626, opened 2021-01-26
LF: Move transaction validation is in is own object.

remyhaemmerle-da:
+ cleanup
 + rewrite value comparison. Do not use Equality as it may not be symetric.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -0,0 +1,342 @@
...
+                } &&
+                loop((values(recordedFields) zip values(recordedFields)) ++: rest)
+            case (
+                  ValueVariant(recordedTyCon, recordedVariant, recordedValue),
+                  ValueVariant(replayedTyCon, replayedVariant, replayedValue),
+                ) =>
+              recordedTyCon == replayedTyCon &&
+                recordedVariant == replayedVariant &&
+                loop((recordedValue, replayedValue) +: rest)
+            case (ValueList(recordedValues), ValueList(replayedValues)) =>
```
cocreature, line 89: I know this didn’t change but I don’t understand why we don’t have to check the length here.

cocreature (approved):
lgtm assuming this just shuffling things around

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
remyhaemmerle-da: It is in fact changed (I update the PR description).
We have to check the length (you already catch that in another PR that I try to slice apart)
Thanks.

cocreature (approved):
Thanks! Took another look given that there have been changes. My only worry is still the lack of testing but you’re aware of that. I think it’s probably a good idea to prioritize that once the type erasure changes have landed.

PR #8629, opened 2021-01-26
fix: daml-extension: turn breadcrumbs off.

### LINKED ISSUE #8410: DAML Studio shows DA.Internal.Record in breadcrumb
I don't know when daml studio started showing this "breadcrumb" but it always seems to point to DA.Internal.Record:

image

This seems undesirable. I wonder if we can fix the breadcrumb, or disable it.

robin-da:
This fixes #8410.

This turns the symbol breadcrumbs off for daml workspaces, because they
point to internals.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-extension/src/extension.ts
+++ b/compiler/daml-extension/src/extension.ts
@@ -33,6 +33,8 @@ var damlLanguageClient: LanguageClient;
 // and then `Toggle Developer Tools` in VSCode. This will show
 // output in the Console tab once the extension is activated.
 export async function activate(context: vscode.ExtensionContext) {
+    // Turn symbol breadcrumbs off because they point to internals.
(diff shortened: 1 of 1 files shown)


robin-da:
Would be nice if someone could test this. It seems to work for me locally but I'm not sure if it's just because I turned that preference off and the it got persisted.

cocreature (commented):
This seems a bit backwards. Rather than trying to hide it, should we instead disable the ghcide request which is used to produce those breadcrumbs in the first place? I think it might be document outline but haven’t double checked.

cocreature:
Or even better, fix it to produce a useful outline.

robin-da:
I thought fixing might be fairly difficult, but on second thought, maybe we can try to filter the symbols and only show non-internal ones. I'll give it a try.

PR #8558, opened 2021-01-19
Upgrade grpc and protobuf libraries

cocreature:
as requested by @danilofaria. The one thing still missing from this is the gRPC C++ libraries. Those are only used in our Haskell libs internally so while we should upgrade that, I care a bit less about doing that now.

changelog_begin

- [Dependencies] Upgrade
  iou.grpc:grpc-(api|core|netty|protobuf|services|stub) to
  1.35.0.

- [Dependencies] Upgrade
  io.netty:netty-(codec-http2|handler|handler-proxy|resolver)
  to 4.1.58.Final.

- [Dependencies] Upgrade
  io.netty:netty-tcnative-boringssl-static to 2.0.36.Final.

- [Dependencies] Upgrade com.google.protobuf:protobuf-java to 3.14.0.

changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8638, opened 2021-01-26
Add test to verify that TLSv1.2 and TLSv1.3 work

stefanobaghino-da:
As raised here: https://discuss.daml.com/t/what-version-of-tls-is-supported-by-the-ledger-api/1982

This has been effectively fixed by upgrading Netty here: https://github.com/digital-asset/daml/pull/8558

Note that this is an integration test for sandbox-classic only, but the
infrastructure used is shared with other Ledger API server
implementations, so this should be enough.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/TlsConfiguration.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/TlsConfiguration.scala
@@ -8,7 +8,7 @@ import java.io.File
 import io.grpc.netty.GrpcSslContexts
 import io.netty.handler.ssl.{ClientAuth, SslContext}
 
-import scala.collection.JavaConverters.asJavaIterableConverter
+import scala.jdk.CollectionConverters._
 
 final case class TlsConfiguration(
     enabled: Boolean,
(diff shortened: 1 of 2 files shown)


ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/TlsConfiguration.scala
```diff
@@ -40,6 +43,7 @@ final case class TlsConfiguration(
           .forClient()
           .keyManager(keyCertChainFile.orNull, keyFile.orNull)
           .trustManager(trustCertCollectionFile.orNull)
+          .protocols(if (protocols.nonEmpty) protocols.asJava else null)
```
cocreature, line 46: I assume `protocols(null)` is equivalent to not calling `protocols` and not equivalent to an empty list?

cocreature (approved):
Thanks!

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/TlsConfiguration.scala
stefanobaghino-da: Unfortunately that's the case. Passing an empty collection will cause all hell to break loose when using this configuration. Using `null` means that you will use the default.

PR #8642, opened 2021-01-27
Correctly document and report malformed party names when allocating

stefanobaghino-da:
As reported here: https://discuss.daml.com/t/issues-allocating-parties-with-large-number-of-characters/1992

changelog_begin
[Ledger API] Documented the hard-coded limit of 255 characters for Daml-LF party names.
[Ledger API] Malformed party names are now correctly reported back with an INVALID_ARGUMENT error
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
@@ -61,7 +61,7 @@ service PartyManagementService {
   // canton: completely different globally unique identifier is allocated.
   // Behind the scenes calls to an internal protocol are made. As that protocol
   // is richer than the surface protocol, the arguments take implicit values
-  // Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
+  // The party identifier suggestion must be a valid party name. Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
   // colon, minus and underscore limited to 255 chars
   rpc AllocateParty (AllocatePartyRequest) returns (AllocatePartyResponse);
 }
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
@@ -61,6 +61,8 @@ service PartyManagementService {
   // canton: completely different globally unique identifier is allocated.
   // Behind the scenes calls to an internal protocol are made. As that protocol
   // is richer than the surface protocol, the arguments take implicit values
+  // Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
+  // colon, minus and underscore limited to 255 chars
   rpc AllocateParty (AllocatePartyRequest) returns (AllocatePartyResponse);
 }
 
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
@@ -110,6 +110,24 @@ final class PartyManagementServiceIT extends LedgerTestSuite {
   })
 
   test(
+    "PMRejectInvalidPartyHints",
+    "A party identifier that contains invalid characters should be rejected with the proper error",
+    allocate(NoParties),
+  )(implicit ec => { case Participants(Participant(ledger)) =>
+    for {
+      error <- ledger
+        .allocateParty(
+          // Assumption: emojis will never be acceptable in a party identifier
+          partyIdHint = Some("\uD83D\uDE00"),
+          displayName = None,
+        )
+        .failed
+    } yield {
+      assertGrpcError(error, Status.Code.INVALID_ARGUMENT, "non expected character")
+    }
+  })
+
+  test(
     "PMAllocateOneHundred",
     "It should create unique party names when allocating many parties",
     allocate(NoParties),
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
@@ -4,11 +4,13 @@
 package com.daml.ledger.api.testtool.suites
 
 import com.daml.ledger.api.testtool.infrastructure.Allocation._
+import com.daml.ledger.api.testtool.infrastructure.Assertions.assertGrpcError
 import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
 import com.daml.ledger.api.v1.admin.party_management_service.PartyDetails
 import com.daml.ledger.client.binding
 import com.daml.ledger.test.model.Test.Dummy
 import com.daml.lf.data.Ref
+import io.grpc.Status
 import scalaz.Tag
 import scalaz.syntax.tag.ToTagOps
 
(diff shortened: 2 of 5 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
```diff
@@ -81,30 +81,41 @@ private[apiserver] final class ApiPartyManagementService private (
...
+
+    val validatedPartyIdentifier =
+      if (request.partyIdHint.isEmpty) {
+        Future.successful(None)
+      } else {
+        Ref.Party
+          .fromString(request.partyIdHint)
+          .fold(
+            error => Future.failed(ErrorFactories.invalidArgument(error)),
+            party => Future(Some(party)),
```
rautenrieth-da, line 93:
```suggestion
            party => Future.successful(Some(party)),
```
?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
```diff
@@ -81,30 +81,41 @@ private[apiserver] final class ApiPartyManagementService private (
...
           )
-        )
       }
+
+    // Submission identifiers are restricted to 255 characters just like party identifiers
+    // so they cannot include the party name to prevent long strings to cause errors
+    lazy val submissionId = v1.SubmissionId.assertFromString(UUID.randomUUID().toString)
```
rautenrieth-da, line 99: IIRC (cc @mziolekda), the prefix was added to the submission ID to facilitate analyzing logs. Could we take up to (255 - length of a UUID) characters from the party name, and append the UUID to it?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
```diff
@@ -90,6 +92,23 @@ final class PartyManagementServiceIT extends LedgerTestSuite {
     }
   })
 
+  test(
```
rautenrieth-da, line 95: Maybe also test that party identifiers containing invalid characters are rejected (with the correct description)?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
```diff
@@ -61,6 +61,8 @@ service PartyManagementService {
   // canton: completely different globally unique identifier is allocated.
   // Behind the scenes calls to an internal protocol are made. As that protocol
   // is richer than the surface protocol, the arguments take implicit values
+  // Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
+  // colon, minus and underscore limited to 255 chars
```
rautenrieth-da, line 65:
```suggestion
  // The party identifier suggestion must be a valid party name. Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
  // colon, minus and underscore limited to 255 chars
```
This might not be obvious, given that the ledger is allowed to arbitrarily change the suggested identifier.

stefanobaghino-da:
This is actually an interesting problem. A compatibility test is failing because older versions cannot pass this test due to the fix.

```
### FAILURES

PartyManagementServiceIT
- [PartyManagementServiceIT:PMRejectLongPartyHints] A party identifier which is too long should be rejected with the proper error ... Assertion failed at line 108
  Expected code [INVALID_ARGUMENT], but got [INTERNAL].
```

How should this be addressed? More lenient tests? Or can we somehow whitelist this change?

rautenrieth-da (approved):
Looks good, but please check whether prefixing the submission ID with the party identifier is something that should be kept.

rautenrieth-da:
> How should this be addressed? More lenient tests? Or can we somehow whitelist this change?

Which ledger is that? This should only happen if you use a new version of the test tool against a ledger built with an old version of the SDK, right?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
stefanobaghino-da: Truncating the party identifier seems reasonable, I'll do it.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
stefanobaghino-da: Good catch, thanks.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
stefanobaghino-da: Makes sense, thanks!

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
stefanobaghino-da: Absolutely, I'll add one, thanks for pointing this out.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
stefanobaghino-da: To be more specific: truncating only for the submission identifier.

stefanobaghino-da:
> > How should this be addressed? More lenient tests? Or can we somehow whitelist this change?
> 
> Which ledger is that? This should only happen if you use a new version of the test tool against a ledger built with an old version of the SDK, right?

As @cocreature made me notice, our compatibility tests have [exclusion lists](https://github.com/digital-asset/daml/blob/9fff5027f6c2ea87b2f3562f9450fe29d0a0b64a/compatibility/bazel_tools/testing.bzl#L28) to ensure we can whitelists cases like this in which we are fixing a bug.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto
```diff
@@ -61,6 +61,8 @@ service PartyManagementService {
   // canton: completely different globally unique identifier is allocated.
   // Behind the scenes calls to an internal protocol are made. As that protocol
   // is richer than the surface protocol, the arguments take implicit values
+  // The party identifier suggestion must be a valid party name. Party names are required to be non-empty US-ASCII strings built from letters, digits, space,
```
cocreature, line 64: is this actually true in general? I know it applies to Sandbox because it literally tries to use the hint as the party id but does it apply to canton, corda, …?

cocreature: oh nvm, I missed the code below. Seems fine to me then.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8642/commits/788c43133a5236aa210e7b61cc298acfdc15aab4

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/PartyManagementServiceIT.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8642/commits/8abe8ee690a8d192fb9634430242e9cb2f36a0e2

@remyhaemmerle-da Is there something else I should use as a test case to ensure we're future proof?

remyhaemmerle-da: I think it is fine like that.

dasormeter:
@stefanobaghino-da we are still allowing spaces in party names, correct?

stefanobaghino-da:
> @stefanobaghino-da we are still allowing spaces in party names, correct?

Nothing changed, only the error we return (`INVALID_ARGUMENT` instead of `INTERNAL`).

stefanobaghino-da:
@oliverse-da FYI I had to exclude these new tests for Canton (see https://github.com/digital-asset/daml/pull/8642/commits/de7eefe8498e995d4759185b609a1a56d38f140c). Right now Canton returns `INTERNAL` instead of `INVALID_ARGUMENT`, you can have a look at this PR for what I imagine could be the fix your may want to apply.

oliverse-da:
> @oliverse-da FYI I had to exclude these new tests for Canton (see [de7eefe](https://github.com/digital-asset/daml/commit/de7eefe8498e995d4759185b609a1a56d38f140c)). Right now Canton returns `INTERNAL` instead of `INVALID_ARGUMENT`, you can have a look at this PR for what I imagine could be the fix your may want to apply.

Thanks, was able to enable these tests with the upgrade to canton 0.21.0

PR #8644, opened 2021-01-27
LF: cleanup value.rst and transaction.rst

remyhaemmerle-da:
Clean up as of version 11.

We will update as of version 12 in a upcoming PR. 

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/spec/transaction.rst
+++ b/daml-lf/spec/transaction.rst
@@ -5,7 +5,7 @@
 Daml-LF Transaction Specification
 =================================
 
-**version 11, 21 January 2019**
+**version 11, 21 January 2021**
 
 This specification, in concert with the ``transaction.proto``
 machine-readable definition, defines a format for *Daml LF
--- a/daml-lf/spec/transaction.rst
+++ b/daml-lf/spec/transaction.rst
@@ -5,7 +5,7 @@
 Daml-LF Transaction Specification
 =================================
 
-**version 10, 25 March 2020**
+**version 11, 21 January 2019**
 
 This specification, in concert with the ``transaction.proto``
 machine-readable definition, defines a format for *Daml LF
@@ -107,12 +107,12 @@ define some semantics vital to understanding your transaction.
 "since version"
 ~~~~~~~~~~~~~~~
 
-Every message type and field is accompanied by one or more *since
-version x* annotations in this document, preceding some description of
-semantics.  This defines when that message, field, or semantic rule
+Every message type and field is accompanied by one or more *(since
+version x)* annotations in this document, preceding some description
+of semantics.  This defines when that message, field, or semantic rule
 was introduced in the transaction specification.  Where there are
 multiple overlapping definitions of semantics for the same field of
-the form *since version x*, the correct interpretation for any
+the form *(since version x)*, the correct interpretation for any
 transaction of version *y* is that under the greatest *x* such that
 *x≤y*, failing that the second-greatest such *x*, and so on.
 
@@ -185,10 +185,10 @@ later.
 message Transaction
 ^^^^^^^^^^^^^^^^^^^
 
-*since version 10*
-
 A list of `message Node`_, implicitly forming a forest starting at
 ``roots``.
+
+(*since version 10*)
 
 As of version 10, these fields are included:
 
@@ -196,33 +196,31 @@ As of version 10, these fields are included:
 * repeated ``string`` roots
 * repeated `message Node`_ nodes
 
-``version`` is required and constrained as described under `field version`_.  Consumers can expect this field to be present and to
-have the semantics defined here without knowing the version of this
-value in advance.
+``version`` is required and constrained as described under `field
+version`_.  Consumers can expect this field to be present and to have
+the semantics defined here without knowing the version of this value
+in advance.
 
 ``roots`` is constrained as described under `field node_id`_.
-
-As of version 11, the
 
 field version
 ~~~~~~~~~~~~~
 
-``version`` and must be a version of this specification.
-For example, for version 11 of this specification, ``version`` must be
-``"11"``
+``version`` and must be a version of this specification.  For example,
+for version 11 of this specification, ``version`` must be ``"11"``
 
-Known versions are listed in ascending order in `Version history`_; any
-``version`` not in this list should be considered newer than any version
-in same list, and consumers must reject values with such unknown
-versions.
+Known versions are listed in ascending order in `Version history`_;
+any ``version`` not in this list should be considered newer than any
+version in same list, and consumers must reject values with such
+unknown versions.
 
 message ContractInstance
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-*since version 10*
-
 An instance of a Daml-LF template, represented by the Daml-LF value used
 to construct that instance.
+
+(*since version 10*)
 
 As of version 10, these fields are included:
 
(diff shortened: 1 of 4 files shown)


daml-lf/spec/transaction.rst
```diff
@@ -232,12 +230,7 @@ As of version 10, these fields are included:
 
 ``template_id`` and ``value`` are required; ``agreement`` is optional.
 
-``value`` must conform to the type of the Daml-LF associated data type
-indicated by ``template_id``.
-
-``template_id``'s structure is defined by `the value specification`_;
-the version of that specification to use when consuming it is the
-``version`` field of ``value``.
```
remyhaemmerle-da, line 240: I dropped everything that is relative to type. This is handle by type checking not (de)serialization processes.

daml-lf/spec/transaction.rst
```diff
@@ -434,27 +417,21 @@ As of version 10, these fields are included:
 * `message VersionedValue`_ return_value
 * `message KeyWithMaintainers`_ key_with_maintainers
 
-``contract_id_struct`` is required. Its structure is defined by `the value
-specification`_, version 3.
```
remyhaemmerle-da, line 438: should have be dropped when we dropped everything older than 10

daml-lf/spec/transaction.rst
cocreature: what happened to the version field of value? Don’t we still need that for version 10?

daml-lf/spec/transaction.rst
```diff
@@ -348,36 +344,28 @@ specification`_.
...
   The stakeholders of a contract are the signatories and the observers of
   said contract.
 
   The signatories of a contract are specified in the Daml-LF definition of
   the template for said contract. Conceptually, they are the parties that
   agreed for that contract to be created.
 
-``key_with_maintainers`` is optional. If present:
-
-* Its ``maintainers`` must be a subset of the ``signatories``;
```
cocreature, line 365: does this also count as validation or why did you remove it?

daml-lf/spec/transaction.rst
```diff
@@ -434,27 +417,21 @@ As of version 10, these fields are included:
...
 
-``children`` may be empty; all other fields are required, and required
-to be non-empty.
+``children`` and ``key_with_maintainers`` may be empty; all other
+fields are required, and required to be non-empty.
 
-``template_id``'s structure is defined by `the value specification`_;
-the version of that specification to use when consuming it is the
-``version`` field of ``chosen_value``.
+``template_id``'s structure is defined by `the value specification`_.
```
cocreature, line 425: same question what happened to the version field? I thought we still have that in 10 or am I off by one?

daml-lf/spec/value.rst
```diff
@@ -150,35 +150,39 @@ Support for value versions 1 to 5 was dropped on 2020-12-10.
...
 ``version`` is required, and must be a version of this specification.
-For example, for version 1 of this specification, ``version`` must be
-``"1"``.  Consumers can expect this field to be present and to
-have the semantics defined here without knowing the version of this
-value in advance.
+For backward compatibility reasons:
+- the version `10` is encoded as the string "6";
+- string "10" is reserved and will be never used to encoded any future version;
+- versions 11 or latter will be encoded as string, for instance
+  version 11 of this specification, ``version`` must be ``"1"``.
```
cocreature, line 181:
```suggestion
  version 11 of this specification, ``version`` must be ``"11"``.
```
or am I missing something? It looks like otherwise we’ll get a collision at 16.

daml-lf/spec/value.rst
```diff
@@ -218,11 +222,11 @@ As of version 6, may be any one of these:
 * `message Numeric`_ numeric
 
 ``Value`` is recursive by virtue of occurrences in some of the above
-cases, e.g. ``list`` contains any number of ``Value``. The maximum depth
-of nested ``Value``, including the outermost, is 100; any more yields an
-invalid value.
+cases, e.g. ``list`` contains any number of ``Value``. The maximum
+depth of nested ``Value``, including the outermost, is 100; any more
```
cocreature, line 226:
```suggestion
depth of a nested ``Value``, including the outermost, is 100; any more
```

daml-lf/spec/value.rst
```diff
@@ -560,14 +552,19 @@ Both ``key`` and ``value`` are required.
...
 In this version, these fields are included:
 
 * repeated `message GenMap.Entry`_ entries
 
-The ``keys`` and the ``value`` fields of every member of ``entries`` must
-conform to the same types. Entries occur in insertion order. If two ore more
-entries have the same keys, the last one override the former entries.
+.. note: *this section is non-normative*
+
+   The ``value`` field of every member of ``entries`` must conform to
```
cocreature, line 565: it seems like you deleted some of those type of notes completely while others got moved into a non-normative section. Is there any specific reason why you chose one or the other in the various places?

daml-lf/spec/value.rst
```diff
@@ -461,49 +457,44 @@ As of version 6, these fields are included:
...
 message Optional
 ^^^^^^^^^^^^^^^^
 
-*since version 6*
+(*since version 10*)
 
 An optional value (equivalent to Scala's ``Option`` or Haskell's
 ``Maybe``).
 
-Only ``Optional`` may be used to encode a Value that conforms to an LF
```
cocreature, line 488: what’s wrong with that?

daml-lf/spec/value.rst
```diff
@@ -461,49 +457,44 @@ As of version 6, these fields are included:
 
   [A-Za-z0-9._:-]+
 
-If ``relative`` is unset or false, this is an absolute contract ID;
```
cocreature, line 464: what happened to this? you still include the boolean field above. If we got rid of relative contract ids, don’t we need to kill both?

daml-lf/spec/value.rst
```diff
@@ -307,73 +313,71 @@ be rejected with error by conforming consumers.
...
 * ``string`` label
 * `message Value`_ value
 
-``label`` may be an empty string.  However, if ``label`` is non-empty,
-it must match the name of the field in this position in the Daml-LF
-record type under consideration.  For example, if the second field of an
-LF record type is named ``bar``, then label of the second element of
-`field fields`_ may be ``"bar"``, or an empty string in circumstances
-mentioned above.  Any other label produces an invalid LF value.
+All fields are required and must be non-empty.
```
cocreature, line 362: that’s not true, `label` can be empty.

cocreature (approved):
lgtm with a few minor comments, thanks!

daml-lf/spec/transaction.rst
remyhaemmerle-da: This is part of validation, nothing constrains `maintainers` from serialization point of view.

daml-lf/spec/transaction.rst
remyhaemmerle-da: `the version of that specification to use when consuming it is the ``version`` field of ``value``.`

This was relevant for legacy value versions that have been dropped

daml-lf/spec/transaction.rst
remyhaemmerle-da: In original design the field ``template_id`` was interpreted according the version of the ``chosen_value``.
In practice this feature has never been used, and was drop when redesigning the versioning scheme. 
no need to speak of the``version`` field of ``chosen_value` here.

daml-lf/spec/value.rst
remyhaemmerle-da: We should have kill that for some time. 
We always set to `false` and never read it. 
It kill it in this PR.

daml-lf/spec/value.rst
remyhaemmerle-da: I do not see the point of this statement. There is currently only one way to encode an `Optional`, I do not see why we cannot change that in the future.

daml-lf/spec/value.rst
remyhaemmerle-da: I put back the few I deleted

PR #8645, opened 2021-01-27
resources: Remove the default factories.

SamirTalwar-DA:
They're no longer necessary, and could cause confusion. Users should use their own factory object, not a generic one.

This extracts out `Resource.apply` to a separate builder function, as this is the only place where the factories were useful.

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):
LGTM, thanks!

SamirTalwar-DA:
This has the nice side effect of making the Scaladoc links in the `ResourceOwnerFactories` class actually work. (I noticed they were broken, which is how I realized this temporary hack was still around.)

PR #8647, opened 2021-01-27
limit blackduck scans to main

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END


dasormeter (approved):
should be fine for BAU, but just need to remember to comment this out if we ever need to test the scan on a branch — anyhow you have to do the commenting of other compat jobs too so probably fine

cocreature (commented):
Thanks!

PR #8648, opened 2021-01-27
LF: update transaction and value spec for LF 1.12

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/spec/transaction.rst
+++ b/daml-lf/spec/transaction.rst
@@ -179,6 +179,8 @@ later.
 +--------------------+-----------------+
 |                 11 |      2021-01-19 |
 +--------------------+-----------------+
+|       (preview) 12 |      2021-01-27 |
++--------------------+-----------------+
 |                dev |      2020-12-14 |
 +--------------------+-----------------+
 
@@ -225,10 +227,11 @@ to construct that instance.
 As of version 10, these fields are included:
 
 * `message Identifier`_ template_id
-* `message VersionedValue`_ value
+* `message VersionedValue`_ arg_versioned
 * ``string`` agreement
 
-``template_id`` and ``value`` are required; ``agreement`` is optional.
+``template_id`` and ``arg_versioned`` are required; ``agreement`` is
+optional.
 
 ``template_id``'s structure is defined by `the value specification`_.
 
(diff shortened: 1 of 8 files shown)


daml-lf/spec/transaction.rst
```diff
@@ -225,10 +227,11 @@ to construct that instance.
 As of version 10, these fields are included:
 
 * `message Identifier`_ template_id
-* `message VersionedValue`_ value
+* `message VersionedValue`_ arg_versioned
```
cocreature, line 230: don’t we need `arg_unversioned` in newer versions?

daml-lf/spec/transaction.rst
```diff
@@ -358,6 +370,16 @@ identifier.
...
 
+(*since version 12*)
+
+As of version 12, these fields are included:
+
+* `message Identifier`_ template_id
+* `message VersionedValue`_ arg_unversioned
+* ``string`` agreement
+
+``template_id`` and ``arg_unversioned`` is required while
```
cocreature, line 381: This seems to be a bit confusing, you state above that the contract instance is required and then here you say it’s optional for 12. I’d put the two statements close to each other to make it clear that it’s required in 10 and the alternative is arg_unversioned in 12.

daml-lf/spec/value.rst
```diff
@@ -374,21 +379,25 @@ All fields are required and must be non-empty.
    containing record, as declared by the LF record type.  It must be
    supplied in all cases.
 
+(*since version 12*)
+
+As of version 12, the field ``label`` is not used anymore.
```
cocreature, line 384: you state above that the field is required. I think the message should be closer to that, otherwise it gets a bit confusing.

daml-lf/spec/transaction.rst
cocreature: ah nvm, this is `contract_instance` while the other one is not in `ContractInstance`.

cocreature (approved):
Thanks!

daml-lf/spec/transaction.rst
remyhaemmerle-da: Since `ContractInstance` is not used any more in Create Node, the only possible usage is alone (for instance in the indexer). In this later case we need the version. 
I add the suffix  `versioned` for the sake of consistency.

daml-lf/spec/transaction.rst
remyhaemmerle-da: This is the original design of the document. I do not feel like changing it now. 
Furthermore it is stated at the beginning: 

```
Do not read this without ``transaction.proto``
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

In the proto definition, there is `since` and `until` comments that make that clear.

PR #8652, opened 2021-01-27
LF: preview of LF 1.12

remyhaemmerle-da:
CHANGELOG_BEGIN

* LF: preview of LF 1.11. Preview versions can be changed only to
include bug fixes. Changes of LF 1.12 include:
  - reduce transaction size by erasing type information in user-defined
    type.

CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/archive/BUILD.bazel
+++ b/daml-lf/archive/BUILD.bazel
@@ -51,7 +51,6 @@ load("@rules_pkg//:pkg.bzl", "pkg_tar")
         ),
     ]
     for version in LF_VERSIONS
-    if version != "1.12"
 ]
 
 # FIXME(JM): Clean this up
@@ -131,6 +130,7 @@ da_scala_test_suite(
         ":DarReaderTest.dalf",
         ":DarReaderTest.dar",
         ":daml_lf_1.11_archive_proto_srcs",
+        ":daml_lf_1.12_archive_proto_srcs",
         ":daml_lf_1.6_archive_proto_srcs",
         ":daml_lf_1.7_archive_proto_srcs",
         ":daml_lf_1.8_archive_proto_srcs",
@@ -147,7 +147,7 @@ da_scala_test_suite(
         "2.12": ["@maven//:org_scala_lang_modules_scala_collection_compat"],
     },
     deps = [
-        ":daml_lf_1.11_archive_proto_java",
+        ":daml_lf_1.12_archive_proto_java",
         ":daml_lf_1.6_archive_proto_java",
         ":daml_lf_1.7_archive_proto_java",
         ":daml_lf_1.8_archive_proto_java",
--- a/daml-lf/archive/BUILD.bazel
+++ b/daml-lf/archive/BUILD.bazel
@@ -51,6 +51,7 @@ load("@rules_pkg//:pkg.bzl", "pkg_tar")
         ),
     ]
     for version in LF_VERSIONS
+    if version != "1.12"
 ]
 
 # FIXME(JM): Clean this up
--- a/daml-lf/encoder/BUILD.bazel
+++ b/daml-lf/encoder/BUILD.bazel
@@ -10,20 +10,14 @@ load(
 )
 load(
     "//daml-lf/language:daml-lf.bzl",
+    "LF_VERSIONS",
     "lf_dev_version",
     "lf_latest_version",
     "lf_preview_version",
     "lf_stable_version",
 )
 
-lf_targets = [
-    "1.6",
-    "1.7",
-    "1.8",
-    "1.11",
-    "1.12",
-    "1.dev",
-]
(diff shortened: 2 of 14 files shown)


daml-lf/archive/BUILD.bazel
```diff
@@ -51,6 +51,7 @@ load("@rules_pkg//:pkg.bzl", "pkg_tar")
         ),
     ]
     for version in LF_VERSIONS
+    if version != "1.12"
```
cocreature, line 54: Why don’t we produce protobufs for preview versions? Did we do the same for 1.11?

cocreature: Or is this because the ledger API side (including daml-lf) is unchanged? I think we should still have those libs. Omitting certain versions just gets confusing.

daml-lf/encoder/BUILD.bazel
```diff
@@ -21,6 +21,7 @@ lf_targets = [
     "1.7",
     "1.8",
     "1.11",
+    "1.12",
```
cocreature, line 24: Can we just replace this list by `LF_VERSIONS`?

daml-lf/language/daml-lf.bzl
```diff
@@ -18,17 +18,29 @@ LF_VERSIONS = [
...
     "1.8",
     "1.11",
+    "1.12",
     "dev",
 ]
 
 # The subset of LF versions accepted by the compiler in the syntax
 # expected by the --target option.
-COMPILER_LF_VERSIONS = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS]
+COMPILER_LF_VERSIONS = [
```
cocreature, line 27: Please filter `LF_VERSIONS` instead of making this an explicit list. I deliberately changed this since the two got out of sync.

cocreature: And add a TODO for removing the filter once the compiler side has caught up.

daml-lf/language/daml-lf.bzl
```diff
@@ -18,17 +18,29 @@ LF_VERSIONS = [
...
 
 # We need Any in DAML Script so we require DAML-LF >= 1.7
 SCRIPT_LF_VERSIONS = [ver for ver in COMPILER_LF_VERSIONS if ver != "1.6"]
 
-LF_VERSION_PACKAGE_DIGITALASSET = {"1.6": "digitalasset", "1.7": "digitalasset", "1.8": "digitalasset", "1.11": "daml"}
+LF_VERSION_PACKAGE_DIGITALASSET = {
+    "1.6": "digitalasset",
+    "1.7": "digitalasset",
+    "1.8": "digitalasset",
+    "1.11": "daml",
```
cocreature, line 42: Did we include 1.11 here when it was still a preview? If so, we should do the same for 1.12.

release/util.bzl
```diff
@@ -143,6 +143,8 @@ protos_zip = rule(
             default = [
                 Label("//daml-lf/archive:daml_lf_{}_archive_proto_tar.tar.gz".format(version))
                 for version in LF_VERSIONS
+                # we skip LF 1.12 as it is exactly as LF 1.11 for archive point of view
```
cocreature, line 146: I’m not sure that’s a good idea in released artifacts. I think this is perfectly sensible to skip such versions in tests but we should have a released artifact for that version, otherwise it just becomes confusing.

cocreature (approved):
LGTM with a bit of bikeshedding

daml-lf/archive/BUILD.bazel
remyhaemmerle-da: I did not want to release the proto for 1.12.  I will.

daml-lf/encoder/BUILD.bazel
```diff
@@ -10,19 +10,14 @@ load(
...
 )
 
-lf_targets = [
-    "1.6",
-    "1.7",
-    "1.8",
-    "1.11",
-    "1.dev",
-]
+lf_targets = ["1.dev" if ver == "dev" else ver for ver in LF_VERSIONS]
```
cocreature, line 20: Actually isn’t this just `COMPILER_LF_VERSIONS`?

cocreature (approved):
Thanks!

daml-lf/encoder/BUILD.bazel
remyhaemmerle-da: no. `COMPILER_LF_VERSIONS` still does not include "1.12".

sofiafaro-da: I can clean this up when I add the compiler support (probably by moving this to the other file and calling it `ENCODER_LF_VERSIONS` to make its relation with `COMPILER_LF_VERSIONS` more explicit).

sofiafaro-da (approved):
Thank you!

PR #8653, opened 2021-01-27
clean-up infra after Ubuntu 20.04 upgrade

garyverhaegen-da:
See #8617.

This has to be a separate PR so we could drain all the jobs still
expecting the linux-pool pool to exist.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
:fire:

PR #8655, opened 2021-01-27
fix perf_speedy report

garyverhaegen-da:
The report currently chokes on quotes in the commit message (see 550aa48fc52cb6dfa19de6b86e27b481d3835048). Rather than trying to correctly excape things in Bash, this PR delegates the quote handling to jq, because having to deal with Bash embedded in YAML is hard enough.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/daily-compat.yml
+++ b/ci/cron/daily-compat.yml
@@ -116,7 +116,7 @@ jobs:
           GCRED: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
       - template: ../daily_tell_slack.yml
         parameters:
-          success-message: '$(cat $(Build.StagingDirectory)/perf-results.json | jq . | jq -sR ''"perf for ''"$COMMIT_LINK"'':```\(.)```"'')'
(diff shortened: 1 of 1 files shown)


ci/cron/daily-compat.yml
```diff
@@ -23,53 +23,6 @@ schedules:
   always: true
 
 jobs:
-  - job: compatibility_ts_libs
```
cocreature, line 26: I’m confused, why is there so much deleted stuff?

garyverhaegen-da: Second commit will be dropped before merging; only the first one needs review. Sorry, should have mentioned that.

garyverhaegen-da: Well actually, that failed, so hold off on reviewing.

garyverhaegen-da:
Oh, forgot to update: this is now fully working and thus ready for review. You can see there's a quote in the commit title and the compat tests (and thus the perf) have run successfully.

cocreature (approved):
Thanks!

PR #8656, opened 2021-01-27
Retrofit types in verbose mode [DPP-154]

rautenrieth-da:
This PR implements the retrofitting of field names and type identifiers in verbose mode. The first commit is also split into a separate PR, to simplify reviewing: https://github.com/digital-asset/daml/pull/8670 (i.e., feel free to ignore the first commit when reviewing).

## Implementation notes

Most of the changes involve feeding the required context to LfValueTranslation, because it now needs access to:
    -  a `ValueEnricher` (and thus an `Engine`)
    - package loading, which involves `Future`s, `ExecutionContext`s, and `LoggingContexts`

## Existing tests that cover value retrofitting

- `ValueEnricher` has unit tests
- `JdbcLedgerDao*Spec` requests values in verbose mode, which does cover the modified `LfValueTranslation` methods
 
## Missing tests

- The ledger API test tool requests values in verbose mode, but it's never run with packages compiled with `1.dev` - see https://github.com/digital-asset/daml/pull/8695


cocreature:
> * 
> 
> * Figure out why the ledger test tool tests pass, as they use verbose mode

I think we don’t test the ledger API test tool against 1.dev. We should probably change that.

remyhaemmerle-da (approved):
looks good to me. 
I will let somebody more familiar with the participant code for approving.

ledger/participant-integration-api/src/main/scala/platform/packages/DeduplicatingPackageLoader.scala
```diff
@@ -0,0 +1,64 @@
...
+import com.daml.lf.language.Ast.Package
+import com.daml.metrics.Timed
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.util.{Failure, Success, Try}
+
+/** Deduplicates package load requests, such that concurrent requests
+  * only access the data store and decode the package once.
+  */
+private[platform] class DeduplicatingPackageLoader() {
```
mziolekda, line 22: It would be great to have a test for this class

mziolekda (approved):
It looks good, but it would be great to provide a test for the DeduplicatingPackageLoader

PR #8662, opened 2021-01-28
Minor docs fixes

bame-da:
Fixes a number of small docs issues:

1. In Chrome, I'd always get "null" in the search bar
2. The favicons were the old DAML logos
3. We had some z-indices that were so large we overlaid popups added by plugins
4. The hover effect in the search was wrong and just turned everything white
5. Our copyright message started "(C) Copyright (C) Copyright..."

### Pull Request Checklist

trigger the build.


docs/configs/html/conf.py
```diff
@@ -52,7 +52,7 @@
 
 # General information about the project.
 project = u'Daml SDK'
-copyright = u'© Copyright 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved. Any unauthorized use, duplication or distribution is strictly prohibited. "Digital Asset" and "Daml" are Registered in the U.S. Patent and Trademark Office.'
+copyright = u'2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved. Any unauthorized use, duplication or distribution is strictly prohibited. "Digital Asset" and "Daml" are Registered in the U.S. Patent and Trademark Office.'
```
garyverhaegen-da, line 55: Any chance you could change that on the template side instead? Having a different string just for this one copyright notice across the entire repo means it will get forgotten every year when we need to update it.

bame-da: Good point. FIxed.

cocreature (commented):
Do we have to commit the `__pycache__` stuff? I thought that was only a temporary cache file that is usually not comitted.

bame-da:
Thanks for spotting that. I don't know what action created that folder, but I've added it to gitignore now.

docs/configs/pdf/conf.py
```diff
@@ -54,7 +54,7 @@
 
 # General information about the project.
 project = u'Daml SDK'
-copyright = u'Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved. Any unauthorized use, duplication or distribution is strictly prohibited.'
+copyright = u'© Copyright 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved. Any unauthorized use, duplication or distribution is strictly prohibited. "Digital Asset" and "Daml" are Registered in the U.S. Patent and Trademark Office.'
```
garyverhaegen-da, line 57: Apologies for the nitpicking here, but the rest of the repo uses `Copyright (c)`. Any reason to diverge from that? (Should we update the rest of the repo to use © instead?)

bame-da: I just went with what was there. I really don't mind what it is as long as legal is satisfied and it doesn't look glitchy. I've changed it to (c).

cocreature (approved):
Thanks!

PR #8664, opened 2021-01-28
ci: use setvar to set variables

garyverhaegen-da:
My goal here is to investigate the new warning Azure has been showing for the past few days:

> ##[warning]%25 detected in ##vso command. In March 2021, the agent command parser will be updated to unescape this to %. To opt out of this behavior, set a job level variable DECODE_PERCENTS to false. Setting to true will force this behavior immediately. More information can be found at https://github.com/microsoft/azure-pipelines-agent/blob/master/docs/design/percentEncoding.md

As far as I'm aware we are not deliberately passing in any `%25` in any of our `vso` commands, so I was a bit surprised by this.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -70,14 +70,18 @@ jobs:
     - checkout: self
       persistCredentials: true
     - template: ci/clean-up.yml
+    - template: ci/bash-lib.yml
+      parameters:
+        var_name: bash-lib
     - bash: |
         set -euo pipefail
+        source $(bash-lib)
         if git tag v$(release_tag) $(release_sha); then
           git push origin v$(release_tag)
           mkdir $(Build.StagingDirectory)/release
           mkdir $(Build.StagingDirectory)/artifactory
         else
-          echo "##vso[task.setvariable variable=skip-github]TRUE"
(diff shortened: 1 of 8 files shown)


cocreature (approved):
lgtm, thanks!

SamirTalwar-DA:
Is there a reason this isn't running on CI?

garyverhaegen-da:
> Is there a reason this isn't running on CI?

That usually means I screwed up the YAML syntax somehow so Azure ails to parse it. In other words, this PR is off to a great start 🙂

cocreature:
afaict, the warning here comes’s from azure’s builtin jobs not our own uses of `setvar`.

cocreature:
upstream issue https://github.com/microsoft/azure-pipelines-tasks/issues/14302

garyverhaegen-da:
> afaict, the warning here comes’s from azure’s builtin jobs not our own uses of `setvar`.

The goal of unifying all of our vso commands through one function was to make it easy to add debug statements to find out which of our commands actually passes a `%`. I couldn't find one, though.

PR #8666, opened 2021-01-28
Increased compatibility test size for *classic-postgresql to 'large' [DPP-195]

kamil-da:
This change is our current best guess on a fix (an experiment really) to get rid of timeouts in `*-classic-postgresql` conformance tests.
Change is done according to [the Bazel docs](https://docs.bazel.build/versions/master/be/common-definitions.html).

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compatibility/bazel_tools/testing.bzl
+++ b/compatibility/bazel_tools/testing.bzl
@@ -466,6 +466,7 @@ def sdk_platform_test(sdk_version, platform_version):
 
     client_server_test(
         name = name + "-classic-postgresql",
+        size = "large",
         client = ledger_api_test_tool,
         client_args = [
             "localhost:6865",


cocreature (commented):
As mentioned yesterday, I don’t think this does anything. The size only affects scheduling according to @aherrmann-da and this is an exclusive test so it cannot be scheduled any more exclusive than it already is.

kamil-da:
> As mentioned yesterday, I don’t think this does anything. The size only affects scheduling according to @aherrmann-da and this is an exclusive test so it cannot be scheduled any more exclusive than it already is.

I've done local testing with this change and it didn't make much difference which could confirm what you say.
But the team's opinion on this was "let's give it a try", hence the PR.

cocreature:
not opposed to merging it, it shouldn’t do any harm either.

mziolekda (approved):
Let's get it in. There is a theory that large size might boost bazel resource allocation. If true it would help fix flay compatibility tests.

PR #8668, opened 2021-01-28
Clean broken entries from the Bazel cache

cocreature:
This is hopefully a somewhat reasonable workaround for the "output not
created" errors that keep annoying us.

For now, this is just part of the hourly cronjob but we could move it
somewhere else if desired.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature:
@garyverhaegen-da do we have delete permissions on the gcp token? if not, we probably need a new token for this.

cocreature:
Obligatory note that while I tested this locally, I couldn’t test deletions.

aherrmann-da (approved):
Great work! Thanks for digging into this!

garyverhaegen-da:
> @garyverhaegen-da do we have delete permissions on the gcp token? if not, we probably need a new token for this.

Yes, CI has [full access to the bazel cache](https://github.com/digital-asset/daml/blob/main/infra/bazel_cache.tf#L22-L31), because I was apparently not able to remove its delete permission.

azure-cron.yml
```diff
@@ -48,6 +48,28 @@ jobs:
...
+          set -euo pipefail
+          eval "$(dev-env/bin/dade assist)"
+
+          bazel build //ci/cron:cron
+          # 90 minutes should provide enough overlap for an hourly
+          # cronjob.
+          ./bazel-bin/ci/cron/cron bazel-cache --age 90 --delete
+        env:
+          AWS_ACCESS_KEY_ID: $(AWS_ACCESS_KEY_ID)
+          AWS_SECRET_ACCESS_KEY: $(AWS_SECRET_ACCESS_KEY)
```
garyverhaegen-da, line 70: These are not the keys you're looking for.

You want:
```
        env:
          GCRED: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
```

and you also need to replicate the logic from [bash-lib](https://github.com/digital-asset/daml/blob/main/ci/bash-lib.yml#L74-L94), either in the Bash script here or in your Haskell code. (Or call Bash from Haskell, but some people find that messy.)

azure-cron.yml
```diff
@@ -57,17 +57,21 @@ jobs:
...
           set -euo pipefail
           eval "$(dev-env/bin/dade assist)"
 
           bazel build //ci/cron:cron
+
+          source $(bash-lib)
           # 90 minutes should provide enough overlap for an hourly
           # cronjob.
-          ./bazel-bin/ci/cron/cron bazel-cache --age 90 --delete
+          gcs ./bazel-bin/ci/cron/cron bazel-cache --age 90 --delete
```
garyverhaegen-da, line 72: As it stands this would end up running `gsutil ./bazel-bin/ci/cron bazel-cache --age 90 --delete`, but I like the idea. Perhaps it's worth changing `bash-lib` so that `gcs` really is the `with-gcs-credentials` type of wrapper you're going for here.

cocreature: Yeah I just realized that, I copied the logic for now instead of calling gcs. Do you want to align them in this PR or separately?

azure-cron.yml
cocreature: Duh, I clearly should have taken a look at the credentials instead of blindly copying the docs cron. Thanks for catching it!

garyverhaegen-da (approved):

azure-cron.yml
garyverhaegen-da: Separately is fine, I've made a note of it. I don't want to delay this.

PR #8671, opened 2021-01-28
Use cliopts.Http for OAuth 2.0 middleware

aherrmann-da:
- Use `com.daml.cliopts.Http` for OAuth 2.0 middleware host and port command-line flags.
  - This is consistent with the flags that the trigger service offers.
  - This allows configuring the list host to something else than "localhost", e.g. "0.0.0.0".
  - This adds a `--port-file` flag as well.
- Add test-case for the `--port-file` option.
- Update the docs.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
+++ b/triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
@@ -6,15 +6,17 @@ package com.daml.auth.middleware.oauth2
 import java.nio.file.{Path, Paths}
 
 import akka.http.scaladsl.model.Uri
+import com.daml.cliopts
 import com.daml.jwt.{JwtVerifierBase, JwtVerifierConfigurationCli}
-import com.daml.ports.Port
 
 import scala.concurrent.duration
 import scala.concurrent.duration.FiniteDuration
 
 case class Config(
-    // Port the middleware listens on
-    port: Port,
+    // Host and port the middleware listens on
+    address: String,
+    port: Int,
+    portFile: Option[Path],
     // The URI to which the OAuth2 server will redirect after a completed login flow.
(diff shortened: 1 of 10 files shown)


triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
```diff
@@ -65,10 +70,12 @@ object Config {
     new scopt.OptionParser[Config]("oauth-middleware") {
       head("OAuth2 Middleware")
 
-      opt[Int]("port")
-        .action((x, c) => c.copy(port = Port(x)))
-        .required()
-        .text("Port to listen on")
+      cliopts.Http.serverParse(this, serviceName = "Trigger")(
```
cocreature, line 73:
```suggestion
      cliopts.Http.serverParse(this, serviceName = "OAuth2 Middleware")(
```

triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
```diff
@@ -36,13 +38,16 @@ case class Config(
 )
 
 object Config {
+  val DefaultHttpPort: Int = 3000
```
cocreature, line 41: any particular reason to choose 3000? I don’t really understand any of our default port numbers so no issue with that just curious.

cocreature (approved):
lgtm, thanks!

triggers/service/auth/src/main/scala/com/daml/auth/middleware/oauth2/Config.scala
aherrmann-da: No, no particular reason. This just happened to be the port number used for the oauth middleware in the docs examples.

PR #8672, opened 2021-01-28
Non-repudiation: use binary headers for signature and fingerprint

stefanobaghino-da:
This should reduce the times we encode and decode to and from base64 strings.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/runtime-components/non-repudiation-core/src/main/java/com/daml/nonrepudiation/Headers.java
+++ b/runtime-components/non-repudiation-core/src/main/java/com/daml/nonrepudiation/Headers.java
@@ -10,13 +10,13 @@ public final class Headers {
     private Headers() {
     }
 
-    public static final Metadata.Key<String> SIGNATURE =
-            Metadata.Key.of("signature", Metadata.ASCII_STRING_MARSHALLER);
+    public static final Metadata.Key<byte[]> SIGNATURE =
+            Metadata.Key.of("signature-bin", Metadata.BINARY_BYTE_MARSHALLER);
 
     public static final Metadata.Key<String> ALGORITHM =
             Metadata.Key.of("algorithm", Metadata.ASCII_STRING_MARSHALLER);
 
-    public static final Metadata.Key<String> FINGERPRINT =
-            Metadata.Key.of("fingerprint", Metadata.ASCII_STRING_MARSHALLER);
+    public static final Metadata.Key<byte[]> FINGERPRINT =
+            Metadata.Key.of("fingerprint-bin", Metadata.BINARY_BYTE_MARSHALLER);
 
 }
--- a/runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
+++ b/runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
@@ -11,13 +11,14 @@ import scala.collection.concurrent.TrieMap
 
 object KeyRepository {
 
+  type Fingerprint <: Array[Byte]
+
   object Fingerprint {
-    def wrap(bytes: Array[Byte]): Fingerprint = new Fingerprint(bytes)
+    def wrap(bytes: Array[Byte]): Fingerprint = bytes.asInstanceOf[Fingerprint]
     def wrap(key: PublicKey): Fingerprint = wrap(Fingerprints.compute(key))
   }
 
-  @SuppressWarnings(Array("org.wartremover.warts.ArrayEquals"))
-  final class Fingerprint(val bytes: Array[Byte]) extends AnyVal {
+  implicit final class FingerprintBase64(val bytes: Fingerprint) extends AnyVal {
     def base64: String = BaseEncoding.base64().encode(bytes)
   }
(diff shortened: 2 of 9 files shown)


runtime-components/non-repudiation-core/src/main/java/com/daml/nonrepudiation/Headers.java
```diff
@@ -10,13 +10,13 @@
     private Headers() {
     }
 
-    public static final Metadata.Key<String> SIGNATURE =
-            Metadata.Key.of("signature", Metadata.ASCII_STRING_MARSHALLER);
+    public static final Metadata.Key<byte[]> SIGNATURE =
```
cocreature, line 13: Is there a good reason to have a separate header name for this at this point given that we don’t care about backwards compat?

runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
```diff
@@ -5,24 +5,20 @@ package com.daml.nonrepudiation
...
-    def assertFromString(string: String): Fingerprint =
-      string.asInstanceOf[Fingerprint]
-
-    def fromString(string: String): Try[Fingerprint] =
-      Try(assertFromString(string))
-
-    def compute(key: PublicKey): Fingerprint =
-      assertFromString(Base64Fingerprint.compute(key))
-
+  @SuppressWarnings(Array("org.wartremover.warts.ArrayEquals"))
```
cocreature, line 19: Do we trigger this via some autoderived equals instance? Should we overwrite equality to do something sensible?

cocreature (approved):
lgtm with one comment on the header names, thanks!

runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
stefanobaghino-da: I'll have a look if we can do something about it.

runtime-components/non-repudiation-core/src/main/java/com/daml/nonrepudiation/Headers.java
stefanobaghino-da: I would have kept it the same, but gRPC actually [throws an exception](https://github.com/grpc/grpc-java/blob/a6df2b2ff403f9aa9097e3fdc6908a53a8f3c4bc/api/src/main/java/io/grpc/Metadata.java#L851-L860) if I don't append `-bin` to the name of binary headers.

runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
stefanobaghino-da: Value classes cannot redefine `equals` and I believe `==` is simply synthesized by the compiler. @S11001001 do you have suggestions?

S11001001: Do not use a value class for this kind of thing. [They are only good for syntax extension.](https://discuss.daml.com/t/extends-anyval-is-good-for-extension-methods-and-thats-it-scala-ritual/1548)

You can then manually write an `equals` or, if you change to a `case class`, mix in [`SomeArrayEquals`](https://github.com/digital-asset/daml/blob/81dc6b947374886c0ba41fe27fa0546b09659071/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SomeArrayEquals.scala).

If you really want a newtype, use the abstract type with existential implementation pattern we use for all sorts of things around the codebase like `LedgerString`, `NewBoolean`, `com.daml.concurrent.Future`, or `DiscriminatedSubtype`.  No synthetic `equals` implementations are made in those cases; of course this is not much help to you if you _need_ `equals` to work (but this is completely incompatible with using a newtype at all).

stefanobaghino-da: I went with the abstract type initially but It accepted a simple `Array[Byte]` in its place. I'll have a second look if I can make it work.

runtime-components/non-repudiation-core/src/main/java/com/daml/nonrepudiation/Headers.java
cocreature: huh, interesting, makes sense thanks!

runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/KeyRepository.scala
stefanobaghino-da: Used it, this time raw `Array[Byte]` are rejected. No idea what I got wrong the first time.

Addressed by https://github.com/digital-asset/daml/pull/8672/commits/a099a8f618b0769b742920281ddc27e345524b2e

PR #8674, opened 2021-01-28
Relax party validation in DAML Script over JSON API

cocreature:
Since the JSON API infers parties from the token instead of accepting
them explicitly, we have to ensure that the parties in the token match
the parties passed to `submit`/`submitMulti` exactly. However, we were
a bit too strict and required a party to be in `readAs` even if it is
already in `actAs`. This caused issues on DamlHub because they issue
tokens with the party in both `actAs` and `readAs` so regular `submit`
doesn’t work.

changelog_begin

- [DAML Script] When running DAML Script over the JSON API, the check
  that the parties in the DAML Script match the parties in the token
  has been relaxed and now allows for duplicate parties as well as
  parties that are only in actAs in the DAML Script but in both actAs
  and readAs in the token.

changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):
Lots of tests. ❤️

PR #8676, opened 2021-01-28
Improve logging in ledger API server [DPP-231]

kamil-da:
**Motivation**
Troubleshooting is made unnecessarily hard by lack of adequate logging. Every client request should be logged at the info level as it enters and leaves the system.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -49,8 +49,12 @@ private[apiserver] final class ApiCommandCompletionService private (
 
         completionsService
           .getCompletions(offset, request.applicationId, request.parties)
+          .via(logger.debugStream(completionsLoggable))
           .via(logger.logErrorsOnStream)
     }
+
+  private def completionsLoggable(response: CompletionStreamResponse): String =
+    s"Responding with transactions: ${response.completions.map(_.toProtoString).mkString("[", ",", "]")}"
 
   override def getLedgerEnd(ledgerId: domain.LedgerId): Future[LedgerOffset.Absolute] =
     completionsService.currentLedgerEnd().andThen(logger.logErrorsOnCall[LedgerOffset.Absolute])
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -54,7 +54,7 @@ private[apiserver] final class ApiCommandCompletionService private (
     }
 
   private def completionsLoggable(response: CompletionStreamResponse): String =
-    s"Responding with transactions: ${response.completions.map(_.toProtoString).mkString("[", ",", "]")}"
+    s"Responding with completions: ${response.completions.map(_.toProtoString).mkString("[", ",", "]")}"
 
   override def getLedgerEnd(ledgerId: domain.LedgerId): Future[LedgerOffset.Absolute] =
     completionsService.currentLedgerEnd().andThen(logger.logErrorsOnCall[LedgerOffset.Absolute])
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -54,7 +54,17 @@ private[apiserver] final class ApiCommandCompletionService private (
     }
 
   private def completionsLoggable(response: CompletionStreamResponse): String =
-    s"Responding with completions: ${response.completions.map(_.toProtoString).mkString("[", ",", "]")}"
+    s"Responding with completions: ${response.completions.toList
+      .map(c => singleCompletionLoggable(c.commandId, c.status.map(_.code)))}"
+
+  private def singleCompletionLoggable(
+      commandId: String,
+      statusCode: Option[Int],
+  ): Map[String, String] =
+    Map(
+      logging.commandId(commandId),
+      "statusCode" -> statusCode.map(_.toString).getOrElse(""),
+    )
 
   override def getLedgerEnd(ledgerId: domain.LedgerId): Future[LedgerOffset.Absolute] =
     completionsService.currentLedgerEnd().andThen(logger.logErrorsOnCall[LedgerOffset.Absolute])
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -63,7 +63,7 @@ private[apiserver] final class ApiCommandCompletionService private (
   ): Map[String, String] =
     Map(
       logging.commandId(commandId),
-      "statusCode" -> statusCode.mkString,
+      "statusCode" -> statusCode.map(_.toString).getOrElse(""),
     )
 
   override def getLedgerEnd(ledgerId: domain.LedgerId): Future[LedgerOffset.Absolute] =
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -63,7 +63,7 @@ private[apiserver] final class ApiCommandCompletionService private (
   ): Map[String, String] =
     Map(
       logging.commandId(commandId),
-      "statusCode" -> statusCode.map(_.toString).getOrElse(""),
+      "statusCode" -> statusCode.mkString,
     )
 
   override def getLedgerEnd(ledgerId: domain.LedgerId): Future[LedgerOffset.Absolute] =
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -111,7 +111,7 @@ private[apiserver] final class ApiSubmissionService private (
   override def submit(request: SubmitRequest): Future[Unit] =
     withEnrichedLoggingContext(logging.commands(request.commands)) { implicit loggingContext =>
       logger.info("Submitting transaction")
-      logger.info(s"Commands: ${request.commands.commands.commands}")
+      logger.trace(s"Commands: ${request.commands.commands.commands}")
       ledgerConfigProvider.latestConfiguration
         .map(deduplicateAndRecordOnLedger(seedService.nextSeed(), request.commands, _))
         .getOrElse(Future.failed(ErrorFactories.missingLedgerConfig()))
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -110,8 +110,8 @@ private[apiserver] final class ApiSubmissionService private (
 
   override def submit(request: SubmitRequest): Future[Unit] =
     withEnrichedLoggingContext(logging.commands(request.commands)) { implicit loggingContext =>
-      logger.debug("Submitting transaction")
-      logger.trace(s"Commands: ${request.commands.commands.commands}")
+      logger.info("Submitting transaction")
+      logger.info(s"Commands: ${request.commands.commands.commands}")
       ledgerConfigProvider.latestConfiguration
         .map(deduplicateAndRecordOnLedger(seedService.nextSeed(), request.commands, _))
         .getOrElse(Future.failed(ErrorFactories.missingLedgerConfig()))
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala
@@ -21,7 +21,9 @@ import com.daml.ledger.participant.state.v1.{
   WriteConfigService,
 }
 import com.daml.lf.data.Time
+import com.daml.platform.apiserver.services.logging
 import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.logging.LoggingContext.withEnrichedLoggingContext
 import com.daml.platform.api.grpc.GrpcApiService
 import com.daml.platform.apiserver.services.admin.ApiConfigManagementService._
 import com.daml.platform.configuration.LedgerConfiguration
@@ -57,11 +59,13 @@ private[apiserver] final class ApiConfigManagementService private (
   override def bindService(): ServerServiceDefinition =
     ConfigManagementServiceGrpc.bindService(this, executionContext)
 
-  override def getTimeModel(request: GetTimeModelRequest): Future[GetTimeModelResponse] =
+  override def getTimeModel(request: GetTimeModelRequest): Future[GetTimeModelResponse] = {
+    logger.info("Getting time model")
     index
       .lookupConfiguration()
       .map(_.fold(defaultConfigResponse) { case (_, conf) => configToResponse(conf) })
       .andThen(logger.logErrorsOnCall[GetTimeModelResponse])
+  }
 
   private def configToResponse(config: Configuration): GetTimeModelResponse = {
     val tm = config.timeModel
@@ -77,55 +81,58 @@ private[apiserver] final class ApiConfigManagementService private (
     )
   }
 
-  override def setTimeModel(request: SetTimeModelRequest): Future[SetTimeModelResponse] = {
-    val response = for {
-      // Validate and convert the request parameters
-      params <- validateParameters(request).fold(Future.failed(_), Future.successful)
+  override def setTimeModel(request: SetTimeModelRequest): Future[SetTimeModelResponse] =
+    withEnrichedLoggingContext(logging.submissionId(request.submissionId)) {
+      implicit loggingContext =>
+        logger.info("Setting time model")
+        val response = for {
+          // Validate and convert the request parameters
+          params <- validateParameters(request).fold(Future.failed(_), Future.successful)
 
-      // Lookup latest configuration to check generation and to extend it with the new time model.
-      optConfigAndOffset <- index.lookupConfiguration()
-      ledgerEndBeforeRequest = optConfigAndOffset.map(_._1)
-      currentConfig = optConfigAndOffset.map(_._2)
+          // Lookup latest configuration to check generation and to extend it with the new time model.
+          optConfigAndOffset <- index.lookupConfiguration()
+          ledgerEndBeforeRequest = optConfigAndOffset.map(_._1)
+          currentConfig = optConfigAndOffset.map(_._2)
 
-      // Verify that we're modifying the current configuration.
-      expectedGeneration = currentConfig
-        .map(_.generation)
-        .getOrElse(LedgerConfiguration.NoGeneration)
-      _ <-
-        if (request.configurationGeneration != expectedGeneration) {
-          Future.failed(
-            ErrorFactories.invalidArgument(
-              s"Mismatching configuration generation, expected $expectedGeneration, received ${request.configurationGeneration}"
-            )
+          // Verify that we're modifying the current configuration.
+          expectedGeneration = currentConfig
+            .map(_.generation)
+            .getOrElse(LedgerConfiguration.NoGeneration)
+          _ <-
+            if (request.configurationGeneration != expectedGeneration) {
+              Future.failed(
+                ErrorFactories.invalidArgument(
+                  s"Mismatching configuration generation, expected $expectedGeneration, received ${request.configurationGeneration}"
+                )
+              )
+            } else {
+              Future.unit
+            }
+
+          // Create the new extended configuration.
+          newConfig = currentConfig
+            .map(config => config.copy(generation = config.generation + 1))
+            .getOrElse(ledgerConfiguration.initialConfiguration)
+            .copy(timeModel = params.newTimeModel)
+
+          // Submit configuration to the ledger, and start polling for the result.
+          submissionId = SubmissionId.assertFromString(request.submissionId)
+          synchronousResponse = new SynchronousResponse(
+            new SynchronousResponseStrategy(
+              writeService,
+              index,
+              ledgerEndBeforeRequest,
+            ),
+            timeToLive = JDuration.ofMillis(params.timeToLive.toMillis),
           )
-        } else {
-          Future.unit
-        }
+          entry <- synchronousResponse.submitAndWait(
+            submissionId,
+            (params.maximumRecordTime, newConfig),
+          )(executionContext, materializer)
+        } yield SetTimeModelResponse(entry.configuration.generation)
 
-      // Create the new extended configuration.
-      newConfig = currentConfig
-        .map(config => config.copy(generation = config.generation + 1))
-        .getOrElse(ledgerConfiguration.initialConfiguration)
-        .copy(timeModel = params.newTimeModel)
-
-      // Submit configuration to the ledger, and start polling for the result.
-      submissionId = SubmissionId.assertFromString(request.submissionId)
-      synchronousResponse = new SynchronousResponse(
-        new SynchronousResponseStrategy(
-          writeService,
-          index,
-          ledgerEndBeforeRequest,
-        ),
-        timeToLive = JDuration.ofMillis(params.timeToLive.toMillis),
-      )
-      entry <- synchronousResponse.submitAndWait(
-        submissionId,
-        (params.maximumRecordTime, newConfig),
-      )(executionContext, materializer)
-    } yield SetTimeModelResponse(entry.configuration.generation)
-
-    response.andThen(logger.logErrorsOnCall[SetTimeModelResponse])
-  }
+        response.andThen(logger.logErrorsOnCall[SetTimeModelResponse])
+    }
 
   private case class SetTimeModelParameters(
       newTimeModel: v1.TimeModel,
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPackageManagementService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPackageManagementService.scala
@@ -22,6 +22,8 @@ import com.daml.ledger.participant.state.v1.{SubmissionId, SubmissionResult, Wri
 import com.daml.lf.archive.{Dar, DarReader, Decode}
 import com.daml.lf.engine.Engine
 import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.platform.apiserver.services.logging
+import com.daml.logging.LoggingContext.withEnrichedLoggingContext
 import com.daml.platform.api.grpc.GrpcApiService
 import com.daml.platform.apiserver.services.admin.ApiPackageManagementService._
 import com.daml.platform.server.api.validation.ErrorFactories
@@ -64,6 +66,7 @@ private[apiserver] final class ApiPackageManagementService private (
   override def listKnownPackages(
       request: ListKnownPackagesRequest
   ): Future[ListKnownPackagesResponse] = {
+    logger.info("Listing known packages")
     packagesIndex
       .listLfPackages()
       .map { pkgs =>
@@ -81,7 +84,7 @@ private[apiserver] final class ApiPackageManagementService private (
 
   private[this] val darReader = DarReader[Archive] { case (_, x) => Try(Archive.parseFrom(x)) }
 
-  def decodeAndValidate(stream: ZipInputStream): Try[Dar[Archive]] =
+  private def decodeAndValidate(stream: ZipInputStream): Try[Dar[Archive]] =
     for {
       dar <- darReader.readArchive("package-upload", stream)
       packages <- Try(dar.all.iterator.map(Decode.decodeArchive).toMap)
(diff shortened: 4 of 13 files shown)


kamil-da:
@mziolekda Could you review changes and point out which information worth logging I haven't included in this draft?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPackageManagementService.scala
```diff
@@ -65,6 +67,7 @@ private[apiserver] final class ApiPackageManagementService private (
   override def listKnownPackages(
       request: ListKnownPackagesRequest
   ): Future[ListKnownPackagesResponse] = {
+    logger.info("Listing known packages")
```
stefanobaghino-da, line 70: Should you add something to the logging context here?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiParticipantPruningService.scala
```diff
@@ -39,33 +41,35 @@ final class ApiParticipantPruningService private (
...
-    submissionIdOrErr.fold(
-      Future.failed,
-      submissionId =>
-        LoggingContext.withEnrichedLoggingContext("submissionId" -> submissionId) {
-          implicit logCtx =>
-            (for {
+      submissionIdOrErr.fold(
+        Future.failed,
+        submissionId =>
+          LoggingContext.withEnrichedLoggingContext("submissionId" -> submissionId) {
```
stefanobaghino-da, line 57: It looks like the submission identifier has already been put in the logging context in the first line of this method, is it not?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
```diff
@@ -57,6 +59,7 @@ private[apiserver] final class ApiPartyManagementService private (
   override def getParticipantId(
       request: GetParticipantIdRequest
   ): Future[GetParticipantIdResponse] = {
+    logger.info("Getting Participant ID")
```
stefanobaghino-da, line 62: Is there some submission identifier that should be added to the logging context here?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
```diff
@@ -69,52 +72,59 @@ private[apiserver] final class ApiPartyManagementService private (
...
-          .submitAndWait(CreateSubmissionId.withPrefix(party), (party, displayName))
-          .map { case PartyEntry.AllocationAccepted(_, partyDetails) =>
-            AllocatePartyResponse(
-              Some(
-                PartyDetails(
-                  partyDetails.party,
-                  partyDetails.displayName.getOrElse(""),
-                  partyDetails.isLocal,
+  override def allocateParty(request: AllocatePartyRequest): Future[AllocatePartyResponse] =
+    withEnrichedLoggingContext(logging.party(request.displayName)) { implicit loggingContext =>
```
stefanobaghino-da, line 94: Would it make more sense to rather have the `partyIdHint` here? Even though both are optional, the `displayName` has more relaxed uniqueness requirements and it's probably less useful to track information when debugging.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
```diff
@@ -69,52 +72,59 @@ private[apiserver] final class ApiPartyManagementService private (
...
+      logger.info("Allocating party")
+      logger.debug(s"Party ID hint: '${request.partyIdHint}'")
```
stefanobaghino-da, line 96: Ideally this kind of information should be part of the logging context. In this way, it can also be used both for normal and structured logging.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -292,6 +294,7 @@ private class JdbcLedgerDao(
       offsetStep: OffsetStep,
       partyEntry: PartyLedgerEntry,
   )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info(s"Storing party entry")
```
stefanobaghino-da, line 297:
```suggestion
    logger.info("Storing party entry")
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -450,21 +453,25 @@ private class JdbcLedgerDao(
...
         preparedInsert.writeState(metrics)(_)
       )
       .map(_ => Ok)(servicesExecutionContext)
+  }
 
   override def storeTransactionEvents(
       preparedInsert: PreparedInsert
-  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info(s"Storing transaction events")
```
stefanobaghino-da, line 468:
```suggestion
    logger.info("Storing transaction events")
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -489,20 +496,23 @@ private class JdbcLedgerDao(
...
-        validate(ledgerEffectiveTime, transaction, divulged) match {
-          case None =>
-            preparedInsert.writeState(metrics)
-            preparedInsert.writeEvents(metrics)
-            insertCompletions(submitterInfo, transactionId, recordTime, offsetStep)
-          case Some(error) =>
-            submitterInfo.foreach(handleError(offsetStep.offset, _, recordTime, error))
-        }
+    withEnrichedLoggingContext(Logging.transactionId(transactionId)) { implicit loggingContext =>
+      logger.info(s"Storing transaction")
```
stefanobaghino-da, line 500:
```suggestion
      logger.info("Storing transaction")
```

libs-scala/contextualized-logging/BUILD.bazel
```diff
@@ -37,6 +37,8 @@ da_scala_test_suite(
     name = "contextualized-logging-tests",
     srcs = glob(["src/test/suite/**/*.scala"]),
     scala_deps = [
+        "@maven//:com_typesafe_akka_akka_actor",
+        "@maven//:com_typesafe_akka_akka_stream",
```
stefanobaghino-da, line 41: I don't think that this library should depend on Akka. Do you think you could move the helper to the only place where it's used?

stefanobaghino-da (commented):
Thanks, it basically looks ok, I've added a few notes here and there.

libs-scala/contextualized-logging/BUILD.bazel
kamil-da: It already depends on Akka (which is a bit strange, I agree). I've only added Akka to test suite dependencies. I could extract stream utils into a separate helper but I'm not sure if that's relevant for this PR. What do you think?

stefanobaghino-da: Ha, you're right, thanks for making me notice. Feel free to split the Akka part off, but don't feel forced to. 🙂

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
kamil-da: I agree - I just wasn't sure if that wouldn't cause context cluttered.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPackageManagementService.scala
kamil-da: There's no identifier available in the request. Should we generate one in such case?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiParticipantPruningService.scala
kamil-da: Good catch!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
kamil-da: The same as above - no identifier available. What to do in such case?

stefanobaghino-da: Nothing, really. I thought that was the case, but felt like commenting just to made sure. Thanks for confirming!

(same applies to other comments)

kamil-da: Got it, thanks!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
stefanobaghino-da: It's definitely not a problem, put in the context everything you think is valuable for post-hoc observability. Consider that this is printed on the console only when using the normal logging configuration, in production you would use a structured logging framework and the extra information would not be in the message.

kamil-da: Thanks for the explanation. I'll add the party hint to the context then 👍

libs-scala/contextualized-logging/BUILD.bazel
kamil-da: I'll address that in a separate PR.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiParticipantPruningService.scala
kamil-da: Fixed.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
kamil-da: Fixed.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPartyManagementService.scala
kamil-da: Fixed.

stefanobaghino-da (approved):
LGTM (but maybe wait for someone from your team to approve as well, if you think that makes sense)

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -72,9 +72,16 @@ private[apiserver] final class ApiTransactionService private (
...
+        .via(logger.debugStream(transactionsLoggable))
         .via(logger.logErrorsOnStream)
     }
 
+  private def transactionsLoggable(transactions: GetTransactionsResponse): String =
+    s"Responding with transactions: ${transactions.transactions.map(_.toProtoString).mkString("[", ",", "]")}"
+
+  private def transactionTreesLoggable(trees: GetTransactionTreesResponse): String =
+    s"Responding with transaction trees: ${trees.transactions.map(_.toProtoString).mkString("[", ",", "]")}"
+
```
kamil-da, line 84: I would appreciate a comment on whether we want complete transactions and transaction trees logged or should we log only a selection of fields? @mziolekda

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
```diff
@@ -49,9 +49,13 @@ private[apiserver] final class ApiCommandCompletionService private (
 
         completionsService
           .getCompletions(offset, request.applicationId, request.parties)
+          .via(logger.debugStream(completionsLoggable))
           .via(logger.logErrorsOnStream)
     }
 
+  private def completionsLoggable(response: CompletionStreamResponse): String =
+    s"Responding with completions: ${response.completions.map(_.toProtoString).mkString("[", ",", "]")}"
```
mziolekda, line 57: We need to dial down the verbosity of this a bit. All we really need is a command id and status. Running it through proto introduces  yet another naming flavour i.e. `command_id`

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
mziolekda: Thanks, it is very thoughtful of you.
Those logs end up being way to verbose when passed through proto. We only need transactionId, commandId, workflowId and offset under those very names.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -294,6 +296,7 @@ private class JdbcLedgerDao(
       offsetStep: OffsetStep,
       partyEntry: PartyLedgerEntry,
   )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info("Storing party entry")
```
mziolekda, line 299: PartyEntry contains submissionId, please log it as well

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -452,21 +455,25 @@ private class JdbcLedgerDao(
 
   override def storeTransactionState(
       preparedInsert: PreparedInsert
-  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info("Storing transaction state")
```
mziolekda, line 459: We can skip this log, the final one when completing is enough

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -452,21 +455,25 @@ private class JdbcLedgerDao(
...
         preparedInsert.writeState(metrics)(_)
       )
       .map(_ => Ok)(servicesExecutionContext)
+  }
 
   override def storeTransactionEvents(
       preparedInsert: PreparedInsert
-  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info("Storing transaction events")
```
mziolekda, line 470: We can skip this log, the final one when completing is enough

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -491,20 +498,23 @@ private class JdbcLedgerDao(
...
-      .executeSql(metrics.daml.index.db.storeTransactionDbMetrics) { implicit conn =>
-        validate(ledgerEffectiveTime, transaction, divulged) match {
-          case None =>
-            preparedInsert.writeState(metrics)
-            preparedInsert.writeEvents(metrics)
-            insertCompletions(submitterInfo, transactionId, recordTime, offsetStep)
-          case Some(error) =>
-            submitterInfo.foreach(handleError(offsetStep.offset, _, recordTime, error))
-        }
+    withEnrichedLoggingContext(Logging.transactionId(transactionId)) { implicit loggingContext =>
```
mziolekda, line 501: This message looks unnecessarily verbose in the logfile and every field is prefixed with `update`:
```
updateCommandId=PRPruneQueryEmptyRangeOk-alpha-39f3ccbfdb89-command-220, updateLedgerTime=2021-02-03T15:58:31.515Z, updateOffset=0000000000000a190000000000000000, updateSubmitter=[PRPruneQueryEmptyRangeOk-alpha-39f3ccbfdb89-party-0], transactionId=0A2463633665333436652D373965652D333034322D613862642D653164633433626661636463, updateTransactionId=0A2463633665333436652D373965652D333034322D613862642D653164633433626661636463})
```
Would it be possible to dial it down a bit?

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -544,19 +554,22 @@ private class JdbcLedgerDao(
       recordTime: Instant,
       offsetStep: OffsetStep,
       reason: RejectionReason,
-  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info("Storing rejection")
```
mziolekda, line 558: Submitter info contains submissionId, please log it when provided

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
```diff
@@ -720,7 +734,8 @@ private class JdbcLedgerDao(
       offsetStep: OffsetStep,
       packages: List[(Archive, PackageDetails)],
       optEntry: Option[PackageLedgerEntry],
-  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] = {
+    logger.info("Storing package entry")
```
mziolekda, line 738: PackageLedgerEntry contains submissionId, please log it when provided

mziolekda (changes requested):

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
stefanobaghino-da: I would recommend keeping it as is. Consider that you only see this when using a console logger, in production you should use structured logging and this would not be duplicated in the message.

mziolekda: but why do we have the prefix, i.e. updateCommandId vs commandId everywhere else, and why do we have transactionId and updateTransactionId both with the same content?

stefanobaghino-da: Had a look at the code, the transaction identifier is injected in the context upstream, as soon as it becomes available, using `loggingContextFor` in the trait `com.daml.platform.indexer.ExecuteUpdate`. I believe there's a number of keys that is being inserted twice.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
stefanobaghino-da: Re: this and other comments.

Before acting on this you may want to check whether these are already injected in the logging context upstream.

See `ExecuteUpdate#loggingContextFor`

```scala
      case PartyAllocationRejected(submissionId, participantId, _, rejectionReason) =>
        Map(
          "updateSubmissionId" -> submissionId,
          "updateParticipantId" -> participantId,
          "updateRejectionReason" -> rejectionReason,
        )
      case PublicPackageUpload(_, sourceDescription, _, submissionId) =>
        Map(
          "updateSubmissionId" -> submissionId.getOrElse(""),
          "updateSourceDescription" -> sourceDescription.getOrElse(""),
        )
```

stefanobaghino-da:
Judging from @mziolekda's most recent comments it appears that many of the keys that are being inserted in this PR were actually already there. It would probably be best to have a second look at this.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: The `ExecuteUpdate#loggingContextFor` method enriches logging context for data writes in `JdbcLedgerDao`. There's no need to enrich the logging context in the latter.
However, I think we should consider removing `update` prefix from the context keys to be consistent with other places where we log correlation ids (`commandId`, `transactionId`, `submissionId`). What do you think @stefanobaghino-da @mziolekda ?

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Same comment as [above](https://github.com/digital-asset/daml/pull/8676#discussion_r571638038).

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Fixed.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Same comment as [above](https://github.com/digital-asset/daml/pull/8676#discussion_r571638038).

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Same comment as [above](https://github.com/digital-asset/daml/pull/8676#discussion_r571638038).

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Same comment as [above](https://github.com/digital-asset/daml/pull/8676#discussion_r571638038).

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
kamil-da: Fixed 👍

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
kamil-da: Fixed 👍

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
mziolekda: I am strongly for removing that prefix. It leaks a bit of architecture detail to the log (updateCommandId differs from commandId only by the place in the code that prints them). For an end-user or support personnel it can be confusing.

mziolekda: Also, the chain of logging context must be broken somewhere, because the example you posted elsewhere did not contain a context:
```
INFO: Storing package entry
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
mziolekda: It is still here :-)

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
```diff
@@ -49,9 +49,23 @@ private[apiserver] final class ApiCommandCompletionService private (
...
+    s"Responding with completions: ${response.completions.toList
+      .map(c => singleCompletionLoggable(c.commandId, c.status.map(_.code)))}"
+
+  private def singleCompletionLoggable(
+      commandId: String,
+      statusCode: Option[Int],
+  ): Map[String, String] =
+    Map(
+      logging.commandId(commandId),
+      "statusCode" -> statusCode.map(_.toString).getOrElse(""),
```
tudor-da, line 66:
```suggestion
      "statusCode" -> statusCode.mkString,
```

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala
```diff
@@ -57,11 +59,13 @@ private[apiserver] final class ApiConfigManagementService private (
   override def bindService(): ServerServiceDefinition =
     ConfigManagementServiceGrpc.bindService(this, executionContext)
 
-  override def getTimeModel(request: GetTimeModelRequest): Future[GetTimeModelResponse] =
+  override def getTimeModel(request: GetTimeModelRequest): Future[GetTimeModelResponse] = {
+    logger.info("Getting time model")
```
tudor-da, line 63: Does this log line bring enough in its current form? Should it be `info` ?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
kamil-da: Good catch!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala
kamil-da: That's a good question. There's no information in the request itself that is suitable for putting in the `LoggingContext`. We could create a correlation id at this point but according to @stefanobaghino-da 's comment we're fine with this simple logging. Please see [this thread](https://github.com/digital-asset/daml/pull/8676#discussion_r568441915).

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
kamil-da: Actually it's not possible with our current Scala compiler settings. Specifically, compilation will fail with:
```
error: [wartremover:Option2Iterable] Implicit conversion from Option to Iterable is disabled - use Option#toList instead
```
I could use `statusCode.toList.mkString` but it looks a bit strange to create a list only for `mkString`. What do you think?

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: From what I see in the codebase, there's no broken chain of logging context but rather the same method for storing package entries is used in different places. In some of them the logging context is not used at all. However, for the cases pertinent to this PR, the logging context is preserved.

kamil-da: > I am strongly for removing that prefix. It leaks a bit of architecture detail to the log (updateCommandId differs from commandId only by the place in the code that prints them). For an end-user or support personnel it can be confusing.

I've removed the prefix. The `LoggingContext`'s keys should now have uniform naming.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: Fixed fixed.

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
```diff
@@ -243,63 +256,98 @@ trait ExecuteUpdate {
...
+    def rejectionReason(reason: String): (String, String) =
+      "rejectionReason" -> reason
+    def displayName(name: String): (String, String) =
+      "displayName" -> name
+    def sourceDescriptionOpt(description: Option[String]): (String, String) =
+      "sourceDescription" -> description.getOrElse("")
+    def submitter(parties: List[Party]): (String, String) =
+      "submitter" -> parties.mkString("[", ", ", "]")
+
+  }
```
kamil-da, line 350: I've created this object to ensure consistent logging for different keys. This is very similar to `com.daml.platform.apiserver.services.logging`. I wonder if it would be worth merging both of these helper utils. What do you think @mziolekda ?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala
mziolekda: those simple get* requests have very little meat on the bone. They are also very shallow i.e. don't reach to committer. It is ok to log little info here.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
mziolekda: Let's have a look at the example log-line from an actual run and close this thread if the context appears

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
mziolekda: I am ok with splitting it to a follow-up PR. This one is getting big as it stands.

ledger/participant-integration-api/src/main/scala/platform/store/dao/JdbcLedgerDao.scala
kamil-da: An example log line for this case is:
```
2021-02-10T18:07:01.537Z INFO  c.d.p.s.d.JdbcLedgerDao@[kka.actor.default-dispatcher-6] -
                Storing package entry (context: {participantId=sandbox-participant})
```

mziolekda (approved):
It looks ok now, thanks

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
```diff
@@ -110,7 +110,7 @@ private[apiserver] final class ApiSubmissionService private (
 
   override def submit(request: SubmitRequest): Future[Unit] =
     withEnrichedLoggingContext(logging.commands(request.commands)) { implicit loggingContext =>
-      logger.debug("Submitting transaction")
+      logger.info("Submitting transaction")
```
oliverse-da, line 113: @kamil-da , don't you feel logging every submission at info level might turn out to be a bit verbose?

kamil-da: @oliverse-da Thank you for the comment.
Actually I think it may be more reasonable to lower the level to debug.
This change was explicitly requested by @mziolekda , so his input might be valuable here as well.

PR #8684, opened 2021-01-29
Disable deriving-defaults per file

cocreature:
Disabling it per target works nicely for compilation but it gets
annoying in ghci since the warnings are still triggered. We could
disable it everywhere but I think the warning is generally useful. I
tried patching proto3-suite to use DerivingStrategies but that doesn’t
work because haskell-src is dead and doesn’t support it. So for now
adding it to the per-file list seems like the best option.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Thanks!

PR #8689, opened 2021-01-29
Deprecate DA.Next.Set and DA.Next.Map

### LINKED ISSUE #8449: Deprecate `DA.Next.Map` and `DA.Next.Set`
In favour of the new GenMap counterparts. Depends on #8448

sofiafaro-da:
Fixes #8449

changelog_begin

- [DAML Standard Library] DA.Next.Map and DA.Next.Set are deprecated.
  Please use DA.Map and DA.Set instead, or use DA.TextMap directly
  for ledger API backward compatibility.

changelog_end

### Pull Request Checklist

trigger the build.


compiler/damlc/tests/daml-test-files/Examples.daml
```diff
@@ -36,8 +36,8 @@ import Trace()
 import PreludeTest()
 import Text()
 import Optional()
-import NextMap()
-import NextSet()
+-- import NextMap()
```
cocreature, line 39: can we just delete it? Or if it actually matters, add the deprecation to the expected warnings. We probably need to split it in 2, one with `SINCE-LF` and one with `UNTIIL-LF` since only one triggers the warning.

compiler/damlc/tests/daml-test-files/Stdlib.daml
```diff
@@ -16,8 +16,6 @@ import DA.List()
 import DA.Logic()
 import DA.Maybe()
 import DA.Monoid()
-import DA.Next.Map()
```
cocreature, line 19: is this file useful for anything? if so, why don’t we need those modules anymore?

cocreature (approved):
Thanks!

compiler/damlc/tests/daml-test-files/Examples.daml
sofiafaro-da: I don't think it matters, I can just delete these lines.

compiler/damlc/tests/daml-test-files/Stdlib.daml
sofiafaro-da: The file describes itself as "Check all the modules in stdlib compile without warnings"

It doesn't make sense to have DA.Next.Map / DA.Next.Set in here anymore since they are supposed to add deprecation warnings. That said this test is not up-to-date. I'd either delete it or update it in a separate PR.

PR #8690, opened 2021-01-29
ledger-api-test-tool: Provide context when expecting a failing Future.

SamirTalwar-DA:
We use `Future#failed` a _lot_ in the conformance tests. Usually, this works fine, but it has the unfortunate effect of yielding a useless error message when the future was actually a success:

```
Future.failed not completed with a throwable.
```

Combined with the lack of useful stack traces, because futures, this makes the test failure completely useless.

This changes all calls of `Future#failed` in our conformance tests to use a new extension method, `#mustFail`, which takes a mandatory "context" parameter to provide context, and includes both the context and the value in the future in case of success.

### Pull Request Checklist

trigger the build.


fabiotudone-da (approved):
That's a significant usability improvement, thanks!

PR #8693, opened 2021-01-29
Upgrade gRPC C library

cocreature:
We can’t upgrade to the latest version 1.35.0 since it depends on a
newer version of Bazel but we can at least go to 1.34.1. We also have
to patch absl and grpc itself to fix mingw support and some undeclared
inclusions on Windows. Both patches are taken from upstream PRs (the
mingw one has not been merged yet).

There is also a small patch to grpc-haskell-core. I’ve started working
on upstreaming that but the nix build system is a mess and I haven’t
quite managed to get it working there.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


bazel_tools/grpc-haskell-core-upgrade.patch
```diff
@@ -0,0 +1,70 @@
...
+ 
+-{#fun grpc_channel_ping as ^
+-  {`Channel', `CompletionQueue', unTag `Tag',unReserved `Reserved'} -> `()' #}
+-
+ {#fun grpc_channel_destroy as ^ {`Channel'} -> `()'#}
+ 
+ -- | Starts executing a batch of ops in the given 'OpArray'. Does not block.
+diff --git a/tests/LowLevelTests/Op.hs b/tests/LowLevelTests/Op.hs
+index 52b6d55..ff0751a 100644
+--- a/tests/LowLevelTests/Op.hs
```
cocreature, line 31: this doesn’t actually matter to us since it’s only a test but I feel more confident including the full patch including the fix for the tests here. Also makes it easier to upstream since we don’t have to figure out how to fix the tests again.

bazel_tools/grpc-haskell-core-upgrade.patch
```diff
@@ -0,0 +1,70 @@
...
+ //This is a hack to work around GHC being unable to deal with raw struct params.
+diff --git a/src/Network/GRPC/Unsafe.chs b/src/Network/GRPC/Unsafe.chs
+index d11f533..6eb08b8 100644
+--- a/src/Network/GRPC/Unsafe.chs
++++ b/src/Network/GRPC/Unsafe.chs
+@@ -219,9 +219,6 @@ castPeek p = do
+     unTag `Tag'}
+   -> `()'#}
+ 
+-{#fun grpc_channel_ping as ^
```
cocreature, line 23: this has just been removed upstream and we don’t need it.

bazel_tools/grpc-haskell-core-upgrade.patch
```diff
@@ -0,0 +1,70 @@
...
+index ebbcbb1..236c73e 100644
+--- a/cbits/grpc_haskell.c
++++ b/cbits/grpc_haskell.c
+@@ -504,8 +504,7 @@ grpc_auth_metadata_processor* mk_auth_metadata_processor(
+ 
+ grpc_call_credentials* grpc_metadata_credentials_create_from_plugin_(
+   grpc_metadata_credentials_plugin* plugin){
+-
+-  return grpc_metadata_credentials_create_from_plugin(*plugin, NULL);
++  return grpc_metadata_credentials_create_from_plugin(*plugin, GRPC_PRIVACY_AND_INTEGRITY, NULL);
```
cocreature, line 11: this technically could be make configurable but the python bindings also don’t make it configurable atm. This is the highest security level which afaict was the default before as well.

aherrmann-da (approved):
Thank you!

PR #8695, opened 2021-01-29
Run the ledger api test tool with dev packages

rautenrieth-da:
This PR:
- Extends `//ledger/test-common` to also compile the test DAML code with the DAML-LF dev version
- Extends the `conformance_test` bzl function so that it's possible to override the location of the bundled dar files
- Extends `//ledger/ledger-api-test-tool` to run the ledger api test tool with the dev-version-compiled dar files against `ledger-on-memory`

**Note**: This PR will fail against the current master, but should succeed once https://github.com/digital-asset/daml/pull/8656 is merged (tested locally).


ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -243,3 +243,16 @@ conformance_test(
         "--include=IdentityIT:IdNotEmpty",
     ],
 )
+
+conformance_test(
+    name = "damlf-dev-version",
```
SamirTalwar-DA, line 248:
```suggestion
    name = "daml-lf-dev-version",
```

ledger/test-common/BUILD.bazel
```diff
@@ -37,40 +38,42 @@ da_scala_library(
...
         ),
         (
             "semantic",
-            lf_stable_version,
             ["@maven//:org_scala_lang_modules_scala_collection_compat_2_12"],
         ),
-        ("performance", lf_stable_version, []),
+        ("performance", []),
+    ]
+    for (target, target_name) in [
```
SamirTalwar-DA, line 74: How do you feel about `target_suffix` instead of `target_name`?

SamirTalwar-DA (approved):

SamirTalwar-DA:
`TransactionServiceIT` failed, but it looks flaky to me. Does this somehow consume more resources?

rautenrieth-da:
> TransactionServiceIT failed, but it looks flaky to me. Does this somehow consume more resources?

Actually, the test should never have succeeded (not until https://github.com/digital-asset/daml/pull/8656 is merged). It turns out my PR was wrong, I have pushed a second attempt.

ledger/ledger-api-test-tool/conformance.bzl
```diff
@@ -25,9 +25,7 @@ def conformance_test(
         timeout = "long",
         client = "//ledger/ledger-api-test-tool",
         client_args = test_tool_args + ["localhost:%s" % port for port in ports],
-        data = extra_data + [
-            "//ledger/test-common:dar-files",
```
rautenrieth-da, line 29: By default, the DAR files are bundled with the test tool binary as resources, they don't need to be added as data here.

mziolekda (approved):
Looks good

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -322,6 +322,13 @@ object Config {
           "Enable preview version of the next Daml-LF language. Should not be used in production."
         )
 
+      opt[Unit]("daml-lf-dev-mode")
```
miklos-da, line 325: Why not add a single command-line parameter for setting the supported Daml-LF version? For example, the user could switch between the non-stable versions via `--use-daml-lf=early-access` or `--use-daml-lf=dev`. Another option is to have two separate arguments such as `--use-early-access-daml-lf` or `--use-dev-daml-lf`.
What do you think?

rautenrieth-da: > Another option is to have two separate arguments such as --use-early-access-daml-lf or --use-dev-daml-lf.

That's how it works right now, but the names are not very nice. We have `--daml-lf-dev-mode-unsafe` (a new, hidden option added in this PR) and `--early-access` (an existing option, added in a different PR today). There were some discussions regarding the naming of the `--early-access` option, I'd leave it to a separate PR to rename both options (cc @mziolekda).

rautenrieth-da:
FYI to the reviewers (@miklos-da @SamirTalwar-DA @mziolekda): I have completely rewritten the PR. My first attempts failed and I realized the ledger API test tool has a compile time dependency on the Daml-Lf compiler version (through the scala codegen), so I had to duplicate the entire ledger API test tool build chain in Bazel for each compiler version we want to test.

@moritzkiefer-da @remyhaemmerle-da This PR should run the ledger API test tool with `latest`, `preview`, and `dev` compiler versions against `ledger-on-memory`. It probably increases the build time of `//ledger/...` by a few minutes, because of the duplicated builds.

ledger/test-common/BUILD.bazel
```diff
@@ -15,10 +15,24 @@ load(
...
 
+# Tuples of (lf_version, name_suffix)
+# The DAML files are compiled once for each daml-lf version listed here.
+test_common_configurations = [
+    (lf_stable_version, ""),
+    (lf_latest_version, "-latest"),
+    (lf_dev_version, "-dev"),
+] + [(
+    lf_preview_version[0],
+    "-preview",
```
remyhaemmerle-da, line 33: `lf_preview_version` contains at most one version, but may be empty. 
Think about it as an option (I did not find option in starlark, so I used list)

you can look how I deal with that in [sandbox-classic-tests](https://github.com/digital-asset/daml/blob/f13c96bddef5a71b4de876535462fc1f9a10c6d4/ledger/sandbox-classic/BUILD.bazel#L242) (It is ugly I know)

rautenrieth-da: That's what happens here as well, see the line below:
```
if lf_preview_version != [] else []
```

remyhaemmerle-da: I missed that. Sorry for the noise.

rautenrieth-da: I have updated the code to use the same list comprehension as in sandbox-classic-tests.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
miklos-da: @rautenrieth-da I'm fine with renaming the command-line options in a separate PR. Who needs to agree on the new, consistent set of names between sandbox and kvutils-app?

PR #8697, opened 2021-01-29
Bump JVM on Darwin and force TLS 1.2 

cocreature:
Details are in the comments but this seems to workaround the annoying
tls issues we have been dealing with for the past few days.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


nix/nixpkgs.nix
```diff
@@ -8,6 +8,20 @@ let
 
   # package overrides
   overrides = _: pkgs: rec {
+    # nixpkgs ships with a very old version of openjdk on darwin
```
cocreature, line 11: I don’t think this is actually required but the fact that we are on a pretty old JDK doesn’t inspire confidence and I’d rather not waste time debugging issues that have already been fixed.

nix/nixpkgs.nix
```diff
@@ -27,7 +41,7 @@ let
         # and/or documentation.
         (pkgs.fetchurl {
           url = "https://patch-diff.githubusercontent.com/raw/bazelbuild/bazel/pull/8983.patch";
-          sha256 = "1j25bycn9q7536ab3ln6yi6zpzv2b25fwdyxbgnalkpl2dz9idb7";
+          sha256 = "1qdyqymsylinkdwqhbxbm8bvbyznrdn74n744pi5xhdwb6lw1r8a";
```
cocreature, line 44: this is unrelated, changing the JVM forced a rebuild and needed a hash change because this file has changed upstream. There don’t seem to be any relevant changes and the PR has now been closed (looks like the change has made it into Bazel 4.0) so hopefully it will stay stable until we get to Bazel 4.0 and can drop it.

stefanobaghino-da (approved):

aherrmann-da (approved):

PR #8699, opened 2021-01-29
lsp-tests: test for no internal imports

robin-da:
We add a test to check that there are no internal imports returned by
the `getDocumentSymbols` endpoint for an empty module.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/lsp-tests/src/Main.hs
+++ b/compiler/lsp-tests/src/Main.hs
@@ -5,7 +5,7 @@
 module Main (main) where
 
 import Control.Applicative.Combinators
-import Control.Lens hiding (List)
+import Control.Lens hiding (List, children)
 import Control.Monad
 import Control.Monad.IO.Class
 import DA.Bazel.Runfiles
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

robin-da:
@cocreature note that the result in Daml studio is a lot better, we still see internal modules for `HasField` instances of templates.

cocreature:
Do you mean we get imports for `HasField` stuff or we just have the instances which we should probably also filter out? Afaict, the imports are marked as generated. I agree on filtering out the instances, unfortunately there is no flag for generated instances so we’ll have to match on name there at least for `HasField`. Not sure if there are other instances worth filtering out.

robin-da:
@cocreature here is how it looks with one template:
2021-01-29-220123_949x882_scrot

cocreature:
makes sense, we should probably filter out everything in `DA.Internal.Record`, `DA.Internal.Desugar` and the `_choice_` stuff.

cocreature:
I’d perhaps try adding a hook to ghcide where we can plug in the filtering mechanism from the daml repo.

robin-da:
@cocreature you have something specific in mind for the hook?

cocreature:
good question, I think it might be easiest to just postprocess the resulting symbols so something of type `List DocumentSymbol -> List DocumentSymbol` (haven’t checked the exact name). Looking at this again, I think `PartialHandlers` might already allow us to override the document symbol header and postprocess the result of the existing handler but it’s been a while since I looked at this.

PR #8708, opened 2021-02-01
lsp-tests: test for no internal symbols

robin-da:
This updates the daml-ghcide dependency and also adds a test to check
that we are not showing any internal symbols and dependencies in the
outline in daml studio.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/lsp-tests/src/Main.hs
+++ b/compiler/lsp-tests/src/Main.hs
@@ -717,6 +717,23 @@ symbolsTests run =
               foo <- openDoc' "Foo.daml" damlId $ T.unlines ["module Foo where"]
               syms <- getDocumentSymbols foo
               liftIO $ preview (_Left . _head . children . _Just) syms @?= Just (List [])
+        , testCase "no internal imports for module with one template" $
+          run $ do
+              foo <-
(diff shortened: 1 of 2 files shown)


compiler/lsp-tests/src/Main.hs
```diff
@@ -717,6 +717,16 @@ symbolsTests run =
               foo <- openDoc' "Foo.daml" damlId $ T.unlines ["module Foo where"]
               syms <- getDocumentSymbols foo
               liftIO $ preview (_Left . _head . children . _Just) syms @?= Just (List [])
+        , testCase "no internal imports for module with one template" $
+          run $ do
+              foo <-
+                  openDoc' "Foo.daml" damlId $
+                  T.unlines
+                      ["module Foo where", "template T with p1 : Party where", "  signatory p1"]
```
cocreature, line 725: Maybe worth adding a choice in here as well?

cocreature (approved):
Thanks!

compiler/lsp-tests/src/Main.hs
robin-da: added a choice.

Issue #8634, opened 2021-01-26
Non-repudiation: add persistent, SQL-backed signed commands repo


stefanobaghino-da:
The current prototype of the non-repudiation middleware stores signed commands in a volatile in-memory map. Add a persistent, SQL-backed storage for signed commands. PostgreSQL will likely be the backend used for the first early access versions but a Sqlite-backed version may be evaluated for testing purposes.

PR #8709, opened 2021-02-01
Revise signed payload repository

stefanobaghino-da:
A few improvements made as part of https://github.com/digital-asset/daml/issues/8634.

- revised signed payload repository to hold more relevant data on the signed payload
- more specific types when handling bytes and strings

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/SignedPayload.scala
```diff
@@ -0,0 +1,24 @@
...
+      algorithm: AlgorithmString,
+      fingerprint: FingerprintBytes,
+      payload: PayloadBytes,
+      signature: SignatureBytes,
+  ): SignedPayload =
+    new SignedPayload(algorithm, fingerprint, payload, signature)
+
+}
+
+// Purposefully not a case class because it contains mutable arrays
```
cocreature, line 18: Do we actually need to carry mutable arrays through everything or could we wrap it in `ArraySeq.unsafeWrapArray` once at the library border and then stop worrying about it?

cocreature (approved):
Nice, thank you!

runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/SignedPayload.scala
stefanobaghino-da: That's a very good point, thanks!

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8709/commits/a5e1b941f3408e9456d6d6a183b382ada2b2cae3

PR #8711, opened 2021-02-02
implement base64 encoding/decoding as builtin

kpinter-da:
### Pull Request Checklist

trigger the build.


cocreature:
What exactly are you trying to achieve by base64 encoding? Usually, that’s intended to represent binary data as text but you already have text as the input here. Do you need to encode unicode as ascii for some reason? If so, could you just escape special characters like we do in DAML-LF instead of going full base64?

Assuming you cannot avoid base64 encoding, do you have benchmarks to demonstrate that this is a significant performance improvement?

PR #8716, opened 2021-02-02
Ensure the pre-load executor is reused across calls

stefanobaghino-da:
changelog_begin
changelog_end

The current behavior is that of creating a new preload executor at every call,
which is a very complex way of simply spawning a thread and defeats the purpose
of having a shared executor, sequencing the pre-loading of packages to maximize
the usage of cache (among other things).

### Pull Request Checklist

trigger the build.


SamirTalwar-DA (approved):
:ship:

I think there's probably a better solution for this, but I think it involves Akka Streams, and I do _not_ want to introduce another one of those right now.

PR #8718, opened 2021-02-02
KV: flag to enable preview version in kvutils

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
+++ b/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
@@ -11,6 +11,7 @@ import java.util.concurrent.TimeUnit
 
 import com.daml.caching
 import com.daml.ledger.api.tls.TlsConfiguration
+import com.daml.ledger.participant.state.kvutils.app.Config.EngineMode
 import com.daml.ledger.participant.state.v1.ParticipantId
 import com.daml.ledger.participant.state.v1.SeedService.Seeding
 import com.daml.ledger.resources.ResourceOwner
@@ -37,6 +38,7 @@ final case class Config[Extra](
     metricsReporter: Option[MetricsReporter],
     metricsReportingInterval: Duration,
     trackerRetentionPeriod: FiniteDuration,
+    engineMode: EngineMode,
     extra: Extra,
 ) {
   def withTlsConfig(modify: TlsConfiguration => TlsConfiguration): Config[Extra] =
@@ -65,6 +67,7 @@ object Config {
       metricsReporter = None,
       metricsReportingInterval = Duration.ofSeconds(10),
       trackerRetentionPeriod = DefaultTrackerRetentionPeriod,
+      engineMode = EngineMode.Stable,
       extra = extra,
     )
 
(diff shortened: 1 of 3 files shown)


ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala
```diff
@@ -79,7 +80,10 @@ final class Runner[T <: ReadWriteService, Extra](
     )
     implicit val materializer: Materializer = Materializer(actorSystem)
 
-    val sharedEngine = Engine.StableEngine()
+    val sharedEngine =
```
cocreature, line 83: rather than hardcoding a different value here, should we just pass it through and make it configurable?

SamirTalwar-DA: Yes please. I would prefer if this was opt-in with a CLI argument.

remyhaemmerle-da: should the option be hidden ?

cocreature (approved):
LGTM, I don’t have a strong opinion on whether this should be hidden or not. Given the clear disclaimer it seems fine to expose it.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala
```diff
@@ -79,7 +81,13 @@ final class Runner[T <: ReadWriteService, Extra](
     )
     implicit val materializer: Materializer = Materializer(actorSystem)
 
-    val sharedEngine = Engine.StableEngine()
+    val allowedLanguageVersions =
+      config.engineMode match {
+        case EngineMode.Stable => LanguageVersion.StableVersions
+        case EngineMode.EarlyAccess => LanguageVersion.EarlyAccessVersions
+      }
+
```
miklos-da, line 89:
```suggestion
```

miklos-da (approved):

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala
SamirTalwar-DA: Eh, this runner is still intended mostly for people who know what they're doing, so it's probably fine.

(It was never intended to actually be used, but c'est la vie.)

SamirTalwar-DA:
How did you test this?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -312,10 +315,24 @@ object Config {
           success
       })
 
+      opt[Unit]("early-access-unsafe")
```
mziolekda, line 318: can we change it to just early-access?

remyhaemmerle-da:
```suggestion
      opt[Unit]("early-access")
```

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -312,10 +315,24 @@ object Config {
           success
       })
 
+      opt[Unit]("early-access")
```
cocreature, line 318: why do we want this to be inconsistent with sandbox? That just seems annoying. And it really is unsafe.

remyhaemmerle-da: cc @mziolekda

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
miklos-da: If we don't need to be consistent with sandbox then an option name that implies that it is about `lf` would be preferred (i.e., the expression "early access" is fairly generic).

PR #8722, opened 2021-02-02
Add a test for DeduplicatingPackageLoader

rautenrieth-da:
This PR adds a simple test for DeduplicatingPackageLoader, added in https://github.com/digital-asset/daml/pull/8656


mziolekda (approved):
LGTM
The fact that the outcome of repeated attempt after failure is time sensitive is a bit annoying. But I don't think this is a major issue. Plugging that hole would add a lot of complexity for marginal returns.

PR #8730, opened 2021-02-03
Retry asset downloads in check_releases

cocreature:
We’ve seen a number of "resource vanished (connection reset by peer)"
errors. Slapping some retries on that should hopefully make the CI job
a bit more robust.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -14,6 +14,7 @@ import Control.Exception.Safe
 import qualified Control.Monad as Control
 import qualified Control.Monad.Extra
 import qualified Control.Monad.Loops
+import Control.Retry
 import qualified Data.Aeson as JSON
 import qualified Data.ByteString
 import qualified Data.ByteString.UTF8 as BS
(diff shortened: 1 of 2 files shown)


ci/cron/src/Main.hs
```diff
@@ -317,20 +318,32 @@ download_assets :: FilePath -> GitHubRelease -> IO ()
...
+                retryPolicy
+                [retryHandler]
+                (\_ -> downloadFile req manager url)
+          )
   where while = Control.Monad.Loops.whileJust_
         readFrom body = ifNotEmpty <$> HTTP.brRead body
         ifNotEmpty bs = if Data.ByteString.null bs then Nothing else Just bs
         writeTo = Data.ByteString.hPut
+        -- Retry for 5 minutes total, doubling delay starting with 20ms
+        retryPolicy = limitRetriesByCumulativeDelay (5 * 60 * 1000 * 1000) (exponentialBackoff (20 * 1000))
```
stefanobaghino-da, line 337: Just to double check: the base unit for these is microseconds, right?

ci/cron/src/Main.hs
```diff
@@ -317,20 +318,32 @@ download_assets :: FilePath -> GitHubRelease -> IO ()
 download_assets tmp release = do
     manager <- HTTP.newManager TLS.tlsManagerSettings
     tokens <- Control.Concurrent.QSem.newQSem 20
-    Control.Concurrent.Async.forConcurrently_ (map uri $ assets release) (\url ->
+    Control.Concurrent.Async.forConcurrently_ (map uri $ assets release) $ \url ->
```
stefanobaghino-da, line 321: More `$` more happiness.

stefanobaghino-da (approved):

ci/cron/src/Main.hs
cocreature: yep, see https://hackage.haskell.org/package/retry-0.8.1.2/docs/Control-Retry.html#v:limitRetriesByCumulativeDelay
We really need units in the types system.

ci/cron/src/Main.hs
cocreature: @garyverhaegen-da is not going to be happy , sorry Gary :)

stefanobaghino-da: I don't think I am either, but I don't think I can really judge Haskell code properly. 😂

cocreature: I don’t feel too strongly about it so if you both prefer parentheses here I’m happy to go back. Personally, I find it quite hard to track parentheses beyond a few lines without an editor that does it for me so I tend to opt for `$` for larger blocks and parentheses for small stuff.

stefanobaghino-da: I was mostly joking at my own lack of experience with Haskell. 😉

garyverhaegen-da: Why would you ever use an editor that does not track parentheses? 😕

I really don't mind using `$`; I even use it myself sometimes.

garyverhaegen-da (approved):

ci/cron/src/Main.hs
cocreature: Reading git diffs unfortunately doesn’t track parentheses for me :slightly_smiling_face:

PR #8731, opened 2021-02-03
Add //libs-scala/doobie-slf4j and replace usages of jdkLogHandler

stefanobaghino-da:
The `jdkLogHandler` provided by Doobie exists purely as an example and the library
itself does not recommend using it in production.

```
/**
 * A LogHandler that writes a default format to a JDK Logger. This is provided for demonstration
 * purposes and is not intended for production use.
 * @group Constructors
 */
```

Note that this slightly changes the runtime behavior, logging successful queries
at debug level rather then info. The message itself is preserved from the original
MIT-licensed example.

This uses Slf4j as most of our components, instead of java.util.logging.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/libs-scala/doobie-slf4j/src/main/scala/com/daml/doobie/logging/Slf4jLogHandler.scala
+++ b/libs-scala/doobie-slf4j/src/main/scala/com/daml/doobie/logging/Slf4jLogHandler.scala
@@ -1,0 +1,43 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.doobie.logging
+
+import org.slf4j.{Logger, LoggerFactory}
+import doobie.util.log.{ExecFailure, LogHandler, ProcessingFailure, Success}
+
+object Slf4jLogHandler {
+
+  def apply(clazz: Class[_]): LogHandler =
+    apply(LoggerFactory.getLogger(clazz))
+
+  def apply(logger: Logger): LogHandler =
+    LogHandler {
+      case Success(s, a, e1, e2) =>
+        logger.debug(s"""Successful Statement Execution:
+                        |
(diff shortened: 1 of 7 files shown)


cocreature (approved):
lgtm, can you please add the change to the log-level to the changelog, thanks!

stefanobaghino-da:
> lgtm, can you please add the change to the log-level to the changelog, thanks!

Addressed in https://github.com/digital-asset/daml/pull/8731/commits/6bfd1bccfbe78b395d5c14c50bd74a5749aa59ce

libs-scala/doobie-slf4j/src/main/scala/com/daml/doobie/logging/Slf4jLogHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def apply(clazz: Class[_]): LogHandler =
+    apply(LoggerFactory.getLogger(clazz))
+
+  def apply(logger: Logger): LogHandler =
+    LogHandler {
+      case Success(s, a, e1, e2) =>
+        logger.debug(s"""Successful Statement Execution:
+                        |
+                        |  ${s.linesIterator.dropWhile(_.trim.isEmpty).mkString("\n  ")}
```
S11001001, line 19: If `|` occurs in any of the interpolated data it will be `stripMargin`ed too; it might be better to not use `stripMargin` here at all.

stefanobaghino-da: Ohhh, this is interesting, thanks.

Failing example.

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

val trololo = "   | trololo"
s"""this
   |  is
   |  an
   |  $trololo
   |  experiment
   |""".stripMargin

// Exiting paste mode, now interpreting.

trololo: String = "   | trololo"
res0: String =
"this
  is
  an
 trololo
  experiment
"
```

PR #8734, opened 2021-02-03
Implement fmt.sh --diff for the Java style check

stefanobaghino-da:
When run with --diff, fmt.sh now only checks for the .java files that have been
changed since the fork point. This makes local git pre-push hooks more quiet
while preserving the safe CI behavior (which doesn't use --diff).

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/fmt.sh
+++ b/fmt.sh
@@ -15,7 +15,7 @@ eval "$(dev-env/bin/dade-assist)"
 is_test=
 scalafmt_args=()
 javafmt_args=(--set-exit-if-changed --replace)
-hlint_diff=false
+diff_mode=false
 dade_copyright_arg=update
 buildifier_target=//:buildifier-fix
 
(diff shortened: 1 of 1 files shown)


stefanobaghino-da:
I tested that the changes behaved as expected on Java files using various combinations of with and without `--test` and `--diff`.

What is an `hlint` violation that I can use to make sure I'm not breaking anything there?

cocreature:
> What is an `hlint` violation that I can use to make sure I'm not breaking anything there?

redundant parentheses iirc

fmt.sh
```diff
@@ -112,24 +125,21 @@ run dade-copyright-headers "$dade_copyright_arg" .
...
-  fi
+if [ "$diff_mode" = "true" ]; then
+  check_diff $merge_base '\.hs$' hlint -j4
+  check_diff $merge_base '\.java$' javafmt "${javafmt_args[@]:-}"
 else
-  "${HLINT[@]}" --git
+  run hlint -j4 --git
+  java_files=$(find . -name "*.java")
+  if [[ -z "$java_files" ]]; then
+    echo "Unexpected: no Java file in the repository"
```
cocreature, line 135: accepting bets for when we’ll reach that.

cocreature (approved):
lgtm, thanks!

fmt.sh
stefanobaghino-da: Do you have a Daml driven betting platform? 😉

stefanobaghino-da:
> > What is an `hlint` violation that I can use to make sure I'm not breaking anything there?
> 
> redundant parentheses iirc

Tested, works. 🙇🏻

PR #8737, opened 2021-02-03
kvutils: Export writes when using pre-execution. [KVL-822]

SamirTalwar-DA:
This allows the user to specify `KVUTILS_LEDGER_EXPORT` to export writes to a file in pre-execution mode.

To facilitate this, we wrap the `SubmissionAggregator.WriteSetBuilder` in a `LedgerStateWriteOperations` implementation which means that the committer doesn't need to know about them.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA:
I will switch the integrity checker over next. I tried doing it as part of this PR but it's a big change.

miklos-da (approved):
Approving with the assumption that you have verified that you can reproduce the same exports as before for e.g. conformance tests.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/validator/CombinedLedgerStateWriteOperations.scala
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.validator
+
+import com.daml.ledger.participant.state.kvutils.Raw
+
+import scala.concurrent.{ExecutionContext, Future}
+
+final class CombinedLedgerStateWriteOperations[ALogResult, BLogResult, LogResult](
```
fabiotudone-da, line 10: I have a feeling we won't use much this abstraction beyond this specific need, but I am OK with it.

fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/validator/CombinedLedgerStateWriteOperations.scala
SamirTalwar-DA: I agree, but I prefer to generalize when possible because there's fewer things you can get wrong.

SamirTalwar-DA:
I have not completely proved that this works yet, but I'm going to merge anyway and not tell anyone about this until we have tests that show it works (and any fixes required).

PR #8738, opened 2021-02-03
tweak local cache cleanup

garyverhaegen-da:
For various reasons, my attempts at improving the cache cleanup process have been delayed. There are, however, two simple, non-controversial changes I can "backport" without having to wait for consensus on the whole thing:

1. Increase the threshold. At least for the compat jobs, we have seen builds failing after starting with ~32GB free.
2. Kill dangling Bazel processes, which keep some files open and sometimes cause the clean-up process to crash.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks! I think this might also just be good enough for now. The permission denied issue is annoying, the long wait times haven’t been that big of an issue.

PR #8739, opened 2021-02-03
Switch from public keys to certificates for non-repudiation

stefanobaghino-da:
changelog_begin
changelog_end

Un-butchering the original design document

### Pull Request Checklist

trigger the build.


runtime-components/non-repudiation/src/test/scala/com/daml/nonrepudiation/NonRepudiationProxySpec.scala
```diff
@@ -138,21 +143,36 @@ final class NonRepudiationProxySpec
...
   )
 
   object Setup {
 
+    def generateKeyAndCertificate(): (PrivateKey, X509Certificate) = {
+      val generator = new CertAndKeyGen(AlgorithmString.RSA, AlgorithmString.SHA256withRSA)
+      generator.generate(2048)
+      val privateKey = generator.getPrivateKey
+      val certificate = generator.getSelfCertificate(
+        new X500Name("CN=Non-Repudiation Test,O=Digital Asset,L=Zurich,C=CH"),
```
cocreature, line 161: Hm, I wonder if the fact that we use CNs but don’t validate them (or really subject alt names because nobody uses CNs these days) will cause trouble. I don’t really want to get into the business of doing that so maybe just needs to be very clear in the docs that we do not validate this.

cocreature (approved):
Nice, thanks!

runtime-components/non-repudiation/src/test/scala/com/daml/nonrepudiation/NonRepudiationProxySpec.scala
stefanobaghino-da: Ultimately should we not simply trust the certificate authority (which is indeed not yet in, but will ultimately be added)?

cocreature: The certificate authority validates that it’s a valid certificate. Subject alt names validate that it’s a certificate for the correct IP/hostname. You can’t really replace one by the other afaict.

stefanobaghino-da: I see, thanks for the clarification. Should we add this validation at all? If yes, I can create a ticket for this to make sure we don't lose this.

cocreature: I’d probably not add it so the task here is only to document that we do not validate this.

stefanobaghino-da: Tracked by https://github.com/digital-asset/daml/issues/8746

PR #8742, opened 2021-02-04
kvutils: Add the record time to the pre-execution log entries. [KVL-822]

SamirTalwar-DA:
This _feels_ like an omission, but I could be totally wrong.

This makes it easier to support pre-execution in the kvutils integrity checker, because we can use the log entry timestamp for the second run.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
@@ -25,6 +25,7 @@ private[kvutils] case class CommitContext(
     private val inputs: DamlStateMap,
     recordTime: Option[Timestamp],
     participantId: ParticipantId,
+    preExecute: Boolean,
 ) {
   private[this] val logger = LoggerFactory.getLogger(this.getClass)
 
@@ -43,8 +44,6 @@ private[kvutils] case class CommitContext(
   // Rejection log entry used for generating an out-of-time-bounds log entry in case of
   // pre-execution.
   var outOfTimeBoundsLogEntry: Option[DamlLogEntry] = None
-
-  def preExecute: Boolean = recordTime.isEmpty
 
   /** Retrieve value from output state, or if not found, from input state.
     * Throws an exception if the key is not found in either.
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/Committer.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/Committer.scala
@@ -14,8 +14,8 @@ import com.daml.ledger.participant.state.kvutils.DamlKvutils.{
   DamlSubmission,
 }
 import com.daml.ledger.participant.state.kvutils.KeyValueCommitting.PreExecutionResult
-import com.daml.ledger.participant.state.kvutils.committer.Committer._
 import com.daml.ledger.participant.state.kvutils._
+import com.daml.ledger.participant.state.kvutils.committer.Committer._
 import com.daml.ledger.participant.state.v1.{Configuration, ParticipantId}
 import com.daml.lf.data.Time
 import com.daml.lf.data.Time.Timestamp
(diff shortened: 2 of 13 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -185,52 +185,57 @@ private[kvutils] class TransactionCommitter(
...
+        .setTransactionRejectionEntry(
+          buildRejectionLogEntry(
+            transactionEntry,
+            RejectionReason.InvalidLedgerTime(
+              s"Record time is outside of valid range [$minimumRecordTime, $maximumRecordTime]"
+            ),
+          )
+        )
+        .build
+      commitContext.outOfTimeBoundsLogEntry = Some(outOfTimeBoundsLogEntry)
```
fabiotudone-da, line 214: This is work you only need to do for pre-execution and it feels wasteful to do it anyway.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -185,52 +185,57 @@ private[kvutils] class TransactionCommitter(
...
-                  buildRejectionLogEntry(
-                    transactionEntry,
-                    RejectionReason.InvalidLedgerTime(reason),
+      if (commitContext.preExecute) {
+        // Pre-execution: propagate the time bounds and defer the checks to post-execution.
+        StepContinue(transactionEntry)
+      } else {
+        commitContext.recordTime match {
+          case None =>
+            StepContinue(transactionEntry)
```
fabiotudone-da, line 222: What does this case correspond to, i.e. batching without a record time?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/Committer.scala
```diff
@@ -81,18 +81,19 @@ private[committer] trait Committer[PartialResult] extends SubmissionExecutor {
...
   ): (DamlLogEntry, Map[DamlStateKey, DamlStateValue]) =
     runTimer.time { () =>
-      val commitContext = CommitContext(inputState, recordTime, participantId)
+      val commitContext = CommitContext(inputState, recordTime, participantId, preExecute = false)
       val logEntry = runSteps(commitContext, submission)
       logEntry -> commitContext.getOutputs.toMap
     }
 
   def runWithPreExecution(
+      recordTime: Option[Time.Timestamp],
```
fabiotudone-da, line 90: What does this represent? Pre-execution has no record time because the record time is established at post-execution, after the deferred time bounds checks and conflict detection.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/CommitContext.scala
```diff
@@ -44,8 +45,6 @@ private[kvutils] case class CommitContext(
   // pre-execution.
   var outOfTimeBoundsLogEntry: Option[DamlLogEntry] = None
 
-  def preExecute: Boolean = recordTime.isEmpty
```
fabiotudone-da, line 47: Pre-execution and the absence of a record time (because it is only determined at post-execution) are equivalent. Why are you separating them?

fabiotudone-da (commented):
I am not sure what you are trying to achieve but I don't see where you are adding the record time to the log entries produced by pre-execution; as a matter of fact you shouldn't and you can't, because the record time is determined only at post-execution. Can you clarify?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/Committer.scala
SamirTalwar-DA: Your other comments are useful but I think this one gets to the heart of why I said "I could be totally wrong".

This makes sense; I can understand why we should be establishing the record time during post-execution, not pre-execution. I added the logic here because it's the last point at which we talk about `DamlLogEntry`, rather than `Raw.Value`. I can see now why that's unhelpful.

I think we could _probably_ do this correctly in post-execution by deserializing the log entry, adding the record time, and then re-serializing it, but that's a lot of work for something that's only useful in testing. So I'm going to propose I close this PR and go with the slightly less efficient pre-executing integrity checker instead.

Thanks for explaining. :smile:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
SamirTalwar-DA: You're right, but we're planning on deleting the other code path at some point soon anyway, so I didn't worry about it too much.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/Committer.scala
fabiotudone-da: My pleasure! Yeah I wouldn't add compute for the sake of testing, especially not at post-execution that needs to be as fast as possible. That's actually the whole point of pre-execution, i.e. pre-validating in parallel (optimistically) so that writing to the ledger can not only be safe but also significantly faster.

PR #8744, opened 2021-02-04
remove dependency-check

garyverhaegen-da:
It relies on a Bintray URL, which is likely to break soon as Bintray sunsets.

It's also not been referenced by anything in this repo since at least open-sourcing, so the value it provides seems low. There's no documentation of why it's here and what its planned usage was, but based on the [project page](https://github.com/jeremylong/DependencyCheck) I imagine it's been supplanted by BlackDuck now.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
:fire:

SamirTalwar-DA (approved):

PR #8745, opened 2021-02-04
Script to build docker locally

remyhaemmerle-da:
based on code by Moritz

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/dev-env/bin/daml-docker-head
+++ b/dev-env/bin/daml-docker-head
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -eu
+
+version=0.0.0
+sdk_tarball=sdk-release-tarball.tar.gz
+
+# Make sure they are in the right repo
+readonly repo_root="$(bazel info workspace)"
+if [[ "${repo_root}/dev-env/bin/daml-docker-head" != "${BASH_SOURCE[0]}" ]]; then
+  echo "Repo root does not match the script being run."
+  echo "Make sure you run this script from a directory within the 'daml' repository "
+  exit 1
+fi
+
+# Make sure they have docker
+if ! which docker > /dev/null ; then
+  echo "docker does not seems to be installed."
+  echo "See https://docs.docker.com/engine/install/ for more detail about docker installation."
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
I'm a bit concerned about long-term maintenance on this. Would you mind elaborating on the use-case a bit?

cocreature:
> Would you mind elaborating on the use-case a bit?

This came up when we tried to test a project relying on those images against SDK HEAD. It is definitely a rather hacky solution, however our docker images don’t change too open and this is only for testing and debugging so if it does get out of sync, we can just fix it next time someone tries to use it. That said, if you have a better idea for making it work, I’m all ears.

dev-env/bin/daml-docker-head
```diff
@@ -0,0 +1,42 @@
...
+RUN tar xf /tmp/${sdk_tarball} -C /tmp && \
+  /tmp/sdk-${version}/install.sh && \
+  rm -rf /tmp/${sdk_tarball} /tmp/sdk-${version}
+ENV PATH="/home/daml/.daml/bin:${PATH}"
+WORKDIR /home/daml
+EOF
+
+docker build ${tmpdir}
+
+rm -rf ${tmpdir}
```
cocreature, line 42: please throw this in a `trap` to make sure we cleanup if building the docker image fails.

dev-env/bin/daml-docker-head
```diff
@@ -0,0 +1,42 @@
...
+  echo "docker does not seems to be installed."
+  echo "See https://docs.docker.com/engine/install/ for more detail about docker installation."
+  exit 1
+fi
+
+tmpdir=$(mktemp -d -t docker-daml-XXXXXXXXXX)
+
+bazel build //release:sdk-release-tarball
+
+cp ${repo_root}/bazel-bin/release/${sdk_tarball} ${tmpdir}/
```
cocreature, line 26: perhaps worth adding a comment that we need this because docker really doesn’t like `COPY`ing from files outside of the build directory and it also doesn’t like symlinks.

cocreature (approved):
lgtm, thanks!

PR #8747, opened 2021-02-04
Desperate attempt at reducing the flakiness in the hs bindings tests

cocreature:
At the very least this should produce better error messages and at
least on my machine it does reduce flakiness significantly by bumping
the timeout and increasing Bazel’s resource allocation (and thereby
reducing parallel runs).

Unfortunately, a proper fix here is much more involved than I have
time for right now.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/hs/bindings/BUILD.bazel
+++ b/language-support/hs/bindings/BUILD.bazel
@@ -53,7 +53,13 @@ da_haskell_test(
         ":for-upload.dar",
         "//ledger/sandbox:sandbox-binary",
     ],
-    # The tests are flaky probably due to issues in gRPC-haskell.
+    # The tests throw flaky timeout errors. It looks like this comes
+    # from a fundamental issue in the Haskell bindings: They eagerly pull
+    # from streams and those stream receives have a timeout. If we create the
+    # stream too early and do a bunch of other stuff in between, this will then timeout.
+    # We could increase the timeout to an absurd value but that also seems silly
+    # and affects other operations as well so for now, we still keep it flaky
+    # and try not to cry.
     flaky = True,
     hackage_deps = [
         "aeson",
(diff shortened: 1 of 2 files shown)


language-support/hs/bindings/BUILD.bazel
```diff
@@ -53,7 +53,13 @@ da_haskell_test(
...
         "//ledger/sandbox:sandbox-binary",
     ],
-    # The tests are flaky probably due to issues in gRPC-haskell.
+    # The tests throw flaky timeout errors. It looks like this comes
+    # from a fundamental issue in the Haskell bindings: They eagerly pull
+    # from streams and those stream receives have a timeout. If we create the
+    # stream too early and do a bunch of other stuff in between, this will then timeout.
+    # We could increase the timeout to an absurd value but that also seems silly
+    # and affects other operations as well so for now, we still keep it flaky
+    # and try not to cry.
```
stefanobaghino-da, line 62: Thanks for the extended explanation. 🙇🏻

language-support/hs/bindings/BUILD.bazel
```diff
@@ -80,6 +86,9 @@ da_haskell_test(
     ],
     main_function = "DA.Ledger.Tests.main",
     src_strip_prefix = "test",
+    # We spin up Sandbox as a separate process, so
+    # try not to overload the machine.
+    tags = ["cpu:2"],
```
stefanobaghino-da, line 91: Should these tests run with `"exclusive"`?

stefanobaghino-da (approved):
🙇🏻

language-support/hs/bindings/BUILD.bazel
cocreature: Good question, I don’t know tbh. Exclusive makes everything slower since nothing else can run in parallel. Of course if things run in parallel but they fail then things might get slower. Locally, this reduces most flakes for me so I think this is probably faster than `exclusive` but we can revisit if it still is super flaky.

PR #7934, opened 2020-11-10
Add a prototype for DAML Script dumps

cocreature:
This is still fairly rough unfortunately but it does at least have
some tests and it doesn’t interact with anything else, so hopefully we
can land this and then parallelize the work from there on.

changelog_begin
changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8750, opened 2021-02-04
Expose transaction trees in DAML Script

### LINKED ISSUE #7847: Access to transaction trees in DAML Script
With multi-party submissions, we should in principle be able to query a transaction stream and turn that into a DAML Script (a “DAML Script dump” of a ledger)

However, that runs into issues around contract ids: Contract ids are not fixed so we clearly cannot simply copy over the ones that we found in the original transaction stream. Instead, we have to keep track of where they were created and use that in the DAML Script. This works fine for contracts that arise from bare creates or that are the result value of an exercise which we can bind as part of the `submit` and then use later. However, if the contract is created as part of an exercise and not returned via the return value, we have no way to access it.

If we provide access to the full transaction tree produced by a `submit`, we can access all created contracts and this might also be useful for implementing tests.

cocreature:
This is spun off from #7934 for ease of review.

For now, all of this is hidden (and for extra caution also marked as
early access). I expect we’ll tweak the API in a bunch of ways before
marking it stable but this at least is sufficient for the prototype.

At this point, we cannot support this over the JSON API. Not too
worried about that at least for now.

fixes #7847

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/src/DA/Test/ScriptService.hs
+++ b/compiler/damlc/tests/src/DA/Test/ScriptService.hs
@@ -659,7 +659,41 @@ main =
                 expectScriptSuccess rs (vr "testSucceed") $ \r ->
                   matchRegex r "Active contracts:  #0:0, #1:0"
                 expectScriptFailure rs (vr "testFail") $ \r ->
-                  matchRegex r "missing authorization from 'p1'"
+                  matchRegex r "missing authorization from 'p1'",
+              testCase "submitTree" $ do
+                rs <-
+                  runScripts
+                    scriptService
+                    [ "module Test where"
+                    , "import DA.Assert"
+                    , "import DA.Foldable"
+                    , "import Daml.Script"
+                    , "template T"
+                    , "  with"
+                    , "    p : Party"
+                    , "    v : Int"
+                    , "  where"
+                    , "    signatory p"
+                    , "    nonconsuming choice CreateN : ()"
+                    , "      with n : Int"
+                    , "      controller p"
+                    , "      do forA_ [ 1 .. n ] $ \\i -> create (T p i)"
+                    , "test = do"
+                    , "  p <- allocateParty \"p\""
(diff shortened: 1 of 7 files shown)


compiler/damlc/tests/src/DA/Test/ScriptService.hs
```diff
@@ -659,7 +659,41 @@ main =
                 expectScriptSuccess rs (vr "testSucceed") $ \r ->
                   matchRegex r "Active contracts:  #0:0, #1:0"
                 expectScriptFailure rs (vr "testFail") $ \r ->
-                  matchRegex r "missing authorization from 'p1'"
+                  matchRegex r "missing authorization from 'p1'",
+              testCase "submitTree" $ do
```
cocreature, line 663: I finally got annoyed enough to open an issue to improve our testing infrastructure here https://github.com/digital-asset/daml/issues/8753

daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/LedgerInteraction.scala
```diff
@@ -528,96 +580,100 @@ class IdeClient(val compiledPackages: CompiledPackages) extends ScriptLedgerClie
...
-                      exercise.templateId,
-                      exercise.choiceId,
-                      exercise.exerciseResult.get,
-                    )
-                  case n =>
-                    // Root nodes can only be creates and exercises.
-                    throw new RuntimeException(s"Unexpected node: $n")
+  )(implicit ec: ExecutionContext): Future[
+    Either[StatusRuntimeException, (RichTransaction, Seq[ScriptLedgerClient.CommandResult])]
+  ] =
```
cocreature, line 585: I recommend reviewing this with whitespace changes hidden. Unfortunately, scalafmt is pretty dumb here.

aherrmann-da (approved):
Looks good!

PR #8751, opened 2021-02-04
ghcide: cached goto definitions

robin-da:
We update ghcide, which enables cached goto definitions that keep
working when the document doesn't compile anymore. This also adds a test
to lsp-tests for this feature.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-doc/src/DA/Daml/Doc/Extract.hs
+++ b/compiler/damlc/daml-doc/src/DA/Daml/Doc/Extract.hs
@@ -54,6 +54,8 @@ import qualified Data.Set as Set
 import qualified Data.Text as T
 import Data.Either
 
+
+
 -- | Extract documentation in a dependency graph of modules.
 extractDocs ::
     ExtractOptions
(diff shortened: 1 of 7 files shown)


compiler/damlc/daml-doc/src/DA/Daml/Doc/Extract.hs
```diff
@@ -192,9 +194,12 @@ haddockParse diagsLogger opts f = MaybeT $ do
           Service.discardInternalModules (optUnitId opts) f
       liftIO $ Service.setFilesOfInterest service (HashSet.fromList nonInternal)
       Service.runActionSync service $ runMaybeT $ do
-          deps <- Service.usesE Service.GetDependencies f
-          Service.usesE Service.TypeCheck $ nubOrd $ f ++ concatMap Service.transitiveModuleDeps deps
+          deps <- usesE' Service.GetDependencies f
+          usesE' Service.TypeCheck $ nubOrd $ f ++ (concatMap Service.transitiveModuleDeps $ deps)
```
cocreature, line 198:
```suggestion
          usesE' Service.TypeCheck $ nubOrd $ f ++ concatMap Service.transitiveModuleDeps deps
```
I don’t really care but hlint probably yells at you for it.

compiler/lsp-tests/src/Main.hs
```diff
@@ -398,6 +398,28 @@ requestTests run _runScenarios = testGroup "requests"
           -- answerFromTest
           locs <- getDefinitions main' (Position 2 8)
           liftIO $ locs @?= [Location (test ^. uri) (Range (Position 1 0) (Position 1 14))]
+
+          -- introduce syntax error
+          changeDoc main' [TextDocumentContentChangeEvent (Just (Range (Position 6 6) (Position 6 6))) Nothing "+"]
```
cocreature, line 403: It would be great to insert a newline in addition to this to make sure position remapping works correctly.

cocreature (approved):
Thanks! One minor comment on the test.

compiler/lsp-tests/src/Main.hs
robin-da: ah good idea, I'll do that.

compiler/damlc/daml-doc/src/DA/Daml/Doc/Extract.hs
```diff
@@ -54,6 +54,8 @@ import qualified Data.Set as Set
 import qualified Data.Text as T
 import Data.Either
 
+
```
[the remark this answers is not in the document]
robin-da: probably reformatting messup, I'll delete them.

PR #8752, opened 2021-02-04
ci/prs: add artifactory credentials

garyverhaegen-da:
Changes already made on both Artifactory and Azure; this is mainly to test.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -197,7 +197,7 @@ jobs:
         push daml-script-runner $(script-runner)
         push daml-script-runner $(script-runner).asc
       env:
-        AUTH: $(ARTIFACTORY_USERPASS)
+        AUTH: $(ARTIFACTORY_USERNAME):$(ARTIFACTORY_PASSWORD)
       condition: not(eq(variables['skip-github'], 'TRUE'))
     - template: ci/bash-lib.yml
       parameters:


azure-pipelines.yml
```diff
@@ -19,6 +19,21 @@ pr: none
 jobs:
 - template: ci/build.yml
 
+- job: test_artifactory_credentials
```
garyverhaegen-da, line 22: There isn't really a way to test the main branch credentials without merging this, but at least this will let us test it without having to make a release.

cocreature (approved):
Thanks!

azure-pipelines.yml
```diff
@@ -197,7 +215,7 @@ jobs:
         push daml-script-runner $(script-runner)
         push daml-script-runner $(script-runner).asc
       env:
-        AUTH: $(ARTIFACTORY_USERPASS)
+        AUTH: $(ARTIFACTORY_USERNAME):$(ARTIFACTORY_PASSWORD)
```
cocreature, line 218: we probably want to keep this and delete the rest now that we have an actual oracle job that uses the credentials.

garyverhaegen-da: 👍

PR #8757, opened 2021-02-04
ledger-on-memory: Only support pre-execution. [KVL-821]

SamirTalwar-DA:
Turns out we don't need batching after all.

I switched out the `materializer.executionContext` for a dedicated committer execution context (which _is_ the materializer's in tests) because our standard test materializer only has one thread available, which is not enough for pre-execution.

### Pull Request Checklist

trigger the build.


ledger/recovering-indexer-integration-tests/src/test/suite/scala/com/digitalasset/platform/indexer/RecoveringIndexerIntegrationSpec.scala
```diff
@@ -256,13 +254,22 @@ object RecoveringIndexerIntegrationSpec {
...
-        ledgerId,
-        BatchingLedgerWriterConfig.reasonableDefault,
-        participantId,
-        metrics = metrics,
-        engine = Engine.DevEngine(),
-      ).map(readerWriter => new KeyValueParticipantState(readerWriter, readerWriter, metrics))
+      for {
+        dispatcher <- memory.dispatcherOwner
+        committerExecutionContext <- ResourceOwner
+          .forExecutorService(() => Executors.newCachedThreadPool())
```
fabiotudone-da, line 260: Are we sure we don't want to limit the threads?

fabiotudone-da:
Nice negative diff! 😃

ledger/recovering-indexer-integration-tests/src/test/suite/scala/com/digitalasset/platform/indexer/RecoveringIndexerIntegrationSpec.scala
SamirTalwar-DA: We could, but this is a test case, so it will be thrown away in about 5 seconds, so who cares?

fabiotudone-da: Oh sorry, you're right. 👍🏼

fabiotudone-da (approved):

PR #8760, opened 2021-02-04
tag the docker image build locally

remyhaemmerle-da:
This was forgotten in #8745

also add a missing '\' in the bash script.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/dev-env/bin/daml-docker-head
+++ b/dev-env/bin/daml-docker-head
@@ -37,7 +37,7 @@ COPY --chown=daml:daml ${sdk_tarball} /tmp/
 RUN tar xf /tmp/${sdk_tarball} -C /tmp && \
   /tmp/sdk-${version}/install.sh && \
   rm -rf /tmp/${sdk_tarball} /tmp/sdk-${version}
-ENV PATH="/home/daml/.daml/bin:${PATH}"
+ENV PATH="/home/daml/.daml/bin:\${PATH}"
 WORKDIR /home/daml
 EOF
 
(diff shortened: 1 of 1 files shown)


dev-env/bin/daml-docker-head
```diff
@@ -41,4 +41,4 @@ ENV PATH="/home/daml/.daml/bin:${PATH}"
 WORKDIR /home/daml
 EOF
 
-docker build ${tmpdir}
+docker build -t "digitalasset/daml-sdk:${version}" ${tmpdir}
```
garyverhaegen-da, line 44: Usually the first segment (before `/`) in a Docker tag is the repo, so local tags are (by convention) not nested. Perhaps `daml-sdk-head:${version}` instead?

Also, if we're considering names, it may be worth not tagging them with 0, but instead something like `$(git show -s --format=%cd-%h --date=format:%Y%m%d-%H%M%S)`? (Only talking about changing the image name here; the build inside Docker should definitely use 0.0.0 otherwise it will take ages.)

garyverhaegen-da (approved):

dev-env/bin/daml-docker-head
cocreature: Not sure it makes sense to keep the docker images more persistent than SDK head itself. I think just overwriting is fine and arguably convenient since you get a fixed hash every time.

cocreature (approved):
Thanks! I agree with @garyverhaegen-da that removing the slash might be less confusing.

cocreature:
for reference, the current docker image name makes sense since it matches the one we push to dockerhub making it as easy as possible to swap out a build relying on a released version against one relying on HEAD so let’s keep it.

PR #8763, opened 2021-02-05
Add a CI job for testing against Oracle DB

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -199,6 +199,65 @@ jobs:
         trigger_sha: '$(trigger_sha)'
     - template: report-end.yml
 
+- job: Linux_oracle
+  dependsOn:
+    - da_ghc_lib
+  timeoutInMinutes: 360
+  pool:
+    name: 'ubuntu_20_04'
+    demands: assignment -equals default
+  steps:
+    - template: report-start.yml
+    - checkout: self
+    - bash: ci/dev-env-install.sh
+      displayName: 'Build/Install the Developer Environment'
+    - template: clean-up.yml
+    - bash: |
+        source dev-env/lib/ensure-nix
+        ci/dev-env-push.py
+      displayName: 'Push Developer Environment build results'
+      condition: eq(variables['System.PullRequest.IsFork'], 'False')
+      env:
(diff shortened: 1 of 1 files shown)


S11001001 (approved):

ci/build.yml
```diff
@@ -199,6 +199,61 @@ jobs:
...
+        # to upload to the bazel cache
+        GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+    - bash: |
+        set -euo pipefail
+        eval "$(./dev-env/bin/dade-assist)"
+        export ORACLE_PORT=1521
+        export ORACLE_PWD=hunter2
+        docker login --username $ARTIFACTORY_USERNAME --password $ARTIFACTORY_PASSWORD digitalasset-daml-on-oracle-docker.jfrog.io
+        docker pull digitalasset-daml-on-oracle-docker.jfrog.io/oracle-xe-preloaded:2021-02-03-5a88b29
+        docker stop oracle || true
```
garyverhaegen-da, line 237: I'm a bit confused here. What is this for? I don't see any oracle container being started before this line. Is `docker pull` starting anything?

On the other hand, I would have expected something like `trap "docker rm -f oracle" EXIT` somewhere.

garyverhaegen-da (approved):

ci/build.yml
cocreature: duh, I had the trap before but accidentally added it. I kept the `docker stop` here as well since I don’t entirely trust the trap triggering on cancelled builds and things like that.

garyverhaegen-da: Is `stop` actually enough here? It's been a while since I actually used Docker, but IIRC if you just `stop` a container but don't remove it, the name is still not available for reuse.

cocreature: there is a `--rm` on the `docker run` which means the stopped container will be autodeleted. But yeah I’ve switched to `rm -f` anyway, I don’t entirely trust that and I got a few errors on CI testing this.

PR #8764, opened 2021-02-05
try to improve macOS install instructions

garyverhaegen-da:
Based on [feedback from Khuram] on the forum.

[feedback from Khuram]: https://discuss.daml.com/t/my-feedback-on-daml-in-general-and-documentation-as-a-non-coder/2043/6?u=gary_verhaegen

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_. For best experience, make sure the ``code`` command is available (it is enabled by default on both Linux and Windows, though in both case only for new terminals opened _after_ the installation; for macOS, follow additional instructions `here <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_).
+1. `Visual Studio Code <https://code.visualstudio.com/download>`_. For best experience, make sure the ``code`` command is available (it is enabled by default on both Linux and Windows, though in both cases only for new terminals opened _after_ the installation; for macOS, follow additional instructions `here <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_).
 2. JDK 8 or greater. If you don't already have a JDK installed, try `AdoptOpenJDK <https://adoptopenjdk.net>`_.
 
    As part of the installation process you might need to set up the ``JAVA_HOME`` variable. You can find here the instructions on how to do it on :doc:`Windows,macOS, and Linux <path-variables>`.
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_. For best experience, make sure the ``code`` command is available (it is enabled by default on both Linux and Windows, though in both cases only for new terminals opened _after_ the installation; for macOS, follow additional instructions `here <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_).
+1. `Visual Studio Code <https://code.visualstudio.com/download>`_. You do not need the ``code`` command-line tool to use Daml, but if you want to have it working on macOS, you can follow `these instructions <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_. (It is automatically added on Windows and Linux.)
 2. JDK 8 or greater. If you don't already have a JDK installed, try `AdoptOpenJDK <https://adoptopenjdk.net>`_.
 
    As part of the installation process you might need to set up the ``JAVA_HOME`` variable. You can find here the instructions on how to do it on :doc:`Windows,macOS, and Linux <path-variables>`.
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_. You do not need the ``code`` command-line tool to use Daml, but if you want to have it working on macOS, you can follow `these instructions <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_. (It is automatically added on Windows and Linux.)
+1. `Visual Studio Code <https://code.visualstudio.com/download>`_.
 2. JDK 8 or greater. If you don't already have a JDK installed, try `AdoptOpenJDK <https://adoptopenjdk.net>`_.
 
    As part of the installation process you might need to set up the ``JAVA_HOME`` variable. You can find here the instructions on how to do it on :doc:`Windows,macOS, and Linux <path-variables>`.
--- a/docs/source/getting-started/installation.rst
+++ b/docs/source/getting-started/installation.rst
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_.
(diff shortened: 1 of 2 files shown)


docs/source/getting-started/path-variables.rst
```diff
@@ -24,48 +24,31 @@ If you have downloaded and installed the SDK using our `Windows installer <https
...
-Verifying the changes
-=====================
-
-In order for the changes to take effect you will need to restart your computer, or, if you're using
-the macOS Terminal app, you only need to quit the Terminal app (Command+Q in the Terminal window) and
-reopen it. Afterward, please follow the instructions below to verify that everything was set
-up correctly.
-
-Please verify the JAVA_HOME variable by running::
+This will update the configuration for future, newly opened terminals, but will not affect any exsting one. To test the configuration of ``JAVA_HOME``, open a new terminal and run::
```
nemanja-da, line 35: Read the forum thread on the installation instructions. Found this a bit more confusing as it's mixing instructions for both `zsh` and `bash`. Would maybe make it clearer at the beginning that current instructions are for `zsh` 

garyverhaegen-da: I would expect people who know the difference between Bash and Zsh to not need this page at all. Under this assumption, I don't see a point in mentioning either; following these instructions will work for both without the user having to know about either, nor which one they're using.

nemanja-da: > I would expect people who know the difference between Bash and Zsh to not need this page at all.

Think this is a pretty big assumption. I don't know the exact diff, just that there is one and that I had to go through different processes when setting up the `PATH` and `JAVA_HOME` variables. 

@moritzkiefer-da also mentioned that there are a number of shells out there. Being explicit about our instructions does help.

garyverhaegen-da: Again, I very much expect that people who have installed a non-default shell know what they are doing enough to:
1. Recognize that all we're doing here is set the values of two env vars, and
2. Know how to set up env vars in their shell.
In my view these instructions are strictly for people who don't know how a shell works.

I may totally be wrong about this, which is what the review process is for. 🙂

nemanja-da: Being explicit about 1) the shell and 2) what they're setting up 3) helps the user look for answers in case they run into issues. Giving them just the commands leaves them dry.

1. We could be removing the bit on `bash`, as this might have caused the confusion. Our instructions are for `zsh` and we mention `bash` as an example. 

2. The command that helped krmmalik set up the `JAVA_PATH` is slightly different form the one that we currently have. He used 

`echo "export JAVA_HOME=`/usr/libexec/java_home`" >> ~/.profile    
. ~/.profile` 

while our instructions have 
`echo 'export JAVA_HOME="$(/usr/libexec/java_home)"' >> ~/.zprofile` 

We might wanna change to that one (not a `shell` expert I'm afraid)

SamirTalwar-DA: `$(...)` is the modern way of doing backticks. They're basically equivalent but `$(...)` can be nested, so most people favor it. It works in `bash` (and `zsh`, `ash`, etc.) but not `sh`.

docs/source/getting-started/installation.rst
```diff
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_.
+1. `Visual Studio Code <https://code.visualstudio.com/download>`_. For best experience, make sure the ``code`` command is available (it is enabled by default on both Linux and Windows, though in both case only for new terminals opened _after_ the installation; for macOS, follow additional instructions `here <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_).
```
robin-da, line 14: typo: `though in both cases`

docs/source/getting-started/path-variables.rst
robin-da: also not sure if adding shell profiles for possibly non-existing shells is a good thing to do.

garyverhaegen-da: > also not sure if adding shell profiles for possibly non-existing shells is a good thing to do.

Adding shell profiles for unused shells is harmless, they are just text files that will never be read.

docs/source/getting-started/installation.rst
garyverhaegen-da: Thanks, fixed.

docs/source/getting-started/path-variables.rst
garyverhaegen-da: @nemanja-da I've added a step to figure out which shell is running, and (hopefully) clarified which commands are for which shell.

docs/source/getting-started/installation.rst
```diff
@@ -11,7 +11,7 @@ The Daml Connect SDK currently runs on Windows, macOS and Linux.
 
 You need to install:
 
-1. `Visual Studio Code <https://code.visualstudio.com/download>`_.
+1. `Visual Studio Code <https://code.visualstudio.com/download>`_. For best experience, make sure the ``code`` command is available (it is enabled by default on both Linux and Windows, though in both cases only for new terminals opened _after_ the installation; for macOS, follow additional instructions `here <https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line>`_).
```
nemanja-da, line 14: Not trying to be nit-picky really, but what's `best experience` within this context? Is this a must or nice_to_have? What's getting improved here?

docs/source/getting-started/path-variables.rst
nemanja-da: @garyverhaegen-da Love the `PATH` variable changes! Left a comment for the installation instructions.

docs/source/getting-started/installation.rst
cocreature: `daml studio` deliberately does not require `code` being in PATH and that’s what we tell people to use. So I’d just omit this.

garyverhaegen-da: I'm a bit confused as to why Khuram thought he had to set that up then. 🤔

I'll update the wording to mention it's not needed.

garyverhaegen-da: Or would you prefer I remove it completely?

garyverhaegen-da:
@nemanja-da I've removed the VSCode install changes; is there anything else blocking this PR?

docs/source/getting-started/path-variables.rst
```diff
@@ -24,48 +24,48 @@ If you have downloaded and installed the SDK using our `Windows installer <https
...
 
-        echo 'export JAVA_HOME="$(/usr/libexec/java_home)"' >> ~/.zprofile
+        echo $SHELL - $0
 
-Setting the PATH variable
-=========================
-The installer will ask you and set the ``PATH`` variable for you. If you want to set your ``PATH`` variable
-manually instead, run the following command in your terminal::
+This should return either ``/bin/bash - bash``, in which case you are running
+Bash, or ``/bin/zsh - -zsh``, in which case you are running zsh. We provide
```
sofiafaro-da, line 33: IME this is slightly inaccurate. I get `/bin/bash - -bash` if Terminal is in bash mode, and `/bin/bash - bash` if I start up `bash` manually. And `/bin/zsh - zsh` if I start up `zsh` manually. Basically the presence of the extra dash (`-`) seems to depend on if you're running your default shell. So my suggestion is to put the extra dash in there in both cases:  

```suggestion
This should return either ``/bin/bash - -bash``, in which case you are running
Bash, or ``/bin/zsh - -zsh``, in which case you are running zsh. We provide
```

nemanja-da (approved):
Thnx a bunch @garyverhaegen-da 😃

docs/source/getting-started/path-variables.rst
garyverhaegen-da: I did not realize this was so fragile, so instead I've just removed `$0`. Do you see any issue with that approach? Unless you're doing something special already (in which case these instructions are probably not for you), `$SHELL` should be correct, I think.

sofiafaro-da: I think that should work! Thanks :)

PR #8765, opened 2021-02-05
ledger-api-test-tool: Increase the dedup time in CommandDeduplicationIT.

SamirTalwar-DA:
The deduplication time was hard-coded to 3 seconds, which can be too short for some ledgers; the command may have expired by the time it gets to the commit stage.

In order to avoid spurious errors, this multiplies the deduplication time by the timeout scale factor, allowing the ledger operator to control the deduplication timeouts in the same way as all test timeouts.

### Changelog

- **[Ledger API Test Tool]** The CommandDeduplicationIT now sets the deduplication timeout on commands using the parameter passed to `--timeout-scale-factor`, allowing it to run on ledger configurations that require more than 3 seconds to commit a submission.

### Pull Request Checklist

trigger the build.


fabiotudone-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import scala.concurrent.duration.{DurationInt, FiniteDuration}
+
+object Defaults {
+
+  val LedgerClockGranularity: FiniteDuration = 1.second
+
+  val TimeoutScaleFactor: Double = 1.0
+
+}
```
stefanobaghino-da, line 14: It probably makes sense to keep these in the same file as `Config` (it's used only there, right?)

stefanobaghino-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
SamirTalwar-DA: _Config_ is not accessible from _Tests_, unfortunately, and it's used there too.

PR #8777, opened 2021-02-08
Pick up user-defined exceptions in LFConversion.

sofiafaro-da:
This uses a "DamlException" tag to annotate which types should
be exported as exceptions. This tag will be added during the
desugaring of exceptions.

The code that scrapes for the HasMessage instance is based on the
corresponding template instance scraping code, but simplified since
we only have to pick up one instance. If we decide to pack more
information into the DefException structure in the future, we can
easily extend this.

This PR also adds a small test to make sure that a user-defined
exception is exported as such.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice, thank you!

PR #8190, opened 2020-12-08
Upgrade nixpkgs

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8780, opened 2021-02-08
update Terraform files to match reality

garyverhaegen-da:
Two changes have happened recently that have invalidated the current Terraform files:

1. The Terraform version has gone through a major, incompatible upgrade (#8190); the required updates for this are reflected in the first commit of this PR.
2. The certificate used to serve [Hoogle](https://hoogle.daml.com) was about to expire, so Edward created a new one and updated the config directly. The second commit in this PR updates the Terraform config to match that new, already-in-prod setting.

Note: This PR applies cleanly, as there are no resulting changes in Terraform's perception of the target state from 1, and the change from 2 has already been applied through other channels.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/main.tf
+++ b/infra/main.tf
@@ -23,11 +23,11 @@ provider "secret" {
 }
 
 provider "template" {
-  version = "2.1.2"
+  version = "~>2.2"
 }
 
 data "google_project" "current" {
-  project_id = "${local.project}"
+  project_id = local.project
 }
 
 locals {
@@ -36,12 +36,10 @@ locals {
     host-group      = "buildpipeline"
     infra-owner     = "daml-language"
     managed         = "true"
-
-    # default the target name to be the name of the folder
-    target = "${basename(path.module)}"
+    target          = "infra"
   }
 
-  machine-labels = "${merge(local.labels, map("env", "production"))}"
+  machine-labels = merge(local.labels, map("env", "production"))
 
   project = "da-dev-gcp-daml-language"
   region  = "us-east4"
--- a/infra/main.tf
+++ b/infra/main.tf
@@ -45,7 +45,7 @@ locals {
   region  = "us-east4"
   zone    = "us-east4-a"
 
-  ssl_certificate_hoogle = "https://www.googleapis.com/compute/v1/projects/da-dev-gcp-daml-language/global/sslCertificates/daml-lang-hoogle-app-service-https-cert"
+  ssl_certificate_hoogle = "https://www.googleapis.com/compute/v1/projects/da-dev-gcp-daml-language/global/sslCertificates/hoogle-google-cert"
 }
 
 resource "secret_resource" "vsts-token" {}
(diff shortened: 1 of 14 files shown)


infra/main.tf
```diff
@@ -36,18 +36,16 @@ locals {
     host-group      = "buildpipeline"
     infra-owner     = "daml-language"
     managed         = "true"
-
-    # default the target name to be the name of the folder
-    target = "${basename(path.module)}"
+    target          = "infra"
```
garyverhaegen-da, line 39: `basename(path.module)` somehow changed from `infra` to `.` in this new version. I could not find the documentation for `path.module`, and the expression `basename(path.module)` still produces `infra` in the `terraform console`, so I gave up.

infra/vsts_agent_ubuntu_20_04.tf
```diff
@@ -46,18 +46,10 @@ resource "google_compute_instance_template" "vsts-agent-ubuntu_20_04" {
...
-    shutdown-script = <<EOS
-#!/usr/bin/env bash
-set -euo pipefail
-cd /home/vsts/agent
-su vsts <<SHUTDOWN_AGENT
-export VSTS_AGENT_INPUT_TOKEN='${secret_resource.vsts-token.value}'
-./config.sh remove --unattended --auth PAT
-SHUTDOWN_AGENT
-    EOS
+    shutdown-script = "#!/usr/bin/env bash\nset -euo pipefail\ncd /home/vsts/agent\nsu vsts <<SHUTDOWN_AGENT\nexport VSTS_AGENT_INPUT_TOKEN='${secret_resource.vsts-token.value}'\n./config.sh remove --unattended --auth PAT\nSHUTDOWN_AGENT\n    "
```
garyverhaegen-da, line 52: Because the existing heredoc was, technically, malformed, there does not seem to exist a new heredoc syntax that produces the same output as before. Specifically, it does not seem to be possible to end a heredoc on spaces (vs. a newline) anymore.

This is probably not the end state we want, but this was not the point of this PR and adding a newline here requires taking down the entire CI cluster, which seemed like a bad idea while people are working.

infra/vsts_agent_windows.tf
```diff
@@ -85,8 +85,8 @@ Invoke-WebRequest https://dl.google.com/cloudagents/windows/StackdriverLogging-v
 iex (New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1')
 
 # Install git, bash
-& choco install git --no-progress --yes 2>&1 | %{ "$_" }
-& choco install windows-sdk-10.1 --no-progress --yes 2>&1 | %{ "$_" }
+& choco install git --no-progress --yes 2>&1 | %%{ "$_" }
+& choco install windows-sdk-10.1 --no-progress --yes 2>&1 | %%{ "$_" }
```
garyverhaegen-da, line 89: The new Terraform version has a new escape sequence `%{}`, so we now need to escape it with `%%{}`.

cocreature (approved):
Thank you! Sorry for making you jump through the upgrade troubles.

PR #8786, opened 2021-02-08
integrity-checker: Reintroduce an interface for comparing write sets. [KVL-822]

SamirTalwar-DA:
Apparently this is used, just outside of this repository. Sorry for breaking it.

This turns `WriteSetComparison` into a trait to be constructed by the `CommitStrategySupport`, allowing implementors to override how we parse, compare, and render write sets.

This code is only used in testing; it does not run in any production system.

### Pull Request Checklist

trigger the build.


fabiotudone-da (approved):
I totally missed that too, thanks for fixing it. I guess this teaches us that we should check against other repositories more often.

PR #8793, opened 2021-02-09
docs: try to improve script --input-file

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/docs/source/daml-script/index.rst
+++ b/docs/source/daml-script/index.rst
@@ -194,28 +194,40 @@ Running a Script
 ----------------
 
 To run our script, we first build it with ``daml build`` and then run
-it by pointing to the DAR, the name of our script, the host and
-port our ledger is running on and the time mode of the ledger.
+it by pointing to the DAR, the name of our script, and the host and
+port our ledger is running on.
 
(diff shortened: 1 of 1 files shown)


docs/source/daml-script/index.rst
```diff
@@ -212,7 +224,7 @@ We can then initialize our ledger passing in the json file via ``--input-file``.
 
 If you open Navigator, you can now see the contracts that have been created.
 
-While we will not use it here, there is also an ``--output-file``
+As mentioned, there is also an ``--output-file``
```
cocreature, line 227: Can we fold this into the paragraph above? It seems a bit weird to mention the option above without explaining it and then mention it somewhere down here.

cocreature (approved):
Thanks!

docs/source/daml-script/index.rst
garyverhaegen-da: I've tried moving this up, which I think flows a little bit better. Thoughts?

cocreature: I like it, thanks!

PR #8803, opened 2021-02-10
Disable MacOS CI jobs

cocreature:
5/6 macos nodes are down and we cannot fix it quickly, so to unblock
everyone, let’s disable those jobs for now.

I deliberately did not remove MacOS from releases. Those really should run on MacOS.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -411,9 +411,11 @@ jobs:
   condition: always()
   dependsOn:
     - Linux
+    - macOS
     - Windows
     - release
     - git_sha
+    - compatibility_macos
     - compatibility_linux
     - compatibility_windows
     - Linux_scala_2_13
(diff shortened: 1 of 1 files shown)


ci/build.yml
```diff
@@ -409,11 +411,9 @@ jobs:
   condition: always()
   dependsOn:
     - Linux
-    - macOS
```
garyverhaegen-da, line 412: I don't think you need to make any changes here, the `collect_build_data` job should be OK with some of its dependencies not running.

garyverhaegen-da: See [prs](https://github.com/digital-asset/daml/blob/main/ci/prs.yml#L76-L78): dependencies need to exist, but not running is OK.

garyverhaegen-da:
Note for monthly standard-change review: this is a change made as part of mitigating [INCIDENT-126](https://digitalasset.atlassian.net/browse/INCIDENT-126).

ci/build.yml
cocreature: ff2a948d333cf8d55080228f215f4b1e32886f89 thanks!

garyverhaegen-da (approved):

ci/build.yml
```diff
@@ -540,10 +542,8 @@ jobs:
         #
         # release only run on releases and is skipped otherwise.
         if [[ "$(Linux.status)" != "Succeeded"
-            || "$(macOS.status)" != "Succeeded"
```
cocreature, line 543: without this, the job fails since the status is skipped.

garyverhaegen-da: 🤦 
Sorry, should have realized that.

Issue #8020, opened 2020-11-20
Exceptions in DAML


sofiafaro-da:
Over-arching ticket for implementing exceptions in DAML (_"one ring to rule them all"_), so we have an issue number to keep track of work to be done. See also [the exception milestone](https://github.com/digital-asset/daml/milestone/54).

TODO:

- [x] update LF specification
  - [x] syntax & semantics of exception handling (implemented in #8068 #8111)
  - [x] update existing primitives & constructs to throw more appropriate exceptions
- [x] update LF protobuf (implemented in #8087)
- [x] change DAML-LF AST
  - [x] on Haskell side (implemented in #8112)
  - [x] on Scala side
- [x] update type checker
  - [x] on Haskell side (#8112, #8137)
  - [x] on Scala side
- [x] update serializability checker
  - [x] on Haskell side (#8112, #8151)
  - [x] on Scala side
- [x] update LF engine
  - [x] implement exception builtins
  - [x] implement exception handling
  - [x] update existing primitives to throw appropriate exceptions
- [x] update standard library
  - [x] add `DAML_EXCEPTIONS` feature (#8114)
  - [x] add `DA.Exception` module (#8023)
  - [x] run through existing standard library functions and see if they should throw more appropriate exceptions (#9783)
- [x] update surface syntax
  - [x] prepare examples of surface syntax & its desugaring
  - [x] implement syntax for declaring exception types
  - [x] implement try-catch block syntax
  - [x] update daml studio language file to highlight exception keywords
  - [x] (not implemented, left as separate issue) implement more powerful ensure statement (issue #3571 but read exception design doc)
- [x] update data-dependencies
  - [x] recognize built-in exception types
  - [x] re-export exceptions
  - [x] add a test that involves raising and catching built-in exceptions across data-dependency boundaries (#9799)
- [x] release a new LF version
  - [x] update the LF version for the `DAML_EXCEPTIONS` feature in `DA.Daml.Lf.Ast.Version`
  - [x] update the LF version for `DA.Internal.Exception` in the "generate stable package" rule
  - [x] update stable package test (`compiler/damlc/tests/src/stable-packages.sh`) and other references to exception stable packages.
- [x] update ledger-api-test-tool with tests that involve catching and throwing exceptions

cocreature:
Conclusion from today’s meeting:

We cannot catch exceptions from old LF versions. A participant might need to validate nodes from that older LF version while it doesn’t understand the new LF version. There are two versions that are possible here:
[...]

PR #8804, opened 2021-02-10
LF: Refactor PartialTransaction Context

remyhaemmerle-da:
This prepares the introduction of roolback node.

This is part of #8020.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -4,7 +4,6 @@
 package com.daml.lf
 package speedy
 
-import com.daml.lf.crypto.Hash
 import com.daml.lf.data.Ref.{ChoiceName, Location, Party, TypeConName}
 import com.daml.lf.data.{BackStack, ImmArray, Ref, Time}
 import com.daml.lf.ledger.Authorize
@@ -15,8 +14,8 @@ import com.daml.lf.transaction.{
   Node,
   NodeId,
   SubmittedTransaction,
-  Transaction => Tx,
   TransactionVersion => TxVersion,
+  Transaction => Tx,
 }
 import com.daml.lf.value.Value
 
@@ -42,7 +41,7 @@ private[lf] object PartialTransaction {
   private[PartialTransaction] final class SeededPartialTransactionRootContext(
       seeds: ImmArray[Option[crypto.Hash]]
   ) extends RootContextInfo {
-    override val childSeed: Int => Hash = { idx =>
+    override val childSeed: Int => crypto.Hash = { idx =>
       seeds.get(idx) match {
         case Some(Some(value)) =>
           value
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -4,6 +4,7 @@
 package com.daml.lf
 package speedy
 
+import com.daml.lf.crypto.Hash
 import com.daml.lf.data.Ref.{ChoiceName, Location, Party, TypeConName}
 import com.daml.lf.data.{BackStack, ImmArray, Ref, Time}
 import com.daml.lf.ledger.Authorize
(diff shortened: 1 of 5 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -27,52 +28,69 @@ private[lf] object PartialTransaction {
...
+      extends RootContextInfo {
+    override private[PartialTransaction] def childrenSeeds(idx: Int) =
+      crypto.Hash.deriveNodeSeed(seed, idx)
+  }
+
+  private[PartialTransaction] final class SeededPartialTransactionRootContext(
+      seeds: ImmArray[Option[crypto.Hash]]
+  ) extends RootContextInfo {
+    override def childrenSeeds(idx: NodeIdx): Hash = {
+      val opt = if (0 <= idx && idx < seeds.length) {
```
cocreature, line 47: should we add something like `.get` to `ImmArray`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -27,52 +28,69 @@ private[lf] object PartialTransaction {
...
+  private[PartialTransaction] object NoneSeededTransactionRootContext extends RootContextInfo {
+    override private[PartialTransaction] def childrenSeeds(idx: NodeIdx): Nothing = {
+      throw new RuntimeException(s"the machine is not configure to create transaction")
+    }
+  }
+
   /** Contexts of the transaction graph builder, which we use to record
     * the sub-transaction structure due to 'exercises' statements.
     */
   final class Context private (
```
cocreature, line 65: this seems stateless. any reason not to make it a case class?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -27,52 +28,69 @@ private[lf] object PartialTransaction {
...
+      val info: ContextInfo,
+      val children: BackStack[NodeId] = BackStack.empty,
   ) {
     def addChild(child: NodeId): Context =
-      new Context(exeContext, children :+ child, childrenSeeds)
-    def nextChildrenSeed: crypto.Hash =
-      childrenSeeds(children.length)
+      new Context(info, children :+ child)
+    // needs to be lazy as info.childrenSeeds may be undefined on children.length
+    lazy val nextChildrenSeed: crypto.Hash =
```
cocreature, line 72: we only expect to call this once right? if so, I think the `def` might be less confusing while having the same performance.

And just for my own understanding. The case where this crashes is during validation when we have a partial transaction context and we find more nodes during reinterpretation than the original (partial) transaction contained?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -111,7 +129,10 @@ private[lf] object PartialTransaction {
       nodeId: NodeId,
       parent: Context,
       byKey: Boolean,
-  )
+  ) extends ContextInfo {
+    override private[PartialTransaction] def childrenSeeds(idx: NodeIdx) =
+      crypto.Hash.deriveNodeSeed(parent.nextChildrenSeed, idx)
```
cocreature, line 134: hm, I would find it more intuitive if we crash when constructing `ExercisesContextInfo` instead of only when we call `childrenSeeds`. Is there an issue with doing that?

cocreature (approved):
Nice, a few bikeshedding comments but looks very sensible.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: > we only expect to call this once right?

No. 

>  if so, I think the `def` might be less confusing while having the same performance.

I change the code so we used it at most one, and then change to `def`.

> And just for my own understanding. The case where this crashes is during validation when we have a partial transaction context and we find more nodes during reinterpretation than the original (partial) transaction contained?

Yes.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: Good point.

I changed the code, so we use `nextChildrenSeed` only once and crash when constructing.

remyhaemmerle-da (commented):
Thanks @cocreature for the very good feedback.

PR #8811, opened 2021-02-10
replace deprecated blackduck property

dasormeter:
Replace use of deprecated property with upgraded 6.8.0 synopsys-detect


cocreature (approved):
Thanks!

PR #8812, opened 2021-02-10
bump check_releases timeout

garyverhaegen-da:
It's become quite flaky at just 4h.

Yes, I do plan to actually fix it, but it may take a while.

CHANGELOG_BEGIN
CHANGELOG_END


leonidr-da (approved):
LGTM.

PR #8821, opened 2021-02-11
Replace ai.x.diff library with com.softwaremill.diffx [DPP-192]

kamil-da:
**Motivation**
At present the Ledger API Test Tool relies on `ai.x.diff` for capturing difference between case classes. This project is no longer maintained and there is no version supporting scala 2.13. We have to switch to using another library so that a full upgrade to scala 2.13 can be performed.

**Changes**
Replaced the `ai.x.diff` library by `diffx` by SoftwareMill.
[https://github.com/softwaremill/diffx](https://github.com/softwaremill/diffx)

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -148,8 +148,10 @@ EOF
                 "//ledger/test-common:semantic-tests%s.scala" % lf_version_suffix,
                 "//libs-scala/grpc-utils",
                 "//libs-scala/timer-utils",
-                "@maven//:ai_x_diff_2_12",
                 "@maven//:com_chuusai_shapeless_2_12",
+                "@maven//:com_propensive_magnolia_2_12",
+                "@maven//:com_propensive_mercator_2_12",
```
kamil-da, line 153: While these libraries are not directly used in our code, they are transitive dependencies of the `diffx`. Without them being explicitly listed Bazel will not put them on the classpath and automatic type class derivation will not work.

bazel-java-deps.bzl
```diff
@@ -173,6 +173,7 @@ def install_java_deps():
             "io.gatling:gatling-http-client:3.3.1",
             "com.fasterxml.jackson.core:jackson-core:2.12.0",
             "com.fasterxml.jackson.core:jackson-databind:2.12.0",
+            "com.softwaremill.diffx:diffx-core_{}:0.4.1".format(scala_major_version),
```
cocreature, line 176: can we kill the ai.x.diff dependency now?

ledger/ledger-api-test-tool/BUILD.bazel
SamirTalwar-DA: Perhaps add a comment to the Bazel file as such?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/DiffExtensions.scala
```diff
@@ -3,51 +3,45 @@
...
+      }
 
-  implicit def diffShowSeq[T](implicit diffShowT: DiffShow[T]): DiffShow[Seq[T]] =
-    new DiffShow[Seq[T]] {
-      override def show(t: Seq[T]): String = t.toString
+    }
+  }
 
-      override def diff(left: Seq[T], right: Seq[T]): Comparison = {
+  implicit def diffSeq[T](implicit diffT: Diff[T]): Diff[Seq[T]] =
```
SamirTalwar-DA, line 20: So the new library doesn't have a native `Diff[Seq[T]]` either?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/DiffExtensions.scala
```diff
@@ -3,51 +3,45 @@
...
+    new Diff[Seq[T]] {
+      override def apply(left: Seq[T], right: Seq[T], toIgnore: List[FieldPath]): DiffResult = {
```
SamirTalwar-DA, line 22: You seem to be ignoring `toIgnore`.

maven_install_2.12.json
```diff
@@ -1,6 +1,6 @@
 {
     "dependency_tree": {
-        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": 1485856285,
+        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": 1031077523,
```
cocreature, line 3: you also need to repin the `2.13` lockfile.

ledger/ledger-api-test-tool/BUILD.bazel
stefanobaghino-da: I think at this point everybody working with Bazel is already familiar with the fact that transitive dependencies must be listed explicitly.

bazel-java-deps.bzl
stefanobaghino-da: kill it with fire

ledger/ledger-api-test-tool/BUILD.bazel
cocreature: the errors here look pretty different from the ones you get usually so I think it’s worth adding a comment even for people familiar with Bazel.

SamirTalwar-DA: That, and so I know exactly what's depending on them. Something like `// transitive dependency of diffx` would help me a lot.

kamil-da: I've added comments for both `magnolia` and `mercator`. As @cocreature mentioned errors in this case were rather misleading, especially for the `mercator`.

bazel-java-deps.bzl
kamil-da: Done 🚮

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/DiffExtensions.scala
kamil-da: Good catch! There's already available instance of `Diff[Seq[T]]`, so I've removed the custom one.

maven_install_2.12.json
kamil-da: Done.

cocreature (approved):
Awesome, thanks! Do you have an example of a diff we got before and the diff we get now?

kamil-da:
> Awesome, thanks! Do you have an example of a diff we got before and the diff we get now?

Sure, see here:

**Simple status comparison**
`ai.x.diff`:
```
  SERVING -> UNKNOWN
```

`com.softwaremill.diffx`:
```
  io.grpc.health.v1.health.HealthCheckResponse.ServingStatus.SERVING -> io.grpc.health.v1.health.HealthCheckResponse.ServingStatus.UNKNOWN
```

**Objects comparison**
`ai.x.diff`:
```
  Seq(
    RecordField( ...,
      value = Some( Value( Text( "CKExposedByTemplate-alpha-150ec2f0cb212-key-0" -> "CKExposedByTemplate-alpha-150ec2f0cb212-key-1" ) ) )
    )
  )
```

`com.softwaremill.diffx`:
```
  List(
       0: RecordField(_1,Some(Value(Party(CKExposedByTemplate-alpha-1515516bb823d-party-0)))),
       1: RecordField(
            label: _2,
            value: Some(
                 value: Value(
                      sum: Text(
                           value: CKExposedByTemplate-alpha-1515516bb823d-key-0 -> CKExposedByTemplate-alpha-1515516bb823d-key-1)))))
```

**Sequence comparison (additional elements)**
`ai.x.diff` with our in-house instance of `DiffShow[Seq[T]]`:
```
  Seq( ...,
    2 = RecordField( label = "giver", value = Some( Value( Party( "CSDiscloseCreateToObservers-alpha-150ec2f0cb212-party-0" ) ) ) ),
    3 = RecordField(
      label = "observers",
      value = Some( Value( List( List( Vector(Value(Party(CSDiscloseCreateToObservers-alpha-150ec2f0cb212-party-1)), Value(Party(CSDiscloseCreateToObservers-alpha-150ec2f0cb212-party-2))) ) ) ) )
    )
  )
```

`com.softwaremill.diffx`:
```
  List(
       0: RecordField(giver,Some(Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-0)))),
       1: RecordField(observers,Some(Value(List(List(Vector(Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-1)), Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-2)))))))),
       2: RecordField(giver,Some(Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-0)))),
       3: RecordField(observers,Some(Value(List(List(Vector(Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-1)), Value(Party(CSDiscloseCreateToObservers-alpha-1515516bb823d-party-2)))))))))
```

I would say that `diffx` has nicer diffs. The only problematic case might be for comparing sequences in monochrome terminals because the `diffx` uses colours to differentiate additional/missing elements in their default `Diff[Iterable]` instance. Our custom instance for `DiffShow[Seq[T]]` had such problem solved by not only printing values but also their corresponding indices.
@cocreature Do you think this may be an issue for us?

kamil-da:
A screenshot for more colorful experience:
Screenshot 2021-02-15 at 18 02 26

cocreature:
Being harder to read if you don’t have colors is fine imho. Being unusable not, e.g., in CI environments it’s not that uncommon for colors to not work properly. Does the library have a way to detect if you’re writing to a terminal and  disable colors?

kamil-da:
I cannot see an option to disable colours in favour of some other graphical differentiation.
A solution would be to leave our custom instance for iterables.

cocreature:
That seems slightly disappointing. I’ll leave it to you and your team to decide whether that’s worth addressing or not.

stefanobaghino-da:
> That seems slightly disappointing. I’ll leave it to you and your team to decide whether that’s worth addressing or not.

Given that this was the situation we had before, I'd say it's fine to leave it like that.

kamil-da:
> > That seems slightly disappointing. I’ll leave it to you and your team to decide whether that’s worth addressing or not.
> 
> Given that this was the situation we had before, I'd say it's fine to leave it like that.

@stefanobaghino-da  I'm not sure if I understand correctly - you'd vote for leaving the default behaviour of `diffx` for iterables which is printing diffs in colours rather than explicitly with index annotation as we had before?

stefanobaghino-da:
> > > That seems slightly disappointing. I’ll leave it to you and your team to decide whether that’s worth addressing or not.
> > 
> > 
> > Given that this was the situation we had before, I'd say it's fine to leave it like that.
> 
> @stefanobaghino-da I'm not sure if I understand correctly - you'd vote for leaving the default behaviour of `diffx` for iterables which is printing diffs in colours rather than explicitly with index annotation as we had before?

I was just saying the differentiation by color is what we already had before, so it should be fine to keep it like that. That means, I don't think there's a need to work on additional differentiation.

kamil-da:
@stefanobaghino-da Ok 👍  Thank you

cocreature:
:shipit:

PR #8823, opened 2021-02-11
Daml Driver for PostgreSQL additional startup options

dasormeter:
Provide ability to startup the Daml Driver for PostgreSQL by running scripts and then completing without starting the index server
1) --sql-start-mode=migrate-only -- Only run schema migration scripts but do not start ledger api services -- error only if the execution of the flyway scripts fails.  Successful termination without error indicates that scripts were executed successfully

These are to address operational feedback requests from clients to run with restricted service account rights in the normal course of operations (mode #2) and only run with elevated rights when explicitly migrating to a new version of the schema when upgrading to a new version of the driver (mode #1)

i initially tried changing interfaces to allow for a Migrated mode or the existing `Resource[Either[Migrated, Ledger]]`Either option but that led to a preponderance of interface changes up and down the stack and through a whole mess of tests so instead i did a different mode only from the daml on sql main where i either run the migration scripts (and nothing else) or we do things as they were previously

i also opted not to tackle rationalizing and harmonizing all the start mode enumerations as this is just a patch for daml-on-sql, so the harmonizing can be done after this is merged back to main

### Pull Request Checklist

trigger the build.


ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlLedger.scala
```diff
@@ -88,21 +88,27 @@ private[sandbox] object SqlLedger {
...
-            Resource.unit
-        }
         retrievedLedgerId <- Resource.fromFuture(dao.lookupLedgerId())
         ledgerId <- Resource.fromFuture(retrievedLedgerId.fold(initialize(dao))(resume))
         _ <- Resource.fromFuture(initOrCheckParticipantId(dao, Tag.unwrap(participantId)))
         ledgerEnd <- Resource.fromFuture(dao.lookupLedgerEnd())
         ledgerConfig <- Resource.fromFuture(dao.lookupLedgerConfiguration())
         dispatcher <- dispatcherOwner(ledgerEnd).acquire()
         persistenceQueue <- new PersistenceQueueOwner(dispatcher).acquire()
+        _ <- startMode match {
```
SamirTalwar-DA, line 99: Why has this moved down?

How are all those DAO lookups above gonna work if the migrations haven't run yet?

ledger/sandbox/src/main/scala/platform/sandboxnext/StartupMode.scala
```diff
@@ -7,8 +7,12 @@ sealed trait StartupMode
 
 object StartupMode {
 
+  case object MigrateOnly extends StartupMode
```
SamirTalwar-DA, line 10: I like your style.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlStartMode.scala
```diff
@@ -7,10 +7,15 @@ private[sandbox] sealed abstract class SqlStartMode extends Product with Seriali
...
+  /** Run Migration scripts but do not start the ledger api service */
+  final case object MigrateOnly extends SqlStartMode
+
   /** Will continue using an initialised ledger, otherwise initialize a new one */
-  final case object ContinueIfExists extends SqlStartMode
+  final case object MigrateAndStart extends SqlStartMode
 
   /** Will always reset and initialize the ledger, even if it has data. */
-  final case object AlwaysReset extends SqlStartMode
+  final case object ResetAndStart extends SqlStartMode
```
SamirTalwar-DA, line 17: Thank you so much for standardizing these.

ledger/daml-on-sql/src/main/scala/on/sql/Cli.scala
```diff
@@ -43,6 +43,17 @@ private[sql] final class Cli(
...
+    parser
+      .opt[String]("sql-start-mode")
+      .optional()
+      .text(
+        s"The mode under which the schema SQL script will be run.  Defaults to ${DefaultConfig.sqlStartMode}.  Possible values are :" +
+          s"${PostgresStartupMode.MigrateOnly} : Run schema migration scripts and exit" +
+          s"${PostgresStartupMode.MigrateAndStart} : Run schema migration scripts and start service" +
+          s"${PostgresStartupMode.ValidateAndStart} : Validate schema migration scripts and start service"
+      )
+      .action((mode, config) => config.copy(sqlStartMode = PostgresStartupMode.fromString(mode)))
```
SamirTalwar-DA, line 55: Can we make these lowercase and hyphenated, to align with other parameters (and general Linux conventions)?

Something like `--sql-start-mode=migrate-and-start`, etc.

ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxServer.scala
```diff
@@ -102,6 +103,16 @@ object SandboxServer {
       }
     } yield server
 
+  // Run only the flyway migrations but do not initialize any of the ledger api or indexer services
+  def migrateOnly(config: SandboxConfig): ResourceOwner[Unit] =
```
SamirTalwar-DA, line 107: Nice solution!

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlLedger.scala
SamirTalwar-DA: You need to move this back up to get the tests to pass.

ledger/daml-on-sql/src/main/scala/on/sql/Cli.scala
dasormeter: changed and now there are only `migrate-only` and `migrate-and-start` options 

also i removed `validate-and-start` as we really cannot do that given we always run schema migrate before initializing dao for the reset behaviour stuff

ledger/daml-on-sql/README.rst
```diff
@@ -31,6 +31,24 @@ To specify the PostgreSQL instance you wish to connect, use the
...
+
+If you want to setup the database schema for the *Daml Driver for PostgreSQL* with a
+more highly permissioned database service account that is able to create database schemas
+and tables but only run your application with an account with lesser permissions, you
+can leverage the combination of the connection string specifying the account to use
+and the ``--sql-start-mode <value>`` parameter specifying the mode to run the driver.
+
+The possible modes available are
+``MigrateOnly`` - Create database schema but do not start the driver
+``MigrateAndStart`` - Create database schema and start the driver.  If the database already exists, this will simply validate and start.  This is the default mode
```
SamirTalwar-DA, line 50: These need updating.

SamirTalwar-DA (approved):
README needs updating, but otherwise I am satisfied.

ledger/daml-on-sql/README.rst
dasormeter: I think these are correct actually -- `MigrateAndStart` does actually just validate and start if the schemas are already there (if the scripts have already been run).  So there is effectively no difference in running flyway validate vs flyway migrate if scripts have already been run and initialized from what i can see.

SamirTalwar-DA: No, I mean the names: `migrate-only` and `migrate-and-start`. :slightly_smiling_face:

ledger/daml-on-sql/src/main/scala/on/sql/Main.scala
```diff
@@ -17,5 +17,12 @@ object Main {
   }
 
   private[sql] def run(config: SandboxConfig): Unit =
-    new ProgramResource(SandboxServer.owner(Name, config)).run(ResourceContext.apply)
+    config.sqlStartMode.foreach {
+      case PostgresStartupMode.MigrateOnly =>
+        new ProgramResource(SandboxServer.migrateOnly(config)).run(ResourceContext.apply)
+        sys.exit()
```
SamirTalwar-DA, line 23: You don't wanna do this. `ProgramResource#run` doesn't block.

Instead, make `migrateOnly` return a `Future[Unit]`, and then do this:

```suggestion
        SandboxServer.migrateOnly(config)
          .onComplete {
             case Success(_) =>
               sys.exit(0)
             case Failure(exception) =>
               System.err.println(exception.getMessage)
               sys.exit(1)
           }
```

ledger/daml-on-sql/README.rst
dasormeter: thanks for spotting - now resolved

ledger/daml-on-sql/src/main/scala/on/sql/Main.scala
dasormeter: resolved in latest push with move to Future

PR #8829, opened 2021-02-12
Improve daml script dump testing infrastructure

### LINKED ISSUE #8772: Improve Daml Script dump integration tests
Currently, we have a fairly hardcoded daml script dump integration test. We should have something a bit more generic to make it easy to test things.

I would suggest is to write a function that can compare transaction trees (or really streams thereof, potentially starting with an ACS) modulo alpha renaming of contract ids and party ids. That should make it relatively easy to make sure that the daml script dump does exactly what it should do in a generic fashion. We could then have tests of the form "run this daml script (or whatever other mechanism we use for producing a transaction tree), run the daml script dump for a different set of parties, assert that the transaction trees are equivalent (using the function mentioned above).

cocreature:
This PR adds a generic function to compare to transaction trees (or
rather sequences thereof) modulo alpha equivalence wtr to contract ids
and party ids.

This allows us to abstract away the concrete integration test for Daml
Script dump we had before and turn it into a generic process:

1. Run some setup code to generate transactions.
2. Query and store transaction stream.
3. Create the dump, build it and run it for a different set of
parties.
4. Query and store transaction stream for new parties.
5. Diff transaction streams.

This PR does not change tests or add new ones. I think it might be
nice to use Daml Scripts for 1 as well but I’ll leave that for a
separate PR. The infrastructure here already supports this.

fixes #8772

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
+++ b/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
@@ -28,15 +28,20 @@ import com.daml.ledger.client.configuration.{
   LedgerClientConfiguration,
   LedgerIdRequirement,
 }
+import com.daml.ledger.testing.utils.TransactionEq
 import com.daml.lf.archive.{Dar, DarReader, Decode}
 import com.daml.platform.sandbox.services.TestCommands
 import com.daml.platform.sandboxnext.SandboxNextFixture
 import com.daml.SdkVersion
 import scalaz.syntax.tag._
+import scalaz.std.scalaFuture._
+import scalaz.std.list._
 import scalaz.syntax.traverse._
 import spray.json._
 
+import scala.concurrent.Future
 import scala.sys.process._
+import org.scalatest._
 import org.scalatest.freespec.AsyncFreeSpec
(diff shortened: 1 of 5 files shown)


daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
```diff
@@ -202,51 +149,113 @@ final class IT
...
-          Some(LedgerOffset().withBoundary(LedgerOffset.LedgerBoundary.LEDGER_END)),
-          transactionFilter(p3),
+      // check that the new transaction trees are the same
+      after <- collectTrees(client, newParties)
+    } yield {
+      TransactionEq.equivalent(before, after).fold(fail(_), _ => succeed)
+    }
+
+  "Generated dump for IOU transfer compiles" in {
+    test(2) { case (client, Seq(p1, p2)) =>
```
cocreature, line 177: I recommend viewing that with whitespace changes hidden.

aherrmann-da (approved):
LGTM

PR #8831, opened 2021-02-12
Port //daml-script/... to Scala 2.13

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -183,12 +183,12 @@ jobs:
           //ledger/... \
           //ledger-service/... \
           //triggers/... \
+          //daml-script/... \
           -//ledger-service/http-json:tests \
           -//ledger-service/http-json:integration-tests \
           -//ledger-service/http-json:failure-tests \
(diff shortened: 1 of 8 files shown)


aherrmann-da (approved):

ci/build.yml
```diff
@@ -183,12 +183,12 @@ jobs:
           //ledger/... \
           //ledger-service/... \
           //triggers/... \
+          //daml-script/... \
```
garyverhaegen-da, line 186: Was it deliberate to add `//daml-script/` twice?

cocreature: The upper one is the `bazel build` step, the lower one is the `bazel test` step. Arguably those should be factored out but given that I’d like to get this to just `//...`, I couldn’t be bothered to do that so far.

garyverhaegen-da: Oh right, missed that as the change in command was in the collapsed lines.

PR #8834, opened 2021-02-12
damlc test: show full coverage report

robin-da:
We add a flag `show-coverage` to `damlc-test` to show templates that
are never created and choices that are never executed during the tests.

CHANGELOG_BEGIN
[damlc test] A new flag `show-coverage` shows full test coverage
reports.
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/Test.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/Test.hs
@@ -138,10 +138,7 @@ printTestCoverage ShowCoverage {getShowCoverage} dalfs results
         , Just contractInstance <- [node_CreateContractInstance]
         ]
     missingTemplates =
-        [ (LF.moduleNameString $ LF.moduleName m) <> ":" <>
-        (T.intercalate "." $ LF.unTypeConName $ LF.tplTypeCon t)
-        | (m, t) <- templates
-        ] \\
+        [fullTemplateName m t | (m, t) <- templates] \\
         [TL.toStrict $ SS.identifierName tId | Just tId <- coveredTemplates]
     coveredChoices =
         nubSort $
@@ -153,16 +150,15 @@ printTestCoverage ShowCoverage {getShowCoverage} dalfs results
         , Just templateId <- [node_ExerciseTemplateId]
         ]
     missingChoices =
-        [ ( (LF.moduleNameString $ LF.moduleName m) <> ":" <>
-            (T.concat $ LF.unTypeConName $ LF.tplTypeCon t)
-          , LF.unChoiceName n)
-        | (m, t, n) <- choices
-        ] \\
+        [(fullTemplateName m t, LF.unChoiceName n) | (m, t, n) <- choices] \\
         [(TL.toStrict $ SS.identifierName t, TL.toStrict c) | (t, c) <- coveredChoices]
     nrOfTemplates = length templates
     nrOfChoices = length choices
     coveredNrOfChoices = length coveredChoices
     coveredNrOfTemplates = length coveredTemplates
+    fullTemplateName m t =
+        (LF.moduleNameString $ LF.moduleName m) <> ":" <>
+        (T.concat $ LF.unTypeConName $ LF.tplTypeCon t)
 
 printScenarioResults :: [(VirtualResource, SS.ScenarioResult)] -> UseColor -> IO ()
 printScenarioResults results color = do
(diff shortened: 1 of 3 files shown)


compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -91,36 +101,49 @@ failedTestOutput h file = do
...
         | n <- allScenarioNodes
         , Just (SS.NodeNodeCreate SS.Node_Create {SS.node_CreateContractInstance}) <-
               [SS.nodeNode n]
         , Just contractInstance <- [node_CreateContractInstance]
         ]
-    coveredNrOfChoices =
-        length $
+    missingTemplates =
+        [ (LF.moduleNameString $ LF.moduleName m) <> ":" <>
+        (T.concat $ LF.unTypeConName $ LF.tplTypeCon t)
```
cocreature, line 142: don’t you need to intersperse dots instead of simply concatenating?

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -91,36 +101,49 @@ failedTestOutput h file = do
...
         [ SS.contractInstanceTemplateId contractInstance
         | n <- allScenarioNodes
         , Just (SS.NodeNodeCreate SS.Node_Create {SS.node_CreateContractInstance}) <-
               [SS.nodeNode n]
         , Just contractInstance <- [node_CreateContractInstance]
         ]
-    coveredNrOfChoices =
-        length $
+    missingTemplates =
+        [ (LF.moduleNameString $ LF.moduleName m) <> ":" <>
```
cocreature, line 141: You’re ignoring package ids in your comparison. I don’t think we should. I’d also suggest factoring this into a helper function, something like `Qualified TypeCon -> Text`.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -129,6 +152,18 @@ printTestCoverage dalfs results
                                                      }) <- [SS.nodeNode n]
         , Just templateId <- [node_ExerciseTemplateId]
         ]
+    missingChoices =
+        [ (LF.moduleNameString $ LF.moduleName m) <> ":" <>
+        (T.concat $ LF.unTypeConName $ LF.tplTypeCon t) <>
```
cocreature, line 157: Instead of mixing the template and the choice into a single string, can we just work with a tuple? For templates, the string makes sort of sense since we get that from the scenario protobuf but choice are separate. No need to turn everything into a single string.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -91,36 +101,49 @@ failedTestOutput h file = do
...
               [SS.nodeNode n]
         , Just contractInstance <- [node_CreateContractInstance]
         ]
-    coveredNrOfChoices =
-        length $
+    missingTemplates =
+        [ (LF.moduleNameString $ LF.moduleName m) <> ":" <>
+        (T.concat $ LF.unTypeConName $ LF.tplTypeCon t)
+        | (m, t) <- templates
+        ] \\
```
cocreature, line 144: Let’s turn this into a set lookup. I don’t want to end up on accidentally quadratic :)

cocreature (approved):
Very nice, thanks! A few small comments

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: Totally agreed.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: not quite sure, but I think you're right.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: shouldn't all modules that we test come from the same package (the one we're testing?). Factored out the helper function.

cocreature: the modules we test against atm come from the same package but we list templates from all packages. That’s the mismatch I was trying to get at in the original PR for test coverage.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: turned it to sets.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: done.

PR #8838, opened 2021-02-12
Only bind contract ids that are used

### LINKED ISSUE #8775: Only bind contract ids that are used
Currently DAML Script dumps bind every contract id returned by `submitTree`. We should limit that to only the ones that are actually used later.

aherrmann-da:
Closes #8775 

Adds an additional traversal of the tree to determine all referenced contract ids.
Filters out contract ids that are never referenced before creating bindings for them.
Adds tests for the feature.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
@@ -81,7 +81,7 @@ object TreeUtils {
 
   case class CreatedContract(cid: String, tplId: Identifier, path: List[Selector])
 
-  def treeCids(tree: TransactionTree): Seq[CreatedContract] = {
+  def treeCreatedCids(tree: TransactionTree): Seq[CreatedContract] = {
     var cids: Seq[CreatedContract] = Seq()
     traverseTree(tree) { case (selectors, kind) =>
       kind match {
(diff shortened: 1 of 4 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -94,6 +94,20 @@ object TreeUtils {
...
 
+  def treeReferencedCids(tree: TransactionTree): Set[String] = {
+    var cids: Set[String] = Set.empty
+    traverseTree(tree) { case (_, kind) =>
+      kind match {
+        case Kind.Empty =>
+        case Kind.Exercised(value) =>
+          cids += value.contractId
+          cids ++= value.choiceArgument.foldMap(arg => valueCids(arg.sum))
+        case Kind.Created(_) =>
```
cocreature, line 105: We also need to recurse here on the create argument. A template can contain a contract id as a field.

cocreature (approved):
Nice! I think there is one issue with references in create nodes that we should address.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Good catch, added.

PR #8840, opened 2021-02-12
LF: (fix) compute exercise node seed is only once.

remyhaemmerle-da:
This PR fix a regression performence intoriduce in #8804, were we
calculate exercise node seed twice.

Here is the result of a benchmark:

before:
CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:test  avgt   20  48.968 ± 0.667  ms/op

after:
CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:test  avgt   20  41.772 ± 0.374  ms/op

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -64,7 +64,7 @@ private[lf] object PartialTransaction {
     */
   final case class Context(info: ContextInfo, children: BackStack[NodeId], nextChildIdx: Int) {
     def addChild(child: NodeId): Context = Context(info, children :+ child, nextChildIdx + 1)
-    // This function may be costly, it must be call at most one for each node.
+    // This function may be costly, it must be call at most once for each node.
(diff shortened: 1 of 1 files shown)


cocreature (approved):
good catch!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -64,7 +64,7 @@ private[lf] object PartialTransaction {
     */
   final case class Context(info: ContextInfo, children: BackStack[NodeId], nextChildIdx: Int) {
     def addChild(child: NodeId): Context = Context(info, children :+ child, nextChildIdx + 1)
-    // needs to be lazy as info.childrenSeeds may be undefined on children.length
+    // This function may be costly, it must be call at most one for each node.
```
remyhaemmerle-da, line 67:
```suggestion
    // This function may be costly, it must be call at most once for each node.
```

PR #8849, opened 2021-02-15
kvutils: Split `Raw.Key` into two, and rename `Raw.Value` to `Raw.Envelope`.

SamirTalwar-DA:
`Key` was a bit too generic: it could refer to either a log entry ID or a state key.

This splits `Raw.Key` into `Raw.LogEntryId` and `Raw.StateKey`, and changes all usages to use one of the two. `Raw.Key` is now a supertype of both of these new types, with an implicit conversion to a third subtype, `Raw.UnknownKey`, to maintain some backwards-compatibility.

The only exception is the integrity checker. Because of the way we export data, we don't know if any given item is a log entry or a state key-value pair. We therefore parse everything as a `Raw.UnknownKey`, and then convert based on usage. This means the type safety is about the same as before in this particular case.

`Raw.Value` is renamed to `Raw.Envelope`, because it's always a serialized envelope. `Raw.Value` is kept as a supertype for now to parallel the `Raw.Key` trait, but might be removed in the future.

Finally, `@deprecated` annotations are added for all conversions between `Raw` types, as well as conversions to and from `ByteString`. Users will have until DAML SDK v1.12 to migrate.

### Changelog

- **[Integration Kit]** The various uses of `ByteString` in kvutils are now strongly typed. Users of kvutils will need to use the `Raw.Bytes` subtypes to represent the various keys and values that flow through a kvutils-based driver. ByteStrings representing keys need to be wrapped in either `Raw.LogEntryId` or `Raw.StateKey`, and ByteStrings representing value envelopes need to be wrapped in `Raw.Envelope`. This prevents accidentally confusing one for the other.

  Implicit conversions have been provided for Scala users, which should ease the transition. There are marked `@deprecated` and will be removed before the release of DAML SDK v1.12.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/tools/integrity-check/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/RawWriteSetComparison.scala
+++ b/ledger/participant-state/kvutils/tools/integrity-check/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/RawWriteSetComparison.scala
@@ -41,11 +41,15 @@ final class RawWriteSetComparison(stateKeySerializationStrategy: StateKeySeriali
   private def writeSetToStrings(writeSet: WriteSet): Seq[String] =
     writeSet.view.map((writeItemToString _).tupled).toVector
 
-  private def writeItemToString(key: Raw.Key, value: Raw.Value): String = {
-    val keyString = stateKeySerializationStrategy.deserializeStateKey(key)
-    val valueString = kvutils.Envelope.open(value)
-    s"$keyString -> $valueString"
-  }
+  private def writeItemToString(rawKey: Raw.Key, rawEnvelope: Raw.Envelope): String =
+    parsePair(rawKey, rawEnvelope) match {
+      case Left(errorMessage) =>
+        throw new IllegalArgumentException(errorMessage)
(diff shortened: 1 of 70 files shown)


miklos-da (approved):

ledger/participant-state/kvutils/tools/integrity-check/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/RawWriteSetComparison.scala
```diff
@@ -54,15 +58,18 @@ final class RawWriteSetComparison(stateKeySerializationStrategy: StateKeySeriali
     val differencesExplained = expectedWriteSet
       .zip(actualWriteSet)
       .map { case ((expectedKey, expectedValue), (actualKey, actualValue)) =>
-        if (expectedKey == actualKey && expectedValue != actualValue) {
+        // We need to compare `.bytes` because the type will be different.
+        // Unfortunately, the export format doesn't differentiate between log and state entries.
```
miklos-da, line 62: The reason for that is that on a real ledger (unless we store these separately) we cannot distinguish them either.

PR #8850, opened 2021-02-15
damlc: fix: catch UserError instead  IOException

robin-da:
When the package-db metadata can not be parsed, an exception is thrown
and needs to be catched.  Previously we were catching IOException, but
in this case it's actually a UserError that is thrown.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
@@ -214,7 +214,7 @@ dbNeedsReinitialization projectRoot allDeps sdkVersion damlLfVersion = do
     errOrmetaData <- tryAny $ readMetadata projectRoot
     pure $
         case errOrmetaData of
-            Left (_err :: SomeException)-> (True, depsFingerprint)
+            Left _err -> (True, depsFingerprint)
             Right metaData -> (fingerprintDependencies metaData /= depsFingerprint, depsFingerprint)
 
 disableScenarioService :: Options -> Options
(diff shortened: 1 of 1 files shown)


compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -211,10 +211,10 @@ dbNeedsReinitialization projectRoot allDeps sdkVersion damlLfVersion = do
     let depsFingerprint =
             fingerprintFingerprints $ sdkVersionFingerprint : damlLfFingerprint : fileFingerprints
     -- Read the metadata of an already existing package database and see if wee need to reinitialize.
-    errOrmetaData <- tryIO $ readMetadata projectRoot
+    errOrmetaData <- try $ readMetadata projectRoot
```
cocreature, line 214:
```suggestion
    errOrmetaData <- tryAny $ readMetadata projectRoot
```

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -211,10 +211,10 @@ dbNeedsReinitialization projectRoot allDeps sdkVersion damlLfVersion = do
...
     pure $
         case errOrmetaData of
-            Left _err -> (True, depsFingerprint)
+            Left (_err :: SomeException)-> (True, depsFingerprint)
```
cocreature, line 217:
```suggestion
            Left _err-> (True, depsFingerprint)
```

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -8,7 +8,7 @@ module DA.Cli.Damlc.Packaging
   ) where
 
 import qualified "zip-archive" Codec.Archive.Zip as ZipArchive
-import Control.Exception.Safe (tryIO)
+import Control.Exception.Safe (try, SomeException)
```
cocreature, line 11:
```suggestion
import Control.Exception.Safe (tryAny)
```
semantically identical but you don’t need the extra type annotation.

cocreature (approved):
Good catch! (no pun intended :slightly_smiling_face:)

PR #8853, opened 2021-02-15
Port non-repudiation middleware to Scala 2.13

cocreature:
Tests are still missing and blocked on #8821.

The main change here is the switch from `ArraySeq[Byte]` to
`ArraySeq.ofByte`. `ArraySeq` allows for boxed and unboxed
representaitons. That means that `ArraySeq[Byte]unsafeArray` does not always return
an Array[Byte] (boxed version would be Array[AnyRef]).

Apparently collection-compat has taken the yolo approach and pretends
it can give you an Array[Byte] anyway :shrug: Scala 2.13 on the other
hand, does things properly in this regard which means the code relying
on `unsafeArray` fails to compile.

`ArraySeq.ofByte` is the specialized unboxed version where none of
this is an issue on both 2.13 and 2.12.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -178,12 +178,12 @@ jobs:
         bazel build --config scala_2_13 -- \
           //... \
           -//ledger/ledger-api-test-tool/... \
-          -//runtime-components/...
+          -//runtime-components/non-repudiation-postgresql:test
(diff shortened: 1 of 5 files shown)


stefanobaghino-da (approved):
🤟🏻

PR #8856, opened 2021-02-16
LF: release LF 1.12

remyhaemmerle-da:
CHANGELOG_BEGIN

- LF: Release LF 1.12. This version reduce the size of transaction

- Compiler: Change the default LF output from 1.8 to 1.11.

CHANGELOG_END

### Pull Request Checklist

trigger the build.


language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
```diff
@@ -37,7 +37,7 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
 
     val (interfaces, pkgPrefixes) = CodeGenRunner.collectDamlLfInterfaces(conf)
 
-    assert(interfaces.length == 19)
+    assert(interfaces.length == 20)
```
remyhaemmerle-da, line 40: @sofiafaro-da Can you confirm there is one more package in archive 1.11 than in archive 1.8 ?

sofiafaro-da: Yes, the stable package for DA.Set.Types

remyhaemmerle-da: :+1:

sofiafaro-da (approved):

language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
stefanobaghino-da: Would it make sense and be feasible to compare the expected names instead of just checking whether the count checks out? In this way this question would answer itself in the test. 🙂

remyhaemmerle-da: It is probably possible. let's ask application-runtime team ? ;)

stefanobaghino-da: 😂 

Good point, I'll make a note to have a look at this once this has been merged.

stefanobaghino-da: Would [this](https://github.com/digital-asset/daml/pull/8921) make sense?

remyhaemmerle-da: If you think it is better, I have no objection for your proposal.

remyhaemmerle-da: I have a double though about this. 
As long it is hard-coded I will prefer to not be changed. 
This will just make my life more complicated each time I have to update the compiler version. 
The test should be designed to work with any version of LF.

remyhaemmerle-da: Please take into account we have a lot of test to fix every time we roll the LF version.

stefanobaghino-da: Let's keep the discussion in that thread, sorry for opening it up here.

remyhaemmerle-da: What the test is verifying actually. Is it not enough to check it contains "Foo:Foo" ?

S11001001 (approved):

cocreature:
:partying_face:

PR #8857, opened 2021-02-16
Dpp 239 indexer parallel ingestion poc

nmarton-da:
### Pull Request Checklist

trigger the build.


compatibility/WORKSPACE
```diff
@@ -163,7 +163,7 @@ dev_env_tool(
 
 nixpkgs_package(
     name = "postgresql_nix",
-    attribute_path = "postgresql_9_6",
+    attribute_path = "postgresql_12",  # requirement of the partitioned schema
```
stefanobaghino-da, line 166: It's probably worth mentioning in the documentation. Would this represent a problem for others working with SQL-backed ledgers?

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only.sql
```diff
@@ -0,0 +1,535 @@
...
+
+
+
+
+------------------------------------------------------------------------------------------------------
+-- new
+
+
+
+-- drop the mutable tables
```
stefanobaghino-da, line 230: You are probably already thinking of this, but just as a reminder, the migration (for PostgreSQL) must also migrate the data, not only the schema (and @da-tanabe should probably be made aware that, if I'm not wrong, this will take some time).

stefanobaghino-da: As a further point, if by any chance we are completely reworking all tables, it may be worth squashing the Flyway schema history (not sure whether this could be feasible as a follow-up, @gerolf-da).

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
```diff
@@ -12,6 +12,8 @@ import com.daml.ledger.test.model.Test.{Asset, Divulgence1, Divulgence2, Proposa
 import scalaz.Tag
 
 final class DivulgenceIT extends LedgerTestSuite {
+
+// TODO this one is a flake, at daml-on-sql it fails constantly in orchestration, but never in isolation, on ledger-on-memory it fails like 5% of the runs
```
stefanobaghino-da, line 16: 👀

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
```diff
@@ -125,7 +125,8 @@ private[dao] trait JdbcLedgerDaoPostCommitValidationSpec extends LoneElement {
     }
   }
 
-  it should "be able to use divulged contract in later transaction" in {
+  // TODO it seems as this test would be based on flawed assumptions (a pure divulgance without create contract having a ledger_effective_time)
```
stefanobaghino-da, line 128: Can this be a situation that arises in a distributed setting, with a contract created remotely being divulged to a local party?

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
```diff
@@ -21,7 +21,11 @@ trait JdbcPipelinedInsertionsSpec extends Inside with OptionValues with Matchers
 
   behavior of "JdbcLedgerDao (on PostgreSQL)"
 
-  it should "allow idempotent transaction insertions" in {
+  // TODO ignoring this test: this idempotent re-ingestion approach is backed by unique index on event_id
+  // TODO with future ingestion strategies this might be not necessary
+  // TODO current PoC schema does not support this DB level uniqueness checks
+  // TODO cc @simon@
+  ignore should "allow idempotent transaction insertions" in {
```
stefanobaghino-da, line 28: I think the idea behind this was to show leniency towards the ledger, so that in case of crash it could start sending transactions from some sort of checkpoint. Does it make sense? Is this still something we can do?

stefanobaghino-da (commented):
I may have indulged in certain comments too eagerly, feel free to ignore anything. Probably the most relevant ones are about ignored tests.

stefanobaghino-da:
conflicts!

compatibility/WORKSPACE
nmarton-da: For the final rollout we might stick to manual partitioning. (context: this bump of the postgresql version is needed because current schema using postgre-specific declarative partitioning) This might would mean no version increase eventually.
Otherwise very good comment: we have now uniform ephemeral postgreSQL support, if we want to differentiate, that would require some extra work.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only.sql
nmarton-da: Good observation: this is only for PoC purposes, hence no migration. For prod rollout migration is a must.
We might end up having feature flag guarder rollout here, so it might be even more complicated than that

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
nmarton-da: AFAIK there are two distinct cases for divulgence:
- public: all divulged contracts have created event in the indexer already
- private: the above is not true, but in this case the divulged contract itself is exposed (so create argument, and template id we can get this way)

Furthermore: in the private mode we do not have ledger_effective_time information of the creation of the contract. And therefore also postcommit validation is not supported.

Also: AFAIK post commit validation is only running in case daml-on-sql/sandbox, which are working in public mode.

Also: question is how can we do post-commit validation if we do not have the create event for the contract? AFAIK: the create event's ledger_effective_Time should be considered foll post commit validation

the above are working theories, let's discuss this further and align them!

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
nmarton-da: Currently I do not have an answer for this, @meiersi-da can you comment on this?
@stefanobaghino-da, @tudor-da do you know maybe what exactly is guarding this?

tudor-da: This asserts the capability of the ledger dao to _repeat_ events. It is needed with batching, since there one ledger end update guards multiple submissions.

tudor-da: In the current design, this asserts the capability of the ledger dao to ingest the same events/state more than once. It's needed due to pipelining, since the ledger end is updated in a subsequent SQL transaction.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
stefanobaghino-da: After an offline discussion with @nmarton-da, we realized this test covers an impossible case.

Post-commit validation is only performed in the "write directly to the index without ~collecting $20~ going through the ledger" (a.k.a. sandbox-classic). But in the sandbox-classic case, divulgences will _always_ be empty ([code](https://github.com/digital-asset/daml/blob/7ae28beff98df7bfd9d5368ea22d46717c0cc353/ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlLedger.scala#L382)).

As a matter of fact, the `divulged` parameter of the post-commit validation function can be dropped entirely.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
nmarton-da: After discussing this @stefanobaghino-da I suggest to omit further investigations hence high probability of not functional bug as root cause 
@meiersi-da, what do you think?

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
nmarton-da: @tudor-da thanks for explaining. do you know maybe what exactly is guarding this? Is it a unique index on something?

tudor-da: Yes:
* On `participant_events` we have `event_id`
* On `participant_contracts` we have `contract_id` and `create_key_hash` unique index
* On `participant_contract_witnesses` we have `(contract_id, contract_witness)`
All insertions in these tables are using conflict action `DO NOTHING`. That's what that test is actually asserting

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
```diff
@@ -138,8 +140,7 @@ private[events] object EventsTableFlatEvents {
           join parameters on
               (participant_pruned_up_to_inclusive is null or event_offset > participant_pruned_up_to_inclusive)
               and event_offset <= ledger_end
-          where transaction_id = $transactionId and #$witnessesWhereClause
-          group by (#$groupByColumns)
+          where transaction_id = $transactionId and #$witnessesWhereClause and event_kind != 0 -- we do not want to fetch divulgence events
```
meiersi-da, line 143: Do we have a test that fails if we don't have this restriction? If not, let's add one when implementing this properly.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -308,17 +308,24 @@ private[events] object EventsTableFlatEventsRangeQueries {
...
+            from participant_events as active_cs
+            where active_cs.event_kind = 10 -- create
+                  and active_cs.event_sequential_id > ${range.startExclusive._2: Long}
+                  and active_cs.event_sequential_id <= ${range.endInclusive._2: Long}
+                  and not exists (
+                    select 1
+                    from participant_events as archived_cs
+                    where
+                      archived_cs.contract_id = active_cs.contract_id and
+                      archived_cs.event_kind = 20 and -- consuming and
```
meiersi-da, line 324:
```suggestion
                      archived_cs.event_kind = 20 and -- consuming
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -396,22 +424,29 @@ private[events] object EventsTableFlatEventsRangeQueries {
...
       SQL"""select #$selectColumns, #$filteredWitnesses as event_witnesses,
-                   case when #$submittersInPartiesClause then command_id else '' end as command_id
-            from participant_events
-            where create_argument is not null
-                  and event_sequential_id > ${range.startExclusive._2: Long}
-                  and event_sequential_id <= ${range.endInclusive._2: Long}
-                  and (create_consumed_at is null or create_consumed_at > ${range.endInclusive._1: Offset})
+                   case when #$submittersInPartiesClause then active_cs.command_id else '' end as command_id
+            from participant_events as active_cs
+            where active_cs.event_kind = 10 -- create
```
meiersi-da, line 437: Seeing the repetition here, I wonder whether we couldn't keep the queries as is by introducing an appropriate view?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableH2Database.scala
```diff
@@ -171,15 +172,16 @@ object EventsTableH2Database extends EventsTable {
   private val updateArchived =
     """update participant_events set create_consumed_at={consumed_at} where contract_id={contract_id} and create_argument is not null"""
```
meiersi-da, line 173:
```suggestion
```

We shouldn't need that anymore should we?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableH2Database.scala
```diff
@@ -198,8 +200,8 @@ object EventsTableH2Database extends EventsTable {
       exerciseResults = serialized.exerciseResults,
     )
 
-    val archivals =
-      info.archives.iterator.map(archive(tx.offset)).toList
+    val archivals = Nil
```
meiersi-da, line 203: Where do we need this?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -47,6 +40,10 @@ object TransactionIndexing {
           createArgumentsByContract += ((create.coid, createArgument))
           createArguments += ((nodeId, createArgument))
           createKeyValue.foreach(key => createKeyValues += ((nodeId, key)))
+          createKeyValue.foreach(_ => createKeyHashes += ((nodeId, create.key
+            .map(convertLfValueKey(create.templateId, _))
+            .map(_.hash.bytes.toByteArray)
```
meiersi-da, line 45: Why two consecutive maps instead of one?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -141,16 +108,17 @@ object TransactionIndexing {
...
-      val netCreates = created.filterNot(c => archived(c.coid))
-      val netArchives = archived.filterNot(allCreatedContractIds)
-      val netDivulgedContracts = divulgedContracts.filterNot(c => allContractIds(c.contractId))
-      val netTransactionVisibility =
-        Relation.from(visibility.result()).view.filterKeys(!archived(_)).toMap
-      val netVisibility = Relation.union(netTransactionVisibility, visibility(netDivulgedContracts))
+      val divulgedContractIndex = divulgedContracts
+        .map(divulgedContract => divulgedContract.contractId -> divulgedContract)
+        .toMap
+      val netDivulged = blinding.divulgence.map {
```
meiersi-da, line 114: Why `netDivulged`? It's not clear to me what kind of netting we are doing here.

Suggestion: rename to `divulgenceEvents`

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -161,18 +129,11 @@ object TransactionIndexing {
...
         ),
         contracts = ContractsInfo(
-          netCreates = netCreates,
-          netArchives = netArchives,
-          divulgedContracts = netDivulgedContracts,
-        ),
-        contractWitnesses = ContractWitnessesInfo(
-          netArchives = netArchives,
-          netVisibility = netVisibility,
+          divulgedContracts = netDivulged,
```
meiersi-da, line 136: Consider moving divulgedContracts into the `EventsInfo`, as we can think of a transaction containing a bunch of divulgence events alongside the create and exercise events.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -188,15 +149,19 @@ object TransactionIndexing {
...
+                                         visibility: Set[Party])
+
   final case class ContractsInfo(
-      netCreates: Set[Create],
-      netArchives: Set[ContractId],
-      divulgedContracts: Iterable[DivulgedContract],
+      divulgedContracts: Iterable[TransactionDivulgedContract],
   )
 
   final case class ContractWitnessesInfo(
```
meiersi-da, line 167: This name reads wrong to me. It contains no witness information, and stores data about both create and exercise events.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
```diff
@@ -148,7 +149,8 @@ private[dao] trait JdbcLedgerDaoPostCommitValidationSpec extends LoneElement {
     }
   }
 
-  it should "refuse to insert entries with conflicting transaction ids" in {
+  // TODO figure out what is leading to exception here (maybe the unique constraint on the event_id?), then figure out whether we can re-add it, or if we need this feature at all?
```
meiersi-da, line 152: @gerolf-da do you know why we want the IndexDB to reject this case? It should not happen if upstream works correctly, and if it does not, we wouldn't want to stop ingesting, but issue a warning.

meiersi-da (commented):
Thank you @nmarton-da . This is great; and it's funny how we almost manage to reduce the amount of code ;-)

I've reviewed everything except for the integrity-checker changes. Please ping me in case I've forgot to comment on an `@simon` marker.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcPipelinedInsertionsSpec.scala
nmarton-da: ok, this we certainly need to take into consideration for the rollout.
right now I will postpone this to ship an experimental PR to LR and VMBC.
Until resolved I leave this discussion open.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsReader.scala
```diff
@@ -61,17 +120,60 @@ private[dao] sealed class ContractsReader(
...
+          WHERE contract_id = $contractId
+            AND event_kind = 0 -- divulgence
+            AND event_sequential_id <= parameters.ledger_end_sequential_id
+            AND #$tree_event_witnessesWhere
+          ORDER BY event_sequential_id
+            -- prudent engineering: make results more stable by preferring earlier divulgence events
+            -- Results might still change due to pruning.
+          LIMIT 1
+       ),
+       create_and_divulged_contracts AS (
```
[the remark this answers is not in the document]
nmarton-da: true, fixed

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
```diff
@@ -82,21 +82,22 @@ private[events] object EventsTableFlatEvents {
...
-      "contract_id",
-      "template_id",
-      "create_argument",
-      "create_signatories",
-      "create_observers",
-      "create_agreement_text",
-      "create_key_value",
+      "active_cs.event_offset",
+      "active_cs.transaction_id",
+      "active_cs.node_index",
```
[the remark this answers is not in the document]
nmarton-da: At line 23: it is part of the EventsTable.Entry. Not sure why.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
nmarton-da: added a todo

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEventsRangeQueries.scala
nmarton-da: Yeah these repetitions are a pain I guess it is due to the resilience of anorm's SQL interpolation for re-use :(
Yes that would be a solution for DRY to do it in SQL, but it is something new in a sense that current code-base is consistently keeping SQL logic in these anorm string literals. Views would break that pattern, next to being a code artifact which is not maintained in code, but in code + in flyway update scripts.
We might want to look into ways how we can reuse anorm literals, or switch to something else here which enables reuse, so we have an alternative to the pure SQL approach. Tradeoffs shall be evaluated.
TLDR: I agree completely with the initiative, and added a TODO for later

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableH2Database.scala
nmarton-da: Correct.
Please note that building H2 support was not a goal of these changes.
I only did the minimal corrections to make it compile: which also means all test over H2 fail currently.
There is already a TODO to propagate support of the append only schema to H2 artifacts.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableH2Database.scala
nmarton-da: This is just a cosmetic change to make it compile, all H2 support is broken at the moment.
There is a TODO capturing this aspect.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
nmarton-da: Purely Code style considerations: I cherish the threading style: easy and natural declaration of transformation steps with function literals. No need to introduce names for function parameters/function names/values. :)
If you disagree, or have a better approach please let me know!

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
nmarton-da: At this point they are more divulged contracts, as this is the indexing part...or would it be cleaner jut to drop ContractsInfo entirely and add this to the TransactionInfo, under the name divulgedEvents? That might be even more cleaner :)

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
nmarton-da: ha! even better so :)
done.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
nmarton-da: It is dead code, removed

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
meiersi-da: agreed

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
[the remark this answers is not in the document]
meiersi-da: Ah that might be due to sharing the datastructure with the GetTransactionTree call. Probably not worth the effort to separate.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -250,31 +226,33 @@ object TransactionIndexing {
...
-      netArchives: Set[ContractId],
-      netVisibility: WitnessRelation[ContractId],
+  // most involved change: divulgence is now an event instead encoded in contract/contract_witness tables
+  // also we generate these events directly as part of the other events, so we need to land the necessery
+  // data in the EventsTablePostgresql
+  case class DivulgedContractInfo(
+      contractId: ContractId,
+      contractInst: Option[
+        ContractInst
+      ], // this we do not necessarily have: for public KV ledgers the divulged contract list is empty, since all contracts must be at all ledger as create nodes already
```
meiersi-da, line 241:
```suggestion
      ], // this we do not necessarily have: for public KV ledgers and sanbox-classic the divulged contract list is empty, since all contracts must be at all ledger as create nodes already
```

meiersi-da (approved):
Thanks for the good work @nmarton-da ! This is a great blueprint for planning the production version of this feature.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
rautenrieth-da: We do not have a good story for handling incorrect data from the ReadService. The current index DB schema has some uniqueness constraints that prevent things like duplicate contract keys. These likely were inherited from sandbox classic times and never revisited.

Like you said, we should look over all index/indexer operations and come up with a design for how to handle corrupt upstream data on a best-effort basis.

rautenrieth-da (commented):
Nice work @nmarton-da ! I'll leave a few high level comments relevant for the real implementation here:
- The new queries are quite verbose, and use PostgreSQL specific syntax in a few places (e.g., `::text` or the declarative partitioning). We should carefully think about how to structure the code such that adding support for new RDBMS is easy, especially if that support is implemented in other repositories. Right now we use a combination of helper objects ([SqlFunctions.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala)), scattered matches on `DBType` (["dbType match"](https://github.com/digital-asset/daml/search?q=dbType+match)), and inheritance ([ContractsTableH2.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTableH2.scala) vs [ContractsTablePostgres.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTablePostgres.scala)). At least, try to reduce the amount of Postgres-specific features.
- This work along with #8990 will change many index DB related interfaces (e.g., `LedgerDao` won't need a `prepareTransactionInsert`). Since we are moving the entire code base towards parallel ingestion, we should take the opportunity to clean up all interfaces.

nmarton-da:
> Nice work @nmarton-da ! I'll leave a few high level comments relevant for the real implementation here:
> 
>     * The new queries are quite verbose, and use PostgreSQL specific syntax in a few places (e.g., `::text` or the declarative partitioning). We should carefully think about how to structure the code such that adding support for new RDBMS is easy, especially if that support is implemented in other repositories. Right now we use a combination of helper objects ([SqlFunctions.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala)), scattered matches on `DBType` (["dbType match"](https://github.com/digital-asset/daml/search?q=dbType+match)), and inheritance ([ContractsTableH2.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTableH2.scala) vs [ContractsTablePostgres.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTablePostgres.scala)). At least, try to reduce the amount of Postgres-specific features.
> 
>     * This work along with #8990 will change many index DB related interfaces (e.g., `LedgerDao` won't need a `prepareTransactionInsert`). Since we are moving the entire code base towards parallel ingestion, we should take the opportunity to clean up all interfaces.

100% agree, both of them were non-goals for PoC, and both of them should be goals for the real thing!

PR #8858, opened 2021-02-16
Release Scala 2.13 artifacts

cocreature:
I believe this is somewhat reasonable but as usual for these type of changes, I have no way of testing this and it’s a bunch of bash and yaml so it probably doesn’t work.

The most annoying part is the check if we can release 2.13 artifacts. I don’t see a way around that unfortunately.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build-unix.yml
+++ b/ci/build-unix.yml
@@ -5,7 +5,7 @@ parameters:
   is_release: ''
   name: ''
   release_tag: ''
-  scala_version: '2.12.12'
+  scala_version: ''
 
 steps:
   - bash: ci/dev-env-install.sh
@@ -50,7 +50,7 @@ steps:
       eval "$(./dev-env/bin/dade-assist)"
 
       EXTRA_FLAGS=""
-      if [[ "$DAML_SCALA_VERSION" != "2.12.12" ]]; then
+      if [[ -n "$DAML_SCALA_VERSION" ]]; then
         EXTRA_FLAGS="--only-scala"
       fi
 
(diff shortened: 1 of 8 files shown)


ci/build.yml
```diff
@@ -70,6 +70,47 @@ jobs:
         trigger_sha: '$(trigger_sha)'
     - template: report-end.yml
 
+- job: Linux_scala_2_13
```
cocreature, line 73: this is bad, this is going to fail for older releases. I’m not quite sure how to fix that unfortunately without starting to parse semvers in bash.

ci/build-unix.yml
```diff
@@ -5,6 +5,7 @@ parameters:
   is_release: ''
   name: ''
   release_tag: ''
+  scala_version: '2.12.12'
```
SamirTalwar-DA, line 8: I'd love to have this hard-coded in fewer places. Can we set it to the empty string by default? We might need to handle this in _scala_version.bzl_.

cocreature: Very valid point 9ea71dda2e5e7e59becd037149b96736f2f4bf9b

stefanobaghino-da (approved):

ci/build-unix.yml
SamirTalwar-DA: Yay!

SamirTalwar-DA (approved):
I haven't reviewed the YAML thoroughly because, well, I value my sanity. But I tried.

PR #8862, opened 2021-02-16
Upgrades opentelemetry to 0.16.0

lucb-da:
opentelemetry `0.8.0` is [not compatible](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1839) with io.grpc >= `1.33.0` (upgraded in #8558).
Upgrading to opentelemetry `0.16.0`.

### Pull Request Checklist

trigger the build.


cocreature:
Can we test this? I’d like to avoid breaking it again on the next grpc upgrade.

lucb-da:
It is not simple to test. Without the upgrade, it doesn't break. It just doesn't generate the expected data.
The gRPC support in OpenTelemetry relies on instrumentation. The instrumented code changed in such a way that the instrumentation was not active anymore.

It would be an integration test. Start one of the component with OpenTelemetry instrumentation enable, make a gRPC call, check that the traces looks right (the trace data can be sent to the log).

cocreature:
Fair enough, I’ll leave it up to you whether that is worth the effort.

cocreature (approved):
lgtm, thanks! See discussion for whether we should test this.

Needs rebasing and you also need to repin the json file for Scala 2.13.

lucb-da:
@fabiotudone-da Yes, I have run the changes multiple times in LR benchmarks.

The manually added spans are still there (they were not broken).
image

And the automatic gRPC spans are back.
image

lucb-da:
Rebased on `main`, and updated the `maven_install*` files.

fabiotudone-da:
@lucb-da I think the build wasn't run automatically because the PR is from a fork; it looks like you need some reformatting through `./fmt.sh`.

lucb-da:
Added changelog section, and squashed the commits.

Need a new run to be triggered.

PR #8863, opened 2021-02-16
Remap contract ids between on and off-ledger speedy

### LINKED ISSUE #8860: Match up display contract ids with Show instances in Daml Studio
Currently, we remap contract ids in Daml Studio in the transaction view to event ids to have something shorter and nicer. However, the `Show` instances in LF >= 1.11 expose the underlying contract ids which are super long and ugly.

The mismatch is pretty confusing and we should fix that. There are two options:

1. Fix the `Show` instance to match the current contract ids. We could either do this by always remapping when switching between on-ledger and off-ledger machine or we could make the behavior of `SBToTextContractId` configurable.
2. We could switch away from remapping and move towards static contract id seeding. This leads to slightly unweildly long contract ids so to combat that, we should probably combine this with git-style shortening to the shortest unambiguous string.

cocreature:
fixes #8860

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
@@ -219,7 +219,11 @@ class Context(val contextId: Context.ContextId, languageVersion: LanguageVersion
       case Success(v) =>
         Success(
           Some(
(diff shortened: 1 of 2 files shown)


cocreature:
This sort of works and I actually like the implementation but it has an issue that seems to severe to ignore: There is now a difference in the ordering on contract ids you get on ledger and off ledger. I hate that too much to go forward with this so closing.

PR #8872, opened 2021-02-17
Include --target flags in daml script dumps

cocreature:
As evidenced by #8856, daml script dumps currently fail to compile if
we generate a dump including templates which are newer than the
default LF version in the compiler. This PR addresses this by
including a --target flag in the generated daml.yaml.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
@@ -24,7 +24,7 @@ object LanguageVersion {
 
   private[lf] val All = Major.V1.supportedMinorVersions.map(LanguageVersion(Major.V1, _))
 
-  private[lf] val List(v1_6, v1_7, v1_8, v1_11, v1_12, v1_dev) = All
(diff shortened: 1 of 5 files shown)


daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
```diff
@@ -24,7 +24,7 @@ object LanguageVersion {
 
   private[lf] val All = Major.V1.supportedMinorVersions.map(LanguageVersion(Major.V1, _))
 
-  private[lf] val List(v1_6, v1_7, v1_8, v1_11, v1_12, v1_dev) = All
+  val List(v1_6, v1_7, v1_8, v1_11, v1_12, v1_dev) = All
```
cocreature, line 27: @remyhaemmerle-da I know you’re not a fan of making things public so if you prefer, I can revert this and duplicate the logic in the tests. That said, I think it is pretty reasonable in this case.

aherrmann-da (approved):

stefanobaghino-da (approved):

PR #8876, opened 2021-02-17
Non-Consuming Choice Subscriptions - Ledger API changes [DPP-253]

kamil-da:
**Motivation**
This is a draft of changes to the Ledger API and the event storing mechanism which enables including non-consuming `ExercisedEvent` events in the flat transactions stream.
The motivation behind this change is to enable consumers to access non-consuming exercise events with filtering, which enables efficiently using specific non-consuming choices as signals to Ledger API clients. An example use-case of this kind of signalling is the integration with an outbound off-ledger message queue.

**Key changes**
- add the `include_non_consuming_exercise_events` flag to  the `GetTransactionsRequest` Ledger API request - this flag enables streaming non-consuming exercise events in the flat transactions stream
- add the `ExercisedEvent exercised = 2;` event type in the Ledger API
- store witnesses of non-consuming exercise events in the `participant_events` table - this change allows to query non-consuming exercise events

**Additional comment**
Please see comments below for more detailed explanation of changes introduced with this draft.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
@@ -30,8 +30,8 @@ message Event {
     // This field was removed in https://github.com/digital-asset/daml/issues/960 and has been re-introduced
     // to make it possible for consumers to access exercised events without being forced to subscribe to
     // the expensive transaction trees stream.
-    // ExercisedEvent items emitted in the flat transactions stream are identical to items in transaction trees
-    // stream except for:
+    // Only non-consuming exercised events are emitted.
+    // They are identical to items in transaction trees stream except for:
     // - child_event_ids field being empty
     // - witness_parties field containing union of the signatories and actors
     ExercisedEvent exercised = 2;
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
@@ -13,8 +13,8 @@ option java_outer_classname = "EventOuterClass";
 option java_package = "com.daml.ledger.api.v1";
 option csharp_namespace = "Com.Daml.Ledger.Api.V1";
 
-// An event in the flat transaction stream can either be the creation
-// or the archiving of a contract.
+// An event in the flat transaction stream can be the creation of a contract,
+// the archiving of a contract or an exercising an action on a contract.
 //
 // In the transaction service the events are restricted to the events
 // visible for the parties specified in the transaction filter. Each
@@ -24,12 +24,20 @@ option csharp_namespace = "Com.Daml.Ledger.Api.V1";
 // that have witnesses.
 message Event {
   oneof event {
+
     CreatedEvent created = 1;
+
+    // This field was removed in https://github.com/digital-asset/daml/issues/960 and has been re-introduced
+    // to make it possible for consumers to access exercised events without being forced to subscribe to
+    // the expensive transaction trees stream.
+    // ExercisedEvent items emitted in the flat transactions stream are identical to items in transaction trees
+    // stream except for:
+    // - child_event_ids field being empty
+    // - witness_parties field containing union of the signatories and actors
+    ExercisedEvent exercised = 2;
+
     ArchivedEvent archived = 3;
  }
-  // see https://github.com/digital-asset/daml/issues/960
-  reserved 2;
-  reserved "exercised";
 }
 
 // Records that a contract has been created, and choices may now be exercised on it.
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/transaction_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/transaction_service.proto
@@ -115,6 +115,9 @@ message GetTransactionsRequest {
   // Optional
   TraceContext trace_context = 1000;
 
+  // If set to True, the flat transactions stream events will contain exercised events matching given filters.
+  // Optional
+  bool include_non_consuming_exercise_events = 6;
 }
 
 message GetTransactionsResponse {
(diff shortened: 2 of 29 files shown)


ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
```diff
@@ -24,12 +24,20 @@ option csharp_namespace = "Com.Daml.Ledger.Api.V1";
...
     CreatedEvent created = 1;
+
+    // This field was removed in https://github.com/digital-asset/daml/issues/960 and has been re-introduced
+    // to make it possible for consumers to access exercised events without being forced to subscribe to
+    // the expensive transaction trees stream.
+    // Only non-consuming exercised events are emitted.
+    // They are identical to items in transaction trees stream except for:
+    // - child_event_ids field being empty
+    // - witness_parties field containing union of the signatories and actors
+    ExercisedEvent exercised = 2;
```
kamil-da, line 37: Restoring `exercised` field. More details in the comment.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/transaction_service.proto
```diff
@@ -115,6 +115,9 @@ message GetTransactionsRequest {
   // Optional
   TraceContext trace_context = 1000;
 
+  // If set to True, the flat transactions stream events will contain exercised events matching given filters.
+  // Optional
+  bool include_non_consuming_exercise_events = 6;
```
kamil-da, line 120: This is an entry point for the feature - a flag in the `GetTransactionRequest`. This flag is then passed all the way down to the database queries to configure proper filtering of `participant_events` rows.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
```diff
@@ -43,30 +44,53 @@ private[events] object EventsTableFlatEvents {
...
-          commandId = commandId.getOrElse(""),
-          workflowId = workflowId.getOrElse(""),
-          event = Raw.FlatEvent.Archived(
-            eventId = eventId,
-            contractId = contractId,
-            templateId = templateId,
-            eventWitnesses = ArraySeq.unsafeWrapArray(eventWitnesses),
-          ),
-        )
+  private val exercisedFlatEventParser: RowParser[EventsTable.Entry[Raw.FlatEvent]] =
```
kamil-da, line 47: This is a common parser that can parse both `Archived` and non-consuming `Exercised` events. The differentiation is made by checking the value of `exercise_consuming` column from `participant_events` table.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEvents.scala
```diff
@@ -43,30 +44,53 @@ private[events] object EventsTableFlatEvents {
...
+              exerciseActors = ArraySeq.unsafeWrapArray(exerciseActors),
+              exerciseChildEventIds = ArraySeq.unsafeWrapArray(exerciseChildEventIds),
+              eventWitnesses = ArraySeq.unsafeWrapArray(eventWitnesses),
+            ),
+          )
     }
 
   val rawFlatEventParser: RowParser[EventsTable.Entry[Raw.FlatEvent]] =
-    createdFlatEventParser | archivedFlatEventParser
+    createdFlatEventParser | exercisedFlatEventParser
```
kamil-da, line 93: Replace a simple parser for `Archived` events with a more generic parser for all exercised events. This is required because we may receive both consuming and non-consuming event rows now.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -143,15 +155,33 @@ private[events] object EventsTableFlatEventsRangeQueries {
...
+      """and exercise_consuming IS NOT FALSE"""
+
+    private def filterOutNonConsumingExerciseEventsWhereClause(
+        config: Option[GetTransactionsConfig]
+    ) =
+      config
+        .collect {
+          case GetTransactionsConfig(include) if include => ""
+        }
+        .getOrElse(FilterOutNonConsumingExerciseEventsWhereClause)
```
kamil-da, line 178: Until now non-consuming events always had `flat_event_witnesses` set empty. This provided filtering out non-consuming events in the write path.
With this PR we intend to fill `flat_event_witnesses` column for both consuming and non-consuming exercise events with appropriate witnesses. This change requires introducing additional filtering to make the switch for including non-consuming events work as intended. To achieve that we add the `FilterOutNonConsumingExerciseEventsWhereClause` `WHERE` clause filter.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -178,11 +178,9 @@ object TransactionIndexing {
           addEventAndDisclosure(event)
           addVisibility(exercise.targetCoid, blinding.disclosure(nodeId))
           addDivulgence(exercise.targetCoid)
+          addStakeholders(nodeId, exercise.signatories ++ exercise.actingParties)
           if (exercise.consuming) {
-            addStakeholders(nodeId, exercise.stakeholders)
             addArchive(exercise.targetCoid)
-          } else {
-            addStakeholders(nodeId, Set.empty)
           }
```
kamil-da, line 184: This is the only change we need to make in the way we store exercised events to make the feature work.
We need to store witnesses not only for consuming events as we did until now but also for non-consuming ones. Otherwise querying non-consuming events would not be possible.

kamil-da (commented):
I've added a few comments to make the draft a bit more clear.

meiersi-da:
Thank you very much @kamil-da for this thorough investigation of the changes required for this feature. The changes and comments on the PR are very helpful! Good work!

As we discussed yesterday, this feature has value, but doesn't make the cut right now for the first round of performance improvements we target. Moreover, other team leads pointed out that changing the `GetTransactionTrees` stream with template-id filtering would be a much more clean API change than adding non-consuming exercise events to the `GetTransactions` stream. It's likely a bit trickier, but may be the right thing to do.

Given this evaluationg, I'm closing this PR for later reference once we engage in putting together a second round of performance improvements.

Thanks again for the good work!

PR #8879, opened 2021-02-17
use representational variance for generated Scala classes

### LINKED ISSUE #8844: Variance of type parameters in Scala codegen
Currently, the scala codegen generates all type parameters as [covariant](https://github.com/digital-asset/daml/blob/ff28059c635a71565b47ae80f121f8e3f923a23b/language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/DamlDataTypeGen.scala#L72). This worked fine so far: We only deal with serializable types so no way to use functions to change variance and all LF primitives had type parameters in covariant positions.

However, the introduction of GenMap changed this: We map this to `InsertOrdMap[K, V+]` which is invariant in `K`. This breaks code such as the following:

```
template Foobar
  with
    operator : Party
    things : MySet Text
  where
    signatory operator

data MySet k = MySet { unwrap : Map k () }
  deriving (Show, Eq)
```

```
bazel-out/k8-opt/bin/language-support/scala/codegen-sample-app/tmp6320987924609045777/com/daml/sample/mymain/MySet.scala:51: error: type mismatch;
 found   : com.daml.ledger.client.binding.Primitive.GenMap[?k1,com.daml.ledger.client.binding.Primitive.Unit] where type ?k1 <: k (this is a GADT skolem)
    (which expands to)  com.daml.lf.data.InsertOrdMap[?k1,Unit]
 required: com.daml.ledger.client.binding.Primitive.GenMap[k,com.daml.ledger.client.binding.Primitive.Unit]
    (which expands to)  com.daml.lf.data.InsertOrdMap[k,Unit]
Note: ?k1 <: k, but class InsertOrdMap is invariant in type K.
You may wish to define K as +K instead. (SLS 4.5)
                case _root_.com.daml.sample.MyMain.MySet(unwrap) => unwrap
                                                                    ^
com.daml.ledger.client.binding.Primitive.GenMap[?k1,com.daml.ledger.client.binding.Primitive.Unit] <: com.daml.ledger.client.binding.Primitive.GenMap[k,com.daml.ledger.client.binding.Primitive.Unit]?
false
```

I see two ways out of this:

1. Track variance of type parameters. This should work I believe but it’s annoyingly complex to implement (you need to track this through every type). However, it is non-breaking.
2. Make type parameters invariant everywhere. This also fixes the issue but at a relatively high cost: this will almost certainly break some code. We actually used to do this but switched away, see DEL-3214 for the reasoning.
3. Something else?

S11001001:
Fixes #8844, which calls for inference of variance when generating Scala classes.

We can classify our choice of variance as _representational_, which is the maximum "parametrically sound" variance [illustrated at the top of this article](https://typelevel.org/blog/2016/09/19/variance-phantom.html) typelevel/typelevel.github.com#116, or _nominal_, which is stricter.

Choosing a stricter variance admits more GADT usage, as that article goes on to demonstrate in several ways. However, because we do not allow GADTs to be defined in Daml, it seems to place a needless burden on "ordinary" usage of our Scala classes just to permit _ad hoc_ usage of those classes as GADTs in related Scala code. The maximal _representational_ choice is also the one closest to our current behavior (all covariant).

```rst
CHANGELOG_BEGIN
- [Scala Codegen] Type parameters used in GenMap keys are now supported.
  See `issue #8879 <https://github.com/digital-asset/daml/pull/8879>`__.
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.


language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+      import iface._
+
+      def lookupOrFail(i: I) = lookupType(i) getOrElse sys.error(s"$i not found")
+
+      def goType(dt: I, seen: Set[I], contexts: Map[I, Set[TVar]])(
+          typ: iface.Type
+      ): VarianceConstraint = typ match {
+        case TypeVar(name) =>
+          // while we default to Covariant at a later step,
+          // absence *does not* mean set-to-Covariant at this step
```
cocreature, line 81: absence of what?

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+              typArgs foldMap goType(dt, seen, contexts)
+          }
+
+        case TypeCon(TypeConName(tcName), typArgs) =>
+          val refDdt = lookupOrFail(tcName)
+          val argConstraints = refDdt.typeVars zip typArgs foldMap { case (tv, ta) =>
+            goType(dt, seen, contexts |+| Map(tcName -> Set(tv)))(ta)
+          }
+          val refConstraints =
+            if (seen(tcName)) mzero[VarianceConstraint] else goSdt(tcName, seen + tcName)(refDdt)
```
cocreature, line 111: I don’t understand how the `seen` stuff works. This doesn’t actually stop you from calling `goSdt` multiple times on a type, e.g., you could call `goType` once per type arg in line `102` since you don’t update `seen` across calls to `goType`. Is that intentional, if so why?

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
     case iface.TypeCon(_, xs) => xs.toSet.flatMap(collectTypeParams)
     case iface.TypeNumeric(_) => Set.empty
   }
+
+  final case class LookupType[+RF, +VF](
+      run: iface.TypeConName => Option[(iface.DefDataType[RF, VF], LookupType[RF, VF])]
+  )
+
+  import VarianceConstraint.BaseResolution
+  import ScopedDataType.{Name => I}
```
cocreature, line 46: This is now renaming the type twice. `ScopedDataType.Name` is already an alias for `Identifier`. Can we just use `Identifier`. While I can probably guess that `I` is an identifier, I’d rather not have to guess.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/LFUtil.scala
```diff
@@ -266,6 +266,16 @@ final case class LFUtil(
       case _ => false
     }
   }
+
+  protected[this] override def precacheVariance(interface: Interface) = {
+    import UsedTypeParams.ResolvedVariance
+    val resolved = interface.typeDecls.foldLeft(ResolvedVariance.Empty) {
+      case (resolved, (id, InterfaceType.Template(_, _))) =>
+        resolved.allCovariantVars(id, interface)._1
```
cocreature, line 274: I don’t understand why we can only recurse on templates. Doesn’t the same problem apply to choice types?

cocreature (approved):
Nice work! A few minor comments and questions

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+  import Variance._
+  private[this] type TVar = Ref.Name
+
+  final class ResolvedVariance private (private val prior: Map[I, ImmArraySeq[Variance]]) {
+    def allCovariantVars(
+        dt: I,
+        ei: iface.EnvironmentInterface,
+    ): (ResolvedVariance, ImmArraySeq[Variance]) = {
```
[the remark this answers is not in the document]
S11001001: Not quite. a2ba2bb38b1

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+        resolved.prior.getOrElse(
+          dt,
+          ei.typeDecls.getOrElse(dt, sys.error(s"$dt not found")).`type`.typeVars map (_ =>
+            Covariant
+          ),
+        ),
+      )
+    }
+
+    /** Variance of `sdt.typeVars` in order. */
```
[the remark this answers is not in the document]
S11001001: a2ba2bb38b1

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+    private[this] def covariantVars[RF <: iface.Type, VF <: iface.Type](
+        dt: I,
+        lookupType: I => Option[iface.DefDataType[RF, VF]],
+    ): ResolvedVariance = {
+      import iface._
+
+      def lookupOrFail(i: I) = lookupType(i) getOrElse sys.error(s"$i not found")
```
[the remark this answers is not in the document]
S11001001: I think this error is far more likely to imply an invalid `EnvironmentInterface` (i.e. a bad LF) because every `dt` should have come from the `EnvironmentInterface`, and every reference within a data type therein should be resolvable.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+            goType(dt, seen, contexts |+| Map(tcName -> Set(tv)))(ta)
+          }
+          val refConstraints =
+            if (seen(tcName)) mzero[VarianceConstraint] else goSdt(tcName, seen + tcName)(refDdt)
+          argConstraints |+| refConstraints
+
+        case TypeNumeric(_) => VarianceConstraint.base(Map.empty)
+      }
+
+      def goSdt(dt: I, seen: Set[I])(sdt: DefDataType[RF, VF]): VarianceConstraint =
```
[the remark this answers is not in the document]
S11001001: They should, but the caller in the recursive case also needs the information, and I'd rather minimize the `lookupOrFail` points.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+          .map(VarianceConstraint.reresolve(dt, sdt, _))
+          .getOrElse {
+            val vLookup = sdt.dataType match {
+              case Record(fields) =>
+                fields foldMap { case (_, typ) => goType(dt, seen, Map.empty)(typ) }
+              case Variant(fields) =>
+                fields foldMap { case (_, typ) => goType(dt, seen, Map.empty)(typ) }
+              case Enum(_) => mzero[VarianceConstraint]
+            }
+            vLookup
```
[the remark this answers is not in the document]
S11001001: c6aa4dfb462

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+    /** Variance of `sdt.typeVars` in order. */
+    private[this] def covariantVars[RF <: iface.Type, VF <: iface.Type](
+        dt: I,
+        lookupType: I => Option[iface.DefDataType[RF, VF]],
+    ): ResolvedVariance = {
+      import iface._
+
+      def lookupOrFail(i: I) = lookupType(i) getOrElse sys.error(s"$i not found")
+
+      def goType(dt: I, seen: Set[I], contexts: Map[I, Set[TVar]])(
```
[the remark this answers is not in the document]
S11001001: We can capture that and a few more invariants that way; nice idea. ae0412d7d4d

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
S11001001: e63eb6df09f

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
S11001001: `seen` _only_ prevents cycles in `goSdt`.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
S11001001: 567f98a377f

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/LFUtil.scala
S11001001: We only need to reach for the "roots" here; `allCovariantVars` itself will recur into referenced datatypes. Nor do we need to be perfect here; if we missed any types, they'll just be resolved later when we ask them to be resolved. We are only looking to precache the variances we're actually going to need, while not resolving tons of variances we won't need.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/UsedTypeParams.scala
```diff
@@ -26,10 +31,191 @@ object UsedTypeParams {
...
+    def allCovariantVars(
+        dt: I,
+        ei: iface.EnvironmentInterface,
+    ): (ResolvedVariance, ImmArraySeq[Variance]) = {
+      val resolved = covariantVars(dt, (i: I) => ei.typeDecls get i map (_.`type`))
+      (
+        resolved,
+        resolved.prior.getOrElse(
+          dt,
+          ei.typeDecls.getOrElse(dt, sys.error(s"$dt not found")).`type`.typeVars map (_ =>
```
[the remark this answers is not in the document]
S11001001: If the type doesn't have any type variables that are referenced in its definition, it won't be in `prior`.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/lf/LFUtil.scala
cocreature: But the record template type doesn’t need to reference choice argument or return types. What makes sure we hit them?

S11001001: _Oh_, I read "choice types" as "variant types" and just reacted internally "well that shouldn't make a difference". It's little problem to add choice and key types here, though it's little harm to forget them either.

cocreature: Why is there no harm in forgetting them? Doesn’t that mean we get variance of those types wrong?

S11001001: (As discussed out-of-band) it's a precache, so we just compute later if we miss it. Anyway, 53940cd3696 adds choice and key types to the precache.

PR #8883, opened 2021-02-17
Let migrations not explode memory

gerolf-da:
This is a slightly spicy one, as it touches existing scala DB migrations.

This doesn't cause migration hash conflicts in flyway for existing databases, because java/scala migrations don't have an associated hash.

When running a recent version of daml on sql on a database from last August 2020, I noticed two issues:
1. ValueSerializer ran into a match error for values that weren't deprecated (fixed by 6675887).
2. The database had a bit over 2 million rows in `participant_events`. Without setting the fetch size explicitly, the JDBC driver (un)happily tries to load the entire result set into memory, eventually crashing the participant with an OutOfMemoryError (fixed by 075326a).

I initially suspected the memory explosion in the V38 migration to come from the assembly of the `LazyList` value (`Stream` often caused similar issues by holding on to values too long), so I changed it to `Iterator`. I then kept the `Iterator` based approach, as reasoning about what is kept in memory is easier with `Iterator` than with `Stream`/`LazyList` imho.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/db/migration/postgres/V32_1__Fix_key_hashes.scala
+++ b/ledger/participant-integration-api/src/main/scala/db/migration/postgres/V32_1__Fix_key_hashes.scala
@@ -16,6 +16,8 @@ private[migration] final class V32_1__Fix_key_hashes extends BaseJavaMigration {
   private val FIX_HASH =
     "update participant_contracts set create_key_hash = ? where contract_id = ?"
 
+  private val BATCH_SIZE = 500
+
   override def migrate(context: Context): Unit = {
     val conn = context.getConnection
     var selectKeys: java.sql.Statement = null
@@ -25,7 +27,10 @@ private[migration] final class V32_1__Fix_key_hashes extends BaseJavaMigration {
       fixHash = conn.prepareStatement(FIX_HASH)
 
       selectKeys = conn.createStatement()
+      conn.setFetchSize(BATCH_SIZE)
       keysRows = selectKeys.executeQuery(SELECT_KEYS)
+
+      var currentBatchSize = 0
 
       while (keysRows.next()) {
         val contractId = keysRows.getString("contract_id")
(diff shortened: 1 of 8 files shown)


SamirTalwar-DA (approved):
This looks pretty good to me. I haven't checked that you caught all of them; I figure your large database dump would have helped with that.

ledger/participant-integration-api/src/main/scala/db/migration/postgres/V32_1__Fix_key_hashes.scala
```diff
@@ -25,8 +27,11 @@ private[migration] final class V32_1__Fix_key_hashes extends BaseJavaMigration {
       fixHash = conn.prepareStatement(FIX_HASH)
 
       selectKeys = conn.createStatement()
+      conn.setFetchSize(BATCH_SIZE)
```
SamirTalwar-DA, line 30: I think you meant to use `selectKeys.setFetchSize` here.

gerolf-da: yeah, thanks. 🤦

ledger/participant-integration-api/src/main/scala/db/migration/translation/ValueSerializer.scala
```diff
@@ -39,6 +39,7 @@ private[migration] object ValueSerializer {
         logger.error(
           s"*** Please upgrade your sandbox database by upgrading your SDK to 1.6 first. ***"
         )
+      case _ =>
```
gerolf-da, line 42: @remyhaemmerle-da: When running this migration on an index DB from last August, this match failed on the `Right` side. I only could get it to pass after add this wildcard match. Would this match your expectation of how this should work?

remyhaemmerle-da: That is the only think that make sens.  This should need to be backported, shouldn't  it ?

remyhaemmerle-da: I propose to simplify this unnecessary complexity of mine:

```scala

  private[this] val DeprecatedValueVersions = Set("1", "2", "3", "4", "5")
  private[this] val UnsupportedErrorMessage = """Unsupported transaction version '(\d)'""".r

  private[translation] def handleDeprecatedValueVersions[X](x: Either[ValueCoder.DecodeError, X]): Either[ValueCoder.DecodeError, X] = {
    x match {
      case Left(ValueCoder.DecodeError(UnsupportedErrorMessage(version))) if DeprecatedValueVersions(version) =>
        logger.error(
          s"*** Deserialization of value version $version is not supported by the SDK 1.7.0 or later. ***"
        )
        logger.error(
          s"*** Please upgrade your sandbox database by upgrading your SDK to 1.6 first. ***"
        )
      case _ => 
    }
    x
  }
```

PR #8885, opened 2021-02-17
Stop pretending strings are booleans

cocreature:
Sorry for all the mess here. I’m not capable of programming in yaml.

It turns out is_release is a string not a boolean and

```
and('false', eq('true', 'true'))
```

is true.

I hate everything about this.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -110,7 +110,7 @@ jobs:
         ./bazel-bin/release/release --release-dir "$(mktemp -d)"
       env:
         DAML_SCALA_VERSION: "2.13.3"
-      name: Maven check
+      name: maven_check
       condition: and(succeeded(), ne(variables['is_release'], 'true'))
     - template: tell-slack-failed.yml
       parameters:
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -109,7 +109,7 @@ jobs:
         bazel build //release:release
         ./bazel-bin/release/release --release-dir "$(mktemp -d)"
       env:
-        DAML_SCALA_VERSION: ${{parameters.scala_version}}
+        DAML_SCALA_VERSION: "2.13.3"
       name: Maven check
       condition: and(succeeded(), ne(variables['is_release'], 'true'))
     - template: tell-slack-failed.yml
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -82,7 +82,7 @@ jobs:
     # The easiest option seems to be to still run the Scala 2.13 job but
     # pretend it’s not a release build and thereby just run it against main
     # so that’s what we do here.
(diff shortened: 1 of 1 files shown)


cocreature:
fun fact: the variable will then be converted from a boolean to a string again so within the job you don’t have a boolean either.

stefanobaghino-da (approved):
`yamlConfuses: 'true'`

cocreature:
Check looks better now
```
  is_release:
    Parsing expression: <and(eq(dependencies.check_for_release.outputs['out.is_release'], 'true'), eq(dependencies.check_for_release.outputs['out.scala_2_13'], 'true'))>
    Evaluating: and(eq(dependencies['check_for_release']['outputs']['out.is_release'], 'true'), eq(dependencies['check_for_release']['outputs']['out.scala_2_13'], 'true'))
    Expanded: and(eq('false', 'true'), eq(dependencies['check_for_release']['outputs']['out.scala_2_13'], 'true'))
    Result: 'False'
```

ci/build.yml
```diff
@@ -109,7 +109,8 @@ jobs:
         bazel build //release:release
         ./bazel-bin/release/release --release-dir "$(mktemp -d)"
       env:
-        DAML_SCALA_VERSION: ${{parameters.scala_version}}
+        DAML_SCALA_VERSION: "2.13.3"
```
cocreature, line 112: I pushed another fix here since the step which now triggers after fixing the check didn’t build against 2.13.

ci/build.yml
```diff
@@ -82,7 +82,7 @@ jobs:
     # The easiest option seems to be to still run the Scala 2.13 job but
     # pretend it’s not a release build and thereby just run it against main
     # so that’s what we do here.
-    is_release: $[ and(dependencies.check_for_release.outputs['out.is_release'], eq(dependencies.check_for_release.outputs['out.scala_2_13'], 'true')) ]
+    is_release: $[ and(eq(dependencies.check_for_release.outputs['out.is_release'], 'true'), eq(dependencies.check_for_release.outputs['out.scala_2_13'], 'true')) ]
```
SamirTalwar-DA, line 85: Ew.

ci/build.yml
SamirTalwar-DA: Can you add a comment explaining this in the file?

cocreature: explaining what? that referencing a non-existent variable doesn’t work? that seems a bit pointless.

SamirTalwar-DA: Nah, just explaining that this is a duplicate of another constant elsewhere and that they need to be kept in sync. (And in the other place.)

PR #8898, opened 2021-02-19
Support starting script dumps from a given offset

### LINKED ISSUE #8771: Support starting script dumps from a given offset
Currently, daml script dumps start from ledger begin. This is obviously not very scalable. Instead we should have an option to start from a given ledger offset. In that case the script dump would first restore the ACS at that offset and then continue reading the transaction stream. Note that the ACS needs to be topologically ordered for reconstruction to handle the case where one contract references the contract id of another contract. There is an issue here around divulged contract which are not included in the ACS. For now, we should just emit an error on that until the ledger API provides the necessary information to handle this.

As a further optimization, we could consider only restoring the parts of the ACS that are referenced later in the transaction stream. I think there are valid usecases for both (at least for ledger migrations you want the full ACS, for debugging parts are probably sufficient) so we should support both.

aherrmann-da:
Closes #8771

Before this offset we reproduce the ACS with simple `createCmd` command submissions.

This current approach does not support contracts in the ACS that reference other contracts outside the ACS, e.g. archived contracts.

Adds a unit test for encoding single creates to reconstruct the ACS.
Extends the integration test to test at different offsets.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
+++ b/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
@@ -16,7 +16,7 @@ import com.daml.lf.data.Ref.PackageId
 import com.daml.lf.engine.script.{GrpcLedgerClient, Participants, Runner, ScriptTimeMode}
 import com.daml.ledger.api.domain
 import com.daml.ledger.api.refinements.ApiTypes.ApplicationId
-import com.daml.ledger.api.testing.utils.{AkkaBeforeAndAfterAll, SuiteResourceManagementAroundEach}
+import com.daml.ledger.api.testing.utils.{AkkaBeforeAndAfterAll, SuiteResourceManagementAroundAll}
 import com.daml.ledger.api.v1.command_service.SubmitAndWaitRequest
 import com.daml.ledger.api.v1.commands._
 import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
@@ -49,7 +49,8 @@ final class IT
     extends AsyncFreeSpec
     with Matchers
     with AkkaBeforeAndAfterAll
-    with SuiteResourceManagementAroundEach
+    with BeforeAndAfterEach
+    with SuiteResourceManagementAroundAll
     with SandboxNextFixture
     with TestCommands {
   private val appId = domain.ApplicationId("script-dump")
@@ -63,15 +64,19 @@ final class IT
   val isWindows: Boolean = sys.props("os.name").toLowerCase.contains("windows")
   val exe = if (isWindows) { ".exe" }
   else ""
-  private val tmpDir = Files.createTempDirectory("script_dump")
+  private var tmpDir: Path = null
   private val damlc =
     BazelRunfiles.requiredResource(s"compiler/damlc/damlc$exe")
   private val damlScriptLib = BazelRunfiles.requiredResource("daml-script/daml/daml-script.dar")
   private def iouId(s: String) =
     api.Identifier(packageId, moduleName = "Iou", s)
 
-  override protected def afterAll(): Unit = {
-    super.afterAll()
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    tmpDir = Files.createTempDirectory("script_dump")
+  }
+  override protected def afterEach(): Unit = {
+    super.afterEach()
     deleteRecursively(tmpDir)
   }
 
--- a/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
+++ b/daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
@@ -4,7 +4,7 @@
 package com.daml.script.dump
 
 import java.io.IOException
-import java.nio.file.{Files, FileVisitResult, Path, SimpleFileVisitor}
+import java.nio.file.{FileVisitResult, Files, Path, SimpleFileVisitor}
 import java.nio.file.attribute.BasicFileAttributes
 import java.util.UUID
 
@@ -128,15 +128,26 @@ final class IT
       .range(0, numParties)
       .traverse(_ => client.partyManagementClient.allocateParty(None, None).map(_.party))
 
-  private def createScriptDump(parties: List[Ref.Party]): Future[Dar[(PackageId, Package)]] = for {
+  private val ledgerBegin = LedgerOffset(
+    LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)
+  )
+  private val ledgerEnd = LedgerOffset(
+    LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_END)
+  )
+  private def ledgerOffset(offset: String) = LedgerOffset(LedgerOffset.Value.Absolute(offset))
+
+  private def createScriptDump(
+      parties: List[Ref.Party],
+      offset: LedgerOffset,
+  ): Future[Dar[(PackageId, Package)]] = for {
     // build script dump
     _ <- Main.run(
       Config(
         ledgerHost = "localhost",
         ledgerPort = serverPort.value,
         parties = parties,
-        start = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)),
-        end = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_END)),
+        start = offset,
+        end = ledgerEnd,
         outputPath = tmpDir,
         damlScriptLib = damlScriptLib.toString,
         sdkVersion = SdkVersion.sdkVersion,
@@ -176,8 +187,9 @@ final class IT
     )
   } yield ()
 
-  private def test(
-      numParties: Int
+  private def testOffset(
+      numParties: Int,
+      skip: Int,
   )(f: (LedgerClient, Seq[Ref.Party]) => Future[Unit]): Future[Assertion] =
     for {
       client <- LedgerClient(channel, clientConfiguration)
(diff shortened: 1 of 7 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
```diff
@@ -19,6 +23,15 @@ object Config {
   def parse(args: Array[String]): Option[Config] =
     parser.parse(args, Empty)
 
+  private def parseLedgerOffset(s: String): LedgerOffset = LedgerOffset {
```
cocreature, line 26: I assume you checked that we don’t have this anywhere else? I think extractor accepts offsets as well but from a quick look it seems like it doesn’t quite have the need for this (it only accepts absolute offsets).

daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
```diff
@@ -19,6 +23,15 @@ object Config {
...
+    s match {
+      case s if s == "start" =>
```
cocreature, line 28:
```suggestion
      case s if s == "begin" =>
```
any reason not to match the name on the API here?

daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
```diff
@@ -30,6 +43,18 @@ object Config {
...
+      .optional()
+      .action((x, c) => c.copy(start = parseLedgerOffset(x)))
+      .text(
+        "The transaction offset (exclusive) for the start position of the dump. Optional, by default the dump includes the beginning of the ledger."
+      )
+    opt[String]("end")
+      .optional()
+      .action((x, c) => c.copy(start = parseLedgerOffset(x)))
+      .text(
+        "The transaction offset (inclusive) for the end position of the dump. Optional, by default the dump includes the current end of the ledger."
```
cocreature, line 56: Do we have the invariant that `--start=x --end=x` always produces the acs dump at the given offset?

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
```diff
@@ -57,29 +60,68 @@ object Main {
     ()
   }
 
+  private def getACS(
```
cocreature, line 63: We should move the logic out of `Main` into some more sensible places in separate PRs.

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
```diff
@@ -57,29 +60,68 @@ object Main {
...
+      parties: List[String],
+      offset: LedgerOffset,
+  )(implicit
+      mat: Materializer
+  ): Future[Map[String, CreatedEvent]] = {
+    val ledgerBegin = LedgerOffset(
+      LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)
+    )
+    if (offset == ledgerBegin) {
+      Future.successful(Map.empty)
```
cocreature, line 74: This is purely an optimization right? The code below looks like it’ll work without this.

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
```diff
@@ -57,29 +60,68 @@ object Main {
     ()
   }
 
+  private def getACS(
+      client: LedgerClient,
+      parties: List[String],
+      offset: LedgerOffset,
+  )(implicit
+      mat: Materializer
+  ): Future[Map[String, CreatedEvent]] = {
```
cocreature, line 69: Maybe use `com.daml.ledger.api.refinements.ContractId` instead of a string?

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
```diff
@@ -57,29 +60,68 @@ object Main {
...
-      pkgRefs: Set[PackageId] = trees.toList
+      acs <- getACS(client, config.parties, config.start)
+      trees <- getTransactionTrees(client, config.parties, config.start, config.end)
+      acsPkgRefs = acs.values.toList
+        .foldMap(ev => valueRefs(Sum.Record(ev.getCreateArguments)))
+        .map(i => PackageId.assertFromString(i.packageId))
+      treePkgRefs = trees.toList
         .foldMap(treeRefs(_))
         .map(i => PackageId.assertFromString(i.packageId))
+      pkgRefs = acsPkgRefs ++ treePkgRefs
```
cocreature, line 118: Any reason not to work with sets here?

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -15,6 +17,21 @@ import scalaz.syntax.foldable._
...
+  /** Sort the active contract set topologically,
+    *  such that a contract at a given position in the list
+    *  is only referenced by other contracts later in the list.
+    */
+  def topoSortAcs(acs: Map[String, CreatedEvent]): List[CreatedEvent] = {
+    val graph: Graphs.Graph[String] = acs.mapValues(createdReferencedCids)
+    Graphs
+      .topoSort(graph)
+      .left
+      .map(msg => new RuntimeException(s"Encountered cyclic contract dependencies: $msg"))
```
cocreature, line 29: maybe throw an IllegalArgumentException instead? probably also worth documenting the throwing behavior in the scaladoc comment.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -15,6 +17,21 @@ import scalaz.syntax.foldable._
...
+      .toTry
```
cocreature, line 30: this seems a bit odd, you first handle the `Left` case by throwing to an exception but then you convert it to a `Try` discarding the information that you already handled the `Left` and call a partial `Left`. Maybe a a pattern match or a fold would be more appropriate?

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -15,6 +17,21 @@ import scalaz.syntax.foldable._
 object TreeUtils {
   final case class Selector(i: Int)
 
+  /** Sort the active contract set topologically,
+    *  such that a contract at a given position in the list
+    *  is only referenced by other contracts later in the list.
+    */
+  def topoSortAcs(acs: Map[String, CreatedEvent]): List[CreatedEvent] = {
+    val graph: Graphs.Graph[String] = acs.mapValues(createdReferencedCids)
```
cocreature, line 25:
```suggestion
    val graph: Graphs.Graph[String] = acs.view.mapValues(createdReferencedCids).toMap
```
`mapValues` is lazy and recomputes the function everytime. On Scala 2.13, this should also fail to compile (with a deprecation warning).

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -33,6 +50,14 @@ object TreeUtils {
     }
   }
 
+  def partiesInContracts(contracts: Iterable[CreatedEvent]): Set[String] = {
+    var parties: Set[String] = Set()
+    contracts.foreach { value =>
```
cocreature, line 55: looks like a `foldMap` :slightly_smiling_face:

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -18,13 +19,34 @@ import scalaz.std.set._
...
-    val cids = trees.map(treeCreatedCids(_))
-    val cidRefs = trees.toList.foldMap(treeReferencedCids(_))
-    val cidMap = cidMapping(cids, cidRefs)
-    val refs = trees.toList.foldMap(treeRefs(_))
+
+    val acsCidRefs = acs.values.toList.foldMap(createdReferencedCids)
+    val treeCidRefs = trees.toList.foldMap(treeReferencedCids)
+    val cidRefs = acsCidRefs ++ treeCidRefs
+
+    val archivedCidRefs = acsCidRefs -- acs.keySet
```
cocreature, line 33: Given that these are not just archived contracts but also divulged contracts or simply unknown contracts (you can pass in an arbitrary string via the ledger API), maybe call this `unknownCidRifs` or `missingCidRefs`.

daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
```diff
@@ -118,63 +123,98 @@ final class IT
...
+      // reproduce from dump
+      dar <- createScriptDump(parties, offset)
+      newParties <- allocateParties(client, numParties)
+      _ <- runScriptDump(client, newParties, dar)
       // check that the new transaction trees are the same
       after <- collectTrees(client, newParties)
+      afterCmp = after.drop(after.length - beforeCmp.length)
     } yield {
-      TransactionEq.equivalent(before, after).fold(fail(_), _ => succeed)
+      TransactionEq.equivalent(beforeCmp, afterCmp).fold(fail(_), _ => succeed)
```
cocreature, line 213: Is it worth comparing the ACS before and after at `skip` as well?

cocreature (approved):
Nice work, thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
aherrmann-da: Yes, if the range is going to be empty, then there is no need to issue an API call. The same holds in `getTransactionTrees`. I've added it there as well for consistency.

daml-script/dump/src/main/scala/com/daml/script/dump/Main.scala
aherrmann-da: Not sure I follow, they are `Set[PackageId]`.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Yes, CI caught it. Fixed.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Not quite, the `.left.map` converts the left type to an exception type and `toTry.get` then throws that in the left case. However, I agree it's too much acrobatics. I've changed it to a simple match.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Yup, this was a copy paste from `partiesInTree`. I've changed it a `foldMap`.

daml-script/dump/src/it/scala/com/daml/script/dump/IT.scala
aherrmann-da: Yes, that would be a good check to have. However, unless I'm missing something, this would require a bit more infrastructure. In `before` the ACS up to the offset is created by the original create and exercise commands in their original order. In `after` it is created by plain creates in some topological ordering based on contract dependencies. There is no guarantee that the contracts are created in the same order. I'm not sure that current unification algorithm can handle this.

PR #8899, opened 2021-02-19
BigNumeric spec.

remyhaemmerle-da:
### Pull Request Checklist

trigger the build.


daml-lf/spec/daml-lf-1.rst
```diff
@@ -3662,40 +3665,152 @@ Int64 functions
...
+  be mapped into a ``BigDecimal`` without loss of precision, returns
+  ``None``.
+
+* ``TO_TEXT_BIGDECIMAL : 'BigDecimal' → 'Text'``
+
+  Returns the numeric string representation of the BigDecimal. The result
+  will be returned at the smallest precision that can represent the result exactly, i.e.,
+  without any trailing zeroes.
+
+* ``TO_NUMERIC_BIGDECIMAL : ∀ (α : nat). 'RoundingMode' → 'BigDecimal'  → 'Numeric' α``
```
remyhaemmerle-da, line 3755:
```suggestion
* ``TO_NUMERIC_BIGDECIMAL : ∀ (α : nat).  'BigDecimal'  → 'Numeric' α``
```
I would either use the default Numeric rounding (i.e `HalfEven`) or throw exception (i.e. `NoRoundingNecessary`)
If people need another rounding they still can round before.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3662,40 +3665,152 @@ Int64 functions
...
+MathContext functions
+~~~~~~~~~~~~~~~~~~~~~
+
+*  ``'MATHCTX_UNLIMITED' : MathContext``
+
+   Returns `'UnBondedMathContext'`
+
+* ``MATHCTX_BOUNDED' : RoundingMode -> Int64 -> MathContext``
+
+   Returns ``'BoundedMathContext' v₁ v₂`` where ``v₁`` and ``v₂`` respectively the first and second argument.
```
sofiafaro-da, line 3677: We don't need these functions if we have math context literals.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -602,6 +613,9 @@ Then we can define our kinds, types, and expressions::
        |  LitTimestamp                              -- ExpLitTimestamp: UTC timestamp literal
        |  LitParty                                  -- ExpLitParty: Party literal
        |  cid                                       -- ExpLitContractId: Contract identifiers
+       |  LitRoundingMode                           -- ExpLitRoundingMode: Rounding Mode
+       |  UnBondedMathContext                       -- ExpUnboundedMathContext: Unlimited Math Context
```
sofiafaro-da, line 617:
```suggestion
       |  UnBoundedMathContext                       -- ExpUnboundedMathContext: Unlimited Math Context
```

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da, line 3677: I think a function  ``NUMERIC_MATHCTX_BOUNDED' : ∀ (α : nat). RoundingMod -> MathContext``  can be Handy if we want to write library for numeric.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3662,40 +3665,152 @@ Int64 functions
...
+  wrapped in ``Some``. If the input does not match the regexp
+  ``[+-]?\d+(\.d+)?`` or if the result of the conversion cannot
+  be mapped into a ``BigDecimal`` without loss of precision, returns
+  ``None``.
+
+* ``TO_TEXT_BIGDECIMAL : 'BigDecimal' → 'Text'``
+
+  Returns the numeric string representation of the BigDecimal. The result
+  will be returned at the smallest precision that can represent the result exactly, i.e.,
+  without any trailing zeroes.
```
remyhaemmerle-da, line 3753: Do we really need those functions. Cannot we use the one from numeric.
This are tricky to implement, test  and open the door for attack.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -1914,37 +1938,16 @@ need to be evaluated further. ::
...
-     ⊢ᵥ  'ROUNDING_HALF_UP'
+   ——————————————————————————————————————————————————— ValUnBondedMathContext
+     ⊢ᵥ  'UnBondedMathContext'
 
-   ——————————————————————————————————————————————————— ValRoundingModeHalf_Unnecessary
-     ⊢ᵥ  'ROUNDING_HALF_UNNECESSARY'
+     ⊢ᵥ  e₁
+     ⊢ᵥ  e₂
+   ——————————————————————————————————————————————————— ValUnBondedMathContext
+     ⊢ᵥ  'BoundedMathContext ' e₁ e₂
```
sofiafaro-da, line 1950: We also need evaluation semantics rules for the various literals.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3662,40 +3665,152 @@ Int64 functions
   in ``Some``. If the input does not match the regexp ``[+-]?\d+`` or
   if the result of the conversion overflows, returns ``None``.
 
+MathContext functions
```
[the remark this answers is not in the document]
remyhaemmerle-da: reshuffled.

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: We need it, if we do not want to create a expression for MathContext in the proto.
I am fine with both.

cocreature: Why do we need that in daml-lf instead of keeping it as a pure Daml thing?

daml-lf/spec/daml-lf-1.rst
cocreature: Great idea, we can definitely use `round` before. Cherrypicked it to my branch f42c6205188467ad10221e4f43b13148e3c3e8fc

daml-lf/spec/daml-lf-1.rst
cocreature: Removed the read function in 223733b7f7eff2003adeb649625ec54b6aa8d7ac, going via Numeric seems perfectly acceptable here. I kept the write. I think that is too important for debugging to drop it.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3667,35 +3670,51 @@ Numeric functions
...
-  the closest multiple of ``10⁻ᵅ`` using `banker's rounding convention
-  <https://en.wikipedia.org/wiki/Rounding#Round_half_to_even>`_ (where
-  `n` is given as the type parameter).  The type parameters `α₁`,
-  `α₂`, `α` define the scale of the first input, the second input, and
-  the output, respectively. Throws an error in case of overflow.
+      'DIV_NUMERIC' ≡
+        Λ α₁ : ⋆.  Λ α₂ : ⋆. Λ α : ⋆. λ x : 'Numeric' α₁ . λ : 'Numeric' α₂ .
+	        let x' = 'TO_BIGDECIMAL_NUMERIC' α₁ 'UnBondedMathContext' x in
+            let y' = 'TO_BIGDECIMAL_NUMERIC' α₂ 'UnBondedMathContext' y in
+            let r  = ``DIV_BIGDECIMAL`` 'UnBondedMathContext' x' y' in
```
cocreature, line 3711: this is not correct. Unbounded throws if the result cannot be represented exactly. You need to round here to match semantics of Numeric.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3667,35 +3670,51 @@ Numeric functions
...
+            'TO_NUMERIC_BIGDECIMAL' α RoundingHalfEven r
```
cocreature, line 3712: Needs to be adapted to the removal of the rounding mode argument.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3667,35 +3670,51 @@ Numeric functions
...
 
 * ``DIV_NUMERIC : ∀ (α₁ α₂ α : nat) . 'Numeric' α₁ → 'Numeric' α₂ → 'Numeric' α``
 
-  Divides the first decimal by the second one and rounds the result to
-  the closest multiple of ``10⁻ᵅ`` using `banker's rounding convention
-  <https://en.wikipedia.org/wiki/Rounding#Round_half_to_even>`_ (where
-  `n` is given as the type parameter).  The type parameters `α₁`,
-  `α₂`, `α` define the scale of the first input, the second input, and
-  the output, respectively. Throws an error in case of overflow.
+      'DIV_NUMERIC' ≡
```
cocreature, line 3707: Actually, I’m now not sure if this works at all. BigDecimal rounds to a given number of non-zero digits but bigdecimal has a floating scale so it does not round to a given number of decimals.

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: Hmm. If you don't want to add a protobuf expression, it sounds like keeping the math context only as built-ins would be more useful? It's not nice for the LF spec to have an expression form with typing rules, etc but you can't read/write in protobuf.

cocreature (commented):
The changes to rounding mode literals are great. I would love to be convinced that the changes to Numeric are correct but I’m not entirely sure atm.

I incorporated to changes into my branch already:
1. The removal of Text -> BigDecimal.
2. The removal of a rounding mode on BigDecimal <-> Numeric conversions.

I don’t understand the motivation for switching math context from builtins to expressions.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -602,6 +613,9 @@ Then we can define our kinds, types, and expressions::
        |  LitTimestamp                              -- ExpLitTimestamp: UTC timestamp literal
        |  LitParty                                  -- ExpLitParty: Party literal
        |  cid                                       -- ExpLitContractId: Contract identifiers
+       |  LitRoundingMode                           -- ExpLitRoundingMode: Rounding Mode
+       |  UnBondedMathContext                       -- ExpUnboundedMathContext: Unlimited Math Context
+       |  BoundedMathContext e₁ e₂                  -- ExpFoundedMathContext: Unlimited Math Context
```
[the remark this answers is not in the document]
remyhaemmerle-da: In the original design I deliberately do not use Builtin function outside the builtin function section: there is no concrete  instance of Builtin function in the type inference or the operation semantics rules. I just think keeping the coherency is important.

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: It is what we have done for map.
There is expression for maps, but we do not have it in proto.

daml-lf/spec/daml-lf-1.rst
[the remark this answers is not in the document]
cocreature: Alright, I’m fine with going back to expressions, just wanted to understand the reasoning.

daml-lf/spec/daml-lf-1.rst
cocreature: added set scale so I think you can now implement this properly.

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: I see, thanks for the clarification :)

daml-lf/spec/daml-lf-1.rst
```diff
@@ -239,6 +239,10 @@ Version: 1.dev (development)
 ............................
 
   + **Add** exception handling.
+  + **Add** BigDecimal type.
```
cocreature, line 242:
```suggestion
  + **Add** BigNumeric type.
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -483,18 +487,38 @@ We now define all the literals that a program can handle::
...
   ``38``, bounds inclusive.
 * A ``LitInt64`` represents a standard signed 64-bit integer (integer
   between ``−2⁶³`` to ``2⁶³−1``).
-* A ``LitNumeric`` represents a signed number that can be represented
-  in base-10 without loss of precision with at most 38 digits
-  (ignoring possible leading 0 and with a scale (the number of
-  significant digits on the right of the decimal point) between ``0``
-  and ``37`` (bounds inclusive). In the following, we will use
-  ``scale(LitNumeric)`` to denote the scale of the decimal number.
+* A decimal numbers is a signed number that can be represented as a
```
cocreature, line 507:
```suggestion
* A decimal number is a signed number that can be represented as a
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -483,18 +487,38 @@ We now define all the literals that a program can handle::
...
+  product `i * 10^-s` where `i` (the *unscaled value* of the number)
+  is a signed integer not divisible by ten and `s` (the *scale* of the
+  number) is a signed integer. The *precision* of a decimal numbers
+  if the number of digits of its unscaled value (ignoring possible
```
cocreature, line 511:
```suggestion
  is the number of digits of its unscaled value (ignoring possible
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -483,18 +487,38 @@ We now define all the literals that a program can handle::
...
+  leading zeros). By convention the scale and the precision of zero
+  are 0.  Daml-LF distinguishes two kinds of decimal numbers:
+  + A ``LitNumeric`` represents those decimal numbers that have a
+    precision of at most 38 and a scale between ``0`` and ``37``
+    (bounds inclusive).
```
cocreature, line 516: This doesn’t seem right. If I have `replicate 38 '1' <> "." <> replicate 37 '1'`, that falls under the definition here but it’s definitely not a valid numeric.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -515,13 +541,16 @@ The literals represent actual Daml-LF values:
...
+``LitBigNumeric``, ``LitDate``, ``LitTimestamp``) are ordered by
+natural ordering. Text-like literals (``LitText``, ``LitParty``, and
 ``Contract ID``) are ordered lexicographically. Note that in the ASCII
 encoding, the character ``#`` comes before digits, meaning V0 Contract
 ID are ordered before V1 Contract ID. In the following we will denote
-the corresponding (non-strict) order by ``≤ₗ``.
+the corresponding (non-strict) order by ``≤ₗ``.  RoundingMode literals
+are ordered as they appear in the definition of ``LitRoundingMode``,
+e.g. ``ROUNDING_UP`` and ``ROUNDING_UNNECESSARY`` are the smallest and
+greatest rounding mode.
```
cocreature, line 553: Do we need ordering on rounding modes? I would have thought no

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3693,6 +3764,82 @@ Numeric functions
...
+  is not a valid BigNumeric.
+
+* ``SUB_BIGNUMERIC : 'BigNumeric' → 'BigNumeric' → 'BigNumeric'``
+
+  Subtracts the two decimals. Throws an ``ArithmeticError`` if the
+  output is not a valid BigNumeric.
+
+* ``MUL_BIGNUMERIC : 'BigNumeric' → 'BigNumeric' → 'BigNumeric'``
+
+  Multiplies the two numerics. Throws an ``ArithmeticError`` if the
```
cocreature, line 3782: add and sub call them decimals, here you call them numeric.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3693,6 +3764,82 @@ Numeric functions
...
+  output is not a valid BigNumeric.
+
+* ``DIV_BIGNUMERIC : 'RoundingMode' → 'Int' → 'BigNumeric' → 'BigNumeric' → 'BigNumeric'``
+
+  Divides the first decimal by the second one and rounds the result
```
remyhaemmerle-da, line 3787:
```suggestion
  Divides the first numeric by the second one and rounds the result
```

daml-lf/spec/daml-lf-1.rst
cocreature: Oh I’m dumb nvm.

daml-lf/spec/daml-lf-1.rst
cocreature: Can’t we fail in the typechecker instead of adding a dynamic check?

remyhaemmerle-da: We cannot fail in the type checker, as comparison and equality are generic. 
This is precisely why comparison of functions type checks.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -483,18 +487,38 @@ We now define all the literals that a program can handle::
...
   between ``−2⁶³`` to ``2⁶³−1``).
-* A ``LitNumeric`` represents a signed number that can be represented
-  in base-10 without loss of precision with at most 38 digits
-  (ignoring possible leading 0 and with a scale (the number of
-  significant digits on the right of the decimal point) between ``0``
-  and ``37`` (bounds inclusive). In the following, we will use
-  ``scale(LitNumeric)`` to denote the scale of the decimal number.
+* A decimal number is a signed number that can be represented as a
+  product `i * 10^-s` where `i` (the *unscaled value* of the number)
+  is a signed integer not divisible by ten and `s` (the *scale* of the
```
cocreature, line 509: maybe spell out that this is equivalent to no trailing zeroes.

daml-lf/spec/daml-lf-1.rst
cocreature: oh I remember now makes sense.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -680,6 +713,13 @@ Then we can define our kinds, types, and expressions::
   because erasable kinds have no runtime representation. This affects the
   operational semantics. The right hand side of an arrow is always erasable.
 
+.. note:: The explicit syntax for BigNumeric literal (case
+  ``ExpLitBigNumeric``) is forbidden in serialized programs. It is
```
cocreature, line 717: Why? I want to be able to write `\x -> x * 2.0` in Daml without going through a numeric conversion.

cocreature (approved):
Thank you!

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: This does not prevent you to have that `BigNumeric` literal in Daml.
 - The simplest is to limit DAML `BigNumeric` literals to those that can be represented as `Numerics`, and wrapped those in a `TO_BIGNUMERIC_NUMERIC`  when generated the LF code. I think this make complete sens as Numeric are not serializable; `BigNumeric` should be use only for intermediary computation. 
 - If you absolutely want that all valid `BigNumeric` have a valid DAML literal (what I really not sure we should) you can still internally construct them using `TO_BIGNUMERIC_NUMERIC`, `ADD_BIGNUMERIC` and `SHIFT_BIGNUMERIC` and cache the result in an LF  `val`.

cocreature: What do we gain from disallowing `BigNumeric` literals? I disagree that this is related to serializability. We clearly need literals in intermediate computations. I think you’re probably right that the numeric wrapper is sufficient but it seems worse from a performance pov with relatively little benefit.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -483,18 +487,38 @@ We now define all the literals that a program can handle::
...
+  is a signed integer not divisible by ten and `s` (the *scale* of the
+  number) is a signed integer. The *precision* of a decimal numbers
+  is the number of digits of its unscaled value (ignoring possible
+  leading zeros). By convention the scale and the precision of zero
+  are 0.  Daml-LF distinguishes two kinds of decimal numbers:
+  + A ``LitNumeric`` represents those decimal numbers that have a
+    precision of at most 38 and a scale between ``0`` and ``37``
+    (bounds inclusive).
+  + A ``LitBigNumeric`` represents those decimal numbers smaller than
+    `10^(-MaxScale+1)` that have a scale of at least ``MaxScale``.
```
sofiafaro-da, line 518: I'm confused because "MaxScale" sounds like an upper bound, but this sentence says it's a lower bound.

```suggestion
    `10^(MaxScale+1)` that have a scale of at most ``MaxScale``.
```

Also shouldn't there be some restriction on the precision?

daml-lf/spec/daml-lf-1.rst
```diff
@@ -1796,6 +1848,9 @@ need to be evaluated further. ::
    ——————————————————————————————————————————————————— ValExpLitNumeric
      ⊢ᵥ  LitNumeric
 
+   ——————————————————————————————————————————————————— ValExpLitBigNumeric
+     ⊢ᵥ  LitBigDecimal
```
sofiafaro-da, line 1852:
```suggestion
     ⊢ᵥ  LitBigNumeric
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -4663,6 +4810,18 @@ program exception using
   ``ANY_EXCEPTION_MESSAGE``, ``GENERAL_ERROR_MESSAGE``, or
   ``ARITHMETIC_ERROR_MESSAGE`.
 
+BigNumeric
+..........
+
+Daml-LF 1.7 is the first version that supports BigNumeric.
```
sofiafaro-da, line 4816: Wait, Daml-LF 1.7?

sofiafaro-da (approved):
Thanks Rémy! I left some comments but otherwise LGTM.

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: When I mean no literals, I mean there is not literal in the syntax i.e. in the archive, there is of course literal operationally. 
It is a bit like GenMap, we do not have literals in the syntax, you can create Map only bu using the builtin.

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da, line 4816:
```suggestion
Daml-LF 1.dev is the first version that supports BigNumeric.
```

remyhaemmerle-da: Good catch

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: Will try to reformulate

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3693,6 +3763,82 @@ Numeric functions
...
+
+  - ``'ROUNDING_HALF_EVEN'`` : Rounds towards the nearest neighbor
+    unless both neighbors are equidistant, in which case round towards
+    the even neighbor.
+
+  - ``'ROUNDING_HALF_UP'`` : Round towards the nearest neighbor unless
+    both neighbors are equidistant, in which case round away from
+    zero.
+
+  - ``'ROUNDING_UNNECESSARY'`` : Throw if the exact result cannot be
```
sofiafaro-da, line 3812:
```suggestion
  - ``'ROUNDING_UNNECESSARY'`` : Throw `ArithmeticError` if the exact result cannot be
```

sofiafaro-da (approved):
LGTM, thank you :-)

PR #8910, opened 2021-02-22
Relax checking of state updates [KVL-848]

miklos-da:
This PR relaxes checking of `TransactionAccepted` state updates in the following ways by the integrity-checker:
* *Blinding info is not checked.* We have started adding blinding info to transaction updates only recently, hence, it is not present in older ledger exports.
* *Transaction ID is not checked.* Transaction ID is generated from the serialized form of a Daml submission in case of batched submissions which is not expected to stay the same across Daml SDK versions.

### Pull Request Checklist

trigger the build.


fabiotudone-da (approved):
I'd only add the explanation in the PR as comments in the code.

miklos-da:
> I'd only add the explanation in the PR as comments in the code.

Good point -- commented on `ReadServiceStateUpdateComparison.normalizeUpdate`.

SamirTalwar-DA (commented):
It looks good to me.

Should we make these optional flags so we _do_ check these on versions which we know contain the relevant information? Perhaps something like `--as-of=YYYY-MM-DD`, and then compare the date against a range for each feature.

miklos-da:
> It looks good to me.
> 
> Should we make these optional flags so we _do_ check these on versions which we know contain the relevant information? Perhaps something like `--as-of=YYYY-MM-DD`, and then compare the date against a range for each feature.

@SamirTalwar-DA Added some knobs so that the comparison can be controlled in a fine-grained way (to be wired in in a follow-up PR). Please let me know if you have concerns with this direction.

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/StateUpdateComparison.scala
```diff
@@ -56,6 +56,17 @@ final class ReadServiceStateUpdateComparison(
...
+
+  case class NormalizationSettings(
+      ignoreBlindingInfo: Boolean = false,
+      ignoreTransactionId: Boolean = false,
+  )
+
+  val DefaultNormalizationSettings: NormalizationSettings = NormalizationSettings(
+    ignoreBlindingInfo = true,
+    ignoreTransactionId = true,
+  )
```
SamirTalwar-DA, line 68: Seems sensible. Happy for this to be configurable in another changeset.

SamirTalwar-DA (approved):

PR #8912, opened 2021-02-22
Factor out utilities from com.daml.script.dump.Main

aherrmann-da:
- Factor out `getACS` and `getTransactionTrees` into `LedgerUtils`.
- Factor out `writeDump` into `Dump`.
- Factor out calculating package references into `Dependencies`.

Following suggestion from https://github.com/digital-asset/daml/pull/8898#discussion_r579146541.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Dependencies.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Dependencies.scala
@@ -6,19 +6,37 @@ package com.daml.script.dump
 import java.nio.file.Path
 
 import com.daml.daml_lf_dev.DamlLf
+import com.daml.ledger.api.v1.transaction.TransactionTree
+import com.daml.ledger.api.v1.event.CreatedEvent
+import com.daml.ledger.api.v1.value.Value.Sum
 import com.daml.ledger.client.LedgerClient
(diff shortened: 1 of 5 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/Dependencies.scala
```diff
@@ -6,20 +6,38 @@ package com.daml.script.dump
...
+    contracts
+      .foldMap(ev => valueRefs(Sum.Record(ev.getCreateArguments)))
+      .map(i => PackageId.assertFromString(i.packageId))
+  }
+
+  def treesReferences(transactions: Iterable[TransactionTree]): Set[PackageId] = {
+    transactions
+      .foldMap(treeRefs(_))
+      .map(i => PackageId.assertFromString(i.packageId))
+  }
```
cocreature, line 39: Definitely bikeshedding but maybe `TreeUtils` would be a better place for those?

cocreature (approved):
Nice, thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/Dependencies.scala
aherrmann-da: Yeah, wasn't sure about these. Moved.

PR #8916, opened 2021-02-22
Add benchmarks for non-repudiation

### LINKED ISSUE #8635: Non-repudiation: track performance and overhead on bare participant
It's extremely important that the overhead of having a non-repudiation middleware is kept to a bare minimum. In order for this to happen, it should be possible to keep track of the performance, in particular in terms of command throughput. Metrics should also be available to track performance of live systems.

stefanobaghino-da:
Closes #8635

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/runtime-components/non-repudiation-perf/BUILD.bazel
+++ b/runtime-components/non-repudiation-perf/BUILD.bazel
@@ -5,16 +5,24 @@ load(
     "//bazel_tools:scala.bzl",
     "da_scala_benchmark_jmh",
 )
-load(
-    "@scala_version//:index.bzl",
-    "scala_major_version_suffix",
-)
 
 da_scala_benchmark_jmh(
     name = "non-repudiation-perf",
     srcs = glob(["src/main/scala/**/*.scala"]),
     max_heap_size = "4g",
     resources = ["src/main/resources/logback.xml"],
+    scala_deps = [
+        "@maven//:com_chuusai_shapeless",
+        "@maven//:com_lihaoyi_sourcecode",
+        "@maven//:org_scala_lang_modules_scala_collection_compat",
+        "@maven//:org_tpolecat_doobie_core",
+        "@maven//:org_tpolecat_doobie_free",
+        "@maven//:org_tpolecat_doobie_hikari",
+        "@maven//:org_typelevel_cats_core",
+        "@maven//:org_typelevel_cats_effect",
+        "@maven//:org_typelevel_cats_free",
+        "@maven//:org_typelevel_cats_kernel",
+    ],
     tags = ["manual"],
     deps = [
         "//ledger-api/grpc-definitions:ledger_api_proto_scala",
@@ -30,15 +38,5 @@ da_scala_benchmark_jmh(
         "@maven//:io_grpc_grpc_services",
         "@maven//:org_postgresql_postgresql",
         "@maven//:org_slf4j_slf4j_api",
-        "@maven//:com_chuusai_shapeless_{}".format(scala_major_version_suffix),
-        "@maven//:com_lihaoyi_sourcecode_{}".format(scala_major_version_suffix),
-        "@maven//:org_scala_lang_modules_scala_collection_compat_{}".format(scala_major_version_suffix),
-        "@maven//:org_tpolecat_doobie_core_{}".format(scala_major_version_suffix),
-        "@maven//:org_tpolecat_doobie_free_{}".format(scala_major_version_suffix),
-        "@maven//:org_tpolecat_doobie_hikari_{}".format(scala_major_version_suffix),
-        "@maven//:org_typelevel_cats_core_{}".format(scala_major_version_suffix),
-        "@maven//:org_typelevel_cats_effect_{}".format(scala_major_version_suffix),
-        "@maven//:org_typelevel_cats_free_{}".format(scala_major_version_suffix),
-        "@maven//:org_typelevel_cats_kernel_{}".format(scala_major_version_suffix),
     ],
 )
--- a/runtime-components/non-repudiation-perf/BUILD.bazel
+++ b/runtime-components/non-repudiation-perf/BUILD.bazel
@@ -1,0 +1,43 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+load(
+    "//bazel_tools:scala.bzl",
+    "da_scala_benchmark_jmh",
+)
+load(
+    "@scala_version//:index.bzl",
+    "scala_major_version_suffix",
+)
+
+da_scala_benchmark_jmh(
+    name = "non-repudiation-perf",
+    srcs = glob(["src/main/scala/**/*.scala"]),
+    max_heap_size = "4g",
+    resources = ["src/main/resources/logback.xml"],
+    tags = ["manual"],
+    deps = [
+        "//ledger-api/grpc-definitions:ledger_api_proto_scala",
+        "//libs-scala/doobie-slf4j",
+        "//libs-scala/postgresql-testing",
+        "//libs-scala/resources",
+        "//libs-scala/resources-grpc",
+        "//runtime-components/non-repudiation",
+        "//runtime-components/non-repudiation-client",
+        "//runtime-components/non-repudiation-postgresql",
+        "//runtime-components/non-repudiation-resources",
+        "@maven//:ch_qos_logback_logback_classic",
+        "@maven//:io_grpc_grpc_services",
+        "@maven//:org_postgresql_postgresql",
+        "@maven//:com_chuusai_shapeless_{}".format(scala_major_version_suffix),
+        "@maven//:com_lihaoyi_sourcecode_{}".format(scala_major_version_suffix),
+        "@maven//:org_scala_lang_modules_scala_collection_compat_{}".format(scala_major_version_suffix),
+        "@maven//:org_tpolecat_doobie_core_{}".format(scala_major_version_suffix),
+        "@maven//:org_tpolecat_doobie_free_{}".format(scala_major_version_suffix),
+        "@maven//:org_tpolecat_doobie_hikari_{}".format(scala_major_version_suffix),
+        "@maven//:org_typelevel_cats_core_{}".format(scala_major_version_suffix),
+        "@maven//:org_typelevel_cats_effect_{}".format(scala_major_version_suffix),
+        "@maven//:org_typelevel_cats_free_{}".format(scala_major_version_suffix),
+        "@maven//:org_typelevel_cats_kernel_{}".format(scala_major_version_suffix),
+    ],
+)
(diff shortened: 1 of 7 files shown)


runtime-components/non-repudiation-perf/BUILD.bazel
```diff
@@ -0,0 +1,43 @@
...
+load(
+    "//bazel_tools:scala.bzl",
+    "da_scala_benchmark_jmh",
+)
+load(
+    "@scala_version//:index.bzl",
+    "scala_major_version_suffix",
+)
+
+da_scala_benchmark_jmh(
```
stefanobaghino-da, line 13: This rule unfortunately does not support `scala_deps` and `runtime_deps` (something we may want to work on at a certain point).

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
```diff
@@ -0,0 +1,167 @@
...
+  var commandPayloadSize: Int = _
+
+  private var stub: CommandSubmissionServiceBlockingStub = _
+  private var proxy: Resource[ExecutionContext, Server] = _
+  private var database: PostgresDatabase = _
+  private var payload: String = _
+
+  @Benchmark
+  def run(): Unit = {
+    // Generating commands adds very little noise to substantial benchmarks
```
stefanobaghino-da, line 52: I checked this by first having commands ready to be shot and comparing it with this. The advantage of this approach is that we can easily create a command with a new identifier for every test, which better simulates actual production conditions.

runtime-components/non-repudiation-perf/BUILD.bazel
cocreature: How’s that https://github.com/digital-asset/daml/pull/8928

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
```diff
@@ -0,0 +1,167 @@
...
+    generator.generate(2048)
+    val key = generator.getPrivateKey
+    val certificate = generator.getSelfCertificate(
+      new X500Name("CN=Non-Repudiation Test,O=Digital Asset,L=Zurich,C=CH"),
+      1.hour.toSeconds,
+    )
+
+    val service = DummyCommandSubmissionService.bind(executionContext)
+
+    val participantName = InProcessServerBuilder.generateName()
```
cocreature, line 77: does the in process server builder involve a network connection (to localhost) or does it bypass that completely? If not, I think it might make sense to have a second benchmark which runs this closer to the scenario we recommend where those are run as separate processes (shouldn’t matter) communicating over a network connection (that should matter).

cocreature (approved):
Nice, thank you!

runtime-components/non-repudiation-perf/BUILD.bazel
cocreature: merged the PR, i’ll leave it up to you if you want to rebase on top of it or do the cleanup separately.

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
stefanobaghino-da: Because we would have to generate a lot of them in advance in order to consume them. I've tried generating 5000 commands beforehand but it consumed all of them after 3 warm-up rounds. With 100 KB payload sizes generating too many of them beforehand becomes unfeasible.

Initially I was re-using the same command over and over again, but this means that this benchmark cannot be used for longer runs to check that performance are stable with a larger number of commands being stored on the back-end. This is what I was referring to when I said that it " better simulates actual production conditions" (as in, not writing a command with the same command identifier over and over again).

Note that I was conservative in writing "very little noise", from my tests the difference is imperceptible. Consider that the cost of generating a command is negligible (especially by keeping the payload the same) compared to accessing the database.

Makes sense? Should I put if differently in the comment? I realize now this comment makes sense to me due to the fact I'm aware of the context, but is probably not very helpful for readers.

runtime-components/non-repudiation-perf/BUILD.bazel
stefanobaghino-da: Well, this was easier than what I thought. 😄 Thanks.

runtime-components/non-repudiation-perf/BUILD.bazel
```diff
@@ -0,0 +1,44 @@
...
+    name = "non-repudiation-perf",
+    srcs = glob(["src/main/scala/**/*.scala"]),
+    max_heap_size = "4g",
+    resources = ["src/main/resources/logback.xml"],
+    tags = ["manual"],
```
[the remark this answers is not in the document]
stefanobaghino-da: Ha, good point, thanks.

runtime-components/non-repudiation-perf/src/main/resources/logback.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+    <!-- Not logging anything on purpose -->
+    <!-- If there is an error, other tests with more precise feedback will log it -->
+    <appender name="nop" class="ch.qos.logback.core.helpers.NOPAppender" />
```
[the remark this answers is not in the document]
stefanobaghino-da: 🙉

runtime-components/non-repudiation-postgresql/src/main/scala/com/daml/nonrepudiation/postgresql/PostgresqlSignedPayloadRepository.scala
```diff
@@ -18,7 +18,7 @@ final class PostgresqlSignedPayloadRepository(transactor: Transactor[IO])(implic
   private val table = Fragment.const(s"${Tables.Prefix}_signed_payloads")
 
   private def getSignedPayload(key: CommandIdString): Fragment =
-    fr"select algorithm, fingerprint, payload, signature from" ++ table ++ fr"where command_id = $key"
+    fr"select algorithm, fingerprint, payload, signature, timestamp from" ++ table ++ fr"where command_id = $key"
```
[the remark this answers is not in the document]
stefanobaghino-da: > Was this just broken before?

Yes.

> If so, can we add a test outside of the benchmark to make sure it doesn’t break again?

Definitely!

---

This is a leftover from code that was previously there to check that I was _actually_ writing stuff to the underlying database in the benchmarks.

I'll retract this change from this PR and make a separate one that fixes this.

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
stefanobaghino-da: > does the in process server builder involve a network connection (to localhost) or does it bypass that completely?

The latter.

I can add it for sure, but what would we be measuring in this way? The overhead of going through the network stack? Is it meaningful?

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
```diff
@@ -0,0 +1,167 @@
...
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext}
+
+@State(Scope.Benchmark)
+class NonRepudiationProxyBenchmark extends PostgresAround {
+
+  import NonRepudiationProxyBenchmark._
+
+  @Param(Array("100000"))
+  var commandPayloadSize: Int = _
```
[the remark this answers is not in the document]
stefanobaghino-da: Honestly I wasn't really thinking about it, I don't really have a preference. Considering that JMH injects the parameters in the generated code, using `_` can (incidentally) be less confusing. Not sure. Thoughts?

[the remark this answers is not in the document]
cocreature: Seems fine, I guess it’s a nice way of saying "this value won’t be used". A bit of a shame that it won’t blow up if the value is used but I guess not much we can do about that.

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
cocreature: I’d say it is meaningful. Less in the sense, that we can do much to reduce it so not super useful for driving optimizations but it is something that our clients will hit so in terms of the numbers they can expect, it probably comes closer when looking at the upper bound on what you could reach in a real setup.

runtime-components/non-repudiation-perf/BUILD.bazel
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/8916/commits/2ab5c17ba236690830f2b60e2092149f41520e88

runtime-components/non-repudiation-perf/BUILD.bazel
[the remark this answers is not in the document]
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/8916/commits/45de2624b27b880b4ebcb5399c9053e4c6ef9593

runtime-components/non-repudiation-postgresql/src/main/scala/com/daml/nonrepudiation/postgresql/PostgresqlSignedPayloadRepository.scala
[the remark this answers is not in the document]
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/8916/commits/aa322169ad2cd85638eadb986f65c0a019fabf02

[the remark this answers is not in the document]
stefanobaghino-da: Issue tracked by https://github.com/digital-asset/daml/issues/8935, I'll fix it right away.

runtime-components/non-repudiation-perf/src/main/scala/com/daml/nonrepudiation/perf/NonRepudiationProxyBenchmark.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8916/commits/d107795813987a3fcd3fbfc6b7d3d72fa9b6c7cd

PR #8919, opened 2021-02-23
damlc test: feature: add --all flag.

robin-da:
This adds a flag `--all` that will test all present scenarios/scripts in
the code and also report code coverage with respect to all present
templates/choices, whereas without the flag, code coverage is reported
relative to a single package.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
+++ b/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
@@ -737,6 +737,24 @@ contextForFile file = do
         , ctxSkipValidation = SS.SkipValidation (getSkipScenarioValidation envSkipScenarioValidation)
         }
 
+contextForPackage :: NormalizedFilePath -> LF.Package -> Action SS.Context
+contextForPackage file pkg = do
+    lfVersion <- getDamlLfVersion
+    encodedModules <-
+        mapM (\m -> fmap (\(hash, bs) -> (hash, (LF.moduleName m, bs))) (encodeModule lfVersion m)) $
+        NM.toList $ LF.packageModules pkg
+    PackageMap pkgMap <- use_ GeneratePackageMap file
+    stablePackages <- useNoFile_ GenerateStablePackages
+    pure
+        SS.Context
+            { ctxModules = Map.fromList encodedModules
+            , ctxPackages =
+                  [ (LF.dalfPackageId pkg, LF.dalfPackageBytes pkg)
+                  | pkg <- Map.elems pkgMap ++ Map.elems stablePackages
+                  ]
+            , ctxSkipValidation = SS.SkipValidation True -- no validation for external packages
+            }
+
 worldForFile :: NormalizedFilePath -> Action LF.World
 worldForFile file = do
     WhnfPackage pkg <- use_ GeneratePackage file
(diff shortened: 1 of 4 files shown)


compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -73,24 +76,36 @@ testRun h inFiles lfVersion coverage color mbJUnitOutput  = do
     let files = nubOrd $ concat $ inFiles : catMaybes deps
 
     results <- runActionSync h $ do
-        dalfs <- Shake.forP files dalfForScenario
         Shake.forP files $ \file -> do
+            dalf <- dalfForScenario file
```
cocreature, line 80: I’m a bit confused. I was expecting us to change which tests we run based on `runAllTests` but that doesn’t seem to be the case. We always run tests against all modules in the main dalf. What we change is whether coverage is reported relative to the main dalf or relative to all dalfs.

I see good usecases for the following configurations:
1. Run all scenarios in the main dalf, report coverage against main dalf. You’ll mainly use that for performance reasons. Running scenarios against all dalfs can clearly be slower.
2. Run all scenarios in all dalfs, report coverage against all dalfs. Doesn’t need much of an explanation I think. You run as much as possible and report it against everything it depends on.

The current `--all` flag implements a third option:

3. Run scenarios of the main dalf and report coverage against all dalfs.

This seems a bit of a weird configuration. You can’t expect to get 100% in that setting without repeating potential tests in your dependencies. If your dependencies don’t have any tests, this is equivalent to 2 so not very useful as a separate configuration. The one usecase I can see is if your dependencies do contain tests but for performance reasons you don’t want to run them and you’re only looking at relative improvements. I think this is a valid usecase but I think it points at a more general feature that might solve this and others:

A way to only run specific tests. Filtering to the main dalf is one specific instance of that but I think there are more, here’s a very rough idea of a syntax for filtering:

1. pkgid:modulename:testname runs exactly that test
2. packagename:modulename:testname runs exactly that test, errors out if package name is ambigous
3. packagename-version:modulename:testname runs exactly that test, errors out if package name & version is ambigous
4. :modulename:testname runs exactly that test in the main dalf
5. :modulename: runs all tests in that module.

You probably also want some form of pattern matching (e.g., regex) in all 3 positions.

That’s definitely not something for this PR and I don’t think anybody has asked for it so far so I would recommend to adjust this PR to switch between 1 and 2 and open a separate issue for 3.

robin-da: @cocreature I think the source of my confusion comes from line 75-76 in Damlc/Test.hs. I was working under the assumption that this will give me all dependencies, not just the main dalf, and hence would implement option 1/2 depending on the flag set or not. Is that not the case?

cocreature: `dalfForScenario` returns a dalf for the given file based on files from the current project so a subset of what `daml build` gives you (the reason why it’s a subset is that we want to run tests even if other files fail to compile at least in daml studio). `files` refers to the files in the current project. `getExternalPackages` provides you with the dependencies.

robin-da: @cocreature I tested this manually. A test would probably have to go to the integration tests for creating and importing an external package, not sure if we should do it since those are fairly slow already.

cocreature: @robin-da you don’t need the integration tests for building code that depends on an external package. We already have a code that does this, see https://github.com/digital-asset/daml/blob/4fcd2b833f40c1c4a4d92b45b87380d9514e85ba/compiler/damlc/tests/src/DamlcTest.hs#L273.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -72,25 +77,71 @@ testRun h inFiles lfVersion coverage color mbJUnitOutput  = do
...
+            return (file, dalf, mbResults)
+
+    extResults <- runActionSync h $ do
+        if runAllTests
+        then do
+            projPathM <- liftIO getProjectPath
+            case projPathM of
+                Nothing -> do
+                    liftIO $
+                        putStrLn "Warning: Project root not found, not testing package dependencies"
```
cocreature: why do we need a project root for this? We already have `extPkgs` here.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -72,25 +77,71 @@ testRun h inFiles lfVersion coverage color mbJUnitOutput  = do
     deps <- runActionSync h $ mapM getDependencies inFiles
     let files = nubOrd $ concat $ inFiles : catMaybes deps
 
+    -- get all external dependencies
+    extPkgs <- fmap (nubSortOn LF.extPackageId . concat) $ runActionSync h $
+      Shake.forP files $ \file -> getExternalPackages file
+    let extDalfs =
+                [ (Just pId, dalf)
+                | pkg <- extPkgs
+                , let dalfs = NM.elems $ LF.packageModules $ LF.extPackagePkg pkg
```
cocreature, line 86:
```
            , let modules = NM.elems $ LF.packageModules $ LF.extPackagePkg pkg
```
or am I missing something here? Same for the other names here, you seem to return a list of `(pkgId, module)`

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -103,11 +154,12 @@ failedTestOutput h file = do
...
 printTestCoverage ::
-       ShowCoverage
-    -> [LF.Module]
+    ShowCoverage
+    -> [LF.ExternalPackage]
+    -> [(Maybe LF.PackageId, LF.Module)]
     -> [(VirtualResource, Either SSC.Error SS.ScenarioResult)]
     -> IO ()
-printTestCoverage ShowCoverage {getShowCoverage} dalfs results
+printTestCoverage ShowCoverage {getShowCoverage} extPkgs dalfs results
```
cocreature, line 157: same name confusion betwene modules and dalfs also exists here (and did before this PR as well).

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -72,25 +77,66 @@ testRun h inFiles lfVersion coverage color mbJUnitOutput  = do
...
-            pure (file, results)
+            let mbResults = liftM2 (++) mbScenarioResults mbScriptResults
+            forM_ mbResults $ \scenarioResults -> do
+                  -- failures are printed out through diagnostics, so just print the successes
+                  let results' = [(v, r) | (v, Right r) <- scenarioResults]
+                  liftIO $ printScenarioResults results' color
+            return (file, dalf, mbResults)
+
+    extResults <- runActionSync h $ do
+        if runAllTests
```
cocreature, line 104: you can pull out the if which would be a bit faster, probably not super measurble but `runAction` is surprisingly slow especially the sync version.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
```diff
@@ -72,25 +77,66 @@ testRun h inFiles lfVersion coverage color mbJUnitOutput  = do
...
+            case headMay inFiles of
+                Nothing -> pure [] -- nothing to test
+                Just file ->
+                    forM extPkgs $ \pkg -> do
+                        mbScenarioResults <- snd <$> runScenariosPkg file pkg extPkgs
+                        mbScriptResults <- snd <$> runScriptsPkg file pkg extPkgs
+                        let mbResults = liftM2 (++) mbScenarioResults mbScriptResults
+                        forM_ mbResults $ \scenarioResults -> do
+                            let results' = [(v, r) | (v, Right r) <- scenarioResults]
+                            liftIO $ printScenarioResults results' color
```
cocreature, line 115: seems like this logic is shared between modules of the current project and external packages, worth factoring out?

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
```diff
@@ -838,6 +856,119 @@ runScriptsRule opts =
       let (diags, results) = unzip scenarioResults
       pure (catMaybes diags, Just results)
 
+runScriptsPkg ::
+       NormalizedFilePath
+    -> LF.ExternalPackage
+    -> [LF.ExternalPackage]
+    -> Action ([FileDiagnostic], Maybe [(VirtualResource, Either SS.Error SS.ScenarioResult)])
+runScriptsPkg projRoot extPkg pkgs = do
```
cocreature, line 864: `runScriptsPkg` and `runScenariosPkg` look almost identical and they should also be able to share the same context. Can we merge them into one function that runs both given that we never run them separately?

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
```diff
@@ -838,6 +856,119 @@ runScriptsRule opts =
...
+        maybe (LF.unPackageId pkgId) (LF.unPackageName . LF.packageName) $ LF.packageMetadata pkg
+    world = LF.initWorldSelf pkgs pkg
+    scenarios =
+        map fst $
+        concat
+            [ scriptsInModule (EnableScripts True) mod
+            | mod <- NM.elems $ LF.packageModules pkg
+            , LF.moduleName mod /= LF.ModuleName ["Daml", "Script"]
+            ]
+    toDiagnostic :: Either SS.Error SS.ScenarioResult -> Maybe FileDiagnostic
```
cocreature, line 903: this looks duplicated with what we do for scenarios/scripts in the current package. Probably worth factoring this and perhaps some of the other logic in this function.

compiler/damlc/tests/src/DamlcTest.hs
```diff
@@ -219,6 +219,69 @@ testsForDamlcTest damlc = testGroup "damlc test" $
...
+            , "template T with p : Party where"
+            , "  signatory p"
+            , "  choice X : () with controller p"
+            , "    do pure ()"
+            , "x = scenario do"
+            , "      alice <- getParty \"Alice\""
+            , "      c <- submit alice $ create T with p = alice"
+            , "      submit alice $ exercise c X with"
+            ]
+          (exitCode, stdout, stderr) <- readProcessWithExitCode damlc
```
cocreature, line 267: maybe test both the coverage with --all and without?

cocreature (approved):
Looks great! A few small comments. Thanks a lot for adjusting this PR!

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: changed it in the meantime. we need a file to create contexts for the scenario service. I changed it to use one of the given input files (if any).

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: changed.

compiler/damlc/tests/src/DamlcTest.hs
robin-da: added a test with an import but without `-all` flag.

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
robin-da: merged the two functions.

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
robin-da: factored this out.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: it's already weird that the function `dalfForScenario` returns a module and not a dalf. I changed it to `moduleForScenario` and did a `s/dalf/module/`.

compiler/damlc/lib/DA/Cli/Damlc/Test.hs
robin-da: factored out.

PR #8921, opened 2021-02-23
Check the exact names of type read in Java codegen tests

### LINKED ISSUE #8920: Make Java codegen runner test explicitly the expected types
Right now the Java codegen runner tests check the number of interfaces read from a DAR matches the expected number.

A more accurate test could check for the names, so that changes to the test itself become more self-descriptive.

_Originally posted by @stefanobaghino-da in https://github.com/digital-asset/daml/pull/8856#r581072132_

stefanobaghino-da:
Closes #8920

As raised initially in https://github.com/digital-asset/daml/pull/8856#discussion_r581072132

This should make the test itself more self-explanatory.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
+++ b/language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
@@ -37,7 +37,63 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
 
     val (interfaces, pkgPrefixes) = CodeGenRunner.collectDamlLfInterfaces(conf)
 
-    assert(interfaces.length == 19)
+    interfaces.flatMap(
+      _.typeDecls.keys.map(_.qualifiedName)
+    ) should contain theSameElementsAs List(
+      "DA.Date.Types:DayOfWeek",
+      "DA.Date.Types:Month",
+      "DA.Generics:Associativity",
+      "DA.Generics:DecidedStrictness",
+      "DA.Generics:Fixity",
+      "DA.Generics:Infix0",
+      "DA.Generics:K1",
+      "DA.Generics:Par1",
+      "DA.Generics:SourceStrictness",
+      "DA.Generics:SourceUnpackedness",
+      "DA.Generics:U1",
+      "DA.Internal.Down:Down",
+      "DA.Internal.Template:Archive",
+      "DA.Logic.Types:Formula",
+      "DA.Monoid.Types:All",
+      "DA.Monoid.Types:Any",
+      "DA.Monoid.Types:Product",
+      "DA.Monoid.Types:Sum",
+      "DA.Next.Map:Map",
+      "DA.Next.Set:Set",
+      "DA.NonEmpty.Types:NonEmpty",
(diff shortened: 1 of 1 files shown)


language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
```diff
@@ -37,7 +37,63 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
...
+      "DA.Types:Tuple7",
+      "DA.Types:Tuple8",
+      "DA.Types:Tuple9",
+      "DA.Validation.Types:Validation",
+      "Foo:Foo",
+      "GHC.Stack.Types:CallStack",
+      "GHC.Stack.Types:SrcLoc",
+      "GHC.Tuple:Unit",
+      "GHC.Types:Ordering",
+    )
```
remyhaemmerle-da, line 95: This is depending on the output target of the compiler. 
Ideally it will be nice if this can be linked to the compiler.
Probably a bit difficult. cc @sofiafaro-da

remyhaemmerle-da: Also here we lost the archive structure. I am not sure what we want to test here.
Any opinion @sofiafaro-da ?

stefanobaghino-da: This is just a unit test for the wrapper around the interface reader, to check that it's reading all the expected output from a fixed DAR. Being compatible with multiple Daml-LF versions is checked in the integration tests.

The DAR is "fixed" within the context of the test, but it changes over time due to LF's evolution, which means that periodically you will need to touch this (as you did [here](https://github.com/digital-asset/daml/pull/8856#discussion_r580898288)).

The point of this change would be to make this change easier to understand. If you think there's no point in that, we can simply close this. 🙂

language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
```diff
@@ -37,7 +37,63 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
...
+      "DA.Types:Tuple20",
+      "DA.Types:Tuple3",
+      "DA.Types:Tuple4",
+      "DA.Types:Tuple5",
+      "DA.Types:Tuple6",
+      "DA.Types:Tuple7",
+      "DA.Types:Tuple8",
+      "DA.Types:Tuple9",
+      "DA.Validation.Types:Validation",
+      "Foo:Foo",
```
remyhaemmerle-da, line 90: Is it not sufficient to check it contains "Foo:Foo" ?

stefanobaghino-da: No, it checks whether the interfaces read are all those for which code should be generated, which includes everything in the DAR.

PR #8922, opened 2021-02-23
document HasField

garyverhaegen-da:
This is stolen directly from @sofiafaro-da's excellent [forum answer].

[forum answer]: https://discuss.daml.com/t/generating-record-update-functions-to-be-used-as-parameters/2091/3?u=gary_verhaegen

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/Internal/Record.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/Internal/Record.daml
@@ -12,6 +12,51 @@ import GHC.Types
 import GHC.Tuple() -- reqired for Tuple's themselves
 import GHC.Integer.Type() -- required for Integer primitives
 
+-- | `HasField` gives you getter and setter functions for each record field
+-- automatically.
+--
+-- `HasField x r a` is a typeclass that takes three parameters. The first
+-- parameter `x` is the field name, the second parameter `r` is the record type,
+-- and the last parameter `a` is the type of the field in this record. For
+-- example, if we define a type:
+--
+-- ```
+-- data MyRecord = MyRecord with
+--     foo : Int
+--     bar : Text
+-- ```
+--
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Current page:

Screenshot 2021-02-23 at 16 59

Local preview of this changeset:

Screenshot 2021-02-23 at 16 59 1

compiler/damlc/daml-stdlib-src/DA/Internal/Record.daml
```diff
@@ -12,6 +12,51 @@ import GHC.Types
...
+-- ```
+-- getFoo : MyRecord -> Int
+-- getFoo r = getField @"foo" r
+--
+-- getBar : MyRecord -> Text
+-- getBar = getField @"bar"
+-- ```
+--
+-- Note that this uses the “type application” syntax ( `f @t` ) to specify the
+-- field name, and that the `getBar` definition relies on currying to omit the
```
cocreature, line 47: Maybe just eta-expand it? Either a user already knows about currying in which case, they can figure this out themselves. Or they don’t, and this is just confusing.

cocreature (commented):
I think the docs are very useful but we should prefix them by recommending RecordDotSyntax where possible.

sofiafaro-da (approved):
Thank you!

I agree with Moritz that it's probably better to just add the parameters explicitly here, and to encourage the use of record syntax when it's more convenient.

garyverhaegen-da:
New version:

Screenshot 2021-02-24 at 16 34

I was not able to find a way to render the URL as a link, nor to get the current version number in the URL. I think that's still acceptable (i.e. better than what we have now) as it stands, but if you know how to do either please tell me.

cocreature (approved):
Thanks!

PR #8923, opened 2021-02-23
document potentially-surprising Month <-> Enum

garyverhaegen-da:
Based on a [forum question] by Alxander Bernauer.

[forum question]: https://discuss.daml.com/t/enum-instance-for-month-is-misleading/2098?u=gary_verhaegen

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/Date/Types.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/Date/Types.daml
@@ -17,6 +17,9 @@ data DayOfWeek
     | Sunday
 
 -- | The `Month` type represents a month in the Gregorian calendar.
+--
+-- Note that, while `Month` has an `Enum` instance, the `toEnum` and `fromEnum`
+-- functions start counting at 0, i.e. `toEnum 1 :: Month` is `Feb`.
 data Month
     = Jan
     | Feb


garyverhaegen-da:
Before:

Screenshot 2021-02-23 at 17 21 1

After:

Screenshot 2021-02-23 at 17 21

cocreature (approved):
lgtm, I assume daml-doc does not render docs on instances? haddock does and if we support it, we probably should put the docs there. If not, happy to merge this for now but maybe worth opening an issue for.

garyverhaegen-da:
> lgtm, I assume daml-doc does not render docs on instances?

I'm not familiar with Haddock syntax at all, so I may have missed something. I tried:

```
-- | test
deriving instance Enum Month
```

and that seemed to produce no output at all, as well as

```
deriving instance Enum Month -- ^ test
```

which resulted in a parse error from `damlc`:

```
File:     compiler/damlc/daml-stdlib-src/DA/Date.daml
Hidden:   no
Range:    41:30-41:39
Source:   parser
Severity: DsError
Message:  compiler/damlc/daml-stdlib-src/DA/Date.daml:41:30: error:parse error on input ‘-- ^ test’
```

Happy to try whatever you had in mind if it wasn't either of the above.

PR #8930, opened 2021-02-24
skip Oracle test on forks

garyverhaegen-da:
Fork builds do not have the required credentials.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -249,6 +249,7 @@ jobs:
         ARTIFACTORY_USERNAME: $(ARTIFACTORY_USERNAME)
         ARTIFACTORY_PASSWORD: $(ARTIFACTORY_PASSWORD)
       displayName: 'Build'
+      condition: eq(variables['System.PullRequest.IsFork'], 'False')
     - template: tell-slack-failed.yml
       parameters:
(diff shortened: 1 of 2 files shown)


ci/build.yml
```diff
@@ -249,6 +249,7 @@ jobs:
         ARTIFACTORY_USERNAME: $(ARTIFACTORY_USERNAME)
         ARTIFACTORY_PASSWORD: $(ARTIFACTORY_PASSWORD)
       displayName: 'Build'
+      condition: eq(variables['System.PullRequest.IsFork'], 'False')
```
cocreature, line 252: I’m sure you’ve checked it but I probably wouldn’t have so let me ask the silly question anyway: Is this set on `main`?

cocreature (approved):
Thanks!

ci/build.yml
garyverhaegen-da: Yes, it's a [predefined variable](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml) and is always set to either `"False"` or `"True"`.

PR #8931, opened 2021-02-24
Run PR build on notices updates

garyverhaegen-da:
Second commit is just there for testing, I'll remove before merging.


cocreature (approved):
Thank you!

PR #8942, opened 2021-02-24
new release

garyverhaegen-da:
Last one failed because Sonatype somehow decided to abort our deployment. Let's try again.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature:
Given that the scala 2.13 issue seems real and this will only succeed if we get lucky, I’ll suggest to fix it first.

garyverhaegen-da:
> Given that the scala 2.13 issue seems real and this will only succeed if we get lucky, I’ll suggest to fix it first.

Yes, this will need to be rebased on top of #8943 after that's merged.

garyverhaegen-da:
Actually less error prone to just kill it. We'll make a new one tomorrow.

PR #8943, opened 2021-02-24
Merge Maven uploads for different Scala versions

cocreature:
It turns out Maven will abort an existing staging operation if you
create a new one. This means our jobs race against each other. We
could try to fix that by either sequencing the jobs in a clever
way (annoying and can break things like rerunning if only parts
failed), or by creating more profiles (unclear if you can even have
two profiles for the same group id, even if you do, it’s annoying to
merge).

So in this PR I (grudgingly) merged both uploads into the Haskell
script. This isn’t all bad:

1. It moves some logic from bash embedded in yaml string literals into
Haskell code.
2. It duplicates some versions but it removes duplication in other
places so overall not too much worse.
3. It does however, make things slower. We don’t run this stuff in
parallel. That said, the release step is relatively small (< 5min) and
it only runs on Linux.

We could add CLI arguments to make the Scala versions configurable for
local development. Given that this is blocking releases, I wanted to
get something in that works first and then see what we need in that regard.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/bazel_tools/scala_version.bzl
+++ b/bazel_tools/scala_version.bzl
@@ -1,6 +1,7 @@
 # Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Keep in sync with /release/src/Main.hs
 default_scala_version = "2.12.12"
 
 def _impl(ctx):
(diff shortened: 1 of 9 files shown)


garyverhaegen-da (approved):

cocreature:
fwiw on CI, the dry run which runs through everything apart from publishing took 39s. Given that, I think the benefits of trying to parallelize this are probably neglible .

PR #8950, opened 2021-02-25
Fix closure references in Daml Repl

cocreature:
Turns out the comment "we probably need to extend this to merge the
modules from each line" was exactly correct: If the result evaluates
to a closure (or a value including a closure), it can reference
definitions from the current module. This happens if the simplifier
lifted something out of the current definition (otherwise we have only
one and it cannot be recursive so no chance of leaking a reference).

This PR fixes this by checking whether the result references the
current module and if it does, we keep it around.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/ReplServiceMain.scala
+++ b/compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/ReplServiceMain.scala
@@ -13,7 +13,7 @@ import com.daml.lf.engine.script._
 import com.daml.lf.language.Ast._
 import com.daml.lf.language.{Ast, LanguageVersion, Util => AstUtil}
 import com.daml.lf.speedy.SExpr._
-import com.daml.lf.speedy.{Compiler, SDefinition, SValue, SExpr, SError}
+import com.daml.lf.speedy.{Compiler, SDefinition, SError, SExpr, SValue}
 import com.daml.grpc.adapter.{AkkaExecutionSequencerPool, ExecutionSequencerFactory}
 import com.daml.ledger.api.refinements.ApiTypes.ApplicationId
 import com.daml.ledger.api.tls.{TlsConfiguration, TlsConfigurationCli}
@@ -24,6 +24,8 @@ import io.grpc.stub.StreamObserver
 import java.net.{InetAddress, InetSocketAddress}
 import java.nio.file.{Files, Path, Paths}
 import java.util.logging.{Level, Logger}
+
+import com.daml.lf.speedy.iterable.SExprIterable
 
 import scala.jdk.CollectionConverters._
 import scala.concurrent.duration._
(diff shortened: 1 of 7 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/iterable/SExprIterable.scala
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf.speedy.iterable
+
+import com.daml.lf.speedy.{SExpr, SValue}
+import scala.jdk.CollectionConverters._
+
+object SExprIterable {
```
cocreature, line 9: @remyhaemmerle-da I don’t like any of this but I don’t know how to do it nicer. If you have suggestions, I’m very open.

aherrmann-da: @cocreature Maybe add a comment explaining how this iterates. As I understand it, it iterates only over the top layer but doesn't recurse into depth.

aherrmann-da (approved):

compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/ReplServiceMain.scala
```diff
@@ -177,6 +179,22 @@ object ReplService {
...
+          if (ref.packageId == homePackageId) {
+            acc + ref.modName
+          } else {
+            acc
+          }
+        case e => SExprIterable(e).foldLeft(acc)(moduleRefs)
+      }
+
+    SExprIterable(v).foldLeft(Set.empty[ModuleName])(moduleRefs)
+  }
```
remyhaemmerle-da, line 197: Can we move that to `com.daml.lf.speedy.SValue` and make the `SExprIterable` `speedy` private ?

remyhaemmerle-da (approved):
:cry:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/iterable/SExprIterable.scala
cocreature: added a comment, yes your observation is exactly right. Works like our other iterables (e.g. `ExprIterable`) or Haskell’s plated stuff

compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/ReplServiceMain.scala
cocreature: I kept this function here but moved the code under `com.daml.lf.speedy.repl.ReplServiceMain` and made the other stuff `private`.

remyhaemmerle-da: Thanks.

Issue #8774, opened 2021-02-08
Avoid submitTree where it is not required


cocreature:
While `submitTree` is the most flexible option, we’re probably not going to need it in 95% of the daml script dumps we produce. To make them easier to read, we should fallback to `submit` (or `submitMulti`) where possible. Given that this is only for readability, we don’t have to try to get every case but at least the common ones:

1. A bunch of bare creates. We can just return the contract ids directly and bind them.
2. An `exercise` which creates exactly one contract that is in the exercise result.

PR #8952, opened 2021-02-25
Avoid submitTree for a sequence of creates

aherrmann-da:
Addresses part of #8774 

This avoids `submitTreeMulti` if a transaction only consists of a sequence of creates, instead it uses `submitMulti`.
It avoids binding contract ids that are not referenced later and keeps the binds of contract ids within the submit body minimal.
This factors out some common functionality between encoding the ACS and encoding transactions.
The tests are extended to cover more relevant cases.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -253,9 +253,11 @@ private[dump] object Encode {
     val cids = ContractId.subst(evs.map(_.contractId))
     val referencedCids = cids.filter(cid => cidRefs.contains(cid))
     val (bind, returnStmt) = referencedCids match {
+      case Seq() if cids.length == 1 =>
+        (Doc.text("_ <- "), Doc.empty)
       case Seq() =>
-        (Doc.empty, Doc.empty)
-      case Seq(cid) if cid == cids.last =>
+        (Doc.empty, Doc.text("pure ()"))
+      case Seq(cid) if cids.length == 1 =>
         val encodedCid = encodeCid(cidMap, cid)
         (encodedCid :+ " <- ", Doc.empty)
       case Seq(cid) =>
@@ -268,8 +270,12 @@ private[dump] object Encode {
     val submit = "submitMulti " +: encodeParties(partyMap, submitters)
     val actions = Doc.stack(evs.map { ev =>
       val cid = ContractId(ev.contractId)
-      val bind = if (returnStmt.nonEmpty && cidRefs.contains(cid)) {
-        encodeCid(cidMap, cid) :+ " <- "
+      val bind = if (returnStmt.nonEmpty) {
+        if (cidRefs.contains(cid)) {
+          encodeCid(cidMap, cid) :+ " <- "
+        } else {
+          Doc.text("_ <- ")
+        }
       } else {
         Doc.empty
       }
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -261,7 +261,7 @@ private[dump] object Encode {
       case _ => (tuple(encodedCids) :+ " <- ", "pure " +: tuple(encodedCids))
     }
     val submit = "submitMulti " +: encodeParties(partyMap, submitters)
-    val actions = Doc.stack(evs.map { case ev =>
+    val actions = Doc.stack(evs.map { ev =>
       val cid = ContractId(ev.contractId)
       val bind = if (returnStmt.nonEmpty && cidRefs.contains(cid)) {
         encodeCid(cidMap, cid) :+ " <- "
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -216,7 +216,6 @@ private[dump] object Encode {
   ): Doc = ev match {
     case TreeEvent.Kind.Created(created) =>
       encodeCreatedEvent(partyMap, cidMap, created)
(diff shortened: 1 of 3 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -67,7 +70,7 @@ private[dump] object Encode {
       (Doc.text("dump Parties{..} = do") /
         Doc.stack(
           sortedAcs.map(createdEvent =>
-            encodeCreatedEvent(partyMap, cidMap, cidRefs, createdEvent)
+            encodeSubmitCreatedEvents(partyMap, cidMap, cidRefs, Seq(createdEvent))
```
aherrmann-da, line 73: We could change this to issue a single `submit` that issues all `createCmd`s to reconstruct the ACS in one go.
However, with a large ACS this might require a pretty large tuple to bind all the contract Ids. I'm not sure if this could cause issues.

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -230,21 +243,34 @@ private[dump] object Encode {
...
-    (bind + Doc.text("submitMulti ") + encodeParties(partyMap, submitters) + Doc.text(
-      " [] do"
-    ) / createCmd).hang(2)
+    val submitters = evs.flatMap(ev => evParties(Kind.Created(ev))).toSet
+    val cids = ContractId.subst(evs.map(_.contractId))
+    val referencedCids = cids.filter(cid => cidRefs.contains(cid))
+    val encodedCids = referencedCids.map(cid => encodeCid(cidMap, cid))
+    val (bind, returnStmt) = referencedCids.length match {
+      case 0 => (Doc.empty, Doc.empty)
+      case 1 if referencedCids.last == cids.last => (encodedCids.last :+ " <- ", Doc.empty)
```
cocreature, line 258: if you pattern match you can avoid the `.last` call here.

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -230,21 +243,34 @@ private[dump] object Encode {
...
+    val submit = "submitMulti " +: encodeParties(partyMap, submitters)
+    val actions = Doc.stack(evs.map { ev =>
+      val cid = ContractId(ev.contractId)
+      val bind = if (returnStmt.nonEmpty && cidRefs.contains(cid)) {
+        encodeCid(cidMap, cid) :+ " <- "
+      } else {
+        Doc.empty
+      }
+      bind + encodeCreatedEvent(partyMap, cidMap, ev)
+    })
```
cocreature, line 271: It looks like there is some potential to factor this into something that works for generic expressions that return applicative results. Doesn’t have to be in this PR but I think it might bmake the logic a bit clearer.

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -254,14 +280,30 @@ private[dump] object Encode {
...
-    val treeBind =
-      (Doc.text("tree <- submitTreeMulti ") + encodeParties(partyMap, submitters) + Doc.text(
-        " [] do"
-      ) / Doc.stack(rootEvs.map(ev => encodeEv(partyMap, cidMap, ev)))).hang(2)
-    val cidBinds = cids.filter(c => cidRefs.contains(c.cid)).map(bindCid(cidMap, _))
-    Doc.stack(treeBind +: cidBinds)
+    val createsOnly = {
+      import scalaz.Scalaz._
+      rootEvs.toList
+        .map {
```
cocreature, line 286: `.map_.sequence` should be equivalent to `.traverse`, maybe also worth factoring into a helper function.

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeCreatedSpec.scala
```diff
@@ -11,51 +11,77 @@ import org.scalatest.matchers.should.Matchers
...
+          ContractId("cid1") -> "contract_0_0",
+          ContractId("cid2") -> "contract_0_1",
+          ContractId("cid3") -> "contract_0_2",
+        )
+        val cidRefs = Set(ContractId("cid1"), ContractId("cid3"))
+        val events = Seq(mkCreated(1), mkCreated(2), mkCreated(3))
+        encodeSubmitCreatedEvents(parties, cidMap, cidRefs, events).render(80) shouldBe
+          """(contract_0_0, contract_0_2) <- submitMulti [alice_0] [] do
+            |  contract_0_0 <- createCmd Module.Template
+            |  createCmd Module.Template
```
cocreature, line 62: should we encode this as `_ <- createCmd …`? I don’t think we turn this warning on by default but it’s generated code so we might as well generate some that doesn’t trigger it.

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeTreeSpec.scala
```diff
@@ -46,10 +47,127 @@ class EncodeTreeSpec extends AnyFreeSpec with Matchers {
...
+                  signatories = Seq("Alice"),
+                  createArguments = Some(
+                    Record(
+                      recordId = Some(Identifier("package", "Module", "Template")),
+                      fields = Seq.empty,
+                    )
+                  ),
+                )
+              )
+            ),
```
cocreature, line 82: looks like a helper similar to `mkCreated` might come in handy. and probably also a helper to create the wrapping trees.

cocreature (approved):
Nice, thanks!

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeTreeSpec.scala
aherrmann-da: I've created simple `mkCreated` and `mkExercised` helpers. We can think of something more sophisticated going forward.

PR #8962, opened 2021-02-26
Fix daml-sdk-head Maven install

cocreature:
There were two issues:

1. The generated pom.xml used to install everything in one step
referenced the source files in the bazel directory rather than the
released artifacts. This luckily worked so far but falls apart now
since the 2.13 build overwrites the 2.12 build. Same issues that also
blew up the actual release on Thursday.

2. Two artifacts are published as shaded jars. We treated those like
Scala artifacts while they should be treated more like fat
jars. They’re not quite like those since the file paths are different
so I gave them a new release type.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/release/src/Util.hs
+++ b/release/src/Util.hs
@@ -81,6 +81,8 @@ data JarType
       -- ^ A java protobuf library (*-speed.jar).
     | Scala
       -- ^ A scala library jar, with source and scaladoc jars.
+    | JarJar
+      -- ^ A scala library jar, with source and scaladoc jars.
     deriving (Eq, Show)
 
 instance FromJSON ReleaseType where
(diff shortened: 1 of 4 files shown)


release/src/Util.hs
```diff
@@ -81,6 +81,8 @@ data JarType
       -- ^ A java protobuf library (*-speed.jar).
     | Scala
       -- ^ A scala library jar, with source and scaladoc jars.
+    | JarJar
+      -- ^ A scala library jar, with source and scaladoc jars.
```
garyverhaegen-da, line 85: Should this comment line be slightly different from the previous one? As it stands it makes it a bit difficult to understand why we need a separate type.

cocreature: absolutely that was dumb copypasta c83cea73522203e0b8e52776bff69422fabc054b

garyverhaegen-da (approved):
I don't understand this diff as well as I'd like, but I can confirm it fixes `daml-sdk-head` for me.

cocreature:
> I don't understand this diff as well as I'd like, but I can confirm it fixes `daml-sdk-head` for me.

Happy to explain anything in here, I definitely did not intend for it to be confusing :slightly_smiling_face:

PR #8969, opened 2021-02-28
Do not require a JWT token for Health and Reflection services [DPP-277]

kamil-da:
**Motivation**
Currently a JWT token is required for all services if a JWT authorisation is enabled. We would like to loosen this restriction and not require providing a token for Health and Reflection services.

**Main changes**
- ~~Replaced the usage of `Claims.empty` with a new `ClaimSet.Unauthorized` for rejecting a request immediately.~~
- ~~`ClaimSet.Claims` with empty claim list does not cause immediate request rejection. Its used for authorisation validation in appropriate services. Even empty claim list in `ClaimSet.Claims` may result in a successful service response if no explicit restrictions on required claims is specified for a service (this exact mechanism is used for the Health and Reflection services)~~
- Introduced `ClaimSet` trait with `Unauthenticated` object and `Claims` case class. `AuthServiceJWT#decodeMetadata()` returns now `ClaimSet`. `Unauthenticated` is used to signal that the token is missing or was malformed.
- `AuthorizationInterceptor` inserts decoded `ClaimSet` in the request context no matter if the decoded set is `Unauthorized` or `Claims`. The early rejection in `AuthorizationInterceptor` with `UNAUTHENTICATED` status has been removed. This means that all requests (even those without a token) get past this interceptor.
- Authorisation logic is shifted to the service layer and is handled by `*Authorization` services which use `Authorizer` internally. Services now can inspect the context and decide how to handle a request basing on the `ClaimSet` from the request context. Specifically, for `Health` service we handle requests with either `ClaimSet.Unauthenticated` or `ClaimSet.Claims` but for secured services, we require `ClaimSet.Claims` with appropriate claims inside.

**Side comment**
I left testing the `AuthServiceJWT` for the next PR but I'm open to suggestions on that.

```
CHANGELOG_BEGIN
- A JWT token is no longer required to call methods of Health and Reflection services
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Authorizer.scala
+++ b/ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Authorizer.scala
@@ -24,7 +24,7 @@ final class Authorizer(now: () => Instant, ledgerId: String, participantId: Stri
   /** Validates all properties of claims that do not depend on the request,
     * such as expiration time or ledger ID.
     */
-  private def valid(claims: Claims): Either[AuthorizationError, Unit] =
+  private def valid(claims: ClaimSet.Claims): Either[AuthorizationError, Unit] =
     for {
       _ <- claims.notExpired(now())
       _ <- claims.validForLedger(ledgerId)
@@ -168,7 +168,10 @@ final class Authorizer(now: () => Instant, ledgerId: String, participantId: Stri
         )
     }
 
-  private def ongoingAuthorization[Res](scso: ServerCallStreamObserver[Res], claims: Claims) =
+  private def ongoingAuthorization[Res](
+      scso: ServerCallStreamObserver[Res],
+      claims: ClaimSet.Claims,
+  ) =
     new OngoingAuthorizationObserver[Res](
       scso,
       claims,
@@ -180,7 +183,7 @@ final class Authorizer(now: () => Instant, ledgerId: String, participantId: Stri
     )
 
   private def authorize[Req, Res](call: (Req, ServerCallStreamObserver[Res]) => Unit)(
-      authorized: Claims => Either[AuthorizationError, Unit]
+      authorized: ClaimSet.Claims => Either[AuthorizationError, Unit]
   ): (Req, StreamObserver[Res]) => Unit =
     (request, observer) => {
       val scso = assertServerCall(observer)
@@ -206,7 +209,7 @@ final class Authorizer(now: () => Instant, ledgerId: String, participantId: Stri
     }
 
   private def authorize[Req, Res](call: Req => Future[Res])(
-      authorized: Claims => Either[AuthorizationError, Unit]
+      authorized: ClaimSet.Claims => Either[AuthorizationError, Unit]
   ): Req => Future[Res] =
     request =>
       AuthorizationInterceptor
--- a/ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Authorizer.scala
+++ b/ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Authorizer.scala
@@ -7,12 +7,13 @@ import java.time.Instant
 
 import com.daml.ledger.api.auth.interceptor.AuthorizationInterceptor
 import com.daml.ledger.api.v1.transaction_filter.TransactionFilter
-import com.daml.platform.server.api.validation.ErrorFactories.permissionDenied
+import com.daml.platform.server.api.validation.ErrorFactories.{permissionDenied, unauthenticated}
 import io.grpc.stub.{ServerCallStreamObserver, StreamObserver}
 import org.slf4j.LoggerFactory
 
 import scala.collection.compat._
 import scala.concurrent.Future
+import scala.util.{Failure, Success, Try}
 
 /** A simple helper that allows services to use authorization claims
   * that have been stored by [[AuthorizationInterceptor]].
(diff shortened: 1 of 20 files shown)


ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Claims.scala
```diff
@@ -46,119 +46,126 @@ final case class ClaimActAsParty(name: Ref.Party) extends Claim
...
+    * | LedgerConfigurationService          | GetLedgerConfiguration     | isPublic                                 |
+    * | PackageService                      | *                          | isPublic                                 |
+    * | PackageManagementService            | *                          | isAdmin                                  |
+    * | PartyManagementService              | *                          | isAdmin                                  |
+    * | ResetService                        | *                          | isAdmin                                  |
+    * | TimeService                         | GetTime                    | isPublic                                 |
+    * | TimeService                         | SetTime                    | isAdmin                                  |
+    * | TransactionService                  | LedgerEnd                  | isPublic                                 |
+    * | TransactionService                  | *                          | for each requested party p: canReadAs(p) |
+    * +-------------------------------------+----------------------------+------------------------------------------+
```
meiersi-da, line 78: Why not add the `HealthService` to this list so later developers don't forget about it?

meiersi-da (commented):
This is not a complete review, as I'm not top of this codebase. I merely peeked because I was interested. Apart from the one comment, I've made, I was looking for tests that check that the HealthService endpoint is accessible without JWT in a JWT configured server. @kamil-da can you point me to them?

ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Claims.scala
kamil-da: Good catch! I'll add the missing items to the list.

kamil-da:
> This is not a complete review, as I'm not top of this codebase. I merely peeked because I was interested. Apart from the one comment, I've made, I was looking for tests that check that the HealthService endpoint is accessible without JWT in a JWT configured server. @kamil-da can you point me to them?

They are being prepared at the moment :)

cocreature (commented):
I’ll leave the detailed review to someone else but I have a question: Did you consider bypassing the auth interceptor completely for services that do not require authorization instead of making the interceptor handle this case?

kamil-da:
> I’ll leave the detailed review to someone else but I have a question: Did you consider bypassing the auth interceptor completely for services that do not require authorization instead of making the interceptor handle this case?

Do you mean enabling interceptor only for specific services?
I am thinking about making interceptor common for all services and producing either `Unauthenticated` or `Claims` and let services decide (by using `Authorizer`) how to handle that. What do you think?

cocreature:
>  Do you mean enabling interceptor only for specific services?

Exactly. I’m not sure if that’s better or worse, it really was just a question since you clearly thought more about this than I have. It seems like it might be less invasive of a change than this one but that doesn’t mean it’s better in the long run.

kamil-da:
> > Do you mean enabling interceptor only for specific services?
> 
> Exactly. I’m not sure if that’s better or worse, it really was just a question since you clearly thought more about this than I have. It seems like it might be less invasive of a change than this one but that doesn’t mean it’s better in the long run.

I haven't found a way to intercept specific services other than write a dedicated interceptor aware of such filtering. This doesn't seem right to me - I would prefer keep service logic separated from the authorisation machinery.

ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/interceptor/AuthorizationInterceptor.scala
```diff
@@ -53,12 +51,8 @@ final class AuthorizationInterceptor(protected val authService: AuthService, ec:
...
             call.close(internalAuthenticationError, new Metadata())
             new ServerCall.Listener[Nothing]() {}
-          case Success(Claims.empty) =>
-            logger.debug(s"Auth metadata decoded into empty claims, returning UNAUTHENTICATED")
-            call.close(Status.UNAUTHENTICATED, new Metadata())
-            new ServerCall.Listener[Nothing]() {}
-          case Success(claims) =>
-            val nextCtx = prevCtx.withValue(contextKeyClaim, claims)
+          case Success(claimSet) =>
+            val nextCtx = prevCtx.withValue(AuthorizationInterceptor.contextKeyClaimSet, claimSet)
```
mziolekda, line 55: We need to return a `ServerCall.Listener[Nothing]()` when the authentication failed, instead of a listener with context

kamil-da: Please correct me if I'm wrong but this way we would lose all the information from the context. The idea of changes in this PR is to deliberately allow even unauthenticated calls at the interceptor level, enrich the context with such information and let service authorisation wrappers decide if the `ClaimSet` from the context of a call is sufficient to call a method.
Side note: Maybe it would be worth renaming `AuthorizationInterceptor` to `AuthenticationInterceptor`.

ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Claims.scala
kamil-da: Fixed.

ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/auth/UnsecuredServiceCallAuthTests.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.daml.platform.sandbox.SandboxRequiringAuthorization
+import com.daml.platform.sandbox.services.SandboxFixture
+import io.grpc.stub.AbstractStub
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.must.Matchers
+import org.scalatest.Assertion
+
+import scala.concurrent.Future
+
+trait UnsecuredServiceCallAuthTests
```
stefanobaghino-da, line 17: Very nice. 👏🏻

stefanobaghino-da (commented):
Nicely integrated in existing tests, good job. I'd prefer having someone else directly involved approving, but from my point of view this looks fine. 🙇🏻

ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/auth/UnsecuredServiceCallAuthTests.scala
kamil-da: Thank you! I've just refactored those traits a bit to reduce code duplication. Please have a look if you're interested.

ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/interceptor/AuthorizationInterceptor.scala
mziolekda: You are right about that.

ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Authorizer.scala
```diff
@@ -179,15 +183,27 @@ final class Authorizer(now: () => Instant, ledgerId: String, participantId: Stri
...
     (request, observer) => {
       val scso = assertServerCall(observer)
-      AuthorizationInterceptor
-        .extractClaimsFromContext()
+      authenticatedClaimsFromContext()
         .fold(
-          observer.onError(_),
+          ex => {
+            logger.debug(
+              s"No authenticated claims found in the request context. Returning UNAUTHENTICATED"
```
mziolekda, line 203: Thanks for retaining the logs. This is very important for our ability to troubleshoot authorization issues.

mziolekda (approved):
Good job.

PR #8978, opened 2021-03-01
Batch creation of ACS

### LINKED ISSUE #8954: Batch creation of ACS
When starting daml script dumps from a given acs offset, we should provide an option to batch creates by a given batch size. Ledgers usually get a bit unhappy for thousands of creates in a single transaction (and batch size 1 is too slow).

aherrmann-da:
Closes #8954 

Let's the user configure the size by which to batch create commands into a single transaction when recreating the ACS in a Daml script dump.
Adds test cases to cover the various edge-cases of batch size compared to ACS size.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
@@ -14,6 +14,7 @@ final case class Config(
     parties: List[String],
     start: LedgerOffset,
     end: LedgerOffset,
+    acsBatchSize: Int,
     outputPath: Path,
     sdkVersion: String,
     damlScriptLib: String,
@@ -55,6 +56,14 @@ object Config {
       .text(
         "The transaction offset (inclusive) for the end position of the dump. Optional, by default the dump includes the current end of the ledger."
       )
+    opt[Int]("acs-batch-size")
+      .optional()
+      .action((x, c) => c.copy(acsBatchSize = x))
+      .validate(x =>
+        if (x <= 0) { failure("ACS batch size must be greater than zero") }
+        else { success }
+      )
+      .text("Batch this many create commands into one transaction when recreating the ACS.")
     opt[File]('o', "output")
       .required()
       .action((x, c) => c.copy(outputPath = x.toPath))
@@ -69,6 +78,7 @@ object Config {
     parties = List(),
     start = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)),
     end = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_END)),
+    acsBatchSize = 1,
     outputPath = null,
     sdkVersion = "",
     damlScriptLib = "daml-script",
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
@@ -78,7 +78,7 @@ object Config {
     parties = List(),
     start = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)),
     end = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_END)),
-    acsBatchSize = 1,
+    acsBatchSize = 10,
     outputPath = null,
     sdkVersion = "",
     damlScriptLib = "daml-script",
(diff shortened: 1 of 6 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -285,6 +287,27 @@ private[dump] object Encode {
...
 
+  private[dump] def encodeACS(
+      partyMap: Map[Party, String],
+      cidMap: Map[ContractId, String],
+      cidRefs: Set[ContractId],
+      acs: Seq[CreatedEvent],
+      batchSize: Int,
+  ): Doc = {
+    val acsSize = acs.length
+    val numBatches: Int = if (acsSize % batchSize == 0) {
```
cocreature, line 298: can we use [grouped](https://www.scala-lang.org/api/current/scala/collection/Iterator.html#grouped[B%3E:A](size:Int):Iterator.this.GroupedIterator[B]) here?

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -68,13 +69,11 @@ private[dump] object Encode {
...
       Doc.text("dump : Parties -> Script ()") /
       (Doc.text("dump Parties{..} = do") /
-        Doc.stack(
-          sortedAcs.map(createdEvent =>
-            encodeSubmitCreatedEvents(partyMap, cidMap, cidRefs, Seq(createdEvent))
-          ) ++
-            trees.map(t => encodeTree(partyMap, cidMap, cidRefs, t))
-        ) /
-        Doc.text("pure ()")).hang(2)
+        stackNonEmpty(
```
cocreature, line 72: Why do we have to filter empty docs now? I’m not really sure which of them can be empty.

daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
```diff
@@ -55,6 +56,14 @@ object Config {
       .text(
         "The transaction offset (inclusive) for the end position of the dump. Optional, by default the dump includes the current end of the ledger."
       )
+    opt[Int]("acs-batch-size")
+      .optional()
+      .action((x, c) => c.copy(acsBatchSize = x))
+      .validate(x =>
+        if (x <= 0) { failure("ACS batch size must be greater than zero") }
```
cocreature, line 63: maybe 0 would be a nice way to disable batching?

cocreature (approved):
Thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
aherrmann-da: Thanks, I was looking for this but somehow missed it.

daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
aherrmann-da: Depends what you mean by disabling batching. If you mean one command per submit, then that's already equivalent to a batch size of 1. If you mean batch all commands into one submit, then yes, we can do that. However, as discussed in https://github.com/digital-asset/daml/pull/8952#discussion_r582953935 that would likely cause issues when more of the ACS contracts are referenced than fit into a tuple. So, I'm not sure if we should provide this option.

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
aherrmann-da: `Doc.stack` will include empty items in it's arguments as empty lines, with `hang` annoyingly prefixed by whitespace. If we don't generate an ACS, i.e. `--start start`, then `encodeACS` will return an empty doc. So, without the filter the dump would include an empty line at the top. Before, this wasn't an issue because we used `++` on sequences. So, if `sortedAcs` was empty, then the sequence passed to `Doc.stack` would just not include any ACS lines.

daml-script/dump/src/main/scala/com/daml/script/dump/Config.scala
```diff
@@ -69,6 +78,7 @@ object Config {
     parties = List(),
     start = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_BEGIN)),
     end = LedgerOffset(LedgerOffset.Value.Boundary(LedgerOffset.LedgerBoundary.LEDGER_END)),
+    acsBatchSize = 1,
```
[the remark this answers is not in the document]
aherrmann-da: 100 may be too large considering https://github.com/digital-asset/daml/pull/8952#discussion_r582953935

PR #8979, opened 2021-03-01
feature: damlc lints all files in project directory

### LINKED ISSUE #8887: daml damlc lint should support linting a whole project
Currently we accept only a single file which is pretty annoying. We should make the flie optional and default to linting the whole project.

robin-da:
fixes #8887.

This changes the input options of `damlc lint` to take several files. If
no file is given, all `.daml` files contained in the directory are
linted.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc.hs
@@ -548,22 +548,22 @@ execLint inputFiles opts =
        withProjectRoot' projectOpts $ \relativize ->
        do
          loggerH <- getLogger opts "lint"
-         inputFiles <- getInputFiles relativize inputFiles
-         forM_ inputFiles $ \inputFile -> do
-           opts <- setDlintDataDir opts
-           withDamlIdeState opts loggerH diagnosticsLogger $ \ide -> do
-               setFilesOfInterest ide (HashSet.singleton inputFile)
-               runActionSync ide $ getDlintIdeas inputFile
-               diags <- getDiagnostics ide
-               if null diags then
-                 hPutStrLn stderr "No hints"
-               else
-                 exitFailure
+         opts <- setDlintDataDir opts
+         withDamlIdeState opts loggerH diagnosticsLogger $ \ide -> do
+             inputFiles <- getInputFiles relativize inputFiles
+             setFilesOfInterest ide (HashSet.fromList inputFiles)
+             diags <- forM inputFiles $ \inputFile -> do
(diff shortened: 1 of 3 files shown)


compiler/damlc/lib/DA/Cli/Damlc.hs
```diff
@@ -539,25 +539,32 @@ execCompile inputFile outputFile opts (WriteInterface writeInterface) mbIfaceDir
...
+               setFilesOfInterest ide (HashSet.singleton inputFile)
+               runActionSync ide $ getDlintIdeas inputFile
+               diags <- getDiagnostics ide
+               if null diags then
+                 hPutStrLn stderr "No hints"
+               else
+                 exitFailure
+     getInputFiles = \case
+       [] -> do
+           fs <- listFilesInside (pure . not . isPrefixOf ".daml" . takeFileName) "."
```
cocreature, line 565: this doesn’t seem quite right. We only want files in the source directory from your daml.yaml. We can probably share the logic that `daml build` uses to find those.

robin-da: using `getDamlRootFiles` now.

compiler/damlc/lib/DA/Cli/Damlc.hs
```diff
@@ -539,25 +539,31 @@ execCompile inputFile outputFile opts (WriteInterface writeInterface) mbIfaceDir
...
-         withDamlIdeState opts loggerH diagnosticsLogger $ \ide -> do
-             setFilesOfInterest ide (HashSet.singleton inputFile)
-             runActionSync ide $ getDlintIdeas inputFile
-             diags <- getDiagnostics ide
-             if null diags then
-               hPutStrLn stderr "No hints"
-             else
-               exitFailure
+         inputFiles <- getInputFiles relativize inputFiles
+         forM_ inputFiles $ \inputFile -> do
```
cocreature, line 552: can we move the loop into `withDamlIdeState` and only set files of interest once to all of them? I think that should be significantly faster. You probably also want to always lint all files and only print "No hints" once (or call `exitFailure`).

robin-da: moved it to the main loop

cocreature (approved):
Thanks! Would be great to have a test for this somewhere. Probably easiest to just create a new test suite.

robin-da:
@cocreature cool, I'll add one.

PR #8981, opened 2021-03-01
Move non-repudiation in-memory backend to testing module

stefanobaghino-da:
This will allow to share it with the non-repudiation API tests.

changelog_begin
changelog_end

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #8982, opened 2021-03-01
Add non-repudiation API

### LINKED ISSUE #8658: Non-repudiation: add a persistent key store with an upload interface
The current non-repudiation prototype has a volatile key store. Add a persistent key store and expose a way to upload new keys from remote processes (some kind of interface over HTTP should be sufficient as a first step).

stefanobaghino-da:
The Web API for the non-repudiation proxy allows to retrieve the signed
payload for a given command, add a new certificate or query for an
existing one.

Closes #8658

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da:
Note: the tests currently only test the happy path, more tests for error conditions will be added before moving this out of draft (along with rebasing after #8981 is merged).

runtime-components/non-repudiation-api/src/test/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpointSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+        algorithm = AlgorithmString.wrap(algorithm),
+        fingerprint = FingerprintBytes.wrap(fingerprint),
+        payload = PayloadBytes.wrap(payload),
+        signature = SignatureBytes.wrap(signature),
+        timestamp = Instant.ofEpochMilli(timestamp),
+      )
+
+    SignedPayloadsEndpoint.toResponse(signedPayload) shouldBe SignedPayloadsEndpoint.Response(
+      algorithm = algorithm,
+      fingerprint = "ZmluZ2VycHJpbnQ=",
```
cocreature, line 41: I assume that’s base64 encoded? maybe worth adding a comment to make the test a bit clearer (or just do the base64 encoding but then this test is pretty close to just replicating the logic which is also not super useful).

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/NonRepudiationApi.scala
```diff
@@ -0,0 +1,87 @@
...
+      signedPayloadRepository,
+    ).map(_ => ())
+
+  final case class Configuration(
+      interface: String,
+      port: Int,
+      shutdownTimeout: FiniteDuration,
+  )
+
+  // Non-repudiation -> NR -> N = 78 and R = 82 in ASCII
```
cocreature, line 40: omg, is that how all our weird port numbers are constructed?

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
```diff
@@ -0,0 +1,117 @@
...
+import spray.json.{DefaultJsonProtocol, RootJsonFormat}
+
+import scala.util.Try
+
+private[api] final class CertificatesEndpoint private (certificates: CertificateRepository)
+    extends CertificatesEndpoint.JsonProtocol {
+
+  import CertificatesEndpoint._
+
+  private def putCertificate(certificate: X509Certificate): Route =
```
cocreature, line 24: what happens if I add the same certificate twice?

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+  private def getCertificate(fingerprint: FingerprintBytes): Route =
+    handleExceptions(logAndReport(logger)(UnableToRetrieveTheCertificate)) {
+      certificates.get(fingerprint).fold(reject)(certificate => complete(encode(certificate)))
+    }
+
+  private val route: Route =
+    concat(
+      put {
```
cocreature, line 37: Related question to the one above, if we’re not idempotent, we probably want post. Couldn’t find a test to check whether we are.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
```diff
@@ -0,0 +1,117 @@
...
+
+object CertificatesEndpoint {
+
+  def apply(certificates: CertificateRepository): Route =
+    new CertificatesEndpoint(certificates).route
+
+  private val logger: Logger = LoggerFactory.getLogger(classOf[CertificatesEndpoint])
+
+  private[api] val InvalidCertificateString: String =
+    "Invalid upload, the 'certificate' field must contain a URL-safe base64-encoded X.509 certificate"
```
cocreature, line 61: what does url-safe mean here? I thought this is in the post body not a query param.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpoint.scala
```diff
@@ -0,0 +1,75 @@
...
+      algorithm = signedPayload.algorithm,
+      fingerprint = base64Url(signedPayload.fingerprint),
+      payload = base64Url(signedPayload.payload),
+      signature = base64Url(signedPayload.signature),
+      timestamp = signedPayload.timestamp.toEpochMilli,
+    )
+
+  private[api] trait JsonProtocol extends SprayJsonSupport with DefaultJsonProtocol {
+
+    protected implicit val apiSignedPayloadFormat: RootJsonFormat[Response] =
```
cocreature, line 70: for the JSON API and the trigger service we follow a common format of `result` and `error` fields and a `status` field in all responses. Have you considered adapting that here as well? If not, what do you think of doing so? totally fine not to do this at all or do it in a separate PR.

cocreature (approved):
Nice, thanks!

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/NonRepudiationApi.scala
stefanobaghino-da: Wait. I assumed you knew about it.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: https://tools.ietf.org/html/rfc4648#page-7

stefanobaghino-da: I assumed this was common knowledge but I can add a link to the RFC if you think that makes sense.

stefanobaghino-da: (the assumption was towards the user, not towards you 😉)

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpoint.scala
stefanobaghino-da: Sure, I was not aware of this. What is the `status` field though?

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: Good question. My intention was to be idempotent but right know it's likely to fail on the backend, I'll add a test for this and fix it.

runtime-components/non-repudiation-api/src/test/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpointSpec.scala
stefanobaghino-da: Base64, URL-safe, I'll add a comment, thanks for pointing this out.

I preferred to not replicate the logic but rather have a "golden sample" of some sort. I was mostly interesting in making sure that this was not utterly broken.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: > I thought this is in the post body not a query param.

It would be `/v1/certificate/<fingerprint>`.

stefanobaghino-da: Whoops, sorry, I misread it. Yes, it's in the body. I just preferred to use a single encoding, does that make sense?

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpoint.scala
cocreature: It’s the status code. It makes more sense in the JSON API since annoyingly it can be different from the http status code in some cases (we stream some responses and we don’t know the status code when sending the header, fun side note: that’s why grpc uses http trailers instead of headers afaik).

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
cocreature: Sorry for being unclear. I wasn’t asking what url-safe means, I was asking why we care about it if it’s not in the url and therefore also why we base64 encode it.

I’m not opposed to using base64 encoding here as well but I think you can make a reasonable argument that the ux is slightly nicer for sending the unencoded contents in something like a multi-part form request, e.g., you can pass it via `-F "parametername=@filename"` to curl whereas you have to call `base64` yourself.

How about the following approach for now:
1. Remove the comment on url-safe since it is misleading given that it is not in a url (the rest of the comment can stay)
2. Stick with base64 for now. We’re early access so we could break it if we want to and even if we don’t and we decide the multi-part is sufficiently nicer, keeping both around is relatively cheap.

stefanobaghino-da: Ok.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpoint.scala
stefanobaghino-da: Ok.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: Mh, that would mean, however, that you cannot be guaranteed to be able to lookup a certificate (`GET /v1/certificate/<fingerprint>`) using the fingerprint returned when uploading (`PUT /v1/certificate`). Are you sure this is fine?

stefanobaghino-da: After some thinking, I'm not 100% sure about this. It would mean we would have places where we use base64 encoding and other where we use base64 URL-safe encoding (`GET /v1/certificate/<fingerprint>`). Are you sure it makes sense to use two different encodings? In particular, if we decide to use base64 encoding in all cases where a URL is not involved, it would mean we would return a base64 encoded fingerprint when uploading a certificate but then we would require a base64 URL-safe encoding to retrieve it.

cocreature: I think I’m confused, I thought this was only about the format used when uploading the certificate. I fully agree that the fingerprint returned in the response to that should be the same format as the one in the URL.

stefanobaghino-da: Sure, I was wondering if it makes sense to use two different encodings, even though it's not strictly needed to be URL-safe for certificates. Since we're forced to use URL-safe encoding for fingerprints, does it make sense to use a different encoding for certificates? I'm mostly trying to understand what an API user would think would make more sense.

cocreature: I guess I view the certificate as binary data which I already have (the certificate lies around somewhere locally) while I view the fingerprint as an opaque string that is returned and produced by the API. So I think it’s not entirely unreasonable to have different encodings for them. That said, I think two different base64 encodings are confusing as you mentioned and doesn’t solve anything So I think we have the following two options:

1. Use url-safe base64 encoding everywhere, i.e., do what you have. Probably just needs a comment somewhere on why we do this for cases where something is not in a url.
2. Use url-safe base64 encoding for things that are strings and/or used in urls, e.g., fingerprints. Upload binary data like certificates without base64 encoding them first, e.g., as multipart form requests.

I would suggest to go with 1 for now and keep the option open to later add/switch to 2 but I don’t feel strongly about this so if you disagree, happy to go with your approach instead.

stefanobaghino-da: Ok, I'll go with number 1 for now then and keep URL-safe base64 encoding everywhere. At this point changing this behavior at a later point is really not difficult. Thanks for the feedback!

runtime-components/non-repudiation-api/src/test/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpointSpec.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8982/commits/cdd891222d7f92ba6a0da3aca5a3ab95a13f0045

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/SignedPayloadsEndpoint.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8982/commits/9faaed6d13deedf47fd8f07b7ddd4dfce425dcb7

I'm not 100% sure that we should include the status code in this API, but we can review the decision at a later point.

Following the `{ "error": "..." }`/`{ "result": { ... } }` pattern actually makes sense, it will hopefully allow users to reuse some code, or at least to be familiar with the output format.

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/8982/commits/e7ef80b57672191ea6bc907dc5032d0d6b39271c

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: Follow-up in https://github.com/digital-asset/daml/pull/8982#discussion_r585666749

runtime-components/non-repudiation-api/src/main/scala/com/daml/nonrepudiation/api/v1/CertificatesEndpoint.scala
stefanobaghino-da: This requires a few changes to the PostgreSQL part, I'll merge this as is and open a separate PR to guarantee that the back-end guarantees idempotency.

stefanobaghino-da: Tracked in https://github.com/digital-asset/daml/issues/9021

PR #8983, opened 2021-03-02
Speedy: building transactions with rollback

nickchapman-da:
Speedy: building transactions with rollback

- call `beginTry`, `endTry`, `rollbackTry` and `abortTry` during speedy _on-ledger_ execution
- test: by checking the expected contracts are created by the transaction

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1397,8 +1397,31 @@ private[lf] object SBuiltin {
         case SOptional(opt) =>
           opt match {
             case None =>
+              machine.ledgerMode match {
+                case OffLedger => ()
+                case onLedger: OnLedger =>
+                  onLedger.ptx = onLedger.ptx.abortTry
+              }
               unwindToHandler(machine, payload) //re-throw
             case Some(handler) =>
+              machine.ledgerMode match {
+                case OffLedger => ()
+                case onLedger: OnLedger =>
+                  payload match {
+                    case SAnyException(typ, _, sv) =>
+                      // TODO https://github.com/digital-asset/daml/issues/8020
+                      // Must convert speedy value to LF value, but currently this crashes on SBuiltinException
+                      // so make a hack workaround:
+                      val value =
+                        sv match {
+                          case _: SBuiltinException => V.ValueInt64(999)
+                          case _ => sv.toValue
+                        }
+                      onLedger.ptx = onLedger.ptx.rollbackTry(typ, value)
+                    case _ =>
+                      crash(s"SBTryHandler, expected payload to be SAnyException: $payload")
+                  }
+              }
               machine.enterApplication(handler, Array(SEValue(token)))
           }
         case v =>
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1395,34 +1395,28 @@ private[lf] object SBuiltin {
       checkToken(token)
       args.get(0) match {
         case SOptional(opt) =>
-          opt match {
-            case None =>
-              machine.ledgerMode match {
-                case OffLedger => ()
-                case onLedger: OnLedger =>
-                  onLedger.ptx = onLedger.ptx.abortTry
-              }
-              unwindToHandler(machine, payload) //re-throw
-            case Some(handler) =>
-              machine.ledgerMode match {
-                case OffLedger => ()
-                case onLedger: OnLedger =>
-                  payload match {
-                    case SAnyException(typ, _, sv) =>
-                      // TODO https://github.com/digital-asset/daml/issues/8020
-                      // Must convert speedy value to LF value, but currently this crashes on SBuiltinException
-                      // so make a hack workaround:
-                      val value =
-                        sv match {
-                          case _: SBuiltinException => V.ValueInt64(999)
-                          case _ => sv.toValue
-                        }
-                      onLedger.ptx = onLedger.ptx.rollbackTry(typ, value)
-                    case _ =>
-                      crash(s"SBTryHandler, expected payload to be SAnyException: $payload")
-                  }
-              }
-              machine.enterApplication(handler, Array(SEValue(token)))
+          machine.withOnLedger("SBTryHandler") { onLedger =>
+            opt match {
+              case None =>
+                onLedger.ptx = onLedger.ptx.abortTry
+                unwindToHandler(machine, payload) //re-throw
+              case Some(handler) =>
+                payload match {
+                  case SAnyException(typ, _, sv) =>
+                    // TODO https://github.com/digital-asset/daml/issues/8020
+                    // Must convert speedy value to LF value, but currently this crashes on SBuiltinException
+                    // so make a hack workaround:
+                    val value =
+                      sv match {
+                        case _: SBuiltinException => V.ValueInt64(999)
+                        case _ => sv.toValue
+                      }
+                    onLedger.ptx = onLedger.ptx.rollbackTry(typ, value)
+                  case _ =>
+                    crash(s"SBTryHandler, expected payload to be SAnyException: $payload")
+                }
+                machine.enterApplication(handler, Array(SEValue(token)))
+            }
           }
         case v =>
           crash(s"invalid argument to SBTryHandler (expected SOptional): $v")
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
@@ -378,6 +378,11 @@ object SExpr {
     def execute(machine: Machine): Unit = {
       machine.pushKont(KTryCatchHandler(machine, handler))
       machine.ctrl = body
+      machine.ledgerMode match {
+        case OffLedger => ()
+        case onLedger: OnLedger =>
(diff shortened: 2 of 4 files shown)


remyhaemmerle-da (commented):
Nice !

remyhaemmerle-da (approved):

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1397,8 +1397,31 @@ private[lf] object SBuiltin {
         case SOptional(opt) =>
           opt match {
             case None =>
+              machine.ledgerMode match {
+                case OffLedger => ()
```
cocreature, line 1401: This seems off, we don’t expect to ever hit this do we? I think this should extend `OnLedgerBuiltin` and crash in off-ledger mode.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
```diff
@@ -378,6 +378,11 @@ object SExpr {
     def execute(machine: Machine): Unit = {
       machine.pushKont(KTryCatchHandler(machine, handler))
       machine.ctrl = body
+      machine.ledgerMode match {
+        case OffLedger => ()
```
cocreature, line 382: same here, I think this should extend `OnLegderBuiltin` and crash here.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  private def typeAndCompile(pkg: Package): PureCompiledPackages = {
+    val rawPkgs = Map(defaultParserParameters.defaultPackageId -> pkg)
+    Validation.checkPackage(rawPkgs, defaultParserParameters.defaultPackageId, pkg)
+    data.assertRight(
+      PureCompiledPackages(rawPkgs, Compiler.Config.Default.copy(stacktracing = FullStackTrace))
+    )
+  }
+
+  private def runUpdateExpr_GetTx(pkgs1: PureCompiledPackages)(e: Expr): SubmittedTransaction = {
```
cocreature, line 34:
```suggestion
  private def runUpdateExprGetTx(pkgs1: PureCompiledPackages)(e: Expr): SubmittedTransaction = {
```
unless there is a good reason to deviate from Scala naming conventions, let’s stick to them.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
```diff
@@ -0,0 +1,185 @@
...
+        template (record : T1) = {
+          precondition True,
+          signatories Cons @Party [(M:T1 {party} record)] (Nil @Party),
+          observers Nil @Party,
+          agreement "Agreement",
+          choices {
+          }
+        };
+
+        val create0 : Party -> Scenario Unit = \(party: Party) ->
```
cocreature, line 65: is there actually a reason why any of these have to go via scenarios? Given that we’d like to get rid of them evenutally, I’d rather avoid building more tests on top of them. You should be able to just call update expressions directly and afterwards call `ptx.finish`

cocreature (commented):
lgtm with a few minor comments. Thanks!

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
remyhaemmerle-da: In general,  I prefer our first level of test depend on the compiler. when test depend on the compiler: 
1 - It make debugging much more involve 
2 - You do not control as well the LF produces, it is maybe not an issue for you compiler guys that have a clear idea how this are compile, but it is not for me. 
3 - A test that tackles a particular feature will survive even if the compiler does not handle it anymore. 

To sum up, in my point of view we should have unit test that do not depends on the compiler **and** integration test.

For sure, we should not rewrite this test now, since it rely on invariant over the cid seeding that will very problably go away.

nickchapman-da: I guess not. It was just convenient. I'll try and change to avoid Scenarios.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
nickchapman-da: Yes, I think you are right.
For some reason I was thinking the exception handling could happen on or off ledger.
But I support the fact that `try`..`catch` is typed as an `Update` prevent this.
I'll change to crash as you suggest.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
remyhaemmerle-da: I aggree with @moritzkiefer-da. 
However I will not rewrite them now, as the invariants (on cid seeding) they are based will probably go away soon. 
Let's postpone that until we produce actual rollback node.

cocreature: We don’t rely on contract ids in the test do we? We rely on the integer value in the contract which should be independent of any changes here.

remyhaemmerle-da: @cocreature. You are right.

cocreature: thanks for making the change!

nickchapman-da: I didn't see this conversation, and already changed to avoid Scenarios.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
nickchapman-da: yup

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: It will happen off ledger but it won’t happen via the speedy builtin.

nickchapman-da: What. You can have _off_ ledger execution of `Update` expressions?
I'm confused now!

cocreature: No you can’t have off-ledger execution of `Update`. But you can throw an exception outside of `Update` and we will add a `Script`-specific catch (which won’t go via a speedy builtin, see https://github.com/digital-asset/daml/pull/7368 for an ancient prototype).

nickchapman-da: Ok. that's clear.  A throw can happen anywhere. But the `try-catch` (`Update` expression) can only be on ledger.

And it is the speedy execution of the `try-catch` which performs the `{begin,end,abort,rollback}Try`, which is where the `OnLedger` is required, and so may be assumed.

nickchapman-da: done!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1395,11 +1395,28 @@ private[lf] object SBuiltin {
       checkToken(token)
       args.get(0) match {
         case SOptional(opt) =>
-          opt match {
-            case None =>
-              unwindToHandler(machine, payload) //re-throw
-            case Some(handler) =>
-              machine.enterApplication(handler, Array(SEValue(token)))
+          machine.withOnLedger("SBTryHandler") { onLedger =>
```
cocreature, line 1398: is there a reason why we cannot extend `OnLedgerBuiltin`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
```diff
@@ -378,6 +378,9 @@ object SExpr {
     def execute(machine: Machine): Unit = {
       machine.pushKont(KTryCatchHandler(machine, handler))
       machine.ctrl = body
+      machine.withOnLedger("SETryCatch") { onLedger =>
```
cocreature, line 381: same here

cocreature (approved):
Nice! One minor question left

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
nickchapman-da: Yes, probably could here.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
nickchapman-da: But not here (without some reorganisation) because this is an expression, not a builtin.

PR #8990, opened 2021-03-02
Dpp 267 indexer parallel ingestion poc

nmarton-da:
### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only.sql
```diff
@@ -0,0 +1,543 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+/*
+------------------------------------------------------------------- OLD
+
+CREATE TABLE public.configuration_entries (
```
oliverse-da, line 7: Note that we can't assume that the tables are in the public schema . cc @sormeter

oliverse-da: But I guess you know that since the `public.` prefix only seems to appear in comments (from pg_dump?)

nmarton-da: Precisely ;) The comments are only there for convenience (since it is Flyway, we don't have diff on full schema), mostly for the work already done, and a little for clarity on what changed exactly.

Also please note: this is PoC code, this "migration" will just renders previous data corrupt, so for prod rollout a sound strategy must be chosen, and implemented properly.

ledger/participant-integration-api/src/main/scala/platform/indexer/poc/BatchingParallelIngestionPipe.scala
```diff
@@ -0,0 +1,156 @@
...
+      runStageUntil: Int,
+  ) // 1 - nothing   2 - input mapping   3 - seq mapping   4 - db ingestion   5 - tailing   6 - persist params
+  (source: Source[IN, NotUsed]): Source[Unit, NotUsed] = {
+    assert(runStageUntil >= 1 && runStageUntil <= 6)
+    val stage1 = source
+    val stage2 = stage1
+      .groupedWithin(
+        submissionBatchSize.toInt,
+        FiniteDuration(batchWithinMillis, "millis"),
+      ) // TODO .batch adds no latency to the pipe, but batch and mapAsync combination leads to dirac impulses in the forming batch sizes, which leads practically single threaded ingestion throughput at this stage.
```
tudor-da, line 37: `..leads to dirac impulses.. ` It's been a while since I read this. Fancy analogy though

ledger/participant-integration-api/src/main/scala/platform/indexer/poc/BatchingParallelIngestionPipe.scala
```diff
@@ -0,0 +1,156 @@
...
+//      .drop(1) // remove the zero element from the beginning of the stream
+//      // Stage 4: Inserting data into the database. Almost no CPU load here, threads are executing SQL commands over JDBC, and waiting for the result. This defines the parallelism on the SQL database side, same amount of PostgreSQL Backend processes will do the ingestion work.
+//      .async
+//      .mapAsync(ingestingParallelism)(ingester)
+//      // Stage 5: Preparing data sequentially for throttled mutations in database (tracking the ledger-end, corresponding sequential event ids and latest-at-the-time configurations)
+//      .conflate(tailer)
+//      .throttle(tailingRateLimitPerSecond, FiniteDuration(1, "seconds"))
+//      // Stage 6: Updating ledger-end and related data in database (this stage completion demarcates the consistent point-in-time)
+//      .mapAsync(1)(ingestTail)
+//      .map(_ => ()) // TODO what should this out? linking to in-mem fan-out?
```
meiersi-da, line 77: Yes, I'd expect that we do that link-up and the link-up to the mutable state cache provided they run in the same process.

FYI: @tudor-da -- I believe we'll want to have these link-ups given the findings you've had wrt deserialization cost from the DB. They should though be optional, at least in architecture, so we can support the split-indexer architeture.

meiersi-da (commented):
@nmarton-da : I've done a second high-level pass, building on the knowledge that you shared in the online-session with me a week ago.

From my perspective, this code is a great blueprint for planning the production roll-out. Thanks for the good work!

ledger/participant-integration-api/src/main/scala/platform/indexer/poc/BatchingParallelIngestionPipe.scala
nmarton-da: This point is particularly useful for tapping on the move of the ledger end.
It would require some work to have here the reconstructed batches of updates as well, but entirely doable.
Would be nice to collect potential consumers of a ledger-end stream, and come up with a union requirement, put a interface on top of it, and expose that as well as we instantiate the indexer, so this broadcast stream can be injected to other, non-dependent consumers.

ledger/participant-integration-api/src/main/scala/platform/indexer/poc/UpdateToDBDTOV1.scala
```diff
@@ -0,0 +1,296 @@
...
+object UpdateToDBDTOV1 {
+
+  def apply(
+      participantId: ParticipantId,
+      translation: LfValueTranslation,
+      compressionStrategy: CompressionStrategy,
+  ): Offset => Update => Iterator[DBDTOV1] = { offset =>
+    {
+      case u: Update.CommandRejected =>
+        // TODO we might want to tune up deduplications so it is also a temporal query @simon@?
```
rautenrieth-da, line 25: As far as I remember, deduplication was noticeable in the metrics (hence the effort to get rid of participant-side command deduplication), so any speedup will be nice.

Just note that the deduplication query (`com.daml.ledger.participant.state.index.v2.IndexSubmissionService.deduplicateCommand()`) needs to be atomic - if it returns that a command is new, it needs to write the deduplication key to the DB and no other concurrent call can return that the command is new. Think about what SQL transaction isolation level we use, and whether the new queries are still atomic.

nmarton-da: A general note first: I think deduplication is a different thing than indexing, and therefore I would applaud any future or current efforts to separate that entirely out from this db schema/code etc.

Since efforts already made to make this happen by @andreaslochbihler-da, I am not sure how much effort we should put in here to harden the implementation.

With all that said I do not understand what that means that the `deduplaceCommand()` needs to be atomic: Isn't it that this is not part of indexing, but the submission processing on participant write side? -- that behavior is not changed.

At indexing we delete these deduplication entries only, races there apply the same fashion how it was before I would say. But to be sure let's discuss this offline as well! 

cc
@meiersi-da

rautenrieth-da: I had another look at the command deduplication in this PR. It executes the same SQL query as the current production code when deleting deduplication keys for failed transactions, and it does not change deduplication queries for the write path. So there is no issue there.

As discussed offline, I agree that deduplication is independent from the index database and should be factored out of LedgerDao. We just have to remember that deduplication is expected to work in horizontally scaled participants (i.e., where multiple Ledger API servers access the same index database fed by a single indexer) and across ledger API server crash recoveries. Implementing deduplication in LedgerDao was done only for convenience (access to a persistent shared data store with atomic operations).

PR #8993, opened 2021-03-03
Properly async pipeline stages in JdbcIndexer

gerolf-da:
`mapAsync` does not introduce asynchronous boundaries, to the pipeline stages aren't truly separate. The asynchronous boundaries are now introduced via `.async`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
@@ -6,7 +6,6 @@ package com.daml.platform.indexer
 import java.time.{Duration, Instant}
 
 import akka.NotUsed
-import akka.stream.OverflowStrategy
 import akka.stream.scaladsl.Flow
 import com.codahale.metrics.Timer
 import com.daml.daml_lf_dev.DamlLf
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
```diff
@@ -429,11 +428,14 @@ class PipelinedExecuteUpdate(
   private[indexer] val flow: ExecuteUpdateFlow =
     Flow[OffsetUpdate]
       .mapAsync(updatePreparationParallelism)(prepareUpdate)
-      .buffer(16, OverflowStrategy.backpressure)
+      .async
       .map(PipelinedUpdateWithTimer(_, metrics.daml.index.db.storeTransaction.time()))
       .mapAsync(1)(insertTransactionState)
+      .async
```
tudor-da, line 434: Do you know what's the improvement of adding these async barriers? I considered adding them a while ago, but decided against since I didn't believe they could bring anything when we have a buffer before the insertion stages.

tudor-da (approved):

ledger/participant-integration-api/src/main/scala/platform/indexer/ExecuteUpdate.scala
gerolf-da: Without the async barriers, the three mapAsync stages are being fused together, causing some additional overhead that is not reflected in the individual metrics.

PR #8995, opened 2021-03-03
Use single party submit/submitTree where appropriate

aherrmann-da:
Part of #8774 

If the list of submitters is a singleton set we can use `submit`/`submitTree` instead of `submitMulti`/`submitTreeMulti`.
This extends the test cases to cover multi-party submissions.

### Pull Request Checklist

trigger the build.


daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -335,6 +333,15 @@ private[dump] object Encode {
 
   private def encodeSelector(selector: Selector): Doc = Doc.str(selector.i)
 
+  private def submit(partyMap: Map[Party, String], submitters: Set[Party], isTree: Boolean): Doc = {
```
aherrmann-da, line 336: I'm not too happy with this `isTree` parameter. However, for now it seems like an improvement over the previous state.

cocreature (approved):
Thanks!

PR #9003, opened 2021-03-03
compat: fix killing stale sandbox(es)

garyverhaegen-da:
We've seen this happen once; I don't think it's worth investigating.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

stefanobaghino-da (approved):

PR #9023, opened 2021-03-04
Use DA.Map in triggers if available

aherrmann-da:
As discussed internally, this transitions the trigger library to use `DA.Map` instead of the deprecated `DA.Next.Map`, if available. This is to avoid deprecation warnings if users import the triggers library. This is a breaking change, but was deemed the best available option.

changelog_begin
- [Triggers] The trigger library now uses `DA.Map` instead of the deprecated `DA.Next.Map` if the targeted Daml-LF version supports it. This is a breaking change: Code that interfaced with the triggers library using `DA.Next.Map`, e.g. with `Daml.Trigger.getCommandsInFlight` or `Daml.Trigger.Assert.testRule`, will need to be changed to use `DA.Map` instead.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/upgrade/example/carbon-upgrade-trigger/daml/UpgradeTrigger.daml
+++ b/docs/source/upgrade/example/carbon-upgrade-trigger/daml/UpgradeTrigger.daml
@@ -1,11 +1,16 @@
 -- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
 
 module UpgradeTrigger where
 
 import DA.Assert
 import DA.Foldable
+#ifdef DAML_GENMAP && DAML_GENERIC_COMPARISON
+import qualified DA.Map as Map
+#else
 import qualified DA.Next.Map as Map
+#endif
 import Daml.Trigger
 import Daml.Trigger.Assert
 import qualified Daml.Script as Script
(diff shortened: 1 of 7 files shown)


cocreature (approved):
Nice thank you! We should probably also figure out how to fix `-Werror` (iirc it fails with cpp since it gets interpreted as a file name) so we can avoid this type of stuff sneaking in accidentally in the future.

cocreature:
```

**Comment by cocreature** [comment]
Looks like one of the docs examples needs updating
```
  docs/source/upgrade/example/carbon-upgrade-trigger/daml/UpgradeTrigger.daml:57:54: error:
  • Couldn't match expected type ‘DA.Internal.LF.Map
Id [Command]’
  with actual type ‘Map.Map k0 v0’
  NB: ‘Map.Map’ is defined in ‘DA.Next.Map’
  ‘DA.Internal.LF.Map’ is defined in ‘DA.Internal.LF’
  • In the fourth argument of ‘testRule’, namely ‘Map.empty’
  In a stmt of a 'do' block:
  (_, commands) <- testRule upgradeTrigger alice acs Map.empty ()
  In the first argument of ‘script’, namely
  ‘do alice <- Script.allocateParty "Alice"
  bob <- Script.allocateParty "Bob"
  certProposal <- submit alice
  $ Script.createCmd (CarbonCertProposal alice bob 10)
  cert <- submit bob
  $ Script.exerciseCmd certProposal CarbonCertProposal_Accept
```

PR #9024, opened 2021-03-04
Make non-repudiation PostgreSQL back-end certificate adding idempotent

### LINKED ISSUE #9021: Non-repudiation: adding a certificate must be idempotent
This decision can be reviewed in the future, but now for simplicity the certificate back-end should behave live a simple in-memory map and accept repeated certificate uploads with an idempotent behavior.

Tests should exist to check this behavior.

_Originally posted by @cocreature in https://github.com/digital-asset/daml/pull/8982#discussion_r585666749_

stefanobaghino-da:
Closes #9021

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/runtime-components/non-repudiation-postgresql/src/main/scala/com/daml/nonrepudiation/postgresql/PostgresqlCertificateRepository.scala
+++ b/runtime-components/non-repudiation-postgresql/src/main/scala/com/daml/nonrepudiation/postgresql/PostgresqlCertificateRepository.scala
@@ -25,7 +25,7 @@ final class PostgresqlCertificateRepository(transactor: Transactor[IO])(implicit
     getCertificate(fingerprint).query[X509Certificate].option.transact(transactor).unsafeRunSync()
 
   private def putKey(fingerprint: FingerprintBytes, certificate: X509Certificate): Fragment =
-    fr"insert into" ++ table ++ fr"""(fingerprint, certificate) values ($fingerprint, $certificate)"""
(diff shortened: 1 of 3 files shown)


stefanobaghino-da:
@cocreature I was thinking of generalizing the API tests to work both as unit tests against the in-memory back-end and integration tests with the PostgreSQL back-end. I would do it as a follow-up PR as it's a bit involved. Sounds good?

runtime-components/non-repudiation-postgresql/src/main/scala/com/daml/nonrepudiation/postgresql/PostgresqlCertificateRepository.scala
```diff
@@ -25,7 +25,7 @@ final class PostgresqlCertificateRepository(transactor: Transactor[IO])(implicit
     getCertificate(fingerprint).query[X509Certificate].option.transact(transactor).unsafeRunSync()
 
   private def putKey(fingerprint: FingerprintBytes, certificate: X509Certificate): Fragment =
-    fr"insert into" ++ table ++ fr"""(fingerprint, certificate) values ($fingerprint, $certificate)"""
+    fr"insert into" ++ table ++ fr"""(fingerprint, certificate) values ($fingerprint, $certificate) on conflict do nothing"""
```
cocreature, line 28: Answering my own question: We don’t have to worry about someone trying to insert a certificate with the same fingerprint but a different certificate because we compute the fingerprint ourselve and don’t let users specify it.

cocreature (approved):
Thanks!

PR #9036, opened 2021-03-05
Add tests for client bindings

### LINKED ISSUE #8636: Non-repudiation: allow supported bindings to use non-repudiation
The existing bindings clients do not allow to add the necessary interceptors that add support to submit commands over a non-repudiation proxy. Extend supported bindings to do so and possibly also those used internally (e.g. Scala bindings and client). The most flexible way to do this is probably to allow callers to pass in arbitrary interceptors.

stefanobaghino-da:
changelog_begin
changelog_end

Closes #8636

### Pull Request Checklist

trigger the build.

### DIFF
--- a/runtime-components/non-repudiation-client/src/test/scala/com/daml/nonrepudiation/client/SigningInterceptorSpec.scala
+++ b/runtime-components/non-repudiation-client/src/test/scala/com/daml/nonrepudiation/client/SigningInterceptorSpec.scala
@@ -1,0 +1,103 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.nonrepudiation.client
+
+import java.time.{Clock, Instant, ZoneId}
+import java.util.Collections
+
+import com.daml.ledger.api.v1.CommandServiceOuterClass.SubmitAndWaitRequest
+import com.daml.ledger.javaapi.data.Command
+import com.daml.ledger.api.v1.CommandsOuterClass.{Command => ProtoCommand}
+import com.daml.ledger.api.v1.command_service.CommandServiceGrpc.CommandService
+import com.daml.ledger.api.v1.command_submission_service.CommandSubmissionServiceGrpc.CommandSubmissionService
+import com.daml.ledger.rxjava.DamlLedgerClient
+import com.daml.nonrepudiation.testing._
+import com.daml.nonrepudiation.{
+  AlgorithmString,
+  CommandIdString,
+  NonRepudiationProxy,
+  SignatureBytes,
+  SignedPayload,
+}
+import com.daml.resources.grpc.{GrpcResourceOwnerFactories => Resources}
+import io.grpc.netty.NettyChannelBuilder
+import org.scalatest.Inside
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.DurationInt
+
+final class SigningInterceptorSpec extends AsyncFlatSpec with Matchers with Inside {
+
+  behavior of "SigningInterceptor"
+
+  it should "be possible to use it with the Java bindings" in {
+    val (key, certificate) = generateKeyAndCertificate()
+    val builders =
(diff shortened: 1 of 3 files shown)


stefanobaghino-da:
@cocreature After some thinking I realized that thanks to the fact that our bindings accept a channel builder as a parameter, we can simply add interceptors there, without adding anything explicitly to our bindings. Do you think this should be enough for now? I guess we can add something explicitly for interceptors, if there is a request for it.

cocreature:
@stefanobaghino-da sounds like it’s enough for now, didn’t know you could add interceptors there

stefanobaghino-da:
> @stefanobaghino-da sounds like it’s enough for now, didn’t know you could add interceptors there

Yeah, it completely slipped my mind until I had a better look. 😅

runtime-components/non-repudiation-client/src/test/scala/com/daml/nonrepudiation/client/SigningInterceptorSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+          CommandService.scalaDescriptor.fullName,
+          CommandSubmissionService.scalaDescriptor.fullName,
+        )
+      } yield proxy
+
+    proxy.use { _ =>
+      val clientChannelBuilder =
+        builders.proxyChannel
+          .asInstanceOf[NettyChannelBuilder]
+          .intercept(new SigningInterceptor(key, certificate))
```
cocreature, line 70: What happens if I use something other than the command service? From a quick look, I couldn’t find anything that prevents the interceptor from kicking in. iirc we only insert it for command services on the non-repudation server side so it probably doesn’t break anything but unnecessarily signing things also doesn’t seem that great of an idea.

At the very least, I think we should have a test for accessing other services (e.g., one of acs, transaction service, …). Totally fine to do this in another PR.

cocreature (approved):
Thanks! One comment on non-command services

runtime-components/non-repudiation-client/src/test/scala/com/daml/nonrepudiation/client/SigningInterceptorSpec.scala
stefanobaghino-da: Very good point, thanks. Probably we actually need something specific.

stefanobaghino-da: I've addressed this with https://github.com/digital-asset/daml/pull/9050. After that is merged, I can rebase this on `main`, limit signing to commands and add a test for this.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9036/commits/dec170bad4fb2d3a790723ea5343d0f3acba08d3

runtime-components/non-repudiation-client/src/test/scala/com/daml/nonrepudiation/client/SigningInterceptorSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+          .asInstanceOf[NettyChannelBuilder]
+          .intercept(SigningInterceptor.signCommands(key, certificate))
+      val client = DamlLedgerClient.newBuilder(clientChannelBuilder).build()
+      client.connect()
+      val command = ProtoCommand.parseFrom(generateCommand().getCommands.commands.head.toByteArray)
+      val expectedWorkflowId = "workflow-id"
+      val expectedApplicationId = "application-id"
+      val expectedCommandId = "command-id"
+      val expectedParty = "party-1"
+      client.getCommandClient
```
cocreature, line 78: Can we get a test that hits a non-command endpoint to make sure nothing breaks?

stefanobaghino-da: Oops, sorry, I forgot to push a commit.

stefanobaghino-da: https://github.com/digital-asset/daml/pull/9053

PR #9038, opened 2021-03-05
send CI failures on separate channel

garyverhaegen-da:
This will send CI failures to `#ci-failures-daml`.

CHANGELOG_BEGIN
CHANGELOG_END


garyverhaegen-da:
Note that this still sends "release ready for testing" to #team-daml. Open to changing that if that was not the right call.

cocreature (approved):
Thanks! I think keeping release notifications in the main channel is sensible. We can always revisit later.

stefanobaghino-da (approved):
🙇🏻‍♂️

PR #9039, opened 2021-03-05
reduce noise on daily tests

garyverhaegen-da:
Getting a separate message per test was fine when there was one, but this kind of got our of hand at this point.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks! Fully agree with that

PR #9042, opened 2021-03-05
ci/cron/check: low-hanging perf improvement

garyverhaegen-da:
Two quick improvements I made while waiting on #9039:
- Avoid loading Java. When looking at the logs flow by this seemed to be taking a huge amount of time.
- Isolate the gcloud config files, which allows for running gcloud downloads in parallel.

Together these reduce the `check_releases` runtime from about 5 hours to about 2. There's much more (and smarter) work needed on this, but this was really easy to do.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -344,6 +344,7 @@ verify_signatures :: FilePath -> FilePath -> String -> IO ()
 verify_signatures bash_lib tmp version_tag = do
     System.callCommand $ unlines ["bash -c '",
         "set -euo pipefail",
+        "export DADE_SKIP_JAVA=1",
         "eval \"$(dev-env/bin/dade assist)\"",
         "source \"" <> bash_lib <> "\"",
         "shopt -s extglob", -- enable !() pattern: things that _don't_ match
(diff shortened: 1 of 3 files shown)


garyverhaegen-da:
Oh and this code has been extracted from #9039 and has already run [a successful compat build](https://dev.azure.com/digitalasset/daml/_build/results?buildId=71260&view=results), so hopefully it's fully debugged already. 🤞

ci/cron/src/Main.hs
```diff
@@ -344,6 +344,7 @@ verify_signatures :: FilePath -> FilePath -> String -> IO ()
 verify_signatures bash_lib tmp version_tag = do
     System.callCommand $ unlines ["bash -c '",
         "set -euo pipefail",
+        "export DADE_SKIP_JAVA=1",
```
cocreature, line 347: instead of trying to get this right everywhere in here, can we just call `dade-assist` once outside of the script and then skip it? I don’t understand why we need to call it separately for each of those invocations.

cocreature (approved):
lgtm with one question on whether we need to call dade-assist more than once.

ci/cron/src/Main.hs
garyverhaegen-da: Actually, we already do call `dade-assist` before running this. 🤔

PR #9047, opened 2021-03-08
LF: fix archive decoder for GREATER_DATE

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/archive/DecodeV1Spec.scala
+++ b/daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/archive/DecodeV1Spec.scala
@@ -714,9 +714,9 @@ class DecodeV1Spec
 
     s"reject comparison builtins as is if version >= ${LV.Features.genComparison}" in {
 
-      forEveryVersionSuchThat(_ < LV.Features.genComparison) { version =>
+      forEveryVersionSuchThat(_ >= LV.Features.genComparison) { version =>
         val decoder = moduleDecoder(version)
-        forEvery(genericComparisonBuiltinCases) { (proto, _) =>
+        forEvery(comparisonBuiltinCases) { (proto, _) =>
           a[ParseError] shouldBe thrownBy(decoder.decodeExpr(toProtoExpr(proto), "test"))
         }
       }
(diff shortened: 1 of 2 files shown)


daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/archive/DecodeV1Spec.scala
```diff
@@ -714,9 +714,9 @@ class DecodeV1Spec
 
     s"reject comparison builtins as is if version >= ${LV.Features.genComparison}" in {
 
-      forEveryVersionSuchThat(_ < LV.Features.genComparison) { version =>
+      forEveryVersionSuchThat(_ >= LV.Features.genComparison) { version =>
```
cocreature, line 717: :facepalm:

daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/archive/DecodeV1Spec.scala
```diff
@@ -714,9 +714,9 @@ class DecodeV1Spec
...
         val decoder = moduleDecoder(version)
-        forEvery(genericComparisonBuiltinCases) { (proto, _) =>
+        forEvery(comparisonBuiltinCases) { (proto, _) =>
```
cocreature, line 719: I’m confused, where did we rename that? Are we still using `genericComparisonBuiltinCases`?

remyhaemmerle-da: The test ` s"reject comparison builtins as is if version >= ${LV.Features.genComparison}" `  was just wrong. 
Its body was copied from `"translate generic comparison builtins as is if version < ${LV.Features.genComparison}"` but it was never adapted it. 
You can check the body of those tests were excatly the same before my fix.

cocreature: ah right, thanks for the explanaiton.

cocreature (approved):
Good catch!

PR #9048, opened 2021-03-08
ci/cron/check: remove dade-assist calls

garyverhaegen-da:
We can only run this in a context where the dev-env is already set up anyway, as that's how we get Bazel to build the script in the first place.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/src/Main.hs
+++ b/ci/cron/src/Main.hs
@@ -344,8 +344,6 @@ verify_signatures :: FilePath -> FilePath -> String -> IO ()
 verify_signatures bash_lib tmp version_tag = do
     System.callCommand $ unlines ["bash -c '",
         "set -euo pipefail",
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Thanks!

ci/cron/src/Main.hs
```diff
@@ -344,8 +344,6 @@ verify_signatures :: FilePath -> FilePath -> String -> IO ()
 verify_signatures bash_lib tmp version_tag = do
     System.callCommand $ unlines ["bash -c '",
         "set -euo pipefail",
-        "export DADE_SKIP_JAVA=1",
```
cocreature, line 347: should we remove the skip java logic again given that it’s now unused?

garyverhaegen-da: Good point.

PR #9051, opened 2021-03-08
LF: Experiment Hook for 1.dev

remyhaemmerle-da:
This PR adds a hook in the compiler and the engine to allow easy prototyping of new features. 

In particular by using this hook, one can add a new feature in `1.dev` with modifying: 
- compiler 
- type checkers (both Haskell and Scala)
- archive Protobuf
- archive decoder 

In addition of development speed, this also adds a bit of confidence, as the peaces enumerated above are not touch when adding the feature: The feature can be added by modifying only the Dams standard library and Speedy. 

[aa7991f8](https://github.com/digital-asset/daml/pull/9052/commits/aa7991f89977b8987070d8ff8733c3200d9ec58d)  shows a use case of this hook. 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
+++ b/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
@@ -15,6 +15,7 @@ module DA.Daml.LF.Ast.Recursive(
     ) where
 
 import Data.Functor.Foldable
+import qualified Data.Text as T
 
 import DA.Daml.LF.Ast.Base
 
@@ -49,6 +50,7 @@ data ExprF expr
   | EToAnyExceptionF !Type !expr
   | EFromAnyExceptionF !Type !expr
   | EThrowF !Type !Type !expr
+  | EExperimentalF !T.Text !Type
   deriving (Foldable, Functor, Traversable)
 
 data BindingF expr = BindingF !(ExprVarName, Type) !expr
@@ -177,7 +179,7 @@ instance Recursive Expr where
     ETyLam      a b   -> ETyLamF        a b
     ENil        a     -> ENilF          a
     ECons       a b c -> EConsF         a b c
-    ECase       a b   -> ECaseF         a (map projectCaseAlternative b)
+    ECase       a b   -> ECaseF         a (Prelude.map projectCaseAlternative b)
     ELet        a b   -> ELetF          (projectBinding a) b
     EUpdate     a     -> EUpdateF       (projectUpdate a)
     EScenario   a     -> EScenarioF     (projectScenario a)
@@ -190,6 +192,7 @@ instance Recursive Expr where
     EToAnyException a b -> EToAnyExceptionF a b
     EFromAnyException a b -> EFromAnyExceptionF a b
     EThrow a b c -> EThrowF a b c
+    EExperimental a b -> EExperimentalF a b
 
 instance Corecursive Expr where
   embed = \case
@@ -210,7 +213,7 @@ instance Corecursive Expr where
     ETyLamF      a b   -> ETyLam        a b
     ENilF        a     -> ENil          a
     EConsF       a b c -> ECons         a b c
-    ECaseF       a b   -> ECase         a (map embedCaseAlternative b)
+    ECaseF       a b   -> ECase         a (Prelude.map embedCaseAlternative b)
     ELetF        a b   -> ELet          (embedBinding a) b
     EUpdateF     a     -> EUpdate       (embedUpdate a)
     EScenarioF   a     -> EScenario     (embedScenario a)
@@ -223,3 +226,4 @@ instance Corecursive Expr where
     EToAnyExceptionF a b -> EToAnyException a b
     EFromAnyExceptionF a b -> EFromAnyException a b
     EThrowF a b c -> EThrow a b c
+    EExperimentalF a b -> EExperimental a b
--- a/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
+++ b/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
@@ -179,7 +179,7 @@ instance Recursive Expr where
     ETyLam      a b   -> ETyLamF        a b
     ENil        a     -> ENilF          a
     ECons       a b c -> EConsF         a b c
-    ECase       a b   -> ECaseF         a (Prelude.map projectCaseAlternative b)
+    ECase       a b   -> ECaseF         a (map projectCaseAlternative b)
     ELet        a b   -> ELetF          (projectBinding a) b
     EUpdate     a     -> EUpdateF       (projectUpdate a)
     EScenario   a     -> EScenarioF     (projectScenario a)
@@ -213,7 +213,7 @@ instance Corecursive Expr where
     ETyLamF      a b   -> ETyLam        a b
     ENilF        a     -> ENil          a
     EConsF       a b c -> ECons         a b c
-    ECaseF       a b   -> ECase         a (Prelude.map embedCaseAlternative b)
+    ECaseF       a b   -> ECase         a (map embedCaseAlternative b)
     ELetF        a b   -> ELet          (embedBinding a) b
     EUpdateF     a     -> EUpdate       (embedUpdate a)
     EScenarioF   a     -> EScenario     (embedScenario a)
--- a/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
+++ b/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
@@ -199,6 +199,8 @@ applySubstInExpr subst@Subst{..} = \case
     ELocation l e -> ELocation
         l
         (applySubstInExpr subst e)
+    EExperimental name ty ->
+        EExperimental name ty
 
 applySubstInAlternative :: Subst -> CaseAlternative -> CaseAlternative
 applySubstInAlternative subst (CaseAlternative p e) =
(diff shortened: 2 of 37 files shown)


daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/iterable/ExprIterable.scala
```diff
@@ -12,9 +12,9 @@ private[validation] object ExprIterable {
 
   private[iterable] def iterator(x: Expr): Iterator[Expr] = {
     x match {
-      case EVar(_) | EBuiltin(_) | EPrimCon(_) | EPrimLit(_) | EVal(_) | EEnumCon(_, _) | ETypeRep(
-            _
-          ) =>
+      case EVar(_) | EBuiltin(_) | EPrimCon(_) | EPrimLit(_) | EVal(_) | EEnumCon(_, _) |
+          // stupid formatter !
```
cocreature, line 16: :grin:

daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/archive/DecodeV1Spec.scala
```diff
@@ -1113,4 +1113,29 @@ class DecodeV1Spec
       }
     }
   }
+
+  s"reject experiment expression  version <= ${LV.v1_dev}" in {
```
remyhaemmerle-da, line 1117:
```suggestion
  s"reject experiment expression if LF version < ${LV.v1_dev}" in {
```

remyhaemmerle-da:
```suggestion
  s"reject experiment expression version <= ${LV.v1_dev}" in {
```

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
```diff
@@ -177,7 +179,7 @@ instance Recursive Expr where
     ETyLam      a b   -> ETyLamF        a b
     ENil        a     -> ENilF          a
     ECons       a b c -> EConsF         a b c
-    ECase       a b   -> ECaseF         a (map projectCaseAlternative b)
+    ECase       a b   -> ECaseF         a (Prelude.map projectCaseAlternative b)
```
cocreature, line 182: What’s up with that? You imported `Text` qualified (and it should be) so I don’t see why this needs to be qualified.

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Recursive.hs
```diff
@@ -210,7 +213,7 @@ instance Corecursive Expr where
     ETyLamF      a b   -> ETyLam        a b
     ENilF        a     -> ENil          a
     EConsF       a b c -> ECons         a b c
-    ECaseF       a b   -> ECase         a (map embedCaseAlternative b)
+    ECaseF       a b   -> ECase         a (Prelude.map embedCaseAlternative b)
```
cocreature, line 216: same here

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Version.hs
```diff
@@ -157,6 +157,13 @@ featureExceptions = Feature
     , featureCppFlag = Just "DAML_EXCEPTIONS"
     }
 
+featureExperimental :: Feature
+featureExperimental = Feature
+    { featureName = "DAML Experimental"
+    , featureMinVersion = versionDev
+    , featureCppFlag = Just "DAML_EXPERIMENTAL"
```
cocreature, line 164: Do we need a cpp pragma here? If not, you can also set this to `Nothing`

compiler/damlc/daml-stdlib-src/DA/Experimental/Example.daml
```diff
@@ -0,0 +1,13 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+module DA.Experimental.Example(
```
cocreature, line 4: This needs at least a `-- | HIDE` to hide it from the documentation but I’m not sure we need to commit this at all?

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/iterable/TypeIterable.scala
```diff
@@ -79,8 +79,8 @@ private[validation] object TypeIterable {
         Iterator(typ) ++
           iterator(value)
       case EVar(_) | EVal(_) | EBuiltin(_) | EPrimCon(_) | EPrimLit(_) | EApp(_, _) | ECase(_, _) |
-          ELocation(_, _) | EStructCon(_) | EStructProj(_, _) | EStructUpd(_, _, _) |
-          ETyAbs(_, _) =>
+          ELocation(_, _) | EStructCon(_) | EStructProj(_, _) | EStructUpd(_, _, _) | ETyAbs(_, _) |
+          EExperimental(_, _) =>
         ExprIterable.iterator(expr0).flatMap(iterator(_))
```
cocreature, line 84: doesn’t `EExperimental` have a type we need to traverse?

cocreature (approved):
lgtm, I’m slightly sceptical of committing the `answer` stuff but I don’t know how to test it without this either so maybe it’s the right thing to do. Apart fom that, I think it might be nice to have some documentation somewhere (sorry if I missed it in this PR) on why we have this and what you would use it for.

remyhaemmerle-da:
@cocreature wrote:
> lgtm, I’m slightly sceptical of committing the `answer` stuff but I don’t know how to test it without this either so maybe it’s the right thing to do. 

I see to option:
 - we drop the `answer` stuff. I am fine to go without test, as it is only for `dev` (and will be forever).
   We just keep the test that checks that the new expression is rejected is rejected for any stable version of LF.
- we find another example (I could move the experimental string builtin)

> Apart fom that, I think it might be nice to have some documentation somewhere (sorry if I missed it in this PR) on why we have this and what you would use it for.

I have like a commit in the description of the PR that show how this can be used.
I will try to add a bit more doc.

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Version.hs
remyhaemmerle-da: I the cpp pragma in `compiler/damlc/daml-stdlib-src/LibraryModules.daml` to be sure DA.Experimental module are visible only in 1.dev.

compiler/damlc/daml-stdlib-src/DA/Experimental/Example.daml
remyhaemmerle-da: I though that because it was not linked by `LibraryModules.daml`, it will not end up in the doc. 
I will add the flag.

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/iterable/TypeIterable.scala
remyhaemmerle-da: :+1:

compiler/damlc/daml-stdlib-src/DA/Experimental/Example.daml
cocreature: Maybe it is but that seems easy to change accidentally so regardless I think the `-- | HIDE` is worth adding for clarity.

cocreature:
> I see to option:

Your choice, I’m fine with keeping it as well.



> I have like a commit in the description of the PR that show how this can be used.
> I will try to add a bit more doc.

The description in the PR is great, maybe just copy that to the protobuf/lf spec/another place.

compiler/damlc/daml-stdlib-src/LibraryModules.daml
```diff
@@ -52,4 +54,9 @@ import DA.Time
 import DA.Traversable
 import DA.Tuple
 import DA.Validation
+
+#ifdef DAML_EXPERIMENTAL
+import DA.Experimental.Example
+#endif
```
sofiafaro-da, line 60: In general I prefer to do this the other way: don't use CPP in LibraryModules, and use it only in DA.Experimantal.Example instead, where you export an empty module if the feature flag is not available. This is how it's done everywhere else in the standard library. (See DA.Map for example.)

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
```diff
@@ -199,6 +199,8 @@ applySubstInExpr subst@Subst{..} = \case
     ELocation l e -> ELocation
         l
         (applySubstInExpr subst e)
+    EExperimental name ty ->
+        EExperimental name ty
```
sofiafaro-da, line 203: You definitely need to substitute in type.

```suggestion
        EExperimental name (applySubstInType subst ty)
```

compiler/damlc/daml-stdlib-src/LibraryModules.daml
remyhaemmerle-da: I dislike the fact that we will end up with empty `DA.Experimental` module in the non dev standard library. 
This is specially hugely for something that will not be in future version.

sofiafaro-da: Ok, makes sense.

daml-lf/spec/experimental.rst
```diff
@@ -1,3 +1,6 @@
+.. Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
 Copyright © 2021, `Digital Asset (Switzerland) GmbH
 <https://www.digitalasset.com/>`_ and/or its affiliates.  All rights
 reserved.
```
sofiafaro-da: The copyright notice is duplicated now (-: (-:

PR #9054, opened 2021-03-09
Move Daml Profiler to EE version of sandbox/sandbox-classic

cocreature:
This splits Sandbox targets into EE/CE targets and exposes the option
in the EE version. The option still exists in the CE option for now
until we have released EE artifacts to not break users that might know
about it without an alternative.

There is also a small test that makes sure that this actually works
since classpaths are dumb and it didn’t work at first.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-assistant/daml-sdk/validate.sh
```diff
@@ -0,0 +1,26 @@
...
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v2 ---
+
+JAVA=$(rlocation "$TEST_WORKSPACE/$1")
+SDK_CE=$(rlocation "$TEST_WORKSPACE/$2")
+SDK_EE=$(rlocation "$TEST_WORKSPACE/$3")
+
+for cmd in sandbox sandbox-classic; do
+    ! $JAVA -jar $SDK_CE $cmd --help | grep -q profile-dir
```
garyverhaegen-da, line 21: Have you seen this fail? Without `-e` my understanding is that Bash will completely ignore all but the last iteration of the second loop.

ledger/sandbox-common/BUILD.bazel
```diff
@@ -57,6 +57,47 @@ da_scala_library(
...
+    exports = [":sandbox-common"],
+    deps = [
+        ":sandbox-common",
+        "//daml-lf/data",
+        "//ledger/caching",
+        "//ledger/ledger-api-common",
+        "//ledger/participant-integration-api",
+        "//ledger/participant-state",
+        "//libs-scala/ports",
+    ],
```
garyverhaegen-da, line 78: Would it be worth extracting those deps to a variable? Do we expect ce & ee to diverge here?

garyverhaegen-da: Actually could this be turned into a loop like `sandbox-classic-{}` above?

sofiafaro-da (approved):
LGTM, thank you!

ledger/sandbox-common/BUILD.bazel
cocreature: They start diverging as soon as we remove the profiler option so yes, I think we can expect this.

garyverhaegen-da: Something like
```
for (edition, tags) in [("ee", []), ("ce", [...])]
```

cocreature: hm, I guess my reasoning was that there is no reason to believe that those two targets share anything apart from the name in general so I’d rather not tie them together artificially.

garyverhaegen-da (approved):

ledger/sandbox-common/BUILD.bazel
garyverhaegen-da: 👍

daml-assistant/daml-sdk/validate.sh
cocreature: so if I replace this with `exit 1` it actually does fail so I think `!` doesn’t do what I thought it does.

cocreature: ahrg apparently `!` negates `set -e`. I hate bash so much.

garyverhaegen-da: That copy-pasted segment is hard to parse. Maybe add an explicit `set -euo pipefail` after it? AFAIK `set` is idempotent. I'm also not quite sure about the priorities of `!` and `|`; it may be worth adding parentheses and/or moving the exclamation mark to the grep command.

garyverhaegen-da: You could expend a few more characters and just be explicit:
```
for cmd in sandbox sandbox-classic; do
    if ! $JAVA -jar $SDK_CE $cmd --help | grep -q profile-dir; then
        exit 1
    fi
done
```

PR #9055, opened 2021-03-09
Upgrade elliptic dependency to address vulnerabilities

cocreature:
This PR bumps our already existing resolution on elliptic to address a
security issue.

I don’t believe in keeping deps at arbitrarily outdated versions so
this PR fully regenerates the yarn.lock files.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/navigator/frontend/package.json
+++ b/navigator/frontend/package.json
@@ -91,7 +91,7 @@
     "whatwg-fetch": "^3.5.0"
   },
   "resolutions": {
-    "**/elliptic": "^6.5.3",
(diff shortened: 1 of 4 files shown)


cocreature:
> Totally agree with the sentiment, but it's a bit unclear to me how much we can trust our CI to catch all possible issues this could cause. Is there any manual test required here?

I reverted the change to Navigator so it actually only upgrades elliptic there because that did break CI. That’s the one I would have tested manually normally. For the other stuff, I trust CI and I’m honestly not sure what a manual test would be.

Issue #8976, opened 2021-03-01
daml package manager


robin-da:
This is an issue for the implementation of a package management command of `daml` assistant which uses `daml` ledger to query and install available packages.

I imagine having a `daml pkg` top-level command and the subcommands described in the following. 

1. The `ledger` stanza `daml.yaml` is extended with an optional `name` field, such that packages can be queried from multiple ledgers and listed under their ledger names. A new field `default` will indicate which ledger to use in commands when several ledger entries are present but no ledger name is given for the command.
2. `update` - query ledgers for available packages and update the local cache with a list of available packages, their name, a short description, and their providing ledger.
3. `list` - list packages in the local cache with package hash for unique identification, package name, description, and location.
4. `add` - adds the package to the `daml.yaml` file under the dependencies stanza. 
5. `install` - fetches all listed packages in `daml.yaml` from the ledger(s) and installs them into the package database for the project, unless they are already installed.

cocreature:
Summarizing our discussion, `daml build` should walk through the following steps:

1. Look at the `daml.yaml` and figure out external dependencies (probably just names in `data-dependencies` that do not end in `.dar` and are not sdk packages). As a first step, we will only support package ids. As a second step, we will support names and versions.
2. Based on that, download all transitive dalfs into `.daml/dependencies` (or something like that). We explicitly sidestep `daml ledger fetch-dar` since it would download transitive dependencies multiple times potentially. Ideally, we also include local dependencies (note that those have more information than just the dalf namely interface files and package conf) and data-dependencies in there to use the same approach for all.
3. In addition to that, write out a lock file. This lock file is required to resolve package name + version to package id and could potentially also serve as a cache to get the dependencies of a dalf without having to decode it again. I think we might want to include local packages in this as well not just remote dependencies to unify the package management story.
4. Once we’ve downloaded/extracted everything, we now proceed similarly to how packaging works now. the dalfs in the database are more or less equivalent to a dalf specified in data-dependencies.

As a general guideline, if my lockfile is up2date and the database is up2date, no network connection should be made. The lockfile should also be generated outside of `.daml` and is expected to be checked in.

In terms of implementation, we decided to roughly proceed in the following steps:
1. Split the current packaging information into a step that extracts DARs and dalfs into the database and a second step which then builds based on that information.
2. Support package ids in daml.yaml.
3. Support package names and versions in daml.yaml.

@robin-da if anything in here doesn’t make sense or sounds different from what we discussed, please correct it.

[...]

PR #9056, opened 2021-03-09
daml pkg: split installation of deps and package db inititialization

robin-da:
This is the next step for the daml package manager program #8976.
This splits the installation of dependencies from the initialization of
the (ghc) package database.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
@@ -63,7 +63,12 @@ dataDepMarkerFile depPath = depPath </> "_data_dependency_"
 --------------------------
 -- | Install all dependencies to the .daml/dependencies directory.
 installDependencies ::
-       NormalizedFilePath -> Options -> PackageSdkVersion -> [String] -> [String] -> IO ()
+   NormalizedFilePath
+   -> Options
+   -> PackageSdkVersion
+   -> [String] -- Package dependencies. Can be base-packages, sdk-packages or filepath.
+   -> [FilePath] -- Data Dependencies. Can be filepath to dars/dalfs.
+   -> IO ()
 installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pDataDeps = do
     deps <- expandSdkPackages (optDamlLfVersion opts) (filter (`notElem` basePackages) pDeps)
     (needsUpdate, newFingerprint) <-
--- a/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
@@ -4,14 +4,10 @@
 module DA.Cli.Damlc.DependencyDb
     ( installDependencies
     , dependenciesDir
-    , queryDalfsFromDependencies
-    , queryDalfsFromDataDependencies
-    , queryMainDalfs
-    , mainsDir
-    , dalfsDir
-    , configsDir
-    , sourcesDir
-    , normalDepsDir
+    , queryDalfs
+    , mainMarker
+    , depMarker
+    , dataDepMarker
     ) where
 
 import qualified "zip-archive" Codec.Archive.Zip as ZipArchive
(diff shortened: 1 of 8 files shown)


compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,204 @@
...
+sourcesDir depPath = depPath </> "sources"
+sdkVersionFile depPath = depPath </> "sdk-version"
+depMarkerFile depPath = depPath </> "_dependency_"
+dataDepMarkerFile depPath = depPath </> "_data_dependency_"
+
+
+
+-- Dependency installation
+--------------------------
+-- | Install all dependencies to the .daml/dependencies directory.
```
cocreature, line 64: Are the arguments filepaths to dars, dalfs, both? If so, let’s use `FilePath` instead of `String` and either way let’s document this.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -323,6 +297,22 @@ settings =
...
+registerDepInPkgDb :: FilePath -> FilePath -> IO ()
+registerDepInPkgDb dbPath depPath = do
+    linkDirRec (configDir depPath) (dbPath </> "package.conf.d")
+    linkDirRec (sourcesDir depPath) dbPath
+    linkDirRec (mainDir depPath) dbPath
+    recachePkgDb dbPath
+
+linkDirRec :: FilePath -> FilePath -> IO ()
+linkDirRec from to = do
+    srcs <- listFilesRecursive from
```
cocreature, line 310: why symlink individual files instead of symlinking the directory? Do we merge multiple directories?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,204 @@
...
+    if isDataDep
+        then write (dataDepMarkerFile depPath) ""
+        else write (depMarkerFile depPath) ""
+    let path = mainDir depPath </> (takeFileName $ ZipArchive.eRelativePath edMain)
+    write path (ZipArchive.fromEntry edMain)
+    forM_ edConfFiles $ \conf ->
+        write
+            (configDir depPath </> (takeFileName $ ZipArchive.eRelativePath conf))
+            (ZipArchive.fromEntry conf)
+    forM_ edSrcs $ \src ->
```
cocreature, line 117: What happens if we don’t extract those? In other words, can we just drop them?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,204 @@
...
+--------------------------
+dependenciesDir :: Options -> NormalizedFilePath -> FilePath
+dependenciesDir opts projRoot =
+    fromNormalizedFilePath projRoot </> projectDependenciesDatabase </>
+    lfVersionString (optDamlLfVersion opts)
+
+fingerprintFile :: FilePath -> FilePath
+fingerprintFile depsDir = depsDir </> "fingerprint.json"
+
+mainDir, dalfsDir, configDir, sourcesDir, sdkVersionFile, depMarkerFile, dataDepMarkerFile :: FilePath -> FilePath
```
cocreature, line 51: I think it would be great to have a detailed comment somewhere here on the layout of the database. I have one main suggestion/question here at the moment:

The way this is currently setup (and that’s absolutely inherited from the previous code), you still operate on DARs. If you have two DARs that overlap in their dependencies, those dependencies will be recorded multiple times.

Can we instead flatten the layout and only store each dependency exactly once? That seems to bring a few advantages:

1. Less wasted space.
2. Each dalf can have exactly one location addressed by its package id.
3. The logic for creating the package db becomes much simpler. It only has a flat list of dalfs to deal with.

wdyt? Worth doing? I don’t mind doing that in a separate PR but I’d like to agree on whether we want to go there or not before merging this PR.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: can be `FilePath` or the names of base-packages, or soon package ids. +1 on the documentation.

robin-da: Changed the type of data-dependencies to `FilePath` and added docs.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: changed this back to copying files, so there it makes sense again.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: `Message: 
  daml/Main.daml:7:1: error:
  Could not find module ‘Daml.Script’
  There are files missing in the ‘daml-script-0.0.0’ package,
  try running 'ghc-pkg check'.
  Use -v (or `:set -v` in ghci) to see a list of the files searched for.
`
I think you need them.

cocreature: Oh does `edSrcs` include iface files and not just source files?

robin-da: yes, and .hie files.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: I switched it to a flat dependency db now, but running into a problem where `daml-script-0.0.0` is a hidden package when compiling `daml-trigger`.

cocreature: sounds like we’re not properly tracking which ones are the main dalfs, i.e., which ones should be exposed?

robin-da: this is resolved now, still battling an issue with data-dependencies. The db is flat now and there is a comment describing the layout at the beginning of the file.

robin-da: data-dependency issue is resolved as well.

robin-da:
@cocreature This is ready, can you have another look?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,251 @@
...
+import System.FilePath
+import System.IO.Extra
+
+-- Dependency Database Layout
+-----------------------------
+-- Here is an exemplary dependecy database:
+-- .
+-- ├── data-deps
+-- │   ├── configs
+-- │   │   └── proj2-0.0.1.conf
```
cocreature, line 45: Can we store them in the directory of the corresponding dalf?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,251 @@
...
+-- │   │   ├── 3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3
+-- │   │   │   └── proj2-0.0.1-3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3.dalf
+-- │   │   └── e9eeb8cf890e5a15bc65031f1a8373a4bb7f89229473dabaa595c25d35768b21
+-- │   │       └── daml-prim-e9eeb8cf890e5a15bc65031f1a8373a4bb7f89229473dabaa595c25d35768b21.dalf
+-- │   ├── mains
+-- │   │   └── 3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3
+-- │   │       └── proj2-0.0.1-3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3.dalf
+-- │   ├── sdk-version
+-- │   └── sources
+-- │       └── proj2-0.0.1-3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3
```
cocreature, line 58: Same qestion here as for `configs`.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,251 @@
...
+-- │   ├── dalfs
+-- │   │   ├── 057eed1fd48c238491b8ea06b9b5bf85a5d4c9275dd3f6183e0e6b01730cc2ba
+-- │   │   │   └── daml-stdlib-DA-Internal-Down-057eed1fd48c238491b8ea06b9b5bf85a5d4c9275dd3f6183e0e6b01730cc2ba.dalf
+-- │   │   ├── 3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3
+-- │   │   │   └── proj2-0.0.1-3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3.dalf
+-- │   │   └── e9eeb8cf890e5a15bc65031f1a8373a4bb7f89229473dabaa595c25d35768b21
+-- │   │       └── daml-prim-e9eeb8cf890e5a15bc65031f1a8373a4bb7f89229473dabaa595c25d35768b21.dalf
+-- │   ├── mains
+-- │   │   └── 3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3
+-- │   │       └── proj2-0.0.1-3c1af853a9bc7d6aa214b98f0e4d2099187bd5fa7ab07852b4165b84a82325d3.dalf
```
cocreature, line 55: Instead of duplicating `main` dalfs, can we add a marker in their corresponding directory, e.g., a file called `_main_`?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -0,0 +1,251 @@
...
+-- │       └── daml-script-0.0.0-bf4a8f6f897b4d9639bab0f384ea7c06a7ad4b6b856cb0b511e1d79c9c46ecf2
+-- │           └── Daml
+-- │               ├── Script
+-- │               │   ├── Free.daml
+-- │               │   ├── Free.hi
+-- │               │   └── Free.hie
+-- │               ├── Script.daml
+-- │               ├── Script.hi
+-- │               └── Script.hie
+-- └── fingerprint.json
```
cocreature, line 84: What happens if a dalf is pulled in as a `dependency` from one DAR and a `data-dependency` from another DAR? Maybe a good way to avoid that ambiguity is to not have separate directories and instead use a marker file?

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -71,22 +68,14 @@ import SdkVersion
...
-createProjectPackageDb :: NormalizedFilePath -> Options -> PackageSdkVersion -> MS.Map UnitId GHC.ModuleName -> [FilePath] -> [FilePath] -> IO ()
-createProjectPackageDb projectRoot (disableScenarioService -> opts) (PackageSdkVersion thisSdkVer) modulePrefixes deps' dataDeps
+createProjectPackageDb :: NormalizedFilePath -> Options -> MS.Map UnitId GHC.ModuleName -> IO ()
+createProjectPackageDb projectRoot (disableScenarioService -> opts) modulePrefixes
   = do
-    deps <- expandSdkPackages (optDamlLfVersion opts) (filter (`notElem` basePackages) deps')
-    (needsReinitalization, depsFingerprint)
-        <- dbNeedsReinitialization projectRoot deps thisSdkVer (show $ optDamlLfVersion opts)
+    allDepFiles <- listFilesRecursive depsDir
+    (needsReinitalization, depsFingerprint) <- dbNeedsReinitialization projectRoot allDepFiles
```
cocreature, line 75: Can we just move listing files into `dbNeedsReinitialization`?

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -110,23 +96,16 @@ createProjectPackageDb projectRoot (disableScenarioService -> opts) (PackageSdkV
...
+      mainDalfs <- queryMainDalfs depsDir dependenciesInPkgDbIds
 
       let dependencyInfo = DependencyInfo
               { dependenciesInPkgDb
               , dalfsFromDependencies
               , dalfsFromDataDependencies
-              , mainUnitIds =
-                    map (decodedUnitId . mainDalf)
-                        (darsFromDataDependencies ++ darsFromDependencies)
+              , mainUnitIds = map decodedUnitId mainDalfs
```
cocreature, line 108: Let’s leave it from this PR but the checks below for inconsistent versions, incompatible lf versions & unit id conflicts look like we can already do them when setting up `.daml/dependencies` instead of only failing here.

cocreature (approved):
Nice work! A few small comments on the layout.

I also wonder if we should filter out the magic stable deps already during installation but even if we do perfectly fine to leave it out of this PR.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: moved it.

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: it's a bit tricky because you need to know the unit id's as registered in the package database for these checks.

cocreature: What’s stopping us from figuring out the unit ids during the install phase?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: that would mean it's not flat anymore and we need to do a traversal when we set up the package db.

cocreature: I guess when I said "flat" I meant that a dalf does not contain its dependencies. Grouping together data for the same dalf (the actual dalf, configs, sources, …) seems fine? Yes we have to do a traversal but we already traverse everything once to find all dalfs, doing another doesn’t seem like an issue (and we could merke them into one).

compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
robin-da: you can do that. I like the idea of the dependencies db to be completely independent of the tricks we need to apply to the package database.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: there's a marker now.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: there's a marker now.

PR #9057, opened 2021-03-09
inline help: scripts JSON API clarification

garyverhaegen-da:
While writing a forum response and trying to use `daml script` against
the JSON API, I was a bit confused by the existing help texts.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #9059, opened 2021-03-09
Fix double push in CommandTracker

gerolf-da:
The command tracker detects timeouts outside the regular pull/push
mechanism of the custom akka graph stage. Basically the timeout
detection jumps the line when emitting outputs on `resultOut`. If it
then processes a regular completion, it tries to push to `resultOut`
even though it hasn't been pulled again in the meantime. Using `emit`
instead of `push` when a completion arrives makes akka take care of
handling the signaling properly.

Fixes DPP-285

CHANGELOG_BEGIN
[Integration Kit, Sandbox, Daml on SQL, Scala Ledger Bindings]:
Fixed a bug that caused the CommandTracker used by the synchronous
CommandService to crash in case of a lot of timeouts.
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -167,6 +167,13 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
       )
 
       private def pushResultOrPullCommandResultIn(compl: Option[Ctx[Context, Completion]]): Unit = {
+        // The command tracker detects timeouts outside the regular pull/push
+        // mechanism of the input/output ports. Basically the timeout
+        // detection jumps the line when emitting outputs on `resultOut`. If it
+        // then processes a regular completion, it tries to push to `resultOut`
+        // even though it hasn't been pulled again in the meantime. Using `emit`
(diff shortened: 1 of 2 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -167,7 +167,7 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
       )
 
       private def pushResultOrPullCommandResultIn(compl: Option[Ctx[Context, Completion]]): Unit = {
-        compl.fold(pull(commandResultIn))(push(resultOut, _))
+        compl.fold(if (!hasBeenPulled(commandResultIn)) pull(commandResultIn))(emit(resultOut, _))
```
mziolekda, line 170: This line deserves a comment. The problem was hard to find and can be easily overlooked in future refactorings

gerolf-da: I tried to give the reasoning in the commit message, but I'll also add it as a comment.

gerolf-da: Done.

mziolekda (approved):

PR #9067, opened 2021-03-10
Enhanced docker-compose Graphite/Grafana

matthiasS-da:
Enhanced docker-compose project for deploying Graphite/Grafana:
    - Fixed incorrect aggregation of ".count" metrics by DropWizard.
    - Changed Graphite retention schedule to 10s granularity during 7 days, 1m granularity during 30 days, 10m granularity during 1800 days.
    - Preserve Grafana configuration during restarts.
    - Added script to manually purge all data/configuration.
    - Mount Graphite configuration directories on host.
    - Enhanced documentation.
    - Tidied up directory structure.

### Pull Request Checklist

trigger the build.


ledger/metrics/collection/graphite/default_conf/storage-aggregation.conf
```diff
@@ -0,0 +1,32 @@
...
+aggregationMethod = max
+
+[sum]
+pattern = \.sum$
+xFilesFactor = 0
+aggregationMethod = sum
+
+[default_average]
+pattern = .*
+xFilesFactor = 0
```
matthiasS-da, line 31: Here I replaced a 0.3 by 0, because otherwise, if you report only once per minute, graphite will consider your data as "sparse" and discard it at the next aggregation round.

ledger/metrics/collection/graphite/default_conf/storage-aggregation.conf
```diff
@@ -0,0 +1,32 @@
...
+[max]
+pattern = \.upper(_\d+)?$
+xFilesFactor = 0.1
+aggregationMethod = max
+
+[sum]
+pattern = \.sum$
+xFilesFactor = 0
+aggregationMethod = sum
+
```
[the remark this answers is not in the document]
SamirTalwar-DA: I think that's the expected behavior for counters; they're _supposed_ to be cumulative. If you want averages I think you might want to use a meter instead.

ledger/metrics/collection/reset-all.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+docker-compose down -v 
+
+rm -rf graphite/conf/*
+rm -rf graphite/statsd_conf/*
```
SamirTalwar-DA, line 9: Can we use Docker volumes for these?

matthiasS-da: We could. Currently, I perceive that it would be more complicated to inspect the config, if it is stored in a volume. Which benefit do you expect from using volumes?

ledger/metrics/collection/graphite/default_conf/storage-aggregation.conf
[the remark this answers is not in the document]
matthiasS-da: I have been using a `com.codahale.metrics.Meter`. Internally, it has a variable `count` that is incremented on every call to `Meter.mark()`. The `GraphiteReporter` reports the value of `count` to Graphite.

In my test, I was reporting 6 values per minute of the metric `my-metric.count` to Graphite. Graphite was only storing one value per minute. Due to the rule I have now deleted, it was summing up the 6 values instead of taking the average.

Looking at the source code of `com.codahale.metrics.Counter` it seems to be the same story for counters.

ledger/metrics/collection/reset-all.sh
SamirTalwar-DA: Just makes it easier to tear them down. But I don't mind too much either way.

ledger/metrics/collection/graphite/default_conf/storage-aggregation.conf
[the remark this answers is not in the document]
matthiasS-da: As discussed over Slack, will extend README.md to explain the config changes.

ledger/metrics/collection/reset-all.sh
matthiasS-da: Ok, resolving.

PR #9070, opened 2021-03-10
Ignore fetch and lookup-by-key nodes for state update comparison [KVL-854]

miklos-da:
Summary of changes:
* Introduced flag `ignoreFetchAndLookupByKeyNodes` in `NormalizationSettings`.
* Made `StateUpdateComparison` agnostic to the presence of fetch and/or lookup-by-key nodes in a `TransactionAccepted` update in case the previous flag is enabled.

**Motivation:** When comparing historical ledger exports, earlier versions included fetch and lookup-by-key nodes in their events, hence, the generated updates are different from the ones generated by recent code. This PR will enable integrity checking to pass on exports generated by earlier versions of the Daml SDK.

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/StateUpdateComparison.scala
```diff
@@ -162,6 +177,26 @@ object ReadServiceStateUpdateComparison {
...
+    val filteredNodes = nodes.map {
+      case (nid, node: Node.NodeExercises[Nid, Cid]) =>
+        val filteredNode = node.copy(children = node.children.filter(nodes.contains))
+        (nid, filteredNode)
+      case keep => keep
+    }
+    val filteredRoots = tx.roots.filter(filteredNodes.contains)
+    val version = filteredRoots.iterator.foldLeft(TransactionVersion.minVersion)((acc, nodeId) =>
+      TransactionVersion.Ordering.max(acc, nodes(nodeId).version)
+    )
```
fabiotudone-da, line 196: Isn't there an utility for that? I'd rather avoid replicating this logic here.

miklos-da: @remyhaemmerle-da Do you have a suggestion?

remyhaemmerle-da: This is KV specific.  I will rather let it here.

fabiotudone-da: If there isn't already encapsulated logic to decide a transaction's version based on its nodes, I'd at least extract it as a separate method for improved readability.

fabiotudone-da (approved):

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/StateUpdateComparison.scala
remyhaemmerle-da: Daml-LF code assumes there is **only** the engine and the deserialization that creates transaction object. 
The logic to infer the version is nested deeply inside the engine.

remyhaemmerle-da: The question is: do you update the version in KV when you rid the transaction of its Fetch and Lookup Nodes ?

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/StateUpdateComparison.scala
```diff
@@ -162,6 +177,26 @@ object ReadServiceStateUpdateComparison {
...
+    val nodes = tx.nodes.filter {
+      case (_, _: Node.NodeFetch[Cid] | _: Node.NodeLookupByKey[Cid]) => false
+      case _ => true
+    }
+    val filteredNodes = nodes.map {
+      case (nid, node: Node.NodeExercises[Nid, Cid]) =>
+        val filteredNode = node.copy(children = node.children.filter(nodes.contains))
+        (nid, filteredNode)
+      case keep => keep
+    }
```
remyhaemmerle-da, line 192: Please make at least one of the two patern matching exhaustive, so we get warning when we add new type of node.

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/StateUpdateComparison.scala
miklos-da: No, we don't update it, we just set it to the transaction's version (see [here](https://github.com/digital-asset/daml/blob/ecd3110f488566c3ceef3c19c023e8c2e81c960a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala#L355)).

remyhaemmerle-da (approved):
LGTM.

PR #9071, opened 2021-03-10
Move artifact publishing out of yaml files

cocreature:
artifacts we publish in the yaml config. This is a problem because we
always release from `main` so the yaml files are always
identical. However, we will add new artifacts over time and this
starts falling apart. This PR changes this such that the process
described in the yaml files is very generic and just uploads and
downloads everything in a directory whereas the details are handled in
bash scripts that will come from the respective release branch and are
therefore version-dependent.

As usual for these type of changes, I don’t have a great way to test
this. I did do some due diligence to test that at least the artifacts
are published correctly and I can download them but I can’t test the
actual publishing.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -51,17 +51,6 @@ jobs:
   pool:
     vmImage: "Ubuntu-16.04"
   variables:
-    linux-tarball: $[ dependencies.Linux.outputs['publish.tarball'] ]
-    macos-tarball: $[ dependencies.macOS.outputs['publish.tarball'] ]
-    windows-tarball: $[ dependencies.Windows.outputs['publish.tarball'] ]
-    windows-installer: $[ dependencies.Windows.outputs['publish.installer'] ]
-    protos-zip: $[ dependencies.Linux.outputs['publish.protos-zip'] ]
-    daml-on-sql: $[ dependencies.Linux.outputs['publish.daml-on-sql'] ]
-    json-api: $[ dependencies.Linux.outputs['publish.json-api'] ]
-    script-runner: $[ dependencies.Linux.outputs['publish.script-runner'] ]
-    trigger-runner: $[ dependencies.Linux.outputs['publish.trigger-runner'] ]
-    trigger-service: $[ dependencies.Linux.outputs['publish.trigger-service'] ]
-    oauth2-middleware: $[ dependencies.Linux.outputs['publish.oauth2-middleware'] ]
     release_sha: $[ dependencies.check_for_release.outputs['out.release_sha'] ]
     release_tag: $[ dependencies.check_for_release.outputs['out.release_tag'] ]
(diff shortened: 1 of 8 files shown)


ci/copy-unix-release-artifacts.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
```
garyverhaegen-da, line 4:
```suggestion
set -euo pipefail
```

ci/copy-windows-release-artifacts.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+set -eoux pipefail
```
garyverhaegen-da, line 4:
```suggestion
set -eou pipefail
```

ci/publish-artifactory.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
```
garyverhaegen-da, line 4:
```suggestion
set -euo pipefail
```

azure-pipelines.yml
```diff
@@ -148,7 +97,7 @@ jobs:
         trap cleanup EXIT
         echo "$GPG_KEY" | base64 -d > $KEY_FILE
         gpg --homedir $GPG_DIR --no-tty --quiet --import $KEY_FILE
-        cd $(Build.StagingDirectory)/release
+        cd $(Build.StagingDirectory)/release-artifacts/github
```
garyverhaegen-da, line 100: I'm really not sure what kind of directory merging `DownloadPipelineArtifact@0` does if say both `linux-release` and `macos-release` have a `github` folder.

A possible way to test this would be to run a test commit that:
1. Considers itself a release by tricking the check_release job
2. Disabling the Maven/npm push in build-unix
2. Stops the release process after downloading the artifacts, and instead of all the pushing just publishes its `StagingDirectory/release-artifacts` as a new artifact, so we can inspect and check that all the files are there.

garyverhaegen-da (approved):

azure-pipelines.yml
cocreature: Sorry should have been clearer, I tested this by adding a dummy step to `build.yml` which downloaded all of them and directories got merged correctly.

garyverhaegen-da: :+1:

PR #9072, opened 2021-03-10
LF: move value normalizer to transaction module.

remyhaemmerle-da:
Like that people can use it ouside test.

CHANGELOG_BEGIN
CHANELOG_END

### Pull Request Checklist

trigger the build.


remyhaemmerle-da:
cc @oliverse-da

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
```diff
@@ -0,0 +1,58 @@
...
+
+package com.daml.lf
+package value
+
+import com.daml.lf.transaction.TransactionVersion
+import com.daml.lf.value.Value._
+
+object Util {
+
+  // equivalent to serialization + unserialization.
```
cocreature, line 12: worth mentioning that it will never add types? so if you have something of an older version without types it won’t get normalized.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
```diff
@@ -0,0 +1,58 @@
...
+      version: TransactionVersion,
+  ): Either[String, Value[ContractId]] =
+    try {
+      Right(assertNormalize(value0, version))
+    } catch {
+      case e: IllegalArgumentException => Left(e.getMessage)
+    }
+
+  @throws[IllegalArgumentException]
+  def assertNormalize(value0: Value[ContractId], version: TransactionVersion): Value[ContractId] = {
```
cocreature, line 24: Why not return an `Either` here? Performance? How big is the impact of that in practice?

cocreature (approved):
Thanks!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
```diff
@@ -0,0 +1,58 @@
...
+          if (allowGenMap)
+            throw new IllegalArgumentException(s"GenMap are not allowed in LF $version")
+          ValueGenMap(entries.map { case (k, v) => stripTypes(k) -> stripTypes(v) })
+      }
+
+    if (version >= TransactionVersion.minTypeErasure)
+      stripTypes(value0)
+    else
+      value0
+
```
oliverse-da, line 55: Hi @remyhaemmerle-da , in canton, I currently have normalize functions for nodes and the transaction as well:

https://github.com/DACH-NY/canton/pull/5506/files#diff-a47b1d2bc7af56e249f4d79794fc22953f149beecc6340960a21070bbf670fb2R110

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
remyhaemmerle-da: - old (de)serialization processes accepts values without type information -- We just never used this possibility.
- the new serialization drops type information if present, while the new deserialization check it is not there.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
remyhaemmerle-da: To write it in a functional way I would have to use traverse.
- I found traverse not very readable 
- I have to use scalaZ  
- I never remember which implicits, I have to import to use `traverseU`
- It uses more stack 

This is probably several time faster, but it should not have a big impact in practice.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/Util.scala
```diff
@@ -0,0 +1,76 @@
...
+    val allowGenMap = version >= TransactionVersion.minGenMap
+    val eraseType = version >= TransactionVersion.minTypeErasure
+
+    def handleTypeInfo[X](x: Option[X]) =
+      if (eraseType) {
+        None
+      } else if (x.isEmpty) {
+        throw new IllegalArgumentException(
+          s"Type information is require for transaction version $version"
```
oliverse-da, line 38:
```suggestion
          s"Type information is required for transaction version $version"
```

PR #9076, opened 2021-03-10
Add runner for non-repudiation

### LINKED ISSUE #8633: Non-repudiation: add runner
Right now the non-repudiation middleware is purely a library, add a runner that can parse relevant CLI parameters.

stefanobaghino-da:
Closes #8633

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/OptionParser.scala
```diff
@@ -0,0 +1,100 @@
...
+    .required()
+    .action(setLedgerHost)
+    .text("The host address of the participant. Required.")
+
+  opt[String]("ledger-port")
+    .required()
+    .action(setLedgerPort)
+    .text("The port of the participant. Required.")
+
+  opt[Map[String, String]]("jdbc")
```
cocreature, line 22: At some point we have to unify CLI parameters for this. At the moment, daml on sql, the JSON API, the trigger service and now the non-repudiation middleware all specify the database url slightly differently.

We should also open an issue to make this configurable via an environment variable (not because I think it’s really all that useful for security purposes but it’s often a policy for our customers to not specify any secrets via CLI).

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/OptionParser.scala
```diff
@@ -0,0 +1,100 @@
...
+  opt[String]("api-host")
+    .action(setApiHost)
+    .text(
+      s"The interface over which the non-repudiation API will be exposed. Defaults to ${Configuration.Default.proxyAddress.getAddress}."
+    )
+
+  opt[String]("api-port")
+    .action(setApiPort)
+    .text(
+      s"The port over which the non-repudiation API will be exposed. Defaults to ${Configuration.Default.proxyAddress.getPort}."
```
cocreature: Defaults seem off, this is referring to `proxyAddress` instead of `apiAddress`.

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/Configuration.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  val Default: Configuration =
+    Configuration(
+      participantAddress = new InetSocketAddress(LocalHost, ParticipantDefaultPort),
+      proxyAddress = new InetSocketAddress(LoopbackAddress, ProxyDefaultPort),
+      apiAddress = new InetSocketAddress(LoopbackAddress, ApiDefaultPort),
+      apiShutdownTimeout = 10.seconds,
+      databaseJdbcUrl = "jdbc:postgresql:/",
+      databaseJdbcUsername = "nonrepudiation",
+      databaseJdbcPassword = "nonrepudiation",
```
cocreature, line 27: I’m slightly sceptical that this is a good idea. Can we instead require users to set those or make them optional and don’t pass any password and user when setting up the hikari thingy? I forgot what the json API and the trigger service do but probably worth trying to keep them in sync.

cocreature (approved):
Awesome, thank you! Is it worth adding an integration test that uses the runner somewhere?

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/Configuration.scala
stefanobaghino-da: This is already required from the CLI.

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/OptionParser.scala
stefanobaghino-da: Already fixed, thanks.

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/OptionParser.scala
stefanobaghino-da: I agree, initially I had my own names, then I realized I was completely off compared to the rest of the runtime components.

stefanobaghino-da:
> Awesome, thank you! Is it worth adding an integration test that uses the runner somewhere?

I agree. I mostly ran functional tests and have modified versions of Navigator and the JSON API. Once those changes stabilize and are merged we can have meaningful integration tests.

cocreature:
> have modified versions of Navigator and the JSON API
fwiw I think it’s perfectly fine to ignore Navigator. We only support it as a development tool so supporting non-repudiation is not crucial.

stefanobaghino-da:
> > have modified versions of Navigator and the JSON API
> 
> fwiw I think it’s perfectly fine to ignore Navigator. We only support it as a development tool so supporting non-repudiation is not crucial.

Makes sense, thanks.

runtime-components/non-repudiation-app/src/main/scala/com/daml/nonrepudiation/app/OptionParser.scala
stefanobaghino-da: Tracked by https://github.com/digital-asset/daml/issues/9095

PR #9080, opened 2021-03-11
Avoid submitTree on simple exercise commands

### LINKED ISSUE #8774: Avoid submitTree where it is not required

aherrmann-da:
Closes #8774
Continuation of #8952 which addressed a sequence of creates.

This avoids `submitTree` in favor of `submit` for a sequence of commands which are either creates or simple exercises which create exactly one contract and return its id as a result. In the code these are referred to as `SimpleEvent`s.
Adds unit-tests for the identification of simple events and their encoding.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -321,10 +321,7 @@ private[dump] object Encode {
       tree: TransactionTree,
   ): Doc = {
     val rootEvs = tree.rootEventIds.map(tree.eventsById(_).kind)
-    val simpleEventsOnly: Option[List[SimpleEvent]] = {
-      import scalaz.Scalaz._
-      rootEvs.toList.traverse(ev => SimpleEvent.fromTreeEvent(ev, tree))
-    }
+    val simpleEventsOnly: Option[Seq[SimpleEvent]] = SimpleEvent.fromTree(tree)
     simpleEventsOnly match {
       case Some(evs) =>
         encodeSubmitSimpleEvents(partyMap, cidMap, cidRefs, evs)
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -251,29 +251,22 @@ private[dump] object Encode {
       cidRefs: Set[ContractId],
       evs: Seq[CreatedEvent],
   ): Doc = {
-    encodeSubmitEvents(partyMap, cidMap, cidRefs, evs.map(Kind.Created(_)))
+    encodeSubmitSimpleEvents(
+      partyMap,
+      cidMap,
+      cidRefs,
+      evs.map(ev => SimpleEvent(Kind.Created(ev), ContractId(ev.contractId))),
+    )
   }
 
-  private[dump] def encodeSubmitEvents(
+  private[dump] def encodeSubmitSimpleEvents(
       partyMap: Map[Party, String],
       cidMap: Map[ContractId, String],
       cidRefs: Set[ContractId],
-      evs: Seq[TreeEvent.Kind],
+      evs: Seq[SimpleEvent],
   ): Doc = {
-    val submitters = evs.flatMap(ev => evParties(ev)).toSet
-    val cids = ContractId.subst(evs.flatMap {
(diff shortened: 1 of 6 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -151,6 +154,49 @@ object TreeUtils {
   def createdReferencedCids(ev: CreatedEvent): Set[ContractId] =
     ev.createArguments.foldMap(args => args.fields.foldMap(f => valueCids(f.getValue.sum)))
 
+  /** A simple event causes the creation of a single contract and returns its contract id.
+    *
+    * A create event is a simple event. An exercise event can be a simple event.
+    */
+  case class SimpleEvent(event: TreeEvent.Kind, contractId: ContractId)
```
cocreature, line 161: This doesn’t take into account whether the contract ids are used right? So if an exercise creates two contracts and returns one and we use only that, we still will don’t hit this?

That seems very sensible, just making sure I understand the logic correctly.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -151,6 +154,49 @@ object TreeUtils {
...
+            }
+          }
+          val creates = {
+            var creates = Seq.empty[String]
+            traverseEventInTree(exercised, tree) {
+              case (_, Kind.Created(value)) => creates ++= Seq(value.contractId)
+              case _ =>
+            }
+            creates
+          }
```
cocreature, line 182: maybe worth factoring out into a function?

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -151,6 +154,49 @@ object TreeUtils {
...
+          (result, creates) match {
+            case (Some(cid), Seq(createdCid)) if cid == createdCid =>
+              Some(SimpleEvent(exercised, ContractId(cid)))
+            case _ => None
+          }
+        case _ => None
```
cocreature, line 188: maybe match explicitly instead of a wildcard given that only `Empty` is left?

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -151,6 +154,49 @@ object TreeUtils {
...
+          Some(SimpleEvent(created, ContractId(value.contractId)))
+        case exercised @ Kind.Exercised(value) =>
+          val result = value.exerciseResult.flatMap {
+            _.sum match {
+              case Sum.ContractId(value) => Some(value)
+              case _ => None
+            }
+          }
+          val creates = {
+            var creates = Seq.empty[String]
```
cocreature, line 176: Maybe use a listbuffer to make this more efficient or perhaps a set (probably not very useful given that they should be unique anywhere).

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -251,8 +251,22 @@ private[dump] object Encode {
...
+    )
+  }
+
+  private[dump] def encodeSubmitSimpleEvents(
+      partyMap: Map[Party, String],
+      cidMap: Map[ContractId, String],
+      cidRefs: Set[ContractId],
+      evs: Seq[SimpleEvent],
+  ): Doc = {
+    val submitters = evs.flatMap(ev => evParties(ev.event)).toSet
```
cocreature, line 268: can we `foldMap`?

cocreature (approved):
Thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Yes, that's right. It's a relatively simple heuristic.

PR #9086, opened 2021-03-11
Release EE SDK tarballs and installer

cocreature:
As before, no way of testing this. I’ll do a snapshot afterwards.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/publish-artifactory.sh
+++ b/ci/publish-artifactory.sh
@@ -8,7 +8,7 @@ RELEASE_TAG=$2
 
 # We’ve already published to GH so focus on artifactory
 
-INPUTS=$STAGING_DIR/release-artifacts/other
+INPUTS=$STAGING_DIR/release-artifacts/artifactory
 
 push() {
     local file repository local_path md5 sha1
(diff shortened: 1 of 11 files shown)


ci/publish-artifactory.sh
```diff
@@ -33,3 +33,12 @@ push daml-trigger-runner $TRIGGER_RUNNER
 push daml-trigger-runner $TRIGGER_RUNNER.asc
 push daml-script-runner $SCRIPT_RUNNER
 push daml-script-runner $SCRIPT_RUNNER.asc
+
+for platform in linux macos windows; do
+    EE_TARBALL=daml-sdk-$RELEASE_TAG-$platform.tar.gz
+    push sdk-ee $EE_TARBALL
+    push sdk-ee $EE_TARBALL.asc
+done
+EE_INSTALLER=daml-sdk-$RELEASE_TAG-windows.exe
```
garyverhaegen-da, line 42: Should we keep the `ee` in the name here? I think that'd be better to avoid confusion.

cocreature: My idea was that it might make the assistant a bit easier if we keep the files the same and only change the location but I don’t feel too strongly about this and for the installer it doesn’t matter anyway since that’s not consumed by the assistant.

garyverhaegen-da (approved):

ci/publish-artifactory.sh
garyverhaegen-da: I'm thinking we will inevitably have people download both the EE and the CE at some point, and it would be nice for them if their local file names reflected the difference. We should probably not change the CE name because that's arguably a breaking change, but we can choose the name of the ee to be different. I think convenience for our customers outweighs convenience for ourselves, provided the added difficulty in the assistant code is not too high.

cocreature: makes sense, I’ll change it

cocreature: 52f173137fe5d61e2b4946ea3b3eaccddecb7a98 ptal

garyverhaegen-da: lgtm

PR #9096, opened 2021-03-11
fix poweroff on failure to clean

garyverhaegen-da:
This requires a redpoyment of the CI fleet, so I'll be doing that later in the day. I'll merge once the new machines are deployed and I've seen the shutdown work (likely by adding an `exit 1` to `ci/clean-up.yml`).

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -48,7 +48,7 @@ jobs:
         set -euo pipefail
         git checkout $(release_sha)
       name: checkout_release
-      condition: eq(variables.is_release, 'true')
+      condition: and(succeeded(), eq(variables.is_release, 'true'))
     - template: clean-up.yml
     - template: build-unix.yml
       parameters:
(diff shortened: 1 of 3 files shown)


cocreature (approved):
Thanks!

garyverhaegen-da:
After much experimentation, I've decided to roll back this change. I've updated this PR to reflect that (i.e. no more change to infra).

While it is possible to kill the machine from the job:
- Killing in a synchronous way (i.e. plain `sudo poweroff` in the middle of the script), this leaves the current build hanging, waiting for news from the agent. The job is not marked as failed but as "running" until Azure decides it timed out (which can take quite a long time). I think a quick failure is better.
- I was not able to get the shutdown to happen asynchronously. I tried various combinations of subshells, `&` and `nohup`, but I just could not get anything (not even a simple `echo >`) to run asynchronously. I guess the Azure agent cleans up diligently after itself.
Overall it seems to me that the current state, broken as it is, is better than what we'd get with this PR's original intention.

PR #9105, opened 2021-03-11
Apply new logo to create-daml-app template

stefanobaghino-da:
changelog_begin
[SDK Assistant] New Daml logo on create-daml-app template
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da:
### Screenshots

#### Login screen

##### Before

Screenshot from 2021-03-11 17-44-41

##### After

Screenshot from 2021-03-11 19-33-57

#### Main screen

##### Before

Screenshot from 2021-03-11 17-47-35

##### After

Screenshot from 2021-03-11 19-34-14

stefanobaghino-da:
I'm not 100% convinced by how the new logo aligns. Should I also adjust the font for "Create" and "App"?

cocreature:
This is identical to `daml-logo-mark-light.svg` afaict, can we just deduplicate the files so they don’t get out of sync in the future?

cocreature:
> I'm not 100% convinced by how the new logo aligns. Should I also adjust the font for "Create" and "App"?

I agree it looks off. Not sure if it needs adjusting fonts but at least alignment.

stefanobaghino-da:
> This is identical to `daml-logo-mark-light.svg` afaict, can we just deduplicate the files so they don’t get out of sync in the future?

Sure.

Probably it would make sense to create a Bazel target for graphical assets so that they be shared between the documentation and UI stuff?

cocreature:
> Probably it would make sense to create a Bazel target for graphical assets so that they be shared between the documentation and UI stuff?

Either that or just use the file as is for the docs and copy it in the genrule for the template.

stefanobaghino-da:
I also noticed Daml was capitalized incorrectly in the title and changed it.

The new title alignment looks like this (I think the kerning is off between the logo and the rest of the header but it's probably fine as is for the time being):

Screenshot from 2021-03-12 10-58-58

stefanobaghino-da:
> Either that or just use the file as is for the docs and copy it in the genrule for the template.

I would prefer not doing this in a `genrule`, their errors can be a bit difficult to interpret and simply renaming that file would lead to a bit of a head-scratcher. I don't think duplicating is that bad. I don't expect our branding to change every month and nothing really happens if a couple of client-facing components lag behind a bit.

If we really want to have a common pool of graphical assets, I would prefer doing it with a shared target, but I'm not sure there's much to gain from this right now.

cocreature (approved):
:shipit:

PR #9115, opened 2021-03-12
Fix artifactory upload path

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/publish-artifactory.sh
+++ b/ci/publish-artifactory.sh
@@ -23,7 +23,7 @@ push() {
          -H "X-Checksum-SHA1:${sha1}" \
          -X PUT \
          -T ${local_path} \
-         https://digitalasset.jfrog.io/artifactory/${repository}/$(release_tag)/${file}
+         https://digitalasset.jfrog.io/artifactory/${repository}/$RELEASE_TAG/${file}
 }
 
 TRIGGER_RUNNER=daml-trigger-runner-$RELEASE_TAG.jar


ci/publish-artifactory.sh
```diff
@@ -23,7 +23,7 @@ push() {
          -H "X-Checksum-SHA1:${sha1}" \
          -X PUT \
          -T ${local_path} \
-         https://digitalasset.jfrog.io/artifactory/${repository}/$(release_tag)/${file}
+         https://digitalasset.jfrog.io/artifactory/${repository}/$RELEASE_TAG/${file}
```
stefanobaghino-da, line 26: Would it make sense for this to be quoted? I only bring this up because it's the only thing I learned from Shellcheck. 😄

stefanobaghino-da (approved):

ci/publish-artifactory.sh
garyverhaegen-da: I would say "no" because we've built all of those to not have spaces in them, and I'd rather have the script fail on some syntax error than push files with spaces in their name should we ever screw that up. Overall I must say I've been unimpressed with most shellcheck suggestions when I've looked at it.

But I know @SamirTalwar-DA for example disagrees and prefers quoting things in general, so YMMV. 🙂

cocreature: I think I agree with @garyverhaegen-da here. We assume that there are no spaces here in all kinds of places. Adding quotes in a few select places isn’t super helpful.

PR #9125, opened 2021-03-12
Fix javadocs & sources in daml-sdk-head

cocreature:
The order of the list elements was swapped. I don’t trust myself to
not get this wrong in the future so rather than relying on list order,
I added a proper datatype.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/release/src/Maven.hs
+++ b/release/src/Maven.hs
@@ -37,17 +37,14 @@ generateAggregatePom releaseDir artifacts = do
     where
     execution :: (MonadFail m, E.MonadThrow m) => Artifact PomData -> m Text
     execution artifact = do
-        (mainArtifactFile:pomFile:rest) <- map snd <$> artifactFiles artifact
-        let (javadocFile, sourcesFile) = case rest of
-              [javadoc, sources] -> (Just javadoc, Just sources)
-              _ -> (Nothing, Nothing)
+        ArtifactFiles{..} <- fmap snd <$> artifactFiles artifact
         let configuration =
                 map (\(name, value) -> (name, pathToText (releaseDir </> value))) $
                     Maybe.catMaybes
-                        [ Just ("pomFile", pomFile)
-                        , Just ("file", mainArtifactFile)
-                        , ("javadoc", ) <$> javadocFile
-                        , ("sources", ) <$> sourcesFile
+                        [ Just ("pomFile", artifactPom)
+                        , Just ("file", artifactMain)
+                        , ("javadoc", ) <$> artifactJavadoc
+                        , ("sources", ) <$> artifactSources
                         ]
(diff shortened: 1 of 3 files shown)


hubert-da (approved):
I'm not familiar with Haskell, so someone else might need to take a look. LGTM, anyway

release/src/Maven.hs
```diff
@@ -37,17 +37,14 @@ generateAggregatePom releaseDir artifacts = do
     where
     execution :: (MonadFail m, E.MonadThrow m) => Artifact PomData -> m Text
     execution artifact = do
-        (mainArtifactFile:pomFile:rest) <- map snd <$> artifactFiles artifact
-        let (javadocFile, sourcesFile) = case rest of
-              [javadoc, sources] -> (Just javadoc, Just sources)
-              _ -> (Nothing, Nothing)
+        ArtifactFiles{..} <- fmap snd <$> artifactFiles artifact
```
SamirTalwar-DA, line 40: Didn't know you could do this. Fancy.

release/src/Util.hs
```diff
@@ -265,8 +266,15 @@ javadocJarName Artifact{..}
 customJavadocJarName :: Artifact a -> Maybe Text
 customJavadocJarName Artifact{..} = T.pack . toFilePath <$> artJavadocJar
 
+data ArtifactFiles f = ArtifactFiles
+  { artifactMain :: f
+  , artifactPom :: f
+  , artifactSources :: Maybe f
+  , artifactJavadoc :: Maybe f
+  } deriving (Functor, Foldable)
```
SamirTalwar-DA, line 274: I like the kind of fix that makes it much harder to break in the future.

SamirTalwar-DA (commented):
Looks good. Can we add a test case somewhere to catch this kind of thing?

cocreature:
> Looks good. Can we add a test case somewhere to catch this kind of thing?

In theory yes but an integration test is slower (several minutes) than I think we can justify for something that is purely for development and changes very rarely. Do you have a good idea of what exactly we could test here which isn’t just redundant now that we don’t rely on list ordering anymore?

garyverhaegen-da (approved):
That was very un-Haskell-like, using a random List with an implicit expected length to convey structured data. Almost like what I'd expect in Clojure code or something.

garyverhaegen-da:
> Looks good. Can we add a test case somewhere to catch this kind of thing?

FWIW I for one am convinced by the type and don't feel the need for a test on this.

SamirTalwar-DA:
Eh, if it's tricky, just ship it. :shipit:

SamirTalwar-DA (approved):

PR #9136, opened 2021-03-14
release job: remove extra bash-lib

garyverhaegen-da:
The filesystem is not reset between steps in the same job, so there is
no need to add the `bash-lib` multiple times.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
:fire:

PR #9137, opened 2021-03-14
Fix cache clean up

garyverhaegen-da:
This is a continuation of #8595 and #8599. I somehow had missed that `/etc/fstab` can be used to tell `mount` to let users mount some filesystems with preset options.

This is using the full history of `mount` hardening so should be safe enough. The option `user` in `/etc/fstab` automatically disables any kind of `setuid` feature on the mounted filesystem, which is the main attack vector I know of.

This works flawlessly on my local VM, so hopefully this time's the charm. (It also happens to be my third PR specifically targeted on this issue, so, who knows, it may even work.)

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/clean-up.yml
+++ b/ci/clean-up.yml
@@ -5,25 +5,31 @@ steps:
 - bash: |
     set -euo pipefail
 
+    exec 1> >(while IFS= read -r line; do echo "$(date -Is) [out]: $line"; done)
+    exec 2> >(while IFS= read -r line; do echo "$(date -Is) [err]: $line"; done >&2)
+
     # Location of the disk cache for CI servers set in their init files:
     # infra/macos/2-common-box/init.sh:echo "build:darwin --disk_cache=~/.bazel-cache" > ~/.bazelrc
     # infra/vsts_agent_linux_startup.sh:echo "build:linux --disk_cache=~/.bazel-cache" > ~/.bazelrc
 
+    df -h .
     if [ $(df -m . | sed 1d | awk '{print $4}') -lt 50000 ]; then
         echo "Disk full, cleaning up..."
-        disk_cache="$HOME/.bazel-cache"
-        rm -rf "$disk_cache"
-        echo "removed '$disk_cache'"
-        local_cache="$HOME/.cache/bazel"
-        if [ -d "$local_cache" ]; then
-            for pid in $(pgrep -a -f bazel | awk '{print $1}'); do
-                kill -s KILL $pid
-            done
-            chmod -R +w "$local_cache"
-            rm -rf "$local_cache"
-            echo "removed '$local_cache'"
-        fi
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Thanks!

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -95,6 +95,44 @@ useradd \
...
+set -euo pipefail
+
+reset_cache() {
+    local file mount_point
+    file=$1
+    mount_point=$2
+
+    echo "Cleaning up '$mount_point'..."
+    if [ -d "$mount_point" ]; then
+        for pid in $(pgrep -a -f bazel | awk '{print $1}'); do
```
SamirTalwar-DA, line 116: What happens if there are no PIDs?

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -95,6 +95,44 @@ useradd \
   vsts
 #add docker group to user
 usermod -aG docker vsts
+# let vsts user mount/unmount cache folders
+echo "/tmp/bazel_cache /home/vsts/.cache/bazel auto rw,user,exec" >> /etc/fstab
+echo "/tmp/disk_cache /home/vsts/.bazel-cache auto rw,user,exec" >> /etc/fstab
```
SamirTalwar-DA, line 100: I didn't know you could just mount directories to other places like this, but I will take your word for it that it works.

ci/clean-up.yml
```diff
@@ -5,25 +5,31 @@ steps:
 - bash: |
     set -euo pipefail
 
+    exec 1> >(while IFS= read -r line; do echo "$(date -Is) [out]: $line"; done)
+    exec 2> >(while IFS= read -r line; do echo "$(date -Is) [err]: $line"; done >&2)
```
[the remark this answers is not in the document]
garyverhaegen-da: Fair warning: I discovered that recently and I'm going to use it all the time from now on.

infra/vsts_agent_ubuntu_20_04_startup.sh
garyverhaegen-da: `for` doesn't run if the list to run over is empty, like in most languages.

infra/vsts_agent_ubuntu_20_04_startup.sh
garyverhaegen-da: Mounting directories is possible but a bit more complicated; this is mounting files. Specifically, files that have a file system on them.

garyverhaegen-da:
Applied 6b0a8b168761359673ba02d4d52b439f8ed286c2.

PR #9138, opened 2021-03-14
Race condition ITs [DPP-274]

kamil-da:
**Motivation**
Multiple possible race conditions have been identified:

Write-write conflicts:
- archive-contract vs. archive-contract
- non-transient create-with-key vs. non-transient create-with-key
- archive-contract vs. non-transient create-with-key
- _command-dedup vs. command-dedup_ (not yet covered by this PR)

Read-write conflicts:
- archive-contract vs. non-consuming-choice
- archive-contract vs fetch-contract
- archive-contract vs. successful lookup-by-key
- transient create-with-key vs. non-transient-create-with-key
- failing lookup-by-key vs. non-transient-create-with-key

This PR introduces a test suite for the `ledger-api-test-tool` for catching those race conditions.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -78,7 +78,8 @@ conformance_test(
         "--exclude=ContractKeysIT,ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped" +
         ",ParticipantPruningIT" +  # see "conformance-test-participant-pruning" below
         ",ConfigManagementServiceIT,LedgerConfigurationServiceIT" +  # dynamic config management not supported by Canton
-        ",ClosedWorldIT",  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",ClosedWorldIT" + # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",RaceConditionIT", # Not yet adapted for Canton
     ],
 ) if not is_windows else None
 
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -78,7 +78,8 @@ conformance_test(
         "--exclude=ContractKeysIT,ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped" +
         ",ParticipantPruningIT" +  # see "conformance-test-participant-pruning" below
         ",ConfigManagementServiceIT,LedgerConfigurationServiceIT" +  # dynamic config management not supported by Canton
-        ",ClosedWorldIT",  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",ClosedWorldIT" +  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",RaceConditionIT",
     ],
 ) if not is_windows else None
 
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -78,8 +78,8 @@ conformance_test(
         "--exclude=ContractKeysIT,ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped" +
         ",ParticipantPruningIT" +  # see "conformance-test-participant-pruning" below
         ",ConfigManagementServiceIT,LedgerConfigurationServiceIT" +  # dynamic config management not supported by Canton
-        ",ClosedWorldIT" + # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
-        ",RaceConditionIT", # Not yet adapted for Canton
+        ",ClosedWorldIT" +  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",RaceConditionIT",  # Not yet adapted for Canton
     ],
 ) if not is_windows else None
 
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -78,8 +78,7 @@ conformance_test(
         "--exclude=ContractKeysIT,ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped" +
         ",ParticipantPruningIT" +  # see "conformance-test-participant-pruning" below
         ",ConfigManagementServiceIT,LedgerConfigurationServiceIT" +  # dynamic config management not supported by Canton
-        ",ClosedWorldIT" +  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
-        ",RaceConditionIT",  # Not yet adapted for Canton
+        ",ClosedWorldIT",  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
     ],
 ) if not is_windows else None
 
@@ -111,7 +110,8 @@ conformance_test(
     test_tool_args = [
         "--verbose",
         "--concurrent-test-runs=4",  # lowered from default #procs to reduce flakes - details in https://github.com/digital-asset/daml/issues/7316
-        "--include=ContractKeysIT",
+        "--include=ContractKeysIT" +
+        ",RaceConditionIT",
     ],
 ) if not is_windows else None
 
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
@@ -65,10 +65,7 @@ final class LedgerTestCasesRunner(
     }
 
     (2 to test.repeated).foldLeft(logAndStart(1)) { (result, repetition) =>
-      for {
-        durationSoFar <- result
-        duration <- logAndStart(repetition)
-      } yield durationSoFar + duration
+      result.flatMap(_ => logAndStart(repetition))
     }
   }
 
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
@@ -59,12 +59,26 @@ final class LedgerTestCasesRunner(
   private def start(test: LedgerTestCase, session: LedgerSession)(implicit
       executionContext: ExecutionContext
   ): Future[Duration] = {
+    def logAndStart(repetition: Int): Future[Duration] = {
+      logger.info(s"Starting '${test.description}'. Run: $repetition out of ${test.repeated}")
+      startSingle(test, session, repetition)
+    }
+
+    // TODO: make it return all/mean of all durations, not only the last duration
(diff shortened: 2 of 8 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,357 @@
...
+        ExpectedNumberOfSuccessfulArchivals,
+        transactions.filter(isArchival),
+      )
+      ()
+    }
+  })
+
+  test(
+    "WWArchiveVsNonTransientCreate",
+    "Cannot re-create a contract if it hasn't yet been archived",
```
tudor-da, line 66: I believe here you're referring to **_re-assign_** a contract key. Can you amend the description?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,357 @@
...
+
+      transactions
+        .filter(isTransientCreate)
+        .foreach(assertTransactionOrder(_, nonTransientCreateTransaction))
+    }
+  })
+
+  test(
+    "RWArchiveVsNonConsumingChoice",
+    "Cannot exercise a non-consuming choice after a contract archival",
```
tudor-da, line 145: More generally, one cannot exercise any choice after a contract archival.

tudor-da (changes requested):
Very nice one! I left a couple of comments about ergonomics and descriptions

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,357 @@
...
+    for {
+      contract <- ledger.create(alice, ContractWithKey(alice))
+      _ <- Future.traverse(1 to Attempts) { attempt =>
+        if (attempt % 2 == 1) {
+          ledger.create(alice, ContractWithKey(alice)).transform(Success(_))
+        } else {
+          ledger.exercise(alice, contract.exerciseContractWithKey_Archive).transform(Success(_))
+        }
+      }
+      transactions <- ledger.transactionTrees(alice)
```
tudor-da, line 79: Will there ever be more than 3 elements in this stream?
I believe the last 98 submissions cannot have any effect since:
- The first submission creates a contract **_contract_** with key for alice
- The second submission archives it (index 1)
- The third submission creates a **_different_** contract for the same key (index 2)
From here on, the test will not progress with contract creations since the next archive (index 3) will try to archive the original **contract** instead of the last created one. Is this the intent?

kamil-da: This test is a bit different in its mechanics than the rest because here we execute `archive` 50 times and `create` 50 times.
The race condition we try to examine here is (following @mziolekda explanation):
`[create]  <wait>  [archive]-race-[create]`
In case of a bug causing the second [create] to see a partial result of [archive] command we could end up with two consecutive successful contract creations.
I've added an appropriate comment in the code.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
kamil-da: I've updated the title.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
kamil-da: Done.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,357 @@
...
+    } yield {
+      import TransactionUtil._
+      def txLt(tx1: TransactionTree, tx2: TransactionTree): Boolean =
+        offsetLessThan(tx1.offset, tx2.offset)
+
+      val sorted = transactions.sortWith(txLt)
```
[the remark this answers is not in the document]
kamil-da: You're right - there's no need for sorting the stream 💯

kamil-da:
@tudor-da Thank you again for the review. I've addressed your comments.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,360 @@
...
+import scala.util.Success
+
+final class RaceConditionIT extends LedgerTestSuite {
+
+  test(
+    "WWDoubleNonTransientCreate",
+    "Cannot concurrently create multiple non-transient contracts with the same key",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, alice)) =>
+    val Attempts = 100
```
mziolekda, line 24: I would suggest rephrasing this test. A 100 attempts to archive the same contract is unlikely to expose more problems than doing it 5 times. Creating 5 contracts and archiving them multiple times, otoh might.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,360 @@
...
+        ()
+      }
+  })
+
+  test(
+    "WWDoubleArchive",
+    "Cannot archive the same contract multiple times",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, alice)) =>
+    val Attempts = 100
```
mziolekda, line 45: As above, see also `SemanticConcurrentDoubleSpend` which is a very similar test

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,360 @@
...
+        fail(s"""Invalid transaction sequence: ${transactions.map(printTransaction).mkString("\n")}""")
+    }
+  })
+
+  test(
+    "RWTransientCreateVsNonTransientCreate",
+    "Cannot create a transient contract and a non-transient contract with the same key",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, alice)) =>
+    val Attempts = 100
```
mziolekda, line 118: Same as above, 5x5 would yield more meaningful results

mziolekda (changes requested):
Good job with these race condition tests. I would consider rephrasing them into a format where the races are shorter but repeated several times. Making races 100 transactions long will not be as effective at discovering problems as 3-5 races of 5 transactions each.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -78,7 +78,8 @@ conformance_test(
         "--exclude=ContractKeysIT,ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped" +
         ",ParticipantPruningIT" +  # see "conformance-test-participant-pruning" below
         ",ConfigManagementServiceIT,LedgerConfigurationServiceIT" +  # dynamic config management not supported by Canton
-        ",ClosedWorldIT",  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",ClosedWorldIT" +  # Canton currently fails this test with a different error (missing namespace in "unallocated" party id)
+        ",RaceConditionIT",  # Not yet adapted for Canton
```
oliverse-da, line 82: As discussed over slack, please add `RaceConditionIT` to the `conformance-test-contract-keys` to line 114 below as the suite relies on unique contract keys. - fyi @andreaslochbihler-da

andreaslochbihler-da: Great to see such tests coming! Is there any chance that we split the test into two pieces (one with all the tests that require unique contract keys and one for the rest)? Then we could run Canton on the latter also in the non-unique mode.

oliverse-da: > Great to see such tests coming! Is there any chance that we split the test into two pieces (one with all the tests that require unique contract keys and one for the rest)? Then we could run Canton on the latter also in the non-unique mode.

Good point, I was assuming all `RaceConditionIT` tests were relying on unique contract keys as the condition to race against, but that may not be the case.

kamil-da: @andreaslochbihler-da I think it's a good idea!
But need @mziolekda input to decide if I should do the split in this or another PR. Personally I would opt for making it a new PR.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
```diff
@@ -58,13 +58,27 @@ final class LedgerTestCasesRunner(
 
   private def start(test: LedgerTestCase, session: LedgerSession)(implicit
       executionContext: ExecutionContext
+  ): Future[Duration] = {
+    def logAndStart(repetition: Int): Future[Duration] = {
+      logger.info(s"Starting '${test.description}'. Run: $repetition out of ${test.repeated}")
+      startSingle(test, session, repetition)
+    }
+
+    // TODO: make it return all/mean of all durations, not only the last duration
```
tudor-da, line 67: Is this comment for later? Otherwise, can you remove it if it's already adressed?

kamil-da: Actually this is something I couldn't decide if it's worth doing or not.
I've just change this part to return the sum of all durations for repeating test cases. I'm open to alternatives though.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RaceConditionIT.scala
```diff
@@ -0,0 +1,393 @@
...
+      assert(
+        isCreateNonTransient(transactions.head),
+        "The first transaction is expected to be a contract creation",
+      )
+      assert(
+        transactions.exists(isCreateDummyContract),
+        "A dummy contract creation is missing. A possible reason might be reading the transactions stream in the test case before submitted commands had chance to be processed.",
+      )
+
+      val (_, valid) =
```
tudor-da, line 106: Can't we make a more strict assertion here? Don't we know how many transactions should there be and their order?

kamil-da: Actually after some discussions we came with a conclusion that we only care about the order here and not how many transactions (because of contention that might happen here). That is why this simple check is sufficient for our needs.

tudor-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
```diff
@@ -58,13 +58,26 @@ final class LedgerTestCasesRunner(
 
   private def start(test: LedgerTestCase, session: LedgerSession)(implicit
       executionContext: ExecutionContext
+  ): Future[Duration] = {
+    def logAndStart(repetition: Int): Future[Duration] = {
+      logger.info(s"Starting '${test.description}'. Run: $repetition out of ${test.repeated}")
```
mziolekda, line 63: Print this message only if test.repeated>1

mziolekda (approved):
Looks good, smash it in

PR #9142, opened 2021-03-15
env vars

garyverhaegen-da:
```
$ DAML_SDK_ADDRESS=0.0.0.0 daml-head json-api --ledger-host [...]
[...] - Config([...], address=0.0.0.0, [...])
```

My goal is to find a way to let users set options with environment
variables instead of CLI arguments. This is definitely not pretty, and
only touches the JSON API for now, but it seems to work.

I'm opening a PR to gather feedback on:

- Is this something we want? Is it worth spending more time on it?
- Is the approach completely wrong, and if so, what would you suggest
  instead?

One nice property of the code as it stands is that it uses the CLI arg
definition as the single source of truth, so there is no risk of CLI
args and env vars getting out of sync, and documentation could explain
the correspondance in a couple sentences. This could be reused across all
of our projects that use scopt.

Not so nice things:

- `scopt` is clearly not designed for this; this code, while simple and
  reasonably readable in my opinion, is definitely very hackish and may
  break when changing `scopt` version.
- At the moment required arguments have to be passed through the CLI;
  changing that is possible but a it more invasive. Given that we don't
  have that many required arguments, I think there is a case for not
  addressing that limitation if we want to keeo the code simple.

CHANGELOG_BEGIN
not yet
CHANGELOG_END

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Cli.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Cli.scala
@@ -16,8 +16,27 @@ object Cli extends StrictLogging {
   private[http] def parseConfig(
       args: collection.Seq[String],
       getEnvVar: String => Option[String] = sys.env.get,
-  ): Option[Config] =
-    configParser(getEnvVar).parse(args, Config.Empty)
+  ): Option[Config] = {
+    configParser(getEnvVar).parse(args, configFromEnv(getEnvVar))
+  }
+
+  private def configFromEnv(
+      env: String => Option[String]
+  ): Config = {
+    // Make all options optional, else we can't construct a partial config
+    val ods = configParser(env).optionDefs.map(_.optional)
+    val cliFromEnvVars = ods
+      .map(_.fullName)
+      .filter(_.length > 0)
+      .filter(_ != "--help")
+      .map(cli => (cli, "DAML_SDK_" + cli.toUpperCase.replaceAll("-", "_").substring(2)))
+      .map({ case (cli, n) => (cli, n, env(n))})
+      .filter({ case (_, _, o) => o.isDefined})
+      .flatMap({ case (cli, _, o) => Seq(cli, o.get)})
+    val (config, _) = scopt.OParser.runParser(scopt.OParser(ods.head, ods.tail.toList), cliFromEnvVars, Config.Empty)
+
+    config.getOrElse(Config.Empty)
+  }
 
   private def setJdbcConfig(
       config: Config,
@@ -48,9 +67,13 @@ object Cli extends StrictLogging {
       conf <- parseJdbcConfig(v)
     } yield conf
 
+  trait ExposeData[C] {
+    def optionDefs: Seq[scopt.OptionDef[_, C]]
+  }
+
   @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
-  private def configParser(getEnvVar: String => Option[String]): scopt.OptionParser[Config] =
-    new scopt.OptionParser[Config]("http-json-binary") {
+  private def configParser(getEnvVar: String => Option[String]): scopt.OptionParser[Config] with ExposeData[Config] =
+    new scopt.OptionParser[Config]("http-json-binary") with ExposeData[Config] {
 
       override def renderingMode: RenderingMode = RenderingMode.OneColumn
 
@@ -166,5 +189,7 @@ object Cli extends StrictLogging {
         .optional()
         .valueName(WebsocketConfig.usage)
         .text(s"Optional websocket configuration string. ${WebsocketConfig.help}")
+
+      override def optionDefs = options.toSeq
     }
 }


cocreature:
I’m broadly in support of supporting env vars in general. See also @bame-da’s [proposal](https://docs.google.com/document/d/1Lh6GmSHLPkwv-vGpImVtR-fwSxy8ilC4slMBavN07UA/edit#)  (private doc). While I’m in support of supporting them, I think I’m not quite convinced that this is that big of an issue in practice at the moment, e.g., whether you specify `args` for your k8s pod or `env` seems irrelevant. That said, at least for secret management it comes up pretty regularly and while you can debate the practical security implications, things like k8s secrets are definitely geared towards env vars over CLI args.

As for the implementation, I’m against adhoc solutions here. If we want to address this, we should agree on Bernhard’s proposal or make a comparable one and then based on that build a library (which can wrap scopt) which we can swap in everywhere. Hacking this into just the JSON API without a clear agreement that we can follow this pattern everywhere is only going to bite us later.

da-tanabe:
Strong preference to using environment variables; they're easier to compose (building up command-line flags in the right order in a flat list is a pain) and as @cocreature mentioned, the only way to effectively get things like secrets into a container in Kubernetes is through environment variables (otherwise you have to reference those environment variables in a command-line, which is even more annoying).

And separately I also agree this makes sense to tackle across the board. The disparity in command-line arguments across applications is annoying, I hope that we can do better in environment variables.

The only special case environment variable IMHO should be `PORT`; it's commonly used as a service provider's way of telling you which port you should be binding to:
* Google Cloud Run: https://cloud.google.com/run/docs/configuring/environment-variables#reserved
* Heroku: https://devcenter.heroku.com/articles/runtime-principles#web-servers
Via Heroku, it seems that at least in the node.js environment, `PORT` is widely used as convention.

I don't have strong opinions about namespacing the rest of the environment variables any more; I used to think they were a great idea but now I'm not so sure.
* MySQL's environment variables are perhaps a guide showing that not being disciplined about _how_ to namespace could leave a bit of a mess behind: https://dev.mysql.com/doc/refman/8.0/en/environment-variables.html
* Postgres's vars are much tidier: https://www.postgresql.org/docs/13/libpq-envars.html

garyverhaegen-da:
@cocreature I suppose I should have been a bit clearer about my intentions here — I've opened this as a draft for early feedback, but it's never been my intention to merge a change limited to the JSON API.

This draft was meant as a design proposal that is actually quite different from @bame-da's document, in that it is much more limited in scope.

First off, I'm only concerned with "production" applications. I believe CLI args are fine for interactive CLI applications, and I have no desire to touch any of the assistant commands or the `daml.yaml` configuration file.

Second, this is not a difference between passing an option through `env` and passing the same option through `args`; it is the difference between a Docker image that works out of the box and spending half a day wondering why Kubernetes keeps restarting a container.

The concrete proposal is this: let's automatically generate an "environment variables" API based on our existing "CLI arguments" API.

The advantages are:
- Our applications already have CLI arguments, so there's little extra work required. Based on the attached proof of concept, this seems feasible in about twenty lines of shared library code and fairly minimal (~3 lines) one-time change to applications' `main` method.
- We don't need to write extra documentation on a different way to configure our applications; we can just explain how CLI args translate to env var names and point to the existing documentation.
- Our users are unlikely to get confused about which env var corresponds to which CLI arg.
- There is no risk of env vars and CLI args getting out of sync in either naming or capabilities.
- Our developers don't need to change the way they design and write code.

The main drawback, compared to @bame-da's document, is that we do not take the introduction of environment variables as an opportunity to completely redesign the way in which we configure all of our applications. I do not think this adds any additional impediment to that effort, though.

> @da-tanabe The disparity in command-line arguments across applications is annoying, I hope that we can do better in environment variables.

I actually do not think it is a good idea to try to solve that by introducing environment variables, as it will lead to unclear, ad-hoc and possibly incomplete mappings between what can be configured by environment variables vs. CLI args. I believe we need to address that by changing the CLI flags themselves (as always, with an intermediate version that accepts both the new and the old, warning on the old (and crashing if both are specified)).

> I don't have strong opinions about namespacing the rest of the environment variables any more; I used to think they were a great idea but now I'm not so sure.

I had not thought about this much when writing the code, but the more I think about it now, the less appealing prefixes seem. I'd be inclined to drop them entirely at this point. Prefixing makes sense if you're setting up environment variables at the machine level and you expect many different applications to run under the same environment (e.g. `~/.bashrc`), as that allows you to avoid collisions. But here we're targeting specialized containers that for the most part literally do not run any other process. In this case prefixing seems like it would just be noise.

da-tanabe:
>> I don't have strong opinions about namespacing the rest of the environment variables any more; I used to think they were a great idea but now I'm not so sure.

> I had not thought about this much when writing the code, but the more I think about it now, the less appealing prefixes seem. I'd be inclined to drop them entirely at this point. Prefixing makes sense if you're setting up environment variables at the machine level and you expect many different applications to run under the same environment (e.g. ~/.bashrc), as that allows you to avoid collisions. But here we're targeting specialized containers that for the most part literally do not run any other process. In this case prefixing seems like it would just be noise.

Completely agree! I used to do this reflexively for apps that _I_ wrote until I started working more with containers and other people's apps; unless the prefix is _really_ short (like Postgres's `PG`) it's quite visually cluttering. `DAML_` could make a nice prefix for us, but `DAML_COM_DIGITALASSET_LEDGER_PORT` would be…distasteful. :-)

da-tanabe:
>> The disparity in command-line arguments across applications is annoying, I hope that we can do better in environment variables.

> I actually do not think it is a good idea to try to solve that by introducing environment variables, as it will lead to unclear, ad-hoc and possibly incomplete mappings between what can be configured by environment variables vs. CLI args. I believe we need to address that by changing the CLI flags themselves (as always, with an intermediate version that accepts both the new and the old, warning on the old (and crashing if both are specified)).

I wasn't clear in that comment but that's what I meant as well. Specifically, I think if we introduced environment variables based on the flags as they are defined _now_, I think that's _worse_ than just not having them at all. So our environment variables should be consistent across components _and_ be obvious from people that know the CLI too, which implies that we clean up the CLI flags too.

garyverhaegen-da:
> Specifically, I think if we introduced environment variables based on the flags as they are defined _now_, I think that's _worse_ than just not having them at all.

Just to be clear: are you saying you would prefer we wait until the CLI args are cleaned-up _before_ we introduce environment variables?

cocreature:
> Second, this is not a difference between passing an option through `env` and passing the same option through `args`; it is the difference between a Docker image that works out of the box and spending half a day wondering why Kubernetes keeps restarting a container.

This is what I don’t entirely buy. We have a whole bunch of mandatory arguments and most of those can’t go away. To make it work out of the box you need to wrap it in something that can provide sensible defaults like a Helm chart. At that point, it doesn’t matter if the default value of the CLI arg like the address you listen on is set in the helm chart or the container itself.

There is definitely a risk of dying by a thousand paper cuts with those paper cuts being things like specifying the address to listen on but at this point, I don’t think that’s the main thing people are having issues with. Whether I need to specify the address to listen on or not is a marginal improvement and completely disappears if we provide templates (regardless of whether that is helm charts or something else). I’d rather focus on the big stuff first and then try to address small issues like this one afterwards in a consistent manner (imho inconsistent CLI args/env vars are at least as much of an issue as not being able to specify something as an env var when it comes to "death by a thousand paper cuts").

da-tanabe:
>> Specifically, I think if we introduced environment variables based on the flags as they are defined now, I think that's worse than just not having them at all.

> Just to be clear: are you saying you would prefer we wait until the CLI args are cleaned-up before we introduce environment variables?

_Definitely_.

> There is definitely a risk of dying by a thousand paper cuts with those paper cuts being things like specifying the address to listen on but at this point, I don’t think that’s the main thing people are having issues with. Whether I need to specify the address to listen on or not is a marginal improvement and completely disappears if we provide templates (regardless of whether that is helm charts or something else). I’d rather focus on the big stuff first and then try to address small issues like this one afterwards in a consistent manner (imho inconsistent CLI args/env vars are at least as much of an issue as not being able to specify something as an env var when it comes to "death by a thousand paper cuts").

I think it's important to be really clear about what we're really trying to solve for? Is it _possible_ to deploy these images and run them in clusters today? Absolutely it is—we do it Daml Hub today. Is it more work than I want it to be? Also definitely; maintaining our templates for deploying these things is a pain.

So if the goal is "make it deployable", then mission already accomplished; move on to the next thing! But if the goal is to "make it _easily_ deployable", then IMHO some agreement and definition of "easy" is critical to make progress. Selfishly a sensible metric might be "number of bytes required to write a Kubernetes `Deployment`" or something, which takes this out of the realm of the subjective and into a more objective discussion. What _is_ the "big stuff" if we can already deploy these images as-is?

PR #9143, opened 2021-03-15
KVL-203 Share participant state integration test harness

hubert-da:
The motivation of these changes is to eliminate manual work and reduce duplication between the SDK and oem-integration-kit repos by reusing the same test fixture for integration testing participant state implementations. Also, the DARs required for running these tests won't need to be manually updated.

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala
```diff
@@ -45,10 +44,11 @@ import scala.util.{Failure, Success, Try}
 abstract class ParticipantStateIntegrationSpecBase(implementationName: String)(implicit
     testExecutionContext: ExecutionContext = ExecutionContext.global
 ) extends AsyncWordSpec
-    with TestResourceContext
     with BeforeAndAfterEach
     with AkkaBeforeAndAfterAll {
 
+  implicit private val resourceContext: ResourceContext = ResourceContext(testExecutionContext)
```
hubert-da, line 50: This change allowed to reduce the number of published libraries

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala
```diff
@@ -713,11 +713,17 @@ object ParticipantStateIntegrationSpecBase {
...
+  private val archives = {
     val reader = DarReader { (_, stream) => Try(DamlLf.Archive.parseFrom(stream)) }
-    val fileName = new File(rlocation(com.daml.ledger.test.TestDars.fileNames("model")))
-    val Success(testDar) = reader.readArchiveFromFile(fileName)
-    testDar.all
+    val fileName = com.daml.ledger.test.TestDars.fileNames("model")
+    reader
+      .readArchive(
+        fileName,
+        new ZipInputStream(this.getClass.getClassLoader.getResourceAsStream(fileName)),
```
hubert-da, line 722: This change allowed to reduce the number of published libraries

ledger/test-common/BUILD.bazel
```diff
@@ -141,6 +141,7 @@ EOF
             generated_srcs = [":dar-file-names-scala-%s" % target],
             # generated_srcs is required for scaladoc
             resources = ["dar-files-%s" % target],
+            tags = ["maven_coordinates=com.daml:test-dar-files-%s-lib:__VERSION__" % target],
```
remyhaemmerle-da, line 144: If you want to publish those artifacts, you should problably be more 


             for target in lf_versions_aggregate("legacy", "stable", "latest", "preview", "dev", "1.8", "1.11", "1.12")
```

File: ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala (hunk 2 above)
Comment (SamirTalwar-DA) [reply ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala:722]: Much better, thanks!

File: release/src/Main.hs [hunk 4]
```diff
@@ -40,6 +40,13 @@ import qualified SdkVersion
 scalaVersions :: [T.Text]
 scalaVersions = ["2.12.12", "2.13.3"]
 
+-- Build-time dependencies to exclude
+depsToExclude :: T.Text
+depsToExclude = T.intercalate " + " [
+   "//compiler/scenario-service/protos:scenario_service_java_proto",
+   "//compiler/repl-service/protos:repl_service_java_proto",
+   "//daml-script/runner:script-runner-lib-ce"]
```
SamirTalwar-DA, line 48: I'd like @garyverhaegen-da, @moritzkiefer-da, or someone equally familiar with this script to review this part.

ledger-api/testing-utils/src/main/scala/com/digitalasset/ledger/api/testing/utils/AkkaBeforeAndAfterAll.scala
```diff
@@ -23,7 +23,7 @@ trait AkkaBeforeAndAfterAll extends BeforeAndAfterAll {
 
   private implicit lazy val executionContext: ExecutionContext =
     ExecutionContext.fromExecutorService(
-      Executors.newSingleThreadExecutor(
+      Executors.newCachedThreadPool(
```
SamirTalwar-DA, line 26: I'm glad this worked. In a subsequent PR, can you try removing it entirely and let the actor system use the global EC, to see if that also works?

SamirTalwar-DA (approved):
I like it. Please wait for a review from someone else regarding the release script.

SamirTalwar-DA:
I think you're going to need to `git merge origin/main` to get the build to work.

ledger-api/testing-utils/src/main/scala/com/digitalasset/ledger/api/testing/utils/AkkaBeforeAndAfterAll.scala
cocreature: What are we trying to achieve with this change?

release/artifacts.yaml
```diff
@@ -33,15 +33,19 @@
   type: jar-scala
 - target: //daml-lf/data-scalacheck:data-scalacheck
   type: jar-scala
+- target: //daml-lf/encoder:encoder
+  type: jar-scala
```
cocreature, line 37: Why do we need the encoder? This was really only intended for internal testing for things that are hard to produce via the daml compiler.

release/artifacts.yaml
```diff
@@ -33,15 +33,19 @@
...
 - target: //daml-lf/interface:interface
   type: jar-scala
 - target: //daml-lf/interpreter:interpreter
   type: jar-scala
+- target: //daml-lf/kv-transaction-support:kv-transaction-support
+  type: jar-scala
 - target: //daml-lf/language:language
   type: jar-scala
-- target: //daml-lf/kv-transaction-support:kv-transaction-support
+- target: //daml-lf/parser:parser
```
cocreature, line 48: Same question here, why do we need this? This shouldn’t be something you rely on outside of `//daml-lf`.

release/src/Main.hs
cocreature: The workaround is fine with me and we had it before. The problem is that the check that we don’t forget to include transitive dependencies is broken. It should be checking the what the jar depends on. What it actually checks is all build time dependencies via `bazel query`. As soon as you depend on a DAR somehow you pull in the compiler which depends on those 3 libs.

I think trying to fix the `bazel query` is relatively hopeless, we should do the check based on the pom files or something like that but that’s a larger change.

I think we should copy parts of my explanation here into a comment but apart from that, I’m fine with this.

cocreature (changes requested):
Mostly looks good to me but I’d like to avoid publishing the daml-lf encoder & parser or at least a good explanation of why we need this.

release/artifacts.yaml
remyhaemmerle-da, line 37: I am a bit worried we start releasing the parser and the encoder. 
1. the quality of those libraries is not so good (in particular parsing error are pretty bad)
2  the syntax handle by the parser is evolved in a non backward compatible way. 
3. their is not proper documentation

Currently, the language team does not have the resources to improve any of this 3 point.

If the code handle by the parser stays in the main repo, the language team will be able to fix it as the syntax as the parser evolved.

ledger-api/testing-utils/src/main/scala/com/digitalasset/ledger/api/testing/utils/AkkaBeforeAndAfterAll.scala
hubert-da: @cocreature for one of our tests in the oem integration kit repo, one thread is not sufficient as it gets blocked and another one has to unblock it.

release/artifacts.yaml
hubert-da: As discussed offline, I'm going to try to avoid publishing the parser and the encoder by building the required DAR files beforehand.

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/SimplePackage.scala
```diff
@@ -114,25 +48,40 @@ class SimplePackage(additionalContractDataType: String) {
       templateId: Ref.Identifier,
       choiceName: Ref.ChoiceName,
   ): Command =
-    ExerciseCommand(templateId, contractId, choiceName, ValueUnit)
+    ExerciseCommand(
+      templateId,
+      contractId,
+      choiceName,
+      ValueRecord(Some(typeConstructorId(s"Simple:$choiceName")), ImmArray.empty),
```
hubert-da, line 55: ValueUnit doesn't work anymore. It took me a while to work around it.

ledger/test-common/BUILD.bazel
```diff
@@ -109,6 +112,12 @@ da_scala_library(
         ),
         ("performance", []),
         ("package_management", []),
+        ("simple_package_party", []),
+        ("simple_package_optional", []),
+        ("simple_package_list", []),
+        ("simple_package_text_map", []),
+        ("simple_package_variant", []),
+        ("simple_package_tuple", []),
```
SamirTalwar-DA, line 120: As discussed earlier, these belong in their own Bazel file under _kvutils_; they're not intended to be used by anything else.

We also don't need to generate them per DAML-LF version, but only for the dev version.

ledger/test-common/BUILD.bazel
```diff
@@ -109,6 +112,12 @@ da_scala_library(
         ),
         ("performance", []),
         ("package_management", []),
+        ("simple_package_party", []),
```
hubert-da, line 115: @SamirTalwar-DA I know you wanted to move these to `kvutils` and I tested it but then I would have to publish `bazel_tools`

SamirTalwar-DA: What do you mean by "publish"?

release/src/Main.hs
hubert-da: I added a comment

ledger/test-common/BUILD.bazel
hubert-da: It would have to go to Artifactory

ledger/test-common/BUILD.bazel
hubert-da: Sorry, answered below.

ledger/test-common/BUILD.bazel
hubert-da: As I would have to add the DARs as bazel `data` and access with `rlocation`.

SamirTalwar-DA: Why can't we package them inside JARs just as we do here?

hubert-da: I can do it but then I'm wondering whether it's more reasonable to increase encapsulation or reduce duplication.

hubert-da: By duplication I mean `BUILD.bazel` files and I won't be able to reuse the `TestDarReader` in this shape. Please let me know which approach is better.

SamirTalwar-DA: I think I'm gonna need @cocreature to weigh in here.

ledger/test-common/src/main/daml/simple_package_list/Simple.daml
```diff
@@ -0,0 +1,30 @@
...
+   with
+     owner : Party
+     observer : Party
+     contractData : [Party]
+   where
+     signatory owner
+     observer  observer
+     key owner : Party
+     maintainer key
+     choice Consume : ()
```
cocreature, line 13: Isn’t this just `Archive`?

ledger/test-common/src/main/daml/simple_package_optional/Simple.daml
```diff
@@ -0,0 +1,30 @@
...
+     observer : Party
+     contractData : Optional Party
+   where
+     signatory owner
+     observer  observer
+     key owner : Party
+     maintainer key
+     choice Consume : ()
+       controller owner
+       do pure ()
```
cocreature, line 15: Same here, looks like `Archive`.

ledger/test-common/src/main/daml/simple_package_optional/Simple.daml
```diff
@@ -0,0 +1,30 @@
+module Simple where
+data SimpleVariant = SV Party
+template SimpleTemplate
+   with
+     owner : Party
+     observer : Party
+     contractData : Optional Party
```
cocreature, line 7: Spotting the difference between those files is quite hard and keeping them in sync even more so. Can we instead have one template and throw `sed` or something at it to replace this one type?

release/artifacts.yaml
```diff
@@ -39,10 +39,10 @@
   type: jar-scala
 - target: //daml-lf/interpreter:interpreter
   type: jar-scala
-- target: //daml-lf/language:language
-  type: jar-scala
 - target: //daml-lf/kv-transaction-support:kv-transaction-support
   type: jar-scala
+- target: //daml-lf/language:language
```
cocreature, line 44: why did you reshuffle this?

ledger/test-common/BUILD.bazel
cocreature: I agree with @SamirTalwar-DA here. If we want to avoid duplication, we can factor it out in a macro that we share between here and `//kvutils/something`. I also prefer embedding it in JARs over publishing the dars directly.

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/SimplePackage.scala
cocreature: Maybe worth factoring out the choice argument given that you reuse it everywhere?

release/artifacts.yaml
hubert-da: I had a merge conflict and moved it manually with the proper alphabetical order :)

ledger/test-common/BUILD.bazel
hubert-da: will do

ledger/test-common/src/main/daml/simple_package_list/Simple.daml
hubert-da: It is but to use that I would have to add logic that apart from checking the package id the type is in, it would have to distinguish between modules. The `Archive` choice is in `DA.Internal.Template` while `Consume` and others are in `Simple`. Please let me leave it as is.

cocreature: Can we just set the type constructor id to `None` in `ValueRecord` to avoid having to figure out the package id? That said, not opposed to keeping it if it makes things significantly easier but then let’s please add a comment documenting why we do this.

hubert-da: Yup, worked, done. Thanks

ledger/test-common/src/main/daml/simple_package_optional/Simple.daml
hubert-da: After simplifying them with `Archive` it should be easier. I don't think deduplicating them would be trivial, given the current shape of the Bazel files. Let me know if you think it's still worth spending more time on it. I would leave it as is, now.

bazel_tools/scala.bzl
```diff
@@ -652,3 +654,77 @@ def da_scala_benchmark_jmh(
     deps = resolve_scala_deps(deps, scala_deps, versioned_scala_deps)
     runtime_deps = resolve_scala_deps(runtime_deps, scala_runtime_deps)
     _wrap_rule_no_plugins(scala_benchmark_jmh, deps, runtime_deps, **kwargs)
+
+def to_camel_case(name):
```
cocreature, line 658: This file is mainly intended for very generic Scala rules. Maybe move this to something more specific in `//ledger/test-common`?

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -29,6 +29,7 @@ import com.daml.ledger.api.refinements.{ApiTypes => lar}
 import com.daml.ledger.api.v1.{value => v}
 import com.daml.ledger.client.LedgerClient
 import com.daml.ledger.service.MetadataReader
+import com.daml.ledger.test.ModelTestDar
```
cocreature, line 32: Why did the name change?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Dars.scala
```diff
@@ -3,14 +3,14 @@
 
 package com.daml.ledger.api.testtool.infrastructure
 
+import com.daml.ledger.test.TestDar
```
cocreature, line 6: Same question here, can we leave the existing targets untouched?

cocreature (approved):
LGTM,

I’ll leave it to @SamirTalwar-DA and you to decide whether you want to factor out the daml sources. I’d probably do it but I’m not the one who has to keep them in sync so your choice :slightly_smiling_face: 

You might want to reshuffle the bazel code to a more specific file and ideally I’d preserve the existing target and package names in `test-common` to make this change less invasive.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
hubert-da: I introduced a sealed trait and case objects to more type-safe and less stringly-typed :)

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Dars.scala
hubert-da: Not really, I think (see https://github.com/digital-asset/daml/pull/9143/files#diff-d5719626ac0f56dec36be3531a23a53ba46b74c7990ca1f862bab3e31255e6d0R689)

bazel_tools/scala.bzl
hubert-da: You mean I can create a `bzl` file there and not in `bazel_tools`?

cocreature: exactly

hubert-da: sure, will do

bazel_tools/scala.bzl
```diff
@@ -652,3 +654,77 @@ def da_scala_benchmark_jmh(
...
+
+object TestDar {
+  val paths: List[String] = List(
+EOF
+""" + "\n".join(["""
+echo "    \\"%s/%s-tests-%s.dar\\"," >> $@
+""" % (native.package_name(), test_name, lf_version) for test_name in daml_dir_names]) + """
+echo "  )\n}\n" >> $@
+""" + "\n".join(["""
+echo "case object %sTestDar extends TestDar { val path = \\"%s/%s-tests-%s.dar\\" }" >> $@
```
SamirTalwar-DA, line 699: This magic is why we pay you the big bucks. :moneybag:

bazel_tools/scala.bzl
```diff
@@ -652,3 +654,77 @@ def da_scala_benchmark_jmh(
...
+        # 2. Generate lookup objects
+        genrule_name = "test-dar-lookup-%s" % lf_version
+        genrule_command = """
+cat > $@ <<EOF
+package com.daml.ledger.test
+
+sealed trait TestDar { val path: String }
+
+object TestDar {
+  val paths: List[String] = List(
```
SamirTalwar-DA, line 692: Can we make this a `List[Path]`?

bazel_tools/scala.bzl
SamirTalwar-DA, line 699: And perhaps make `path` a `Path`?

SamirTalwar-DA (approved):

bazel_tools/scala.bzl
hubert-da: I thought about this and didn't do it because:
1. We use it mostly with `getResouceAsStream` which takes a String, so I don't see benefits here. I also checked for alternatives - with no luck.
2. This would require an additional import.
Is it fine to leave it?

bazel_tools/scala.bzl
hubert-da: If you see magic anywhere it might be a bad sign that something needs to be changed 😃

hubert-da:
@SamirTalwar-DA @cocreature I had to make `test-common` lf version dependent, as the conformance tests run into jar hell. Please let me know if you are fine with the last changes. Shall I use `Squash and merge`, then?

release/src/Main.hs
```diff
@@ -40,6 +40,13 @@ import qualified SdkVersion
 scalaVersions :: [T.Text]
 scalaVersions = ["2.12.12", "2.13.3"]
 
+-- Hack: the logic that looks for missing transitive dependencies is broken, as it checks all build-time dependencies with `bazel query`
```
miklos-da, line 43: Do we have a GH/KVL issue to keep track of this? If not, let's create one.

miklos-da (approved):

release/src/Main.hs
hubert-da: Created: https://github.com/digital-asset/daml/issues/9195

PR #9144, opened 2021-03-15
Make async commit configurable and add "LOCAL" async commit option

oliverse-da:
In replicated Postgres setups, running with SYNCHRONOUS_COMMIT=LOCAL is a
suitable option as it is a trade off to only locally commit synchronously
but not to wait for a replication / network ack.

Therefore we'd like to make the async-commit configurable and also change it to an enum.

cc @matthiasS-da 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
@@ -40,4 +40,17 @@ private[platform] object DbType {
     case _ =>
       sys.error(s"JDBC URL doesn't match any supported databases (h2, pg)")
   }
+
+  sealed trait AsyncCommitMode {
+    def setting: String
+  }
+  object SynchronousCommit extends AsyncCommitMode {
+    override val setting: String = "ON"
+  }
+  object AsynchronousCommit extends AsyncCommitMode {
+    override val setting: String = "OFF"
+  }
+  object LocalSynchronousCommit extends AsyncCommitMode {
+    override val setting: String = "LOCAL"
+  }
 }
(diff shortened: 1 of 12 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
```diff
@@ -40,4 +40,17 @@ private[platform] object DbType {
...
+  sealed trait AsyncCommitMode {
+    def setting: String
+  }
+  object SynchronousCommit extends AsyncCommitMode {
+    override val setting: String = "ON"
+  }
+  object AsynchronousCommit extends AsyncCommitMode {
+    override val setting: String = "OFF"
+  }
+  object LocalSynchronousCommit extends AsyncCommitMode {
```
tudor-da, line 53: I am not sure what `LOCAL` brings to the plate here. As I understand it behaves the same as `ON` (sync commit), unless there are standby PG replicas. And in that case, do we want `LOCAL` or would we be better off with just `ON`?

tudor-da: Nevermind, I just noticed the linked PR. Thanks!

tudor-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
oliverse-da: I added the following link to the description above: https://www.percona.com/blog/2020/08/21/postgresql-synchronous_commit-options-and-synchronous-standby-replication/

PR #9146, opened 2021-03-16
Support fetching SDK EE tarball in the assistant

cocreature:
This PR adds an `artifactory-api-key` field to `daml-config.yaml`. If
this is set, we will download the SDK EE tarball if it exists, falling
back to GH releases on 404s (for old releases).

changelog_begin

- [Daml Assistant] The assistant now supports an `artifactory-api-key`
  field in `daml-config.yaml`. If you have a license for Daml Connect
  EE you can specify this and the assistant will automatically fetch
  the EE edition which provides additional functionality.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-assistant/src/DA/Daml/Assistant/Install.hs
+++ b/daml-assistant/src/DA/Daml/Assistant/Install.hs
@@ -25,14 +25,13 @@ import DA.Daml.Assistant.Version (getLatestSdkSnapshotVersion, getLatestReleaseV
 import DA.Daml.Project.Consts
 import DA.Daml.Project.Config
 import Safe
-import Data.Maybe
 import Data.List
 import Conduit
 import qualified Data.Conduit.List as List
 import qualified Data.Conduit.Tar.Extra as Tar
 import qualified Data.Conduit.Zlib as Zlib
 import Data.Either.Extra
-import Network.HTTP.Client.Conduit (HttpExceptionContent(StatusCodeException))
+import qualified Data.SemVer as SemVer
 import Network.HTTP.Simple
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.UTF8 as BS.UTF8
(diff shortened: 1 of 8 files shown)


daml-assistant/src/DA/Daml/Assistant/Install.hs
```diff
@@ -300,20 +307,34 @@ extractAndInstall env source =
...
-        when (getResponseStatusCode response /= 200) $
-            throwIO . assistantErrorBecause "Failed to download release."
-                    . pack . show $ getResponseStatus response
-        let totalSizeM = readMay . BS.UTF8.toString =<< headMay
-                (getResponseHeader "Content-Length" response)
-        extractAndInstall env
-            . maybe id (\s -> (.| observeProgress s)) totalSizeM
-            $ getResponseBody response
+    requiredAny "Failed to download SDK release." $
+        firstNon404 (map tryLocation (maybeToList (fmap (Artifactory.versionLocation version) artifactoryApiKeyM) <> [Github.versionLocation version]))
```
sofiafaro-da, line 314: IIUC, when the EE SDK release isn't available on artifactory, it installs the CE SDK from github instead.

Is this really desirable behavior? Would it be better to just fail, and let the user know that the EE release couldn't be found, and give them a command-line that lets them install the CE instead?

I'm (also) concerned people will get stuck with the wrong version by accident.

cocreature: Older versions (before the most recent snapshot) don’t have a separate EE edition. I think asking users to be aware since when we started publishing the enterprise edition is too annoying.

That leaves us with two options. Either we do what we do here and try artifactory first and go to GH releases on 404s. The alternative would be to hardcode the version number of the first snapshot and only try artifactory for newer versions and only try GH releases for older versions.

I don’t feel too strongly about this. Hardcoding the version in the assistant seemed a bit uglier and less robust but if you disagree, I’m happy to switch to that model.

sofiafaro-da: I prefer hardcoding the version number for the first release that's available on artifactory. Or alternatively, have a cut-off point at which it will no longer use Github as an automatic backup. I think these will lead to better behaviour for our users.

cocreature: changed in 435c5b2208e1191e3724d02fa2794aae50644b6b

sofiafaro-da: Thanks!

sofiafaro-da (approved):
LGTM, thank you!

PR #9150, opened 2021-03-16
Handle createAndExecuteCmd

### LINKED ISSUE #9127: Handle createAndExecuteCmd
Right now `createAndExecuteCmd` is not handled properly by Daml script dumps. Instead of one `createAndExecuteCmd` invocation it will generate separate `createCmd` and `executeCmd` invocations. The latter will depend on an undefined variable.

E.g. the following Daml script:

``` haskell
template Contract
  with owner : Party
  where signatory owner

setup : Script ()
setup = script do
  alice <- allocatePartyWithHint "Alice" (PartyIdHint "Alice")

  submit alice do
    createAndExerciseCmd
      Contract with owner = alice
      Archive
```

Results in the following Daml script dump:

``` haskell
dump : Parties -> Script ()
dump Parties{..} = do
  tree <- submitTree alice_0 do
    createCmd Main.Contract with
      owner = alice_0
    exerciseCmd contract_1_0 DA.Internal.Template.Archive
  let contract_1_0 = createdCid @Main.Contract [0] tree
  pure ()
```

`contract_1_0` is not in scope when `exerciseCmd` is called on it.

aherrmann-da:
Closes #9127

Identifies `createAndExercise` commands in Daml script dumps.
- Adds a new type to distinguish create, exercise, and createAndExercise commands.
- Transaction trees are converted to a sequence of such commands before further processing for encoding.
- Other components, such as `SimpleEvent` are based on commands instead, i.e. `SimpleCommand`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
@@ -230,6 +230,34 @@ private[dump] object Encode {
     case TreeEvent.Kind.Empty => throw new IllegalArgumentException("Unknown tree event")
   }
 
+  private def encodeCmd(
+      partyMap: Map[Party, String],
+      cidMap: Map[ContractId, String],
+      cmd: Command,
+  ): Doc = cmd match {
+    case CreateCommand(createdEvent) =>
+      encodeCreatedEvent(partyMap, cidMap, createdEvent)
+    case ExerciseCommand(exercisedEvent) =>
+      Doc.text("exerciseCmd ") + encodeCid(
+        cidMap,
+        ContractId(exercisedEvent.contractId),
+      ) + Doc.space + encodeValue(
+        partyMap,
+        cidMap,
+        exercisedEvent.getChoiceArgument.sum,
+      )
+    case CreateAndExerciseCommand(createdEvent, exercisedEvent) =>
+      Doc
+        .stack(
+          Seq(
+            Doc.text("createAndExerciseCmd"),
+            encodeRecord(partyMap, cidMap, createdEvent.getCreateArguments),
+            encodeValue(partyMap, cidMap, exercisedEvent.getChoiceArgument.sum),
+          )
+        )
+        .hang(2)
+  }
+
   private def encodeCreatedEvent(
       partyMap: Map[Party, String],
       cidMap: Map[ContractId, String],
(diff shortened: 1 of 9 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
```diff
@@ -30,7 +29,11 @@ private[dump] object Encode {
     val partyMap = partyMapping(parties)
 
     val acsCidRefs = acs.values.foldMap(createdReferencedCids)
-    val treeCidRefs = trees.foldMap(treeReferencedCids)
+    // TODO[AH] Avoid constructing Commands twice. Currently we do this once here and once more in encodeTree.
```
aherrmann-da, line 32: I'll address this in a follow-up PR.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -159,39 +160,97 @@ object TreeUtils {
...
+          val creates = treeEventCreatedCids(Kind.Exercised(exercisedEvent), tree)
           (result, creates) match {
             case (Some(cid), Seq(createdCid)) if cid == createdCid =>
-              Some(SimpleEvent(exercised, ContractId(cid)))
+              Some(SimpleCommand(command, ContractId(cid)))
             case _ => None
           }
-        case Kind.Empty => None
+        case _: CreateAndExerciseCommand =>
+          // TODO[AH] Identify simple createAndExercise commands.
```
aherrmann-da, line 246: I'll address this in a follow-up PR.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -159,39 +160,97 @@ object TreeUtils {
...
+                    )
+                  case _ =>
+                    throw new RuntimeException(
+                      "Internal error: Encountered `created` entry without corresponding `commands` entry."
+                    )
+                }
+            }
+        }
+      }
+      commands.valuesIterator.toSeq
```
cocreature, line 217: This seems to return a more or less random order. I think you have to keep track of the order in which you encounter events (I don’t think event ids provide ordering guarantees).

cocreature (approved):
Thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -159,39 +160,97 @@ object TreeUtils {
...
+      val created = mutable.HashMap.empty[ContractId, EventId]
+      tree.rootEventIds.foreach { eventId =>
+        val event = tree.eventsById(eventId)
+        event.kind match {
+          case Kind.Empty =>
+          case Kind.Created(createdEvent) =>
+            commands += EventId(eventId) -> CreateCommand(createdEvent)
+            created += ContractId(createdEvent.contractId) -> EventId(eventId)
+          case Kind.Exercised(exercisedEvent) =>
+            created.get(ContractId(exercisedEvent.contractId)) match {
```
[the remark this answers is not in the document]
cocreature: https://github.com/digital-asset/daml/issues/9155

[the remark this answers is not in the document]
aherrmann-da: Good point, the logic only applies to the root events of the tree. I was only taking `create`, `createAndExercise`, and `exercise` into consideration. In which case, I don't think this would pose an issue. However, I forgot about `exerciseByKeyCmd`. Using that it is possible to construct root events where the ordering matters as described here. See https://github.com/digital-asset/daml/issues/9155#issuecomment-800924621.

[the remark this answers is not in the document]
cocreature: The roots that you can use when submitting are not identical to the roots that you will get on a transaction tree. The projection my party sees can have nodes as roots which were not roots originally.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: `commands` is a `LinkedHashMap` which returns items in insertion order. So the order is not random, but the order in which we encounter events.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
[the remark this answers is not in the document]
aherrmann-da: I've changed it so that only adjacent create & exercise events can be converted to `createAndExerciseCmd` and added a test-case for this situation.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: https://github.com/digital-asset/daml/pull/9154

daml-script/dump/src/main/scala/com/daml/script/dump/Encode.scala
aherrmann-da: https://github.com/digital-asset/daml/pull/9179

PR #9151, opened 2021-03-16
Add a test for create and exercise events in the profiler

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
@@ -243,7 +243,7 @@ private[lf] final class Compiler(
     ref ->
       SDefinition(
         unsafeClosureConvert(
-          SEAbs(arity, withLabel(ref, body(List.fill(arity)(nextPosition()))))
+          withLabel(ref, SEAbs(arity, body(List.fill(arity)(nextPosition()))))
         )
       )
 
--- a/daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ProfilerTest.scala
+++ b/daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ProfilerTest.scala
@@ -1,0 +1,124 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf
+package speedy
+
+import com.daml.lf.data._
+import com.daml.lf.language.Ast._
+import com.daml.lf.speedy.PartialTransaction._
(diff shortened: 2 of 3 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -241,7 +241,7 @@ private[lf] final class Compiler(
     ref ->
       SDefinition(
         unsafeClosureConvert(
-          SEAbs(arity, withLabel(ref, body(List.fill(arity)(nextPosition()))))
+          withLabel(ref, SEAbs(arity, body(List.fill(arity)(nextPosition()))))
```
cocreature, line 244: This was just broken. We want the label around the closure. Otherwise, we get to https://github.com/digital-asset/daml/blob/98c2651998cce1fd1b8776e488d8f8cfbcf3023d/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala#L1429 and we don’t have a closure and we don’t generate any entries. If you look at the original PR https://github.com/digital-asset/daml/commit/caedc7255111d564d872e70582c371086b30a679# that’s also what happened there afaict. Haven’t tracked down when exactly it broke. That’s why we should have tests for things :slightly_smiling_face:

remyhaemmerle-da: Not sure I should review this PR as I am the culprit for the mess.

cocreature: The culprit is not having tests not changing code :slightly_smiling_face: 

Let me try to provide some more context: First in terms of the tests you can easily see the difference. Without this change, we don’t get any of the create and exercise nodes. This is clearly wrong.

Now how does profiling work?
1. In the compiler we wrap `SELabelClosure` around some closures, e.g., this one (the fact that we didn’t wrap it around the closure anymore already indicates that it was probably broken).
2. When executing `SELabelClosure` we push a `KLabelClosure`.
3. When we execute `KLabelClosure`, the return value should now be a closure. We add the label stored in `KLabelClosure` to that closure.
4. Now when we get to the `KFun` that applies this closure, we start the timer for the label of the closure and we push a `KLeaveClosure`.
5. When executing `KLeaveClosure we stop the timer for the label.

Everything here is centered around timing the application of a closure. Afaict, `KLabelClosure` should even just error out if it doesn’t get a closure but I’ll leave that for a separate PR.

Does that help?

remyhaemmerle-da: make sens.

remyhaemmerle-da (approved):
LGTM.

nickchapman-da (approved):
LGTM. Thanks!

PR #9152, opened 2021-03-16
Drop profiler from Sandbox CE

cocreature:
Now that we have an EE edition, we can drop this. Approved by @bame-da

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):
Is this something that should be mentioned in the changelog?

cocreature:
> Is this something that should be mentioned in the changelog?

It was a hidden option so I don’t think so. The addition of the EE edition has a changelog entry and the release notes will of course mention the differences in the EE edition.

PR #9153, opened 2021-03-16
pkgid data deps

robin-da:
### Pull Request Checklist

trigger the build.


compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -175,24 +184,58 @@ dalfFileName bs fp = do
...
+    pure $ DataDeps {dataDepsDars = dars, dataDepsDalfs = dalfs, dataDepsPkgIds = pkgIds}
+  where
+    (dars, rest) = partition ("dar" `isExtensionOf`) fpOrIds
+    (dalfs, pkgIds0) = partition ("dalf" `isExtensionOf`) rest
+
+-- | A rudimentary check that no bad package ID's are present in the data-dependency section of
+-- daml.yaml.
+validatePkgId :: String -> IO LF.PackageId
+validatePkgId pkgId = do
+    unless (all (`elem` ['a' .. 'z'] ++ ['A' .. 'Z'] ++ ['0' .. '9']) pkgId) $
```
cocreature, line 215: Don’t we store this in base16 with only lowercase letters? Checking that it’s a 64-letter hex string seems easy enough so I’d rather do it properly than use the approximation here.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -175,24 +184,58 @@ dalfFileName bs fp = do
...
+        fail $ "Invalid package ID dependency in daml.yaml: " <> pkgId
+    pure $ LF.PackageId $ T.pack pkgId
+
+-- Ledger interactions
+----------------------
+
+getDalfsFromLedger :: [LF.PackageId] -> IO [RemoteDalf]
+getDalfsFromLedger pkgIds
+    | null pkgIds = pure []
+    | otherwise = runLedgerGetDalfs (defaultLedgerFlags Grpc) pkgIds
```
cocreature, line 225: Can we move this logic in `runLedgerGetDalfs`? Seems less error prone than shortcircuiting here.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -110,8 +111,14 @@ installDependencies ::
    -> IO ()
 installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pDataDeps = do
     deps <- expandSdkPackages (optDamlLfVersion opts) (filter (`notElem` basePackages) pDeps)
+    DataDeps {dataDepsDars, dataDepsDalfs, dataDepsPkgIds} <- readDataDeps pDataDeps
     (needsUpdate, newFingerprint) <-
-        depsNeedUpdate depsDir (deps ++ pDataDeps) thisSdkVer (show $ optDamlLfVersion opts)
+        depsNeedUpdate
+            depsDir
+            (deps ++ dataDepsDalfs)
```
cocreature, line 118: What about `dataDepsDars`? Don’t we need to include them in the check as well?

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -293,9 +297,9 @@ downloadAllReachablePackages args pid = loop Map.empty [pid]
         then loop acc morePids
         else do
           pkg <- downloadPackage args pid
-          loop (Map.insert pid pkg acc) (packageRefs pkg ++ morePids)
+          loop (Map.insert pid pkg acc) (nubSort $ packageRefs pkg ++ morePids)
```
cocreature, line 300: can we use a Set instead of `nubSort`ing a list? I think it might also be benefitial to only insert `packageRefs pkg` that aren’t already in the map instead of first inserting them here only to then drop them in the next step.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -323,6 +327,24 @@ downloadPackage args pid = do
...
+    }
+runLedgerGetDalfs :: LedgerFlags -> [LF.PackageId] -> IO [RemoteDalf]
+runLedgerGetDalfs lflags pkgIds = do
+    args <- getDefaultArgs lflags
+    m <- downloadAllReachablePackages args pkgIds
+    pure
+        [ RemoteDalf {..}
+        | (i :: Int, (_pid, pkg)) <- zip [0 ..] $ Map.toList m
+        , let (bsl, pid) = LFArchive.encodeArchiveAndHash pkg
+        , let remoteDalfFp = (T.unpack $ recoverPackageName pkg ("dep" <> show i, pid)) <.> "dalf"
```
cocreature, line 343: Can we default to the package id as the name intead of `dep$i`? The latter doesn’t really provide any meaningful info.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -323,6 +327,24 @@ downloadPackage args pid = do
     convPid :: LF.PackageId -> L.PackageId
     convPid (LF.PackageId text) = L.PackageId $ TL.fromStrict text
 
+data RemoteDalf = RemoteDalf
+    { remoteDalfFp :: FilePath
```
cocreature, line 331: I think it might be helpful to add a comment that this isn’t the filepath we’ve written to, it’s the file path we’re planning to write it to. Alternatively calli t something like `remoteDalfName`, make it a string and add the extension at the callsite.

compiler/damlc/tests/src/DamlcPkgManager.hs
```diff
@@ -0,0 +1,68 @@
...
+                      unlines
+                          [ "module A where"
+                          , "import PkgManagerTest (S)"
+                          , "type U = S"
+                          , "template T with p : Party where"
+                          , "  signatory p"
+                          ]
+                  setEnv "DAML_PROJECT" projDir True
+                  (exitCode, _stdout, stderr) <-
+                      readProcessWithExitCode damlc ["build", "--project-root", projDir] ""
```
cocreature, line 65: you shouldn’t need `--project-root` if `DAML_PROJECT` is set.

cocreature (approved):
Nice! A few small comments.

I’d like to see some more detailed tests for the caching logic in particular now that we have two levels of caching (maybe we can factor the logic for that nicely so we don’t have to setup a bunch of integration tests for that) but that can be in a separate PR.

compiler/damlc/tests/src/DamlcPkgManager.hs
robin-da: that's true, changed.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: that's a good catch, fixed.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: sure, I moved it.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: switched to `isBase64` and `all isLower` for the check.

cocreature: why base64? it should be base16 with 64 characters (256 bits because sha256).

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
robin-da: sure, I choose the `dep` prefix in order to be aligned on how the `fetch-dar` command names the dalfs. should we change that as well?k

cocreature: yeah I think I’d change both

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
robin-da: renamed to `remoteDalfName`.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
robin-da: ok, done

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
robin-da: I added a filter for the `packageRefs pkg` so that we exclude package ids that already in the accumulator. I think we should keep the list and `nubSort` because we loop over the list. Otherwise, you'll end up with a lot of `toList/fromList` in the iteration.

cocreature: You can use `minView` https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Set.html#v:minView but fair enough, I don’t feel strongly about this at all.

robin-da: ah cool, I didn't know `minView`. Changed it to use this.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -175,24 +186,55 @@ dalfFileName bs fp = do
...
+  where
+    (dars, rest) = partition ("dar" `isExtensionOf`) fpOrIds
+    (dalfs, pkgIds0) = partition ("dalf" `isExtensionOf`) rest
+
+-- | A check that no bad package ID's are present in the data-dependency section of daml.yaml.
+validatePkgId :: String -> IO LF.PackageId
+validatePkgId pkgId = do
+    unless ((Base64.isBase64 $ T.pack pkgId) && all isLower (filter isAlpha pkgId)) $
+        fail $ "Invalid package ID dependency in daml.yaml: " <> pkgId
+    pure $ LF.PackageId $ T.pack pkgId
```
sofiafaro-da, line 218: This isn't right ... package ids are lowercase hex strings (with length 64), like @cocreature said, not base64. This `validatePkgId` will accept strings like `"helloworld"` when it should reject them.

robin-da: ah no, I must have misread the comment. I thought the request was explicitly to check for base64. I'll fix it in the next PR.

PR #9154, opened 2021-03-16
Recognize simple createAndExercise commands

aherrmann-da:
**depends on #9150**

This addresses https://github.com/digital-asset/daml/pull/9150#discussion_r595105162

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
@@ -241,6 +241,7 @@ object TreeUtils {
           if (exercisedEvent.consuming) {
             // If the choice is not consuming then we have two resulting contracts:
             // The created one and the result of the exercise. I.e. not a simple command.
+            // A child exercised event might still consume the contract, but this heuristic is kept simple for now.
             simpleExercise(exercisedEvent).map(SimpleCommand(command, _))
           } else {
             None
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
@@ -214,25 +214,33 @@ object TreeUtils {
 
   object SimpleCommand {
     def fromCommand(command: Command, tree: TransactionTree): Option[SimpleCommand] = {
+      def simpleExercise(exercisedEvent: ExercisedEvent): Option[ContractId] = {
+        val result = exercisedEvent.exerciseResult.flatMap {
+          _.sum match {
+            case Sum.ContractId(value) => Some(value)
+            case _ => None
+          }
+        }
+        val creates = treeEventCreatedCids(Kind.Exercised(exercisedEvent), tree)
+        (result, creates) match {
+          case (Some(cid), Seq(createdCid)) if cid == createdCid =>
+            Some(ContractId(cid))
+          case _ => None
+        }
+      }
       command match {
         case CreateCommand(createdEvent) =>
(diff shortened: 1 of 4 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -214,25 +218,33 @@ object TreeUtils {
...
-          val creates = treeEventCreatedCids(Kind.Exercised(exercisedEvent), tree)
-          (result, creates) match {
-            case (Some(cid), Seq(createdCid)) if cid == createdCid =>
-              Some(SimpleCommand(command, ContractId(cid)))
-            case _ => None
+          simpleExercise(exercisedEvent).map(SimpleCommand(command, _))
+        case CreateAndExerciseCommand(_, exercisedEvent) =>
+          if (exercisedEvent.consuming) {
+            // If the choice is not consuming then we have two resulting contracts:
+            // The created one and the result of the exercise. I.e. not a simple command.
```
cocreature, line 243: Non-consuming choices can still call consuming choices (e.g., `Archive`). Totally fine not to handle this case but perhaps worth mentioning in the comment.

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeSimpleSpec.scala
```diff
@@ -33,15 +35,26 @@ class EncodeSimpleSpec extends AnyFreeSpec with Matchers {
               exerciseResult = Some(ContractId("cid3")),
               actingParties = Seq(Party("Bob")),
             ),
+            TestData.Created(ContractId("cid4"), submitters = Seq(Party("Alice"))),
```
cocreature, line 38: Can we differentiate between `createAndExercise` and `create` + `exerciseByKey`? Seems perfectly fine if we translate the latter in the former, just curious.

cocreature (approved):
Thanks!

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: True, I've expanded the comment. It shouldn't be too hard to expand this to take nested consuming choices into account. I'll look into it separately, though.

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeSimpleSpec.scala
aherrmann-da: I'm not aware of a way to differentiate them in general. However, we do have access to a contract's contract key, if any, in the `createdEvent`. This could help us handle a few more cases of https://github.com/digital-asset/daml/issues/9155. Here's the approach I have in mind:
- If the exercise follows right after the create, just use `createAndExerciseCmd`.
- If the exercise follows later and the contract has a contract key, then use `exerciseByKeyCmd`.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Addressed in https://github.com/digital-asset/daml/pull/9185

daml-script/dump/src/test/scala/com/daml/script/dump/EncodeSimpleSpec.scala
aherrmann-da: Addressed in https://github.com/digital-asset/daml/pull/9185

PR #9158, opened 2021-03-16
Expose unhandled exception values and types

cocreature:
This PR addresses the TODO to expose the actual exception values and
type instead of the message. This allows us to simply speedy a bit by
removing the continuation used for that but more importantly it means
we can now catch and handle the exception in Daml script.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


remyhaemmerle-da (approved):
LGTM.

nickchapman-da (approved):
Nice! I really like the simplification here.
- not having to deal with weird special cases like message functions throwing
- and only calling the message function when explicitly requested by the builtin.

PR #9159, opened 2021-03-16
Oracle json-api: more queries

S11001001:
This extends #8596 with Oracle-compatible contract query compilation (excepting range queries), lookup by contract ID and key, `read committed`-safe contract insertion, and issues brought up in #8596 review. Like #8596, it is not meant to be complete support, only a checkpoint.

Addressed here and in followup PR:

* [x] query @>
* [ ] query `cmpContractPathToScalar`
* [x] special scalar =
* [ ] use or lose Array get/put
* [x] key =
* [x] [divergent text types](https://github.com/digital-asset/daml/pull/8596#discussion_r576311748)
* [ ] duplicate contracts for multiple sig/obs match (different sig/obs agg orders?)
* [x] OR instead of ANDing [the sig/obs check](https://github.com/digital-asset/daml/pull/9159/files#r603179621)
* [x] on conflict ignore
* [ ] use 2c14ff1's approach to [on conflict ignore](https://github.com/digital-asset/daml/pull/9159/files#r603176801)
* [ ] retrySqlStates
* [ ] >4kb JSON data tests
* [ ] variant value path test
* [ ] test comparing a string to "& bar"

### Pull Request Checklist

trigger the build.


ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -577,26 +675,117 @@ private object OracleQueries extends Queries {
...
+      case JsBoolean(_) | JsNull | JsNumber(_) | JsString(_) =>
+        equalAtContractPath(path, literal)
+
+      case JsObject(fields) =>
+        fields.toVector match {
+          case hp +: tp =>
+            // this assertFromString is forced by the aforementioned too-big type
+            val fieldPreds = OneAnd(hp, tp).map { case (ok, ov) =>
+              containsAtContractPath(path objectAt Ref.Name.assertFromString(ok), ov)
+            }
```
S11001001, line 766: This is a little simplistic, as explained in #9256.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -577,26 +675,117 @@ private object OracleQueries extends Queries {
...
+  // I cannot believe this function exists in 2021
+  // ORA-40454: path expression not a literal
+  private[this] def oraclePathEscape(readyPath: Cord): Fragment = {
+    val s = readyPath.toString
+    assert(
+      !s.startsWith("'") && !s.endsWith("'"),
+      "Oracle JSON query syntax doesn't allow ' at beginning or ending",
+    )
+    Fragment const0 ("'" + s.replaceAllLiterally("'", "''") + "'")
+  }
```
S11001001, line 732: Reviewers please take special care to examine this function.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -577,26 +675,117 @@ private object OracleQueries extends Queries {
...
+  // I cannot believe this function exists in 2021
+  // ORA-40454: path expression not a literal
+  private[this] def oraclePathEscape(readyPath: Cord): Fragment = {
+    val s = readyPath.toString
+    assert(
+      !s.startsWith("'") && !s.endsWith("'"),
+      "Oracle JSON query syntax doesn't allow ' at beginning or ending",
+    )
+    Fragment const0 ("'" + s.replace("'", "''") + "'")
+  }
```
S11001001, line 732: Reviewers, please take special care to examine this function.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -503,9 +594,9 @@ private object OracleQueries extends Queries {
     sql"""
       CREATE TABLE
         signatories
-          (contract_id NVARCHAR2(100) NOT NULL REFERENCES contract(contract_id) ON DELETE CASCADE
-          ,party NVARCHAR2(100) NOT NULL
-          ,UNIQUE (contract_id, party)
+          (contract_id NVARCHAR2(255) NOT NULL REFERENCES contract(contract_id) ON DELETE CASCADE
+          ,party NVARCHAR2(255) NOT NULL
```
cocreature, line 598: Any reason not to use `contractIdType` and `partyIdType` here?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -515,9 +606,9 @@ private object OracleQueries extends Queries {
     sql"""
       CREATE TABLE
         observers
-          (contract_id NVARCHAR2(100) NOT NULL REFERENCES contract(contract_id) ON DELETE CASCADE
-          ,party NVARCHAR2(100) NOT NULL
-          ,UNIQUE (contract_id, party)
+          (contract_id NVARCHAR2(255) NOT NULL REFERENCES contract(contract_id) ON DELETE CASCADE
+          ,party NVARCHAR2(255) NOT NULL
```
cocreature, line 610: same here

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -527,43 +618,50 @@ private object OracleQueries extends Queries {
...
   protected[this] override def primInsertContracts[F[_]: cats.Foldable: Functor](
       dbcs: F[DBContract[SurrogateTpId, DBContractKey, JsValue, Array[String]]]
   )(implicit log: LogHandler, pas: Put[Array[String]]): ConnectionIO[Int] = {
     println("insert contracts")
     println(dbcs)
-    val r = Update[(String, SurrogateTpId, JsValue, JsValue, String)](
+    val r = Update[(String, String, SurrogateTpId, JsValue, JsValue, String)](
       """
-        INSERT INTO contract(contract_id, tpid, key, payload, agreement_text)
+        MERGE INTO contract USING (SELECT 1 FROM DUAL) ON (contract_id = ?)
```
cocreature, line 631: Assuming the `MERGE INTO` stuff is to handle `on conflict do nothing`, I noticed that you can do that a bit easier https://github.com/digital-asset/daml/commit/2c14ff1e80538883156aacbb04ee2798ebfa8092#

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -527,43 +618,50 @@ private object OracleQueries extends Queries {
...
+          (c.contractId, c.contractId, c.templateId, c.key, c.payload, c.agreementText)
         }
     )
     println("inserted")
     import cats.syntax.foldable._, cats.instances.vector._
-    val r2 = Update[(String, String)](
+    val r2 = Update[(String, String, String, String)](
       """
-        INSERT INTO signatories(contract_id, party)
+        MERGE INTO signatories USING (SELECT 1 FROM DUAL) ON (contract_id = ? AND party = ?)
```
cocreature, line 647: same here

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -527,43 +618,50 @@ private object OracleQueries extends Queries {
...
       logHandler0 = log,
-    ).updateMany(dbcs.foldMap(c => c.signatories.view.map(s => (c.contractId, s)).toVector))
-    val r3 = Update[(String, String)](
+    ).updateMany(
+      dbcs.foldMap(c => c.signatories.view.map(s => (c.contractId, s, c.contractId, s)).toVector)
+    )
+    val r3 = Update[(String, String, String, String)](
       """
-        INSERT INTO observers(contract_id, party)
+        MERGE INTO observers USING (SELECT 1 FROM DUAL) ON (contract_id = ? AND party = ?)
```
cocreature, line 657: and here

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -577,26 +675,117 @@ private object OracleQueries extends Queries {
...
     println("selecting")
-    val q = sql"""SELECT contract_id, key, payload, agreement_text
-                  FROM contract
-                  WHERE tpid = $tpid""" // TODO SC AND (""" ++ predicate ++ sql")"
+    import Queries.CompatImplicits.catsReducibleFromFoldable1
+    // % is explicitly reserved by specification as a delimiter
+    val q = sql"""SELECT c.contract_id, key, payload, agreement_text, sd.parties, od.parties
+                  FROM (contract c
+                        JOIN signatories sm ON (c.contract_id = sm.contract_id)
+                        JOIN observers om ON (c.contract_id = om.contract_id))
```
cocreature, line 684: what happens if there are no observers or signatories for a given contract? Doesn’t the inner join here just filter them out?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -577,26 +675,117 @@ private object OracleQueries extends Queries {
...
   ): T[Query0[DBContract[Mark, JsValue, JsValue, Vector[String]]]] = {
     val Seq((tpid, predicate @ _)) = queries // TODO SC handle more than one
-    val _ = parties
     println("selecting")
-    val q = sql"""SELECT contract_id, key, payload, agreement_text
-                  FROM contract
-                  WHERE tpid = $tpid""" // TODO SC AND (""" ++ predicate ++ sql")"
+    import Queries.CompatImplicits.catsReducibleFromFoldable1
+    // % is explicitly reserved by specification as a delimiter
+    val q = sql"""SELECT c.contract_id, key, payload, agreement_text, sd.parties, od.parties
```
cocreature, line 681: maybe add a todo to clean up this query. We should probably either try to switch to arrays or try to only do one join per table.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
cocreature: Can you provide an example of a query that needs this? I think I’m not quite following which weird restriction we’re trying to work around here.

cocreature (commented):
Mostly lgtm, I’d like to understand the escaping before approving and it would also be great to have at least some very simple tests that hit some of this code.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: Every location that uses the function needs it. Function calls of the form `json_query(x, $y)` where `y: String` fail with `ORA-40454: path expression not a literal`. In other words, while paths use string _syntax_, you cannot put anything but a literal string in the SQL.

cocreature: Let me rephrase, in which case do we get a value that we need to escape? Daml-LF names cannot have `'` in them so record fields or variants are not an option. What can hit this?

ledger-service/http-json/src/main/scala/com/digitalasset/http/query/ValuePredicate.scala
```diff
@@ -76,18 +77,22 @@ sealed abstract class ValuePredicate extends Product with Serializable {
...
         raw: SqlWhereClause,
         safe_== : Option[JsValue],
         safe_@> : Option[JsValue],
     ) {
       def flush_@>(path: Path): Option[Fragment] =
-        safe_@> map (jq => path ++ sql" @> $jq::jsonb")
+        safe_@> map (containsAtContractPath(path, _))
       def flush_==(path: Path): Option[Fragment] =
-        safe_== map (jq => path ++ sql" = $jq::jsonb")
+        safe_== map (equalAtContractPath(path, _))
```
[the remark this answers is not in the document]
S11001001: Record queries are contains queries, which lift neatly into variants and optionals, so you can express a bunch of contains queries at once.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: I see, yes. I'll address that in the followup.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: Nice. I'll incorporate that into the next PR, though.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: `$.p?(@ == "I'm a string")` as produced by https://github.com/digital-asset/daml/pull/9159/files#diff-570f7c791ece7e9dc9ef545dae8b7bb7375560172359ddc5a6ed86477713d6c1R736-R739

cocreature: Makes sense, what is stopping us from relying on doobie’s escaping, e.g., something like `fr"${escapethis}"`?

cocreature: Oh I guess that goes via jdbc to do escaping so it’s not a literal to oracle?

S11001001: Exactly.

cocreature: ugh, alright, don’t have a better idea then. This is very sad :sob:

cocreature (approved):

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: Or 07059d0ac4b197dbc3568bc3e333fa8b2c84cead actually

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: e6a53be22a7b34c233110d63d1e5cd20cd87d50b

PR #9166, opened 2021-03-17
fix release job

garyverhaegen-da:
Three issues here:

1. The release job runs on an Azure-hosted agent, so it doesn't have the `reset_caches.sh` script (and doesn't need it).
2. The `bash-lib` step should not run if the current job has already failed.
3. The `skip-github` jobs should also not run if the job  has failed.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -62,7 +62,6 @@ jobs:
         set -euo pipefail
         git checkout $(release_sha)
       name: checkout_release
-    - template: ci/clean-up.yml
     - template: ci/bash-lib.yml
       parameters:
         var_name: bash-lib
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -79,17 +78,17 @@ jobs:
       inputs:
         artifactName: linux-release
         targetPath: $(Build.StagingDirectory)/release-artifacts
-      condition: not(eq(variables['skip-github'], 'TRUE'))
+      condition: and(succeeded(), not(eq(variables['skip-github'], 'TRUE')))
```
cocreature, line 81: I’m starting to wonder whether I should write a linter for azure pipelines configs :slightly_smiling_face:

cocreature (approved):
Thanks!

PR #9169, opened 2021-03-17
skip compat tests on release

garyverhaegen-da:
As requested by @coceature.

Note: skipping the ts_lib job is enough to skip all compat tests because they all depend on it, and in the Azure model if one of your dependencies was skipped you get skipped too.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -250,6 +250,8 @@ jobs:
   dependsOn:
     - da_ghc_lib
     - check_for_release
+  condition: and(succeeded(),
+                 not(eq(dependencies.check_for_release.outputs['out.is_release'], 'true')))
   timeoutInMinutes: 360
   pool:
     name: ubuntu_20_04


ci/build.yml
```diff
@@ -250,6 +250,9 @@ jobs:
   dependsOn:
     - da_ghc_lib
     - check_for_release
+  condition: and(succeeded(),
+                 not(and(eq(dependencies.check_for_release.outputs['out.is_release'], 'true'),
+                         eq(variables['Build.SourceBranchName'], 'main'))))
```
cocreature, line 255: do we want to run them on PRs? I guess there is no harm but it also doesn’t seem super useful

cocreature (approved):

ci/build.yml
garyverhaegen-da: Fair enough.

PR #9179, opened 2021-03-17
Avoid constructing Commands twice

aherrmann-da:
Addresses https://github.com/digital-asset/daml/pull/9150#discussion_r595105037

Adds case classes for the four cases
- `submit`
- `submitMulti`
- `submitTree`
- `submitTreeMulti`

Also groups them by single/multi submitter and simple/tree commands using additional traits.

The logic to classify submits is called earlier to avoid identifying commands twice, as we did before to correctly identify referenced contract ids taking `createAndExerciseCmd` into account.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice, thanks!

PR #9185, opened 2021-03-18
Recognize (some) exerciseByKey commands in Daml script dumps

aherrmann-da:
Addresses https://github.com/digital-asset/daml/pull/9154#discussion_r596678341

- Adds a separate command type for `exerciseByKeyCmd` next to `createCmd`, `exerciseCmd`, and `createAndExerciseCmd`.
- `exerciseByKeyCmd` are used when an exercise event exercises a choice on a contract that was created by the same transaction, is not eligible for `createAndExerciseCmd`, and has a contract key.
- This adds test-cases for recognizing `exerciseByKeyCmd`, distinguishing simple and non-simple commands, and encoding.
- This PR also takes the opportunity to broaden which exercises are considered simple as described in https://github.com/digital-asset/daml/pull/9154#discussion_r596678312. I.e. child events that consume contracts are taken into account when determining whether the command only creates one lasting contract in total.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
+++ b/daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
@@ -137,17 +137,20 @@ object TreeUtils {
     cids
   }
 
-  def treeEventCreatedCids(event: TreeEvent.Kind, tree: TransactionTree): Set[ContractId] = {
-    var creates = Set.empty[ContractId]
+  def treeEventCreatedConsumedCids(
+      event: TreeEvent.Kind,
+      tree: TransactionTree,
+  ): (Set[ContractId], Set[ContractId]) = {
+    var created = Set.empty[ContractId]
     var consumed = Set.empty[ContractId]
     traverseEventInTree(event, tree) {
       case (_, Kind.Created(value)) =>
-        creates += ContractId(value.contractId)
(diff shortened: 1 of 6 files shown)


daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -181,29 +187,47 @@ object TreeUtils {
...
         case Kind.Empty =>
         case Kind.Created(createdEvent) =>
+          createdEvent.contractKey.foreach { contractKey =>
+            contractKeys += ContractId(createdEvent.contractId) -> contractKey
+          }
           commands += CreateCommand(createdEvent)
         case Kind.Exercised(exercisedEvent) =>
-          commands.lastOption match {
-            case Some(CreateCommand(createdEvent))
+          lazy val optCreateAndExercise = commands.lastOption.flatMap {
```
cocreature, line 213: seems like we always force this so not sure it needs to be lazy.

cocreature (approved):
Nice, thank you!

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: True, removed.

PR #9189, opened 2021-03-18
PoC: Cache signatories and observers on exercises

cocreature:
On the extremely artifical benchmark the results look much better than I expected:

before: CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:testNew  avgt  200  16.470 ± 0.184  ms/op

After: CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:testNew  avgt  200  9.531 ± 0.149  ms/op
### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -916,7 +916,7 @@ private[lf] object SBuiltin {
         )
         .fold(err => throw DamlETransactionError(err), identity)
 
-      machine.addLocalContract(coid, templateId, createArg)
+      machine.addLocalContract(coid, templateId, createArg, sigs, obs)
       onLedger.ptx = newPtx
       checkAborted(onLedger.ptx)
       machine.returnValue = SContractId(coid)
@@ -998,7 +998,7 @@ private[lf] object SBuiltin {
       }
 
       onLedger.localContracts.get(coid) match {
-        case Some((tmplId, contract)) =>
+        case Some((tmplId, contract, _, _)) =>
           if (tmplId != templateId)
             crash(s"contract $coid ($templateId) not found from partial transaction")
           else
@@ -1010,7 +1010,7 @@ private[lf] object SBuiltin {
               templateId,
               onLedger.committers,
               cbMissing = _ => machine.tryHandleSubmitMustFail(),
-              cbPresent = { case V.ContractInst(actualTmplId, V.VersionedValue(_, arg), _) =>
+              cbPresent = { case V.ContractInst(actualTmplId, V.VersionedValue(_, arg), _, _, _) =>
                 // Note that we cannot throw in this continuation -- instead
                 // set the control appropriately which will crash the machine
                 // correctly later.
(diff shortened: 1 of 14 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1026,6 +1026,67 @@ private[lf] object SBuiltin {
...
+                // correctly later.
+                machine.ctrl =
+                  if (actualTmplId != templateId)
+                    SEDamlException(DamlEWronglyTypedContract(coid, templateId, actualTmplId))
+                  else
+                    SBStructCon(Struct.assertFromSeq(
+                      List(
+                        Name.assertFromString("arg"),
+                        Name.assertFromString("signatories"),
+                        Name.assertFromString("stakeholders")).zipWithIndex)).apply(
```
remyhaemmerle-da, line 1079: this could be computed only once.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1026,6 +1026,67 @@ private[lf] object SBuiltin {
...
+                  SEImportValue(typ, arg),
+                  SEValue(SList(signatories.get.map(SParty(_)).to(FrontStack))),
+                  SEValue(SList(stakeholders.get.map(SParty(_)).to(FrontStack))))
```
remyhaemmerle-da, line 1082: Why not caching the SList ?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -111,7 +111,7 @@ private[lf] object Speedy {
       /* Flag to trace usage of get_time builtins */
       var dependsOnTime: Boolean,
       // local contracts, that are contracts created in the current transaction)
-      var localContracts: Map[V.ContractId, (Ref.TypeConName, SValue)],
+      var localContracts: Map[V.ContractId, (Ref.TypeConName, SValue, SValue, SValue)],
```
remyhaemmerle-da, line 114: Why not caching the full node ?

remyhaemmerle-da: I was planning to move the `localContracts` Map directly in the `PatialTransaction`. Here all the data are available.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: look at `keyWithMaintainersStructFields`

cocreature: meh I’m used to compilers that do cse :slightly_smiling_face:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: We can’t expect `ContractInst` to give us an `SValue` for external contracts. We could add a cache in the engine for this but for now I mainly wanted to see how much we gain by avoiding to recompute the signatories in Daml-LF. The latter should pay off for any exercise (asuming the ledger can supply them) whereas the SValue caches will only help for multiple exercises on the same cid.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: you mean cache the create node? No good reason apart from the fact that it gets us back to having to cache the `SValue` computation separately. I mostly just hacked up the simplest thing I could come up with to see if it’s worth pursuing this further or not.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: I’ve actually gone back and used this approach even for local contracts. I don’t want to skew the benchmark to much towards exercises on local contracts since I expect those are relatively rare in practice compared to exercises on non-transient contracts.

cocreature:
It makes a less impressive but still pretty significant difference on the standard collect authority benchmark:
Before:
```
CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:test  avgt  200  20.423 ± 0.093  ms/op
```
After
```
CollectAuthority.bench  //daml-lf/scenario-interpreter/CollectAuthority.dar  CollectAuthority:test  avgt  200  16.476 ± 0.066  ms/op
```

So roughly ~1.24x speedup

cocreature:
Closing in favor of #9192 which does not require ledger changes and seems to provide most of the benefits. We could revisit this later.

PR #9191, opened 2021-03-19
Factor out an Oracle test fixture

cocreature:
We need to add Oracle support to other components like the trigger
service as well. At that point, we cannot assume that things don’t
stomp on each other so this PR adds a fixture that generates a random
user (which also comes with its own schema, creating new databases is
a bit different in oracle) which we use testing.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
+++ b/libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
@@ -10,8 +10,6 @@ import scala.util.{Random, Using}
 private[oracle] final case class User(name: String, pwd: String)
 
 trait OracleAround {
-  Class.forName("oracle.jdbc.OracleDriver")
-
   @volatile
   private var systemUser: String = _
   @volatile
@@ -34,9 +32,8 @@ trait OracleAround {
     createNewUser(u)
   }
 
-  private val pwd = "hunter2"
-
   protected def createNewUser(name: String): User = {
+    val pwd = "hunter2"
     Using.Manager { use =>
       val con = use(
         DriverManager.getConnection(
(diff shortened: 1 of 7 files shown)


libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  protected def createNewRandomUser(): User = {
+    // See https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm#i27570
+    // for name restrictions.
+    val u = "u" + Random.alphanumeric.take(29).mkString("")
+    createNewUser(u)
+  }
+
+  private val pwd = "hunter2"
```
stefanobaghino-da, line 37: `val`s in `trait`s can lead to weird issues based on the order of inheritance, I think this can be safely put in the `createNewUser` method as IIUC it's only used there.

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.testing.oracle
+
+import com.daml.ports._
+import java.sql._;
```
stefanobaghino-da, line 7:
```suggestion
import java.sql._
```

stefanobaghino-da (approved):

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
cocreature: I’m surprised scalafmt doesn’t fix this

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
cocreature: til, thx 5bbc67d903a0c298d2304bb1001e6e61fd05e3d3 Also moved away the `Class.forName` stuff into `beforeAll` which seems like a more sensible place.

PR #9192, opened 2021-03-19
Cache contract fetches in speedy

cocreature:
This PR adds a cache to speedy to cache contract fetches and
information only derived from the contract argument, namely,
signatories, observers and keys.

The cache is engine-internal so on the first fetch of a global
contract in a transaction, we recompute that information.

This does not change observable semantics:

Ledgers must be consistent within a transaction so caching is safe. We
still recompute signatories, observers & keys the first time so if
they fail, we still blow u.

We also never compute more than before. While `SBUFetch` itself did
not compute that information, it was immediately folowed by either
`SBUBeginExercise` or `SBUInsertFetchNode` which compute that
information.

We also keep the optimization that we do not have to compute key and
maintainers on by-key operations

On the collect authority benchmark, this is around a ~1.48x
improvement which is pretty decent.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -1806,6 +1806,63 @@ class EngineTest
     }
   }
 
+  "wrongly typed contract id" should {
+    val withKeyId = Identifier(basicTestsPkgId, "BasicTests:WithKey")
+    val simpleId = Identifier(basicTestsPkgId, "BasicTests:Simple")
+    val fetcherId = Identifier(basicTestsPkgId, "BasicTests:Fetcher")
+    val cid = toContractId("#BasicTests:WithKey:1")
+    val fetcherCid = toContractId("#42")
+    val fetcherInst = ContractInst(
+      fetcherId,
+      assertAsVersionedValue(
+        ValueRecord(
+          None,
+          ImmArray((None, ValueParty(alice)), (None, ValueParty(alice)), (None, ValueParty(alice))),
+        )
+      ),
+      "",
+    )
+    val contracts = defaultContracts + (fetcherCid -> fetcherInst)
+    val lookupContract = contracts.get(_)
+    val correctCommand =
+      ExerciseCommand(withKeyId, cid, "SumToK", ValueRecord(None, ImmArray((None, ValueInt64(42)))))
+    val incorrectCommand =
+      ExerciseCommand(simpleId, cid, "Hello", ValueRecord(None, ImmArray.empty))
+    val incorrectFetch =
+      ExerciseCommand(
+        fetcherId,
+        fetcherCid,
+        "DoFetch",
+        ValueRecord(None, ImmArray((None, ValueContractId(cid)))),
+      )
+
+    val now = Time.Timestamp.now()
+    val submissionSeed = hash("wrongly-typed cid")
+    def run(cmds: ImmArray[Command]) =
+      engine
+        .submit(Set(alice), Commands(cmds, now, ""), participant, submissionSeed)
+        .consume(lookupContract, lookupPackage, lookupKey)
+
+    "error on fetch" in {
+      val result = run(ImmArray(incorrectFetch))
+      inside(result) { case Left(e) =>
+        e.msg should include("wrongly typed contract id")
+      }
+    }
+    "error on exercise" in {
+      val result = run(ImmArray(incorrectCommand))
+      inside(result) { case Left(e) =>
+        e.msg should include("wrongly typed contract id")
+      }
+    }
+    "error on exercise even if used correctly before" in {
+      val result = run(ImmArray(correctCommand, incorrectCommand))
+      inside(result) { case Left(e) =>
+        e.msg should include("wrongly typed contract id")
+      }
+    }
+  }
+
   "nested transactions" should {
 
     val forkableTemplate = "BasicTests:Forkable"
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1830,7 +1830,7 @@ private[lf] object SBuiltin {
     v match {
       case SStruct(_, vals) if vals.size == cachedContractStructFields.size =>
         CachedContract(
-          templateId,
+          templateId = templateId,
           value = vals.get(cachedContractArgIdx),
           signatories = extractParties(vals.get(cachedContractSignatoriesIdx)),
           observers = extractParties(vals.get(cachedContractObserversIdx)),
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -916,7 +916,7 @@ private[lf] object SBuiltin {
         )
         .fold(err => throw DamlETransactionError(err), identity)
 
-      machine.addLocalContract(coid, templateId, createArg)
+      machine.addLocalContract(coid, templateId, createArg, sigs, obs, mbKey)
       onLedger.ptx = newPtx
       checkAborted(onLedger.ptx)
       machine.returnValue = SContractId(coid)
@@ -939,7 +939,7 @@ private[lf] object SBuiltin {
       choiceId: ChoiceName,
       consuming: Boolean,
       byKey: Boolean,
-  ) extends OnLedgerBuiltin(7) {
+  ) extends OnLedgerBuiltin(4) {
 
     override protected final def execute(
         args: util.ArrayList[SValue],
@@ -951,12 +951,13 @@ private[lf] object SBuiltin {
         case SContractId(coid) => coid
         case v => crash(s"expected contract id, got: $v")
       }
-      val sigs = extractParties(args.get(2))
-      val templateObservers = extractParties(args.get(3))
-      val ctrls = extractParties(args.get(4))
-      val choiceObservers = extractParties(args.get(5))
+      val cached = onLedger.cachedContracts.get(coid).get
+      val sigs = cached.signatories
+      val templateObservers = cached.observers
+      val ctrls = extractParties(args.get(2))
+      val choiceObservers = extractParties(args.get(3))
 
-      val mbKey = extractOptionalKeyWithMaintainers(args.get(6))
+      val mbKey = cached.key
       val auth = machine.auth
 
       onLedger.ptx = onLedger.ptx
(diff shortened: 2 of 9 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -951,12 +951,13 @@ private[lf] object SBuiltin {
         case SContractId(coid) => coid
         case v => crash(s"expected contract id, got: $v")
       }
-      val sigs = extractParties(args.get(2))
-      val templateObservers = extractParties(args.get(3))
-      val ctrls = extractParties(args.get(4))
-      val choiceObservers = extractParties(args.get(5))
+      val cached = onLedger.cachedContracts.get(coid).get
```
stefanobaghino-da, line 954:
```suggestion
      val cached = onLedger.cachedContracts(coid)
```
The `apply` method is defined on `Map`s to ultimately have the same effect as this, but the error message is more helpful:
```
scala> Map(1 -> "one")(2)
java.util.NoSuchElementException: key not found: 2
  at scala.collection.immutable.Map$Map1.apply(Map.scala:245)
  ... 32 elided

scala> Map(1 -> "one").get(2).get
java.util.NoSuchElementException: None.get
  at scala.None$.get(Option.scala:627)
  at scala.None$.get(Option.scala:626)
  ... 32 elided
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -114,6 +121,7 @@ private[lf] object Speedy {
       var localContracts: Map[V.ContractId, (Ref.TypeConName, SValue)],
       // global contract discriminators, that are discriminators from contract created in previous transactions
       var globalDiscriminators: Set[crypto.Hash],
+      var cachedContracts: Map[V.ContractId, CachedContract],
```
stefanobaghino-da, line 124: What is the lifecycle of this cache? IIUC it prevents fetching the same contract more than once while evaluating a single transaction. If I'm wrong and it survives across evaluations, would it make sense (and be feasible) to apply some eviction policy to prevent this from blowing off and causing OOMs?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: hah, I actually deliberately didn’t use `apply` to remind me that I need to emit a proper error message here since I find `.get` much easier to spot than an `apply` somewhere. Obviously I still forgot, will fix on Monday :slightly_smiling_face:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: It never survives a transaction (which applies to all fields in here, speedy is pretty much only alive for one transaction with the exception of scenarios and there we reset it after each submit) and the size of the cache is at most proportional to the size of the transaction we’re building (each cache entry corresponds to a fetch or an exercise). The values in the cache are also mostly shared with the actual transaction node so we don’t keep signatories twice in memory or something like that.

cocreature: In other words, given that it doesn’t grow more than the transaction and the overhead to that seems quite small given that most values are shared, I wouldn’t add an eviction policy but I’m more than happy to be proven wrong here and add one.

stefanobaghino-da: Nah, if this is only for a transaction, a simple map works. Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: 2fbd75434cb8198a46429a65d02a7aa8639e6281

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1004,24 +1008,47 @@ private[lf] object SBuiltin {
...
-                // correctly later.
-                machine.ctrl =
-                  if (actualTmplId != templateId)
-                    SEDamlException(DamlEWronglyTypedContract(coid, templateId, actualTmplId))
-                  else
-                    SEImportValue(typ, arg)
-              },
-            )
-          )
+          onLedger.cachedContracts.get(coid) match {
```
remyhaemmerle-da, line 1011: We need to check the that coid has the proper type.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -98,6 +98,13 @@ private[lf] object Speedy {
 
   private[lf] sealed trait LedgerMode
 
+  private[lf] final case class CachedContract(
+      value: SValue,
+      signatories: Set[Party],
+      observers: Set[Party],
+      key: Option[Node.KeyWithMaintainers[V[Nothing]]],
+  )
```
remyhaemmerle-da, line 106:
```suggestion
  private[lf] final case class CachedContract(
      templateId: Ref.TypeConName,
      value: SValue,
      signatories: Set[Party],
      observers: Set[Party],
      key: Option[Node.KeyWithMaintainers[V[Nothing]]],
  )
```

You need the type to check a run time.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -329,14 +337,25 @@ private[lf] object Speedy {
...
+        observers: Set[Party],
+        key: Option[Node.KeyWithMaintainers[V[Nothing]]],
+    ) =
       withOnLedger("addLocalContract") { onLedger =>
         coid match {
           case V.ContractId.V1(discriminator, _)
               if onLedger.globalDiscriminators.contains(discriminator) =>
             crash("Conflicting discriminators between a global and local contract ID.")
           case _ =>
             onLedger.localContracts = onLedger.localContracts.updated(coid, templateId -> SValue)
```
remyhaemmerle-da, line 354:
```suggestion
            onLedger.localContracts = (onLedger.localContracts + coid)
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -329,14 +337,25 @@ private[lf] object Speedy {
...
+            onLedger.cachedContracts = onLedger.cachedContracts.updated(
+              coid,
+              CachedContract(SValue, signatories, observers, key),
```
remyhaemmerle-da, line 357:
```suggestion
              CachedContract(templateId, SValue, signatories, observers, key),
```

remyhaemmerle-da (commented):
Looks good in general. 
We are missing however that cached contracts are well type. 
This need to be fixed and a test have to be add. 

Less important I think we change the signature of local Contracts to `Set[ContractId]` and factorize the lookup of local contract and global cached contract.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: Added in 5d3a7eb2274f0cebbc499e768391ae94113502c0 and added a test in cee52077c3a510f6e4e713bfd5af0db284363a87 which fails without this change.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: As discussed, let’s leave refactoring of `localContracts` to another PR.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/WronglyTypedContractIdIT.scala
```diff
@@ -41,4 +41,26 @@ final class WronglyTypedContractIdIT extends LedgerTestSuite {
         assertGrpcError(fetchFailure, Code.INVALID_ARGUMENT, "wrongly typed contract id")
       }
   })
+
+  test(
+    "WTMultipleExerciseFails",
+    "Exercising on a wrong type fails after correct exercise in same transaction",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
```
remyhaemmerle-da, line 49: 1. I definitively would like a Unit test in daml-lf module. I think there is everything you need in `EngineTest` (this is mess I know :disappointed: )
2. I am inclined to not test at all this engine specific behavior  in the conformance test.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1769,6 +1805,40 @@ private[lf] object SBuiltin {
...
+    cachedContractStructFields.indexOf(Ast.observersFieldName)
+
+  private[speedy] def extractCachedContract(
+      templateId: Ref.TypeConName,
+      v: SValue,
+  ): CachedContract =
+    v match {
+      case SStruct(_, vals) if vals.size == cachedContractStructFields.size =>
+        CachedContract(
+          templateId,
```
remyhaemmerle-da, line 1833:
```suggestion
          templateId = templateId,
```
very cosmetic

remyhaemmerle-da (approved):
LGTM.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/WronglyTypedContractIdIT.scala
cocreature: I’ve looked at `EngineTest` but it doesn’t have any tests for wrongly typed contract ids atm afaict and it also has no tests that run across transactions in general :disappointed: I’ll see if I can figure something out.

Given that we already have tests for wrongly typed contract ids, I think it’s reasonable to keep this one.

cocreature: 220e52fb3b287fd25f626355bc15755008ca9467

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -1806,6 +1806,63 @@ class EngineTest
     }
   }
 
+  "wrongly typed contract id" should {
```
remyhaemmerle-da, line 1809: :bow:

PR #9197, opened 2021-03-20
Update to Flyway 6.5.7

dasormeter:
Flyway 7 looks to have some nice added configurability.

One particular thing that jumps out is the configurable retry count whenever there is a lock between different tests/setups trying to run the migration scripts in parallel.
https://flywaydb.org/documentation/configuration/parameters/lockRetryCount

Currently with v6 this looks to be 50s with no configurability (and we are seeing the retry limit exception being hit with Postgres on some tests)


maven_install_2.12.json
```diff
@@ -1,6 +1,6 @@
 {
     "dependency_tree": {
-        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": 218875396,
+        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": 1207618494,
```
cocreature, line 3: You also need to repin the Scala 2.13 file.

bazel-java-deps.bzl
```diff
@@ -136,7 +136,7 @@ def install_java_deps():
             "org.apache.commons:commons-text:1.4",
             "org.awaitility:awaitility:3.1.6",
             "org.checkerframework:checker:2.5.4",
-            "org.flywaydb:flyway-core:6.5.0",
+            "org.flywaydb:flyway-core:6.5.7",
```
cocreature, line 139: Is there anything useful in there? Your PR description presumably refers to the changes in 7.0? Not opposed to upgrading at all, quite the opposite, just curious.

dasormeter: Just WIP for now -- I was trying a less intrusive minor version upgrade as second order and that looks to be fine

What we really want is the 7.x changes, but while those seem to run locally and even across a test suite they fail miserably when run on CI, presumably because of some classpath issues and maybe some race conditions?

Anyway this is still a WIP pr -- I need to do some more work on it to get things to mergeable state -- sorry for prematurely putting you as a reviewer (thought I had removed after seeing the 7 build fail spectacularly)

dasormeter:
Closing for now -- this was more of an experiment -- upgrade to next major version of flyway has some more significant breaking api changes to deal with than expected

PR #9207, opened 2021-03-23
Remove contract arguments from localContracts

cocreature:
We already store them in cachedContracts, no reason to store them
twice. This should also be slightly more efficient since we only do
one lookup in cachedContracts in the common case and only look at
localContracts if there’s a wrongly typed contract. I doubt it’s
measurable but at least it’s not worse.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -341,7 +341,7 @@ private[lf] object Speedy {
     def addLocalContract(
         coid: V.ContractId,
         templateId: Ref.TypeConName,
-        SValue: SValue,
+        arg: SValue,
```
remyhaemmerle-da, line 344: :+1:

remyhaemmerle-da (approved):
Thanks for the cleanup.

PR #9208, opened 2021-03-23
Support passing sandbox port via --sandbox-option --port

cocreature:
--sandbox-port has caused confusion several times and cannot be
specified in daml.yaml which is a bit annoying. There isn’t really any
good reason not to support --sandbox-option --port since we can pick
up the port via the port file either way.

This change is fully backwards compatible.

changelog_begin

- [Daml Assistant] The sandbox port can now also be configured via
  `--sandbox-option=--port=12345` instead of `--sandbox-port`. Other
  tools like Navigator, the JSON API and Daml Script will pick up the
  modified port automatically.

changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Thank you!

garyverhaegen-da (approved):

PR #9216, opened 2021-03-23
Improve syntax for selecting contract ids from a transaction tree

### LINKED ISSUE #9083: Improve syntax for selecting contract ids from a transaction tree
Currently this looks a bit crude and is fairly hard to understand.. Not quite sure what the endresult should be. One possibility would be to include choice names or something like this.

aherrmann-da:
Closes #9083 

Implements more readable transaction tree accessors. These use the template id and choice name to select a path in a transaction tree instead of plain numerical indices. E.g.

```
let contract_1_0 = fromTree tree $
      exercised @Module.SomeTemplate "Choice" $
      created @Module.OtherTemplate
```

instead of

```
let contract_1_0 = createdCid @Module.OtherTemplate [0, 0]
```

In case multiple events at a given level in the transaction tree match the given template id (and choice name) the generated Daml script dump will use the forms `createdN`/`exercisedN` to index into the list of matching events.

This adapts the test cases and makes sure that all of `created`, `createdN`, `exercised`, and `exercisedN` are covered.

This removes the now unused `createdCid` accessors.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/daml/Daml/Script.daml
+++ b/daml-script/daml/Daml/Script.daml
@@ -10,7 +10,6 @@ module Daml.Script
   , submitMultiMustFail
   , submitTree
   , submitTreeMulti
-  , createdCid
   , query
   , queryContractId
   , queryContractKey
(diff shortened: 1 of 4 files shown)


daml-script/daml/Daml/Script.daml
```diff
@@ -566,6 +543,81 @@ instance Show Exercised where
...
+    offset : Int
+    child : TreeIndex t
+
+-- | HIDE This is an early access feature.
+fromTree : Template t => TransactionTree -> TreeIndex t -> ContractId t
+fromTree tree index = fromTreeGo index tree.rootEvents
+
+fromTreeGo : Template t => TreeIndex t -> [TreeEvent] -> ContractId t
+fromTreeGo (CreatedIndex index) events =
+  case mapOptional fromCreated events of
```
cocreature, line 571: Probably perfectly fine for this PR but having just spent some time on optimizations: This type of code is usually much faster in Daml if you hand-write the selector since there is no list fusion so you pay for the full `mapOptional` and the allocation of the list + the second traversal for the index.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
```diff
@@ -50,9 +59,29 @@ object TreeUtils {
     }
   }
 
+  private def layerSelectors(events: Seq[TreeEvent.Kind]): Seq[Selector] = {
+    val created = mutable.HashMap.empty[TemplateId, Int].withDefaultValue(0)
+    val exercised = mutable.HashMap.empty[(TemplateId, Choice), Int].withDefaultValue(0)
+    events.collect {
```
cocreature, line 65: You might want to error out on `Kind.Empty` here instead of using `collect`. If you ever actually get this (which could happen if we expand the ledger API), the `zip` below will blow up since you have fewer selectors than root events.

cocreature (approved):
Nice! I like it.

daml-script/dump/src/main/scala/com/daml/script/dump/TreeUtils.scala
aherrmann-da: Good point. I've avoided the issue by pulling the iteration into `layerSelectors`, renaming it to `foreachEventWithSelector`. This way we can just skip `Kind.Empty` without having to throw an error or providing an empty selector of some sort.

PR #9217, opened 2021-03-23
More manual list fusion in stdlib

cocreature:
Both of these seem sufficiently common that it seems worth optimizing
them.

I realize I’m switching somewhat randomly between foldr and
hand-written recursion. Either should be better than traversing the
list twice but I think it’s a clear sign that we need some benchmarks
to establish the benefits or downsides of one over the other properly.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Thanks! Maybe worth doing this for `concat` too?

cocreature:
> Thanks! Maybe worth doing this for `concat` too?

concat at least only traverses the list once. There is some overhead to jump through the typeclass indirection but I think we only pay this once and then cache it so I’m less worried about this for now.

PR #9219, opened 2021-03-23
Make non-repudiation integration exclusive to enterprise edition

### LINKED ISSUE #9182: Non-repudiation: make it exclusive to enterprise edition
The non-repudiation middleware and the related components are enterprise edition exclusive. Make sure that the non-repudiation CLI options are available only to enterprise edition users.

stefanobaghino-da:
Closes #9182

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/it/edition/ce/com/daml/http/NonRepudiationTest.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import akka.http.scaladsl.model.StatusCodes
+import com.daml.nonrepudiation.CommandIdString
+
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+final class NonRepudiationTest extends AbstractNonRepudiationTest {
+
+  import HttpServiceTestFixture._
+
+  "fail " in withSetup { (db, uri, encoder) =>
```
stefanobaghino-da, line 16: Just noticed the message here is incomplete. Working on it.

ledger-service/http-json/src/it/edition/ce/com/daml/http/NonRepudiationTest.scala
```diff
@@ -0,0 +1,31 @@
...
+    val expectedParty = "Alice"
+    val expectedNumber = "abc123"
+    val expectedCommandId = UUID.randomUUID.toString
+    val meta = Some(domain.CommandMeta(commandId = Some(domain.CommandId(expectedCommandId))))
+    val domainParty = domain.Party(expectedParty)
+    val command = accountCreateCommand(domainParty, expectedNumber).copy(meta = meta)
+    postCreateCommand(command, encoder, uri)
+      .flatMap { case (status, _) =>
+        status shouldBe StatusCodes.InternalServerError
```
stefanobaghino-da, line 25: This is not what we actually want to happen, I'd say 403 (Forbidden) makes more sense. Working on this.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
```diff
@@ -0,0 +1,169 @@
...
+import scopt.{Read, RenderingMode}
+
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.util.Try
+
+object OptionParser {
+
+  private val logger = LoggerFactory.getLogger(classOf[OptionParser])
+
+}
```
stefanobaghino-da, line 19: I did this on a rush, rather than using raw Slf4j we probably want to use the same approach as in the rest of the HTTP JSON codebase and use Scala Logging instead.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -106,7 +106,12 @@ private[http] final case class JdbcConfig(
 )
 
 private[http] object JdbcConfig extends ConfigCompanion[JdbcConfig]("JdbcConfig") {
-  import dbbackend.ContractDao.supportedJdbcDriverNames
+
+  // FIXME Duplicated to not introduce a circular dependency between ContractDao and this
```
stefanobaghino-da, line 110: Should I create a separate target for this that both `ContractDao` and this can depend on? Otherwise I'm not 100% of how to fix this.

stefanobaghino-da:
I have a CI failure that I cannot reproduce locally. I've never seen this error. Any suggestion? Going back and forth with CI could lead to terribly slow feedback loop to fix this.

```
> bazel build //ledger-service/http-json-oracle/...
INFO: Invocation ID: 6284d51a-8cad-45f3-ab92-4e6859935354
INFO: Analyzed 2 targets (1 packages loaded, 1004 targets configured).
INFO: Found 2 targets...
INFO: Elapsed time: 3.427s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
> bazel build //ledger-service/http-json-oracle:integration-tests
INFO: Invocation ID: 5a11d58b-a62f-4315-a45f-03600b92e689
ERROR: Target //ledger-service/http-json-oracle:integration-tests is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //ledger-service/http-json-oracle:integration-tests   <-- target platform didn't satisfy constraint @platforms//:incompatible
INFO: Elapsed time: 0.804s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

stefanobaghino-da (commented):
Offline comment by @S11001001: add aliases for `*-ce` targets.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
cocreature: I don’t quite understand the issue. What is stopping us from either moving this to `ContractDao` and making the CLI depend on that or the other way around? That said, splitting things into separate targets is always a reasonable option imho.

ledger-service/http-json-perf/BUILD.bazel
```diff
@@ -13,96 +13,109 @@ scalacopts = lf_scalacopts + [
...
-        "@maven//:com_fasterxml_jackson_core_jackson_databind",
-        "@maven//:io_gatling_gatling_app",
-        "@maven//:io_gatling_gatling_commons",
-        "@maven//:io_gatling_gatling_core",
-        "@maven//:io_gatling_gatling_http",
-        "@maven//:io_gatling_gatling_http_client",
-        "@maven//:org_slf4j_slf4j_api",
-    ],
-) if scala_major_version == "2.12" else None
+[
```
cocreature, line 16: I think our daily benchmarks run this. Probably need an updated target name.

ledger-service/http-json/src/it/edition/ce/com/daml/http/NonRepudiationTest.scala
cocreature: I think InternalServerError is fine for now. There is a broader topic here on the JSON API reporting any failure coming from the ledger as 500 but I wouldn’t address this specifically for non-repudiation.

cocreature (commented):
Looks very sensible, thanks for pushing this through!

I slightly wonder if there might be a way to factor things in a way that requires less duplication of targets but I don’t have a concrete proposal so fine with this approach.

cocreature:
> I have a CI failure that I cannot reproduce locally. I've never seen this error. Any suggestion? Going back and forth with CI could lead to terribly slow feedback loop to fix this.

I don’t see the failure on CI. Can you point me to a log? The error probably comes from somewhere around https://github.com/digital-asset/daml/blob/e492603808fa2c9068f58bd60dbc2d01cb2044ac/ledger-service/http-json-oracle/BUILD.bazel#L18. We disable the oracle targets by default to make sure that something like `bazel test //...` still works for users that don’t have Oracle installed. That said, I don’t see anything in your PR that would change how this works.

ledger-service/http-json/src/it/edition/ce/com/daml/http/NonRepudiationTest.scala
stefanobaghino-da: Ok, thanks.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
stefanobaghino-da: Because `ContractDao` is in `http-json` and if we make `http-json-cli` depend on `http-json` the dependency becomes circular. Or am I missing something?

cocreature: What I was trying to ask is, why can’t ContractDao use the definition here given that it already depends on it?

stefanobaghino-da: Because `ContractDao` defines which are the supported drivers, which I think makes a lot of sense. I don't think we should put this in the CLI.

cocreature: Makes sense, then either a separate target just for this or a separate target for all of `ContractDao` to break the circular dependency seems sensible.

stefanobaghino-da: This is going to be quite a large change due to some dependencies between various packages that are requiring me to break more and more packages apart. I would recommend we accept a one line code duplication for the time being while I break apart the modules into a separate PR.

stefanobaghino-da: Tracked in https://github.com/digital-asset/daml/issues/9229

cocreature: perfectly fine

ledger-service/http-json/src/it/edition/ce/com/daml/http/NonRepudiationTest.scala
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/9219/commits/703142d146a372895fe560f757f1d8f697edc0f5

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
stefanobaghino-da: Address https://github.com/digital-asset/daml/pull/9219/commits/25a62a77e6b4e11ceaa4870aa3a134430f8348c7

ledger-service/http-json-perf/BUILD.bazel
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/9219/commits/9eeec4193bf92fcc0c2b5dfb56c4970b6de0d11a

stefanobaghino-da:
> > I have a CI failure that I cannot reproduce locally. I've never seen this error. Any suggestion? Going back and forth with CI could lead to terribly slow feedback loop to fix this.
> 
> I don’t see the failure on CI. Can you point me to a log? The error probably comes from somewhere around
> 
> https://github.com/digital-asset/daml/blob/e492603808fa2c9068f58bd60dbc2d01cb2044ac/ledger-service/http-json-oracle/BUILD.bazel#L18
> 
> . We disable the oracle targets by default to make sure that something like `bazel test //...` still works for users that don’t have Oracle installed. That said, I don’t see anything in your PR that would change how this works.

@cocreature https://dev.azure.com/digitalasset/daml/_build/results?buildId=72836&view=logs&j=52c0dd40-3465-5e8e-2f7c-ae591b3bddfa&t=d7d170c8-bc3c-5788-66a8-cddc87f82282

cocreature:
> @cocreature https://dev.azure.com/digitalasset/daml/_build/results?buildId=72836&view=logs&j=52c0dd40-3465-5e8e-2f7c-ae591b3bddfa&t=d7d170c8-bc3c-5788-66a8-cddc87f82282

That just fails with
```
ledger-service/http-json-oracle/src/it/scala/http/HttpServiceOracleInt.scala:15: error: not found: type JdbcConfig
```
and a bunch of other errors of this type not the error you described?

stefanobaghino-da:
> > @cocreature https://dev.azure.com/digitalasset/daml/_build/results?buildId=72836&view=logs&j=52c0dd40-3465-5e8e-2f7c-ae591b3bddfa&t=d7d170c8-bc3c-5788-66a8-cddc87f82282
> 
> That just fails with
> 
> ```
> ledger-service/http-json-oracle/src/it/scala/http/HttpServiceOracleInt.scala:15: error: not found: type JdbcConfig
> ```
> 
> and a bunch of other errors of this type not the error you described?

Sorry for not being clear.

What I'm trying to say is that I cannot build the target locally because of the error I've shown.

I would like to reproduce the same error I see on CI locally but apparently there's something that prevents `bazel build //ledger-service/http-json-oracle:integration-tests` from running.

How can I work around it?

cocreature:
Oh in that case just add `oracle_testing=yes` like we do on CI https://github.com/digital-asset/daml/blob/4fd1cf7c28a3576967f241623154a962bb5be7a6/ci/build.yml#L236. If you want to run tests locally and not just build you also have to start the docker container, again the CI config should be a useful reference for this.

stefanobaghino-da:
> Oh in that case just add `oracle_testing=yes` like we do on CI
> 
> https://github.com/digital-asset/daml/blob/4fd1cf7c28a3576967f241623154a962bb5be7a6/ci/build.yml#L236
> 
> . If you want to run tests locally and not just build you also have to start the docker container, again the CI config should be a useful reference for this.

Cool, thanks.

stefanobaghino-da:
@cocreature Do you prefer to wait for @S11001001's review before merging?

cocreature (approved):
Nice, thank you!

cocreature:
I’ll leave it up to @S11001001 if he wants to block merging or not in this case. I usually tend to lean on the side of merging quickly but treating comments made after merging at the same level as once made before merging.

stefanobaghino-da:
Ok, I'll merge and address any open point at a later time, if there's any.

PR #9224, opened 2021-03-24
Add Ord instances for Any(Template|Choice|ContractKey)

cocreature:
This is useful to use them as map keys. We have a large application
that currently does a linear search over a potentially very large set
of contracts which can be replaced with a map lookup with this.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
@@ -224,45 +224,23 @@ instance IsParties (Optional Party) where
 
 #ifdef DAML_GENERIC_COMPARISON && DAML_ANY_TYPE
 
--- We do not have a general Eq instance for Any since Any is not exposed
--- to users directly and that instance will crash on some values but
--- we can have total Eq instances for AnyTemplate, AnyChoice and AnyContractKey
--- since those are guaranteed to be serializable.
-eqAny : Any -> Any -> Bool
-eqAny = primitive @"BEEqual"
-
-ltAny : Any -> Any -> Bool
-ltAny = primitive @"BELess"
-
-lteAnyTemplate : AnyTemplate -> AnyTemplate -> Bool
-lteAnyTemplate = primitive @"BELessEq"
-
-lteAnyChoice : AnyChoice -> AnyChoice -> Bool
-lteAnyChoice = primitive @"BELessEq"
-
-lteAnyContractKey : AnyContractKey -> AnyContractKey -> Bool
-lteAnyContractKey = primitive @"BELessEq"
-
 instance Eq AnyTemplate where
-  AnyTemplate a == AnyTemplate b = eqAny a b
+  (==) = primitive @"BEEqual"
 
 instance Eq AnyChoice where
-  AnyChoice aAny aRep == AnyChoice bAny bRep =
-    eqAny aAny bAny && aRep == bRep
+  (==) = primitive @"BEEqual"
 
 instance Eq AnyContractKey where
-  AnyContractKey aAny aRep == AnyContractKey bAny bRep =
-    eqAny aAny bAny && aRep == bRep
-
+  (==) = primitive @"BEEqual"
 
 instance Ord AnyTemplate where
-  (<=) = lteAnyTemplate
+  (<=) = primitive @"BELessEq"
 
 instance Ord AnyChoice where
-  (<=) = lteAnyChoice
+  (<=) = primitive @"BELessEq"
 
 instance Ord AnyContractKey where
-  (<=) = lteAnyContractKey
+  (<=) = primitive @"BELessEq"
 #endif
 
 #ifdef DAML_TYPE_REP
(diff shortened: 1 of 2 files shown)


compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
```diff
@@ -242,6 +248,21 @@ instance Eq AnyContractKey where
   AnyContractKey aAny aRep == AnyContractKey bAny bRep =
     eqAny aAny bAny && aRep == bRep
 
+
+instance Ord AnyTemplate where
+  AnyTemplate a <= AnyTemplate b = lteAny a b
```
sofiafaro-da, line 253: Can we just use `primitive @"BELessEq"` here and for all the other `Ord` instances? If the goal is to make use of these in maps, it makes a lot more sense to use the built-in order for all of them.

```suggestion
  (<=) = primitive @"BELessEq"
```

cocreature: I’m a bit worried that sorting by value before we sort by template id is going to cause a fair amount of confusion. We can’t change the order of those since they are in stable types.

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
```diff
@@ -242,6 +248,21 @@ instance Eq AnyContractKey where
...
+
+instance Ord AnyChoice where
+  (AnyChoice aAny aRep) <= (AnyChoice bAny bRep)
+    | aRep < bRep = True
+    | aRep == bRep = lteAny aAny bAny
+    | otherwise = False
```
sofiafaro-da, line 259: Same here. This is a different order (the choice any is compared after the template type rep) but it makes more sense to do this if we're going to use these in maps.

```suggestion
  (<=) = primitive @"BELessEq"
```

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
cocreature: hm fair point, if we get the same in map, we probably should do that.

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
```diff
@@ -242,6 +248,21 @@ instance Eq AnyContractKey where
...
+
+instance Ord AnyContractKey where
+  (AnyContractKey aAny aRep) <= (AnyContractKey bAny bRep)
+    | aRep < bRep = True
+    | aRep == bRep = lteAny aAny bAny
+    | otherwise = False
```
sofiafaro-da, line 265: Same here.

```suggestion
  (<=) = primitive @"BELessEq"
```

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
cocreature: fixed in 64cd504949da6ca0d38ce46ee97ab78d92bf2b30, what you suggested fails in the lf conversion due to the way we use those primitives but fixed it slightly differently.

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
cocreature: 64cd504949da6ca0d38ce46ee97ab78d92bf2b30

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
cocreature: 64cd504949da6ca0d38ce46ee97ab78d92bf2b30

compiler/damlc/daml-stdlib-src/DA/Internal/Template/Functions.daml
sofiafaro-da: It fails? That's unexpected.

cocreature:
```
error type checking value DA.Internal.Template.Functions.$c<=:
|  unknown value:   
|  518032f41fd0175461b35ae0c9691e08b4aea55e62915f8360af2cc7a1f2ba6c:GHC.Types:primitive
```
Haven’t looked too closely at it, I expect it probably ends up looking slightly different in typeclass methods which breaks the way we match on it.

sofiafaro-da: e.g. this is the `Ord Int` instance in the stdlib. What's different?

image

cocreature: duh, I forgot the `@` in front of the string. Fixed in 7eca99ddf95c8dee21c92cd66566086326ac97e2

sofiafaro-da (approved):
Thanks!

PR #9226, opened 2021-03-24
Add DA.List.BuiltinOrder module

cocreature:
This provides variants of `dedup*` and `sort*` which rely on Daml-LF’s
builtin ordering (using Map internally). I don’t have microbenchmarks
but even in macrobenchmarks this is a measurable improvement which
isn’t particularly surprising.

changelog_begin

- [Daml Stdlib] Add `DA.List.BuiltinOrder` module. This module provides
  variants of `sort*` and `dedup*` which rely on Daml-LF’s builtin
  ordering and are significantly more efficient in some cases.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/List/BuiltinOrder.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/List/BuiltinOrder.daml
@@ -1,0 +1,125 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+
+#ifndef DAML_GENMAP
+module DA.List.BuiltinOrder where
+#else
+
+-- | Note: This is only supported in DAML-LF 1.11 or later.
+--
+-- This module provides variants of other standard library
+-- functions that are based on the builtin Daml-LF ordering rather
+-- than user-defined ordering. This is the same order also used
+-- by `DA.Map`.
+--
+-- These functions are usually much more efficient than their
+-- `Ord`-based counterparts.
+--
+-- Note that the functions in this module still require `Ord`
+-- constraints. This is purely to enforce that you don’t
+-- pass in values that cannot be compared, e.g., functions. The
+-- implementation of those instances is not used.
+module DA.List.BuiltinOrder
+  ( dedup
+  , dedupOn
+  , dedupSort
+  , dedupOnSort
+  , sort
+  , sortOn
+  ) where
+
+-- import DA.Map (Map)
+import qualified DA.Map as Map
+
+-- | `dedup l` removes duplicate elements from a list. In particular,
+-- it keeps only the first occurence of each element.
+--
+-- `dedup` is stable so the elements in the output are ordered
+-- by their first occurence in the input. If you do not need
+-- stability, consider using `dedupSort` which is more efficient.
+--
+-- ```
+-- >>> dedup [3, 1, 1, 3]
+-- [3, 1]
+-- ```
+dedup : Ord a => [a] -> [a]
+dedup = dedupOn identity
+
+-- | A version of `dedup` where deduplication is done
+-- after applying the given function. Example use: `dedupOn (.employeeNo) employees`.
+--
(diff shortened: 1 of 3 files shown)


compiler/damlc/daml-stdlib-src/DA/Ord/Builtin.daml
```diff
@@ -0,0 +1,125 @@
...
+-- [1, 3]
+-- ```
+dedupSort : Ord a => [a] -> [a]
+dedupSort = dedupOnSort identity
+
+-- | `dedupOnSort` is a more efficient variant of `dedupOn`
+-- that does not preserve the order of the input elements.
+-- Instead the output will be sorted on the values returned by the function.
+--
+-- For duplicates, the last element in the list will be included in the output.
```
cocreature, line 88: I’m slightly unhappy about this. It seems nicer if `dedupOnSort` wouldn’t also change whether we preserve the first or the last element. We can easily change that but it means we pay for an extra lookup before the insert. I don’t have particularly strong feelings here either way and I don’t have the microbenchmarks to classify the impact so happy to change it either way.

sofiafaro-da: You could use a foldr to make it preserve the first element instead of the last. (I imagine foldr vs foldl is about the same in terms of impact, compared to doing a lookup.)

compiler/damlc/daml-stdlib-src/DA/Ord/Builtin.daml
```diff
@@ -0,0 +1,125 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+{-# LANGUAGE CPP #-}
+
+#ifndef DAML_GENMAP
+module DA.Ord.Builtin where
```
sofiafaro-da, line 6: Since these are all List functions, how about `DA.List.Fast` or `DA.List.Dedup` or `DA.List.Sort` or something?

compiler/damlc/daml-stdlib-src/DA/Ord/Builtin.daml
cocreature: duh, fixed in 1214abe31ce3c28fbcd4d94054ed66aa2a6ae7bd

sofiafaro-da (approved):
LGTM, thank you :-)

compiler/damlc/daml-stdlib-src/DA/Ord/Builtin.daml
cocreature: I think I don’t really like `Fast` in a module name. `Dedup` or `Sort` both work but also only cover half of it. Something like `DA.List.BuiltinOrd` could work. Not sure if that’s better or worse.

sofiafaro-da: Ok. I personally don't like `DA.Ord.Builtin` but this isn't important.

cocreature: Do you have another suggestion that somehow combines lists with the fact that the difference is the builtin ordering? Or are you opposed to mentioning the builtin order in the module name at all?

sofiafaro-da: How about `DA.List.Order`? As in, both "dedup" and "sort" are imposing some kind of "order" on a list.

cocreature: Not opposed to it but that doesn’t really explain why there is `dedup` in `DA.List` as well. Maybe we can rely on the docs for that but it seemed nice to have that very explicit in the module name.

sofiafaro-da: Ok. `DA.List.BuiltinOrder` (or `DA.List.BuiltinOrd`) is fine too. I do think it's better than `DA.Ord.Builtin`.

cocreature: Changed to `DA.List.BuiltinOrder`, agree that it’s better

PR #9228, opened 2021-03-24
use coreutils `date` on macos

garyverhaegen-da:
macOS uses BSD date by default which has slightly different options.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #9234, opened 2021-03-24
fix slack notifs

garyverhaegen-da:
Looks like I screwed up in #9166. I'm not really sure whether it's better to enable piecemeal as I'm doing here or just put the "always run" flag back.

CHANGELOG_BEGIN
CHANGELOG_END


garyverhaegen-da:
... and of course I managed to break the entire YAML parser with this. Oh well. That resolves the question, at least: I'll go with enabling all the time as that's the only thing we _can_ do, apparently.

cocreature (approved):

PR #9245, opened 2021-03-25
macos cache cleaning 

garyverhaegen-da:
This is adapting the same approach as #9137 to the macOS machines. The setup is very similar, except macOS apparently doesn't require any kind of `sudo` access in the process.

The main reason for the chance here is that while `~/.bazel-cache` is reasonably fast to clean, cleaning just that has finally caught up to us with a recent cleanup step that proudly claimed:

```
before: 638Mi free
after: 1.2Gi free
```

So we do need to start cleaning the other one after all.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/clean-up.yml
+++ b/ci/clean-up.yml
@@ -17,21 +17,8 @@ steps:
     df -h .
     if [ $(df -m . | sed 1d | awk '{print $4}') -lt 50000 ]; then
         echo "Disk full, cleaning up..."
-        case $(uname) in
-        Linux)
-            $HOME/reset_caches.sh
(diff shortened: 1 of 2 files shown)


cocreature (approved):
lgtm, thank you!

garyverhaegen-da:
@cocreature @nycnewman I'll be deploying that over the weekend.

nycnewman:
Ack (and seeing whether this retriggers CLA)

garyverhaegen-da:
Change deployed.

PR #9246, opened 2021-03-26
DPP-313 Append only schema migration

rautenrieth-da:
## Index database schema

This PR adds a first version of the new append-only index database schema.

The corresponding flyway migration is experimental. It is protected by the following measures:
- It is disabled by default
- Participants based on the kvutils runner app can enable it through a CLI flag. The flag is hidden and clearly marked as unsafe.
- The SQL migration code will fail if the index database contains any data.

Because of the above, the new migration should not affect anything.

## Postgres version

This PR also bumps the PostgreSQL version used in the SDK from 9.x to 12. A bump to at least 10 is required because hash indices were unsafe prior to that version. A bump to 12 is required for table partitioning used in the new migration, which could be removed if necessary.


rautenrieth-da:
@nmarton-da The migration script itself is based on your PoC #8990. I have deleted the comments showing the old schema, and added a short statement that will fail the migration if there is any data.

@gerolf-da Note the version bump of Postgres

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -362,6 +364,14 @@ object Config {
             "Enable the development version of the Daml-LF language. Highly unstable. Should not be used in production."
           )
 
+        opt[Unit]("index-append-only-schema-unsafe")
+          .optional()
+          .hidden()
+          .text(
+            s"Enable the use of the experimental append-only index database schema. Highly unstable. Should not be used in production."
```
miklos-da, line 371: Nit: this may be shortened as
```suggestion
            s"Use the experimental append-only index database schema. Highly unstable. Should not be used in production."
```
or
```suggestion
            s"Enable the experimental append-only index database schema. Highly unstable. Should not be used in production."
```

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -32,14 +36,16 @@ private[platform] class FlywayMigrations(jdbcUrl: String)(implicit loggingContex
...
+      allowExistingSchema: Boolean = false,
+      enableExperimentalSchema: Boolean = false,
   )(implicit resourceContext: ResourceContext): Future[Unit] =
     dataSource.use { ds =>
       Future.successful {
         val flyway = configurationBase(dbType)
           .dataSource(ds)
           .baselineOnMigrate(allowExistingSchema)
           .baselineVersion(MigrationVersion.fromVersion("0"))
+          .target(if (enableExperimentalSchema) "latest" else lastStableSchemaVersion)
```
miklos-da, line 48: We are assuming we won't need to add a new iteration to the 'stable' schema, right?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -39,6 +39,7 @@ final case class Config[Extra](
     metricsReportingInterval: Duration,
     trackerRetentionPeriod: FiniteDuration,
     engineMode: EngineMode,
+    enableExperimentalSchemaForIndex: Boolean,
```
miklos-da, line 42: Is there a reason for calling this 'experimental schema' instead of 'append-only schema'? If we only intend to have additional 'experimental' schema changes on top of the append-only schema then let's call this `enableAppendOnlySchema` or similar, so that it is not confusing (i.e., 'experimental' and 'append-only' refers to the same concept in the code).

miklos-da (approved):
I don't have context to review the migration script itself.

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
gerolf-da: Adding a new migration for the generic index is still possible:
1. current last stable schema version is 49
2. increase versions for existing experimental schema migrations by 1
3. add migration for generic index schema as version 50
4. set `lastStableSchemaVersion` to 50

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -0,0 +1,172 @@
...
+
+    -- * create events only
+    create_signatories text[],
+    create_observers text[],
+    create_agreement_text text,
+    create_key_value bytea,
+    create_key_hash bytea,
+
+    -- * exercise events (consuming and non_consuming)
+    exercise_choice text,
```
gerolf-da, line 100: Should this already contain the key value for exercise nodes as well to enable less complex maintenance of the mutable state cache?
@tudor-da: Or were you planning to add this later as part of the mutable state cache work?

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -0,0 +1,172 @@
...
+    exercise_result_compression SMALLINT
+) PARTITION BY LIST (event_kind);
+
+
+-- Set storage parameters before creating partitions so they follow the pattern.
+-- these columns contain data that is generally incompressible, so don't try it
+ALTER TABLE participant_events ALTER COLUMN create_key_hash  SET STORAGE EXTERNAL;
+
+-- Partition the events according to the event type.
+-- TODO: manually partition the table in order to avoid using PostgreSQL-specific features
```
gerolf-da, line 119: Doing manual partitioning would mean that the `participant_events` table doesn't exist at all and the insertion happens directly to these tables, correct?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
gerolf-da: I concur.

gerolf-da (changes requested):
~Temporarily blocking to sort out the bump to postgres 12 before merging.~
Please switch to manual partitioning and bumping to postgresql 10 (for crash tolerant hash indices).

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
tudor-da: Yes, I was planning to do it as part of the mutable state cache epic.

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
miklos-da: That makes sense, assuming we don't guarantee that an 'append-only' schema may be migrated after a 'stable' schema change (which is a perfectly reasonable assumption).

nmarton-da:
> ~Temporarily blocking to sort out the bump to postgres 12 before merging.~
> Please switch to manual partitioning and keep the postgres version at 9.6. We'll decide later when to bump to postgres 10.

@gerolf-da @rautenrieth-da 
As discussed offline:
- We will switch to manual partitioning as part of the epic.
- We won't switch to manual partitioning as part of this PR: this PR's purpose to land an experimental schema, we will stabilize/change later.
- The bump of the ephemeral postgres support is not strictly needed neither here, neither after finishing (since we will switch to manual partitioning, 9.6 postgreSQL would be sufficient. Bumping is only a convenience to reflect the fact that we already are and also will rely on hash indexes, which is durable only after PG version 10). So please @rautenrieth-da remove ephemeral PG version bump related changes.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -0,0 +1,172 @@
...
+--
+-- This is a major redesign of the index database schema. Updates from the ReadService are
+-- now written into the append-only table participant_events, and the set of active contracts is
+-- reconstructed from the log of create and archive events.
+---------------------------------------------------------------------------------------------------
+
+-- This migration is work in progress. Data continuity is not implemented yet.
+-- Even though this migration is hidden behind a feature flag, we add safety checks
+-- that prevent it from running on a non-empty database.
+CREATE TABLE safety_check(
```
nmarton-da, line 16: kudos :)

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
nmarton-da: @gerolf-da  @miklos-da  Please note: as of this is getting merged, the generic indexer DB schema is frozen, until the gradual rollout finishes with the cleanup step. So the step Gerolf is laying out only apply for a change which is a resolution of a major incident impacting all clients or something ;)
Also please note: this is temporary code, as soon as cleanup conducted, these safeguards will disappear.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
nmarton-da, line 16: please place a TODO here that this will be removed as soon as data migration efforts and schema reached production readiness.

nmarton-da, line 16: can you add a test for verifying, that this safeguard is working?

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -21,6 +21,10 @@ private[platform] class FlywayMigrations(jdbcUrl: String)(implicit loggingContex
 
   private val dbType = DbType.jdbcType(jdbcUrl)
 
+  // The last stable version.
+  // Across all database types, experimental migrations must have a higher version number.
+  private val lastStableSchemaVersion = "49"
```
nmarton-da, line 26: please mark it up with `TODO append-only` stating this feature is temporary and will be removed in the cleanup step

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
nmarton-da: I agree as well, I would suggest calling this appendOnly (e.g. enableAppendOnlySchema here), and renaming accordingly above as well if needed

nmarton-da, line 42: I would suggest remove config flag from everywhere above FlywayMigrations: no code exists which can make use of this schema yet, so I would only adapt the factories as the last step of the PoC rollout, at: DPP-316. My working assumption: we do not need a separate flag for turning on the schema and using the new code, this will be directed by one flag.

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -32,14 +36,16 @@ private[platform] class FlywayMigrations(jdbcUrl: String)(implicit loggingContex
     }
 
   def migrate(
-      allowExistingSchema: Boolean = false
+      allowExistingSchema: Boolean = false,
+      enableExperimentalSchema: Boolean = false,
```
nmarton-da, line 40: rename to appendOnly (e.g. enableAppendOnlySchema): this will be the same flag probably throughout the whole rollout

nmarton-da (changes requested):
Looks good, some pruning needed though. Interesting stuff with the check table!

rautenrieth-da:
@gerolf-da @nmarton-da : I have removed the change that bumps the postgres version. The flag to enable the new schema now has the same name everywhere, and is commented with
```
// TODO append-only:
```
which will be the string used to mark all places that need cleanup at the end

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -362,6 +364,15 @@ object Config {
             "Enable the development version of the Daml-LF language. Highly unstable. Should not be used in production."
           )
 
+        // TODO append-only: remove after removing support for the current (mutating) schema
+        opt[Unit]("index-append-only-schema-unsafe")
+          .optional()
+          .hidden()
+          .text(
+            s"Use the experimental append-only index database. Highly unstable. Should not be used in production."
```
nmarton-da, line 372: Can you please adapt the description? This flag is now enabling parallel ingestion as well, and will be gradually non experimental.

nmarton-da (approved):
LGTM

gerolf-da (approved):
Thanks, @rautenrieth-da. 👍

PR #9250, opened 2021-03-26
Extend `scalacheck` testing to generate `NodeRollback` nodes

nickchapman-da:
Extend `scalacheck` testing to generate `NodeRollback` nodes
- temporarily controlled by `allowRollback`
- set `allowRollback = false` for tests which dont yet work. Marked by `TODO`
- `allowRollback = true` used in `NormalizerSpec` tests
- implement `version` support in `NodeRollback` (so `NormalizerSpec` test will pass)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/kv-transaction-support/src/main/scala/com/daml/lf/kv/Normalizer.scala
+++ b/daml-lf/kv-transaction-support/src/main/scala/com/daml/lf/kv/Normalizer.scala
@@ -13,15 +13,9 @@ object Normalizer {
   def normalizeTransaction(
       tx: CommittedTransaction
   ): CommittedTransaction = {
-    // TODO https://github.com/digital-asset/daml/issues/8020
-
-    // We need drop Rollback nodes and all their children. Before we do this we need to
-    // update NormalizerSpec. Which in turn requires we extend our scalagen generators to
-    // generate transactions containing rollback nodes.
     val nodes = tx.nodes.filter {
       case (_, _: Node.NodeRollback[_] | _: Node.NodeFetch[_] | _: Node.NodeLookupByKey[_]) => false
-      case (_, _: Node.NodeExercises[_, _] | _: Node.NodeCreate[_]) =>
-        true
+      case (_, _: Node.NodeExercises[_, _] | _: Node.NodeCreate[_]) => true
     }
     val filteredNodes = nodes.transform {
       case (_, node: Node.NodeExercises[NodeId, ContractId]) =>
--- a/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
+++ b/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
@@ -14,6 +14,7 @@ import com.daml.lf.transaction.Node.{
   NodeExercises,
   NodeFetch,
   NodeLookupByKey,
+  NodeRollback,
 }
 import com.daml.lf.transaction.{
   BlindingInfo,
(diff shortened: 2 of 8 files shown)


daml-lf/kv-transaction-support/src/main/scala/com/daml/lf/kv/Normalizer.scala
```diff
@@ -13,15 +13,9 @@ object Normalizer {
...
-    // We need drop Rollback nodes and all their children. Before we do this we need to
-    // update NormalizerSpec. Which in turn requires we extend our scalagen generators to
-    // generate transactions containing rollback nodes.
     val nodes = tx.nodes.filter {
       case (_, _: Node.NodeRollback[_] | _: Node.NodeFetch[_] | _: Node.NodeLookupByKey[_]) => false
-      case (_, _: Node.NodeExercises[_, _] | _: Node.NodeCreate[_]) =>
-        true
+      case (_, _: Node.NodeExercises[_, _] | _: Node.NodeCreate[_]) => true
     }
     val filteredNodes = nodes.transform {
```
remyhaemmerle-da, line 20: You have to be a bit more clever here. 
Like that, the children of the rollback nodes will not be accessible anymore from the root by will stay in `filterNodes`.

remyhaemmerle-da: You could extends maybe you could *you should anyway at some point) extends `Transaction.foldInExecutionOrder` to handle Rollback nodes and use it here.

cocreature (approved):
lgtm, thanks!

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
```diff
@@ -353,6 +354,17 @@ object ValueGenerators {
     )
   }
 
+  /** Makes rollback node with some random child IDs. */
```
remyhaemmerle-da, line 357: @stephencompall-DA Could you take a look at this file ?

remyhaemmerle-da (commented):
There is an issue with dropping the rollback node and its descendants. 
Otherwise it look good for me.

miklos-da (approved):

daml-lf/kv-transaction-support/src/main/scala/com/daml/lf/kv/Normalizer.scala
nickchapman-da: Thanks, I see the problem.
I will address this in the next PR

remyhaemmerle-da: :+1:

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
S11001001: Most tests shouldn't care that the generated transactions are well-formed in any way. If expecting the IDs in the rollback is an exception, then it's worth worrying about; otherwise random IDs should be fine.

S11001001: However, all existing uses of this kind of node-ID generation use "danglingRef" or "malformed" in their name to indicate this, and this practice has not been carried through here.

PR #9252, opened 2021-03-26
get oracle image from Docker Hub

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -214,12 +214,13 @@ jobs:
         eval "$(./dev-env/bin/dade-assist)"
         export ORACLE_PORT=1521
         export ORACLE_PWD=hunter2
-        docker login --username $ARTIFACTORY_USERNAME --password $ARTIFACTORY_PASSWORD digitalasset-daml-on-oracle-docker.jfrog.io
-        docker pull digitalasset-daml-on-oracle-docker.jfrog.io/oracle-xe-preloaded:2021-02-03-5a88b29
+        docker login --username "$DOCKER_LOGIN" --password "$DOCKER_PASSWORD"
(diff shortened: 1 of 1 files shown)


ci/build.yml
```diff
@@ -214,12 +214,13 @@ jobs:
         eval "$(./dev-env/bin/dade-assist)"
         export ORACLE_PORT=1521
         export ORACLE_PWD=hunter2
-        docker login --username $ARTIFACTORY_USERNAME --password $ARTIFACTORY_PASSWORD digitalasset-daml-on-oracle-docker.jfrog.io
-        docker pull digitalasset-daml-on-oracle-docker.jfrog.io/oracle-xe-preloaded:2021-02-03-5a88b29
+        docker login --username "$DOCKER_LOGIN" --password "$DOCKER_PASSWORD" hub.docker.com
+        IMAGE=digitalasset/oracle-xe-18.4.0-preloaded:20210312-16-7009a64
```
cocreature, line 218: can we also switch to the non-xe version or do you want to do this separately?

cocreature (approved):
Thanks!

ci/build.yml
garyverhaegen-da: Happy to switch; I guess I've not followed this closely enough to know we needed to switch.

cocreature: I _think_ oracle is reasonablly backwards compatible bu tthe other version is what will be run in production so seems better to test against that.

garyverhaegen-da: I did not manage to make it work, so I've reverted this PR to just doing the Docker Hub thing and I'll open a new one with "how far I got" in the switch to Oracle Entreprise.

PR #9257, opened 2021-03-26
Ledger api pruning helper to aid pruning by time

oliverse-da:
Pruning by offset puts the burden of finding a suitable offset on the user
(see https://docs.daml.com/ops/index.html#determining-a-suitable-pruning-offset)
whereas reasoning in terms of time is more meaningful intuitively abeit
more ambiguous particularly against ledger that don't produce monotonically
increasing record time where time can "jump around" making ambiguous the
assignment of an offset to a timestamp.

This adds a "get_offset_from_time" command to the ParticipantPruningService that
returns the largest offset for which no lower offset exists with a record time
larger than the provided input timestamp.

Marked as a draft pending:
- consensus on this helper command
- adding tests to `ParticipantPruningIT`

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
@@ -38,7 +38,7 @@ service ParticipantPruningService {
   rpc Prune (PruneRequest) returns (PruneResponse);
 
   // Look up offset by timestamp to help determine a suitable pruning offset by ``record_time`` as exposed by the
-  // ``CommandCompletionService`` ``Checkpoint``s.
+  // ``CommandCompletionService`` Checkpoints.
   //
   // Returns the last offset of the "first" event/transaction before or at the specified timestamp.
   // Note that for ledger in which ``record_time`` increases monotonically with respect to offsets the resulting offset
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
@@ -5,12 +5,13 @@ syntax = "proto3";
 
 package com.daml.ledger.api.v1.admin;
 
+import "com/daml/ledger/api/v1/ledger_offset.proto";
+
+import "google/protobuf/timestamp.proto";
+
 option java_outer_classname = "ParticipantPruningServiceOuterClass";
 option java_package = "com.daml.ledger.api.v1.admin";
 option csharp_namespace = "Com.Daml.Ledger.Api.V1.Admin";
-
-// Status: experimental interface, will change before it is deemed production
-// ready
 
 // Prunes/truncates the "oldest" transactions from the participant (the participant Ledger Api Server plus any other
 // participant-local state) by removing a portion of the ledger in such a way that the set of future, allowed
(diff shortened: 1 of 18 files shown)


ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
```diff
@@ -36,6 +37,16 @@ service ParticipantPruningService {
...
 
+  // Look up offset by timestamp to help determine a suitable pruning offset by ``record_time`` as exposed by the
+  // ``CommandCompletionService`` ``Checkpoint``s.
+  //
+  // Returns the last offset of the "first" event/transaction before or at the specified timestamp.
+  // Note that for ledger in which ``record_time`` increases monotonically with respect to offsets the resulting offset
+  // is uniquely defined (assuming events before or at the specified timestamp exist). For ledgers in which ``record_time``
+  // does not increase monotonically, there may be events with lower ``record_time`` at higher offsets than the returned
+  // offset. In such cases this command serves as an approximate heuristic.
+  rpc GetOffsetByTime (GetOffsetByTimeRequest) returns (GetOffsetByTimeResponse);
```
cocreature, line 48: Is there a reason why this should be limited to pruning? It seems like there might also be value in running something like "give me the transactions that happened yesterday" (and e.g. produce a daml dump based on that) where this could come in handy. That would suggest placing it closer to `GetLedgerEnd`.

oliverse-da: Yes, I was thinking about this too. In fact the completions/and or transactions service could then also be a good place for the `GetLedgerBegin` that @bame-da  was looking for.

meiersi-da: Good point @moritzkiefer-da . We would need though to make the return value specific to the transaction events that the `read_as` parties can see, as otherwise we leak the information of whether there are further events visible to parties other than the ones in `read_as`.

That in turn means that the pruning service lookup requires a different query, as it should consider all events stored in the IndexDB _independent_ of their per-party visibility.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
```diff
@@ -50,3 +61,13 @@ message PruneRequest {
 message PruneResponse {
   // Empty for now, but may contain fields in the future
 }
+
+message GetOffsetByTimeRequest {
+  // Inclusive ``record_time`` timestamp up to and at which the ledger is to be pruned.
```
meiersi-da, line 66: @oliverse-da : the comment seems off, as the request won't be pruning anything.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/participant_pruning_service.proto
```diff
@@ -50,3 +61,13 @@ message PruneRequest {
...
+  google.protobuf.Timestamp prune_up_to = 1;
+}
+
+message GetOffsetByTimeResponse {
+  // The absolute, inclusive offset of the last event/transaction at or before the ``prune_up_to`` input timestamp.
```
meiersi-da, line 71: @oliverse-da : what do you think about explaining the semantics of the endpoint here by relating a concrete request-response-pair to the observations possible on the completion streams and transaction streams for different parties; i.e., something along the lines that up to the returned offset all completions and transactions will have an `effective_at` field that is lesser or equal to the given timestamp.

PR #9265, opened 2021-03-29
Generate DALFs for data-dependencies in Daml script dump

### LINKED ISSUE #9085: Revisit dependency handling in daml script dumps
Currently daml script dumps try to reconstruct a bunch of DARs for the dependencies. This really makes no sense, we just want dalfs.

data-dependencies actually already support paths to dalfs already. I think we probably need some `--package` flags in addition to that to expose packages but that still seems much nicer than the dumb logic for reconstructing DARs we have now.

aherrmann-da:
Closes #9085 

- DALFs in data-dependencies are no longer marked as main DALFs. This caused the LF version consistency check to fail on DALFs that were generated with a different LF version and also occurred in `dalfsFromDependencies`, e.g. `daml-prim-DA-Types`. This means that DALF data-dependencies that are imported directly need an accompanying `--package` flag.
- Daml script dump now generates DALFs for data-dependencies instead of DARs.
- Updates affected test-cases

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
@@ -133,11 +133,11 @@ installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pD
         -- install data-dependencies
         ----------------------------
         forM_ dataDepsDars $ extractDar >=> installDar depsDir True
-        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf depsDir fp
+        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf False depsDir fp
         exclPkgIds <- queryPkgIds Nothing depsDir
         rdalfs <- getDalfsFromLedger dataDepsPkgIds exclPkgIds
         forM_ rdalfs $ \RemoteDalf {..} -> do
-            installDataDepDalf depsDir (remoteDalfName <.> "dalf") remoteDalfBs
+            installDataDepDalf remoteDalfIsMain depsDir (remoteDalfName <.> "dalf") remoteDalfBs
         -- Mark received packages as well as their transitive dependencies as data dependencies.
         markAsDataRec
             (Set.fromList [remoteDalfPkgId | RemoteDalf {remoteDalfPkgId} <- rdalfs])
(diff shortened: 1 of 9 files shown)


compiler/damlc/lib/DA/Cli/Damlc/Packaging.hs
```diff
@@ -514,9 +514,9 @@ data DependencyInfo = DependencyInfo
   -- Note that for data-dependencies it is sufficient to list a DAR without
   -- listing all of its dependencies.
   , mainUnitIds :: [UnitId]
-  -- ^ Unit id of the main DALFs specified in dependencies and
-  -- data-dependencies. This will be used to generate the --package
-  -- flags which define which packages are exposed by default.
+  -- ^ Unit id of the main DALFs specified in dependencies.
```
moritzkiefer-da, line 517: It still applies to DARs in data-dependencies right? I think the comment is mostly fine. What change is the definition of "main dalf" which no longer applies to indvidual dalfs.

daml-script/dump/src/main/scala/com/daml/script/dump/Dump.scala
```diff
@@ -36,26 +33,31 @@ object Dump {
         .render(80)
         .getBytes(StandardCharsets.UTF_8),
     )
-    val dars: Seq[Dar[(PackageId, ByteString, Ast.Package)]] =
-      pkgRefs.view.collect(Function.unlift(Dependencies.toDar(_, pkgs))).toSeq
+    val packages: Seq[String] =
```
moritzkiefer-da, line 36: maybe call that `exposedPackages` or something like that?

daml-script/dump/src/main/scala/com/daml/script/dump/Dump.scala
```diff
@@ -36,26 +33,31 @@ object Dump {
...
+      pkgRefs.view.collect(Function.unlift(Dependencies.toPackages(_, pkgs))).toSeq
+    val dalfs: Seq[(PackageId, ByteString, Ast.Package)] =
+      pkgRefs.view.flatMap(Dependencies.toDalfs(_, pkgs)).toSeq
```
moritzkiefer-da, line 39: I don’t quite understand this.
Before we were looking at the direct references producing one DAR for each of them because the main dalf of each of those gets exposed.
Now we produce the direct references in `packages`. That’s all good. But then we chase dependencies for each of those to produce `dalfs` which can contain duplicate packages afaik. Don’t we already know that we need all the dalfs in `pkgs`? Can we just map over those filtering out a few and kill the `toDalfs` stuff?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -133,11 +133,11 @@ installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pD
         -- install data-dependencies
         ----------------------------
         forM_ dataDepsDars $ extractDar >=> installDar depsDir True
-        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf True depsDir fp
+        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf depsDir fp
         exclPkgIds <- queryPkgIds Nothing depsDir
         rdalfs <- getDalfsFromLedger dataDepsPkgIds exclPkgIds
         forM_ rdalfs $ \RemoteDalf {..} -> do
-            installDataDepDalf remoteDalfIsMain depsDir (remoteDalfName <.> "dalf") remoteDalfBs
+            installDataDepDalf depsDir (remoteDalfName <.> "dalf") remoteDalfBs
```
moritzkiefer-da, line 140: Why do we need to drop this? If I reference a package id in my daml.yaml I do expect this to be the main.

moritzkiefer-da (approved):
lgtm with a few comments

PR #9270, opened 2021-03-29
remote pkgs: resolve package names in data dependencies via ledger

robin-da:
This adds the ability to specify package names/versions in daml.yaml in
the data-dependencies stanza. They are being resolved via the project
ledger and a newly introduced daml.lock lock file.

More tests for the caching of dependencies and resolution will be added in a follow-up PR.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
@@ -283,8 +283,8 @@ validatePkgId pkgId = do
 ---------------------
 
 data FullPkgName = FullPkgName
-    { pkgName :: LF.PackageName
-    , pkgVersion :: LF.PackageVersion
+    { pkgName :: !LF.PackageName
+    , pkgVersion :: !LF.PackageVersion
     } deriving (Eq, Ord, Show)
 
 data LockFile = LockFile
--- a/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
@@ -15,7 +15,6 @@ import Control.Exception.Safe (tryAny)
 import Control.Lens (toListOf)
 import Control.Monad.Extra
 import DA.Daml.Compiler.Dar
-import Data.Char
 import DA.Daml.Compiler.DecodeDar (DecodedDalf(..), decodeDalf)
 import DA.Daml.Compiler.ExtractDar (ExtractedDar(..), extractDar)
 import DA.Daml.Helper.Ledger
@@ -29,8 +28,10 @@ import qualified Data.Aeson as Aeson
 import Data.Aeson (eitherDecodeFileStrict', encode)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BSL
+import Data.Char
 import Data.List.Extra
 import qualified Data.Map.Strict as M
+import Data.Maybe
 import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Data.Yaml as Yaml
@@ -142,7 +143,7 @@ installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pD
         -- install data-dependencies
         ----------------------------
         forM_ dataDepsDars $ extractDar >=> installDar depsDir True
-        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf True depsDir fp
+        forM_ dataDepsDalfs $ \fp -> BS.readFile fp >>= installDataDepDalf False depsDir fp
         resolvedPkgIds <- resolvePkgs projRoot opts dataDepsNameVersion
         exclPkgIds <- queryPkgIds Nothing depsDir
         rdalfs <- getDalfsFromLedger (dataDepsPkgIds ++ M.elems resolvedPkgIds) exclPkgIds
@@ -150,7 +151,7 @@ installDependencies projRoot opts sdkVer@(PackageSdkVersion thisSdkVer) pDeps pD
             installDataDepDalf
                 remoteDalfIsMain
                 depsDir
-                (packageNameToFp remoteDalfName)
+                (packageNameToFp $ packageNameOrId remoteDalfPkgId remoteDalfName)
                 remoteDalfBs
         -- Mark received packages as well as their transitive dependencies as data dependencies.
         markAsDataRec
@@ -293,13 +294,16 @@ data LockFile = LockFile
 instance Aeson.FromJSON LockFile
 instance Aeson.ToJSON LockFile
 data DependencyInfo = DependencyInfo
-    { name :: String
-    , version :: String
-    , pkgId :: String
+    { name :: LF.PackageName
+    , version :: LF.PackageVersion
+    , pkgId :: LF.PackageId
     } deriving Generic
 instance Aeson.FromJSON DependencyInfo
 instance Aeson.ToJSON DependencyInfo
 
+-- | Resolves the given list of package names/versions to package IDs.
+-- This will fail if any package can't be resolved.
+-- Once all packages have been resolved, a new `daml.lock` file is noting the resolution.
 resolvePkgs :: NormalizedFilePath -> Options -> [FullPkgName] -> IO (M.Map FullPkgName LF.PackageId)
 resolvePkgs projRoot opts pkgs
     | null pkgs = pure M.empty
@@ -329,15 +333,19 @@ writeLockFile lockFp resolvedPkgs = do
     Yaml.encodeFile lockFp $
         LockFile
             [ DependencyInfo
-                { name = T.unpack $ LF.unPackageName pkgName
-                , version = T.unpack $ LF.unPackageVersion pkgVersion
-                , pkgId = T.unpack $ LF.unPackageId pkgId
+                { name = pkgName
+                , version = pkgVersion
+                , pkgId = pkgId
                 }
             | (FullPkgName {pkgName, pkgVersion}, pkgId) <- M.toList resolvedPkgs
             ]
 
--- | Read the package lock file and resolve the given packages. Returns a list of unresolved
--- packages together with the resolved ones.
+-- | Read the package lock file and resolve the given packages.
+-- Returns
+--  Nothing -> At least one of the given packages could not be resolved.
+--  Just map -> Resolution of all packages
+--
+-- It fails if the `daml.lock` file can't be parsed.
 resolvePkgsWithLockFile :: FilePath -> [FullPkgName] -> IO (Maybe (M.Map FullPkgName LF.PackageId))
 resolvePkgsWithLockFile lockFp pkgs = do
     hasLockFile <- doesFileExist lockFp
@@ -352,13 +360,15 @@ resolvePkgsWithLockFile lockFp pkgs = do
                     let m =
                             M.fromList
                                 [ ( FullPkgName
-                                        { pkgName = LF.PackageName $ T.pack $ name d
-                                        , pkgVersion = LF.PackageVersion $ T.pack $ version d
+                                        { pkgName = name d
+                                        , pkgVersion = version d
                                         }
-                                  , LF.PackageId $ T.pack $ pkgId d)
+                                  , pkgId d)
                                 | d <- dependencies
                                 ]
                     pure $
+                        -- Check that all given packages could be resolved, return Nothing
+                        -- otherwise.
                         if Set.fromList pkgs `Set.isSubsetOf` M.keysSet m
                             then Just $ M.restrictKeys m (Set.fromList pkgs)
                             else Nothing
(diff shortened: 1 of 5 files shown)


compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
         fail $ "Invalid package ID dependency in daml.yaml: " <> pkgId
     pure $ LF.PackageId $ T.pack pkgId
 
+-- Package resolution
+---------------------
+
+data FullPkgName = FullPkgName
+    { pkgName :: LF.PackageName
+    , pkgVersion :: LF.PackageVersion
```
cocreature, line 287:
```suggestion
    { pkgName :: !LF.PackageName
    , pkgVersion :: !LF.PackageVersion
```

cocreature: Probably want this on everything here, I don’t think any of these should be lazy.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -352,19 +353,20 @@ downloadPackage args pid = do
     convPid (LF.PackageId text) = L.PackageId $ TL.fromStrict text
 
 data RemoteDalf = RemoteDalf
-    { remoteDalfName :: String
+    { remoteDalfName :: LF.PackageName
+    , remoteDalfVersion :: Maybe LF.PackageVersion
```
cocreature, line 357: This looks like we have to decode all packages even if we only have package ids (either because they are specified in daml.yaml or because we already have a lockfile). Is that required?

docs/source/daml/reference/packages.rst
```diff
@@ -184,6 +184,39 @@ When importing packages this way, the Daml compiler will try to reconstruct the
 
 Because of their flexibility, data-dependencies are a tool that is recommended for performing Daml model upgrades. See the :ref:`upgrade documentation <upgrade-overview>` for more details.
 
+Importing Daml packages from the project ledger
```
cocreature, line 187:
```suggestion
Referencing Daml packages already on the ledger
```

docs/source/daml/reference/packages.rst
```diff
@@ -184,6 +184,39 @@ When importing packages this way, the Daml compiler will try to reconstruct the
...
+If your ledger runs at the default host and port (``localhost:6865``), the ledger stanza can be
+omitted. This will fetch and install the package ``foo-1.0.0``. A ``daml.lock`` file is created at
+the root of your project directory, pinning the resolved packages to their exact package ID:
+
+.. code-block:: yaml
+
+  dependencies:
+  - pkgId: 51255efad65a1751bcee749d962a135a65d12b87eb81ac961142196d8bbca535
+    name: foo
+    version: 1.0.0
```
cocreature, line 214: Given that this is mainly intended as a file for machines rather than humans, should we render it out as json?

compiler/damlc/tests/src/DamlcPkgManager.hs
```diff
@@ -70,6 +71,77 @@ testsForRemoteDataDependencies damlc dar =
...
+                  exitCode @?= ExitSuccess
+                  lock <- readFile lockFp
+                  lines lock @?=
+                    [
+                        "dependencies:"
+                        , "- pkgId: " <> mainPkgId
+                        , "  name: pkg-manager-test"
+                        , "  version: 1.0.0"
+                    ]
+                  -- a second call shouldn't make any network connections. So let's set the sandbox
```
cocreature, line 128: Nice!

compiler/damlc/tests/src/DamlcPkgManager.hs
```diff
@@ -70,6 +71,77 @@ testsForRemoteDataDependencies damlc dar =
...
+                      unlines
+                          [ "sdk-version: " <> sdkVersion
+                          , "name: a"
+                          , "version: 0.0.1"
+                          , "source: ."
+                          , "dependencies: [daml-prim, daml-stdlib]"
+                          , "data-dependencies: [pkg-manager-test:1.0.0]"
+                          , "ledger:"
+                          , "  host: localhost"
+                          , "  port: " <> show (sandboxPort + 1)
```
cocreature, line 140: does -1 work just to be completely sure it’s something that isn’t going to work?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+data DependencyInfo = DependencyInfo
+    { name :: String
+    , version :: String
+    , pkgId :: String
+    } deriving Generic
+instance Aeson.FromJSON DependencyInfo
+instance Aeson.ToJSON DependencyInfo
+
+resolvePkgs :: NormalizedFilePath -> Options -> [FullPkgName] -> IO (M.Map FullPkgName LF.PackageId)
+resolvePkgs projRoot opts pkgs
```
cocreature, line 304: I think a doc comment here would be useful.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+                , pkgId = T.unpack $ LF.unPackageId pkgId
+                }
+            | (FullPkgName {pkgName, pkgVersion}, pkgId) <- M.toList resolvedPkgs
+            ]
+
+-- | Read the package lock file and resolve the given packages. Returns a list of unresolved
+-- packages together with the resolved ones.
+resolvePkgsWithLockFile :: FilePath -> [FullPkgName] -> IO (Maybe (M.Map FullPkgName LF.PackageId))
+resolvePkgsWithLockFile lockFp pkgs = do
+    hasLockFile <- doesFileExist lockFp
```
cocreature, line 343: `doesFileExist` is an antipattern imho since it’s inherently racy. Read and catch `isDoesNotExistError`.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+        LockFile
+            [ DependencyInfo
+                { name = T.unpack $ LF.unPackageName pkgName
+                , version = T.unpack $ LF.unPackageVersion pkgVersion
+                , pkgId = T.unpack $ LF.unPackageId pkgId
+                }
+            | (FullPkgName {pkgName, pkgVersion}, pkgId) <- M.toList resolvedPkgs
+            ]
+
+-- | Read the package lock file and resolve the given packages. Returns a list of unresolved
```
cocreature, line 339: Comment on semantics would be useful:
1. Not exists -> Nothing
2. Exists but invalid format -> Error

Related question, what happens on an SDK upgrade with an incompatible format change? I guess ideally we don’t want to make those so it seems fine to break this but maybe worth documenting.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+    } deriving (Eq, Ord, Show)
+
+data LockFile = LockFile
+    { dependencies :: [DependencyInfo]
+    } deriving Generic
+instance Aeson.FromJSON LockFile
+instance Aeson.ToJSON LockFile
+data DependencyInfo = DependencyInfo
+    { name :: String
+    , version :: String
```
cocreature, line 297: Any reason not to store `LF.PackageName`, `LF.PackageVersion` and `LF.PackageId` here? Would save you a few conversions. Getting a JSON instance for those should be easy if we don’t already have one (just gnd it)

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+                                        { pkgName = LF.PackageName $ T.pack $ name d
+                                        , pkgVersion = LF.PackageVersion $ T.pack $ version d
+                                        }
+                                  , LF.PackageId $ T.pack $ pkgId d)
+                                | d <- dependencies
+                                ]
+                    pure $
+                        if Set.fromList pkgs `Set.isSubsetOf` M.keysSet m
+                            then Just $ M.restrictKeys m (Set.fromList pkgs)
+                            else Nothing
```
cocreature, line 364: What is this checking for?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+    , pkgId :: String
+    } deriving Generic
+instance Aeson.FromJSON DependencyInfo
+instance Aeson.ToJSON DependencyInfo
+
+resolvePkgs :: NormalizedFilePath -> Options -> [FullPkgName] -> IO (M.Map FullPkgName LF.PackageId)
+resolvePkgs projRoot opts pkgs
+    | null pkgs = pure M.empty
+    | otherwise = do
+        mbRes <- resolvePkgsWithLockFile lockFp pkgs
```
cocreature, line 307: What happens if nothing changed but the lock file is not sufficient. Will `daml build` keep refetching everything on every run and fail?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
```diff
@@ -246,24 +279,143 @@ validatePkgId pkgId = do
...
+
+-- | Query the ledger for available packages and try to resolve the given packages. Returns either
+-- the missing packages or a map from package names to package id.
+resolvePkgsWithLedger ::
+       FilePath -> [FullPkgName] -> IO (Either [FullPkgName] (M.Map FullPkgName LF.PackageId))
+resolvePkgsWithLedger depsDir pkgs = do
+    ledgerPkgIds <- listLedgerPackages
+    rdalfs <- getDalfsFromLedger ledgerPkgIds []
+    forM_ rdalfs $ \RemoteDalf {..} ->
+        installDalf [] depsDir (packageNameToFp remoteDalfName) remoteDalfBs
```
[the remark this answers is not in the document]
robin-da: The idea is that I don't want to download the package again after I downloaded everything for resolving. It's installed but not marked as dep/data-dep. In the next step, when packages are downloaded, this package will not be downloaded anymore and only marked with the right markers. 
Where would you want to separate it?

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: No, it will re-fetch once and write the new lock-file.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: this checks, that all required packages could be resolved with the lock-file.

docs/source/daml/reference/packages.rst
robin-da: I prefer yaml to json, because less brackets more readable if I have to take a look for debugging. But I don't have a strong opinion about this one.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: True, too many conversions. I'll change it.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -375,12 +377,23 @@ runLedgerGetDalfs lflags pkgIds exclPkgIds
             [ RemoteDalf {..}
             | (_pid, Just pkg) <- Map.toList m
             , let (bsl, pid) = LFArchive.encodeArchiveAndHash pkg
+            , let LF.Package {packageMetadata} = pkg
             , let remoteDalfPkgId = pid
-            , let remoteDalfName = T.unpack $ recoverPackageName pkg pid
+            , let remoteDalfName =
+                      maybe (LF.PackageName $ LF.unPackageId pid) LF.packageName packageMetadata
```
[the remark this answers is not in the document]
robin-da: Moved it to the callsite.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: This will fail when an incompatible format change is encountered. The error message will however indicate the user to delete the lock file and re-run. The next invocation will write out a properly formatted lock file.

I added a comment for the exact semantics.

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
cocreature: Maybe worth adding an inline comment here.

docs/source/daml/reference/packages.rst
cocreature: Alright, let’s stick with yaml for now.

cocreature (approved):
Nice work, thank you!

compiler/damlc/lib/DA/Cli/Damlc/DependencyDb.hs
robin-da: added an inline comment.

PR #9272, opened 2021-03-29
LF-Repl: add a benchmark function

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/StandardLibraries.scala
```diff
@@ -0,0 +1,109 @@
+package com.daml.lf.language
+
+import com.daml.lf.data.{ImmArray, Ref}
+
+object StandardLibraries {
```
cocreature, line 5: did you intend to commit this? seems unrelated

daml-lf/repl/src/main/scala/com/digitalasset/daml/lf/repl/testing/Main.scala
```diff
@@ -532,9 +546,10 @@ object Repl {
             case Left((err, ledger @ _)) =>
               println(prettyError(err, onLedger.ptx).render(128))
               (false, state)
-            case Right((diff @ _, steps @ _, ledger, value @ _)) =>
+            case Right((diff @ _, steps @ _, ledger, value)) =>
               // NOTE(JM): cannot print this, output used in tests.
               //println(s"done in ${diff.formatted("%.2f")}ms, ${steps} steps")
+              println(value.toValue)
```
cocreature, line 552: intentional?

daml-lf/repl/src/main/scala/com/digitalasset/daml/lf/repl/testing/Main.scala
```diff
@@ -611,6 +626,58 @@ object Repl {
...
   }
 
+  val warmingTime = 20L
+
+  private[this] def standardDeviation(a: Seq[Long]): Double = {
+    def avg = a.map(_.toDouble).sum / a.size
+    math.sqrt(a.map(x => (x - avg) * (x - avg)).sum / a.size)
+  }
+
+  def cmdBench(state: State, testId: String): (Boolean, State) = {
```
cocreature, line 636: I think I feel slightly uneasy about reinventing JMH here. Is there a way we can just delegate to that?

cocreature (approved):
Thank you! I think it might still be worth checking if we can use jmh for this at some point but I’m fine merging it in the current form.

cocreature:
Why did you close this? This seemed like a useful feature.

PR #9278, opened 2021-03-29
Oracle json-api: websockets and testing

S11001001:
Here, the Oracle backend passes all integration tests that were previously run only against PostgreSQL, as well as some new tests introduced specifically to deal with the extra complexity of Oracle query compilation. Output of unconstrained query data as literal SQL was completely removed in this PR, so `oraclePathEscape` is now significantly less worrisome (though still needed, as we cannot avoid compiling path literals).

Extends #9159. Addressed here:

* [x] query `cmpContractPathToScalar`
* [ ] use Array get/put, or
* [ ] lose it, and test for duplicate contracts for multiple sig/obs match (different sig/obs agg orders?)
* [x] websocket (matchedQueries-supporting) queries
* [ ] union/listagg approach to ByInt queries (probably Postgres too) #9375
* [x] enable websocket tests
* [x] use 2c14ff1's approach to [on conflict ignore](https://github.com/digital-asset/daml/pull/9159/files#r603176801)
* [x] retrySqlStates
* [x] add Oracle column to `sqlWheres` in `ValuePredicateTest`, some new rows
* [x] >4kb JSON data tests
  * [x] >4kb string comparison
  * [x] >4kb JSON data output
* [x] variant value path test #9321 
* [x] test comparing a string to "& bar"

### Pull Request Checklist

trigger the build.


ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -674,38 +684,67 @@ private object OracleQueries extends Queries {
...
-              key = key.asJsObject.fields("key"),
-              payload = payload,
-              signatories = unpct(pctSignatories),
-              observers = unpct(pctObservers),
-              agreementText = agreement getOrElse "",
-            )
+      case MatchedQueryMarker.ByInt =>
+        type Ix = Int
+        // TODO we may UNION the resulting queries and aggregate the Ixes SQL-side,
+        // but this will probably necessitate the same PostgreSQL-side
```
S11001001, line 739: #9375

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -728,6 +728,37 @@ abstract class AbstractHttpServiceIntegrationTest
       }: Future[Assertion]
   }
 
+  private[this] def randomTextN(n: Int) = {
+    import org.scalacheck.Gen
+    Gen
+      .buildableOfN[String, Char](n, Gen.alphaNumChar)
```
cocreature, line 734: Any reason to limit this to alphaNumChar?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -40,7 +40,7 @@ sealed abstract class Queries {
     sql"""
       CREATE TABLE
         contract
-        (contract_id """ ++ contractIdType ++ sql""" NOT NULL PRIMARY KEY
+        (contract_id """ ++ contractIdType ++ sql""" NOT NULL CONSTRAINT contract_k PRIMARY KEY
```
cocreature, line 43: Not at all opposed to this quite the opposite but is this needed for something else in this PR or just a drive-by change?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -674,38 +684,67 @@ private object OracleQueries extends Queries {
...
-                                  FROM signatories GROUP BY contract_id) sd
-                              ON (c.contract_id = sd.contract_id)
-                       LEFT JOIN (SELECT contract_id, LISTAGG(party, '%') parties
-                                  FROM observers GROUP BY contract_id) od
-                              ON (c.contract_id = od.contract_id)
-                  WHERE (""" ++ Fragments.in(fr"sm.party", parties) ++
-      fr" OR " ++ Fragments.in(fr"om.party", parties) ++ sql""")
-                        AND tpid = $tpid AND (""" ++ predicate ++ sql")"
+    // we effectively shadow Mark because Scala 2.12 doesn't quite get
+    // that it should use the GADT type equality otherwise
```
cocreature, line 689: Could you expand on that? I don’t understand what the issue is and how this fixes it.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -721,27 +760,48 @@ private object OracleQueries extends Queries {
...
-        val pred = opath ++ "?(@ == " ++ literal.compactPrint :- ')'
-        sql"JSON_EXISTS(" ++ contractColumnName ++ sql", " ++ oraclePathEscape(pred) ++ sql")"
-      case JsObject(_) | JsArray(_) =>
-        sql"JSON_EQUAL(JSON_QUERY(" ++ contractColumnName ++ sql", " ++ oraclePathEscape(
-          opath
-        ) ++ sql"), ${literal.compactPrint: String})"
+    def existsForm[Lit: Put](literal: Lit) =
+      (
+        "?(@ == $X)", // not a Scala interpolation
+        sql" PASSING $literal AS X",
```
cocreature, line 789: What does this change do? Does it sidestep the length limit by moving the literal out of the path? If so, let’s please add a comment.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -782,10 +842,29 @@ private object OracleQueries extends Queries {
     }
   }
 
+  // XXX as with containsAtContractPath, literalScalar is too big a type
```
cocreature, line 845: xxx? Is that a todo? If so, let’s please use TODO so we only have to grep for one thing.

ledger-service/http-json/src/test/scala/com/digitalasset/http/query/ValuePredicateTest.scala
```diff
@@ -232,42 +244,71 @@ class ValuePredicateTest
...
-      import doobie.implicits._, sjd.queries.Implicits._
+      import doobie.implicits._, dbbackend.Queries.Implicits._
       Table(
-        ("query", "type", "sql"),
-        ("42", VA.int64, sql"payload = ${s"""{"$dummyFieldName":42}""".parseJson}::jsonb"),
+        ("query", "type", "postgresql", "oraclesql"),
+        (
+          "42",
+          VA.int64,
+          sql"payload = ${s"""{"$dummyFieldName":42}""".parseJson}::jsonb",
```
cocreature, line 254: I’m a bit confused by the use of `parseJson`. Are you just using that to canonicalize the format? If not, I don’t see what you gain over just writing the literal directly.

cocreature (approved):
Looks great with a few minor comments, thank you!

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -626,41 +636,41 @@ private object OracleQueries extends Queries {
   )(implicit log: LogHandler, pas: Put[Array[String]]): ConnectionIO[Int] = {
     println("insert contracts")
     println(dbcs)
-    val r = Update[(String, String, SurrogateTpId, JsValue, JsValue, String)](
+    val r = Update[(String, SurrogateTpId, JsValue, JsValue, String)](
       """
-        MERGE INTO contract USING (SELECT 1 FROM DUAL) ON (contract_id = ?)
-        WHEN NOT MATCHED THEN INSERT (contract_id, tpid, key, payload, agreement_text)
+        INSERT /*+ ignore_row_on_dupkey_index(contract(contract_id)) */
```
stefanobaghino-da, line 641: There are a few `println`s in this method. Should they be removed? If so, feel free to do it in a separate PR.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -626,41 +636,41 @@ private object OracleQueries extends Queries {
...
         }
     )
     println("inserted")
     import cats.syntax.foldable._, cats.instances.vector._
-    val r2 = Update[(String, String, String, String)](
+    val r2 = Update[(String, String)](
       """
-        MERGE INTO signatories USING (SELECT 1 FROM DUAL) ON (contract_id = ? AND party = ?)
-        WHEN NOT MATCHED THEN INSERT (contract_id, party)
+        INSERT /*+ ignore_row_on_dupkey_index(signatories(contract_id, party)) */
```
stefanobaghino-da, line 657: I'm not sure I follow what `/*` is. It looks like a C-style comment but it's embedded in a multi-line string that ultimately becomes a SQL, how is this parsed? If by any chance this is parsed as a Scala comment, should it stay? If not, I don't understand it.

stefanobaghino-da (approved):

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
cocreature: This is oracle magic fro `on conflict do nothing`. Yes really, magic comments are how this work …

stefanobaghino-da: Ok.

stefanobaghino-da: Thanks.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: I use it as a sort of weaker TODO that shouldn't block anything (i.e. in some respect not finding it with grep is a feature). Is there a better way to express that in a lightweight way?

ledger-service/http-json/src/test/scala/com/digitalasset/http/query/ValuePredicateTest.scala
S11001001: I think, when you have JSON data in code, it is a lot clearer to write it in JSON format when feasible rather than to write constructor calls.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
S11001001: `\0` isn't allowed by the database, `filter` on Gens is unreliable, and this is the closest alternative at hand.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: Drive-by, it seemed nicer to have error messages tell you _which_ constraint failed.

ledger-service/http-json/src/test/scala/com/digitalasset/http/query/ValuePredicateTest.scala
cocreature: agreed, what I’m wondering is why you write it as a JSON string literal, then call `parseJson` on it only to then convert it back to a string literal. The only thing I can come up with is canonicalization.

S11001001: > then convert it back to a string literal

This does not happen in this case.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
cocreature: not sure, I tend to use NOTE and TODO but NOTE mostly for things that are purely informational and shouldn’t be acted upon. Fine to keep XXX as well if you want.

ledger-service/http-json/src/test/scala/com/digitalasset/http/query/ValuePredicateTest.scala
cocreature: I’m confused, I thought you’re interpolating it back into the sql string?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: As a rule Scala is perfectly aware of type equalities in GADTs, i.e. _for all_ arbitrary `F: * -> *`, `F[A] ~ F[B]`. But sometimes it won't _use_ that fact until you give it a nudge. In this code, the "nudge" is the ascription `T[Query0[DBContract[Mark, JsValue, JsValue, Vector[String]]]]`, but this doesn't apply until the _result_ of `queryByCondition`, so letting Scala operate under the fantasy that Int (or SurrogateTpId) is not Mark, until it really must be, is most convenient.

I think they [finally fixed this in Scala 3](https://twitter.com/not_xuwei_k/status/1332751181119184896).

ledger-service/http-json/src/test/scala/com/digitalasset/http/query/ValuePredicateTest.scala
S11001001: Nope 🙂

cocreature: Ohhh, I see :facepalm:

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: Sure. 42c913b178f

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
S11001001: baa61c5fee2

PR #9281, opened 2021-03-29
Fork com.daml.platform.store.dao

nmarton-da:
Forking
- below LedgerDao interface
- into com.daml.platform.store.appendonlydao
- based on PoC code
This commit duplicates JdbcLedgerDao and dependencies as part of
the parallel ingestion rollout work. The fork will be removed as
the complete rollout is finished.

- adds appendonlydao package
- adapts LedgerDao: marks LedgerWriteDao methods for cleanup
- deprecates LedgerWriteDao/storeInitialState

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsReader.scala
@@ -1,0 +1,64 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+import com.daml.ledger.participant.state.v1.Offset
+import com.daml.lf.data.Ref
+import com.daml.ledger.ApplicationId
+import com.daml.ledger.api.v1.command_completion_service.CompletionStreamResponse
+import com.daml.logging.LoggingContext
+import com.daml.metrics.Metrics
+import com.daml.platform.ApiOffset
+import com.daml.platform.store.DbType
+import com.daml.platform.store.appendonlydao.events.{QueryNonPruned, SqlFunctions}
+import com.daml.platform.store.dao.LedgerDaoCommandCompletionsReader
+
+import scala.concurrent.{ExecutionContext, Future}
+
+private[appendonlydao] final class CommandCompletionsReader(
+    dispatcher: DbDispatcher,
+    dbType: DbType,
+    metrics: Metrics,
+    executionContext: ExecutionContext,
+) extends LedgerDaoCommandCompletionsReader {
+
+  private val sqlFunctions = SqlFunctions(dbType)
+
+  private def offsetFor(response: CompletionStreamResponse): Offset =
+    ApiOffset.assertFromString(response.checkpoint.get.offset.get.getAbsolute)
+
+  override def getCommandCompletions(
+      startExclusive: Offset,
+      endInclusive: Offset,
+      applicationId: ApplicationId,
+      parties: Set[Ref.Party],
+  )(implicit
+      loggingContext: LoggingContext
+  ): Source[(Offset, CompletionStreamResponse), NotUsed] = {
+    val query = CommandCompletionsTable.prepareGet(
+      startExclusive = startExclusive,
+      endInclusive = endInclusive,
+      applicationId = applicationId,
+      parties = parties,
+      sqlFunctions = sqlFunctions,
+    )
+    Source
+      .future(
+        dispatcher
+          .executeSql(metrics.daml.index.db.getCompletions) { implicit connection =>
+            QueryNonPruned.executeSql[List[CompletionStreamResponse]](
+              query.as(CommandCompletionsTable.parser.*),
+              startExclusive,
+              pruned =>
+                s"Command completions request from ${startExclusive.toHexString} to ${endInclusive.toHexString} overlaps with pruned offset ${pruned.toHexString}",
+            )
+          }
+          .flatMap(_.fold(Future.failed, Future.successful))(executionContext)
+      )
+      .mapConcat(_.map(response => offsetFor(response) -> response))
+  }
+
+}
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsTable.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsTable.scala
@@ -1,0 +1,61 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao
+
+import java.time.Instant
+
+import anorm.{Row, RowParser, SimpleSql, SqlParser, SqlStringInterpolation, ~}
+import com.daml.ledger.ApplicationId
+import com.daml.ledger.api.v1.command_completion_service.CompletionStreamResponse
+import com.daml.ledger.api.v1.completion.Completion
+import com.daml.ledger.participant.state.v1.Offset
+import com.daml.lf.data.Ref
+import com.daml.platform.store.CompletionFromTransaction.toApiCheckpoint
+import com.daml.platform.store.Conversions._
+import com.daml.platform.store.appendonlydao.events.SqlFunctions
+import com.google.rpc.status.Status
+
+private[platform] object CommandCompletionsTable {
+
+  import SqlParser.{int, str}
+
+  private val sharedColumns: RowParser[Offset ~ Instant ~ String] =
+    offset("completion_offset") ~ instant("record_time") ~ str("command_id")
+
+  private val acceptedCommandParser: RowParser[CompletionStreamResponse] =
+    sharedColumns ~ str("transaction_id") map {
+      case offset ~ recordTime ~ commandId ~ transactionId =>
+        CompletionStreamResponse(
+          checkpoint = toApiCheckpoint(recordTime, offset),
+          completions = Seq(Completion(commandId, Some(Status()), transactionId)),
+        )
+    }
+
+  private val rejectedCommandParser: RowParser[CompletionStreamResponse] =
+    sharedColumns ~ int("status_code") ~ str("status_message") map {
+      case offset ~ recordTime ~ commandId ~ statusCode ~ statusMessage =>
+        CompletionStreamResponse(
+          checkpoint = toApiCheckpoint(recordTime, offset),
+          completions = Seq(Completion(commandId, Some(Status(statusCode, statusMessage)))),
+        )
+    }
+
+  val parser: RowParser[CompletionStreamResponse] = acceptedCommandParser | rejectedCommandParser
+
+  def prepareGet(
+      startExclusive: Offset,
+      endInclusive: Offset,
+      applicationId: ApplicationId,
+      parties: Set[Ref.Party],
+      sqlFunctions: SqlFunctions,
+  ): SimpleSql[Row] = {
+    val submittersInPartiesClause =
+      sqlFunctions.arrayIntersectionWhereClause("submitters", parties)
+    SQL"select completion_offset, record_time, command_id, transaction_id, status_code, status_message from participant_command_completions where completion_offset > $startExclusive and completion_offset <= $endInclusive and application_id = $applicationId and #$submittersInPartiesClause order by completion_offset asc"
+  }
+
+  def prepareCompletionsDelete(endInclusive: Offset): SimpleSql[Row] =
+    SQL"delete from participant_command_completions where completion_offset <= $endInclusive"
+
+}
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
@@ -1,0 +1,903 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+package com.daml.platform.store.appendonlydao
+
+import java.sql.Connection
+import java.time.Instant
+import java.util.Date
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+import anorm.SqlParser._
+import anorm.{Macro, RowParser, SQL, SqlParser}
+import com.daml.daml_lf_dev.DamlLf.Archive
+import com.daml.ledger.api.domain
+import com.daml.ledger.api.domain.{LedgerId, ParticipantId, PartyDetails}
+import com.daml.ledger.api.health.HealthStatus
+import com.daml.ledger.participant.state.index.v2.{
+  CommandDeduplicationDuplicate,
+  CommandDeduplicationNew,
+  CommandDeduplicationResult,
+  PackageDetails,
+}
+import com.daml.ledger.participant.state.v1._
+import com.daml.ledger.resources.ResourceOwner
+import com.daml.ledger.{TransactionId, WorkflowId}
+import com.daml.lf.archive.Decode
+import com.daml.lf.data.Ref
+import com.daml.lf.data.Ref.{PackageId, Party}
+import com.daml.lf.engine.ValueEnricher
+import com.daml.lf.transaction.{BlindingInfo, GlobalKey}
+import com.daml.lf.value.Value.ContractId
+import com.daml.logging.LoggingContext.withEnrichedLoggingContext
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.metrics.{Metrics, Timed}
+import com.daml.platform.configuration.ServerRole
+import com.daml.platform.indexer.OffsetStep
+import com.daml.platform.store.Conversions._
+import com.daml.platform.store.SimpleSqlAsVectorOf.SimpleSqlAsVectorOf
+import com.daml.platform.store._
+import com.daml.platform.store.appendonlydao.CommandCompletionsTable.prepareCompletionsDelete
+import com.daml.platform.store.appendonlydao.events._
+import com.daml.platform.store.appendonlydao.events.EventsTableDelete
+import com.daml.platform.store.dao.events.TransactionsWriter.PreparedInsert
+import com.daml.platform.store.dao.{
+  LedgerDao,
+  LedgerReadDao,
+  MeteredLedgerDao,
+  MeteredLedgerReadDao,
+  PersistenceResponse,
+}
+import com.daml.platform.store.entries.{
+  ConfigurationEntry,
+  LedgerEntry,
+  PackageLedgerEntry,
+  PartyLedgerEntry,
+}
+import scalaz.syntax.tag._
+
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+
+private final case class ParsedPartyData(
+    party: String,
+    displayName: Option[String],
+    ledgerOffset: Offset,
+    explicit: Boolean,
+    isLocal: Boolean,
+)
+
+private final case class ParsedPackageData(
+    packageId: String,
+    sourceDescription: Option[String],
+    size: Long,
+    knownSince: Date,
+)
+
+private final case class ParsedCommandData(deduplicateUntil: Instant)
+
+private class JdbcLedgerDao(
+    dbDispatcher: DbDispatcher,
+    dbType: DbType,
+    servicesExecutionContext: ExecutionContext,
+    eventsPageSize: Int,
+    performPostCommitValidation: Boolean,
+    metrics: Metrics,
+    lfValueTranslationCache: LfValueTranslation.Cache,
+    validatePartyAllocation: Boolean,
+    enricher: Option[ValueEnricher],
+) extends LedgerDao {
+
+  import JdbcLedgerDao._
+
+  private val queries = dbType match {
+    case DbType.Postgres => PostgresQueries
+    case DbType.H2Database => H2DatabaseQueries
+  }
+
+  private val logger = ContextualizedLogger.get(this.getClass)
+
+  override def currentHealth(): HealthStatus = dbDispatcher.currentHealth()
+
+  override def lookupLedgerId()(implicit loggingContext: LoggingContext): Future[Option[LedgerId]] =
+    dbDispatcher.executeSql(metrics.daml.index.db.getLedgerId)(ParametersTable.getLedgerId)
+
+  override def lookupParticipantId()(implicit
+      loggingContext: LoggingContext
+  ): Future[Option[ParticipantId]] =
+    dbDispatcher.executeSql(metrics.daml.index.db.getParticipantId)(
+      ParametersTable.getParticipantId
+    )
+
+  /** Defaults to Offset.begin if ledger_end is unset
+    */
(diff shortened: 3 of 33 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsReader.scala
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao
```
mziolekda, line 4: Nit-pick, so far we have separated words in the class paths with dots i.e.
`com.daml.platform.store.append.only.dao`

nmarton-da: valid point, I was also considering that, my arguments to keep it like this:
- simplicity in general, only one dir
- the old and new code is truely next to each other
- with access modifiers/current package references it would be harder to rename/refactor eventually
- this is ephemeral package: after cleanup we will end up with the same package structure

So therefore if you would agree, I propose to keep it simple on cost of less conforming to code-style principles

nmarton-da:
Dear reviewers, I would recommend to do the review by reviewing the diff between the forked code and the original (in IntelliIDEA you can do this by selecting both directories - right click - compare directories).

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -0,0 +1,897 @@
...
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.metrics.{Metrics, Timed}
+import com.daml.platform.configuration.ServerRole
+import com.daml.platform.indexer.OffsetStep
+import com.daml.platform.store.Conversions._
+import com.daml.platform.store.SimpleSqlAsVectorOf.SimpleSqlAsVectorOf
+import com.daml.platform.store._
+import com.daml.platform.store.appendonlydao.CommandCompletionsTable.prepareCompletionsDelete
+import com.daml.platform.store.appendonlydao.events._
+import com.daml.platform.store.dao.events.EventsTableDelete
```
mziolekda, line 42: this should probably be `appendonlydao`. There is a version of that class in the new hirerachy

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/MeteredLedgerDao.scala
```diff
@@ -0,0 +1,312 @@
...
+  override def completeTransaction(
+      submitterInfo: Option[SubmitterInfo],
+      transactionId: TransactionId,
+      recordTime: Instant,
+      offsetStep: OffsetStep,
+  )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
+    ledgerDao.completeTransaction(submitterInfo, transactionId, recordTime, offsetStep)
+
+  // TODO append-only cleanup
+  override def storeTransactionState(preparedInsert: PreparedInsert)(implicit
```
mziolekda, line 308: For ease of analysis, I would move this method to where the original was located.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -0,0 +1,897 @@
...
+    new CommandCompletionsReader(dbDispatcher, dbType, metrics, servicesExecutionContext)
+
+  private val postCommitValidation =
+    if (performPostCommitValidation)
+      new PostCommitValidation.BackedBy(contractsReader.committedContracts, validatePartyAllocation)
+    else
+      PostCommitValidation.Skip
+
+  // TODO append-only: cleanup
+  override def storeTransactionState(
```
mziolekda, line 646: For ease of analysis, I would move this method to where the original was located.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -0,0 +1,897 @@
...
+        |set referential_integrity true;
+      """.stripMargin
+
+    /** H2 does not support asynchronous commits */
+    override protected[JdbcLedgerDao] def enforceSynchronousCommit(implicit
+        conn: Connection
+    ): Unit = ()
+  }
+
+  def deduplicationKey(
```
mziolekda, line 883: For ease of analysis, I would move this method to where the original was located.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/MeteredLedgerDao.scala
```diff
@@ -0,0 +1,312 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
mziolekda, line 1: Do we need to duplicate this class? I am not convinced.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/CompressionMetrics.scala
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao.events
+
+import com.codahale.metrics.Histogram
+import com.daml.metrics.Metrics
+
+object CompressionMetrics {
```
mziolekda, line 9: Is that a file that was modified in parallel to this PR? It looks very different to its original in `dao`

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsReader.scala
gerolf-da: I agree with Márton too keep it simple. How about removing the `dao` part, it's the append-only store and not the append-only-dao store :)

nmarton-da: @gerolf-da it's a bit different: not the store is getting forked, but rather the store.dao, so basically the original is the dao package, and the forked is the appendonlydao. But if you still think appendonly is enough, I have no strong opinion...anyway this is just a working title :)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: sharp eye! changed

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsReader.scala
gerolf-da: yeah, I actually made the comment 5 seconds before I started my own review. appendonlydao works for me. let's rather continue with the juicy bits and keep the bikeshedding for later :)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/package.scala
```diff
@@ -0,0 +1,115 @@
...
+import com.daml.lf.transaction.Node.KeyWithMaintainers
+
+/** Type aliases used throughout the package
+  */
+package object events {
+
+  type SqlSequence[A] = SqlSequence.T[A]
+
+  import com.daml.lf.value.{Value => lfval}
+  type ContractId = lfval.ContractId
```
mziolekda, line 17: Any particular reason why these members are no longer `private[events]`

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/MeteredLedgerDao.scala
nmarton-da: Good point, it is identical. Although there are multiple things here which is questionable regarding duplication, but I deliberately did not wanted to move in that territory much at the moment, so we have potentially easier hand later on.
TLDR: I would limit "reuse" or most importantly dependency between forked and original code.
Are you okay to draw the line here, and keep the metered dao duplicated?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
```diff
@@ -0,0 +1,122 @@
...
+         (SELECT * FROM divulged_contract)
+       )
+  SELECT ledger_effective_time
+    FROM create_and_divulged_contracts
+   WHERE NOT EXISTS (SELECT 1 FROM archival_event)
+   LIMIT 1;
+               """.as(instant("ledger_effective_time").?.singleOpt)
+
+      val foundIds: List[Option[Instant]] = ids.toList
+        .map(lookup)
```
gerolf-da, line 90: Could you help me understand the reason for switching from _1 query for all contract ids_ over to _1 query per contract id_ to lookup the ledger time? Is there a threshold of number of contract ids where the overhead of the DB round trip becomes cheaper than the inefficiency of a single query due to a higher number of contracts to lookup?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/CompressionMetrics.scala
nmarton-da: Not only this, but also the CompressionStrategy is quite different. As part of the PoC I had to rebase at some point, and I faced all changes regarding compression. There I wanted to have compression easily used in the new PoC indexer, and therefore I made refactorings to the original compression supporting code. I think the refactoring makes sense, and I would go ahead with that if that's fine with you.
Something else to note: the store.serialisation.Compression code was not changing much during my Compression refactorings, so I kept that around (was not forking that as well).
On that account: I forgot to add there a TODO to cleanup that helper method which is not needed any more...just did that as well.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/package.scala
nmarton-da: yes, these type helpers are quite useful for the new indexer code as well, since they simplify greatly complicated types.
In order to be able to use them outside of this package I just simply made them public.
As for me I wouldn't have obligations to let them stay public, but if we want to be more strict in hiding stuff, I am happy to add here a TODO to revisit visibility here.
What do you think, should I add the TODO?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
```diff
@@ -0,0 +1,122 @@
...
+         UNION ALL
+         (SELECT * FROM divulged_contract)
+       )
+  SELECT ledger_effective_time
+    FROM create_and_divulged_contracts
+   WHERE NOT EXISTS (SELECT 1 FROM archival_event)
+   LIMIT 1;
+               """.as(instant("ledger_effective_time").?.singleOpt)
+
+      val foundIds: List[Option[Instant]] = ids.toList
```
gerolf-da, line 89: `foundIds` contains ledger effective times and not ids. 
```suggestion
      val foundLedgerEffectiveTimes: List[Option[Instant]] = ids.toList
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: In this case it is a bit different: that method jumped from the class to the object, so it can be reused from within new ingestion code.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -0,0 +1,503 @@
...
+            from participant_events as active_cs
+            where active_cs.event_kind = 10 -- create
+                  and active_cs.event_sequential_id > ${range.startExclusive._2: Long}
+                  and active_cs.event_sequential_id <= ${range.endInclusive._2: Long}
+                  and not exists (
+                    select 1
+                    from participant_events as archived_cs
+                    where
+                      archived_cs.contract_id = active_cs.contract_id and
+                      archived_cs.event_kind = 20 and -- consuming and
```
gerolf-da, line 380:
```suggestion
                      archived_cs.event_kind = 20 and -- consuming
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/MeteredLedgerDao.scala
mziolekda: In case of this particular class it is more difficult to keep a duplicate than reuse. It is implemented using a decorator pattern so it will ALWAYS follow the interface defined in LedgerDao and no more or less.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/CompressionMetrics.scala
mziolekda: Thanks for the explanation. It makes sense

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/package.scala
mziolekda: TODO sounds fine

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
nmarton-da: Good observation: this is a totally inefficient implementation, the reason for switching to this is nothing else than doing it over the append only schema in one SQL is not trivial. I had a TODO here stating this, but removed it after discussing with Simon (I don't exactly recall why...maybe because mutable caching will change this?)
Actually let me re-add the TODO so it is getting considered again.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
```diff
@@ -0,0 +1,122 @@
...
+        .map(lookup)
+        .collect { case Some(found) =>
+          found
+        }
+      val result =
+        if (foundIds.size != ids.size) Failure(ContractsTable.notFound(ids))
```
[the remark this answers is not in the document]
nmarton-da: This is merely following the semantics of the original implementation. The error message also states: `One or more of the following contract identifiers has been found:` not sure we want to change this now.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
```diff
@@ -0,0 +1,122 @@
...
+        else {
+          val ledgerTimes = foundIds.collect { case Some(ledgerEffectiveTime) =>
+            ledgerEffectiveTime
+          }
+          val optionalMax: Option[Instant] =
+            if (ledgerTimes.isEmpty) None else Some(ledgerTimes.max)
+          Success(optionalMax)
```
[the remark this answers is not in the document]
nmarton-da: Yeah I agree that code sucks :( but unfortunately it is not enough: even if we have found id, it can have no ledgerEffectiveTime, so we cannot know that ledgerTimes will be empty or not.
These are the occasions where I miss scala 2.13 a lot :(

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
nmarton-da: good idea, but this picture of the optionally found optional ledger effective time is a though for nomenclature. Wouldnt be missleading if we say here foundLedgerEffectiveTimes, when in fact those could be none? Also precise naming would be very chatty here

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -0,0 +1,503 @@
...
+      sqlFunctions: SqlFunctions,
+  ) extends EventsTableFlatEventsRangeQueries[EventsRange[(Offset, Long)]] {
+
+    override protected def singleWildcardParty(
+        range: EventsRange[(Offset, Long)],
+        party: Party,
+        pageSize: Int,
+    ): QueryParts = {
+      def witnessesWhereClause(prefix: String) =
+        sqlFunctions.arrayIntersectionWhereClause(s"$prefix.flat_event_witnesses", party)
```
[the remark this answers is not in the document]
nmarton-da: no, good point, it is a leftover after refactorings probably...inlining it to a val

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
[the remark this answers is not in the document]
gerolf-da: Ugh, the error message is also missing the negation. I think it's reasonable to fix this.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -0,0 +1,503 @@
...
+            from participant_events as active_cs
+            where active_cs.event_kind = 10 -- create
+                  and active_cs.event_sequential_id > ${range.startExclusive._2: Long}
+                  and active_cs.event_sequential_id <= ${range.endInclusive._2: Long}
+                  and not exists (
+                    select 1
+                    from participant_events as archived_cs
+                    where
+                      archived_cs.contract_id = active_cs.contract_id and
+                      archived_cs.event_kind = 20 and -- consuming and
```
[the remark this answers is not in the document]
nmarton-da: yeah I the `, wait for it` was missing :D
thanks! removed

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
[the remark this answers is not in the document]
gerolf-da: Not sure I'm following. `foundIds` is of type `List[Option[Instant]]`. The `Ordered` instance for `Option[Instant]` makes it so that `List(None).max` returns `None` and `List(None, Some(instant)).max` returns `Some(instant)`. This matches the type of `optionalMax: Option[Instant]`.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTableFlatEventsRangeQueries.scala
```diff
@@ -0,0 +1,503 @@
...
+        formatPartiesAndTemplatesWhereClause(
+          sqlFunctions,
+          "active_cs.flat_event_witnesses",
+          partiesAndTemplateIds,
+        )
+      val filteredWitnesses =
+        sqlFunctions.arrayIntersectionValues("active_cs.flat_event_witnesses", parties)
+      val submittersInPartiesClause =
+        sqlFunctions.arrayIntersectionWhereClause("active_cs.submitters", parties)
+      // TODO these repetitions are painful to maintain. Ideas: Adding SQL views (do reuse in SQL schema), pursue anorm to enable reuse, using something else then anorm here.
```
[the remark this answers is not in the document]
nmarton-da: yepp, that is the current idea that we will have view-s

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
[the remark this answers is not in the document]
nmarton-da: ah I completely missed you point, you are correct, great idea, fixed it. thanks!!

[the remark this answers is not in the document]
nmarton-da: kind of lucky that the two orderings aligned perfectly for our use case :)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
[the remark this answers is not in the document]
nmarton-da: fixed, thanks

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
nmarton-da: changed

gerolf-da (approved):
👍

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
rautenrieth-da: In the future, we should move this helper to some other place, in order to remove inter-dependencies between `JdbcLedgerDao` and the new ingestion code.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/CommandCompletionsTable.scala
```diff
@@ -0,0 +1,61 @@
...
+      applicationId: ApplicationId,
+      parties: Set[Ref.Party],
+      sqlFunctions: SqlFunctions,
+  ): SimpleSql[Row] = {
+    val submittersInPartiesClause =
+      sqlFunctions.arrayIntersectionWhereClause("submitters", parties)
+    SQL"select completion_offset, record_time, command_id, transaction_id, status_code, status_message from participant_command_completions where completion_offset > $startExclusive and completion_offset <= $endInclusive and application_id = $applicationId and #$submittersInPartiesClause order by completion_offset asc"
+  }
+
+  def prepareCompletionsDelete(endInclusive: Offset): SimpleSql[Row] =
```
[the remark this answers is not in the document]
nmarton-da: Yeah very good points! LedgerDao interface is begging for segregation/improvements, I think we have very nice ideas forming around it. Let's hope we can get back to this relatively soon, and we can make it awesome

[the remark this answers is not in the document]
nmarton-da: cc @mziolekda

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -0,0 +1,398 @@
...
+      endInclusive: Offset,
+      requestingParties: Set[Party],
+      verbose: Boolean,
+  )(implicit
+      loggingContext: LoggingContext
+  ): Source[(Offset, GetTransactionTreesResponse), NotUsed] = {
+    val span =
+      ParticipantTracer
+        .spanBuilder(
+          "com.daml.platform.store.appendonlydao.events.TransactionsReader.getTransactionTrees"
```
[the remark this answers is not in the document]
nmarton-da: yeah good point, changing it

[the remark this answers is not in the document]
nmarton-da: found another two occurrence, fixed those as well

PR #9282, opened 2021-03-30
Publish non-repudiation client libraries to Artifactory

### LINKED ISSUE #9097: Non-repudiation: publish to Artifactory
Publish the fat JAR for the app and the JAR for the client to Artifactory.

stefanobaghino-da:
changelog_begin
changelog_end

Closes #9097

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/copy-unix-release-artifacts.sh
+++ b/ci/copy-unix-release-artifacts.sh
@@ -49,4 +49,13 @@ if [[ "$NAME" == "linux" ]]; then
     NON_REPUDIATION=non-repudiation-$RELEASE_TAG.jar
     bazel build //runtime-components/non-repudiation-app:non-repudiation-app_deploy.jar
     cp bazel-bin/runtime-components/non-repudiation-app/non-repudiation-app_deploy.jar $OUTPUT_DIR/github/$NON_REPUDIATION
+
+    NON_REPUDIATION_CLIENT_JAR=non-repudiation-client-$RELEASE_TAG.jar
+    NON_REPUDIATION_CLIENT_POM=non-repudiation-client-$RELEASE_TAG.pom
+    NON_REPUDIATION_CLIENT_SRC=non-repudiation-client-$RELEASE_TAG-sources.jar
+    bazel build //runtime-components/non-repudiation-client
+    cp bazel-bin/runtime-components/non-repudiation-client/libnon-repudiation-client.jar $OUTPUT_DIR/artifactory/$NON_REPUDIATION_CLIENT_JAR
+    cp bazel-bin/runtime-components/non-repudiation-client/non-repudiation-client_pom.xml $OUTPUT_DIR/artifactory/$NON_REPUDIATION_CLIENT_POM
+    cp bazel-bin/runtime-components/non-repudiation-client/libnon-repudiation-client-src.jar $OUTPUT_DIR/artifactory/$NON_REPUDIATION_CLIENT_SRC
 fi
+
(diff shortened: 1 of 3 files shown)


ci/publish-artifactory.sh
```diff
@@ -37,6 +37,17 @@ push daml-script-runner $SCRIPT_RUNNER.asc
 push non-repudiation $NON_REPUDIATION
 push non-repudiation $NON_REPUDIATION.asc
 
+NON_REPUDIATION_CLIENT_JAR=non-repudiation-client-$RELEASE_TAG.jar
+NON_REPUDIATION_CLIENT_POM=non-repudiation-client-$RELEASE_TAG.pom
+NON_REPUDIATION_CLIENT_SRC=non-repudiation-client-$RELEASE_TAG-sources.jar
+
+push connect-ee-mvn/com/daml $NON_REPUDIATION_CLIENT_JAR
+push connect-ee-mvn/com/daml $NON_REPUDIATION_CLIENT_JAR.asc
```
stefanobaghino-da, line 45: I copied this from above, but I can't see where these files are created and I don't see them locally when I `bazel build` the relative target. Should these stay here?

cocreature: They are created here https://github.com/digital-asset/daml/blob/45c4ba2230854e048027c999d476b3b85f5393d1/azure-pipelines.yml#L92

ci/copy-unix-release-artifacts.sh
```diff
@@ -49,4 +49,13 @@ if [[ "$NAME" == "linux" ]]; then
     NON_REPUDIATION=non-repudiation-$RELEASE_TAG.jar
     bazel build //runtime-components/non-repudiation-app:non-repudiation-app_deploy.jar
     cp bazel-bin/runtime-components/non-repudiation-app/non-repudiation-app_deploy.jar $OUTPUT_DIR/github/$NON_REPUDIATION
+
+    NON_REPUDIATION_CLIENT_JAR=non-repudiation-client-$RELEASE_TAG.jar
+    NON_REPUDIATION_CLIENT_POM=non-repudiation-client-$RELEASE_TAG.pom
+    NON_REPUDIATION_CLIENT_SRC=non-repudiation-client-$RELEASE_TAG-sources.jar
```
cocreature, line 55: Should we include javadoc as well?

cocreature (approved):
Thanks!

ci/publish-artifactory.sh
stefanobaghino-da: As per @cocreature's [approval](https://github.com/digital-asset/daml/pull/9282#pullrequestreview-623962794), it appears that these are safe to stay.

ci/copy-unix-release-artifacts.sh
stefanobaghino-da: Is there a specific target to run to get it? I didn't see it in the `bazel-bin` directory for `//runtime-components/non-repudiation-client`.

stefanobaghino-da: This is the full extend of the javadoc in the library, though:

```
/**
 * A gRPC client-side interceptor that uses a key pair to sign a payload and adds it as metadata to
 * the call, alongside a fingerprint of the public key and the algorithm used to sign.
 */
 ```
 
 Maybe at this point it wouldn't make too much sense to publish it, unless it's important to have it for some reason I'm missing.

Comment (cocreature) [reply ci/copy-unix-release-artifacts.sh:55]: I know Maven mandates having it but I think artifactory doesn’t. I’d probably still publish it. It seems easier to publish javadocs everywhere than to to a per-library decision. As for why the target is not there, it looks like we need to add the following tag for this to work:
```
        "javadoc_root_packages=com.daml.nonrepudiation.client",
```

stefanobaghino-da: Mmmh, I remember trying to add it and having issues with that, I'll see if that's still the case, thanks for the tip.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9282/commits/da3cc126e64751aa86902af5bc0b2833bf1c626c

PR #9286, opened 2021-03-30
Add Oracle support in the trigger service

cocreature:
This PR migrates the ddl & queries and adds tests for this. It does
not yet expose this to users. I’ll handle that in a separate PR.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
```diff
@@ -29,7 +31,7 @@ trait OracleAround {
     // See https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm#i27570
     // for name restrictions.
     val u = "u" + Random.alphanumeric.take(29).mkString("")
-    createNewUser(u)
+    createNewUser(u.toUpperCase)
```
cocreature, line 34: oracle is case insensitive so this seems like a less confusing option.

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V1__Init.sql
```diff
@@ -0,0 +1,22 @@
+-- Copyright (c) 2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Running trigger table.
```
cocreature, line 4: We could in theory use only one migration here since nobody has any historical data on oracle. However, keeping the two in sync seems easier to understand.

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V1__Init.sql
```diff
@@ -0,0 +1,22 @@
...
+create table running_triggers(
+  -- a UUID generated by the service
+  trigger_instance nvarchar2(36) primary key,
+  -- the token corresponding to the party
+  trigger_party nvarchar2(255) not null,
+  -- the identifier for the trigger in its dalf,
+  --  of the form "packageId:moduleName:triggerName"
+  full_trigger_name nvarchar2(1000) not null
```
cocreature, line 12: 1000 is somewhat arbitrary. There is no length limit on module name and trigger name. Oracle really does not like unlimited length (there are options but they all suck). The JSON API and canton also use somewhat arbitrary limits here.

triggers/service/src/test/resources/logback.xml
```diff
@@ -11,7 +11,7 @@
     <logger name="io.grpc.netty" level="WARN">
         <appender-ref ref="stderr-appender"/>
     </logger>
-    <logger name="com.daml.lf.engine.trigger" level="DEBUG">
+    <logger name="com.daml.lf.engine.trigger" level="INFO">
```
cocreature, line 14: happy to revert this but I found it super hard to find the right places with debug logs turned on.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
```diff
@@ -296,14 +288,59 @@ final class DbTriggerDao private (dataSource: DataSource with Closeable, xa: Con
...
+    "oracle.jdbc.OracleDriver" -> ((d, xa) => new DbTriggerDaoOracle(d, xa)),
+  )
+
   def apply(c: JdbcConfig, poolSize: PoolSize = Production)(implicit
       ec: ExecutionContext
   ): DbTriggerDao = {
     implicit val cs: ContextShift[IO] = IO.contextShift(ec)
     val (ds, conn) = Connection.connect(c, poolSize)
-    new DbTriggerDao(ds, conn)
+    val driver = supportedJdbcDrivers.get(c.driver) match {
```
S11001001, line 340: Can use `supportedJdbcDrivers.getOrElse` here.

triggers/service/src/test/resources/logback.xml
S11001001: Seems fine to change

triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceFixture.scala
```diff
@@ -407,26 +408,58 @@ trait TriggerDaoPostgresFixture
...
+    with OracleAroundAll {
+  self: Suite =>
 
+  override def jdbcConfig: Option[JdbcConfig] = Some(jdbcConfig_)
+
+  // Lazy because the postgresDatabase is only available once the tests start
+  private lazy val jdbcConfig_ =
+    JdbcConfig("oracle.jdbc.OracleDriver", oracleJdbcUrl, oracleUser, oraclePwd)
+  // TODO For whatever reason we need a larger pool here, otherwise
+  // the connection deadlocks. I have no idea why :(
```
S11001001, line 446: My guess would be that culprit=flyway, maybe more connections when running on Oracle.

cocreature: Yeah it’s definitely flyway. I just don’t really get why it would use more connections against Oracle than against PostgreSQL for the same number of migrations but :shrug: doesn’t seem that bad to change it.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
cocreature: c3cbb7cfe5d582d0470ac96635b4c767ed8c3200 thanks

S11001001 (approved):

PR #9293, opened 2021-03-30
Allow to clean up more than one process listening on a port

stefanobaghino-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,7 +24,7 @@ fi
 # Note that lsof returns a non-zero exit code if there is no match.
 SANDBOX_PID="$(lsof -ti tcp:6865 || true)"
 if [ -n "$SANDBOX_PID" ]; then
-    kill "$SANDBOX_PID"
+    echo "$SANDBOX_PID" | xargs kill
 fi
 
 # Bazel test only builds targets that are dependencies of a test suite so do a full build first.


build.sh
```diff
@@ -24,7 +24,7 @@ fi
 # Note that lsof returns a non-zero exit code if there is no match.
 SANDBOX_PID="$(lsof -ti tcp:6865 || true)"
 if [ -n "$SANDBOX_PID" ]; then
-    kill "$SANDBOX_PID"
+    echo "$SANDBOX_PID" | xargs kill
```
cocreature, line 27: I think we can also just remove the quotes? If we get so many pids that we need xargs other functionalites we have a problem. That said, I guess there is no harm in doing that even if we never need it.

cocreature (approved):

build.sh
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9293/commits/fe45bb8e0542dcb45881d642a7c3e6273d9b6e07

build.sh
```diff
@@ -24,7 +24,7 @@ fi
 # Note that lsof returns a non-zero exit code if there is no match.
 SANDBOX_PID="$(lsof -ti tcp:6865 || true)"
 if [ -n "$SANDBOX_PID" ]; then
-    kill "$SANDBOX_PID"
+    echo $SANDBOX_PID | xargs kill
 fi
```
SamirTalwar-DA, line 28: You could potentially use `echo "$SANDBOX_PID" | xargs --no-run-if-empty kill` here instead.

stefanobaghino-da: Ah, I didn't know about it thanks. 😃

PR #9299, opened 2021-03-31
StateCache implementation for mutable contract state

tudor-da:
Caching backend implementation for mutable contract state caching
* Implemented as a wrapper around CaffeineCache
* Ensures order-deterministic update conflict resolution in face of multiple in-flight async cache updates
* Benchmarks and correctness tests

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
@@ -1,0 +1,142 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.cache
+
+import com.daml.caching.Cache
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.metrics.Metrics
+import com.daml.platform.store.cache.StateCache.PendingUpdatesTracker
+import com.daml.scalautil.Statement.discard
+
+import scala.collection.mutable
+import scala.concurrent.{ExecutionContext, Future}
+
+/** This class is a wrapper around a Caffeine cache designed to handle
+  * correct resolution of concurrent updates for the same key.
+  */
+private[platform] case class StateCache[K, V](cache: Cache[K, V], metrics: Metrics)(implicit
+    ec: ExecutionContext
+) {
+  private val logger: ContextualizedLogger = ContextualizedLogger.get(getClass)
+  private[cache] val pendingUpdates = mutable.Map.empty[K, PendingUpdatesTracker]
+
+  /** Fetch the corresponding value for an input key, if present.
+    *
+    * @param key the key to query for
+    * @return optionally [[V]]
+    */
+  def get(key: K)(implicit loggingContext: LoggingContext): Option[V] =
+    cache.getIfPresent(key) match {
+      case Some(value) =>
+        logger.debug(s"Cache hit for $key -> ${value.toString.take(100)}")
+        Some(value)
+      case None =>
+        logger.debug(s"Cache miss for $key ")
+        None
+    }
+
+  /** Update the cache synchronously.
+    *
+    * In face of multiple in-flight updates competing for the `key` (see [[putAsync()]]),
+    * this method updates the cache only if the to-be-inserted tuple is the most recent
+    * (i.e. it has `validAt` highest amongst the competing updates).
+    *
+    * @param key the key at which to update the cache
+    * @param validAt ordering discriminator for pending updates for the same key
+    * @param value the value to insert
+    */
+  def put(key: K, validAt: Long, value: V): Unit = pendingUpdates.synchronized {
+    val competingLatestForKey =
+      pendingUpdates
+        .get(key)
+        .map { pendingUpdate =>
+          val oldLatestValidAt = pendingUpdate.latestValidAt
+          pendingUpdate.latestValidAt = Math.max(validAt, pendingUpdate.latestValidAt)
+          oldLatestValidAt
+        }
+        .getOrElse(Long.MinValue)
+
+    if (competingLatestForKey < validAt) cache.put(key, value) else ()
+  }
+
+  /** Update the cache asynchronously.
+    *
+    * In face of multiple in-flight updates competing for the `key`,
+    * this method registers an async update to the cache
+    * only if the to-be-inserted tuple is the most recent
+    * (i.e. it has `validAt` highest amongst the competing updates).
+    *
+    * @param key the key at which to update the cache
+    * @param validAt ordering discriminator for pending updates for the same key
+    * @param eventualValue the eventual result signaling successful enqueuing of the cache async update
+    */
+  final def putAsync(key: K, validAt: Long, eventualValue: Future[V])(implicit
+      loggingContext: LoggingContext
+  ): Future[Unit] = pendingUpdates.synchronized {
+    val pendingUpdatesForKey = pendingUpdates.getOrElseUpdate(key, PendingUpdatesTracker.empty)
+    if (pendingUpdatesForKey.latestValidAt < validAt) {
+      pendingUpdatesForKey.latestValidAt = validAt
+      pendingUpdatesForKey.pendingCount += 1
+      registerEventualCacheUpdate(key, eventualValue, validAt)
+    } else Future.unit
+  }
+
+  private def registerEventualCacheUpdate(
+      key: K,
+      eventualUpdate: Future[V],
+      validAt: Long,
+  )(implicit loggingContext: LoggingContext): Future[Unit] =
+    eventualUpdate.transform[Unit](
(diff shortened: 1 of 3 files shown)


kamil-da (approved):
LGTM 👍

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+    * @param value the value to insert
+    */
+  def put(key: K, validAt: Long, value: V): Unit = pendingUpdates.synchronized {
+    val competingLatestForKey =
+      pendingUpdates
+        .get(key)
+        .map { pendingUpdate =>
+          val oldLatestValidAt = pendingUpdate.latestValidAt
+          pendingUpdate.latestValidAt =
+            if (pendingUpdate.latestValidAt < validAt) validAt else pendingUpdate.latestValidAt
```
meiersi-da, line 56:
```suggestion
          pendingUpdate.latestValidAt = max(oldLatestValidAt,  validAt)
```
Using https://www.scala-lang.org/api/2.12.0/scala/math/index.html#max(x:Long,y:Long):Long

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+    *
+    * In face of multiple in-flight updates competing for the `key`,
+    * this method registers an async update to the cache
+    * only if the to-be-inserted tuple is the most recent
+    * (i.e. it has `validAt` highest amongst the competing updates).
+    *
+    * @param key the key at which to update the cache
+    * @param validAt ordering discriminator for pending updates for the same key
+    * @param eventualValue the eventual result
+    */
```
meiersi-da, line 74: Missing comment on what the return value signals.

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+          removeFromPending(key)
+          logger.warn(s"Failure in pending cache update for key $key", err)
+          err
+        },
+    )
+
+  private def removeFromPending(key: K): Unit =
+    discard(pendingUpdates.synchronized {
+      val pendingUpdateForKey = pendingUpdates(key)
+      pendingUpdateForKey.pendingCount -= 1
```
meiersi-da, line 110: Question on Scala style: we'd expect the resulting of `pendingUpdates(key)` to be not `null`. Is it good style to add an assertion and proper error log in case it is `null`? That's what I'd do in my Haskell code, as I'd get an `Option` on which I have to do a complete case distinction.

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+/** This class is a wrapper around a Caffeine cache designed to handle
+  * correct resolution of concurrent updates for the same key.
+  */
+private[platform] case class StateCache[K, V](cache: Cache[K, V], metrics: Metrics)(implicit
+    ec: ExecutionContext
+) {
+  private val logger: ContextualizedLogger = ContextualizedLogger.get(getClass)
+  private[store] val pendingUpdates = mutable.Map.empty[K, PendingUpdates]
```
meiersi-da, line 22: Why is this scoped to `store` and not to `cache` where the tests reside?

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/StateCacheSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+
+    insertions.foreach { case (_, (promise, value, _)) =>
+      promise.complete(Success(value))
+    }
+
+    Await.result(Future.sequence(insertionFutures.toVector), 5.second).nonEmpty shouldBe true
+    assertCacheElements(stateCache)(insertions, `number of competing updates`)
+  }
+
+  it should "putAsync 100_000 values for the same key in 1 second" in {
```
meiersi-da, line 65: Could this test become flaky in constrained CI elements?

meiersi-da (approved):
Great work @tudor-da . Very clean and simple code.

I have a few cosmetic changes, but otherwise LGTM!

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: Ah, the simplest thing. Thanks

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+      (err: Throwable) =>
+        pendingUpdates.synchronized {
+          removeFromPending(key)
+          logger.warn(s"Failure in pending cache update for key $key", err)
+          err
+        },
+    )
+
+  private def removeFromPending(key: K): Unit =
+    discard(pendingUpdates.synchronized {
```
mziolekda, line 108: since it is a private method that is only used from synchronized sections, I would consider removing a redundant intrinsic lock.

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -0,0 +1,133 @@
...
+    * @param validAt ordering discriminator for pending updates for the same key
+    * @param eventualValue the eventual result
+    */
+  final def putAsync(key: K, validAt: Long, eventualValue: Future[V])(implicit
+      loggingContext: LoggingContext
+  ): Future[Unit] = pendingUpdates.synchronized {
+    val pendingUpdatesForKey = pendingUpdates.getOrElseUpdate(key, PendingUpdates.empty)
+    if (pendingUpdatesForKey.latestValidAt < validAt) {
+      pendingUpdatesForKey.latestValidAt = validAt
+      pendingUpdatesForKey.pendingCount += 1
```
mziolekda, line 81: I don't understand why we need to have a list of pending updates per key, if only the latest one is ever inserted.

mziolekda (commented):
Left two comments

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: Ok, good catch. I'll add a descriptive warn message instead of this hard failure.

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: Thanks. Changed

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: I guess the choice of names is confusing. `PendingUpdates` tracks concurrent updates: it doesn't have a list, but it merely counts them and the most recent `validAt` of the bunch.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/StateCacheSpec.scala
tudor-da: Yes it can but I don't think more than other timeout-sensitive tests. I've set the limit at 100K quite conservatively (on my machine I was getting 300-400K in 1 second).

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: I've left it there for because the method is not thread-safe and can cause annoying bugs if misused.
But you are right that one synchronize is redundant. I removed it from line 103

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
meiersi-da: You could use `PendingUpdatesState` as one option of a name that doesn't evoke the list connotation.

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: Added

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: I was going to go for `PendingUpdatesTracker` but `State` suffix is better

mziolekda (approved):
Looks good, thanks

PR #9303, opened 2021-03-31
Avoid base64 encoding of cache suffixes

cocreature:
base64 includes / which is at the very least pretty confusing and also
broke our cache cleanup which assumes that the cache suffix takes up
one directory. Afaict, we are not length-restricted in gcp paths so we
can just use the hex digits we get from md5

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


aherrmann-da (approved):
LGTM

PR #9304, opened 2021-03-31
Introduce parallel indexer

nmarton-da:
* Adds parallel indexer in PoC quality
* Adds relevant metrics, and wiring to the parallel indexer code
* Minor fixes to tracing in TransactionsReader

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
+++ b/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
@@ -516,6 +516,35 @@ final class Metrics(val registry: MetricRegistry) {
       val stateUpdateProcessing: Timer = registry.timer(Prefix :+ "processed_state_updates")
     }
 
+    // TODO append-only: streamline metrics upon cleanup
+    object parallelIndexer {
+      private val Prefix: MetricName = daml.Prefix :+ "parallel_indexer"
+
+      // bundle of metrics coming from instrumentation of the underlying thread-pool
+      val ingestionExecutor: MetricName = Prefix :+ "ingestion_executor"
+
+      // bundle of metrics coming from instrumentation of the underlying thread-pool
+      val inputMappingExecutor: MetricName = Prefix :+ "input_mapping_executor"
+
+      // the latency, which during an update element is residing in the mapping-stage (since batches are involved, this duration is divided by the batch size)
+      val inputMappingStageDuration: Timer =
+        registry.timer(Prefix :+ "input_mapping_stage_duration")
+
+      // the latency, which during an update element is residing in the ingestion (since batches are involved, this duration is divided by the batch size)
+      val ingestionStageDuration: Timer = registry.timer(Prefix :+ "ingestion_stage_duration")
+
+      // Throughput in #submissions measured on the output of the indexer (after the effect of the corresponding Update is persisted into the database, and before this effect is visible via moving the ledger end forward)
+      val indexerSubmissionThroughput: Counter =
+        registry.counter(Prefix :+ "indexer_submission_throughput")
+
+      // metric tracking the size of the queue before the indexer
+      val indexerInputBufferLength: Counter =
+        registry.counter(Prefix :+ "indexer_input_buffer_length")
+
+      // metric tracking the average batch size
+      val batchSize: MetricName = Prefix :+ "batch_size"
+    }
+
     object services {
       private val Prefix: MetricName = daml.Prefix :+ "services"
 
--- a/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
+++ b/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
@@ -520,29 +520,42 @@ final class Metrics(val registry: MetricRegistry) {
     object parallelIndexer {
       private val Prefix: MetricName = daml.Prefix :+ "parallel_indexer"
 
-      // bundle of metrics coming from instrumentation of the underlying thread-pool
-      val ingestionExecutor: MetricName = Prefix :+ "ingestion_executor"
+      // Number of state updates persisted to the database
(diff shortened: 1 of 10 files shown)


ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -516,6 +516,35 @@ final class Metrics(val registry: MetricRegistry) {
...
+
+      // bundle of metrics coming from instrumentation of the underlying thread-pool
+      val ingestionExecutor: MetricName = Prefix :+ "ingestion_executor"
+
+      // bundle of metrics coming from instrumentation of the underlying thread-pool
+      val inputMappingExecutor: MetricName = Prefix :+ "input_mapping_executor"
+
+      // the latency, which during an update element is residing in the mapping-stage (since batches are involved, this duration is divided by the batch size)
+      val inputMappingStageDuration: Timer =
+        registry.timer(Prefix :+ "input_mapping_stage_duration")
```
mziolekda, line 531: Separate metrics segments translate on presentation platforms to very deep trees. I would suggest removing all the words that are technology specific and don't add any semantic context.
In this case i would just say "input_mapping"

nmarton-da: I am not sure what you mean...AFAIK in grafana it is not causing any issues, at least there is no deep trees formed from underscore separated strings. Are you referring to VMBC related metrics visualisation?
Just for reference: looking around in Metrics this is not really seem to be a rule... is this a new requirement?

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PerfSupport.scala
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.indexer.parallel
+
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.Source
+
+object PerfSupport {
+
+  case class AverageCounter() {
```
mziolekda, line 11: I would consider replacing this counter with standard Histogram. There is prior art, see the metrics called `lookup_active_contract_count_per_execution`. The benefit of using the histogram is that non-integer averages can be collected, also we will get all sorts of other useful stats

mziolekda (commented):
I have left a number of comments mostly regarding metrics. 
I have also reviewed the resource allocation and akka pipelines aspect of things.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
mziolekda: Yes it is mostly about wavefront. Every time there is an underscore, it creates a branch in the metrics tree name. You quickly learn to be frugal when you interact with a system like that. We did have a go once trimming long metric names. It got a bit worse since then.

mziolekda: In wavefront this metrics will be represented as:
com.daml.parallel.indexer.input.mapping.count
com.daml.parallel.indexer.input.mapping.min
com.daml.parallel.indexer.input.mapping.mean
com.daml.parallel.indexer.input.mapping.max
com.daml.parallel.indexer.input.mapping.0.75
etc

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/AsyncSupport.scala
```diff
@@ -0,0 +1,74 @@
...
+import com.daml.metrics.MetricName
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object AsyncSupport {
+
+  trait Executor {
+    def execute[FIN, FOUT](f: FIN => FOUT): FIN => Future[FOUT]
+
+    def execute[OUT](block: => OUT): Future[OUT]
```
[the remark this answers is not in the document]
nmarton-da: well it is not a function, but a call by name param

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PerfSupport.scala
```diff
@@ -0,0 +1,49 @@
...
+      val (count, times) = retrieveAndReset
+      if (times == 0) None
+      else Some(count / times)
+    }
+
+  }
+
+  // adds a buffer to the output of the original source, and adds a Counter metric for buffer size
+  // good for detecting consumer vs producer speed
+  def instrumentedBufferedSource[T, U](
```
rautenrieth-da, line 38: Nice helper! I'm sure others working with Akka streams will appreciate this.

@oliverse-da I vaguely remember you were interested in this?

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PerfSupport.scala
rautenrieth-da: This class is only used to feed the Gauge that displays the *current* batch size (or more precisely, the average batch size since the last gauge update).

Changing to a histogram would give you more information about the distribution of batch sizes, but it would be harder to track high-frequency fluctuations in the batch sizes, as histograms use a `ExponentiallyDecayingReservoir` by default, which "produces quantiles which are representative of (roughly) the last five minutes of data".

If we are interested in seeing detailed patterns in the fluctuations of batch sizes, then we should keep the current `AverageCounter + Gauge` combo. Otherwise, I would remove `AverageCounter` and replace the batch size measurement by a histogram, optionally switching to a sliding (time) window reservoir if we are interested in capturing high-frequency changes.

mziolekda (approved):
Thanks, Smash it in

PR #9311, opened 2021-03-31
ledger-api-test-tool: Show a pleasant error message if DAR upload fails.

SamirTalwar-DA:
Previously, we would show a big scary "UNCAUGHT EXCEPTION", which was unnecessary. This is normal on connection failure.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
@@ -137,13 +137,19 @@ final class LedgerTestCasesRunner(
   ): Future[Either[Result.Failure, Result.Success]] =
     result(start(test, session))
 
-  private def uploadDar(context: ParticipantTestContext, name: String)(implicit
-      executionContext: ExecutionContext
-  ): Future[Unit] = {
+  private def uploadDar(
+      context: ParticipantTestContext,
+      name: String,
+  )(implicit executionContext: ExecutionContext): Future[Unit] = {
     logger.info(s"""Uploading DAR "$name"...""")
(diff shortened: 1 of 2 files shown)


fabiotudone-da (approved):
If it's not too expensive to write a test for that, please consider doing it.

SamirTalwar-DA:
@fabiotudone-da: Unfortunately we don't have _any_ tests for the Ledger API Test Tool. This is generally OK because we run it against a bunch of ledgers, but there are code paths like this one that will never be triggered in a normal CI run.

I am up for adding unit tests but would rather not do it in this PR, as it's going to be expensive at first. It will require a bunch of Bazel configuration, and a lot of refactoring to extract out the methods under test from the depths of `LedgerTestCasesRunner`.

SamirTalwar-DA:
(BTW, tested manually with _ledger-on-memory_ and _Ctrl+C_.)

stefanobaghino-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestCasesRunner.scala
```diff
@@ -208,10 +214,10 @@ final class LedgerTestCasesRunner(
             )(materializer, materializer.executionContext)
           } yield concurrentTestResults ++ sequentialTestResults
 
-        testResults
-          .recover { case NonFatal(e) =>
+        testResults.recover {
```
miklos-da, line 217: You may avoid having to resort to `isInstanceOf` by pattern matching on and rethrowing the `Errors.FrameworkException` first.

SamirTalwar-DA: Yeah, I tried that but it seemed uglier to me. I don't really mind either way though.

PR #3882, opened 2019-12-17
Change variant JSON encoding, so it is easier to pattern match on it in TypeScript

leo-da:
Closes: #3622, #3390

### TODO
- [x] Resolve choice type by templateID and choice name, get rid of choice recordID, related to #3390
- [x] Resolve key type by templateID
- [x] Add `contracts/lookup` test where `Key = (Party, Text)`
- [x] Add `contracts/search` test case query by `"status":{"tag":"Enabled","value":"2019-12-21T03:14:36.710Z"}`
- [x] Add `Key = (Party, Int)` JSON decoding test
- [x] Add `Key = Party` JSON decoding test
- [x] Add `Key = (Party, (Int, String, SomeRecord))` JSON decoding test

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags, if appropriate
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9321, opened 2021-04-05
use {tag, value} format for SQLizing advanced JSON queries

S11001001:
#3882 changed the JSON format of variants, but the code that produces SQL predicates from the query language wasn't updated in one respect: for advanced queries within the variant data. The easiest example is a range query, included as a test case here. (We also test simpler queries of variants, which did not have this problem.)

Prerequisite to #9278.

```rst
CHANGELOG_BEGIN
- [HTTP JSON API] Range queries within variant data would not return matching
  data when using the PostgreSQL backend with JSON API; this is fixed.
  See `issue #9321 <https://github.com/digital-asset/daml/pull/9321>`__.
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/query/ValuePredicate.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/query/ValuePredicate.scala
@@ -128,12 +128,13 @@ sealed abstract class ValuePredicate extends Product with Serializable {
 
         case VariantMatch((dc, q)) =>
           val Rec(vraw, v_==, v_@>) = go(path objectAt JsonVariant.valueKey, q)
-          // @> is safe because in a variant-typed context, all JsObjects
-          // have exactly two keys
+          // @> induction is safe because in a variant-typed context, all JsObjects
+          // have exactly two keys. @> is conjoined with raw so we use it to add
+          // the tag check
           Rec(
             vraw,
-            v_== map (jv => JsonVariant(dc, jv)),
-            v_@> map (jv => JsonVariant(dc, jv)),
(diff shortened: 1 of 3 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/query/ValuePredicate.scala
```diff
@@ -127,13 +127,14 @@ sealed abstract class ValuePredicate extends Product with Serializable {
...
+          val Rec(vraw, v_==, v_@>) = go(path objectAt JsonVariant.valueKey, q)
+          // @> induction is safe because in a variant-typed context, all JsObjects
+          // have exactly two keys. @> is conjoined with raw so we use it to add
+          // the tag check
           Rec(
             vraw,
-            v_== map (jv => JsonVariant(dc, jv)),
-            v_@> map (jv => JsonVariant(dc, jv)),
+            v_== map (JsonVariant(dc, _)),
+            v_@> map (JsonVariant(dc, _)) orElse Some(JsonVariant withoutValue dc),
```
cocreature, line 137: Do we have a test that hits this? I was expecting to see something with only a `tag` but no `value` but the tests you added do have an entry for the value.

cocreature (approved):
good catch!

ledger-service/http-json/src/main/scala/com/digitalasset/http/query/ValuePredicate.scala
S11001001: Yes, the last-added test is actually a tag-with-no-value test.

Issue #9315, opened 2021-03-31
Restore non-repudiation client libraries publishing on Artifactory


stefanobaghino-da:
The part that publishes the non-repudiation client libraries on Artifactory has been commented out of `ci/publish-artifactory.sh`. Restore it and find out why uploading either POM causes an HTTP 409 Conflict error code.

Please coordinate with @remyhaemmerle-da who's in charge of the release next Wednesday.

Temporarily assigning to @cocreature to make sure this is tackled next week, feel free to assign it to whomever. 🙂

cocreature:
@garyverhaegen-da could you please take a look at that? Happy to catch you up on what we tried so far but I think this needs someone with artifactory access to debug it efficiently.

cocreature:
fixed in #9323, thanks @garyverhaegen-da!

PR #9323, opened 2021-04-06
try to fix non-repudiation pushing

garyverhaegen-da:
I've seen reports of Artifactory returning 409 when it detects an invalid POM file, which would map cleanly to our observed behaviour (as other files do seem to upload fine). I'm not a POM expert so not entirely sure how to check the actual files, but I do see one error in the existing, commented code: the path is not a valid Maven repository path. It should be `groupid/artifactid/version`, i.e. it is currently missing the `artifactid` bit. So I'd like to try adding that.

I don't know how to test this without making a release, so my plan is to make a release once this is merged. Open to suggestion on faster ways to test this.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/publish-artifactory.sh
+++ b/ci/publish-artifactory.sh
@@ -37,38 +37,13 @@ push daml-script-runner $SCRIPT_RUNNER.asc
 push non-repudiation $NON_REPUDIATION
 push non-repudiation $NON_REPUDIATION.asc
 
-# FIXME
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
Oh probably worth mentioning this is an attempt at fixing #9315.

cocreature (approved):
Thanks! Looks very plausible

stefanobaghino-da:
Argh. Thanks!

PR #9328, opened 2021-04-06
Remove exception message from AnyException

cocreature:
This doesn’t really make that much sense in there and it makes some
things like Daml Script harder because I need to duplicate the logic
to get the exception message.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1506,7 +1506,7 @@ private[lf] object SBuiltin {
                 unwindToHandler(machine, payload) //re-throw
               case Some(handler) =>
                 payload match {
-                  case SAnyException(typ, _, value) =>
+                  case SAnyException(typ, value) =>
                     onLedger.ptx = onLedger.ptx.rollbackTry(typ, value.toValue)
                   case _ =>
                     crash(s"SBTryHandler, expected payload to be SAnyException: $payload")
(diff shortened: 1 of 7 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1562,14 +1562,23 @@ private[lf] object SBuiltin {
         machine: Machine,
     ): Unit = {
       args.get(0) match {
-        case SAnyException(_, messageFunction, innerValue) =>
-          machine.ctrl = SEApp(messageFunction, Array(SEValue(innerValue)))
+        case SAnyException(ty, innerValue) =>
+          machine.ctrl = SEApp(exceptionMessage(ty), Array(SEValue(innerValue)))
```
cocreature, line 1566: I don’t like computing this at runtime all that much. I’m not really worried about performance and it seems nicer than what we had before but if someone has a better proposal I’m all ears.

sofiafaro-da: This function should be called really infrequently, so I think it should be fine.

sofiafaro-da (approved):
This change makes sense. Thank you!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
nickchapman-da: I agree. Can't imagine there is any perf issue here.

nickchapman-da (approved):
LGTM

PR #9343, opened 2021-04-07
fix macOS cache cleaning

garyverhaegen-da:
The script needs to run once before the first build, otherwise the cache
folders get created on the main partition.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
good catch, thanks!

PR #9345, opened 2021-04-07
LF: Freeze archive proto for LF 1.13

### LINKED ISSUE #8719: Arbitrary Precision Decimals in DAML
### Context

We already provide a Numeric type but it has a limited precision which can cause issues in some applications where the precision is insufficient. In addition to that, it is limited to bankers’s rounding on inexact operations which can cause issues if that’s not the rounding mode you need to use. Arbitrary precision decimals are quite common (cf Java’s BigDecimal) type so we should also provide support for them in DAML. The new type will be unserializable so it allows for precise intermediate calculations but requires a conversion back to the limited Numeric at the end.

### DAML-LF Changes

1. Add a new primitive, unserializable type `BigDecimal`. This type corresponds to Java’s BigDecimal, i.e., it consist of an arbitrary precision, unscaled integer _value_ and a 32bit _scale_. The value of the number represented by a `BigDecimal` is `unscaledValue × 10^{-scale}`. Note that the scale is never exposed. Values that are numerically equivalent with a different scale are equivalent.
2. Add the following 4 arithmetic operations on `BigDecimal`: Add, Subtract, Multiply, Divide. Divide accepts an optional scale (if unspecified, non-terminating expansions throw) and a rounding mode.
3. Add unserializable type `RoundingMode` roundingMode` is one of the 8 rounding mode supported by Java.
4.Addition, subtraction and multiplication are always exact.
5. Add a conversion `fromNumeric: Numeric n -> BigDecimal`, this is always exact
6. Add a conversion `toNumeric : forall (n: Nat). BigDecimal -> Numeric n`
7. Add a rounding function `round : RoundingMode -> Int64 -> BigDecimal -> BigDecimal`
11. ofc we also need conversion to `Text` and ordering but I’ll omit the details here.

### DAML Changes

IMHO the DAML API has to be a bit less explicit than the DAML-LF API. Passing the math context on every addition is annoying, it forces you to use a prefix operator (which probably has to be imported qualified) and overall just seems messy. We could move the math context to the type level but that also gets rather messy.

So the API here is split into two parts:
1. The Num instances all default to arbitrary precision and no rounding.
2. We expose methods that accept a math context in a separate module. Users can still wrap BigDecimal in their own type and specialize it to a given math context if they want to.

This leads to the following changes:

1. Add a new type `data BigDecimal`.
2. Provide `Additive`, `Multiplicative`, `Signed`, `Number`, `Divisible` and `Fractional` instances. These default to being exact and throw when rounding is required?
3. `fromNumeric` is always exact`, `tuNumeric` as well.
4. RoundingMode types.
5. Add the `div` methods in `DA.BigDecimal` that accept the scale and rounding mode (non-optional, / already covers the optional case).

### Open Questions

1. Can we avoid adding MathContext and RoundingMode as primitive types in LF? Even if they are not serializable there is still a certain complexity associated with them, e.g., we might need to pattern match on it, …. It seems like we probably could use a structural record with one Int64 field for precision and one Int64 field for rounding mode. That is a bit hacky and we’ll have some runtime errors for invalid values (that will never be reachable via Daml code) but maybe still nicer, than adding two completely separate types, pattern match operations, … just for this? 

 Answer: We don’t really need pattern matching primitives. That makes the cost here relatively low so for now the plan is to make them proper types.

2. Should we expose getScale? 

Answer: We decided against this. If you look at the Java spec https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html, it leads to a whole mess around preferred scales with no clear benefit. We still expose `setScale` though.

3. Can we get away with a java < 5 API, i.e., no limits on precision, only limit to scale on div. The main usecase this API is designed for, can get away with it so that’s what we’re doing for now.

### Alternative designs

1. We consider representing the MathContext at the type-level in DAML. However, that seems potentially confusing to a lot of our users who are not that familiar with type-level operations and doesn’t provide any extra flexibility (you can implement the same API yourself on top of what we provide in daml-stdlib).
2. We also considered unifying Numeric and BigDecimal. However, that leads to a pretty complex design with two type parameters which didn’t seem worth the complexity. See https://github.com/digital-asset/daml/issues/8719#issuecomment-771877299 for an explanation of the design.

remyhaemmerle-da:
part of #8719

### Pull Request Checklist

trigger the build.


daml-lf/archive/BUILD.bazel
```diff
@@ -128,6 +128,7 @@ da_scala_test_suite(
         ":DarReaderTest.dalf",
         ":DarReaderTest.dar",
         ":daml_lf_1.11_archive_proto_srcs",
+        ":daml_lf_1.13_archive_proto_srcs",
```
cocreature, line 131: bazel is probably going to yell at you if you don’t sort ths and even if not, sorting is probably a good idea.

cocreature (approved):
Thanks!

daml-lf/archive/BUILD.bazel
remyhaemmerle-da: fixed.

PR #9351, opened 2021-04-07
infra: hoogle blue/green tf

garyverhaegen-da:
Making it a bit easier to manage rollouts.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -20,6 +20,69 @@ wrapper script that you can run to apply the latest changes:
 $ ./apply
 ```
 
+### Zero-downtime deployments
+
+Changing the template of an instance manager tends to kill all the machines in
+the current group. Therefore, we're experimenting with keeping two groups per
+logical group. (This is currently done only for the Hoogle servers.)
+
+The default state should be to split machines between the two groups evenly,
+and have the exact same configuration in each group. When a change needs to be
+made, one can set the size of one of the groups to 0, make the appropriate
+changes, then ramp up the size of the "changed" group. Once enough nodes in the
+new group have been started and are confirmed to work, the old group can be
+resized to 0 (then updated to match so we don't accidentally roll back later).
+
+For example, if we were to change the way in which the hoogle servers
+initiailize, we could:
+
+1. Change the size of the green group to 0 and bump the blue group to 3.
+2. Add a new field to both groups containing the lines we want to change, and
+   splice those lines in the init script. At this point a `terraform plan`
+   should report no change.
+3. Change the value for the green group and bump its size to 1. Deploy, wait
+   for the green machine to start.
+4. If everything went well, bump the size of the green group to 2 and zero the
+   size of the blue group.
+5. Update the variables in the blue group to match the green group, and bump
+   the size of the blue group to 1.
+
+This way, we can do a rolling upgrade while controlling how many working
+machines we have at all times, and thus perform zero-downtime upgrades with
+hopefully no impact on users. (Assuming the only failure mode of the new
+machines is to fail at startup.)
+
+At the Terraform level, this is achieved by using the special `count` field,
+which allows one to duplicate (multiplicate) a Terraform resource description.
+Within a resource that has a `count` entry, the special variable `count.index`
+will be the index of the "current" iteration of a resource. When other
+resources reference one that has a `count` property, they have to address is as
+an array. For example, if we have a resouce defined by:
+
+```terraform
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Note: this is not technically a change, as this is what I did to deploy #9344. I usually do this process by hand, but this time I did it through Terraform, so I can keep the process around to make next time much easier.

infra/hoogle_server.tf
```diff
@@ -18,16 +18,32 @@ resource "google_compute_firewall" "hoogle" {
   }
 }
 
+locals {
+  h_clusters = [
+    {
+      suffix         = "-blue",
+      ubuntu_version = "1604",
```
cocreature, line 25: should we check this in with `2004` given that we already deployed #9344 so next time we use this infrastructure, we definitely don’t want either blue or green to be 1604? Or are you keeping it so we can rollback?

infra/hoogle_server.tf
```diff
@@ -197,7 +214,7 @@ resource "google_compute_backend_service" "hoogle-http" {
   port_name     = "http"
 
   backend {
-    group = google_compute_instance_group_manager.hoogle.instance_group
+    group = google_compute_instance_group_manager.hoogle[0].instance_group
```
cocreature, line 217: How does the index work here? in `h_clusters`, index 0 is the deactivated blue deployment whereas I was expecting to see this point to green. What am I missing?

cocreature (commented):
lgtm except for the index which I don’t understand. It might be nice to have a brief readme or comment describing how this is supposed to be used to roll out changes.

infra/hoogle_server.tf
garyverhaegen-da: In the general case, I don't know what will change next time (in this case I have a strong suspicion it's going to be related to #9352). I'm not entirely decided whether we should keep this specific field at all vs. defining the fields based on the change being made. This is still a bit new to me.

infra/hoogle_server.tf
garyverhaegen-da: It's me who missed this: it should be a dynamic block like the https one. Updated.

infra/hoogle_server.tf
```diff
@@ -196,8 +213,11 @@ resource "google_compute_backend_service" "hoogle-http" {
   health_checks = [google_compute_health_check.hoogle-http.self_link]
   port_name     = "http"
 
-  backend {
-    group = google_compute_instance_group_manager.hoogle.instance_group
+  dynamic backend {
```
cocreature, line 216: What happens for a dynamic backend if both have size /= 0? Does it do round-robin or something like that?

cocreature (approved):

infra/hoogle_server.tf
garyverhaegen-da: This took a bit of research, but yes, I believe the final answer is round-robin across all machines (independent of group affiliation). At least that's my understanding of [this table](https://cloud.google.com/load-balancing/docs/features#load_balancing_methods), and the fact that we're using the EXTERNAL balancing scheme with the UTILIZATION balancing mode (with max=0.8).

So it would be round-robin amongst the machines that have CPU utilization below 0.8.

garyverhaegen-da: Also note that this is not a "dynamic backend"; the `dynamic` here is a Terraform feature. This expands to:

```
resource "google_compute_backend_service" "hoogle-http" {
  name          = "hoogle-http"
  health_checks = [google_compute_health_check.hoogle-http.self_link]
  port_name     = "http"

  backend {
    group = google_compute_instance_group_manager.hoogle[0].instance_group
  }

  backend {
    group = google_compute_instance_group_manager.hoogle[1].instance_group
  }

}
```

So from Google's perspective this is a backend service with two backends.

infra/README.md
```diff
@@ -20,6 +20,69 @@ wrapper script that you can run to apply the latest changes:
...
+  value = "the count is ${text_field.my_counter[count.index]}"
+}
+```
+
+A word of warning: because of the way in which Terraform keeps track of state,
+if you swapped two such resources, Terraform will destroy and recreate both. In
+other words, it keeps track of each indivividual element in the list, not of
+the list itself. Similarly, if you deleted an element in the list, all elements
+after it would be destroyed and recreated.
+
```
garyverhaegen-da, line 85: @cocreature does this help?

PR #9354, opened 2021-04-07
Deduplicate LfValueTranslation cache

rautenrieth-da:
This PR only includes refactoring.

It extracts and deduplicates the definition of `LfValueTranslation` caches. The definition was duplicated in `dao` and `appendonlydao`, and it was hard to pass around the cache for sharing between the read and write side.


tudor-da (approved):
Very welcomed change. One question though: why not entirely deduplicate the `LfValueTranslation` ? Currently they are identical, publicly-visible. Do you know if they're going to diverge in functionality?

rautenrieth-da:
> why not entirely deduplicate the LfValueTranslation

`LfValueTranslation` uses `com.daml.platform.store.dao.events.Raw` objects, we would have to deduplicate those as well (along with probably most of the `dao.events` package).

Note that `LfValueTranslation` objects are not shared between the read and write path, only the cache is shared and every consumer creates its own `LfValueTranslation` instance.

> Do you know if they're going to diverge in functionality?

I don't expect so. One of them is going to be deleted eventually.

PR #9355, opened 2021-04-08
Run Canton tests with JDK11

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -52,13 +52,14 @@ conformance_test(
     # Ideally these would be part of the script definition above, but that doesn't seem to work.
     extra_data = [
         ":bootstrap.canton",
-        ":canton",
+        ":canton_deploy.jar",
         ":canton.conf",
         ":logback-debug.xml",
         "@coreutils_nix//:bin/base64",
         "@curl_nix//:bin/curl",
         "@grpcurl_nix//:bin/grpcurl",
         "@jq_dev_env//:jq",
+        "@jdk11_nix//:bin/java",
     ],
     # FIXME: drop the following line once canton-test-runner supports LF 1.12
     lf_versions = ["legacy"],
@@ -98,6 +99,7 @@ conformance_test(
         "@curl_nix//:bin/curl",
         "@grpcurl_nix//:bin/grpcurl",
         "@jq_dev_env//:jq",
+        "@jdk11_nix//:bin/java",
     ],
     # FIXME: drop the following line once canton-test-runner supports LF 1.12
     lf_versions = ["legacy"],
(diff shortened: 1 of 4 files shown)


ledger/ledger-api-test-tool-on-canton/canton-test-runner.sh
```diff
@@ -6,8 +6,11 @@ set -e
 set -u
 set -o pipefail
 
+# Canton has some issues with JDK8 so we run it with JDK11.
+JAVA="$(rlocation jdk11_nix/bin/java)"
```
garyverhaegen-da, line 10: May be worth setting JAVA_HOME. I'm not sure what happens if you run a djk11 binary with JAVA_HOME pointing to jdk8. (Which might be the case here? I'm not sure who/what sets JAVA_HOME for the rest of our build.)
```
export JAVA_HOME=$(realpath $(dirname $(rlocation jdk11_nix/bin/java)/..))
```
should work I think.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -98,6 +99,7 @@ conformance_test(
         "@curl_nix//:bin/curl",
         "@grpcurl_nix//:bin/grpcurl",
         "@jq_dev_env//:jq",
+        "@jdk11_nix//:bin/java",
```
oliverse-da, line 102: on line 94 above we need `s/canton/canton_deploy.jar/` and ideally also add these two changes to the manual test below at lines 130 and 138

oliverse-da (commented):
Thank you @moritzkiefer-da for helping with this. Much appreciated 🙇 

With the proposed changes inline, all tests pass on my machine.

ledger/ledger-api-test-tool-on-canton/canton-test-runner.sh
cocreature: hm, I thought JAVA_HOME doesn’t matter if you invoke the binary directly but maybe I’m wrong.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
cocreature: good catch, fixed

oliverse-da (approved):

ledger/ledger-api-test-tool-on-canton/canton-test-runner.sh
cocreature: Looking through https://stackoverflow.com/questions/2025290/what-is-java-home-how-does-the-jvm-find-the-javac-path-stored-in-java-home, the JVM itself does not care about it at all. It’s used by other stuff to find java which doesn’t matter if we already invoke java.

garyverhaegen-da: It's also sometimes used to find bits of java other than the executable. I'm not concerned about it not being set, as the failures modes for that are well understood (non-zero exit), but I am a little bit concerned about what happens if it is set but to a different JDK than the executable you're running.

It's likely we're not using anything that uses it, though. It's mostly used by "old" programs like Java IDEs and the like.

PR #9356, opened 2021-04-08
Update canton tests to pre 0.23.0 release

oliverse-da:
- switch lf_version to stable and latest from legacy
- enable non-contract-key race condition tests in regular canton deployment (on non-unique contract key domain)
- reenable temporarily disabled semantic tests (see KVL-876 - fyi @hubert-da)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -952,7 +952,7 @@ java_import(
     jars = glob(["lib/**/*.jar"]),
 )
 """,
-    sha256 = "81e590a6854eaaf724356c0b7068502174de6457af6ec2a24624b0698eeb7362",
-    strip_prefix = "canton-community-0.21.0",
-    urls = ["https://www.canton.io/releases/canton-community-0.21.0.tar.gz"],
+    sha256 = "9fb38c922d302c94e200c161ec8467ea39f4c31c9a7fad0b32f1493170a51fa7",
+    strip_prefix = "canton-community-0.23.0-SNAPSHOT",
+    urls = ["https://www.canton.io/releases/canton-community-20210407.tar.gz"],
 )
--- a/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
+++ b/ledger/ledger-api-test-tool-on-canton/BUILD.bazel
@@ -61,9 +61,10 @@ conformance_test(
         "@jq_dev_env//:jq",
         "@jdk11_nix//:bin/java",
     ],
-    # FIXME: drop the following line once canton-test-runner supports LF 1.12
-    lf_versions = ["legacy"],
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -952,7 +952,7 @@ java_import(
     jars = glob(["lib/**/*.jar"]),
 )
 """,
-    sha256 = "81e590a6854eaaf724356c0b7068502174de6457af6ec2a24624b0698eeb7362",
-    strip_prefix = "canton-community-0.21.0",
-    urls = ["https://www.canton.io/releases/canton-community-0.21.0.tar.gz"],
+    sha256 = "f9930166a52d436cf422577d68df0097b38e4b2d3b7a90e228a5117eb8494425",
```
cocreature, line 955: let’s leave it out of this PR but this should really be in `deps.bzl` not `WORKSPACE`

cocreature (approved):
Thanks!

WORKSPACE
oliverse-da: I have taken a stab at moving this to `deps.bzl`

deps.bzl
```diff
@@ -384,3 +384,19 @@ genrule(
 )
             """,
         )
+
+    if "canton" not in native.existing_rules():
```
cocreature, line 388: lgtm, thank you!

deps.bzl
```diff
@@ -384,3 +384,19 @@ genrule(
...
+        package(default_visibility = ["//visibility:public"])
+
+        java_import(
+            name = "lib",
+            jars = glob(["lib/**/*.jar"]),
+        )
+        """,
+            sha256 = "9fb38c922d302c94e200c161ec8467ea39f4c31c9a7fad0b32f1493170a51fa7",
+            strip_prefix = "canton-community-0.23.0-SNAPSHOT",
+            urls = ["https://www.canton.io/releases/canton-community-20210407.tar.gz"],
```
cocreature, line 401: Is there a good reason why there is a mismatch between the version in the url and the prefix in the tarball? that seems a bit confusing.

cocreature (approved):

WORKSPACE
oliverse-da: Looks like this needs a bit more - leaving this as is for now.

deps.bzl
oliverse-da: This is an artifact of canton doing nightly releases instead of as-needed snapshots.

cocreature: What’s stopping you from using the date in the prefix as well?

oliverse-da: Makes sense.

PR #9363, opened 2021-04-08
Publish trigger service EE fat JARs

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/copy-unix-release-artifacts.sh
+++ b/ci/copy-unix-release-artifacts.sh
@@ -30,10 +30,11 @@ if [[ "$NAME" == "linux" ]]; then
     cp bazel-bin/ledger-service/http-json/http-json-binary_deploy.jar $OUTPUT_DIR/github/$JSON_API
 
     TRIGGER_SERVICE=trigger-service-$RELEASE_TAG.jar
+    TRIGGER_SERVICE_EE=trigger-service-$RELEASE_TAG-ee.jar
     bazel build //triggers/service:trigger-service-binary-ce_deploy.jar
     cp bazel-bin/triggers/service/trigger-service-binary-ce_deploy.jar $OUTPUT_DIR/github/$TRIGGER_SERVICE
     bazel build //triggers/service:trigger-service-binary-ee_deploy.jar
-    cp bazel-bin/triggers/service/trigger-service-binary-ee_deploy.jar $OUTPUT_DIR/artifactory/$TRIGGER_SERVICE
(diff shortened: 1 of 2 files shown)


ci/copy-unix-release-artifacts.sh
```diff
@@ -30,8 +30,10 @@ if [[ "$NAME" == "linux" ]]; then
     cp bazel-bin/ledger-service/http-json/http-json-binary_deploy.jar $OUTPUT_DIR/github/$JSON_API
 
     TRIGGER_SERVICE=trigger-service-$RELEASE_TAG.jar
-    bazel build //triggers/service:trigger-service-binary_deploy.jar
-    cp bazel-bin/triggers/service/trigger-service-binary_deploy.jar $OUTPUT_DIR/github/$TRIGGER_SERVICE
+    bazel build //triggers/service:trigger-service-binary-ce_deploy.jar
+    cp bazel-bin/triggers/service/trigger-service-binary-ce_deploy.jar $OUTPUT_DIR/github/$TRIGGER_SERVICE
+    bazel build //triggers/service:trigger-service-binary-ee_deploy.jar
+    cp bazel-bin/triggers/service/trigger-service-binary-ee_deploy.jar $OUTPUT_DIR/artifactory/$TRIGGER_SERVICE
```
cocreature, line 36: We could include `-ee` in the file name. I have a slight preference for not doing it here since we don’t do it for other fat JARs like non-repudiation but I don’t feel all that strongly about it.

garyverhaegen-da (approved):

ci/copy-unix-release-artifacts.sh
garyverhaegen-da: I have a preference for including the `ee` for the same reason we include the version number in the filename (rather than just as part of the path in the download URL): I expect people will download both and it's awkward if they have two files with the same name and different contents.

cocreature: how do you feel about including it for non-repudiation fat jar (I think I’d leave it out of the maven artifacts) as well? I think nobody has access to that so far so we can change it and we would be consistent for everything that is officially in connect (the script and trigger runner are only for DBAG).

cocreature: changed both in d1aa19993e2073eebcd8ba0258935d4cfb7c6d4d ptal

garyverhaegen-da: I'd include it everywhere. Obviously the naming conflict doesn't apply to EE-only components, but I think consistency has value here.

garyverhaegen-da: d1aa199 looks good, thanks!

garyverhaegen-da (approved):

PR #9362, opened 2021-04-08
infra/hoogle: use nix

garyverhaegen-da:
This is a demonstration, commit-by-commit, of how to use the blue/green deployment for Hoogle servers.

The actual change (using nix) is stolen from #9352.

CHANGELOG_BEGIN
CHANGELOG_END

PR #7370, opened 2020-09-10
hoogle db as tarbal

garyverhaegen-da:
We want to be able to support more than one package in our [Hoogle] instance. In order to not have to list each file individually, we assume the collection of Hoogle text files will be published as a tarball.

Note: we keep trying the existing file for now, because the deployment of this change needs to be done in separate, asynchronous steps if we want everything to keep working with no downtime:

1. We deploy the new version of the Hoogle configuration, which supports both the new and old file structure on the docs website (this PR).
2. After the next stable version (likely 1.6) is published, the docs site actually changes to the new format.
3. We can then clean-up the Hoogle configuration.

Any other sequence will require turning off Hoogle and coordinating with the docs update, which seems impractical.

[Hoogle]: https://hoogle.daml.com

CHANGELOG_BEGIN
CHANGELOG_END

PR #9364, opened 2021-04-08
fix hoogle

garyverhaegen-da:
After #9362, hoogle stopped returning any result. It was up and running, but with an empty database.

Problem was two-fold:
1. In the switch to `nix`, I lost track of the fact that we were previously doing all the work in `~/hoogle` rather than `~/`, which broke the periodic refresh.
2. The initial setup has been broken for months; machines have been initializing with an empty DB and only getting data on the first refresh since #7370.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/hoogle_server.tf
+++ b/infra/hoogle_server.tf
@@ -115,34 +115,26 @@ NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/c50e680b03adecae01fdd1
 HOOGLE_PATH=$(nix-build --no-out-link -E '(import <nixpkgs> {}).haskellPackages.hoogle')
 mkdir -p /home/hoogle/.local/bin
 ln -s $HOOGLE_PATH/bin/hoogle /home/hoogle/.local/bin/hoogle
-export PATH=/home/hoogle/.local/bin:$PATH
-mkdir daml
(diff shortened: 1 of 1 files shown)


infra/hoogle_server.tf
```diff
@@ -115,45 +115,40 @@ NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/c50e680b03adecae01fdd1
 HOOGLE_PATH=$(nix-build --no-out-link -E '(import <nixpkgs> {}).haskellPackages.hoogle')
 mkdir -p /home/hoogle/.local/bin
 ln -s $HOOGLE_PATH/bin/hoogle /home/hoogle/.local/bin/hoogle
-export PATH=/home/hoogle/.local/bin:$PATH
```
cocreature, line 118: So the idea is that we just give up on initialization and only do the diff thing? Seems sensible

cocreature (approved):

infra/hoogle_server.tf
garyverhaegen-da: I see it more as using the same script for init and cron, that way they don't get out of sync as they did in #7370. It does rely on the fact that `diff a b` is just as non-zero when `a` and `b` differ as when one doesn't exist.

PR #9366, opened 2021-04-08
Patch hoogle binary to include bugfix

cocreature:
This includes https://github.com/ndmitchell/hoogle/pull/367.

As usual, I unfortunately cannot test this myself so please review
carefully.

Note that this will slightly increase compile times since we will now
build hoogle. However, we still only build hoogle rather than
everything which takes less than 2min on my very weak personal
laptop. We could integrate this with our nix cache but for now, I’m
not that worried about this.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/infra/hoogle_server.tf
+++ b/infra/hoogle_server.tf
@@ -112,7 +112,21 @@ curl -sSfL https://nixos.org/nix/install | sh
 # Feel free to bump the commit, this was the latest
 # # at the time of creation.
 NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/c50e680b03adecae01fdd1ea4e44c82e641de0cf.tar.gz
-HOOGLE_PATH=$(nix-build --no-out-link -E '(import <nixpkgs> {}).haskellPackages.hoogle')
+cat << EOF > /home/hoogle/hoogle_overlay.nix
+self: super:
+{
+  haskellPackages = super.haskellPackages.override {
+    overrides = haskellSelf: haskellSuper: {
+      hoogle = super.haskell.lib.appendPatch haskellSuper.hoogle
+        (super.fetchurl {
+          url = "https://patch-diff.githubusercontent.com/raw/ndmitchell/hoogle/pull/367.patch";
+          sha256 = "1p0xdnfjicl5zp6g0fkqjk9mgm6fqzl7sz0v5m51chzd7lwx181y";
+        });
+    };
+  };
+}
+EOF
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):

garyverhaegen-da:
Patch already applied. Let's get this merged so main matches reality again.

garyverhaegen-da:
FYI startup time is now at 5:10 for the one machine I looked at.

infra/hoogle_server.tf
```diff
@@ -112,7 +112,21 @@ curl -sSfL https://nixos.org/nix/install | sh
...
+      hoogle = super.haskell.lib.appendPatch haskellSuper.hoogle
+        (super.fetchurl {
+          url = "https://patch-diff.githubusercontent.com/raw/ndmitchell/hoogle/pull/367.patch";
+          sha256 = "1p0xdnfjicl5zp6g0fkqjk9mgm6fqzl7sz0v5m51chzd7lwx181y";
+        });
+    };
+  };
+}
+EOF
+HOOGLE_PATH=$(nix-build --no-out-link -E '((import /home/hoogle/hoogle_overlay.nix) (import <nixpkgs> {})).haskellPackages.hoogle')
```
cocreature, line 129:
```suggestion
HOOGLE_PATH=$(nix-build --no-out-link -E 'import <nixpkgs> { overlays = (import /home/hoogle/hoogle_overlay.nix)}.haskellPackages.hoogle')
```
Not important to include this but this was what I was trying to do.

infra/hoogle_server.tf
```diff
@@ -112,7 +112,21 @@ curl -sSfL https://nixos.org/nix/install | sh
 # Feel free to bump the commit, this was the latest
 # # at the time of creation.
 NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/c50e680b03adecae01fdd1ea4e44c82e641de0cf.tar.gz
-HOOGLE_PATH=$(nix-build --no-out-link -E '(import <nixpkgs> {}).haskellPackages.hoogle')
+cat << EOF > /home/hoogle/hoogle_overlay.nix
+super:
```
cocreature, line 116:
```suggestion
self: super:
```

infra/hoogle_server.tf
garyverhaegen-da: There still seem to be a couple syntactic issues with this version, but I see where you were going. I wasn't aware of the `overlays` option to `<nixpkgs>`. Locally at least, running
```nix
nix-build --no-out-link -E '(import <nixpkgs> { overlays = [(import ./overlay.nix)];}).haskellPackages.hoogle'
```
(extra `;` and `()`) yields the same derivation as the version in production, so I don't think it's worth going through another deployment to amend, as I can't imagine needing another overlay on these servers.

PR #8808, opened 2021-02-10
LF: first draft of the exception internal API for PartialTransaction

remyhaemmerle-da:
This naive API will allow Speedy to interact with the transaction builder when adding support of exception. 

This is part of #8020

CHANGELOG_BEGIN
CHANEGLOG_END

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9367, opened 2021-04-09
Create rollback nodes

nickchapman-da:
Create rollback nodes in `PartialTransaction`
- and test in `RollbackTest`

This PR contains an implementation for `rollbackTry` which creates the rollback node, but more interesting is the behaviour of `endTry/abortTry` which manipulate the `BackStack` of children in ways unlike existing code. Specifically, two backstacks have to be merged if we leave a try context without creating a rollback node. Does this seem correct?

@remyhaemmerle-da 
Please look carefully at what I have so far, because I'm sure there is stuff missing. For example, the handling of seeds.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -18,7 +18,6 @@ import com.daml.lf.transaction.{
   TransactionVersion => TxVersion,
   Transaction => Tx,
 }
-import com.daml.lf.transaction.TransactionVersion.VDev
 import com.daml.lf.value.Value
 import com.github.ghik.silencer.silent
 
@@ -31,20 +30,20 @@ private[lf] object PartialTransaction {
   type LeafNode = Node.LeafOnlyNode[Value.ContractId]
 
   sealed abstract class ContextInfo {
-    val childSeed: Int => crypto.Hash
+    val actionChildSeed: Int => crypto.Hash
   }
 
   sealed abstract class RootContextInfo extends ContextInfo
 
   private[PartialTransaction] final class SeededTransactionRootContext(seed: crypto.Hash)
       extends RootContextInfo {
-    val childSeed = crypto.Hash.deriveNodeSeed(seed, _)
+    val actionChildSeed = crypto.Hash.deriveNodeSeed(seed, _)
   }
 
   private[PartialTransaction] final class SeededPartialTransactionRootContext(
       seeds: ImmArray[Option[crypto.Hash]]
   ) extends RootContextInfo {
-    override val childSeed: Int => crypto.Hash = { idx =>
+    override val actionChildSeed: Int => crypto.Hash = { idx =>
       seeds.get(idx) match {
         case Some(Some(value)) =>
           value
@@ -55,7 +54,7 @@ private[lf] object PartialTransaction {
   }
 
   private[PartialTransaction] object NoneSeededTransactionRootContext extends RootContextInfo {
-    val childSeed: Any => Nothing = { _ =>
+    val actionChildSeed: Any => Nothing = { _ =>
       throw new IllegalStateException(s"the machine is not configure to create transaction")
     }
   }
@@ -63,10 +62,17 @@ private[lf] object PartialTransaction {
   /** Contexts of the transaction graph builder, which we use to record
     * the sub-transaction structure due to 'exercises' statements.
     */
-  final case class Context(info: ContextInfo, children: BackStack[NodeId], nextChildIdx: Int) {
-    def addChild(child: NodeId): Context = Context(info, children :+ child, nextChildIdx + 1)
+  final case class Context(
+      info: ContextInfo,
+      children: BackStack[NodeId],
+      nextActionChildIdx: Int,
+  ) {
+    def addActionChild(child: NodeId): Context =
+      Context(info, children :+ child, nextActionChildIdx + 1)
+    def addRollbackChild(child: NodeId, nextActionChildIdx: Int): Context =
+      Context(info, children :+ child, nextActionChildIdx)
     // This function may be costly, it must be call at most once for each node.
-    def nextChildSeed: crypto.Hash = info.childSeed(nextChildIdx)
+    def nextActionChildSeed: crypto.Hash = info.actionChildSeed(nextActionChildIdx)
   }
 
   object Context {
@@ -119,12 +125,12 @@ private[lf] object PartialTransaction {
       parent: Context,
       byKey: Boolean,
   ) extends ContextInfo {
-    val nodeSeed = parent.nextChildSeed
-    val childSeed = crypto.Hash.deriveNodeSeed(nodeSeed, _)
+    val actionNodeSeed = parent.nextActionChildSeed
+    val actionChildSeed = crypto.Hash.deriveNodeSeed(actionNodeSeed, _)
   }
 
   final case class TryContextInfo(nodeId: NodeId, parent: Context) extends ContextInfo {
-    val childSeed: NodeIdx => crypto.Hash = parent.info.childSeed
+    val actionChildSeed: NodeIdx => crypto.Hash = parent.info.actionChildSeed
   }
 
   def initial(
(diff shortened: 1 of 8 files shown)


remy:
Yeah, you don't mean @remy :)

nickchapman-da:
> Yeah, you don't mean @remy :)

oops, sorry!
I think I've done that before.
I mean:
@remyhaemmerle-da

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -525,8 +536,19 @@ private[lf] case class PartialTransaction(
...
+        val version: TxVersion = VDev
+        val rollbackNode = Node.NodeRollback(
+          context.children.toImmArray,
+          version,
+        )
+        val nodeId = info.nodeId
+        copy(
+          context = info.parent.addChild(nodeId),
+          nodes = nodes.updated(nodeId, rollbackNode),
+          //nodeSeeds = nodeSeeds :+ (nodeId -> info.nodeSeed), //TODO 8020, need this?
```
remyhaemmerle-da, line 550:
```suggestion
          nodeSeeds = nodeSeeds.update(nodeId, info.nodeSeed),
```
Lets be simple and add everything naively.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -308,8 +308,8 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
   final def fold[A](z: A)(f: (A, (Nid, Node.GenNode[Nid, Cid])) => A): A = {
     var acc = z
     foreach { (nodeId, node) =>
-      // make sure to not tie the knot by mistake by evaluating early
-      val acc2 = acc
+      // make sure to not tie the knot by mistake by evaluating early //TODO: ???
+      val acc2 = acc //TODO: why is this necessary?
       acc = f(acc2, (nodeId, node))
```
remyhaemmerle-da, line 313: @stephencompall-DA could you comment on that ?

S11001001: If `f` was lazy, then `acc = f(acc, ...)` is circular. Whether you wish to keep defending against that (I assume copy-paste or signature change) is your choice; it's harmless, anyhow.

S11001001: By the way, there actually was such a bug, fixed in this way. I don't remember where, though.

remyhaemmerle-da:
cc @oliverse-da @oggy-

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
```diff
@@ -194,43 +195,60 @@ class ExceptionTest extends AnyWordSpec with Matchers with TableDrivenPropertyCh
...
-    ("exer2", List(100, 300)),
+    ("create1", List(C(100))),
+    ("create2", List(C(100), C(200))),
+    ("create3", List(C(100), C(200), C(300))),
+    ("create3nested", List(C(100), C(200), C(300))),
+    ("create3catchNoThrow", List(C(100), C(200), C(300))),
+    ("create3throwAndCatch", List[Tree](R(List(C(100), C(200))), C(300))),
+    ("create3throwAndOuterCatch", List[Tree](R(List(C(100), C(200))), C(300))),
+    ("exer1", List[Tree](C(100), X(List(C(400), C(500))), C(200), C(300))),
+    ("exer2", List[Tree](C(100), R(List(X(List(C(400))))), C(300))), // TODO: Is the X correct?
```
cocreature, line 209: What are you unsure about here? I’m probably missing something but it looks fine to me.

remyhaemmerle-da: @nickchapman-da is unsure. 
I am not sure about what it is unsure. 
I will review this test again.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -308,8 +308,8 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
   final def fold[A](z: A)(f: (A, (Nid, Node.GenNode[Nid, Cid])) => A): A = {
     var acc = z
     foreach { (nodeId, node) =>
-      // make sure to not tie the knot by mistake by evaluating early
-      val acc2 = acc
+      // make sure to not tie the knot by mistake by evaluating early //TODO: ???
+      val acc2 = acc //TODO: why is this necessary?
```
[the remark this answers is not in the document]
remyhaemmerle-da: :+1:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -54,18 +54,25 @@ private[lf] object PartialTransaction {
...
-    def addChild(child: NodeId): Context = Context(info, children :+ child, nextChildIdx + 1)
+  final case class Context(
+      info: ContextInfo,
+      children: BackStack[NodeId],
+      nextActionChildIdx: Int,
+  ) {
+    def addActionChild(child: NodeId): Context =
+      Context(info, children :+ child, nextActionChildIdx + 1)
+    def addRollbackChild(child: NodeId, nextActionChildIdx: Int): Context =
+      Context(info, children :+ child, nextActionChildIdx)
```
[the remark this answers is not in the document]
remyhaemmerle-da: We decided some time ago,  that the rollback node will not interact with the contract ID scheme, the contract ID scheme will consider only Action Node. This was the purpose of #8808.
So since the rollback node is not an action we do not increment the `nextActionChildIdx`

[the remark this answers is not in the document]
cocreature: I think I’m still slightly confused. The rollback node itself should not affect the CID scheme but children of that node which we’re adding here do don’t they? If they don’t, I would expect CID collisions between contracts created in rollback nodes and those created outside which we don’t want.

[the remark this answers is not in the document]
cocreature: Oh I see what you’re saying. I got slightly confused by the naming. `addRollbackChild` does not add a child to a rollback node. It adds a rollback node as the child of whatever is the parent. makes sense :+1:

cocreature (approved):
Thanks! Nice work @nickchapman-da and @remyhaemmerle-da

cocreature:
@remyhaemmerle-da This is currently failing in the damlc integration tests. I recommend to simply disable those in this PR and merge this and then we can parallelize on fixing the scenario service to reenable those tests (happy to take that myself) and fixing other stuff.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/RollbackTest.scala
remyhaemmerle-da: Looks fine for me.

remyhaemmerle-da:
@nickchapman-da  wrote:
> @remyhaemmerle-da
> Please look carefully at what I have so far, because I'm sure there is stuff missing. For example, the handling of seeds.

I fixed the seeds and the contract ID scheme. Please review how Rollback node does not impact for contract ID seeding.

remyhaemmerle-da:
part of #8020

moritzkiefer-da:
Simplest fix is probably to simply add a `()` argument to the scenarios to make sure they are not run. That way we still test compilation.

remyhaemmerle-da:
@sofiafaro-da please note that I deactivated some tests as @moritzkiefer-da suggested.

cocreature:
s/deactivated/switched from testing compilation + scenario execution to only testing compilation/
that sounds a bit less scary :) 

With https://github.com/digital-asset/daml/pull/9416/files and https://github.com/digital-asset/daml/pull/9410 we should be able to reenable scenario execution.

PR #9370, opened 2021-04-09
LF: fix contract ID freshness check

remyhaemmerle-da:
Local contract ids are collected in exercise children during
preprocessing transaction for replay.

CHANGELOG_BEGIN
- [Engine] Fix contract ID fresness check when validating transaction

CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
       tx: GenTransaction[NodeId, Cid]
   ): (ImmArray[speedy.Command], Set[ContractId]) = {
 
-    type Acc = ((Set[Value.ContractId], Set[Value.ContractId]), BackStack[speedy.Command])
+    type Acc = (Set[Value.ContractId], Set[Value.ContractId], BackStack[speedy.Command])
```
cocreature, line 64: maybe a good idea turning this into a case class? It’s a bit hard to see what the role of the two cid sets is.

cocreature: Or alternatively, add a comment.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
...
 
-    val ((localCids, _), cmds) =
-      tx.roots.foldLeft[Acc](((Set.empty, Set.empty), BackStack.empty)) {
-        case ((cids, stack), id) =>
-          tx.nodes.get(id) match {
+    val (_, globalCids, cmds) =
+      tx.roots.foldLeft[Acc]((Set.empty, Set.empty, BackStack.empty)) {
+        case ((localCids0, globalCids0, stack), id) =>
+          val (newGlobals, newLocals, cmd) = tx.nodes.get(id) match {
```
cocreature, line 69: I recommend keeping the order of global and local consistent, so use locals first here as well.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
...
+                case exe: Node.NodeExercises[_, Cid] =>
+                  val templateId = exe.templateId
+                  val (cmd, newCids) =
+                    commandPreprocessor.unsafePreprocessExercise(
+                      templateId,
+                      exe.targetCoid,
+                      exe.choiceId,
+                      exe.chosenValue,
+                    )
+                  val newGlobalCids = newCids.filterNot(localCids0)
```
cocreature, line 88: Instead of filtering at each step, can we first accumulate everything and then drop the local cids from the global cids at the end? That seems easier to get right.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
...
               fail(s"invalid transaction, root refers to non-existing node $id")
             case Some(node) =>
               node match {
+                case create: Node.NodeCreate[Cid] =>
+                  val (cmd, newCids) =
+                    commandPreprocessor.unsafePreprocessCreate(create.templateId, create.arg)
+                  val newGlobalCids = newCids.filterNot(localCids0)
+                  (newGlobalCids, List(create.coid), cmd)
+                case exe: Node.NodeExercises[_, Cid] =>
+                  val templateId = exe.templateId
```
cocreature, line 80: nit: seems odd, that this is the only thing that you pull out into a local variable and then you still only use it once.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
...
-                  (acc, stack :+ cmd)
+                case _: Node.NodeRollback[_] =>
+                  // TODO https://github.com/digital-asset/daml/issues/8020
+                  // how on earth can we turn a rollback node back into a speedy command?
+                  sys.error("rollback nodes are not supported")
               }
           }
+          val globalCids = globalCids0 | newGlobals
+          if (newLocals.exists(globalCids))
+            fail("Conflicting discriminators between a global and local contract ID.")
```
cocreature, line 104: Can we get a testcase for this? This seems like a related but not the same bug. I think you can trigger it on a transaction like the following:

1. First do an exercise on a global cid.
2. Then do a create which produces the same discriminator as the global cid.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
cocreature: nvm, the added check below wouldn’t work in that case.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
cocreature: Missed that this just got shuffled around but I still don’t see an existing test for it so it would be nice to add it.

cocreature (approved):
LGTM, thanks for fixing this so quickly!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
remyhaemmerle-da: :+1:

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
remyhaemmerle-da: :+1:

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
cocreature: Not sure what you’re getting at. All I was suggesting is to inline this or either assign all the `exe._` fields to variables.

remyhaemmerle-da: Of course. thanks.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,28 +61,52 @@ private[preprocessing] final class TransactionPreprocessor(
...
-                case _ =>
-                  val (cmd, acc) = unsafeTranslateNode(cids, node)
-                  (acc, stack :+ cmd)
+                case _: Node.NodeRollback[_] =>
+                  // TODO https://github.com/digital-asset/daml/issues/8020
+                  // how on earth can we turn a rollback node back into a speedy command?
+                  sys.error("rollback nodes are not supported")
               }
           }
+          val globalCids = globalCids0 | newGlobals
```
andreaslochbihler-da, line 102: Are you sure that this works? I'm a bit uneasy that we can't distinguish the following two scenarios:

Scenario 1:
1. There is a contract with cid `cid0` on the ledger.
2. We submit a two-command transaction: Command 1's transaction tree contains a create node that creates a contract with local cid `cid0`. Command 2 is a plain exercise on the contract `cid0` that exists on the ledger. However, replay now interprets this command as acting on the newly created contract.

Scenario 2:
1. There is no contract with cid `cid0` on the ledger.
2. We submit a two-command transaction: Command 1's transaction tree contains a create node that creates a contract with local cid `cid0` and key `k`. Command 2 is a by-key exercise on the key `k`.

One could probably argue that scenario 1 should have failed during interpretation in the submission service, but I'm worried that this subtlety gets overlooked at some point in the future and we get the next bug here.

andreaslochbihler-da (approved):
I've left a concern in one comment. I think this works for the moment, but there's the danger that two commands with different semantics might yield the same preprocessing result. It's more a case of being cautious.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
cocreature: That is a good point, we definitely violate the spec here, quoting:
```
* The *local* contract IDs are the IDs of the contracts created by the
  transaction.

* The *global* contract IDs are the contract IDs that:
   
   * appear in the commands that produced the transaction. This
     includes the IDs of the exercised contract, together with all the
     IDs referenced in the arguments of the create and exercise
     commands;
   * that are fetched or looked up by key unless they are local;
   * are referenced in the input contracts.
```
so `cid0` should definitely be global due to the exercise and we should get a conflict.

Not really sure how to fix that. It seems like the only way out is to preserve `byKey` info on exercise which we should do either way but that doesn’t help for old transaction versions.

andreaslochbihler-da: After another offline discussion with @remyhaemmerle-da: There are probably two ways to approach this;

1. Properly distinguish between normal exercise and exerciseByKey nodes using the `byKey` flag. For backwards compatibility, we could treat all exercise nodes without a `byKey` field as normal exercises. This preserves the current behaviour, but still needs special treatment for `CreateAndExercise` commands.
2. In general allow exercises in later commands to use contract IDs created in earlier commands of the same transaction (both as input contract and in their choice arguments). This is the direction of this PR.

In scenario 1, the created contract clashes with an existing contract. So the Daml ledger should reject the submission with `Contract cid0 already exists`. The interesting part here is that the existing `cid0` might have been archived long ago (and just sits in the submitting participant's store of divulged contracts, so pruning may have removed the existence of `cid0` from the ledger's memory, but then the skew bounds on the ledger time (a contract ID commits to the ledger time of the creating transaction) should be tight enough that `cid0` cannot be created in this transaction.

So, we could update the spec to say that if a later exercise command refers to a contract created by an earlier command, then that's not considered a global contract ID for the transaction. What should we do about an exercise command referring to a contract ID that's created in its consequences or by a later command? I'd argue that this is an internally inconsistent transaction and should be rejected, and AFAICS this PR implements this.

cocreature: I think there are two mismatches which bite us here both of which we should fix but they are slightly orthogonal:
1. The mismatch between transaction nodes and LF primitives. This affects by-key flags but not createAndExercise (there is no primitive for that in LF)
2. The mismatch between transaction nodes and ledger API commands. This affects by-key flags but only on root nodes and createAndExercise.

We should absolutely fix 1 and add by-key flags but as you pointed out that does not make the issue go away for createAndExercise. Just doing that leads to what I’d argue are slightly unintuitive semantics: If you modify your scenario 1 such that the exercise in the second command is not the root node, this is already perfectly fine. Your approach 2 (implemented in this PR but probably should also get a spec update) removes that annoying special treatment of root nodes from the cid freshness check afaict. That seems like a more robust solution to address mismatch 2 (in this instance). I’m not really qualified to judge if relaxing the CID check here causes any issues but it sounds like you’re saying it’s fine. Assuming that is the case, my preference would be the following two steps:

1. Implement approach 2, i.e., this PR but also update the spec to match this (can be in a separate PR). This is all we need to fix the CID freshness check.
2. Add byKey flags to transaction nodes after that (soonish). This address mismatch 1 which causes other issues.

wdyt?

andreaslochbihler-da: I don't see a problem with relaxing the CID check here, as this should be caught on a higher level (freshness check for created contracts). I recommend to document the need for this freshness check; otherwise we'll forget and eventually run into a problem with some of the Daml ledgers.

The `byKey` flag is orthogonal in this approach. I would like to see the flag added officially at some point; it's not just at root nodes though, because child nodes can be come root nodes via projection.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
remyhaemmerle-da: It appear the test is not 100% necessary, as the dynamic check performed by the reinterpretation will catch those problematic cases. 
I  still prefer to let the test here. I add a comment and a test.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -28,95 +28,112 @@ private[preprocessing] final class TransactionPreprocessor(
...
+  }
+
+  /*
+   * Translate a transaction tree into a sequence of Speedy commands and collect the global CIDs.
+   *
+   * A contract ID `cid` is *local* w.r.t. a node `n`, if either:
+   *  - it is local in a previous node w.r.t. traversal order, or
+   *  - `n` is a create node such that `n.coid == cid`
+   * A contract ID `cid` is *global* in a root node `n`, if:
+   *  - it is
```
cocreature, line 82: ?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -28,95 +28,112 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * A contract ID `cid` is *global* in a root node `n`, if
+   *  - `cid` is not local w.r.t. `n`, and
+   *  - if `cid` is an input of a `n`, i.e. :
+   *    * `n` is a create node and `cid` appear in the payload of the create contract (`n.arg`)
+   *    * `n` is an exercise node and `cid` appear in the exercise argument (`n.choosenValue`)
+   *    * `n` is an exercise node and `cid` is the ID of the exercise contract (`n.taretCoid`)
```
cocreature, line 88:
```suggestion
   *    * `n` is an exercise node and `cid` is the ID of the exercise contract (`n.targetCoid`)
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -28,95 +28,112 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * A contract ID is *global* w.r.t a transaction `tx` if it is global w.r.t one of the root `tx`.
```
cocreature, line 89: Perhaps worth pointing out in the comment that this is only parts of the definition of global cids. In particular, a contract id is also global if it is referenced in the arguments of a global fetched or lookup’d contract. We detect those cases during interpretation so no need to add it here, just asking to clarify the comment.

remyhaemmerle-da: :+1:

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
remyhaemmerle-da: dropped.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * and collect the global contract IDs.
+   * A contract ID `cid` is considered *local* w.r.t. a node `n`, if
+   * either:
+   *  - it is local in a previous node w.r.t. traversal order, or
+   *  - `n` is a create node such that `n.coid == cid`
+   *
+   * A contract ID `cid` is considered *global* in a root node `n`,
+   * if:
+   *  - `cid` is not considered local w.r.t. `n`, and
+   *  - if `cid` is an input of a `n`, i.e. :
```
andreaslochbihler-da, line 86:
```suggestion
   *  - if `cid` is referenced in `n`, i.e. :
```
Input contract is typically used for the contract ID that an exercise acts on, not the ids that appear in the choice argument.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   *    - `n` is a create node and `cid` appear in the payload of the
```
andreaslochbihler-da, line 87:
```suggestion
   *    - `n` is a create node and `cid` appears in the payload of the
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   *      create contract (`n.arg`)
+   *    - `n` is an exercise node and `cid` appear in the exercise
```
andreaslochbihler-da, line 89:
```suggestion
   *    - `n` is an exercise node and `cid` appears in the exercise
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+    ) = Acc(
+      globalCids ++ newInputCids.filterNot(localCids),
+      localCids ++ newLocalCids,
+      commands :+ cmd,
+    )
   }
 
+  /*
+   * Translates a transaction tree into a sequence of Speedy commands
+   * and collect the global contract IDs.
```
andreaslochbihler-da, line 77:
```suggestion
   * Translates a transaction tree into a sequence of Speedy commands
   * and collects the global contract IDs.
```
Either `translates` and `collects` or both without trailing `s`.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * A contract ID `cid` is considered *local* w.r.t. a node `n`, if
+   * either:
+   *  - it is local in a previous node w.r.t. traversal order, or
```
andreaslochbihler-da, line 80: Do you want to make clear that the traversal order visits all nodes of a transaction (rather than just the root nodes, which is what the rest of this comment is about)?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   *      argument (`n.choosenValue`)
+   *    - `n` is an exercise node and `cid` is the ID of the exercise
+   *      contract (`n.targetCoid`).
+   *
+   * A contract ID is considered *global* w.r.t. a transaction `tx` if
+   * it is global w.r.t. one of the roots of `tx`.
+   *
+   * Note that it is, in general, not possible to recover from a
+   * transaction, the original sequence of command that generate this
+   * latter. In particular:
```
andreaslochbihler-da, line 99:
```suggestion
   * transaction, the original sequence of command that generated this
   * transaction. In particular:
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   *    exercise performed "by key" (as of LF v1.13).
+   *  - we cannot distinguish a createAndExercise from a create
+   *    followed by an exercise.
+   *
+   * Consequently the sequence of commands and the set of global
+   * contract IDs generated by this method may be different from the
+   * original sequence of commands. In particular:
+   * - all exercises are translated into exercise by ID.
+   * - a cid is not considered global if there exists a create node
+   *   within the transaction that create a contract with the same ID.
```
andreaslochbihler-da, line 110:
```suggestion
   *   within the transaction that creates a contract with the same ID.
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   *
+   * Under the assumption that the underline ledger guarantee the
```
andreaslochbihler-da, line 112:
```suggestion
   * Under the assumption that the underlying ledger guarantees the
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * uniqueness of all contract IDs (including transient contracts),
+   * the reinterpretation of the generates transaction will succeeds
```
andreaslochbihler-da, line 114:
```suggestion
   * the reinterpretation of the generated transaction will succeed
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -62,17 +62,61 @@ private[preprocessing] final class TransactionPreprocessor(
...
+   * iff the original submission was valid and succeed.
```
andreaslochbihler-da, line 115:
```suggestion
   * iff the original submission was valid and succeeded.
```

daml-lf/spec/contract-id.rst
```diff
@@ -101,18 +101,19 @@ Contract ID uniqueness
...
-contract IDs on the whole ledger.  This can be done by enforcing
-global uniqueness of the seeds or by appropriately suffixing the
-contract IDs.  No other requirement (except the 94 bytes size limit)
-is assumed for those suffices.
+Ledger implementations are responsible for enforcing global uniqueness
+of all the contract IDs referenced by the ledger, including transient
+contract IDs.  This can be done by enforcing global uniqueness of the
+seeds or by appropriately suffixing the contract IDs.  No other
+requirement (except the 94 bytes size limit) is assumed for those
+suffices.
```
andreaslochbihler-da, line 109:
```suggestion
suffixes.
```

daml-lf/spec/contract-id.rst
```diff
@@ -101,18 +101,19 @@ Contract ID uniqueness
...
 
 The simplest approach consists to suffix all local contract ID with a
 uniquely global transaction ID. Alternatively central committer ledger
-can completely avoid suffixing by enforcing that same pair (submission
-seed, submission time) is never used more that once, the discriminator
-allocation scheme ensuring in this case the uniqueness of allocated
-discriminators.
+can completely avoid suffixing by enforcing that the pair (submission
+seed, submission time) is not used by two different submission, the
```
andreaslochbihler-da, line 114:
```suggestion
seed, submission time) is not used by two different submission, as the
```

andreaslochbihler-da (approved):
Left a few stylistic comments.

Issue #5753, opened 2020-04-28
Improve code generated for pattern matching on integers/enums


hurryabit:
Currently, some cases produce a linear search when pattern matching on integers. We should avoid searches that are too long and produce a binary search tree instead. See https://github.com/digital-asset/daml/blob/ea50c329f28ac1d4507dcdeda42e01d6e491c445/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs#L831

PR #9381, opened 2021-04-12
Pattern matching for RoundingMode

sofiafaro-da:
- Fix the order of RoundingMode constructors in GHC.Types to match the LF built-in order. Try to match this order across all code and documentation, and added a test for this order.
- Implements pattern matching for RoundingMode. The added machinery could also be useful for solving #5753 in the future.
- Implements Show instance for RoundingMode. (Mainly so we can use them in tests via DA.Assert. Nice to have either way.) Moved BigNumeric Show instance to GHC.Show.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
+++ b/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
@@ -1275,6 +1275,13 @@ data EqualityAlternative = EqualityAlternative
         -- ^ Right-hand side of case alternative. Potentially divergent.
     } deriving (Eq, Ord)
 
+-- | Represents the body of a generalised case expression.
+data GeneralizedCaseBody
+    = GCBExpr LF.Expr
+        -- ^ Expression representing the generalised case.
+    | GCBAlts [CaseAlternative]
+        -- ^ Alternatives for a regular case statement.
+
 -- | Make a case expression from GeneralisedCaseAlternatives.
 --
 -- The scrutinee will be evaluated multiple times unless all the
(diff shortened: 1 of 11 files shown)


compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1208,8 +1202,9 @@ convertExpr env0 e = do
             let fldName = fldNames !! fldIndex
             recTyp <- convertType env (varType bind)
             pure $ mkDictProj env (fromTCon recTyp) fldName scrutinee' `ETmApp` EUnit
-    go env o@(Case scrutinee bind _ [alt@(DataAlt con, vs, x)]) args = fmap (, args) $ do
+    go env o@(Case scrutinee bind resultType [alt@(DataAlt con, vs, x)]) args = fmap (, args) $ do
         convertType env (varType bind) >>= \case
+            -- opaque types have no patterns that can be matched
```
cocreature, line 1207: I assume that’s because in Core a pattern match on a literal is already an equality + a case on the result of that?

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1233,32 +1228,93 @@ convertExpr env0 e = do
...
+-- by LF's CaseAlternative syntax. In particular the pattern here
+-- is some expression that we want to use the built-in equality
+-- to match.
+data EqualityAlternative = EqualityAlternative
+    { eqaltPattern :: LF.Expr
+        -- ^ Pattern expression to test against. This is an inert
+        -- expression (can't raise errors, e.g. a constructor).
+    , eqaltBody :: LF.Expr
+        -- ^ Right-hand side of case alternative. Potentially divergent.
+    } deriving (Eq, Ord)
```
cocreature, line 1276: Just an idea:
Currently you factor this into
equality = eqExpr * rhs
other = pattern * rhs

Would it be a good idea to factor it into?
alternative = pattern * rhs
pattern = casePattern | equality

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1233,32 +1228,93 @@ convertExpr env0 e = do
...
+  where
+    finalize :: Either LF.Expr [CaseAlternative] -> LF.Expr
+    finalize (Left e) = e
+    finalize (Right []) = ECase scrutinee
+        [ CaseAlternative CPDefault $ EBuiltin BEError `ETyApp` resultType `ETmApp` EBuiltin (BEText "Unreachable") ]
+        -- GHC only generates empty case alternatives if it is sure the scrutinee will fail.
+        -- LF doesn't support empty alternatives, so we turn this into a non-empty alternative.
+    finalize (Right alts) = ECase scrutinee alts
+
+    addCaseAlternative :: GeneralisedCaseAlternative -> Either LF.Expr [CaseAlternative] -> Either LF.Expr [CaseAlternative]
```
cocreature, line 1297: Maybe worth adding a comment on the meaning of `Left` and `Right` or alternatively define a custom type.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1233,32 +1228,93 @@ convertExpr env0 e = do
...
+    addCaseAlternative (GCANormal alt) (Right alts) = Right (alt : alts)
+    addCaseAlternative (GCANormal alt) (Left e) =
+        case alt of
+            CaseAlternative CPDefault _ -> Right [alt]
```
cocreature, line 1301: Do we ever hit this? I thought `CPDefault` always comes last where we still have `Right []` in the accumulator.

cocreature (approved):
Nice work, thank you!

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
sofiafaro-da: Yes 🙃

compiler/damlc/daml-prim-src/GHC/Show.daml
```diff
@@ -122,6 +122,21 @@ instance Show Decimal where
   show = primitive @"BEToText"
 #endif
 
+#ifdef DAML_BIGNUMERIC
+instance Show BigNumeric where
+  show = primitive @"BEToTextBigNumeric"
+instance Show RoundingMode where
```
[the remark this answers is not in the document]
sofiafaro-da: Unknown. The derived Show instances always implement `shows` and rely heavily on cons being O(1) for strings, though it's O(n) in daml. I wouldn't be surprised if this manual version is twice as fast for `show`, and a lot faster for `shows` ... but it probably isn't a big difference.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
sofiafaro-da: That's what I had at first and it was slightly more annoying to work with. I think I'll keep it like this for now and we can switch it later if it becomes more helpful.

sofiafaro-da: I went ahead and factored it as you suggested :-)

PR #9386, opened 2021-04-12
daml-sdk-head speed improvement proposal

hubert-da:
Recently, I had to switch between the sdk repo and the integration kit repo a lot. Therefore, I had to constantly run `daml-sdk-head` which, initially, was taking ~15 min on my machine:
```
Done installing JARs as 0.0.0.

________________________________________________________
Executed in   17.75 mins    fish           external
   usr time   14.82 secs    0.32 millis   14.82 secs
   sys time    9.12 secs    2.01 millis    9.11 secs
```
It's a bit long feedback loop.

This change reduces the time to ~5 min, as most of the time is spent generating scaladocs:
```
Done installing JARs as 0.0.0.

________________________________________________________
Executed in  289.21 secs    fish           external
   usr time   12.67 secs    0.24 millis   12.67 secs
   sys time    7.50 secs    1.34 millis    7.50 secs
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/bazel_tools/scala.bzl
+++ b/bazel_tools/scala.bzl
@@ -381,9 +381,10 @@ rm -f {jar_output}
 def _scaladoc_jar_impl(ctx):
     # Detect an actual scala source file rather than a srcjar or other label
     srcFiles = [
-        src.path
-        for src in ctx.files.srcs
-        if src.is_source or src in ctx.files.generated_srcs
+# TODO: add a config param to omit the below
+#        src.path
+#        for src in ctx.files.srcs
+#        if src.is_source or src in ctx.files.generated_srcs
     ]
 
     if srcFiles != []:


hubert-da:
@cocreature what do you think? where would you put a config param to omit the scaladoc generation?

bazel_tools/scala.bzl
```diff
@@ -381,9 +381,10 @@ rm -f {jar_output}
...
     # Detect an actual scala source file rather than a srcjar or other label
     srcFiles = [
-        src.path
-        for src in ctx.files.srcs
-        if src.is_source or src in ctx.files.generated_srcs
+# TODO: add a config param to omit the below
+#        src.path
+#        for src in ctx.files.srcs
+#        if src.is_source or src in ctx.files.generated_srcs
     ]
```
SamirTalwar-DA, line 388: We need these Scaladocs when we publish the real thing.

Perhaps you can tell `daml-sdk-head` to run `bazel build` with the parameter `--build_tag_filters '-scaladoc,-pdfdocs'` instead?

cocreature: That’s not going to work, there are two things you need to change:
1. The Haskell release script needs to gain an option to disable scaladoc (and probably also javadoc) builds.
2. daml-sdk-head then needs to pass this option through either optionally or always (I’m fine with both).

hubert-da: @SamirTalwar-DA 
Yeah, it's solely about the dev feedback loop.

Thanks, I might need to take a deeper look, as both:
- `export BAZEL_MODE_FLAGS="--build_tag_filters '-scaladoc,-pdfdocs'" && daml-sdk-head`
- `bazel build ${BAZEL_MODE_FLAGS[@]:-} --build_tag_filters '-scaladoc,-pdfdocs' //release:sdk-release-tarball-ce.tar.gz`

didn't work (still build the scaladocs).

cocreature: Tags aren’t going to get you anywhere here. The haskell script selects individual targets that it depends on. You need to change that and not go via tags.

hubert-da:
Already done by @cocreature in some other PR, thanks!

PR #9387, opened 2021-04-12
Speed up compilation of Ledger API Test Tool

cocreature:
diffx.generic.auto._ is apparently super slow especially if you use it
repeatedly to derive the same implicits. This PR drops all uses of
that and instead adds a few implicits that are defined once in
Assertions.scala.

On my local machine, this speeds up compilation of
TransactionServiceIT from ~100s to <5s.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


mziolekda (approved):
Good job, thanks @cocreature

kamil-da (approved):
👏

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -32,11 +36,58 @@ object Assertions extends DiffExtensions {
     val diff = Diff.compare(actual, expected)
     if (!diff.isIdentical)
       throw AssertionErrorWithPreformattedMessage(
-        diff.show,
+        diff.show(),
         s"$context: two objects are supposed to be equal but they are not",
       )
   }
 
+  // Specialize a few instances via semi-automatic derivation
```
cocreature, line 44: @kamil-da @mzagorski-da I had to make a bunch of changes to actually get this to work so please take another look if you’re fine with those changes:

1. Use `Diff.derived` instead of `implicitly`. Afaict, the default `implicitly` simply uses `==` so doesn’t produce any diffs which is clearly not the point here.
2. This then ran into a bunch of bugs in `diffx` which forced an upgrade to the latest version which I also included here.

cocreature: And now it fails on Scala 2.13, I hate implicits so much.

cocreature:
Turning this into a draft for now, the Scala 2.13 errors are a giant pain and my attempts at fixing the mutually recursive implicits haven’t been fruitful so far.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
kamil-da: The changes look good, thanks! Although it's a bit disappointing that we have to explicitly write all those definitions.

ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -72,27 +72,48 @@ da_scala_binary(
...
+    srcs = diff_instances,
+    scala_deps = [
+        "@maven//:com_propensive_magnolia",  # transitive dependency of diffx
+        "@maven//:com_propensive_mercator",  # transitive dependency of diffx
+        "@maven//:com_softwaremill_diffx_diffx_core",
+    ],
+    deps = [
+        "//language-support/scala/bindings",
+    ],
+)
```
kamil-da, line 90: Nice!

kamil-da (approved):

cocreature:
apparently this still doesn’t work, now it fails with a null pointer exception in the derived instances at runtime and I’m failing to figure out why. giving up on this for now.

PR #9390, opened 2021-04-12
Oracle json-api: use custom party_array type for signatories, observers

### LINKED ISSUE #9388: Oracle json-api: use array types for signatories/observer, or test for duplicates
Unlike the PostgresQueries, OracleQueries separates signatories and observers into separate tables, doing multiple joins to ensure that every signatory and observer (including the non-matching ones) is included in the resulting structure. There are two drawbacks:

1. the insert and query significantly diverge from PostgresQueries, and entail a strange chain of joins; and
2. the result may contain duplicates that may not be fixable with `select distinct`.

You can encounter duplicates because of the party check matching the same party multiple times, or in a multiparty query, any number of parties matching multiple times. In addition, there is no guarantee that the aggregation order of signatories and observers collected up with `LISTAGG` will remain stable, hence the failure of `distinct` to deal with this.

Shifting to arrays and testing for set intersection will solve this problem. In addition, `SupportedJdbcDriver` may then pass in the correct `Get` and `Put` instances for `Array[String]`.

However, if we do not switch over to arrays, we should test that the duplicate problem does not occur, and figure out how to adjust the queries accordingly. We want to avoid adding a layer of Scala-side processing to handle this if at all possible, because we are trying to generally do _less_ processing in dbbackend, not more (e.g. #9375). In addition, in this case, we should change the types of implicits that `SupportedJdbcDriver` holds to _depend_ on the underlying `Queries` instance, so that the nonsense `Array` instance currently in place can be removed.

S11001001:
Fixes #9388.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -331,6 +331,7 @@ object Queries {
   private[dbbackend] object InitDdl {
     final case class CreateTable(name: String, create: Fragment) extends InitDdl
     final case class CreateIndex(create: Fragment) extends InitDdl
+    final case class CreateType(create: Fragment) extends InitDdl
   }
 
   /** Whether selectContractsMultiTemplate computes a matchedQueries marker,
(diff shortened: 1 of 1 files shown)


stefanobaghino-da:
Should I work on this or will you?

PR #9394, opened 2021-04-13
DPP-334 Manually partition the event table

rautenrieth-da:
This PR removes the automatic partitioning of the `participant_events` table in favor of a manual partitioning. This is easier to port to other databases, and removes unnecessary fields from individual partitions.

### Schema changes

- The schema now contains 4 distinct tables: `participant_events_divulgence`, `participant_events_create`, `participant_events_consuming_exercise`, and `participant_events_non_consuming_exercise`. The individual tables are leaner than before - they don't contain useless fields, and the remaining fields have been made `NOT NULL` where possible.
- There is now a view `participant_events` that contains exactly the same fields as the previous partitioned table of the same name.
- In the indexer, events are grouped by their kind, and inserted directly into the respective tables. This increases the number of SQL statements executed (previously: 1 batch of mixed events, now: 4 batches of events of the same kind).

### Testing

The conformance test tool is now run against the append-only schema on Postgres, using `ledger-on-sql`.

### Future work

- Make fields of `DBDTOV1.Event*` non-optional where applicable.
- Change all queries to select from the individual partitions instead of from the view, where it makes sense.
- Implement the removal of partially written divulgence events


ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -121,52 +121,541 @@ CREATE TABLE participant_events_divulgence             PARTITION OF participant_
...
 CREATE TABLE participant_events_consuming_exercise     PARTITION OF participant_events FOR VALUES IN (20);
 CREATE TABLE participant_events_non_consuming_exercise PARTITION OF participant_events FOR VALUES IN (25);
+*/
+
+---------------------------------------------------------------------------------------------------
+-- Events table: divulgence
+---------------------------------------------------------------------------------------------------
+CREATE TABLE participant_events_divulgence (
+    -- * kinds of events
+    event_kind smallint NOT NULL CONSTRAINT divulgence_event_kind CHECK (event_kind = 0),
```
meiersi-da, line 131: Why not add the kind in the view directly as constants?

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -121,52 +121,541 @@ CREATE TABLE participant_events_divulgence             PARTITION OF participant_
...
+            NULL as exercise_argument,
+            NULL as exercise_result,
+            NULL as exercise_actors,
+            NULL as exercise_child_event_ids,
+            create_argument_compression,
+            create_key_value_compression,
+            NULL as exercise_argument_compression,
+            NULL as exercise_result_compression
+        FROM participant_events_create
+    UNION
```
meiersi-da, line 587: I believe we need a 'UNION ALL' as 'UNION' does sorting and removal of duplicates.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -121,52 +121,541 @@ CREATE TABLE participant_events_divulgence             PARTITION OF participant_
...
+
+    -- * divulgence and create events
+    -- [N/A for this event kind]: create_argument bytea,
+
+    -- * create events only
+    -- [N/A for this event kind]: create_signatories text[],
+    -- [N/A for this event kind]: create_observers text[],
+    -- [N/A for this event kind]: create_agreement_text text,
+    -- [N/A for this event kind]: create_key_value bytea,
+    -- [N/A for this event kind]: create_key_hash bytea,
```
meiersi-da, line 441: @rautenrieth-da : what's your view on removing the comments here and adding the comments where necessary on the view below?

rautenrieth-da: The comments are only here to simplify review. I will remove them all in a separate commit before undrafting the PR.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
rautenrieth-da: :+1:

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
rautenrieth-da: 👍

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/RawDBBatchPostgreSQLV1.scala
```diff
@@ -234,20 +324,57 @@ object RawDBBatchPostgreSQLV1 {
...
+    private[this] var eventsBatchBuilderNonConsumingExercise: EventsBatchBuilderExercise = _
+    private[this] var configurationEntriesBatchBuilder: ConfigurationEntriesBatchBuilder = _
+    private[this] var packageEntriesBatchBuilder: PackageEntriesBatchBuilder = _
+    private[this] var packagesBatchBuilder: PackagesBatchBuilder = _
+    private[this] var partiesBatchBuilder: PartiesBatchBuilder = _
+    private[this] var partyEntriesBatchBuilder: PartyEntriesBatchBuilder = _
+    private[this] var commandCompletionsBatchBuilder: CommandCompletionsBatchBuilder = _
+    private[this] var commandDeduplicationBatchBuilder: CommandDeduplicationBatchBuilder = _
+
+    // Sequential ID within this batch. The final sequential ID
```
rautenrieth-da, line 339: Finish this comment and add a description to `offsetSequentialEventIds()`.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTableDelete.scala
```diff
@@ -15,21 +15,33 @@ object EventsTableDelete {
...
+              SELECT 1 FROM participant_events_consuming_exercise as archive_events
+              WHERE
+                archive_events.event_offset <= $endInclusive AND
+                archive_events.contract_id = delete_events.contract_id
+            );
+          -- Exercise events
+          delete from participant_events_consuming_exercise as delete_events
+          where
+            delete_events.event_offset <= $endInclusive and
+          delete from participant_events_non_>consuming_exercise as delete_events
```
rautenrieth-da, line 42: This can't possibly work. Make sure this code path is tested.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
```diff
@@ -49,124 +49,415 @@ CREATE TABLE parameters (
...
+    create_observers text[] NOT NULL,
     create_agreement_text text,
     create_key_value bytea,
     create_key_hash bytea,
 
+    -- * compression flags
+    create_argument_compression SMALLINT,
+    create_key_value_compression SMALLINT
+);
+-- these columns contain data that is generally incompressible, so don't try it
```
tudor-da, line 143: Can you update this comment? It's quite ambiguous

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/UpdateToDBDTOV1.scala
```diff
@@ -212,13 +205,7 @@ object UpdateToDBDTOV1 {
...
                 tree_event_witnesses =
                   blinding.disclosure.getOrElse(nodeId, Set.empty).map(_.toString),
-                create_argument = None,
-                create_signatories = None,
-                create_observers = None,
-                create_agreement_text = None,
-                create_key_value = createKeyValue
-                  .map(compressionStrategy.createKeyValueCompression.compress),
-                create_key_hash = None,
+                // TODO append-only: createKeyValue is not stored, check whether we need it for upcoming ledger API additions
```
meiersi-da, line 208: Yes, AFAIR @tudor-da 's mutable state cache require the create_key_hash to be stored for consuming exercises, so that key-unassignments can be tracked. 

@tudor-da : What else do you need?

@rautenrieth-da : AFAIS, we can add the createKeyValue vai a (potentially costly) migration at a later stage when implementing Ledger API features. Right now, I'd focus on what we need for the current Ledger API and the 1k trade/s.

meiersi-da (commented):
Thanks @rautenrieth-da for this good work! As you said, it is a bit boilerplaty, but it is also simple, which I like. Comments added. Nothing major.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V50__Append_only_schema.sql
rautenrieth-da: Updated the comment.

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PostgresDAO.scala
```diff
@@ -576,8 +809,21 @@ case class JDBCPostgresDAO(jdbcUrl: String) extends PostgresDAO with AutoCloseab
       |FROM participant_command_completions
       |WHERE completion_offset > ?;
       |
+      |-- TODO append-only: how do we remove divulgence events? They don't have an event_offset
```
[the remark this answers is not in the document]
rautenrieth-da: Good point. They didn't use an offset in the schema before, but I don't see a reason why the shouldn't use the offset of the transaction. I'll fix this in this PR.

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/UpdateToDBDTOV1.scala
tudor-da: We need the `create_key_value` and we've already implemented it.

rautenrieth-da: Fixed in this PR.

tudor-da (approved):
LGTM. Tested in an E2E run with no observed performance impact.

PR #9396, opened 2021-04-13
Add a ledger model page for exceptions.

sofiafaro-da:
This page describes the changes to the ledger model as part of the exceptions feature. The changes should be kept separate for now, but once exceptions are stable and out for a while, they should probably be incorporated into the rest of the daml ledger model.

Part of #8020. **This PR is to be merged only when we start releasing exceptions.**

changelog_begin

- [docs] The Daml ledger model has been updated to describe the changes introduced by the new exceptions feature. See  here: https://docs.daml.com/concepts/ledger-model/ledger-exceptions.html

changelog_end


docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+one of which was rolled back, and so they don't break the ledger integrity.
+
+Privacy
++++++++
+
+Rollback nodes also have an interesting effect on the notion of privacy in
+the ledger model. When projecting a transaction for a party `p`, it's
+necessary to preserve some of the rollback structure of the transaction,
+even if `p` does not have the right to observe every action under it. This
+is because we need for `p` to be able to verify that an exercise under a
```
oggy-, line 82: Well we need a bit more than that. E.g., we also need `p` to be able to verify a `Fetch` of a contract on which it's a signatory, or that a rolled-back `Create` where it was a signatory was authorized by it. I don't have an immediate suggestion for how to state this succinctly and accurately though :(

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+rollback (to which they are is party) is conformant, but we also need `p`
+to know that the exercise was rolled back.
+
+As such, the notion of projection for rollback nodes is defined as follows:
+When projecting a rollback node for a party `p`, we project the subtransaction
+contained in the rollback node, and then we wrap the projected subtransaction
+in a new rollback node.
```
oggy-, line 89: What do we do if the projected subtransaction is empty?

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Subsequently, we see whether we can "normalise" the projected rollback node to
```
oggy-, line 91: I know that I'm the one insisting on the "normalization" :) but the rules below feel a bit arbitrary as presented. Let's think whether we can come up with a higher-level criterion, at least for us, and if we can come up with something convincing and simple, we can also include it for the reader.

I guess we want to block "information flows" in a way: things that don't concern you shouldn't change your view of the transaction. Here's the best formulation that I have right now (feels a bit rough): if we take a DAML transaction `tx` and extend it to a transaction `tx'` such that we either:
1. add actions to `tx` where `p` is not an informee
2. wrap individual actions in `tx` in other actions where `p` is not an informee

then `proj(tx', p) = proj(tx, p)`.

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+- If the projected rollback node is empty, we drop it.
+
+Note that all of these transformations preserve the "after" relation among
+actions in the transaction tree. They only affect the structure of the
+transaction by reducing the amount of variation that may occur in a
+projected rollback node. This way, information cannot be leaked through
+the arrangement of rollback nodes.
+
+The privacy section of the ledger model makes a point of saying that a
+ledger model should be **subaction-closed** to support projections. But
```
oggy-, line 131:
```suggestion
contract model should be **subaction-closed** to support projections. But
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+this requirement is not necessarily true once we introduce rollbacks.
+Rollback nodes may contain actions that are not valid as standalone actions,
+since they may have been interrupted prematurely by an exception.
+
+Instead, we require that the ledger model be **projection-closed**, i.e.
```
oggy-, line 136:
```suggestion
Instead, we require that the contract model be **projection-closed**, i.e.
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+closed under projections for any party 'p'. This is a weaker requirement
```
oggy-, line 137: There's actually something more for Canton, which I also can't quite properly formulate yet :roll_eyes: I'll try to explain here. But I think it may be valuable if we can extract the essence, so worth spending some brain cycles on this.

For distributed checks we will have Alice checking what she thinks that Bob should see, and checking that Bob has actually seen it. Alice's view is based on `proj(tx, Alice)`, so she will be computing `proj(proj(tx, Alice), Bob)`. So this itself has to be model-conformant, otherwise, Alice might think that Bob would reject this.

There's probably a bit more that we want there. E.g., if Bob is "above" Alice in the transaction, then we probably want `proj(tx, Alice) == proj(proj(tx, Bob), Alice)`. But not sure how to formulate "above"...

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+This Daml code will try to create a contract, and exercise a choice on this
+contract, before throwing an exception. That exception is caught immediately,
+and then another contract is created.
+
+Thus a rollback node is created, to reset the ledger to the state it had
+at the start of the "try" block. The rollback node contains the create and
+exercise nodes. After the rollback node, another contract is created.
+Thus the final transaction looks like this:
+
+.. code-block:: none
```
oggy-, line 167: I'd represent this graphically.

oggy- (commented):
Hey, thanks a lot. This is looking pretty good, but I do have a bunch of comments :roll_eyes: The most important ones:
1. I'd like us to understand a bit better what are the driving forces for our definitions of projections, as these get more complicated with exceptions
2. I'd really want to see graphical examples (but OK to wait until we have everything else in place, so also happy to do in the second iteration)

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+#. If `act1` and `act2` are two actions that share a parent (e.g. two
+   top-level actions, or two actions under the same rollback node, or two
+   consequences of the same action), and `act1` precedes `act2` in the list
+   of children, then `act2` comes after `act1`.
+
+#. If `node1` is an action, and `node2` is a rollback node, and they both
+   share a parent, and `node1` precedes `node2` in the list of children,
+   then `node2` comes after `node1`.
+
+#. If `node1` and `node2` are actions or rollback nodes in the transaction
```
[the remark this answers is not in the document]
sofiafaro-da: This also includes rollback nodes being ordered before their children, though.

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: > For distributed checks we will have Alice checking what she thinks that Bob should see, and checking that Bob has actually seen it. Alice's view is based on proj(tx, Alice), so she will be computing proj(proj(tx, Alice), Bob). So this itself has to be model-conformant, otherwise, Alice might think that Bob would reject this.

This part should be implied by "projection-closed" I think. If it's closed under one projection, it's also closed under two projections!

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+  We repeat this step until the projected rollback does not start with another
+  rollback node.
+
+- If the projected rollback node ends with another rollback node, for instance:
+
+  .. code-block:: none
+
+    'Rollback' [ node1, ..., nodeN, 'Rollback' [ node1', ..., nodeM' ] ]
+
+  Then we remove the flatten the inner rollback node into its parent:
```
cocreature, line 116:
```suggestion
  Then we flatten the inner rollback node into its parent:
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+
+The privacy section of the ledger model makes a point of saying that a
+contract model should be **subaction-closed** to support projections. But
+this requirement is not necessarily true once we introduce rollbacks.
+Rollback nodes may contain actions that are not valid as standalone actions,
+since they may have been interrupted prematurely by an exception.
+
+Instead, we require that the contract model be **projection-closed**, i.e.
+closed under projections for any party 'p'. This is a weaker requirement
+that matches what we actually the need.
```
cocreature, line 138:
```suggestion
that matches what we actually need.
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+siblings, but never "before". The only time a rollback node comes "before"
+another node, is when that other node is a child or descendant of the
+rollback node.
+
+So in the example of a consuming exercise in a rollback node, followed by a
+consuming exercise on the same contract outside of the rollback node, neither
+exercise comes "after" the other. They are part of separate continuities,
+one of which was rolled back, and so they don't break the ledger integrity.
+
+Privacy
```
cocreature, line 75: It seems implicit in your definition since rollbacks are not actions but I think it might be worth spelling this out explicitly in your doc somewhere:
rollback nodes have neither informees nor required authorizers

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+With this modified "after" relation, the notion of consistency remains the
+same. Note that under this definition, a rollback node can be "after" its
+siblings, but never "before". The only time a rollback node comes "before"
+another node, is when that other node is a child or descendant of the
+rollback node.
+
+So in the example of a consuming exercise in a rollback node, followed by a
+consuming exercise on the same contract outside of the rollback node, neither
+exercise comes "after" the other. They are part of separate continuities,
+one of which was rolled back, and so they don't break the ledger integrity.
```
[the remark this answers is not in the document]
bame-da: Yes, me too. I think it would be useful to hint at some alternative ways of looking at this to help readers built a mental model. What we are effectively saying is that every Rollback Node is a fork. One branch is the _main_ branch, which skips the rolled back subtree, and one branch is the one that goes down the rolled back subtree and terminates at its end.
Ie we get a tree (as defined by your "after" relation) consisting of a committed trunk, with rolled back branches. We require every branch comply with the integrity property from the main Ledger Model. 

This picture should also be easy to illustrate as per Oggy's suggestion.

bame-da (commented):
This is looking good. I don't have much to add to the comments that are already there.

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: We drop the rollback node (third bullet point below).

oggy-:
@sofiafaro-da I think I can formulate somewhat helpful criteria for projections now. Basically, we can take the weakening of action order for consistency as the only manifestation of transaction rollbacks in our ledger model. The Rollback nodes are just a way to represent that order. 

So the sequence of actions in a projection for a party `p` doesn't change with rollbacks - it's still the list of maximal (in sub-action order) actions where the `p` is an informee. Let's call this the "raw projection" for p. But we need to ensure that the consistency order of actions in a projection is the same as in the original ledger. As we use rollback node to encode this order, we need to insert some of those into the projection for p. I can think of two criteria for the definition of projection:
1. privacy criterion: any two transactions whose raw projections for `p` are equal, and whose consistency order is the same when restricted to the actions in the raw projection for `p`, should also have the same projection for p
2. we don't want unnecessary structure in the transaction, so the projection should use the minimal amount of rollback nodes

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,178 @@
...
+As such, the notion of projection for rollback nodes is defined as follows:
+When projecting a rollback node for a party `p`, we project the subtransaction
+contained in the rollback node, and then we wrap the projected subtransaction
+in a new rollback node.
+
+Subsequently, we see whether we can "normalise" the projected rollback node to
+prevent leaking information about the structure of the original transaction.
+In particular, we apply the following steps:
+
+- If the projected rollback node starts with another rollback node, for instance:
```
cocreature, line 95: Do any of these steps still apply, if the party sees the parent exercise? Looking at the first rule for example, it seems unintuitive to me that even if I see the full rollback the structure won’t be what it was originally. I know if I see the parent exercise or not, so leaking information is not an issue here afaict.

sofiafaro-da: Projection (for `p`) does not apply to consequences of an exercise that is seen by the party `p`. So this only happens for top-level rollback nodes OR rollback nodes under an exercise that was projected away (OR nested rollback nodes).

sofiafaro-da: Basically you get the original rollback structure if it wasn't projected away. Otherwise we project under the rollback and try to normalize the result.

cocreature: Makes sense, that’s what I was hoping for :+1:

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,288 @@
...
+
+Integrity
++++++++++
+
+The notion of a ledger action coming "after" another ledger action must
+be revised in the presence of rollback nodes. It is not enough to traverse
+the transaction tree in prefix order, because the actions under a rollback
+were rolled back.
+
+For example, a contract may be consumed by an exercise under a rollback node,
```
oggy-, line 68: Would be good to refer to the concrete example at the end of the previous section

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,288 @@
...
+   or, if it fails, to withdraw 500 USD instead".'
+
+In addition, rollback nodes may be nested. This corresponds to a possible
+case where multiple exceptions are raised and caught within the same
+transaction.
+
+Integrity
++++++++++
+
+The notion of a ledger action coming "after" another ledger action must
```
oggy-, line 63: I think only a few people will make the connection to the "before-after" relation from the consistency section. So I'd either give a direct pointer here. Even better would be to motivate the change by the example from the previous section, saying that intuitively this should be OK, but we need to adapt our integrity definition.

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,288 @@
...
+exercise comes "after" the other. They are part of separate continuities,
+one of which was rolled back, and so they don't break the ledger integrity.
+
+..
+   TODO: Add diagrams to demonstrate the "forking".
+
+Authorization
++++++++++++++
+
+Since they are not ledger actions, rollback nodes do not have authorizers
```
oggy-, line 106: You could also relate this to the existing definition in two ways:
1. a commit/ledger is well-authorized exactly when the same ledger with `Rollback` nodes removed (absorbed into their parent transactions) is well-authorized according to the old definition
2. define "top-level action" and "consequence" to "see through" `Rollback` nodes, and then just keep the existing rules.

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,288 @@
...
+For example, ...
+
+..
+   TODO: Add example of an exercise containing a rollback node
+   with diagram. 'Alice orders her bank to "withdraw 1000 USD,
+   or, if it fails, to withdraw 500 USD instead".'
+
+In addition, rollback nodes may be nested. This corresponds to a possible
+case where multiple exceptions are raised and caught within the same
+transaction.
```
oggy-, line 58: Would also be good to have an example - especially something that you can reuse in the `Privacy` section

docs/source/concepts/ledger-model/ledger-exceptions.rst
oggy-: You're right, and I think the "privacy requirements" now mostly capture what I wanted to say (minus the Canton-distributed-check requirement, but we don't need to solve that for this PR)

oggy- (approved):

oggy-:
Pending the examples, comments and the consistency issue, good to go from my side

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -0,0 +1,277 @@
...
+
+With this modified "after" relation, the notion of internal consistency
+remains the same. Meaning that, for any contract `c`, we still forbid the
+create of `c` coming after any action on `c`, and we forbid any action on
+`c` coming after a consuming exercise on `c`.
+
+In the example of a consuming exercise in a rollback node, followed by a
+consuming exercise on the same contract outside of the rollback node,
+neither consuming exercise comes "after" the other. They are part of
+separate continuities, one of which was rolled back, and so they don't
```
andreaslochbihler-da, line 86:
```suggestion
separate continuations, one of which was rolled back, and so they don't
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: Added a link.

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: Done :-)

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: Added a link and made the connection clearer.

sofiafaro-da:
I'm planning to merge this PR as-is -- this one is very long and old at this point -- and make a separate PR to bring out the normalization section (since we decided that we're only going to deal with normalized rollback nodes).

PR #9399, opened 2021-04-13
fix daml-lf/governance.rst

garyverhaegen-da:
There have been a few changes and the names in the current version are no longer valid.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/daml-lf/governance.rst
+++ b/daml-lf/governance.rst
@@ -56,8 +56,7 @@ Overview
   present in the ``daml-lf-X.proto`` and ``transaction.proto`` files.
 
 * Every change to Daml-LF values, nodes, transactions or languages *must* be
-  checked for governance rules conformance and be approved by Rémy Haemmerle
-  and Sofia Faro. Proposed changes can be filed as tickets under labels
+  checked for governance rules conformance and be approved at least two persons among by Rémy Haemmerle (@remyhaemmerle-da), Sofia Faro (@sofiafaro-da) and Moritz Kiefer (@cocreature). Proposed changes can be filed as tickets under labels
   "component/daml-lf" and "discussion".
 
 Language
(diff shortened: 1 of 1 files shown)


daml-lf/governance.rst
```diff
@@ -55,11 +55,10 @@ Overview
...
   present in the ``daml-lf-X.proto`` and ``transaction.proto`` files.
 
-* Every change to Daml-LF values, nodes, transactions or languages
-  *must* be checked for governance rules conformance and be approved
-  by at least two persons among Rémy Haemmerle, Martin Huschenbett
-  and Sophia Mota. Proposed changes can be filed as tickets under
-  labels "component/daml-lf" and "discussion".
+* Every change to Daml-LF values, nodes, transactions or languages *must* be
+  checked for governance rules conformance and be approved by Rémy Haemmerle
+  and Sofia Faro. Proposed changes can be filed as tickets under labels
```
cocreature, line 60:
```suggestion
  checked for governance rules conformance and be approved at least two persons among by Rémy Haemmerle (@remyhaemmerle-da), Sofia Faro (@sofiafaro-da) and Moritz Kiefer (@cocreature). Proposed changes can be filed as tickets under labels
```

Not at all saying that @remyhaemmerle-da and @sofiafaro-da aren’t sufficient and if they are available they should be involved, but this reduces the bus factor a bit and avoids deadlocks when someone is on vacation or unavailable for other reasons.

cocreature (approved):

daml-lf/governance.rst
remyhaemmerle-da: I agree with @cocreature

sofiafaro-da: Agreed :)

sofiafaro-da (approved):
Thank you!

PR #9400, opened 2021-04-13
Add Ledger API test tool tests for exceptions

cocreature:
Currently testing against:
1. Sandbox classic in-memory
2. Sandbox classic on SQL
3. Ledger on SQL
4. Ledger on Memory

Requests for additional tests are covered in https://github.com/digital-asset/daml/issues/9526 and will be adressed separately.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -72,3 +72,38 @@ LF_MAJOR_VERSIONS = ["1"]
 
 def mangle_for_java(name):
     return name.replace(".", "_")
+
+def _to_major_minor(v):
+    (majorStr, minorStr) = v.partition(".")
+    (int(majorStr), int(minorStr))
+
+def _cmp(a, b):
+    if a == b:
+        return 0
+    elif a > b:
+        return 1
+    else:
+        return -1
+
+def _cmp_lf_version(a, b):
+    if a == "1.dev" and b == "1.dev":
+        return 0
+    elif a == "1.dev":
+        return 1
+    elif b == "1.dev":
+        return -1
+    else:
+        aVer = _to_major_minor(a)
+        bVer = _to_major_minor(b)
+        _cmp(aVer, bVer)
+
+def _gte(a, b):
+    return _cmp(a, b) >= 0
+
+def _lte(a, b):
+    return _cmp(a, b) <= 0
+
+versions = struct(
+    lte = _lte,
+    gte = _gte,
+)
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -98,10 +98,10 @@ def _cmp_lf_version(a, b):
         return _cmp(aVer, bVer)
 
 def _gte(a, b):
-    return _cmp(a, b) >= 0
+    return _cmp_lf_version(a, b) >= 0
 
 def _lte(a, b):
-    return _cmp(a, b) <= 0
+    return _cmp_lf_version(a, b) <= 0
 
 versions = struct(
     lte = _lte,
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -95,7 +95,7 @@ def _cmp_lf_version(a, b):
     else:
         aVer = _to_major_minor(a)
         bVer = _to_major_minor(b)
-        _cmp(aVer, bVer)
+        return _cmp(aVer, bVer)
 
 def _gte(a, b):
     return _cmp(a, b) >= 0
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/dummy/ExceptionsIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/dummy/ExceptionsIT.scala
@@ -1,0 +1,9 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.testtool.suites
+
+import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
+
+// Empty dummy suite for LF versions that do not support exceptions.
+final class ExceptionsIT extends LedgerTestSuite {}
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionsIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionsIT.scala
@@ -1,0 +1,245 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.testtool.suites
+
+import com.daml.ledger.api.testtool.infrastructure.Allocation._
+import com.daml.ledger.api.testtool.infrastructure.Assertions._
+import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
+import com.daml.ledger.api.testtool.infrastructure.TransactionHelpers._
+import com.daml.ledger.api.testtool.infrastructure.Synchronize.synchronize
+import com.daml.ledger.test.semantic.Exceptions.{Divulger, ExceptionTester, Fetcher, WithKey}
+import io.grpc.Status
+
+final class ExceptionsIT extends LedgerTestSuite {
+  test(
+    "ExUncaught",
+    "Uncaught exception returns INVALID_ARGUMENT",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    for {
+      t <- ledger.create(party, ExceptionTester(party))
+      failure <- ledger.exercise(party, t.exerciseThrowUncaught(_)).mustFail("unhandled exception")
+    } yield {
+      assertGrpcError(failure, Status.Code.INVALID_ARGUMENT, "unhandled exception")
+    }
+  })
+
+  test(
+    "ExCaught",
+    "Exceptions can be caught",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    for {
+      t <- ledger.create(party, ExceptionTester(party))
+      tree <- ledger.exercise(party, t.exerciseThrowCaught(_))
+    } yield {
+      assertLength(s"1 successful exercise", 1, exercisedEvents(tree))
+      ()
+    }
+  })
+
+  test(
+    "ExCaughtNested",
+    "Exceptions can be caught when thrown from a nested try block",
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    for {
+      t <- ledger.create(party, ExceptionTester(party))
+      tree <- ledger.exercise(party, t.exerciseNestedCatch(_))
+    } yield {
+      assertLength(s"1 successful exercise", 1, exercisedEvents(tree))
+      ()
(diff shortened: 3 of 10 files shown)


ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -136,6 +136,12 @@ conformance_test(
         "--contract-id-seeding=testing-weak",
         "--participant=participant-id=example,port=6865",
     ],
+    lf_versions = [
```
cocreature, line 139: Obviously not going to merge it like this, this is a draft for a reason.

meiersi-da (commented):
@cocreature : this goes in a good direction. I believe many of these tests are related to the read-write conflict tests introduced as part of #9138. With exceptions we get a new set of read-write conflicts. Here are the existing ones for reference:

1. archive-contract vs. non-consuming-choice
2. archive-contract vs fetch-contract
3. archive-contract vs. successful lookup-by-key
4. non-transient-create-with-key vs. transient create-with-key
5. non-transient-create-with-key vs. failing lookup-by-key 

I believe we can extend this for exceptions as follows:

1. archive-contract vs. consuming or non-consuming choice within rollback
2. archive-contract vs fetch-contract within rollback
3. archive-contract vs. successful lookup-by-key within rollback
4. non-transient-create-with-key vs. transient or non-transient create-with-key within rollback
5. non-transient-create-with-key vs. failing lookup-by-key within rollback

I'd expect to test these once in a sequential fashion to see that the expected failure happens and once within the race-condition setting to see that the conflict detection logic for concurrent transactions works as expected.

What do you think?

cocreature:
@meiersi-da 
So looking through your list, 2 seems already covered by `ExRollbackActive`. 4 is covered by `ExRollbackDuplicateKey` for the non-transient part, happy to add something for the transient part.
1 and 3 seems very sensible, happy to add those. If I understand you correctly, 5 is just 4 but you start with the failing lookup and then create it instead of starting with the successful lookup and then archiving it.

As for the race condition tests, I agree those are sensible but I’d prefer to tackle it in to steps and get something in that we agree on first for the semantics of the sequential cases and then add the race condition tests on top of that.

cocreature:
- 270e5d3a6bf63033dff0690a823b72cbff7928ae for 5 (or rather key fetches)
- ee47d460ebdd27be0c28e5fb10a166ba304cecdf for 1 (consuming and nonconsuming)
- 1bef91812f9785e8f4addec5799757c97c8022c4 as a reverse version of 4
- 2337a99109cfdcbf98011075b22911503c866788 for nested try/catch as requested by @mziolekda 

What did I forget? And are you fine with leaving the race condition tests for a separate step?

meiersi-da:
@cocreature : LGTM. I agree wrt adding RaceCondition tests in a separate step once the sequential versions work.

oggy-:
A few more test cases that I'd add:
1. liveness tests for non-conflicts:
    a. a rolled-back consuming exercise doesn't block later consuming or non-consuming exercises (you have one of these covered I believe with your archive after consuming exercise, but not for a non-consuming one)
    b. a rolled-back key creation doesn't block a subsequent creation of the same key
2. check that conflicting and non-conflicting (liveness above) work if you have them in the same transaction, with interesting projections and two nodes

cocreature:
@oggy- covered 1 in 50eb70090cc3b0db3383932fba75e254948d5475 (at least I tried :))

As for 2, do you have something specific in mind there? I can add some random stuff but I think I don’t really have an intuition of what is interesting there.

oggy-:
I think all of the following would be useful to test the internal consistency checks of the ledger under test:
1. repeat the existing tests as-is, just wrapped  (insofar possible) in a single transaction.
2. same as 1, but roll the transaction back, and see that the accept/reject behavior doesn't change.
3. repeat 1 and 2, but now intersperse the transactions (including their rolled back parts) with actions that have different informees than the other actions, such that these informees are hosted on a different participant. 

Possibly useful to combine with the existing checks, such that we test whether the ledger computes the inputs/outputs of a transaction correctly given partially rolled-back. Or maybe checking the ACS suffices.

I know I'm asking for a lot of tests :) But it's a complicated feature and I see the potential for errors in all of these contexts. Maybe there's a clever way to remove some of the repetition...

cocreature:
I’m a bit worried this is getting slightly out of hand. The original exceptions doc asked for 5 different test cases, we are now up to 15 + requests to run all of those in 3 different variations (race condition, single transaction, rollback the whole transaction) and other tests.

oggy-:
Sure, like I said I understand that I'm asking for a large number of tests. I'm not asking you to built them in this PR. However, without these additional tests, we have no coverage for internal consistency checks in the presence of rollbacks. And we've had several problems in that area with contract keys for example.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/dummy/ExceptionsIT.scala
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.testtool.suites
+
+import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
+
+// Empty dummy suite for LF versions that do not support exceptions.
```
cocreature, line 8: If someone has a nicer suggestion here, I’m all ears. This isn’t super nice but it works and is somewhat extensible to more tests that are version specific.

cocreature:
This now passes on Sandbox classic with the SQL backend. I’ve captured the requests for additional tests in #9526. Given that I’d like it to keep working and test it on more ledgers, I would like to merge it with the current sets of tests and then address those requests separately.

cocreature:
I’ve extended this with more ledgers now that they’re supported.

daml-lf/language/daml-lf.bzl
```diff
@@ -72,3 +72,38 @@ LF_MAJOR_VERSIONS = ["1"]
...
+    if a == "1.dev" and b == "1.dev":
+        return 0
+    elif a == "1.dev":
+        return 1
+    elif b == "1.dev":
+        return -1
+    else:
+        aVer = _to_major_minor(a)
+        bVer = _to_major_minor(b)
+        _cmp(aVer, bVer)
```
sofiafaro-da, line 98:
```suggestion
        return _cmp(aVer, bVer)
```

daml-lf/language/daml-lf.bzl
```diff
@@ -72,3 +72,38 @@ LF_MAJOR_VERSIONS = ["1"]
...
+def _gte(a, b):
+    return _cmp(a, b) >= 0
+
+def _lte(a, b):
+    return _cmp(a, b) <= 0
+
+versions = struct(
+    lte = _lte,
+    gte = _gte,
+)
```
sofiafaro-da, line 109: should these be using `_cmp_lf_version` instead of `_cmp`?

cocreature: very good point 96d8b10dc1cbe4164ae1821ffc85b14f69c93b7c

sofiafaro-da (approved):
This looks good. Thank you! 👍

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionsIT.scala
```diff
@@ -0,0 +1,245 @@
...
+      tExercise <- ledger.create(party, ExceptionTester(party))
+      _ <- ledger.exercise(party, t.exerciseRollbackNonConsuming(_, tExercise))
+      _ <- ledger.exercise(party, t.exerciseArchive(_))
+      failure <- ledger
+        .exercise(party, t.exerciseRollbackNonConsuming(_, tExercise))
+        .mustFail("contract is archived")
+    } yield {
+      assertGrpcError(failure, Status.Code.ABORTED, "Contract could not be found")
+    }
+  })
```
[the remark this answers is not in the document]
cocreature: 00ee366a4609d3f79f6f19a71a28e32ab05835ae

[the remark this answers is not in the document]
sofiafaro-da: Thanks! :-)

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/dummy/ExceptionsIT.scala
gerolf-da: It might be worth adding a "noop" test case with a test title like "empty test cases because DAML LF under test doesn't support exceptions".

gerolf-da (approved):
Thank you, @cocreature!

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Tests.scala
```diff
@@ -7,12 +7,19 @@ import java.nio.file.Path
...
 import com.daml.ledger.api.testtool.suites._
+import com.daml.lf.language.LanguageVersion
+import com.daml.ledger.test.TestDar
 
 import scala.collection.SortedSet
 import scala.concurrent.duration.FiniteDuration
 
 object Tests {
 
+  private val supportsExceptions: Boolean = {
```
nmarton-da, line 18: These are temp code, right? after exceptions table (or preview?) this will be gone, right?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/dummy/ExceptionsIT.scala
cocreature: I’m not quite following. Users should never be able to run this. We just need to get this to compile but we don’t add it to the list of available test cases for an LF version that does not support it.

nmarton-da (commented):
LGTM
on the exhaustiveness of the ExceptionsIT suit regarding the feature I am not sure I am the best person to judge

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Tests.scala
cocreature: Not quite, we build the ledger API test tool for very old LF versions and those are never going to support LF versions. Once we stabilize it this will move to a stable LF version. If we ever stop building the ledger APi test tool for old LF versions, we can kill this completely.

ledger/ledger-on-sql/BUILD.bazel
```diff
@@ -245,12 +245,15 @@ da_scala_test_suite(
...
 [
+    # TODO https://github.com/digital-asset/daml/issues/8020
+    # Consider if we want to keep dev once exceptions are out of 1.dev
     conformance_test(
         name = "conformance-test-{}".format(db["name"]),
         lf_versions = [
             "stable",
             "latest",
             "preview",
+            "dev",
```
nmarton-da, line 256: I see you implemented already the fix for the TransactionCommitter, so this makes sense :+1:

nmarton-da (approved):
:shipit:

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Tests.scala
nmarton-da: Oh, well you know what you do, but going forward like this might increase this type of clutter, so eventually we might need something else. for now it's fine by me.

nmarton-da:
> LGTM
> on the exhaustiveness of the ExceptionsIT suit regarding the feature I am not sure I am the best person to judge

just read that Simon and Oggy where involved on that topic :man_facepalming: all good from my side :+1:

Issue #9397, opened 2021-04-13
Quickstart-scala build error: silencer lib not found


richardkapolnai-da:
Issuing these commands (Daml 1.11), a build error occurs:
```
daml new quickstart-scala --template quickstart-scala
cd quickstart-scala/
sbt "application/runMain com.daml.quickstart.iou.IouMain localhost 6865"
```

The output (suppressed a bit) :
```
$ sbt "application/runMain com.daml.quickstart.iou.IouMain localhost 6865"
[info] Loading settings for project quickstart-scala-build from plugins.sbt ...
[info] Loading project definition from /tmp/quickstart-scala/project
[info] Loading settings for project parent from build.sbt ...
daml.sdk.version = 1.11.1
[info] Set current project to quickstart-scala (in build file:/tmp/quickstart-scala/)
[info] Updating 
[info] Resolved  dependencies
[warn]  Note: Unresolved dependencies path:
[error] sbt.librarymanagement.ResolveException: Error downloading com.github.ghik:silencer-lib_2.12.8:1.7.1
[error]   Not found
[error]   not found: /home/richard/.ivy2/local/com.github.ghik/silencer-lib_2.12.8/1.7.1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/github/ghik/silencer-lib_2.12.8/1.7.1/silencer-lib_2.12.8-1.7.1.pom
[error]         at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:247)
[error]         at lmcoursier.CoursierDependencyResolution.$anonfun$update$35(CoursierDependencyResolution.scala:216)
[error]         at ...
[error] (scala-codegen / update) sbt.librarymanagement.ResolveException: Error downloading com.github.ghik:silencer-lib_2.12.8:1.7.1
[error]   Not found
[error]   not found: /home/richard/.ivy2/local/com.github.ghik/silencer-lib_2.12.8/1.7.1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/github/ghik/silencer-lib_2.12.8/1.7.1/silencer-lib_2.12.8-1.7.1.pom
```

tamaskalcza-da:
Apparently that library does not exist: https://mvnrepository.com/artifact/com.github.ghik/silencer-lib.

tamaskalcza-da:
[...]

PR #9401, opened 2021-04-13
Fix scala version in quickstart-scala

cocreature:
Setting this manually has resulted in #9397 where we use a scala
version that doesn’t have a silencer version. This PR changes the
quickstart template to use the scala version we also use in our builds
which should avoid them getting out of sync in the future.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):
Approved. But do we still need a Scala quickstart?

cocreature:
> Approved. But do we still need a Scala quickstart?

The scala bindings are still supported (for the deprecation period) and this is referenced in the docs so I think dropping it does more harm than good.

PR #9404, opened 2021-04-13
Add speedy rollback command

cocreature:
This isn’t yet used anywhere in the interest of keeping things small
to ease review and avoid conflicts.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
@@ -1627,6 +1627,9 @@ private[lf] final class Compiler(
       )
     case Command.LookupByKey(templateId, contractKey) =>
       LookupByKeyDefRef(templateId)(SEValue(contractKey))
+    case Command.Rollback(_) =>
+      // TODO https://github.com/digital-asset/daml/issues/8020
+      sys.error("rollback cmds are not supported")
   }
 
   private val SEUpdatePureUnit = unaryFunction(_ => SEValue.Unit)
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -105,16 +105,20 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
           globalCids,
         ) map { case (tx, meta) =>
           // Annotate the transaction with the package dependencies. Since
-          // all commands are actions on a contract template, with a fully typed
+          // all commands are actions on a contract template (no rollback nodes as roots
+          // so .packageIds is always a singleton set), with a fully typed
           // argument, we only need to consider the templates mentioned in the command
           // to compute the full dependencies.
-          val deps = processedCmds.foldLeft(Set.empty[PackageId]) { (pkgIds, cmd) =>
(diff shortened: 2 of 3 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -1627,6 +1627,9 @@ private[lf] final class Compiler(
       )
     case Command.LookupByKey(templateId, contractKey) =>
       LookupByKeyDefRef(templateId)(SEValue(contractKey))
+    case Command.Rollback(_) =>
+      // TODO https://github.com/digital-asset/daml/issues/8020
+      sys.error("rollback cmds are not supported")
```
remyhaemmerle-da, line 1632: We do not want Rollback command. 
1 - Canton will not use it.
2 - The will never appear are transaction roots.

cocreature: I’m confused, you said in Slack, that you want a version of `reinterpret` that works on speedy commands. The whole point of `reinterpret` is that it works on non-root nodes. So how do you handle rollback nodes here?

cocreature: turned into a draft until I clear up my confusion

remyhaemmerle-da: In practice, Canton needs a `reinterpret` for the a single speedy commands as they are now: create, exercise, fetch and lookup, i.e.:
```
def reinterpret(
      submitters: Set[Party],
      action: speedy.Command,
      nodeSeed: Option[crypto.Hash],
      submissionTime: Time.Timestamp,
      ledgerEffectiveTime: Time.Timestamp,
      checkAuthorization: CheckAuthorizationMode = CheckAuthorizationMode.On,
  ): Result[(SubmittedTransaction, Tx.Metadata)]
```
Because the interface does not exists, Canton currently rebuilds a dummy node for each of those 4 case. For instance for exercising a contract, Canton build an dummy exercise node with the relevant contract ID, template ID, choice ID, but without result and with no children. Then it call the reinterpret with this dummy node, which is translate into the desired command.

By changing the interface as describe above: 
 - Canton avoids the unnecessary generation of dummy node 
 - We do not have to handle (i.e. crash) Rollback node.

cocreature: I’m still confused. Canton verifies a projection. With rollback nodes, that projection can contain a rollback node at the root. How do you call `reinterpret` on such a projection if you don’t add rollback nodes to speedy commands?

remyhaemmerle-da: Canton already reimplements itself the effect of several actions by using several times the reinterpret on a single node and by managing an "artificial" state of the ledger between each of those call.
Canton will have to reimplements rollback itself in a similar way.

cocreature: Alright, that seems sensible. I think we should move some of the logic that Canton currently has into the Daml repo and maybe a speedy command makes sense at that point, but let’s wait until we actually do that before adding anything.

PR #9410, opened 2021-04-14
Support rollback nodes in scenario service

cocreature:
This focuses on the semantics rather than the display in Daml Studio
which needs more work (and seems not all that important at this
stage).

This already uncovered a bug which also applies outside of scenarios:

The consumedBy field was not affected by rollbacks which breaks the
mustBeActive check in partial transactions. This PR fixes this by
caching on try and restoring on rollback.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
+++ b/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
@@ -1,0 +1,134 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- @SINCE-LF 1.dev
+module ExceptionSemantics where
+
+import DA.Exception (throw)
+
+exception E
+  where
+    message "E"
+
+template K
+  with
+    p : Party
+    v : Int
+  where
+    signatory p
+    key (p, v) : (Party, Int)
+    maintainer key._1
+
+template T
+  with
+    p : Party
+  where
+    signatory p
+    nonconsuming choice Throw : ()
+      controller p
+      do throw E
+    nonconsuming choice Catch : ()
+      controller p
+      do try (exercise self Throw)
+         catch
+           E -> pure ()
+
+    nonconsuming choice TransientDuplicate : ()
+      with
+        i : Int
+      controller p
+      do try do
+           create (K p i)
+           create (K p i)
+           throw E
+         catch
+           E -> pure ()
+    nonconsuming choice NonTransientDuplicate : ()
+      with
+        i : Int
+      controller p
+      do try do
+           create (K p i)
+           throw E
+         catch
+           E -> pure ()
+    nonconsuming choice RollbackKey : ()
+      with
+        i : Int
+      controller p
+      do try do
+           create (K p i)
+           throw E
+         catch
+           E -> create (K p i) >> pure ()
+
+    nonconsuming choice RollbackArchive : ()
+      with
+        i : Int
+      controller p
+      do cid <- create (K p i)
+         try (archive cid >> throw E)
+         catch
+           E -> archive cid
+
+    nonconsuming choice NonRollbackArchive : ()
+      with
+        i : Int
+      controller p
+      do cid <- create (K p i)
+         try (archive cid)
+         catch
+           E -> pure ()
+         archive cid
+
+template Fetcher
+  with
+    sig : Party
+    obs : Party
+  where
+    signatory sig
+    observer obs
+    choice Fetch : K
+      with
+        cid : ContractId K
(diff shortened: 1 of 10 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/NodeInfo.scala
```diff
@@ -89,11 +89,11 @@ object NodeInfo {
 
     final def requiredAuthorizers: Set[Party] =
       // TODO https://github.com/digital-asset/daml/issues/8020
-      sys.error("rollback nodes are not supported")
+      Set.empty
 
     final def informeesOfNode: Set[Party] =
       // TODO https://github.com/digital-asset/daml/issues/8020
-      sys.error("rollback nodes are not supported")
+      Set.empty
```
cocreature, line 96: This is obviously broken but it is good enough to fix the scenario service (who doesn’t use those methods). I would propose to merge this and then fix it in a separate PR.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/scenario/ScenarioLedger.scala
```diff
@@ -427,9 +427,11 @@ object ScenarioLedger {
...
 
                   node match {
-                    case _: NodeRollback[_] =>
-                      // TODO https://github.com/digital-asset/daml/issues/8020
-                      sys.error("rollback nodes are not supported")
+                    case rollback: NodeRollback[_] =>
+                      processNodes(
+                        Right(newCache),
+                        (Some(nodeId) -> rollback.children.toList) :: idsToProcess,
+                      )
```
cocreature, line 434: This is completely broken, will spend some more time thinking about it.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
```diff
@@ -58,48 +58,53 @@ object BlindingTransaction {
         parentExerciseWitnesses: Set[Party],
         nodeId: NodeId,
     ): BlindState = {
-      val action = tx.nodes.get(nodeId) match {
+      tx.nodes.get(nodeId) match {
```
cocreature, line 61: Sorry I shouldn’t be sneaking this in here. Will try to spin it out into a separate PR with tests.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -549,6 +555,8 @@ private[lf] case class PartialTransaction(
         copy(
           context = info.parent.addRollbackChild(info.nodeId, context.nextActionChildIdx),
           nodes = nodes.updated(info.nodeId, rollbackNode),
+          // Undo any consuming choices in the rollback.
```
cocreature, line 558: Same here, will also try to get this into a separate PR with a standalone test.

cocreature: https://github.com/digital-asset/daml/pull/9430

sofiafaro-da (commented):
Tests look good to me. I can't comment on the scala side.

compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
```diff
@@ -0,0 +1,134 @@
...
+
+divulgence = scenario do
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (K p1 0)
+  divulger <- submit p2 $ create (Fetcher p2 p1)
+  fetcher <- submit p1 $ create (Fetcher p1 p2)
+  submitMustFail p2 $ exercise fetcher (Fetch cid)
+  submit p1 $ exercise divulger (RollbackFetch cid)
+  submit p2 $ exercise fetcher (Fetch cid)
```
sofiafaro-da, line 133: Interesting case!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
cocreature: https://github.com/digital-asset/daml/pull/9431

cocreature:
Dependent PRs are all in so should be ready now.

nickchapman-da (approved):
LGTM
The additions to the `unwind()` code in Speedy.scala look fine.
Nice catch!

PR #9411, opened 2021-04-14
cleanup: kill more stuff

garyverhaegen-da:
changelog_begin
changelog_end

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -132,12 +132,6 @@ jobs:
         git checkout $(release_sha)
       name: checkout_release
       condition: and(succeeded(), eq(variables.is_release, 'true'))
-    - bash: |
-        set -x
-        eval "$(dev-env/bin/dade-assist)"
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
@cocreature This seems to have fixed that machine. Do you per chance remember why we switched from `lsof` to `pgrep`? I have no recollection and unfortunately I did not record it in #8748.

cocreature:
> @cocreature This seems to have fixed that machine. Do you per chance remember why we switched from `lsof` to `pgrep`? I have no recollection and unfortunately I did not record it in #8748.

I’m not entirely sure, I remember either something about lsof not working properly or something about us not trusting the results and going for the “full” cleanup (which is apparently not full enough). I guess worts case, we’ll rediscover it.

garyverhaegen-da:
This time we had Java and daml processes lingering, so killing Bazel ones is clearly not enough. 🤷

cocreature (approved):
Thans!

Issue #9374, opened 2021-04-09
Alpha version of daml script exports


cocreature:
This is a tracking ticket for things we have to do before releasing an alpha version of daml script exports (what used to be called daml script dumps).

1. Expose it in the assistant under `daml export script`. This addresses both feedback that `dump` doesn’t sound nice and includes `script` in the name to allow for other export formats (e.g., you could imagine a binary export for certain usecases) under the same command.
2. Add some documentation. Doesn’t have to be super detailed but should at least show the basics and document the caveats that exist at the moment.

sofiafaro-da:
Since this is a ledger command, I think the command would be better exposed under `daml ledger`. Something like `daml ledger export script`, or `daml ledger export-script`, or ... something. Also it ought to pick up ledger.host and ledger.port options from daml.yaml automatically, like the other `daml ledger` commands.

What do you think? @aherrmann-da @cocreature

cocreature:
@sofiafaro-da I’m slightly leaning towards agreeing with you here but where do we draw the line? daml script, daml triggers, the JSON API, … all connect to a ledger so should they also all go under `daml ledger`? The best distinction I can come up with is that `daml ledger` commands are never long-running services (that excludes the ones I mentioned before but wouldn’t exclude script exports) and perhaps maps somewhat directly to ledger API functionality (that one is a bit more fishy for script exports).

I’d probably go for your suggestion (although I have a slight preference for `daml ledger export script` but that also seems a bit longish). I agree we should infer host and port but I don’t think that should block this PR.

sofiafaro-da:
> I’m slightly leaning towards agreeing with you here but where do we draw the line? 

That's a good question. I think that distinction makes sense (`daml ledger` commands are non-long-running commands that communicate with the ledger). I also think of it a bit like a "noun verb" sentence -- "daml ledger <verb>" means "do <verb> with a daml ledger", which is a pattern that "daml ledger export script" fits. But ultimately, I think it's something we just have to decide on a case-by-case basis.

So my argument here is that this command only works by connecting to a ledger, and we already have a series of commands that work this way that are all `daml ledger <something>`, and they all share some CLI flags / daml.yaml behavior, so I think it makes sense here.

I don't think any of this should block the PR for the alpha version of daml script exports. We can hide the export command for now, or mark it as experimental, and add the `daml ledger export` logic in daml-helper in a separate PR.

aherrmann-da:
[...]

PR #9414, opened 2021-04-14
Add daml export script command to Daml assistant

aherrmann-da:
Part of https://github.com/digital-asset/daml/issues/9374

- Renames `daml dump` to `daml export`
- Exposes exporting Daml script as a sub-command `daml export script` (for now the only subcommand)
- Exposes `daml export` in the Daml assistant
- Adds an integration test to the daml-assistant IT suite that runs `daml export script` and checks that the expected files were generated.

changelog_begin

* [Daml export] New feature: Use ``daml export script`` to generate a
  Daml script that will reconstruct a given ledger state. This is an
  early access feature.

changelog_end

Documentation for the feature will follow in a separate PR.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice work! Only open point I see is the naming of the command.

daml-assistant/daml-sdk/src/main/scala/com/digitalasset/daml/sdk/SdkMain.scala
```diff
@@ -22,6 +23,7 @@ object SdkMain {
       case "trigger" => Trigger.main(rest)
       case "script" => Script.main(rest)
       case "test-script" => TestScript.main(rest)
+      case "export" => Export.main(rest)
```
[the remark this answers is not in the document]
aherrmann-da: I've implemented the change, the command is now exposed as `daml ledger export script`. It's an entirely different way of exposing the command. So, I decided to do it right away instead of a follow-up PR, as that follow-up would essentially undo this one.

I've replicated the command-line argument parsing in optparse-applicative. The duplication is unfortunate, however, this way all arguments have user visible help strings. Otherwise, I didn't find a way to access the help of the underlying Scala option parser, even with `daml ledger export -- --help` optparse-applicative swallowed the `--help` instead of passing it through to the Scala side.

daml-script/export/src/main/scala/com/daml/script/export/Main.scala
```diff
@@ -45,29 +45,35 @@ object Main {
...
-        config.outputPath,
-        acs,
-        trees,
-        pkgRefs,
-        pkgs,
-        config.acsBatchSize,
-      )
-    } yield ()
+  ): Future[Unit] = {
+    config.exportType match {
```
stefanobaghino-da, line 49: Looks like a `fold`, but feel free to leave it as is if you think it's more readable.

daml-script/export/src/main/scala/com/daml/script/export/Main.scala
```diff
@@ -45,29 +45,35 @@ object Main {
...
+            exportScript.sdkVersion,
+            exportScript.damlScriptLib,
+            exportScript.outputPath,
+            acs,
+            trees,
+            pkgRefs,
+            pkgs,
+            exportScript.acsBatchSize,
+          )
+        } yield ()
```
stefanobaghino-da, line 69:
```suggestion
        } {
          Export.writeExport(
            exportScript.sdkVersion,
            exportScript.damlScriptLib,
            exportScript.outputPath,
            acs,
            trees,
            pkgRefs,
            pkgs,
            exportScript.acsBatchSize,
          )
        }
```
Removing the `yield` is equivalent to turning the final `map` in a `foreach`. Makes sense?

daml-script/export/src/main/scala/com/daml/script/export/Config.scala
```diff
@@ -0,0 +1,123 @@
...
+    checkConfig(c =>
+      c.exportType match {
+        case None => failure("Must specify export type")
+        case Some(_) => success
+      }
+    )
+  }
+
+  private val EmptyExportScript = ExportScript(
+    outputPath = null,
```
stefanobaghino-da, line 99: This is unlikely to be a cause of concern and I understand that this is hidden behind a required option, but is there something we can do to make errors more telling than a `NullPointerException`? Would it be feasible to have a "raw" version of this that contains `Option`s and have a function that validates it and turns it into something that can be used safely?

stefanobaghino-da (commented):
I left a few comments for the Scala part, LGTM.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -499,6 +502,45 @@ reset args = do
...
         "The reset command is currently not available for the HTTP JSON API. Please use the gRPC API."
 
+data LedgerExportFlags = LedgerExportFlags
+  { parties :: [String]
+  , start :: Maybe String
+  , end :: Maybe String
+  , export :: LedgerExportType
+  }
+
+data LedgerExportType = LedgerExportScript
```
cocreature, line 512: I don’t understand why you need to duplicate everything. `daml codegen java --help` works fine. What makes `daml ledger export script` different and forces us to duplicate the CLI parsing?

daml-assistant/integration-tests/src/DA/Daml/Assistant/IntegrationTests.hs
```diff
@@ -445,6 +445,21 @@ damlStartTests getDamlStart =
                           ]
                   contents <- readFileUTF8 (projDir </> "output.json")
                   lines contents @?= ["{", "  \"_1\": 0,", "  \"_2\": 1", "}"]
+        , testCase "daml export script" $ do
+              DamlStartResource {projDir} <- getDamlStart
+              withTempDir $ \exportDir -> do
+                  withCurrentDirectory projDir $ do
+                      callCommand $ unwords
+                          [ "daml ledger export"
+                          , "--party Alice"
```
cocreature, line 454: Does moving `--party` after `script` work as well? I expect for users it’s probably more intuitive in most cases to throw everything at the end.

daml-script/export/src/main/scala/com/daml/script/export/Config.scala
cocreature: Definitely feasible, I tend to go for the `null/""/…` approach personally mainly because I’m lazy and I’m angry at scopt for making me implement my own validation logic (Haskell’s optparse-applicative is much nicer). That said, that’s a pretty poor excuse so definitely open to reconsider this. But let’s leave it out of this PR, this is just reshuffling.

cocreature (approved):
Nice! I think I don’t fully understand the need to replicate CLI parsing but I’m fine with merging it in this state either way.

daml-script/export/src/main/scala/com/daml/script/export/Main.scala
aherrmann-da: I may be misunderstanding, but this doesn't type-check. The function's return type is `Future[Unit]`, but without the `yield` this returns `Unit`.

daml-script/export/src/main/scala/com/daml/script/export/Main.scala
aherrmann-da: Yeah, personally I find the `match` version more readable. But, I have no strong opinion on this.

daml-assistant/integration-tests/src/DA/Daml/Assistant/IntegrationTests.hs
aherrmann-da: That doesn't work by default. There's an optparse-applicative option ([`subparserInline`](https://hackage.haskell.org/package/optparse-applicative-0.16.1.0/docs/Options-Applicative.html#v:subparserInline)) to enable this. It's a global switch though. At a quick test it seems to cause issues with other commands. E.g. it breaks `daml start` with 
```
Invalid option `--dar'
Usage: daml run-jar JAR [--logback-config ARG] [ARG]
```

The documentation states
> NOTE: When this option is used, preferences for the subparser which effect the parser behaviour (such as noIntersperse) are ignored.

IIUC this also disables `forwardOptions` which we rely on to forward flags to wrapped executables.

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
aherrmann-da: That's good pointer, thanks. I had looked at `daml ledger navigator` instead, which has the same problem. `daml codegen java|scala|...` still at least replicates the command parsing, i.e. the `java|scala|...` part. But, it doesn't replicate the flags themselves. I'll look into avoiding the duplication in a follow-up.

PR #9416, opened 2021-04-14
LF: distninguish Rollback from Action Nodes

remyhaemmerle-da:
The change was already asked by @oggy-. It is actually very useful. 

CHANGELOG_BEGIN
CHANGELOG_END

this is part of #8020 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/NodeInfo.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/NodeInfo.scala
@@ -85,16 +85,4 @@ object NodeInfo {
       keyMaintainers
   }
 
-  trait Rollback extends NodeInfo {
-
-    final def requiredAuthorizers: Set[Party] =
-      // TODO https://github.com/digital-asset/daml/issues/8020
-      sys.error("rollback nodes are not supported")
-
-    final def informeesOfNode: Set[Party] =
-      // TODO https://github.com/digital-asset/daml/issues/8020
-      sys.error("rollback nodes are not supported")
-
-  }
-
 }
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
@@ -281,13 +281,16 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
 
   // TODO: https://github.com/digital-asset/daml/issues/8020
   //  for now we assume that rollback node cannot be a root of a transaction.
+  @throws[IllegalArgumentException]
   def rootNodes: ImmArray[Node.GenActionNode[Nid, Cid]] =
     roots.map(nid =>
       nodes(nid) match {
         case action: Node.GenActionNode[Nid, Cid] =>
           action
         case _: Node.NodeRollback[_] =>
-          sys.error("illed-formed transaction")
+          throw new IllegalArgumentException(
+            s"invalid transaction, root refers to a Rollback node $nid"
+          )
       }
     )
 
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
@@ -142,7 +142,7 @@ final case class GenTransaction[Nid, +Cid](
                       nr.children ++: nids
                     },
                   )
-                case _: Node.LeafOnlyNode[Cid] => go(newErrors, newVisited, nids)
+                case _: Node.LeafOnlyActionNode[Cid] => go(newErrors, newVisited, nids)
                 case ne: Node.NodeExercises[Nid, Cid] =>
                   go(
                     newErrors,
@@ -279,6 +279,18 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
 
   def roots: ImmArray[Nid]
 
+  // TODO: https://github.com/digital-asset/daml/issues/8020
+  //  for now we assume that rollback node cannot be a root of a transaction.
+  def rootNodes: ImmArray[Node.GenActionNode[Nid, Cid]] =
+    roots.map(nid =>
+      nodes(nid) match {
+        case action: Node.GenActionNode[Nid, Cid] =>
+          action
+        case _: Node.NodeRollback[_] =>
+          sys.error("illed-formed transaction")
+      }
+    )
+
   /** This function traverses the transaction tree in pre-order traversal (i.e. exercise node are traversed before their children).
     *
     * Takes constant stack space. Crashes if the transaction is not well formed (see `isWellFormed`)
@@ -293,7 +305,7 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
         f(nodeId, node)
         node match {
           case nr: Node.NodeRollback[Nid] => go(nr.children ++: toVisit)
-          case _: Node.LeafOnlyNode[Cid] => go(toVisit)
+          case _: Node.LeafOnlyActionNode[Cid] => go(toVisit)
           case ne: Node.NodeExercises[Nid, Cid] => go(ne.children ++: toVisit)
         }
     }
@@ -336,7 +348,7 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
         node match {
           case nr: Node.NodeRollback[Nid] =>
             go(nr.children.map(_ -> newPathState) ++: toVisit)
-          case _: Node.LeafOnlyNode[Cid] => go(toVisit)
+          case _: Node.LeafOnlyActionNode[Cid] => go(toVisit)
           case ne: Node.NodeExercises[Nid, Cid] =>
             go(ne.children.map(_ -> newPathState) ++: toVisit)
         }
@@ -372,7 +384,7 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
   final def foreachInExecutionOrder(
       exerciseBegin: (Nid, Node.NodeExercises[Nid, Cid]) => Boolean,
       rollbackBegin: (Nid, Node.NodeRollback[Nid]) => Boolean,
-      leaf: (Nid, Node.LeafOnlyNode[Cid]) => Unit,
+      leaf: (Nid, Node.LeafOnlyActionNode[Cid]) => Unit,
       exerciseEnd: (Nid, Node.NodeExercises[Nid, Cid]) => Unit,
       rollbackEnd: (Nid, Node.NodeRollback[Nid]) => Unit,
   ): Unit = {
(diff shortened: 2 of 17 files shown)


remyhaemmerle-da:
cc @oliverse-da @oggy-

gerolf-da:
Thanks @remyhaemmerle-da. I'm wondering whether it would be useful to have something like: 
```scala
// pseudo code
case class VersionedTransaction(....) {

  def actionsOnly = VersionedTransaction(version, nodes.collect { case (nid, node: Node.GenActionNode) => (nid, node) } )
}
```

However, there's also something to be said for being very explicit about how rollback nodes are handled in each case, than just outright ignoring them via `tx.actionsOnly.foreach(...)`.

cocreature:
> Thanks @remyhaemmerle-da. I'm wondering whether it would be useful to have something like:

I think at least while we’re rolling it out we benefit from being explicit everywhere.

remyhaemmerle-da:
> Thanks @remyhaemmerle-da. I'm wondering whether it would be useful to have something like:
> 
> ```scala
> // pseudo code
> case class VersionedTransaction(....) {
> 
>   def actionsOnly = VersionedTransaction(
>     version, 
>     nodes.collect { case (nid, node: Node.GenActionNode) => (nid, node) },
>    roots,
>   )
> }
> ```
> 
> However, there's also something to be said for being very explicit about how rollback nodes are handled in each case, than just outright ignoring them via `tx.actionsOnly.foreach(...)`.

I though about defining similar method. I am not 100% is a good idea because it is not clear what we want : 
- Do we want all the action Nodes 
- Do we want all the action Nodes that are not rolled back.

For now, I prefer to not have to make a choice, and postpone the implementation of such method once we are cleaning up the prototype. 

What it sure is that I will not go for your exact definition ;) :
- it produces a non well-formed transaction (the rolled back action are dangling in  actionsOnly.nodes)
- I would rather implement something like `forEachAction(a: GenActionNode[Nid, Cid] => Unit)`  that have a more convenient type signature.

gerolf-da:
I agree with both your points, @remyhaemmerle-da, @moritzkiefer-da. I also didn't expect my 5 second draft would be any good ;)

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/NodeInfo.scala
```diff
@@ -85,16 +85,4 @@ object NodeInfo {
       keyMaintainers
   }
 
-  trait Rollback extends NodeInfo {
```
cocreature, line 88: :fire:

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -279,6 +279,18 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
 
   def roots: ImmArray[Nid]
 
+  // TODO: https://github.com/digital-asset/daml/issues/8020
+  //  for now we assume that rollback node cannot be a root of a transaction.
```
cocreature, line 283: I thought the plan was to keep assuming that because the only case where this is not the case is projections and we handle those separately?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -279,6 +279,18 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
...
+  def rootNodes: ImmArray[Node.GenActionNode[Nid, Cid]] =
+    roots.map(nid =>
+      nodes(nid) match {
+        case action: Node.GenActionNode[Nid, Cid] =>
+          action
+        case _: Node.NodeRollback[_] =>
+          sys.error("illed-formed transaction")
```
cocreature, line 290:
```suggestion
          sys.error("ill-formed transaction")
```
Assuming my comment above is correct, we should also throw something sensible like an IllegalArgumentException here instead of `sys.error`.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
```diff
@@ -200,7 +200,15 @@ private[apiserver] final class ApiSubmissionService private[services] (
   )(implicit loggingContext: LoggingContext): Future[Seq[SubmissionResult]] =
     if (configuration.implicitPartyAllocation) {
       val partiesInTransaction =
-        transaction.nodes.valuesIterator.flatMap(_.informeesOfNode).toSeq.distinct
+        transaction.nodes.valuesIterator
+          .flatMap {
+            // TODO https://github.com/digital-asset/daml/issues/8020
```
cocreature, line 205: rollback nodes don’t have informees so this seems perfectly fine.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
```diff
@@ -113,8 +114,13 @@ private[appendonlydao] object PostCommitValidation {
       def foldInformees[T](tx: CommittedTransaction, init: T)(
           f: (T, Party) => T
       ): T =
-        tx.fold(init) { case (accum, (_, node)) =>
-          node.informeesOfNode.foldLeft(accum)(f)
+        tx.fold(init) {
+          case (accum, (_, node: LFNode.GenActionNode[_, _])) =>
+            node.informeesOfNode.foldLeft(accum)(f)
+          case (accum, (_, _: LFNode.NodeRollback[_])) =>
+            // TODO https://github.com/digital-asset/daml/issues/8020
```
cocreature, line 121: Same comment above, there are no informees so this is correct afaict.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/PostCommitValidation.scala
```diff
@@ -113,8 +114,13 @@ private[dao] object PostCommitValidation {
       def foldInformees[T](tx: CommittedTransaction, init: T)(
           f: (T, Party) => T
       ): T =
-        tx.fold(init) { case (accum, (_, node)) =>
-          node.informeesOfNode.foldLeft(accum)(f)
+        tx.fold(init) {
+          case (accum, (_, node: LFNode.GenActionNode[_, _])) =>
+            node.informeesOfNode.foldLeft(accum)(f)
+          case (accum, (_, _: LFNode.NodeRollback[_])) =>
+            // TODO https://github.com/digital-asset/daml/issues/8020
```
cocreature, line 121: and here

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -408,8 +409,13 @@ private[kvutils] class TransactionCommitter(
       def foldInformeeParties(tx: Tx.Transaction, init: Boolean)(
           f: (Boolean, String) => Boolean
       ): Boolean =
-        tx.fold(init) { case (accum, (_, node)) =>
-          node.informeesOfNode.foldLeft(accum)(f)
+        tx.fold(init) {
+          case (accum, (_, node: Node.GenActionNode[_, _])) =>
+            node.informeesOfNode.foldLeft(accum)(f)
+          case (accum, (_, _: NodeRollback[_])) =>
+            // TODO https://github.com/digital-asset/daml/issues/8020
```
cocreature, line 416: and again same comment

cocreature (approved):
Nice work, thank you!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
remyhaemmerle-da: My concern is more about the rolled back nodes. Do we need to take into account the actions that have been rolled back ? It is not clear to me and I prefer to think properly about that in a different PR.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
remyhaemmerle-da: Same answer than above.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/PostCommitValidation.scala
remyhaemmerle-da: also here.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
remyhaemmerle-da: If the answer is the same for the 4 places, I will rather implement directly a method `informees` in GenTransaction class

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: :+1:

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: Yes. I just want to keep track of this and eventually reconsider it before releasing #8020.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
cocreature: Perfectly fine to do it in a separate PR, I’ll leave comments on the individual instances here anyway so the thoughts I put into it don’t get lost:

Implicit party allocation here: We clearly also need to allocate parties in rolled back nodes.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
cocreature: This is the counterpoint to implicit party allocation that validates all parties have been allocated. Parties in rolled back nodes also need to be taken into account.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/PostCommitValidation.scala
cocreature: Duplicate of the one above for non-appendonly so same answer.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
cocreature: this is the kvutils version of `validateParties` above so same answer.

As I said, perfectly fine to do this in a separate PR. I don’t really like adding `informees`. This might be fine now but it seems easy to end up changing something later or adding a new function where rollback nodes shouldn’t be treated the same. Rollback nodes really don’t have informees.

PR #9419, opened 2021-04-15
Don't repeat the Oracle password in ci/build.yml

stefanobaghino-da:
Got it from https://stackoverflow.com/a/13802438

Incidentally, to write this commit message containing a single quote, I had to escape it with this https://stackoverflow.com/a/1250279

I guess today is "bring-your-quotation-marks-to-work day".

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -230,7 +230,7 @@ jobs:
         }
         trap cleanup EXIT
         testConnection() {
-            docker exec oracle bash -c 'sqlplus -L system/hunter2@//localhost:1521/ORCLPDB1 <<< "select * from dba_users;"; exit $?' >/dev/null
+            docker exec oracle bash -c 'sqlplus -L system/'"$ORACLE_PWD"'@//localhost:1521/ORCLPDB1 <<< "select * from dba_users;"; exit $?' >/dev/null
         }
         until testConnection
         do


ci/build.yml
```diff
@@ -230,7 +230,7 @@ jobs:
         }
         trap cleanup EXIT
         testConnection() {
-            docker exec oracle bash -c 'sqlplus -L system/hunter2@//localhost:1521/ORCLPDB1 <<< "select * from dba_users;"; exit $?' >/dev/null
+            docker exec oracle bash -c 'sqlplus -L system/'"$ORACLE_PWD"'@//localhost:1521/ORCLPDB1 <<< "select * from dba_users;"; exit $?' >/dev/null
```
cocreature, line 233: We could also use double quotes for the outer one and then single quotes for the `select` and escape the `$?`. Not sure that’s clearer, wdyt?

cocreature (approved):
*******

ci/build.yml
garyverhaegen-da:
```suggestion
            docker exec oracle bash -c <<BASH
                sqlplus -L system/$ORACLE_PWD@//localhost:1521/ORCLPDB1 <<< "select * from dba_users;" >/dev/null
            BASH
```

I think this should also work, though it's arguably a slightly different semantics:

```
docker exec oracle bash -c "sqlplus -L system/$ORACLE_PWD@//localhost:1521/ORCLPDB1" <<< "select * from dba_users;" >/dev/null
```

garyverhaegen-da:
Note that the branch name is misleading: today you learned that _there is no way to expand a variable within single quotes_ and that the workaround is to close the single quotes before expanding the variable.

ci/build.yml
cocreature: I think the `exit` stuff was required but I forgot why :thinking:

garyverhaegen-da: Bash returns the exit code of the last command by default AFAIK so I don't think it does anything here. Maybe the command used to be more complex?

cocreature: I thought it was some mix of `docker exec` and `bash -c` that required it but I’ve probably just messed it up at the time.

stefanobaghino-da: I don't have any specific opinion, but this works and seems fairly readable.

stefanobaghino-da:
> Note that the branch name is misleading: today you learned that _there is no way to expand a variable within single quotes_ and that the workaround is to close the single quotes before expanding the variable.

😂 

Indeed.

😂

PR #9425, opened 2021-04-15
remove Array[String] instances from Oracle json-api driver layer

S11001001:
We avoid having actual doobie-provided DB-specific instances in the dbbackend, but we still want the set of implicits to _depend_ on the specific driver chosen. We represent that by having Postgres depend on Array instances, while Oracle does not. This lets us clean up after #9423 by removing the never-used fake Array instances from the Oracle `SupportedJdbcDriver`.

### Pull Request Checklist

trigger the build.


cocreature (approved):
lgtm

PR #9361, opened 2021-04-08
LF: Move KV's InputsAndEffects to LF land

remyhaemmerle-da:
We move the code from computing from KV to LF. 
Testing + more refactoring will follows in subsequent PRs.  

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9429, opened 2021-04-16
KV: do not use "Inputs" part of InputsAndEffects

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


miklos-da (commented):
@remyhaemmerle-da If you merge #9361 you won't need this PR. Am I missing something?

remyhaemmerle-da:
> @remyhaemmerle-da If you merge #9361 you won't need this PR. Am I missing something?

I think it is better to go this way. Review #9361 and make it only about Effect and make this PR about Inputs only.

In the concrete case we do not need a special computation of Input because almost everything (expect contract key) is already in GenTransaction. 

I tried to address some of the comment you make in #9361, here.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -384,6 +384,24 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
       case (acc, _) => acc
     } -- localContracts.keySet
 
+  /** Return all the contract keys created, exercised, fetched or lookup by this transation.
+    */
+  final def contractKeys[Cid2 >: Cid]: Set[(Ref.Identifier, Value[Cid2])] = {
```
cocreature, line 389: Should we have a test for this somewhere?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -384,6 +384,24 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
       case (acc, _) => acc
     } -- localContracts.keySet
 
+  /** Return all the contract keys created, exercised, fetched or lookup by this transation.
```
cocreature, line 387: Can we be a bit more precise here? What exactly does `all` mean? Does it include transient contracts, created and archived within this transaction? That probably then also defines what we do with rollback nodes.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueSubmission.scala
```diff
@@ -44,14 +46,28 @@ class KeyValueSubmission(metrics: Metrics) {
...
+      val packageIdStates = meta.optUsedPackages
+        .getOrElse(
+          throw new InternalError("Transaction was not annotated with used packages")
+        )
+        .map(Conversions.packageStateKey)
+      val partyStates =
+        (tx.informees ++ submitterInfo.actAs).toList.map(Conversions.partyStateKey)
+      val contractIdStates =
+        tx.inputContracts[ContractId].map(Conversions.contractIdToStateKey)
+      val contractKeyStates =
```
cocreature, line 59: Why do we treat contract keys created with the transaction as inputs but don’t do the same for contract ids? That seems a bit inconsistent. I assume it has something to do with duplicate key detection or something along those lines?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
cocreature: fine with merging without tests given that there don’t really seem to be tests for the inputs and effects logic either but we should fix this so maybe add a todo.

cocreature (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueSubmission.scala
remyhaemmerle-da: As far I understand, for validation, KV have to collect all the info needed to replay the transaction. For created contract it should know the contract key does not exists.

cocreature: alright so the input is the “non-existence of the key” or something like that, makes sense :+1:

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: done

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: done.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -268,6 +272,106 @@ class TransactionSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPro
 
     }
   }
+
+  "contractKeys" - {
+    "return all the contract key" in {
```
remyhaemmerle-da, line 277:
```suggestion
    "return all the contract keys" in {
```

daml-lf/transaction-test-lib/src/main/scala/lf/transaction/test/TransactionBuilder.scala
```diff
@@ -49,6 +49,8 @@ final class TransactionBuilder(
     nodes(parentId) match {
       case _: TxExercise | _: TxRollback =>
         children += parentId -> (children(parentId) :+ nodeId)
+      case _: TxRollBack =>
+        children += parentId -> (children(parentId) :+ nodeId)
```
SamirTalwar-DA, line 53: Is this clause ever hit, given the above one?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -384,6 +384,27 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
...
     } -- localContracts.keySet
 
+  /** Return all the contract keys references by this transaction.
+    * This includes the keys created, exercised, fetched, or lookup, even those
+    * that refer to transient contracts or that appear under a roolback node.
+    */
+  final def contractKeys(implicit
+      ev: HasTxNodes[Nid, Cid] <:< HasTxNodes[_, Value.ContractId]
+  ): Set[GlobalKey] = {
+    ev(this).fold(Set.empty[GlobalKey]) {
```
SamirTalwar-DA, line 394: I'd find this a little prettier with some kind of `foldMap`. Is it worth implementing that?

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -268,6 +272,108 @@ class TransactionSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPro
...
```
SamirTalwar-DA, line 277:
```suggestion
    "return all the contract keys" in {
```

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
cocreature: Could be sensible but I’d prefer keeping it for a separate PR.

daml-lf/transaction-test-lib/src/main/scala/lf/transaction/test/TransactionBuilder.scala
cocreature: good point, I think this was a botched merge, fixed in 0667cc76a35204a2165b1395ab54d5c457376900

SamirTalwar-DA (approved):
I'm happy but if I were you, I'd wait for @miklos-da or @fabiotudone-da to verify. I am not _too_ familiar with this part of the code.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueSubmission.scala
```diff
@@ -44,14 +46,24 @@ class KeyValueSubmission(metrics: Metrics) {
...
     metrics.daml.kvutils.submission.conversion.transactionToSubmission.time { () =>
-      val inputDamlStateFromTx = InputsAndEffects.computeInputs(tx, meta)
       val encodedSubInfo = buildSubmitterInfo(submitterInfo)
+      val packageIdStates = meta.optUsedPackages
+        .getOrElse(
+          throw new InternalError("Transaction was not annotated with used packages")
+        )
+        .map(Conversions.packageStateKey)
+      val partyStates =
+        (tx.informees ++ submitterInfo.actAs).toList.map(Conversions.partyStateKey)
```
fabiotudone-da, line 56: How about encapsulating `(tx.informees ++ submitterInfo.actAs)` too somehow/somewhere?

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -268,6 +272,108 @@ class TransactionSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPro
 
     }
   }
+
+  "contractKeys" - {
+    "return all the contract keys" in {
+      // TODO: https://github.com/digital-asset/daml/issues/8020
+      // change VDev to  TransactionVersion.StableVersions.max once exception are released
```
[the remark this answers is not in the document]
cocreature: We generally keep these todos around so it’s easy to grep for them before closing the issue.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueSubmission.scala
cocreature: 5f95fd0b3abaf0e7715307993edd36d476ff6e8f if you have a better name for it, I’m happy to change it. I considered `partyStates` but that doesn’t seem much better.

fabiotudone-da (approved):

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -384,6 +384,27 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
       case (acc, _) => acc
     } -- localContracts.keySet
 
+  /** Return all the contract keys referenced by this transaction.
+    * This includes the keys created, exercised, fetched, or looked up, even those
+    * that refer transient contracts or that appear under a rollback node.
+    */
+  final def contractKeys(implicit
+      ev: HasTxNodes[Nid, Cid] <:< HasTxNodes[_, Value.ContractId]
```
miklos-da, line 392: What is an `ev`? Can you use meaningful parameter names, please?

SamirTalwar-DA: `ev` is pretty conventional in Scala-ese for "evidence".

I prefer the latter, but I think it's quite common (kind of like "i" for an iterator or "e" for an exception).

cocreature: https://github.com/digital-asset/daml/pull/9442

miklos-da: Thanks for the explanation, @SamirTalwar-DA.

remyhaemmerle-da: The method [Iterable#toMap](https://github.com/scala/scala/blob/8684ae833dcfeac6107343fcca5501301e509eef/src/library/scala/collection/GenTraversableOnce.scala#L648) is the standard example of method using such implicit (and syntax)

remyhaemmerle-da (commented):
LGTM

PR #9431, opened 2021-04-16
Add tests for blindinginfo

cocreature:
Most of this PR is just tests for blinding info in general since there
seem to be basically none. The actual change to blinding info is
rather simple.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/BlindingSpec.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/BlindingSpec.scala
@@ -1,0 +1,196 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf
+package engine
+
+import com.daml.lf.data.ImmArray
+import com.daml.lf.data.Ref.Party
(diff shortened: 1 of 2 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/BlindingInfo.scala
```diff
@@ -10,8 +10,9 @@ import com.daml.lf.value.Value.ContractId
 /** A transaction's blinding information, consisting of disclosure and
   * divulgence info.
   *
-  * "Disclosure" tells us which transaction nodes to communicate to which
-  * parties.
+  * "Disclosure" tells us which transaction actions to communicate to which
```
cocreature, line 13: I wonder if blinding info should also store information about which nodes have been rolled back or if that’s better left to something else. Either way, I’d postpone that decision until we get to the usesites of blinding info. For the scenario service usecase, we only need this.

remyhaemmerle-da (approved):
Scala looks good. I would like a second opinion on the logic as I have never been quite sure of all the details of the Blinding.

cocreature:
> Scala looks good. I would like a second opinion on the logic as I have never been quite sure of all the details of the Blinding.

will drop the rollback from this PR and reduce it to just tests and then open a separate on just for rollback semantics.

PR #9432, opened 2021-04-16
Don't duplicate export script option parsing

aherrmann-da:
Forward all flags to the wrapped executable instead of handling them in daml-helper with optparse-applicative.

Addresses https://github.com/digital-asset/daml/pull/9414#discussion_r614620153

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
+++ b/daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
@@ -512,7 +512,7 @@ runLedgerExport flags remainingArguments = do
         args <- getDefaultArgs flags
         -- TODO[AH]: Use parties from daml.yaml by default.
         -- TODO[AH]: Use SDK version from daml.yaml by default.
-        pure $ ["--host", host args, "--port", show (port args)]
+        pure ["--host", host args, "--port", show (port args)]
     withJar
       damlSdkJar
       [logbackArg]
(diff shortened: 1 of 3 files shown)


daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -502,42 +500,23 @@ reset args = do
...
-          ]
+    let isHelp = any (\x -> x `elem` ["-h", "--help"]) remainingArguments
+    ledgerFlags <- if isHelp then
+        -- Don't use getDefaultArgs here so that --help can be used outside a daml project.
+        pure []
+      else do
+        args <- getDefaultArgs flags
+        -- TODO[AH]: Use parties from daml.yaml by default.
+        -- TODO[AH]: Use SDK version from daml.yaml by default.
+        pure $ ["--host", host args, "--port", show (port args)]
```
cocreature, line 515: maybe we should switch to reading those from the Scala side to make this a bit nicer. On the other hand, it would duplicate some of that logic so not sure :/

cocreature (approved):
Thanks!

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
aherrmann-da: Yes, might be worth the trouble. I'll look into it in the context of the todo notes about the other inferred flags.

sofiafaro-da (approved):
Looks great, thank you!

PR #9433, opened 2021-04-16
KV: do not use "Effects" part of InputsAndEffects

remyhaemmerle-da:
The PR is based one #9429.
Together with #9429 this make InputsAndEffects useless.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


cocreature:
turning this into a draft, this one needs some more work to be ready for review.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -374,6 +374,97 @@ class TransactionSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPro
...
+      }
+      def exercise(create: Node.NodeCreate[ContractId], consuming: Boolean) =
+        builder.exercise(
+          contract = create,
+          choice = "C",
+          actingParties = parties.toSet,
+          consuming = consuming,
+          argument = V.ValueUnit,
+          byKey = false,
+        )
```
fabiotudone-da, line 402: Perhaps it's possible to share these utilities between test cases? That would also make them shorter and so more readable.

fabiotudone-da (approved):

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
cocreature: Factored out in 60a00a51a2b598954a75da07efc7de1f3e8c9fe3. Probably could do some more aggressive factoring but I’ll leave that for a future PR.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -364,13 +364,30 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
...
     }
 
+  /** Returns the IDs of all the consumed contracts.
+    *  This includes transient contracts but it does not include contracts
+    *  consumed in rollback nodes.
+    */
+  final def consumedContracts[Cid2 >: Cid]: Set[Cid2] =
+    foldInExecutionOrder(Set.empty[Cid2])(
+      exerciseBegin = (acc, _, exe) => {
+        val acc0 = if (exe.consuming) { acc + exe.targetCoid }
```
miklos-da, line 381: You might want to use braces around all the if branches -- the indentation looks confusing otherwise. You could even simplify it by returning directly from the branches:
```scala
  if (exe.consuming) {
    (acc + exe.targetCoid, true)
  } else {
    (acc, true)
  }
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -14,7 +14,14 @@ import com.daml.ledger.participant.state.v1.{PackageId, SubmitterInfo}
 import com.daml.lf.data.Ref.{Identifier, LedgerString, Party}
 import com.daml.lf.data.Relation.Relation
 import com.daml.lf.data.Time
-import com.daml.lf.transaction.{GlobalKey, _}
+import com.daml.lf.transaction.{
+  BlindingInfo,
```
miklos-da, line 18: :+1:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -448,30 +447,31 @@ private[kvutils] class TransactionCommitter(
...
   ): Unit = {
-    val effects = InputsAndEffects.computeEffects(transactionEntry.transaction)
-    val cid2nid: Value.ContractId => NodeId =
-      transactionEntry.transaction.localContracts
+    val localContracts = transactionEntry.transaction.localContracts
+    val consumedContracts = transactionEntry.transaction.consumedContracts
+    val contractKeys = transactionEntry.transaction.updatedContractKeys
     // Add contract state entries to mark contract activeness (checked by 'validateModelConformance').
-    for ((key, createNode) <- effects.createdContracts) {
+    localContracts.foreach { case (cid, (nid, createNode)) =>
```
miklos-da, line 454: Why not use for comprehension for better readability?
```suggestion
    for((cid, (nid, createNode)) <- localContracts) {
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -497,9 +497,10 @@ private[kvutils] class TransactionCommitter(
     }
     // Update contract keys.
     val ledgerEffectiveTime = transactionEntry.submission.getLedgerEffectiveTime
-    for ((key, contractKeyState) <- effects.updatedContractKeys) {
+    contractKeys.foreach { case (contractKey, contractKeyState) =>
```
miklos-da, line 500: Same here regarding for comprehension:
```suggestion
    for ((contractKey, contractKeyState) <- contractKeys) {
```

miklos-da (approved):

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
cocreature: 413ba40cbbdb20b51d31575ccadbd3551e3f323b

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
remyhaemmerle-da: Let me expose the original motivation for the change I made here.

- @stephencompall-DA strongly advices to never use the pattern matching in the left-hand side of a `<-` inside a for comprehension as it is not translate as you expect. (It think the reason is explain [here](https://github.com/lampepfl/dotty/issues/2573, @stephencompall-DA please correct me if I mistake). 
If you want to keep using for comprehension, you should rather use 
```
for {
  entry <- effects.createdContracts) 
  ((key, createNode) = entry 
} { ... }
``` 

- I personally prefer avoiding for-comprehension for imperative code.

remyhaemmerle-da (commented):
Thanks @cocreature

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
S11001001: Correct, and I wrote a mini-article about it [on our forum](https://discuss.daml.com/t/pattern-always-adds-withfilter-scala-ritual/1463). And I also agree with your motivation for avoiding the `for` without `yield`.

PR #9438, opened 2021-04-19
Fix flakiness in //daml-lf/transaction

cocreature:
The current generators are a bit too dumb and relatively
frequently (~1/10) fail to generate 5 successful results before
failing forever. This PR changes things slightly to avoid this. We
could be even more clever and draw the second version after having
selected the first one so we cannot even draw the same but that
doesn’t seem worth the complexity for now.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


garyverhaegen-da (approved):

remyhaemmerle-da (commented):
LGTM

PR #9443, opened 2021-04-19
Test for duplicate contracts when querying on behalf of multiple parties

### LINKED ISSUE #9388: Oracle json-api: use array types for signatories/observer, or test for duplicates

stefanobaghino-da:
Fixes #9388, tests #9423.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -891,7 +891,6 @@ abstract class AbstractWebsocketServiceIntegrationTest
             _ = inside(sharedAccountObject) { case JsObject(obj) =>
               inside((obj get "owners", obj get "number")) {
                 case (Some(JsArray(owners)), Some(JsString(number))) =>
-                  // FIXME Question for the reviewer: should we make guarantees regarding the ordering?
                   owners should contain theSameElementsAs Vector(JsString("Alice"), JsString("Bob"))
                   number shouldBe sharedAccountNumber
               }
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -20,7 +20,7 @@ import scalaz.syntax.std.option._
 import scalaz.syntax.tag._
 import scalaz.syntax.traverse._
 import scalaz.{-\/, \/-}
-import spray.json.{JsNull, JsObject, JsString, JsValue}
+import spray.json.{JsArray, JsNull, JsObject, JsString, JsValue}
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Future}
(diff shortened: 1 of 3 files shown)


ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -857,6 +857,68 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+                uri,
+                headers = headersWithPartyAuth(aliceAndBob),
+              )
+            )
+            _ = status shouldBe a[StatusCodes.Success]
+            ContractDelta(Vector((_, sharedAccountObject)), _, Some(lastSeenOffset)) <- readOne
+            _ = inside(sharedAccountObject) { case JsObject(obj) =>
+              inside((obj get "owners", obj get "number")) {
+                case (Some(JsArray(owners)), Some(JsString(number))) =>
+                  // FIXME Question for the reviewer: should we make guarantees regarding the ordering?
```
S11001001, line 894: No, it's irrelevant to the test.

S11001001 (approved):
Looks good, but I assume this failed prior to #9423 (this can be checked by applying the following)

```diff
diff --git a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
index f0da5aa58bf..23e5c961563 100644
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -728,7 +728,7 @@ private object OracleQueries extends Queries {
                     AND $queriesCondition"""
       // see https://github.com/digital-asset/daml/issues/9388#issuecomment-820538688
       // for a demonstration
-      val q = sql"SELECT $outerSelectList FROM ($dupQ) WHERE rownumber = 1"
+      val q = sql"SELECT $outerSelectList FROM ($dupQ) WHERE 1 = 1"
       q.query[
         (String, Mark0, JsValue, JsValue, Option[String], Option[String], Option[String])
       ].map { case (cid, tpid, key, payload, agreement, pctSignatories, pctObservers) =>
```

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Thanks, comment removed in https://github.com/digital-asset/daml/pull/9443/commits/760b54f901b4311c4a3566ec8e3a46bf64569c6a

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -857,6 +857,67 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+          for {
+            ContractDelta(Vector(), _, Some(liveStartOffset)) <- readOne
+            (status, _) <- liftF(
+              postCreateCommand(
+                sharedAccountCreateCommand(aliceAndBob, sharedAccountNumber),
+                encoder,
+                uri,
+                headers = headersWithPartyAuth(aliceAndBob),
+              )
+            )
```
S11001001, line 888: Any `create` following the `liveStartOffset` message is _guaranteed_ not to go through SQL. To guarantee that a create _does_ go through SQL, you must sequence the `postCreateCommand` future to complete before the `runWith` step on line 915. So if you move your create down there instead of delaying it with the `Consume` language's `liftF`, you should be golden.

stefanobaghino-da: Oh, I thought everything went through the database. This actually explains why breaking the query entirely (`WHERE 1 = 2`) didn't break the test. Thanks, I'll see if I can make it work tomorrow.

stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/9443/commits/d3eb4fdf82b15164521ee49cc506d67afde14b46 and checked that the test does indeed catch a regression with `WHERE 1 = 1` as mentioned above. Thanks for double checking and clarifying how to make this work.

PR #9445, opened 2021-04-19
Support rollback nodes in BlindingInfo

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
@@ -57,51 +57,55 @@ object BlindingTransaction {
         state0: BlindState,
         parentExerciseWitnesses: Set[Party],
         nodeId: NodeId,
-    ): BlindState = {
-      val action = tx.nodes.get(nodeId) match {
+    ): BlindState =
+      tx.nodes.get(nodeId) match {
         case Some(action: Node.GenActionNode[NodeId, ContractId]) =>
-          action
-        case Some(_: Node.NodeRollback[_]) =>
-          // TODO https://github.com/digital-asset/daml/issues/8020
-          throw sys.error("Rollback node not supported")
+          val witnesses = parentExerciseWitnesses union action.informeesOfNode
+
+          // nodes of every type are disclosed to their witnesses
+          val state = state0.discloseNode(witnesses, nodeId)
+
+          action match {
+
+            case _: Node.NodeCreate[ContractId] => state
+            case _: Node.NodeLookupByKey[ContractId] => state
+
+            // fetch & exercise nodes cause divulgence
+
+            case fetch: Node.NodeFetch[ContractId] =>
+              state
+                .divulgeCoidTo(parentExerciseWitnesses -- fetch.stakeholders, fetch.coid)
+
+            case ex: Node.NodeExercises[NodeId, ContractId] =>
+              val state1 =
+                state.divulgeCoidTo(
+                  (parentExerciseWitnesses union ex.choiceObservers) -- ex.stakeholders,
+                  ex.targetCoid,
+                )
+
+              ex.children.foldLeft(state1) { (s, childNodeId) =>
+                processNode(
+                  s,
+                  witnesses,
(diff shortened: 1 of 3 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/BlindingInfo.scala
```diff
@@ -13,6 +13,9 @@ import com.daml.lf.value.Value.ContractId
   * "Disclosure" tells us which transaction nodes to communicate to which
   * parties.
   * See https://docs.daml.com/concepts/ledger-model/ledger-privacy.html
+  * Note that rollback nodes are a bit special here: Even if the node
+  * itself is not disclosed, a party will be disclosed if a node was below a
+  * rollback node. That information is not included in blinding info at this point.
```
cocreature, line 18: We might want to revisit this later but this is enough for the scenario service and sandbox classic and I’d rather start simple and add something later if we do end up needing it.

nickchapman-da (approved):
LGTM :+1: 

One small thing. I see that following the `GenActionNode` / `RollbackNode` split, we have removed `NodeInfo` from rollback nodes. So consequently, unlike other node types, there is no `informeesOfNode` (or `requiredAuthorizers`) for a rollback node. Is this correct?

cocreature:
> One small thing. I see that following the `GenActionNode` / `RollbackNode` split, we have removed `NodeInfo` from rollback nodes. So consequently, unlike other node types, there is no `informeesOfNode` (or `requiredAuthorizers`) for a rollback node. Is this correct?

Exactly, only actions have informees. Rollback nodes are not actions. However, we can still disclose a rollback node to the informees of the parent exercise.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
```diff
@@ -57,51 +57,55 @@ object BlindingTransaction {
...
+          // nodes of every type are disclosed to their witnesses
+          val state = state0.discloseNode(witnesses, nodeId)
 
-      action match {
+          action match {
 
-        case _: Node.NodeCreate[ContractId] => state
-        case _: Node.NodeLookupByKey[ContractId] => state
+            case _: Node.NodeCreate[ContractId] => state
+            case _: Node.NodeLookupByKey[ContractId] => state
```
remyhaemmerle-da, line 71:
```suggestion
    case _: Node.NodeCreate[ContractId] |  _: Node.NodeLookupByKey[ContractId] => state
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
```diff
@@ -57,51 +57,55 @@ object BlindingTransaction {
...
+          // Rollback nodes are disclosed to the witnesses of the parent exercise.
+          val state = state0.discloseNode(parentExerciseWitnesses, nodeId)
+          rollback.children.foldLeft(state) { (s, childNodeId) =>
             processNode(
               s,
-              witnesses,
+              parentExerciseWitnesses,
               childNodeId,
             )
           }
```
remyhaemmerle-da, line 103:
``` suggestion
          rollback.children.foldLeft(state) ((s, childNodeId) =>
            processNode(
              s,
              parentExerciseWitnesses,
              childNodeId,
            )
          )
```

or 

``` suggestion
          rollback.children.foldLeft(state)(processNode(_, parentExerciseWitnesses, _))
```

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/BlindingInfo.scala
remyhaemmerle-da: I agree, but I would have added a  `// TODO https://github.com/digital-asset/daml/issues/8020` with this comment inside the code.

remyhaemmerle-da (commented):
LGTM

remyhaemmerle-da:
part of #8020

PR #9446, opened 2021-04-19
Document daml ledger export script

aherrmann-da:
Closes https://github.com/digital-asset/daml/issues/9374

Adds a chapter on `daml ledger export script` under the "Early Access Features" documentation.

changelog_begin

* [Daml export] Refer to the "Ledger Export" chapter under the "Early
  Access Features" for a description of the new Daml ledger export
  command. This is an early access feature.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/tools/export.rst
+++ b/docs/source/tools/export.rst
@@ -1,0 +1,164 @@
+.. Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
+Ledger Export
+#############
+
+Export is currently an :doc:`Early Access Feature in Labs status </support/status-definitions>`.
+
+Introduction
+************
+
+Daml ledger exports read the transaction history or active contract set (ACS)
+from the ledger and write it to disk encoded as a :doc:`Daml Script
+</daml-script/index>` that will reproduce the ledger state when executed. This
+can be useful to migrate the history or state of a ledger from one network to
+another, or to replicate the ledger state locally for testing or debugging
+purposes.
+
+Usage
+*****
+
+The command to generate a Daml ledger export has the following form. ::
+
+  daml ledger export <format> <options>
+
+Right now Daml script, specified as ``script``, is the only supported export
+format. You can get an overview of the available command-line options using the
+``--help`` flag as follows. ::
+
+  daml ledger export script --help
+
+A full example invocation looks like this: ::
+
+  daml ledger export script --host localhost --port 6865 --party Alice --party Bob --output ../out --sdk-version 0.0.0
+
+The flags ``--host`` and ``--port`` define the Daml ledger to connect to. You
+can omit these flags if you are invoking the command from within a Daml project
+with a running ledger, e.g. with a running ``daml start``.
+
+The ``--party`` flags define which contracts will be included in the export. In
+the above example only contracts visible to the parties Alice and Bob will be
+included in the export. Lack of visibility of certain events may cause exports to fail, see :ref:`caveats <export-caveats>`.
+
+The ``--output`` flag defines the directory prefix under which to generate the
+Daml project that contains the Daml script that represents the ledger export.
+The flag ``--sdk-version`` defines which Daml Connect version to configure in
+the generated ``daml.yaml`` configuration file.
+
+By default an export will reproduce all transactions in the ledger history. The
+:ref:`ledger offsets <export-ledger-offsets>` section describes how to change
+this behavior.
+
+Output
+******
+
+The generated Daml code contains the following top-level definitions:
+
+``data Parties``
+  A record that holds the relevant Daml parties.
+``allocateParties : Script Parties``
+  A Daml script that allocates fresh parties on the ledger and returns them in
+  the ``Parties`` record.
+``export : Parties -> Script ()``
+  The Daml ledger export encoded as a Daml script. Given the relevant parties
+  this script will reproduce the ledger state when executed. You can read this
+  script to understand the exported ledger state or history, and you can modify
+  this script for debugging or testing purposes.
+``testExport : Script ()``
+  A Daml script that will first invoke ``allocateParties`` and then ``export``.
+
+In most simple cases the generated Daml script will use the functions
+``submit`` or ``submitMulti`` to issue ledger commands that reproduce a
+transaction or ACS. In some cases the generated Daml script
+will fall back to the more general functions ``submitTree`` or
+``submitTreeMulti``.
+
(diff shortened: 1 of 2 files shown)


docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+        owner = alice_0
+      Main.ChoiceA
+  let contractB_1_1 = fromTree tree $
+        exercised @Main.ContractA "ChoiceA" $
+        created @Main.ContractB
+  let contractB_1_2 = fromTree tree $
+        exercised @Main.ContractA "ChoiceA" $
+        createdN @Main.ContractB 1
+
+.. TODO[AH] Add a full example project and example export.
```
aherrmann-da, line 98: This could be expanded to a more complete example that includes an initial Daml script to create an example history and then display the resulting Daml script exports for different configurations. This could be imported via `literalinclude` from a working example similar to how other documentation chapters do, to make sure that the documentation stays up-to-date and correct. However, this requires a bit more work, so I'll leave that for a separate PR.

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
+.. Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
+Ledger Export
+#############
+
+Export is currently an :doc:`Early Access Feature in Labs status </support/status-definitions>`.
```
cocreature, line 7: I think this should be Alpha. We do want to get this into a stable feature. See https://docs.daml.com/support/status-definitions.html for the definitions. cc @bame-da

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+
+A full example invocation looks like this: ::
+
+  daml ledger export script --host localhost --port 6865 --party Alice --party Bob --output ../out --sdk-version 0.0.0
+
+The flags ``--host`` and ``--port`` define the Daml ledger to connect to. You
+can omit these flags if you are invoking the command from within a Daml project
+with a running ledger, e.g. with a running ``daml start``.
+
+The ``--party`` flags define which contracts will be included in the export. In
```
cocreature, line 40: definitely not for this PR but might be nice to accept comma-separated lists as well.

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+the above example only contracts visible to the parties Alice and Bob will be
+included in the export. Lack of visibility of certain events may cause exports to fail, see :ref:`caveats <export-caveats>`.
+
+The ``--output`` flag defines the directory prefix under which to generate the
```
cocreature, line 44: (not for this PR) We probably need to check what we do if the directory already exists or if it is not empty. Probably best to fail with a good error by default and maybe have a --force flag.

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+Daml project that contains the Daml script that represents the ledger export.
+The flag ``--sdk-version`` defines which Daml Connect version to configure in
```
cocreature, line 46: Note for other reviewers: We’re already planning to infer this

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+By default an export will reproduce all transactions in the ledger history. The
+:ref:`ledger offsets <export-ledger-offsets>` section describes how to change
+this behavior.
+
+Output
+******
+
+The generated Daml code contains the following top-level definitions:
+
+``data Parties``
```
cocreature, line 58: Side note: There is no nice mapping from old party ids to new party ids atm. One solution for that would be to accept a TextMap instead of a record as an argument with the keys being the old party ids.

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+``allocateParties : Script Parties``
+  A Daml script that allocates fresh parties on the ledger and returns them in
+  the ``Parties`` record.
+``export : Parties -> Script ()``
+  The Daml ledger export encoded as a Daml script. Given the relevant parties
+  this script will reproduce the ledger state when executed. You can read this
+  script to understand the exported ledger state or history, and you can modify
+  this script for debugging or testing purposes.
+``testExport : Script ()``
+  A Daml script that will first invoke ``allocateParties`` and then ``export``.
```
cocreature, line 69:
```suggestion
  A Daml script that will first invoke ``allocateParties`` and then ``export``. This can be useful to test it in Daml Studio.
```

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+  Transactions up to and including the start offset will be reproduced as a
+  sequence of create commands that reproduce the ACS as of the start offset.
+  Later transactions will be reproduced as seen by the configured parties. In
+  particular, ``--start end`` will reproduce the current ACS but no transaction
+  history, ``--start start`` (the default) will reproduce the history of all
+  transactions as seen by the configured parties.
+``--end``
+  Export transactions up to and including this end offset.
+
+.. TODO[AH] Provide a reference or hints how to obtain arbitrary ledger offsets.
```
cocreature, line 121: see https://github.com/digital-asset/daml/pull/9257/files#r602506698 for something that might help

docs/source/tools/export.rst
```diff
@@ -0,0 +1,164 @@
...
+*******
+
+Unknown Contract Ids
+====================
+
+Daml ledger export does currently not handle references to unknown contract
+ids, but will just fail if any such references are encountered. This may occur
+if a contract was divulged to one of the configured parties, but the event that
+initially created that contract is not visible to any of the configured
+parties. This may also occur if a contract was archived before the configured
```
cocreature, line 135: Maybe worth clarifying that this is only an issue if the contract was archived _and_ you also retained a reference to it. Archived contracts in general are fine.

cocreature (approved):
Nice work, thanks!

docs/source/tools/export.rst
aherrmann-da: Oh, copy pasted from the wrong place. Fixed.

docs/source/tools/export.rst
aherrmann-da: Yes, currently overwriting fails with
```
java.nio.file.FileAlreadyExistsException: ../dump-out/deps
```
We should handle this properly, produce a good error message. `--force` sounds useful too.

docs/source/tools/export.rst
aherrmann-da: Indeed, I've raised a ticket for this: https://github.com/digital-asset/daml/issues/9453

PR #9447, opened 2021-04-19
Use sequential id for contract and key activeness checks

tudor-da:
This PR optimizes activeness checks for contracts and keys performed by the `ContractsReader` against the index db, by using the current cache index (from `MutableCacheBackedContractStore`) as the ledger end bound in SQL queries.

**NOTE**: The optimization is targeted at the mutable contract state cache. The legacy `dao` schema and the translation cache do not leverage this optimization.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsReader.scala
@@ -20,6 +20,7 @@ import com.daml.platform.store.appendonlydao.events.SqlFunctions.{
   PostgresSqlFunctions,
 }
 import com.daml.platform.store.appendonlydao.DbDispatcher
+import com.daml.platform.store.cache.MutableCacheBackedContractStore.EventSequentialId
 import com.daml.platform.store.interfaces.LedgerDaoContractsReader
 import com.daml.platform.store.serialization.{Compression, ValueSerializer}
 
@@ -150,13 +151,14 @@ private[appendonlydao] sealed class ContractsReader(
   override def lookupActiveContractAndLoadArgument(
       readers: Set[Party],
       contractId: ContractId,
+      ledgerEndSequentialId: Option[EventSequentialId] = None,
   )(implicit loggingContext: LoggingContext): Future[Option[Contract]] = {
 
     Timed.future(
       metrics.daml.index.db.lookupActiveContract,
       dispatcher
         .executeSql(metrics.daml.index.db.lookupActiveContractDbMetrics) { implicit connection =>
-          lookupActiveContractAndLoadArgumentQuery(contractId, readers)
+          lookupActiveContractAndLoadArgumentQuery(contractId, readers, ledgerEndSequentialId)
             .as(contractRowParser.singleOpt)
         }
         .map(_.map { case (templateId, createArgument, createArgumentCompression) =>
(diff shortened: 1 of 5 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsReader.scala
```diff
@@ -203,28 +206,36 @@ private[appendonlydao] sealed class ContractsReader(
...
   ) = {
     val tree_event_witnessesWhere =
       sqlFunctions.arrayIntersectionWhereClause("tree_event_witnesses", readers)
+
+    val ledgerEndLimit =
+      ledgerEndSequentialId.map(_.toString).getOrElse("parameters.ledger_end_sequential_id")
     SQL"""
   WITH archival_event AS (
          SELECT participant_events.*
            FROM participant_events, parameters
```
rautenrieth-da, line 235: You are unconditionally joining with the parameters table. Maybe something like this?
```suggestion
           FROM participant_events#${ledgerEndSequentialId.map(_ => "").getOrElse(", parameters")}
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsReader.scala
```diff
@@ -344,16 +359,24 @@ private[appendonlydao] sealed class ContractsReader(
            """
   }
 
-  private def lookupContractKeyQuery(readers: Set[Party], key: Key): SimpleSql[Row] = {
+  private def lookupContractKeyQuery(
+      readers: Set[Party],
+      key: Key,
+      // TODO Remove SQL references to `parameters` table when removing the `ledgerEndSequentialId` parameter optionality
```
rautenrieth-da, line 365:
```suggestion
      // TODO append-only: Remove SQL references to `parameters` table when removing the `ledgerEndSequentialId` parameter optionality
```

same change at other places

rautenrieth-da (commented):
This PR just introduces the new `ledgerEndSequentialId` parameter to a few methods, but the parameter is never used, right? I would have expected a change at the call site(s) of `lookupActiveContractAndLoadArgument` and `lookupActiveContractWithCachedArgument`.

tudor-da:
> This PR just introduces the new `ledgerEndSequentialId` parameter to a few methods, but the parameter is never used, right? I would have expected a change at the call site(s) of `lookupActiveContractAndLoadArgument` and `lookupActiveContractWithCachedArgument`.

Totally right, thanks. I was focused too much on how not to impact the legacy `dao`, it seems

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsReader.scala
tudor-da: I was thinking it has no impact if the table is not actually queried. Fine to add it as well.

meiersi-da (changes requested):
@tudor-da : the construction looks good. However, I can't see the changes for changing the contract key lookups, am I overlooking something?

tudor-da:
> @tudor-da : the construction looks good. However, I can't see the changes for changing the contract key lookups, am I overlooking something?

After an offline discussion, we've agreed that the changes applicable here have less impact than what was initially considered. Optimizing the read queries to not join with the `parameters` table will be revisited later

Issue #9283, opened 2021-03-30
Ledger Pruning is incompatible with the use of divulged contracts


bame-da:
The new Ledger Pruning feature causes problems for applications that use divulged contracts as part of interpretation. Since participants may never learn of the archival of a divulged contracts, the only sensible behaviour is to prune all contracts that were only divulged before the pruning offset. However, there is no guarantee that these contracts are actually archived. Thus a fetch on an active contract may fail during interpretation. This behaviour is inconsistent with an unpruned Sandbox and thus harms portability and the separation of concerns between node operations (pruning) and application development.

As a remedy, we should deprecate the use of divulged contracts during interpretation and explicitly say in the pruning documentation that pruning is incompatible with the use of this behaviour. This is a pure documentation change.

With the introduction of explicit disclosure, we should consider disabling this behaviour by default, only maintaining a "legacy" mode in which divulged contracts are usable.

cocreature:
Do we want to deprecate this before we have explicit divulgence? It seems nicer to wait until that is ready.

bame-da:
I don't like having incompatibility between two stable features. And if we know we'll kill a feature, I think it's better to be quick about marking it as such.

cocreature:
Alright, no strong objection from me. Just seemed nicer to have a clear alternative to migrate to when something gets deprecated.

meiersi-da:
Thanks @bame-da . I agree with your reasoning.

daravep:
@bame-da what is the intended timeline for this? can we assume this will be the default mode in our stack by Q3/2021?

bame-da:
The reason I opened this ticket is that I think we should deprecate the use of divulged contracts in interpretation, and add a warning to the pruning docs asap. We do not want anyone to start relying on both pruning and the use of divulged contracts. 

Making the change to actually prune away the divulged contracts can then happen whenever it makes sense to slot it in, but at the latest with the Explicit Disclosure work which is currently targeted for Q3.

cocreature:
Closing this, I think we’ve resolved this.

PR #9451, opened 2021-04-20
Detect transient duplicate key errors in speedy

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
```diff
@@ -44,6 +45,14 @@ final case class ContractNotFound(ci: ContractId) extends Error {
   override def detailMsg: String = msg
 }
 
+/** See com.daml.lf.transaction.Transaction.DuplicateContractKey
+  *    for more information.
+  */
+final case class DuplicateContractKey(key: GlobalKey) extends Error {
```
meiersi-da, line 51: @moritzkiefer-da : do you think it would be possible / sensible to also report the contract-ids on which the collision occurs?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -193,14 +214,15 @@ private[lf] object PartialTransaction {
...
+    contractKeyUniqueness: ContractKeyUniquenessMode,
     submissionTime: Time.Timestamp,
     nextNodeIdx: Int,
     nodes: HashMap[NodeId, PartialTransaction.Node],
     actionNodeSeeds: BackStack[(NodeId, crypto.Hash)],
     consumedBy: Map[Value.ContractId, NodeId],
     context: PartialTransaction.Context,
     aborted: Option[Tx.TransactionError],
-    keys: Map[GlobalKey, Option[Value.ContractId]],
+    keys: Map[GlobalKey, Option[(Value.ContractId, PartialTransaction.VisibleByKey)]],
```
meiersi-da, line 225: What about the alternative of storing the stakeholders together with the contract-id, so that on lookup the authorization rules can be executed?

I believe the current option might be conflating two separate concepts:
1. What is the slice of the ledger state that is accessible to the submitter for command interpretation, i.e., all contracts visbile to `actAs \union readAs`. I believe this has to only be checked outside the engine, as its purpose is to ensure tha the submitter does not learn about data from co-hosted parties. One can see that when thinking about disclosure, where we plan to explicitly expand the amount of state available to interpretation.
2. The authorization rules of fetch and lookupByKey, whose purpose it is to avoid privacy leaks through the ability of a submitter to ask for validation results on lookups touching contracts on which the submitter never got any kind of delegation.

meiersi-da: Slight clarification on point 1: given the diffrence in fetch vs https://docs.daml.com/daml/reference/contract-keys.html#visiblebykey, we actually define two slices of ledger state:
1. The slice for resolving key lookups: all contracts where one of `actAs \union readAs` is a stakeholder
2. The slice for resolving fetches: the one from 1. extended with all divulged contracts.

I still believe that it would be valuable if we could organize our code such that these checks on ledger state access are the responsibility of the caller of the engine, while the engine ensures that all produced transactions are internally consistent.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -657,6 +657,25 @@ object Transaction {
...
+    * In particular, a standalone create will never produce a duplicate key
+    * error if the global contract with the duplicate key is not used in the same
+    * transaction.
+    *
+    * At the moment, Speedy also only if the two contracts are actually active at the same time.
+    * It makes no attempts to detect contract key inconsistencies, e.g.,
+    * a local create with key A, followed by an archive, followed by a global fetch
+    * with key A. This is just a more complex version of the standalone create where
+    * the global key is not fetched at all and should be caught by the ledger.
+    */
```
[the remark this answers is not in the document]
meiersi-da: Thanks for the ping @cocreature . Do you want to say that you the engine only checks internal consistency wrt contract keys (cf. https://docs.daml.com/concepts/ledger-model/ledger-integrity.html#internal-consistency) and this error signals an `InternallyInconsistentContractKey`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -592,18 +620,39 @@ private[lf] case class PartialTransaction(
   def isConsumed(coid: Value.ContractId): Boolean = consumedBy.contains(coid)
 
   /** Guard the execution of a step with the unconsumedness of a
-    * `ContractId`
+    * `ContractId` and check that there is no other contract with the
+    *  same contract key.
```
meiersi-da, line 624: This feels like a strange place to check for key collisions.

The only statement that can introduce a key collision is a create, which is where I'd expect the check to be.

The updates of the `keys` map should happen in the respective mutating commands: create (insert key), archive (set key to None), fetch (insert key if present), lookupByKey (insert result in key map).

meiersi-da (changes requested):
Hi @cocreature thanks for tackling this important area. I've requested changes, as I'd like us to make sure we're aligned on
1. submitter read-authorization checks vs. tx-internal-authorization
2. where to check key uniqueness and where to update the key mapping

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
cocreature: > @moritzkiefer-da : do you think it would be possible / sensible to also report the contract-ids on which the collision occurs?

Maybe but we don’t do that atm in the ledgers that do this and for now I’m interested in getting consistency and fixing the bug not in exposing more information.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: > The only statement that can introduce a key collision is a create, which is where I'd expect the check to be.

This is not true. I can have a `create` first and then a `fetch` of a global contract with the same key. Yes the ledger should also reject this but this is still an internal duplicate key error where there are two contracts with the same key active at the same time and imho speedy should reject this.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: > What about the alternative of storing the stakeholders together with the contract-id, so that on lookup the authorization rules can be executed?

This is not about authorization. We already validate this correctly for everything (afaik). This is about the visibility check for contract key lookups (at least one reader must be a stakeholder). Note that there is a bug here already, see https://github.com/digital-asset/daml/issues/9454. Working on a fix for that next which will address the TODOS around the Transient visibility.

Let me walk you through what `ptx.keys` tracked before and what it tracks now:
Before: `ptx.keys` tracked successful global key lookups and creates. For key lookups/fetches, the ledger already validates visibility. For creates it doesn’t (since it is not involved) and the engine cannot do it atm (it doesn’t know readAs parties) leading to the bug mentioned above.
After: `ptx.keys` now tracks the keys of all contracts used in the current transaction regardless of whether they come from a fetch by key, a regular fetch, local, …. This means we can easily detect duplicate key errors: If we ever get two different entries in `ptx.keys`, we have an error. It does however mean that the caching `ptx.keys` was used for needs to take visibility into account, if we know about the contract from a regular fetch, we might still need the ledger to check for visibility.

I’d argue that “keys of all contracts used in the current transaction” is a pretty clear concept and is not conflating anything. The visibility is indeed separate and we could track this is a separate map. But I’d rather have the two together, that makes it easier to make sure we don’t forget to look at visibility in places where we have to and avoids them getting out of sync.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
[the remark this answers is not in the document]
cocreature: Not really sure what you mean by that exactly. The ledger model does not define a concept of internal key consistency. 

What I’m trying to say is that we detect if we observe two contracts being active with the same key at the same time. 

What we do not do atm is to make inferences about past activeness. In particular, for a global contract we know that even if we have a later fetch it must have been active at the beginning of the transaction as well, so the following transaction will not be rejected by speedy:
```
cid <- create c -- with key K
archive cid
fetch globalCid -- with key k
```
 We could detect those and while I don’t feel strongly about this, I’d be very open to agreeing that speedy should reject this as well but I’d prefer to keep it out of this PR since it’s somewhat orthogonal and this is large enough as it is.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
meiersi-da: Ack. Agreed.

meiersi-da: Sorry for raising this unnecessarily.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
meiersi-da: Here our mental models disagree. What I'd expect speedy to do is to read the key from the ledger as part of the create so that it can check that the key is not already assigned.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -29,6 +30,24 @@ private[lf] object PartialTransaction {
...
+  sealed trait VisibleByKey
+  object VisibleByKey {
+    // We know about the key but we do not know if it is visible to
+    // actAs union readAs. This happens for example, if a contract
+    // that has a contract key was fetched but not fetched by key.
+    // This can be used for duplicate key detection but key lookups
+    // and fetches have to throw SResultNeedKey.
+    final case object Unknown extends VisibleByKey
+    // We fetched or looked up the contract by key via SResultNeedKey
+    // so we know it is visible and can be useud to cache
```
andreaslochbihler-da, line 42:
```suggestion
    // so we know it is visible and can be used to cache
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
moritzkiefer-da: I’m open to rejecting this case, see my comment below. I don’t think we should reject a single `create` without the `fetch` later. This is an internally consistent transaction as in there exists a ledger state it could be applied to. The ledger should make sure that the transaction being created by speedy can actually be applied to the current ledger state.

meiersi-da: Thinking more about my proposal, I believe:
1. it works to detect contract key inconsistencies between the transaction and any contracts whose stakeholders intersect with `actAs \union readAs`
2. it does not detect contract key inconsistencies between the transaction and divulged contracts visible to `actAs \union readAs` that are fetched within the transaction

I'm wondering whether we should aim to detect the conflicts from Case 2, as divulgence is going to go away #9283.

@cocreature @andreaslochbihler-da what do you think?

cocreature: If I understand you correctly, you are suggesting that instead of doing the validation on every fetch & exercise you only do it on a global fetch or exercise? I don’t understand how that affects divulgence. A global fetch could still fetch a divulged contract.

I actually started out with that approach and abandoned for a few reasons:
1. I generally would like to limit different treatment of global & local operations as much as possible. This is exactly how we end up with these type of errors in the first place. I trust the solution here much more to catch all cases and to be robust against future changes.
2. In addition to not making it safer, it doesn’t seem to make anything simpler, you still have to check fetches and exercises which we already check. But now you need logic to track local contracts in `PartialTransaction` and you have some extra code in fetches/exercises to skip the check (which stays exactly the same we have atm).

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
meiersi-da: Thank you for the detailed explanation @cocreature . It makes sense to me now, and I retract my comment.

If you could work your explanation wrt what is being tracked 'keys of all contracts used in the current transaction' into the code somehow, that would likely be helpful for future readers like me :blush:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
meiersi-da: Yes, having made another pass on your implementation I agree. Thanks for your kind explanations.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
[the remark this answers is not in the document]
meiersi-da: Ignore my comment wrt internal key consistency. The definition in https://docs.daml.com/concepts/ledger-model/ledger-integrity.html#internal-consistency, reads

> A transaction is *internally consistent* if it is internally consistent for all contracts and consistent for all keys.

So, it refers to the (global) consistency notion for keys, which matches with what you are aiming for. Your example
```
cid <- create c -- with key K 
archive cid 
fetch globalCid -- with key k
```
does though not satisfy /key consistency/ and should therefore be rejected by the engine as the resulting transaction is not internally consistent.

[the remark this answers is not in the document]
cocreature: Makes sense, as I said, I’m totally fine with rejecting this in speedy. I would just prefer to keep it out of this PR since it only makes things more confusing.

meiersi-da (approved):
@cocreature : thanks for the constructive discussion. My comments are adressed and the PR looks good to me under the assumption that you'll address the remaining undetected key consistency case in this or another PR.

daml-lf/tests/BasicTests.daml
```diff
@@ -615,4 +615,81 @@ template NoMaintainer
...
+    choice DuplicateGlobalFetchByKey : ()
+      controller p
+      do _ <- fetchByKey @WithKey (p, 42)
+         cid1 <- create (WithKey p 42)
+         archive cid1
+
+    choice DuplicateGlobalFetchByKeyAfter : ()
+      controller p
+      do cid1 <- create (WithKey p 42)
+         _ <- fetchByKey @WithKey (p, 42)
```
simonmaxen-da, line 648: Many familiar with database transactions would expected fetchByKey return cid1 (no duplicate).

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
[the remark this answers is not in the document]
simonmaxen-da: Thanks, good to see this tightened up.

daml-lf/tests/BasicTests.daml
cocreature: It does. This does not fail in speedy, it is internally consistent. But it should fail in the ledger if there already exists another contract at the point of creation of `cid1`.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: expanded the docs in d610e8818a4f465a0c8cd70900e85070f566bccb

daml-lf/tests/BasicTests.daml
simonmaxen-da: Cool.  I see now this is a success criteria not a failure one.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -180,27 +201,31 @@ private[lf] object PartialTransaction {
...
+  *              successfully looked up by key and if it was created in the transaction.
+  *              We know it is not active if we observed a consuming exercise on a contract with
+  *              this key or a negative lookup by key.
+  *              In addition to activeness, we also need to track visibility-by-key to ensure
+  *              that fetch-by-key and lookup-by-key only succeed if (actAs union readAs) contains
+  *              a stakeholder. For fetch-by-key and lookup-by-key of global contracts this
+  *              is validated by the ledger, for fetches and exercises we do not know this information
+  *              so we have to ask the legder but can cache it afterwards.
+  *              Note that all this information is restored after a rollback node,
+  *              so it only tracks currently active/not-active keys.
```
meiersi-da, line 216: @cocreature : there might be a problem with the reset in the code
```
  private def resetActiveState(state: ActiveLedgerState): PartialTransaction =
    copy(consumedBy = state.consumedBy, keys = state.keys)
```
at it seems to be too strong. I'd expect it to miss the duplicate key violation in the following example
```
rollback [
   create c -- with key k
   abort
]
create c -- with key k
```
What am I overlooking?

meiersi-da: (changed example to fetch a global key)

cocreature: Yes, that succeeds atm. I’d put that in the same bucket as the one with the `fetchByKey` after an archive of a local contract, if you want to put a name on it something like “assertions about activeness of global contracts”.

cocreature: In other words, there are two types of checks:
1. We know about two contracts being active at the current state. This is what is implemented in this PR and is a somewhat local check in that it only refers to the current state.
2. We observed a key of global contract and a create with the same key anywhere. This a global check in the sense that the order of those doesn’t matter. The fetch can be anywhere including in a rollback node and the same holds for the create.

cocreature:
Following a discussion with @meiersi-da and @andreaslochbihler-da I’ll revise this a bit so turning into a draft in the meantime.

PR #9456, opened 2021-04-20
more robust macOS cleanup

garyverhaegen-da:
We've recently seen a few cases where the macOS nodes ended up not having the cache partition mounted. So far this has only happened on semi-broken nodes (guest VM still up and running but host unable to connect to it), so I haven't been able to actually poke at a broken machine, but I believe this should allow a machine in such a state to recover.

While we haven't observed a similar issue on Linux nodes (as far as I'm aware), I have made similar changes there to keep both scripts in sync.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/clean-up.yml
+++ b/ci/clean-up.yml
@@ -7,29 +7,8 @@ steps:
 
     eval "$(dev-env/bin/dade-assist)"
 
-    exec 1> >(while IFS= read -r line; do echo "$(date -Is) [out]: $line"; done)
-    exec 2> >(while IFS= read -r line; do echo "$(date -Is) [err]: $line"; done >&2)
-
-    # Location of the disk cache for CI servers set in their init files:
-    # infra/macos/2-common-box/init.sh:echo "build:darwin --disk_cache=~/.bazel-cache" > ~/.bazelrc
-    # infra/vsts_agent_linux_startup.sh:echo "build:linux --disk_cache=~/.bazel-cache" > ~/.bazelrc
-
-    case $(uname) in
-    Linux)
-        cache_1=/home/vsts/.cache/bazel
-        cache_2=/home/vsts/.bazel-cache
-    ;;
-    Darwin)
-        cache_1=/var/tmp/_bazel_vsts
-        cache_2=/Users/vsts/.bazel-cache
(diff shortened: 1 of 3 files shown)


ci/clean-up.yml
```diff
@@ -7,29 +7,8 @@ steps:
...
-    ;;
-    esac
-
-    for cache in $cache_1 $cache_2; do
-        for pid in $(lsof $cache | sed 1d | awk '{print $2}' | sort -u); do
-            kill -s KILL $pid
-        done
-    done
+    exec 1> >(while IFS= read -r line; do echo "$(date -uIs) [out]: $line"; done)
+    exec 2> >(while IFS= read -r line; do echo "$(date -uIs) [err]: $line"; done >&2)
```
garyverhaegen-da, line 11: If you're looking for the change in those two lines, it's the `-u` option to `date`.

infra/vsts_agent_ubuntu_20_04_startup.sh
```diff
@@ -117,7 +117,14 @@ reset_cache() {
             echo "Killing $pid..."
             kill -s KILL $pid
         done
-        umount $mount_point
+        for pid in $(lsof $mount_point | sed 1d | awk '{print $2}' | sort -u); do
+            echo "Killing $pid..."
+            kill -s KILL $pid
+        done
+        if mount -l | grep $mount_point; then
```
cocreature, line 124: When do we expect this to fail?

garyverhaegen-da: > When do we expect this to fail?

I'm not sure. I don't understand why we get into that state on macOS nodes sometimes. Currently my best guess is that a build gets cancelled in the middle of this script, i.e. it has unmounted the partition already but hasn't mounted the new one yet.

Under that assumption, this could happen on Linux too.

cocreature (approved):
Thanks!

SamirTalwar-DA (approved):

aherrmann-da (approved):

garyverhaegen-da:
In the process of deploying; I'll merge once this reflects production.

PR #9461, opened 2021-04-20
DPP-335 Add missing compression flag

rautenrieth-da:
This PR adds a missing compression flag to the exercise event tables of the append-only schema.


rautenrieth-da:
@nmarton-da @tudor-da This was not caught by any test, I assume we don't test the compression flags, or at least not the compression of the create key value in exercise events?

nmarton-da (approved):
LGTM

PR #9468, opened 2021-04-21
Extend Transaction proto format for rollback nodes.

nickchapman-da:
Part of #8020.

Extend Transaction proto format for rollback nodes.

- proto format
- encode/decode
- testcase
- always `allowRollback` from scalagen testing; and so remove control flag
- two more `8020` TODOs, for other code which decodes the proto

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
+++ b/daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
@@ -45,6 +45,7 @@ message Node {
         NodeFetch fetch = 4;
         NodeExercise exercise = 5;
         NodeLookupByKey lookup_by_key = 6;
+        NodeRollback rollback = 7;
     }
 
     string version = 63; // *since 11*, optional
@@ -110,6 +111,10 @@ message NodeLookupByKey {
     KeyWithMaintainers key_with_maintainers = 2;
     reserved 3; // was contract_id
     com.daml.lf.value.ContractId contract_id_struct = 4;
+}
+
+message NodeRollback {
+    repeated string children = 1; // node ids
 }
 
 // architecture-handbook-entry-end: Nodes
--- a/daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
+++ b/daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
@@ -45,7 +45,8 @@ message Node {
         NodeFetch fetch = 4;
         NodeExercise exercise = 5;
         NodeLookupByKey lookup_by_key = 6;
-        NodeRollback rollback = 7;
+        // TODO https://github.com/digital-asset/daml/issues/8020
+        NodeRollback rollback = 7; //*since DEV* // TODO: update to version when exceptions land
     }
 
     string version = 63; // *since 11*, optional
@@ -113,7 +114,8 @@ message NodeLookupByKey {
     com.daml.lf.value.ContractId contract_id_struct = 4;
 }
 
-message NodeRollback {
+// TODO https://github.com/digital-asset/daml/issues/8020
+message NodeRollback { // *since DEV* // TODO: update to version when exceptions land
     repeated string children = 1; // node ids
 }
 
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
@@ -260,9 +260,11 @@ object TransactionCoder {
         nodeBuilder.setVersion(node.version.protoValue)
 
       node match {
-        case _: NodeRollback[_] =>
-          // TODO https://github.com/digital-asset/daml/issues/8020
-          sys.error("rollback nodes are not supported")
+        case ne @ NodeRollback(_, _) =>
+          val builder = TransactionOuterClass.NodeRollback.newBuilder()
+          ne.children.foreach { id => builder.addChildren(encodeNid.asString(id)); () }
+          Right(nodeBuilder.setRollback(builder).build())
+
         case nc @ NodeCreate(_, _, _, _, _, _, _, _, _) =>
           val builder = TransactionOuterClass.NodeCreate.newBuilder()
           nc.stakeholders.foreach(builder.addStakeholders)
@@ -449,6 +451,18 @@ object TransactionCoder {
     val nodeId = decodeNid.fromString(protoNode.getNodeId)
 
     protoNode.getNodeTypeCase match {
+      case NodeTypeCase.ROLLBACK =>
+        val protoRollback = protoNode.getRollback
+        val childrenOrError = protoRollback.getChildrenList.asScala
+          .foldLeft[Either[DecodeError, BackStack[Nid]]](Right(BackStack.empty[Nid])) {
+            case (Left(e), _) => Left(e)
+            case (Right(ids), s) => decodeNid.fromString(s).map(ids :+ _)
+          }
+          .map(_.toImmArray)
+        for {
+          ni <- nodeId
+          children <- childrenOrError
+        } yield ni -> NodeRollback(children, nodeVersion)
       case NodeTypeCase.CREATE =>
         val protoCreate = protoNode.getCreate
         for {
@@ -754,6 +768,9 @@ object TransactionCoder {
       protoNode: TransactionOuterClass.Node,
   ): Either[DecodeError, NodeInfo] =
     protoNode.getNodeTypeCase match {
+      case NodeTypeCase.ROLLBACK =>
+        // TODO https://github.com/digital-asset/daml/issues/8020
+        sys.error("protoNodeInfo, rollback nodes are not supported")
       case NodeTypeCase.CREATE =>
         val protoCreate = protoNode.getCreate
         for {
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
@@ -246,6 +246,8 @@ object TransactionCoder {
       enclosingVersion: TransactionVersion,
       nodeId: Nid,
       node: GenNode[Nid, Cid],
+      disableVersionCheck: Boolean =
+        false, //true allows encoding of bad protos (for testing of decode checks)
   ): Either[EncodeError, TransactionOuterClass.Node] = {
     val nodeVersion = node.version
     if (enclosingVersion < nodeVersion)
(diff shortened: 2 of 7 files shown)


daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
```diff
@@ -45,6 +45,7 @@ message Node {
         NodeFetch fetch = 4;
         NodeExercise exercise = 5;
         NodeLookupByKey lookup_by_key = 6;
+        NodeRollback rollback = 7;
```
cocreature, line 48: please add the version in a comment like we do for other stuff in this file.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -260,9 +260,11 @@ object TransactionCoder {
         nodeBuilder.setVersion(node.version.protoValue)
 
       node match {
-        case _: NodeRollback[_] =>
-          // TODO https://github.com/digital-asset/daml/issues/8020
-          sys.error("rollback nodes are not supported")
+        case ne @ NodeRollback(_, _) =>
```
cocreature, line 263: I think this should have a version check similar to the one we have before writing choice observers.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -449,6 +451,18 @@ object TransactionCoder {
     val nodeId = decodeNid.fromString(protoNode.getNodeId)
 
     protoNode.getNodeTypeCase match {
+      case NodeTypeCase.ROLLBACK =>
+        val protoRollback = protoNode.getRollback
+        val childrenOrError = protoRollback.getChildrenList.asScala
+          .foldLeft[Either[DecodeError, BackStack[Nid]]](Right(BackStack.empty[Nid])) {
+            case (Left(e), _) => Left(e)
+            case (Right(ids), s) => decodeNid.fromString(s).map(ids :+ _)
+          }
```
cocreature, line 460: Maybe worth factoring out? Seems very similar to the one to decode the children of exercise nodes.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -449,6 +451,18 @@ object TransactionCoder {
     val nodeId = decodeNid.fromString(protoNode.getNodeId)
 
     protoNode.getNodeTypeCase match {
+      case NodeTypeCase.ROLLBACK =>
```
cocreature, line 454: We also need a version check here.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -754,6 +768,9 @@ object TransactionCoder {
       protoNode: TransactionOuterClass.Node,
   ): Either[DecodeError, NodeInfo] =
     protoNode.getNodeTypeCase match {
+      case NodeTypeCase.ROLLBACK =>
```
cocreature, line 771: :+1:

cocreature: note that this isn’t used at all in the Daml repo but it is used in pkvutils.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -133,10 +133,31 @@ class TransactionCoderSpec
       }
     }
 
+    "do NodeRollback" in {
```
cocreature, line 136: Once we add the version checks we should also add some tests for the unhappy path. choice observers are probably another good example for that type of test.

cocreature (commented):
Nice thank you! Let’s please add the version checks in this PR, then this should be good to :shipit:

daml-lf/transaction/src/main/protobuf/com/daml/lf/transaction.proto
nickchapman-da: done

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
nickchapman-da: done

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
nickchapman-da: ok, done

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
nickchapman-da: done

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -269,6 +290,23 @@ class TransactionCoderSpec
       }
     }
 
+    "fail if try encode rollback node in version < minExceptions" in {
```
cocreature, line 293: Please also add a test for decoding

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
nickchapman-da: added a test for the unhappy encode path

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
nickchapman-da: So what are we trying to test here? .. _that we cannot decode a rollback node as a transaction with version < `minExceptions`_

But I'm not sure how this can be tested:

- We must create a proto with a rollback node
- and this must be created with a tx-version `>= minExceptions`, or else the encode will fail.

- Then we must attempt to decode that proto with a target version of `< minExceptions`.
- But this will fail with the general check in `decodeNodeVersion`, that: _A transaction of version V-OLD cannot contain node of newer version (v-NEWER)_

- So we never reach the code which fails if it find a rollback node in an old version.


By this logic, the version check for `ROLLBACK` in `decodeNode` is redundant.

However, I remember in previous similar situations, Remy is keen we keep such checks even if that cannot be hit.

cocreature: If we assume that our code is correct, we don’t need any of those checks since we will never produce those nodes for the wrong version. The point of those checks is to be extra safeguards so encoding doesn’t make assumption about the producer that passes in the un-encoded version and decoding doesn’t make assumptions about the encoded version.

For your example, can we encode it, then tweak the version in the encoded protobuf and then decode it and see that it fails?

nickchapman-da: Makes sense.
I coded this by disabling the check at encode time, thus allowing the bad proto to be created.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -595,6 +633,31 @@ class TransactionCoderSpec
...
+        val normalizedNode = normalizeNode(node.updateVersion(v1))
+        val Right(encodedNode) =
+          TransactionCoder
+            .encodeNode(
+              TransactionCoder.NidEncoder,
+              ValueCoder.CidEncoder,
+              v1,
+              NodeId(0),
+              normalizedNode,
+              disableVersionCheck = true, //so the bad proto can be created
```
remyhaemmerle-da, line 647:
```suggestion
              normalizedNode.updateVersion(v1 min TransactionVersion.minException),
```

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -595,6 +633,31 @@ class TransactionCoderSpec
...
+            )
+        val result =
+          TransactionCoder
+            .decodeVersionedNode(
+              TransactionCoder.NidDecoder,
+              ValueCoder.CidDecoder,
+              v2,
+              encodedNode,
```
remyhaemmerle-da, line 655:
```suggestion
              encodedNode.toBuilder.setVersion(v1.protoValue).build,
```

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionCoderSpec.scala
```diff
@@ -595,6 +633,31 @@ class TransactionCoderSpec
       }
     }
 
+    "fail if we try to decode a rollback node in a version < minExceptions" in {
+      forAll(danglingRefRollbackNodeGen, versionInIncreasingOrder()) { case (node, (v1, v2)) =>
```
remyhaemmerle-da, line 637: We can avoid using `disableVersionCheck = true` as suggested bellow.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -245,6 +246,8 @@ object TransactionCoder {
       enclosingVersion: TransactionVersion,
       nodeId: Nid,
       node: GenNode[Nid, Cid],
+      disableVersionCheck: Boolean =
+        false, //true allows encoding of bad protos (for testing of decode checks)
```
remyhaemmerle-da, line 250: I am a bit uncomfortable with adding this kind of logic only for testing. 
This can be void by tweaking the proto directly as suggested below.

remyhaemmerle-da (commented):
except the `disableVersionCheck` parameter withing the `encodeNode` function, this looks good to.

PR #9469, opened 2021-04-21
Optionally emit setTime actions in daml export

### LINKED ISSUE #9081: Optionally set time for each transaction in daml script dump
Choices can have logic that depends on the current ledger time so in general, we have to set the time to reproduce issues This isn’t too common and we cannot fake time when running against a real ledger so we shouldn’t always do this. However, we should at least have the option to emit `setTime` calls.

aherrmann-da:
Closes #9081 

- Adds an option `--set-time` to `daml ledger export script`.
- If enabled it will emit `setTime` actions before submit actions if the transactions effective-at time differed from that of the previous one.
- Adds unit tests
- Enables `--set-time` in the integration test to make sure the emitted code compiles and runs. Consequently changes the ledger time mode to static time.
- Fixes the encoding of date and time values. These require `DA.Time`/`DA.Date` to be imported and to be used qualified.
- Updates the docs for `daml ledger export`.

### Pull Request Checklist

trigger the build.


docs/source/tools/export.rst
```diff
@@ -158,8 +158,6 @@ Transaction Time
 ================
 
 Daml ledger exports may fail to reproduce the ledger state or transaction
-history if contracts are sensitive to ledger time. The current implementation
-does not attempt to reproduce the history faithfully with regard to ledger
-time. In future this will be implemented by optionally issuing ``setTime``
-commands in the generated Daml script. However, this is not supported by all
-ledgers.
+history if contracts are sensitive to ledger time. You can enable the
```
cocreature, line 161: Maybe now that you can fix it, we should move it out of caveats?

cocreature (approved):
Nice, thank you!

daml-script/export/src/main/scala/com/daml/script/export/Encode.scala
```diff
@@ -24,12 +24,20 @@ private[export] object Encode {
...
+      setTime: Boolean,
   ): Doc = {
     val parties = partiesInContracts(acs.values) ++ trees.foldMap(partiesInTree(_))
     val partyMap = partyMapping(parties)
 
     val acsCidRefs = acs.values.foldMap(createdReferencedCids)
-    val submits = trees.map(Submit.fromTree)
+    val actions: Seq[Action] = {
+      if (setTime) {
+        Action.fromTrees(trees)
```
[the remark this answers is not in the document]
aherrmann-da: Yes, I was contemplating this. I've moved the `setTime` parameter into `Action.fromTrees`.

docs/source/tools/export.rst
aherrmann-da: Good point, I've moved it into its own section.

PR #9458, opened 2021-04-20
Check visibility for transient key lookups

cocreature:
fixes #9454

changelog_begin
changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9470, opened 2021-04-21
Check visibility for by-key operation of local contracts

### LINKED ISSUE #9454: Missing visibility check for key lookups of transient contracts
The following example currently succeeds but should fail because the key created in `LOL` is not visible to the submitter. It should succeed for a `submitMulti [p1] [p2]`. The check either needs to go into speedy or speedy needs to throw a callback to the ledger for lookups of transient contracts in addition to the lookups for non-transient contracts we already have.

```
module LolNoStakeholder where

import Daml.Script

template T
  with
    p : Party
  where
    signatory p
    key p : Party
    maintainer key

template MyHelper
  with
    p1 : Party
    p2 : Party
  where
    signatory p1, p2
    nonconsuming choice LOL : ()
      controller p2
      do create (T p1)
         Some _ <- lookupByKey @T p1
         pure ()
    nonconsuming choice Meh : ()
      controller p2
      do Some _ <- lookupByKey @T p1
         pure ()

submitterIsNoStakeholder = script do
  p1 <- allocateParty "p1"
  p2 <- allocateParty "p2"
  cid <- submitMulti [p1, p2] [] $ createCmd (MyHelper p1 p2)
  submit p2 $ exerciseCmd cid LOL
  submitMustFail p2 $ exerciseCmd cid Meh
```

cocreature:
fixes #9454

I tried out two approaches for this:

1. The one here where we add a new callback. This has the advantage
   that the engine remains oblivious to visibility checks. They are all
   done outside and the engine doesn’t even know about the reading
   parties.
2. Make the engine aware of the reading parties. A start of that is in
   #9458.

Both work in principle but I ended up going for 1 in the end. Doing
half of the visibility checks outside the engine and half inside just
seems worse than the current state.

changelog_begin

- [Daml Engine] Fix a bug where it was possible to
  fetch/lookup/exercise a local contract by key even if the reading parties
  are not stakeholders. See #9454 for details.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
+++ b/compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
@@ -1,6 +1,8 @@
--- @ERROR Attempt to fetch, lookup or exercise a key associated with a contract
--- @ERROR Attempt to fetch, lookup or exercise a key associated with a contract
--- @ERROR Attempt to fetch, lookup or exercise a key associated with a contract
+-- @ERROR range=18:1-18:10; Attempt to fetch, lookup or exercise a key associated with a contract
+-- @ERROR range=63:1-63:15; Attempt to fetch, lookup or exercise a key associated with a contract
+-- @ERROR range=81:1-81:12; Attempt to fetch, lookup or exercise a key associated with a contract
+-- @ERROR range=111:1-111:12; Attempt to fetch, lookup or exercise a key associated with a contract
+-- @ERROR range=117:1-117:11;  Attempt to fetch, lookup or exercise a key associated with a contract
 module ContractKeyNotVisible where
 
 import DA.Optional
@@ -70,8 +72,9 @@ divulgeeLookup = scenario do
   -- which means the error message changes. The reason is that the
   -- assertion is checked at interpretation time, before the lookup
   -- is checked at validation time.
+  delegationCid <- submit sig $ create (Delegation sig divulgee)
   submit divulgee do
-    mcid <- createAndExercise (Delegation sig divulgee) LookupKeyed
+    mcid <- exercise delegationCid LookupKeyed
     assert (isNone mcid)
     pure ()
 
@@ -80,7 +83,46 @@ blindLookup = scenario do
   blind <- getParty "b" -- Blind
   _ <- submit sig do create Keyed with ..
   -- Blind party can't do positive lookup with maintainer authority.
+  cid <- submit sig $ create (Delegation sig blind)
   submit blind do
-    mcid <- createAndExercise (Delegation sig blind) LookupKeyed
+    mcid <- exercise cid LookupKeyed
     assert (isNone mcid)
     pure ()
+
+template LocalKeyVisibility
+  with
+    sig : Party
+    obs : Party
+  where
+    signatory sig
+    observer obs
+    choice LocalLookup : ()
+      controller obs
+      do create (Keyed sig)
+         Some _ <- lookupByKey @Keyed sig
+         pure ()
+
+    choice LocalFetch : ()
+      controller obs
+      do create (Keyed sig)
+         _ <- fetchByKey @Keyed sig
+         pure ()
+
+localLookup = scenario do
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
+  submit p2 $ exercise cid LocalLookup
+
+localFetch = scenario do
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
+  submit p2 $ exercise cid LocalLookup
+
+localLookupFetchMustFail = scenario do
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
+  submitMustFail p2 $ exercise cid LocalLookup
+  submitMustFail p2 $ exercise cid LocalFetch
--- a/compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
+++ b/compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
@@ -118,7 +118,7 @@ localFetch = scenario do
   p1 <- getParty "p1"
   p2 <- getParty "p2"
   cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
-  submit p2 $ exercise cid LocalLookup
+  submit p2 $ exercise cid LocalFetch
 
 localLookupFetchMustFail = scenario do
   p1 <- getParty "p1"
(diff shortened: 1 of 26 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1182,6 +1182,25 @@ private[lf] object SBuiltin {
       val gkey = GlobalKey(templateId, keyWithMaintainers.key)
       // check if we find it locally
       onLedger.ptx.keys.get(gkey) match {
+        case Some(Some(coid)) if onLedger.localContracts.contains(coid) =>
+          val cachedContract = onLedger.cachedContracts
+            .get(coid)
+            .getOrElse(crash(s"Local contract $coid not in cachedContracts"))
+          val stakeholders = cachedContract.signatories union cachedContract.observers
+          throw SpeedyHungry(
+            SResultNeedLocalKeyVisible(
```
cocreature, line 1191: Note that this is not cached, I’ll implement caching on top of #9451 which provides the necessary infrastructure.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -119,7 +119,8 @@ conformance_test(
         "--include=ContractKeysIT" +
         ",RaceConditionIT",
         # Some of gRPC status codes changed from INVALID_ARGUMENT to ABORTED (see https://github.com/digital-asset/daml/pull/9218):
-        "--exclude=ContractKeysIT:CKFetchOrLookup,ContractKeysIT:CKNoFetchUndisclosed,ContractKeysIT:CKMaintainerScoped",
+        # LocalKeyVisibility can be reenabled on the next canton update.
```
cocreature, line 122: @oliverse-da fyi

oliverse-da: thanks @cocreature

garyverhaegen-da (approved):

garyverhaegen-da (commented):
Oops, GitHub was only showing me the change to `compatibility/bazel_tools/testing.bzl` when I approved. I must have clicked the wrong link in my email. I don't feel qualified to approve the rest of this PR.

cocreature:
> Oops, GitHub was only showing me the change to `compatibility/bazel_tools/testing.bzl` when I approved. I must have clicked the wrong link in my email. I don't feel qualified to approve the rest of this PR.

And I was hoping I finally found someone who wants to review one of my PRs :grin:

compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
```diff
@@ -80,7 +83,46 @@ blindLookup = scenario do
...
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
+  submit p2 $ exercise cid LocalLookup
+
+localFetch = scenario do
+  p1 <- getParty "p1"
+  p2 <- getParty "p2"
+  cid <- submit p1 $ create (LocalKeyVisibility p1 p2)
+  submit p2 $ exercise cid LocalLookup
```
tudor-da, line 121: Here, you should exercise `LocalFetch`, right?

compiler/damlc/tests/src/DA/Test/ScriptService.hs
```diff
@@ -700,6 +700,69 @@ main =
...
+                    , "localLookup = do"
+                    , "  p1 <- allocateParty \"p1\""
+                    , "  p2 <- allocateParty \"p2\""
+                    , "  cid <- submit p1 $ createCmd (LocalKeyVisibility p1 p2)"
+                    , "  submit p2 $ exerciseCmd cid LocalLookup"
+                    , "localFetch = do"
+                    , "  p1 <- allocateParty \"p1\""
+                    , "  p2 <- allocateParty \"p2\""
+                    , "  cid <- submit p1 $ createCmd (LocalKeyVisibility p1 p2)"
+                    , "  submit p2 $ exerciseCmd cid LocalLookup"
```
tudor-da, line 745: Same comment as above

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Result.scala
```diff
@@ -90,6 +98,45 @@ final case class ResultNeedKey[A](
     resume: Option[ContractId] => Result[A],
 ) extends Result[A]
 
+/** Whether a given contract can be fetched by key, i.e., actAs union readAs
+  *    contains at least one stakeholder.
+  */
+sealed trait VisibleByKey {
```
tudor-da, line 104: Isn't the compiler complaining that this is not `Serializable with Product`?

tudor-da: Ah sorry, only when inferring it from its children it is

compiler/damlc/tests/daml-test-files/ContractKeyNotVisible.daml
cocreature: good catch 57ee4f03436d8a323ce42868c6fdc3d9c079a00b

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Result.scala
cocreature: Happy to add it if you prefer, I don’t have strong feelings here either way.

tudor-da: I think it's bloat usually, so if the compiler doesn't want it, it's fine

tudor-da (approved):
Left some minor comments. Looks good otherwise 👍

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/StoreBackedCommandExecutor.scala
```diff
@@ -142,6 +144,10 @@ private[apiserver] final class StoreBackedCommandExecutor(
               )
             }
 
+        case ResultNeedLocalKeyVisible(stakeholders, resume) =>
+          val visible = VisibleByKey.fromSubmitters(actAs, readAs)(stakeholders)
```
gerolf-da, line 148: nit/idea: If you want to make use of the fact that there is an intermediate function, you could extract `lazy val isVisible = VisibleByKey.fromSubmitters(actAs, readAs)` outside `resolveStep` and then call `isvisible(stakeholders)` to avoid the repeated execution and construction of `actAs union readAs`. But this would be entering micro-optimization territory, so take this comment with a mountain of salt.

gerolf-da (approved):
Thanks, @cocreature. Looks good.

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/StoreBackedCommandExecutor.scala
cocreature: 2c9e2aa5036c7c7e1e1607be33f9ce9a33af26dc
easy enough, thanks!

PR #9472, opened 2021-04-22
Test multi-keys behavior

cocreature:
Canton relies on the Engine working correctly even in a setting where
we do not have multiple keys. So far this worked by accident but the
semantics of this are rather unclear. To make things worse, Canton
upgrades rely on those semantics being stable so we really do care
about the choices we make here. (however, canton’s non-unique mode is not GA yet so this is not all that important).

This PR adds a bunch of tests as an executable documentation of the
current behavior. However, we do not yet provide stability guarantees for
the current behavior and therefore these tests can be changed as
needed. But at least we are aware of those changes rather than doing
them by accident.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -2177,6 +2177,112 @@ class EngineTest
       }
     }
 
+    // Note that we provide no stability for multi key semantics so
+    // these tests serve only as an indication of the curren tbehavior
+    // but can be changed freely.
+    "multi keys" should {
+      val (multiKeysPkgId, _, allMultiKeysPkgs) = loadPackage("daml-lf/tests/MultiKeys.dar")
+      val lookupPackage = allMultiKeysPkgs.get(_)
+      val keyedId = Identifier(multiKeysPkgId, "MultiKeys:Keyed")
+      val opsId = Identifier(multiKeysPkgId, "MultiKeys:KeyOperations")
+      val let = Time.Timestamp.now()
+      val submissionSeed = hash("multikeys")
+      val seeding = Engine.initialSeeding(submissionSeed, participant, let)
+
+      val cid1 = toContractId("#1")
+      val cid2 = toContractId("#2")
+      val keyedInst = ContractInst(
+        TypeConName(multiKeysPkgId, "MultiKeys:Keyed"),
+        assertAsVersionedValue(ValueRecord(None, ImmArray((None, ValueParty(party))))),
+        "",
+      )
+      val contracts = Map(cid1 -> keyedInst, cid2 -> keyedInst)
+      val lookupContract = contracts.get(_)
+      def lookupKey(key: GlobalKeyWithMaintainers): Option[ContractId] =
+        (key.globalKey.templateId, key.globalKey.key) match {
+          case (
+                `keyedId`,
+                ValueParty(`party`),
+              ) =>
+            Some(cid1)
+          case _ =>
+            None
+        }
+      def run(choice: String, argument: Value[Value.ContractId]) = {
+        val cmd = CreateAndExerciseCommand(
+          opsId,
+          ValueRecord(None, ImmArray((None, ValueParty(party)))),
+          choice,
+          argument,
+        )
+        val Right((cmds, globalCids)) = preprocessor
+          .preprocessCommands(ImmArray(cmd))
+          .consume(lookupContract, lookupPackage, lookupKey, allKeysVisible)
+        engine
+          .interpretCommands(
+            validating = false,
+            submitters = Set(party),
+            commands = cmds,
+            ledgerTime = let,
+            submissionTime = let,
+            seeding = seeding,
+            globalCids = globalCids,
+          )
+          .consume(lookupContract, lookupPackage, lookupKey, allKeysVisible)
+      }
+      val emptyRecord = ValueRecord(None, ImmArray.empty)
+      // The cid returned by a fetchByKey at the beginning
+      val keyResultCid = ValueRecord(None, ImmArray((None, ValueContractId(cid1))))
+      // The cid not returned by a fetchByKey at the beginning
+      val nonKeyResultCid = ValueRecord(None, ImmArray((None, ValueContractId(cid2))))
+      val twoCids =
+        ValueRecord(None, ImmArray((None, ValueContractId(cid1)), (None, ValueContractId(cid2))))
+      val createOverwritesLocal = ("CreateOverwritesLocal", emptyRecord)
+      val createOverwritesUnknownGlobal = ("CreateOverwritesUnknownGlobal", emptyRecord)
+      val createOverwritesKnownGlobal = ("CreateOverwritesKnownGlobal", emptyRecord)
+      val fetchDoesNotOverwriteGlobal = ("FetchDoesNotOverwriteGlobal", nonKeyResultCid)
+      val fetchDoesNotOverwriteLocal = ("FetchDoesNotOverwriteLocal", keyResultCid)
+      val localArchiveOverwritesUnknownGlobal = ("LocalArchiveOverwritesUnknownGlobal", emptyRecord)
+      val localArchiveOverwritesKnownGlobal = ("LocalArchiveOverwritesKnownGlobal", emptyRecord)
+      val globalArchiveOverwritesUnknownGlobal = ("GlobalArchiveOverwritesUnknownGlobal", twoCids)
+      val globalArchiveOverwritesKnownGlobal1 = ("GlobalArchiveOverwritesKnownGlobal1", twoCids)
+      val globalArchiveOverwritesKnownGlobal2 = ("GlobalArchiveOverwritesKnownGlobal2", twoCids)
+      val rollbackCreateNonRollbackFetchByKey = ("RollbackCreateNonRollbackFetchByKey", emptyRecord)
+      val rollbackFetchByKeyNonRollbackCreate = ("RollbackFetchByKeyNonRollbackCreate", emptyRecord)
+      val rollbackFetchNonRollbackCreate = ("RollbackFetchNonRollbackCreate", keyResultCid)
+      val rollbackGlobalArchiveNonRollbackCreate =
+        ("RollbackGlobalArchiveNonRollbackCreate", keyResultCid)
+      val rollbackCreateNonRollbackGlobalArchive =
+        ("RollbackCreateNonRollbackGlobalArchive", keyResultCid)
+      val rollbackGlobalArchiveUpdates =
+        ("RollbackGlobalArchiveUpdates", twoCids)
+
+      "non-uck mode" in {
+        val allCases = Table(
+          ("choice", "argument"),
+          createOverwritesLocal,
+          createOverwritesUnknownGlobal,
+          createOverwritesKnownGlobal,
+          fetchDoesNotOverwriteGlobal,
+          fetchDoesNotOverwriteLocal,
+          localArchiveOverwritesUnknownGlobal,
+          localArchiveOverwritesKnownGlobal,
+          globalArchiveOverwritesUnknownGlobal,
+          globalArchiveOverwritesKnownGlobal1,
+          globalArchiveOverwritesKnownGlobal2,
+          rollbackCreateNonRollbackFetchByKey,
+          rollbackFetchByKeyNonRollbackCreate,
+          rollbackFetchNonRollbackCreate,
+          rollbackGlobalArchiveNonRollbackCreate,
+          rollbackCreateNonRollbackGlobalArchive,
+          rollbackGlobalArchiveUpdates,
+        )
+        forEvery(allCases) { case (name, arg) =>
+          run(name, arg) shouldBe a[Right[_, _]]
+        }
+      }
+    }
+
     "exceptions" should {
       val (exceptionsPkgId, _, allExceptionsPkgs) = loadPackage("daml-lf/tests/Exceptions.dar")
       val lookupPackage = allExceptionsPkgs.get(_)
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -2178,7 +2178,7 @@ class EngineTest
     }
 
     // Note that we provide no stability for multi key semantics so
-    // these tests serve only as an indication of the curren tbehavior
+    // these tests serve only as an indication of the current behavior
     // but can be changed freely.
     "multi keys" should {
       val (multiKeysPkgId, _, allMultiKeysPkgs) = loadPackage("daml-lf/tests/MultiKeys.dar")
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -180,16 +180,27 @@ private[lf] object PartialTransaction {
   *                 the transaction was in when aborted. It is up to
   *                 the caller to check for 'isAborted' after every
   *                 change to a transaction.
-  *  @param keys A local store of the contract keys. Note that this contains
-  *              info both about relative and contract ids. We must
-  *              do this because contract ids can be archived as
-  *              part of execution, and we must record these archivals locally.
-  *              Note: it is important for keys that we know to not be present
-  *              to be present as [[None]]. The reason for this is that we must
-  *              record the "no key" information for contract ids that
-  *              we archive. This is not an optimization and is required for
-  *              correct semantics, since otherwise lookups for keys for
-  *              locally archived contract ids will succeed wrongly.
+  *  @param keys A local store of the contract keys used for lookups and fetches by keys
+  *              (including exercise by key). Each of those operations will be resolved
+  *              against this map first. Only if there is no entry in here
+  *              (but not if there is an entry mapped to None), will we ask the ledger.
+  *
+  *              This map is mutated by the following operations:
+  *              1. fetch-by-key/lookup-by-key/exercise-by-key will insert an
+  *                 an entry in the map if there wasn’t already one (i.e., if they queried the ledger).
+  *              2. ACS mutating operations if the corresponding contract has a key. Specifically,
+  *                 2.1. A create will set the corresponding map entry to Some(cid) if the contract has a key.
+  *                 2.2. A consuming choice will set the corresponding map entry to None if the contract has a key.
+  *
+  *              On a rollback, we restore the state at the beginning of the rollback. Note that
+  *              This means that at the moment, we will query (by-key) for a global contract again even if
+  *              we queried it in a rollback node already.
+  *              TODO (MK) This should be fixed for performance and to ensure that the engine
+  *              never queries a key more than once thereby giving us consistency within a transaction.
+  *
+  *              Note that the engine is also used in Canton’s non-uck (unique contract key) mode.
+  *              In that mode, duplicate keys should not be an error. We provide no stability
+  *              guarantees for this mode at this point so tests can be changed freely.
   */
 private[lf] case class PartialTransaction(
     packageToTransactionVersion: Ref.PackageId => TxVersion,
--- a/daml-lf/tests/MultiKeys.daml
+++ b/daml-lf/tests/MultiKeys.daml
@@ -1,0 +1,235 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+module MultiKeys where
+
+import DA.Assert
+import DA.Exception (throw)
+import DA.Optional
+
+exception E
+  where
+    message "E"
+
+template Keyed
+  with
+    p : Party
+  where
+    signatory p
+    key p : Party
+    maintainer key
+
+-- All these tests operate under the assumption that there
+-- exists two global contracts with the same key `p` and one
+-- of those will always be returned by fetch/lookp-by-key at
+-- the beginning of the choice.
+-- All choices here are accepted in non-uck (unique contract key) mode.
+-- Behavior in uck-mode is documented in comments.
+-- Note that uck-mode here refers to the uck mode in the engine
+-- (not yet implemented, TODO(MK) remove this comment once it is) which only detects
+-- a subset of unique key errors.
+-- Note: At the moment, we do not provide any stability guarantees for
+-- non-uck at this point so these tests serve only as an indication of the
+-- current behaviro and can be changed freely.
+template KeyOperations
+  with
+    p : Party
+  where
+    signatory p
+
+    -- should be rejected in uck
+    nonconsuming choice CreateOverwritesLocal : ()
+      controller p
+      do cid1 <- create (Keyed p)
+         cid2 <- create (Keyed p)
+         (cid, _) <- fetchByKey @Keyed p
+         cid2 === cid
+
+    -- should be rejected in uck
+    nonconsuming choice CreateOverwritesUnknownGlobal : ()
+      controller p
+      do -- assume the global exists but do not fetch it
+         cid1 <- create (Keyed p)
+         (cid, _) <- fetchByKey @Keyed p
+         cid1 === cid
+
+    -- should be rejected in uck
+    nonconsuming choice CreateOverwritesKnownGlobal : ()
+      controller p
+      do _ <- fetchByKey @Keyed p
+         cid1 <- create (Keyed p)
+         (cid, _) <- fetchByKey @Keyed p
+         cid1 === cid
+
+    -- should be accepted in uck, cid not fetched by key
+    nonconsuming choice FetchDoesNotOverwriteGlobal : ()
+      with
+        cid : ContractId Keyed
+      controller p
+      do (cidByKey, _) <- fetchByKey @Keyed p
+         cid =/= cidByKey -- sanity-check the choice argument
+         c <- fetch cid
+         key c === p
+         (cidByKey', _) <- fetchByKey @Keyed p
+         cidByKey' === cidByKey
+
+    -- should be accepted in uck, cid not fetched by key
+    nonconsuming choice FetchDoesNotOverwriteLocal : ()
+      with
+        cid : ContractId Keyed
+      controller p
+      do local <- create (Keyed p)
+         c <- fetch cid
+         key c === p
+         (cid', _) <- fetchByKey @Keyed p
+         local === cid'
+
+    -- should be accepted in uck
+    nonconsuming choice LocalArchiveOverwritesUnknownGlobal : ()
+      controller p
+      do local <- create (Keyed p)
+         archive local
+         None <- lookupByKey @Keyed p
+         pure ()
+
+    -- should be rejected in uck
+    nonconsuming choice LocalArchiveOverwritesKnownGlobal : ()
+      controller p
+      do _ <- fetchByKey @Keyed p
+         local <- create (Keyed p)
+         archive local
+         None <- lookupByKey @Keyed p
+         pure ()
+
+    -- Should be accepted in uck
+    nonconsuming choice GlobalArchiveOverwritesUnknownGlobal : ()
+      with
+        cid1 : ContractId Keyed
+        cid2 : ContractId Keyed
+      controller p
+      do cid1 =/= cid2
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         archive cid1
+         None <- lookupByKey @Keyed p
+         pure ()
+
+    -- should be accepted in uck
+    nonconsuming choice GlobalArchiveOverwritesKnownGlobal1 : ()
+      with
+        cid1 : ContractId Keyed
+        cid2 : ContractId Keyed
+      controller p
+      do cid1 =/= cid2
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         (keyCid, _) <- fetchByKey @Keyed p
+         assert (keyCid `elem` [cid1, cid2])
+         archive (fromSome (find (/= keyCid) [cid1, cid2]))
+         None <- lookupByKey @Keyed p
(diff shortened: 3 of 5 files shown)


daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+      do (cid', _) <- fetchByKey @Keyed p
+         cid1 === cid'
+         try do
+           -- archive has an implicit read which is not rolled back
+           -- and should therefore update the following lookup.
+           archive cid2
+           throw E
+         catch
+           E -> pure ()
+         (cid', _) <- fetchByKey @Keyed p
```
cocreature, line 231: This is the only test that currently fails. As described in the comments, my intuition is that consuming choices with a key basically have an implicit `fetchByKey` before (ignoring the visibility check that we don’t have for the exercise). That is a read so it isn’t rolled back so therefore the test should succeed.

If that’s not the case, we need semantics that are quite different from the current implementation.

cocreature: To make that example even more annoying, consider that we don’t know if the contract we see via the archive passes the visibility check for contract keys! This could mean that the `fetchByKey` has to fail here. instead of returning `cid2` which seems very confusing.

cocreature: One way out of this mess could be to limit the effect of consuming choices to those cases:
1. Set `keys` to `None` if it was `Some(cid)` before and the choice was exercised on cid.
2. Set `keys` to `None` if there was no entry in `keys`.

Only seems to be getting more convoluted though.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+         (cid, _) <- fetchByKey @Keyed p
+         cid1 === cid
+
+    -- should be accepted in uck, cid not fetched by key
+    nonconsuming choice FetchDoesNotOverwriteGlobal : ()
+      with
+        cid : ContractId Keyed
+      controller p
+      do (cidByKey, _) <- fetchByKey @Keyed p
+         cid =/= cidByKey
```
andreaslochbihler-da, line 64: I suppose that this is supposed to encode a precondition of the choice?
```suggestion
         cid =/= cidByKey -- sanity-check the choice argument
```

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+    signatory p
+    key p : Party
+    maintainer key
+
+-- All these tests operate under the assumption that there
+-- exists two global contracts with the same key `p` and one
+-- of those will always be returned by fetch/lookp-by-key at
+-- the beginning of the choice.
+-- All choices here are accepted in non-uck mode.
+-- Behavior in uck-mode is documented in comments.
```
andreaslochbihler-da, line 27: What is "uck-mode" referring to here? Canton's uck mode will reject some of the choices as part of its contract key conflict detection. So is there a "Daml engine uck mode"?

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+         None <- lookupByKey @Keyed p
+         pure ()
+
+    -- should be rejected in uck
+    nonconsuming choice LocalArchiveOverwritesKnownGlobal : ()
+      controller p
+      do _ <- fetchByKey @Keyed p
+         local <- create (Keyed p)
+         archive local
+         None <- lookupByKey @Keyed p
```
andreaslochbihler-da, line 96: I'm on the fence w.r.t. the one. `lookupByKey` returning `None` is pretty meaningless in non-uck mode. Thinking about projections, it is probably the right choice, though:
The create-archive-lookup steps could be witnessed by some party `q` other than `p` and `q` doesn't witness the `fetchByKey`. So to `q`, this looks the same as `LocalArchiveOverwritesUnknownGlobal`.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         (keyCid, _) <- fetchByKey @Keyed p
+         assert (keyCid `elem` [cid1, cid2])
+         -- A consuming choice is
+         -- implicitly a by-key operation.
+         archive (fromSome (find (/= keyCid) [cid1, cid2]))
+         None <- lookupByKey @Keyed p
```
andreaslochbihler-da, line 130: I'd expect this to return `Some keyCid`, for the following reason:
1. The `fetchByKey` brings the association `Key p -> Some keyCid` into knowledge.
2. The `archive` acts on a different contract, so it should leave the key association unchanged.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+        cid1 : ContractId Keyed
+        cid2 : ContractId Keyed
+      controller p
+      do cid1 =/= cid2
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         archive cid1
+         None <- lookupByKey @Keyed p
```
andreaslochbihler-da, line 111: I wonder why whether `None` is required here. I'd argue the following:
1. Daml engine should ask the ledger `ResultNeedKey` at this point whether there is a contract for key `@Keyed p`.
2. If the ledger responds with `cid1`, then the answer should be `None` because `cid1` has been archived before in the transaction.
3. If the ledger responds with `cid2`, then the answer should be `Some cid2` because `cid2` has not been archived yet.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+      controller p
+      do cid1 =/= cid2
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         (keyCid, _) <- fetchByKey @Keyed p
+         assert (keyCid `elem` [cid1, cid2])
+         archive keyCid
+         None <- lookupByKey @Keyed p
```
andreaslochbihler-da, line 147: I agree that `None` is the only sensible choice here. `fetchByKey` brought the association `key -> Some keyCid` into scope, so it's `key -> None` after the archival.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+        cid : ContractId Keyed
+      controller p
+      do try do
+           create (Keyed p)
+           throw E
+         catch E -> pure ()
+         c <- fetch cid
+         key c === p
+         archive cid
+         None <- lookupByKey @Keyed p
```
andreaslochbihler-da, line 213: I disagree here. Like for `GlobalArchiveOverwritesUnknownGlobal`, Daml engine has not yet seen a read of the key from the underlying ledger, so the outcome depends on the response to the `ResultNeedKey`:
* If it's `Some cid`, then `None` is the `lookupByKey` result.
* If it's `Some otherCid`, then `Some otherCid` is the result.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -180,16 +180,33 @@ private[lf] object PartialTransaction {
...
+  *              In non-uck (Unique Contract Key) mode, i.e., Canton this is all we need to do.
+  *              In uck mode (not yet implemented), Speedy will detect and reject duplicate keys.
+  *              Only the operations that mutate `keys` will be taken into account for that.
+  *              Specifically, non-ACS operations that are not by-key (fetch, non-consuming exercises)
+  *              can never result in a failure due to duplicate keys.
+  *              For duplicate key detection, rollbacks roll back all writes but not reads.
+  *              So a global fetch by key in a rollback node will still conflict with
+  *              a create outside of the rollback node. And a create, comes with an implicit
+  *              NoSuchKey read so a create in a rollback node will conflict with a global fetch
+  *              by key outside of the rollback.
```
andreaslochbihler-da, line 209: Here's my suggestions of how to model keys. We have three kinds of operations on the `keys` map:
1. lookup of an entry in the `keys` map
2. caching a ledger response upon a `ResultNeedKey`
3. inserting/updating a key

A rollback node reverts all inserts/updates by its transitive children, but keeps their cache updates. This ensures that Daml engine queries the ledger for every key at most once during one interpretation run.

The operations are used as follows in non-uck mode:

1. A `Create` node of a contract with a key `k` upserts `keys` at `k` to the new contract ID. It does not care what was stored previously at `k`.
2. `Fetch` and non-consuming `Exercise` do not lookup or cache anything in `keys`.
3. `FetchByKey` and `ExerciseByKey` lookup the key `k` in keys `keys`. If they find an entry for `k`, that's the contract they use. If the don't find an entry, they ask the ledger with `ResultNeedKey` and cache the ledger's response in `keys`. If the `ExerciseByKey` is consuming, it then updates the key `k` to `None`.
4. A consuming `Exercise` on a contract `cid` with a key `k` checks whether `keys` associates `k` with `Some cid`. If so, it updates the association to `None`. Otherwise, `keys` remains as is.

daml-lf/tests/MultiKeys.daml
andreaslochbihler-da: Yes, that's my mental model that I settled upon.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+      do cid1 =/= cid2
+         c1 <- fetch cid1
+         key c1 === p
+         c2 <- fetch cid2
+         key c2 === p
+         (keyCid, _) <- fetchByKey @Keyed p
+         assert (keyCid `elem` [cid1, cid2])
+         -- A consuming choice is
+         -- implicitly a by-key operation.
+         archive (fromSome (find (/= keyCid) [cid1, cid2]))
```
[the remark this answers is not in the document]
meiersi-da: That is probably input for the semantics changes.

daml-lf/tests/MultiKeys.daml
meiersi-da: What about the alternative where we do not update the key-map on consumption, but instead check for non-consumedness on read from the key-map?

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+
+    -- should be rejected in uck
+    nonconsuming choice RollbackGlobalArchiveNonRollbackCreate : ()
+      with
+        cid : ContractId Keyed
+      controller p
+      do try do
+           c <- fetch cid
+           key c === p
+           -- Again, consuming choices are implicitly by-key.
```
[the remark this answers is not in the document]
meiersi-da: You examples seem to have a lot of structure, which potentially already covers that. What I'd find helpful in either case is  a short list at the top in the overall comments that explains how all the test cases are enumerated (they seem to be the cross-product of a few sets).

meiersi-da (commented):
@cocreature : thanks for the PR.

Notes wrt semantic changes that I'd see as beneficial:
- a transaction looks-up every global key at most once from the ledger, and thereby guarantees per-key consistency. This map is not reset on rollbacks, as it represents reads only.
- we design the intra-transaction lookup for multi-keys such that it has a predictable and compositional semantics

One option for this intra-transaction lookup is that we interpret a key lookup as the following SQL query:
```
SELECT * 
  FROM create_events 
 WHERE contract_key = ?key
    AND NOT (EXISTS SELECT 1 FROM archive_events WHERE archive_events.contract_id = create_events.contract_id)
 ORDER BY create_event.tx_node_id DESC
   LIMIT 1
```
What this implies is that new create events shadow old ones until they get archived. This semantics matches what we'd get if we applied the transaction nodes and their effects one-by-one (with separate begin_exercise and end_exercise events) to the ledger.

And it could in principle also be extended so that multiple lookups from the ledger would be supported in a deterministic fashion by supplying an exclusion list.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: Seems like a sensible model. To spell out the diff to the current state explicitly:
1. Cache across rollback nodes making sure the engine only ever looks up a key once. I’d argue the ledger is still broken if it returns two different results during execution but it seems like a very sensible expectation on the engine nevertheless.
2. Change archives to only set a key to None if it would have been returned by a directly preceding `fetchByKey`.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -180,16 +180,33 @@ private[lf] object PartialTransaction {
...
+  *              1. fetch-by-key/lookup-by-key/exercise-by-key will insert an
+  *                 an entry in the map if there wasn’t already one (i.e., if they queried the ledger).
+  *              2. ACS mutating operations if the corresponding contract has a key. Specifically,
+  *                 2.1. A create will set the corresponding map entry to Some(cid) if the contract has a key.
+  *                 2.2. A consuming choice will set the corresponding map entry to None if the contract has a key.
+  *
+  *              On a rollback, we restore the state at the beginning of the rollback. Note that
+  *              This means that at the moment, we will query (by-key) for a global contract again even if
+  *              we queried it in a rollback node already. We can cache this to improve
+  *              performance in the future.
```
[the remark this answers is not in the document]
cocreature: See comment above, I’m definitely in favor of changing this although I’d argue if the ledger is inconsistent here it’s still a bug.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+    p : Party
+  where
+    signatory p
+    key p : Party
+    maintainer key
+
+-- All these tests operate under the assumption that there
+-- exists two global contracts with the same key `p` and one
+-- of those will always be returned by fetch/lookp-by-key at
+-- the beginning of the choice.
```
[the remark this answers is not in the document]
cocreature: This is already done in this PR. I test at the engine level where I can absolutely throw different contracts with the same key in there.

daml-lf/tests/MultiKeys.daml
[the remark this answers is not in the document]
cocreature: Not in the engine, these are engine-level tests. My understanding was that we agreed that the engine should only throw on duplicate key errors if a contract is operated on by-key which in this case is not the case for the `fetch`.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+    -- should be rejected in uck
+    nonconsuming choice LocalArchiveOverwritesKnownGlobal : ()
+      controller p
+      do _ <- fetchByKey @Keyed p
+         local <- create (Keyed p)
+         archive local
+         None <- lookupByKey @Keyed p
+         pure ()
+
+    -- Should be accepted in uck
```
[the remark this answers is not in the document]
cocreature: again, engine-level tests. Only one of them is operated on by-key so my understanding was that the engine should not reject this.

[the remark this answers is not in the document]
meiersi-da: OK, I misunderstood the meaning of 'uck' then. As @andreaslochbihler-da pointed out, there seems to be a 'ledger-level uck' and `transaction-level uck` mode. What do you think about referring to the latter one here to avoid the confusion?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
meiersi-da: Thanks for the concrete model @andreaslochbihler-da. I agree with the diff that you @cocreature list.

@both: I'm no longer convinced, that resolving a key at most once is the right action in non-uck mode. The reason is that it implies that composing two working commands into a single one can lead to command interpretation failures.
```
-- setup
cid1 <- submit p $ create (Keyed p)
cid2 <- submit p $ create (Keyed p)
-- this works
submit p $ archiveByKey p
submit p $ archiveByKey p
-- using the composed version instead of the above two commands would fail
submit p $ do
  archiveByKey p
  archiveByKey p
```

Questions:
1. I believe our current system satisfies this composition property for uck mode, do you agree?
2. Do you believe we should keep this property?
3. If yes, how do we want to define non-uck key semantics so that we can efficiently implement it?

daml-lf/tests/MultiKeys.daml
meiersi-da: I'd say we should carefully revisit this once we've agreed on the semantics of non-uck key lookups. To me `lookupByKey` does not really make sense in non-uck mode, as its not clear at all what guarantee reading a `None` provides.

We might want to just forbid `lookupByKey` in non-uck mode.

daml-lf/tests/MultiKeys.daml
meiersi-da: I'm not sure for non-uck mode. The other option would be to query the ledger again for another contract. See my comment above: https://github.com/digital-asset/daml/pull/9472/files#r620200392

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: 1. agreed
2. seems like a very sensible and desirable property, phrasing it differently: The difference between batching commands into a single submission is transactionality and nothing else.
3. adding a set of exclusions to ResultNeedsKey seems easy on the engine side and hopefully reasonably efficient.

That said, overall I’m increasingly hesitant to tie multi-key semantics to doing duplicate key checks in the engine. The latter is a problem we have right now that results in actual bugs. I want to get rid of the 4 different implementations in the Daml repo sooner rather than later and I don’t think we should rush multi-key semantics. So I think we should have two related but orthogonal discussions:

1. What does an uck-mode for the engine reject? I think we agree on by-key operations and create. The open one seems to be `archive`. I think I still prefer @andreaslochbihler-da’s proposal here.
2. What should multi-key semantics be?

daml-lf/tests/MultiKeys.daml
andreaslochbihler-da: If there's a chance that `lookupByKey` returns a different contract here, then we need to change the Canton implementation of contract keys, because Canton currently supports at most one input contract for a contract key.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
andreaslochbihler-da: 1. With a UCK ledger (e.g., Canton in UCK mode), the second `submit p $ archiveByKey p` will fail. (IMO there's little point to run Daml engine in UCK mode on a non-uck ledger.) So why do you think that this works?
2. We violate the compositionality property already in other places. For example, in a multi-domain setup, transaction 1 could be routed to domain D1 and transaction 2 to domain D2, but there's no domain on which the combination of both transactions could be executed. (This problem will remain even with cross-domain transactions.)

The composition property is actually preserved with the current design under @oggy- 's fetch-Some semantics, which is what we guarantee to Daml developers. Under this semantics, the second `archiveByKey p` may fail even when we submit the transactions individually. And we do not promise to our users that aggregating several transactions into one has the same behaviour w.r.t. resolving the non-determinism around fetchSome.

cocreature: I think the example @meiersi-da provided was intended to be run in non-uck. With the semantics you proposed, the first one will succeed. The second one will not.

You are correct, that in terms of guarantees the first one could also fail but I think there is still a useful difference between "this could fail given the guarantees we provide" and "this will fail on an honest submitter".

meiersi-da: @cocreature : I agree to the separation of solving uck semantics of the engine and non-uck semantics.

As for uck-semantics, I'm in agreement with @andreaslochbihler-da 's proposal.

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,232 @@
...
+         _ <- fetchByKey @Keyed p
+         pure ()
+
+    -- should be rejected in uck
+    nonconsuming choice RollbackFetchByKeyNonRollbackCreate : ()
+      controller p
+      do try do
+           _ <- fetchByKey @Keyed p
+           throw E
+         catch E -> pure ()
```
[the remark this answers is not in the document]
meiersi-da: I'd also expect that
```
      do cid <- try do
           cid <- fetchByKey @Keyed p
           throw (E cid)
           pure cid
         catch (E cid) -> pure cid
         cid' <- fetchByKey @Keyed p
         assert $ cid == cid'
```

cocreature:
There is still an issue with @andreaslochbihler-da’s proposal even in the uck case:
Consider the following example:

1. We have a global contract with key k where Alice is not a stakeholder.
2. Alice submits a command.
3. That command first archives the global contract.
4. Then it does a lookup by key for `k`.

Currently, this will succeed and produce a negative lookup. However if as Andreas suggested, the second lookup will query the ledger, that will fail the visibility check and the whole submission fails. I’m not necessarily saying the new behavior is wrong but it is technically breaking.

In the multi-key case this gets more confusing. Consider the following:
1. We have a global contract with key `k` where Alice is a stakeholder.
2. Alice submits a command.
3. That command creates a contract with the same key `k` where Alice is not a stakeholder.
4. Then you do a lookup by key for `k`.

You cannot return the contract created in 3 since it fails the visibility check. However, it also seems awkward to return the contract in 1 and lose the recency property here. I’m somewhat tempted to say it should fail the submission but that also seems weird.

andreaslochbihler-da:
> There is still an issue with @andreaslochbihler-da’s proposal even in the uck case:
> Consider the following example:
> 
>     1. We have a global contract with key k where Alice is not a stakeholder.
>     2. Alice submits a command.
>     3. That command first archives the global contract.
>     4. Then it does a lookup by key for `k`.
> 

Great example. Let's distinguish between submission and reinterpretation. During submission, the visibility check ensures that we don't find the global contract and therefore put a `None` for the lookupByKey. In contrast, during reinterpretation, we see what should actually be going on: The engine should keep track of what contract the partial transaction has archived. If the ledger returns such an already-archived contract as part of a `ResultNeedKey`, then Daml engine interprets this as `None` for lookupByKey and a failure for exercise/fetchByKey. Does this match the current behaviour?

> In the multi-key case this gets more confusing. Consider the following:
> 
>     1. We have a global contract with key `k` where Alice is a stakeholder.
>     2. Alice submits a command.
>     3. That command creates a contract with the same key `k` where Alice is not a stakeholder.
>     4. Then you do a lookup by key for `k`.
> 
> You cannot return the contract created in 3 since it fails the visibility check. However, it also seems awkward to return the contract in 1 and lose the recency property here. I’m somewhat tempted to say it should fail the submission but that also seems weird.

I'd rather say that the visibility check is the problem here. This check is currently only in place because we wanted to discourage disclosing contracts via divulgence, because there are problems with divulgence. As soon as we have a proper treatment for disclosure (IIRC it's somewhere on the roadmap), we'll have to adjust the visibility check anyway. So I'd rather argue that we should change the visibility rule to say that a lookupByKey/exerciseByKey is allowed to find a contract if that contract is created in the same transaction even if it doesn't host a readAs party.

cocreature:
> Great example. Let's distinguish between submission and reinterpretation. During submission, the visibility check ensures that we don't find the global contract and therefore put a `None` for the lookupByKey. In contrast, during reinterpretation, we see what should actually be going on: The engine should keep track of what contract the partial transaction has archived. If the ledger returns such an already-archived contract as part of a `ResultNeedKey`, then Daml engine interprets this as `None` for lookupByKey and a failure for exercise/fetchByKey. Does this match the current behaviour?

The current behavior is that the lookup returns None and the submission succeeds even during validation.

I just realized my misunderstanding. I somehow thought that a `ResultNeedKey` would fail during submission for a key that is not visible and never return back to the engine. That means with your proposed change it would fail. However it just returns `None` and then as you pointed out we blow up during reinterpretation. Therefore, this seems to not change the current behavior :+1: 

I agree the visibility check needs changing so the second example is probably not too worrying.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -180,16 +180,33 @@ private[lf] object PartialTransaction {
...
-  *              part of execution, and we must record these archivals locally.
-  *              Note: it is important for keys that we know to not be present
-  *              to be present as [[None]]. The reason for this is that we must
-  *              record the "no key" information for contract ids that
-  *              we archive. This is not an optimization and is required for
-  *              correct semantics, since otherwise lookups for keys for
-  *              locally archived contract ids will succeed wrongly.
+  *  @param keys A local store of the contract keys used for lookups and fetches by keys
+  *              (including exercise by key). Each of those operations will be resolved
+  *              against against this map first. Only if there is no entry in here
```
oggy-, line 185:
```suggestion
  *              against this map first. Only if there is no entry in here
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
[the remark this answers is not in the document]
cocreature: Changed to a TODO since we agreed to fix this (but not in this PR).

daml-lf/tests/MultiKeys.daml
cocreature: clarified the comment.

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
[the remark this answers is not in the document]
cocreature: Clarified the comment above..

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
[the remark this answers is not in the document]
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, we’re postponing multikey semantics so resolving this comment for this PR.

daml-lf/tests/MultiKeys.daml
cocreature: As discussed, this is what we’re going to do but not in this PR (and ignoring effects for multi-keys).

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: Resolving since we agreed to postpone non-uck. Will implement the proposal for uck in separate PRs.

cocreature:
Following a fruitful discussion with @andreaslochbihler-da and @meiersi-da (thank you both!), we agreed to postpone stabilizing multi-key semantics for now.

I’ve changed this PR to document the current state as executable tests without implying stability guarantees (as pointed out very clearly in comments) and updated the PR description to reflect that.

In addition to that, we agreed that the engine should start doing unique key checks soonish taking only by-key & create with key operations into account. Archive of a contract which hasn’t been brought into scope by key by a prior by-key operation will result in another `ResultNeedKey` thrown to the ledger and the engine checks if the result has already been archived. All of that will be implemented in separate PRs so if there are open questions (we’re not aware of any), we can resolve them there.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -2177,6 +2177,112 @@ class EngineTest
       }
     }
 
+    // Note that we provide no stability for multi key semantics so
+    // these tests serve only as an indication of the curren tbehavior
```
sofiafaro-da, line 2181:
```suggestion
    // these tests serve only as an indication of the current behavior
```

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,235 @@
...
+-- of those will always be returned by fetch/lookp-by-key at
+-- the beginning of the choice.
+-- All choices here are accepted in non-uck (unique contract key) mode.
+-- Behavior in uck-mode is documented in comments.
+-- Note that uck-mode here refers to the uck mode in the engine
+-- (not yet implemented, TODO(MK) remove this comment once it is) which only detects
+-- a subset of unique key errors.
+-- Note: At the moment, we do not provide any stability guarantees for
+-- non-uck at this point so these tests serve only as an indication of the
+-- current behaviro and can be changed freely.
```
sofiafaro-da, line 33:
```suggestion
-- current behavior and can be changed freely.
```

daml-lf/tests/MultiKeys.daml
```diff
@@ -0,0 +1,235 @@
...
+  with
+    p : Party
+  where
+    signatory p
+    key p : Party
+    maintainer key
+
+-- All these tests operate under the assumption that there
+-- exists two global contracts with the same key `p` and one
+-- of those will always be returned by fetch/lookp-by-key at
```
sofiafaro-da, line 24:
```suggestion
-- of those will always be returned by fetch/lookup-by-key at
```

sofiafaro-da (approved):
Thank you!

PR #9473, opened 2021-04-22
Adding Daml Script to the "create-daml-app" template

nemanja-da:
CHANGELOG_BEGIN
Added Daml Script to the "create-daml-app" template. This template is used a lot at daml.com/learn and we're switching to using Daml Script over scenarios in general
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/templates/create-daml-app/daml.yaml.template
+++ b/templates/create-daml-app/daml.yaml.template
@@ -10,6 +10,7 @@ dependencies:
 - daml-prim
 - daml-stdlib
 - daml-trigger
+- daml-script
 sandbox-options:
 - --wall-clock-time
 - --ledgerid=__PROJECT_NAME__-sandbox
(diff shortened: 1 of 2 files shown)


cocreature (changes requested):
This is just an unused import and an unused dependency. If you want to add the import and the dependency, we should have a script in the file.

garyverhaegen-da:
Do we still want this? There is no scenario currently in create-daml-app, so nothing to port, but  could see an argument for providing an example Script in this template.

Adding the script itself would be easy enough, but I'm not sure that'd be enough. We may want to update the documentation to mention it, and update the build to actually run those scripts. It's not a huge amount of work but it's not entirely trivial either.

cocreature:
[#12089](https://github.com/digital-asset/daml/pull/12089) will add it anyway so I’ll close this.

PR #9474, opened 2021-04-22
KVL-874 Change the TelemetryContext to use SpanAttributes instead of raw AttributeKeys and remove an unused object

hubert-da:
- Removes part of unused code
- Reduces the number of dependencies
- Makes the API a bit more elegant (with additional restrictions)
- Renames the Tracer and its instrumentation name from `participant` to `com.daml.telemetry`

### Pull Request Checklist

trigger the build.


ledger/metrics/src/main/scala/com/daml/telemetry/package.scala
```diff
@@ -7,5 +7,5 @@ import io.opentelemetry.api.GlobalOpenTelemetry
 import io.opentelemetry.api.trace.Tracer
 
 package object telemetry {
-  val ParticipantTracer: Tracer = GlobalOpenTelemetry.getTracer("participant")
+  val GlobalTracer: Tracer = GlobalOpenTelemetry.getTracer("participant")
```
hubert-da, line 10: Are we ok to use this "GlobalTracer" across the sdk and "pkv" i.e. get rid of [this one](https://github.com/DACH-NY/daml-on-kv-oem-integration-kit/blob/0e365299528d63dfce8ea3def589ec4bef3f79b4/pkvutils/src/main/scala/com/daml/ledger/telemetry/package.scala#L10)?

ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,107 @@
...
+  /** Returns a telemetry context from the metadata encoded in the given key-value map.
+    *
+    * Typically, the metadata map has been encoded in a message, which has been sent
+    * across boundaries, from a different process.
+    * Originally, it has been created to carry tracing metadata across boundaries, and
+    * to create complete traces.
+    *
+    * @see [[com.daml.telemetry.TelemetryContext.encodeMetadata()]]
+    */
+  def contextFromMetadata(metadata: Option[jMap[String, String]]): TelemetryContext
```
hubert-da, line 33: There are a few methods like this one, `TelemetryContext.runInOpenTelemetryScope` and `TelemetryContext.encodeMetadata` that are not used in this repos, shall I move them back somehow to the oem-integration-kit repo?

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
```diff
@@ -82,35 +79,19 @@ trait TelemetryContext {
 
 }
 
-object TelemetryContext {
```
hubert-da, line 85: There are a few more methods that are not being used in this repo (yet?), which require more effort to get rid of. Presumably by introducing more inheritance in the oem integration kit repo, which I'm not happy about. I won't address it in this PR.

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
```diff
@@ -6,20 +6,17 @@ package com.daml.telemetry
...
 
 import scala.concurrent.Future
 import scala.util.{Failure, Success}
 
 trait TelemetryContext {
 
-  /** Sets or replaces the value of attribute `key` to `value`.
+  /** Sets or replaces the value of `attribute` to `value`.
     */
-  def setAttribute[T](key: AttributeKey[T], value: T): TelemetryContext
```
hubert-da, line 22: @miklos-da are you fine with supporting just the String values here? 

There are 2-3 places in the oem-integration-kit repo where other types have been used, which I easily changed (to `String`s). I'm not sure if, potentially, the only drawback is in the UI that will represent everything as `String`s. 

Adding a type parameter to the `SpanAttribute` (which I already checked) complicates things a bit. I could also leave this particular method as-is. However, it results in an additional dependency and exposing the raw types from the opentelementry lib.

miklos-da: Yes, it's OK to simplify it now. We can still go back to supporting typed span attributes if requested by users of `TelemetryContext`.

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
```diff
@@ -146,7 +127,7 @@ protected class DefaultTelemetryContext(protected val span: Span) extends Teleme
   protected def createSubSpan(
       spanName: String,
       kind: SpanKind,
-      attributes: (AttributeKey[String], String)*
+      attributes: (SpanAttribute, String)*
   ): Span = {
     val subSpan =
       ParticipantTracer
```
hubert-da, line 133: @miklos-da do you know what would be the consequences of using a global `tracer` across the sdk and the oem-integration-kit? I'm wondering if I really have to make it a parameter.

hubert-da: According to my e2e test, there were correct service names in the UI.

miklos-da: That's fine, based on the [OpenTelemetry documentation](https://github.com/open-telemetry/opentelemetry-java/blob/005b8c97efbc1d20ad1565922a1176d503f439af/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L105) this only means fixing "the name of the instrumentation library, not the name of the instrument*ed* library". We can consider `com.daml.telemetry` as the instrumentation library.

miklos-da (approved):

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
hubert-da: Thanks, I changed the instrumentation name from `participant` to `com.daml.telemetry` as discussed offline.

fabiotudone-da (approved):

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
lucb-da: No special concern from me to limit ourselves to `String`.
The generic support is only used for one attribute in the oem-integration-kit repo, and I don't think it is really needed.
There is already a comment in the code stating that putting the values into a `String`, instead of using `List[String]`, might be as good.
And the trace UI is mainly text search based anyway.

PR #9477, opened 2021-04-22
Fix index size limit issues [DPP-347]

kamil-da:
## Main changes
- Added `key_hash` column to the ledger `*state` table for `ledger-on-sql` - the new column will become a primary key
-  Modified index on the `participant_command_completions` table to not include submitters in the index
- Added `ValueLimitsIT` test suite to the `ledger-api-test-tool`

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ValueLimitsIT.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.concurrent.Future
+import scala.util.Random
+
+final class ValueLimitsIT extends LedgerTestSuite {
+
+  test(
+    "VLLargeSubmittersNumberCreateContract",
+    "Create a contract with a large submitters number",
+    allocate(NoParties),
+    runConcurrently = false,
```
kamil-da, line 20: Added this to avoid `Overloaded` errors on the `sandbox-on-x`. Couldn't figure out what was the reason.

ledger/ledger-on-sql/src/main/resources/com/daml/ledger/on/sql/migrations/sqlite/V4__change_state_primary_key.sql
```diff
@@ -0,0 +1,12 @@
...
+    key_hash VARBINARY(2048) PRIMARY KEY NOT NULL,
+    key   VARBINARY(16384) NOT NULL,
+    value BLOB                         NOT NULL
+);
+
+INSERT INTO ${table.prefix}state_new (key_hash, key, value)
+    SELECT key_hash, key, value FROM ${table.prefix}state;
+
+DROP TABLE ${table.prefix}state;
+ALTER TABLE ${table.prefix}state_new RENAME TO ${table.prefix}state;
```
kamil-da, line 12: AFAIK there's no other way to change the primary key constraint on the SQLite.

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/migrations/V3__Backfill_Key_Hash_State_Table.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import anorm.{BatchSql, NamedParameter}
+import com.daml.ledger.on.sql.queries.StateKeyHashing
+import org.flywaydb.core.api.migration.{BaseJavaMigration, Context}
+
+import java.sql.{Connection, ResultSet}
+import scala.collection.compat.immutable.LazyList
+import scala.jdk.CollectionConverters._
+
+private[migrations] abstract class V3__Backfill_Key_Hash_State_Table extends BaseJavaMigration {
```
kamil-da, line 14: I've tested this migration manually using Postgres. Do you think it's worth writing a test for this?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ValueLimitsIT.scala
kamil-da: After discussing with @nmarton-da we decided to increase the `ReadWriteServiceBridge` queue size. The previous value (200) was too low to handle conformance tests load.
It'd be valuable to implement a retry mechanism in `ReadWriteServiceBridge.submit()` in the `sandbox-on-x`.

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/migrations/V3__Backfill_Key_Hash_State_Table.scala
mziolekda: Ledger-on-sql is a library used by sandbox-next only. Its postgres-backed ledger-on-sql is a configuration that is not built into any client-facing product. Long term we want to get rid of it in favour of sandbox-on-x. So writing test for it would be throwing good money after bad.

mziolekda (commented):
Thank you Kamil for this PR.
Could you please check with @rautenrieth-da if he is ok with the approach taken to update the schema versions 45 and 50 of the index-db?

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/migrations/V3__Backfill_Key_Hash_State_Table.scala
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.on.sql.migrations
+
+import anorm.{BatchSql, NamedParameter}
+import com.daml.ledger.on.sql.queries.StateKeyHashing
```
rautenrieth-da, line 7: Consider inlining the code from this import. Migrations should be immutable, if we ever change `StateKeyHashing`, we will silently break this migration. We had many issues with the `V2_1__Rebuild_Acs` migration of the index database for this reason.

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/queries/PostgresqlQueries.scala
```diff
@@ -30,7 +30,7 @@ final class PostgresqlQueries(override protected implicit val connection: Connec
   }
 
   override protected val updateStateQuery: String =
-    s"INSERT INTO $StateTable VALUES ({key}, {value}) ON CONFLICT(key) DO UPDATE SET value = {value}"
+    s"INSERT INTO $StateTable (key, key_hash, value) VALUES ({key}, {key_hash}, {value}) ON CONFLICT(key_hash) DO UPDATE SET value = {value}"
```
rautenrieth-da, line 33: Technically this would corrupt the database on a hash collision (where the `key_hash` value conflicts with an existing row, but the `key` values are different).

Are we worried about that? If yes, you could do something like `ON CONFLICT(key_hash) DO UPDATE SET value = {value} WHERE EXCLUDED.key = {key}` and then count the number of affected rows (should be exactly 1).

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V45__fix_large_size_index_issues.sql
```diff
@@ -0,0 +1,5 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+DROP INDEX IF EXISTS participant_command_completio_completion_offset_application_idx;
```
rautenrieth-da, line 4: Why the `IF EXISTS` clause? The list of previous migrations should determine whether the index exists or not.

rautenrieth-da (approved):
Left some comments, changes to the schema look good to me.

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/migrations/V3__Backfill_Key_Hash_State_Table.scala
kamil-da: That's a good point. Thanks!

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/queries/PostgresqlQueries.scala
kamil-da: The `key_hash` is generated with the SHA256 algorithm, so I would assume that we're safe here but a green light from you @rautenrieth-da or @mziolekda would be great :)

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/migrations/V3__Backfill_Key_Hash_State_Table.scala
kamil-da: Fixed.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V45__fix_large_size_index_issues.sql
kamil-da: Fixed.

PR #9478, opened 2021-04-22
Support rollback nodes in kvutils ContractKeysValidation

cocreature:
Unfortunately, I couldn’t find any tests to update for this. If
someone can point me to something, I’ll happily add some rollback
tests.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
@@ -66,8 +66,10 @@ private[transaction] object ContractKeysValidation {
       contractKeysToContractIds: Map[DamlContractKey, RawContractId],
       transactionEntry: DamlTransactionEntrySummary,
   ): StepResult[DamlTransactionEntrySummary] = {
+    type KeyValidationStackStatus = Either[KeyValidationError, List[KeyValidationState]]
+
     val keysValidationOutcome = transactionEntry.transaction
-      .foldInExecutionOrder[Either[KeyValidationError, List[KeyValidationState]]](
+      .foldInExecutionOrder[KeyValidationStackStatus](
         Right(List(KeyValidationState(activeStateKeys = contractKeyDamlStateKeys)))
       )(
         exerciseBegin = (status, _, exerciseBeginNode) => {
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
@@ -146,8 +146,8 @@ private[transaction] object ContractKeysValidation {
   ) {
 
     /** Earlier active state keys are discarded.
-      *        submitted contract key mapping is merged biased towards earlier
-      *        assignments.
+      * submitted contract key mapping is merged biased towards earlier
+      * assignments.
       */
     def +(state: KeyValidationState): KeyValidationState = {
       val newContractKeyMappings =
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
@@ -112,14 +112,18 @@ private[transaction] object ContractKeysValidation {
       initialState: KeyValidationState,
   ): KeyValidationStatus =
     for {
-      stateAfterUniquenessCheck <- checkNodeKeyUniqueness(
-        node,
-        initialState,
+      stateAfterUniquenessCheck <- initialState.onActiveStateKeys(
+        checkNodeKeyUniqueness(
+          node,
+          _,
+        )
       )
-      finalState <- checkNodeKeyConsistency(
-        contractKeysToContractIds,
-        node,
-        stateAfterUniquenessCheck,
+      finalState <- stateAfterUniquenessCheck.onSubmittedContractKeysToContractIds(
+        checkNodeKeyConsistency(
+          contractKeysToContractIds,
+          node,
+          _,
+        )
       )
     } yield finalState
 
@@ -145,19 +149,29 @@ private[transaction] object ContractKeysValidation {
       ]],
   ) {
 
-    /** Earlier active state keys are discarded.
-      * submitted contract key mapping is merged biased towards earlier
-      * assignments.
-      */
-    def +(state: KeyValidationState): KeyValidationState = {
-      val newContractKeyMappings =
-        state.submittedContractKeysToContractIds -- submittedContractKeysToContractIds.keySet
-      new KeyValidationState(
-        activeStateKeys = state.activeStateKeys,
-        submittedContractKeysToContractIds =
-          submittedContractKeysToContractIds ++ newContractKeyMappings,
+    def onSubmittedContractKeysToContractIds(
+        f: Map[DamlContractKey, Option[
+          RawContractId
+        ]] => Either[KeyValidationError, Map[DamlContractKey, Option[
+          RawContractId
+        ]]]
+    ): KeyValidationStatus =
+      f(submittedContractKeysToContractIds).map(newSubmitted =>
+        new KeyValidationState(
+          activeStateKeys = this.activeStateKeys,
+          submittedContractKeysToContractIds = newSubmitted,
+        )
       )
-    }
+
+    def onActiveStateKeys(
+        f: Set[DamlStateKey] => Either[KeyValidationError, Set[DamlStateKey]]
+    ): KeyValidationStatus =
+      f(activeStateKeys).map(newActive =>
+        new KeyValidationState(
+          activeStateKeys = newActive,
+          submittedContractKeysToContractIds = this.submittedContractKeysToContractIds,
+        )
+      )
   }
 
   private[keys] object KeyValidationState {
@@ -174,11 +188,6 @@ private[transaction] object ContractKeysValidation {
 
   private[keys] type KeyValidationStatus =
     Either[KeyValidationError, KeyValidationState]
-
-  def currentStatus(
-      statusStack: Either[KeyValidationError, List[KeyValidationState]]
-  ): KeyValidationStatus =
-    statusStack.map(_.head)
 
   def onCurrentStatus[E, A](
       statusStack: Either[E, List[A]]
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
@@ -35,6 +35,7 @@ private[transaction] object ContractKeysValidation {
           case (k, v) if k.hasContractKey && v.getContractKeyState.getContractId.nonEmpty =>
             k -> v.getContractKeyState.getContractId
         }
+      // State before the transaction
       val contractKeyDamlStateKeys: Set[DamlStateKey] =
         contractKeyDamlStateKeysToContractIds.keySet
       val contractKeysToContractIds: Map[DamlContractKey, RawContractId] =
@@ -66,27 +67,23 @@ private[transaction] object ContractKeysValidation {
       transactionEntry: DamlTransactionEntrySummary,
   ): StepResult[DamlTransactionEntrySummary] = {
     val keysValidationOutcome = transactionEntry.transaction
-      .foldInExecutionOrder[KeyValidationStatus](
-        Right(
-          KeyValidationState(activeStateKeys = contractKeyDamlStateKeys)
-        )
+      .foldInExecutionOrder[Either[KeyValidationError, List[KeyValidationState]]](
+        Right(List(KeyValidationState(activeStateKeys = contractKeyDamlStateKeys)))
       )(
-        (keyValidationStatus, _, exerciseBeginNode) =>
-          (
-            checkNodeContractKey(
-              exerciseBeginNode,
-              contractKeysToContractIds,
-              keyValidationStatus,
-            ),
-            true,
-          ),
-        (_, _, _) =>
-          // TODO https://github.com/digital-asset/daml/issues/8020
-          sys.error("rollback nodes are not supported"),
-        (keyValidationStatus, _, leafNode) =>
-          checkNodeContractKey(leafNode, contractKeysToContractIds, keyValidationStatus),
-        (accum, _, _) => accum,
-        (accum, _, _) => accum,
+        exerciseBegin = (status, _, exerciseBeginNode) => {
+          val newStatus =
+            onCurrentStatus(status)(
+              checkNodeContractKey(exerciseBeginNode, contractKeysToContractIds, _)
+            )
+          (newStatus, true)
+        },
+        rollbackBegin = (status, _, _) =>
+          // Store state to restore after rollback
+          (status.map(stack => stack.head +: stack), true),
+        leaf = (status, _, leafNode) =>
+          onCurrentStatus(status)(checkNodeContractKey(leafNode, contractKeysToContractIds, _)),
+        exerciseEnd = (accum, _, _) => accum,
+        rollbackEnd = (status, _, _) => status.map(stack => stack.tail),
       )
 
     keysValidationOutcome match {
@@ -110,12 +107,11 @@ private[transaction] object ContractKeysValidation {
   }
 
   private def checkNodeContractKey(
-      node: Node.GenNode[NodeId, ContractId],
+      node: Node.GenActionNode[NodeId, ContractId],
       contractKeysToContractIds: Map[DamlContractKey, RawContractId],
-      keyValidationStatus: KeyValidationStatus,
+      initialState: KeyValidationState,
   ): KeyValidationStatus =
     for {
-      initialState <- keyValidationStatus
       stateAfterUniquenessCheck <- checkNodeKeyUniqueness(
         node,
         initialState,
@@ -133,12 +129,26 @@ private[transaction] object ContractKeysValidation {
   private[keys] case object Duplicate extends KeyValidationError
   private[keys] case object Inconsistent extends KeyValidationError
 
+  /** The state used during key validation.
+    *
+    * @param activeStateKeys The currently active contract keys for uniqueness
+    *  checks starting with the keys active before the transaction.
+    * @param submittedContractKeysToContractIds Map of keys to their expected assignment at
+    *  the beginning of the transaction starting with an empty map.
+    *  E.g., a create (with nothing before) means the key must have been inactive
+    *  at the beginning.
+    */
   private[keys] final class KeyValidationState private[ContractKeysValidation] (
       private[keys] val activeStateKeys: Set[DamlStateKey],
       private[keys] val submittedContractKeysToContractIds: Map[DamlContractKey, Option[
         RawContractId
       ]],
   ) {
+
+    /** Earlier active state keys are discarded.
+      *        submitted contract key mapping is merged biased towards earlier
+      *        assignments.
+      */
     def +(state: KeyValidationState): KeyValidationState = {
       val newContractKeyMappings =
         state.submittedContractKeysToContractIds -- submittedContractKeysToContractIds.keySet
@@ -164,4 +174,19 @@ private[transaction] object ContractKeysValidation {
 
   private[keys] type KeyValidationStatus =
     Either[KeyValidationError, KeyValidationState]
+
+  def currentStatus(
+      statusStack: Either[KeyValidationError, List[KeyValidationState]]
+  ): KeyValidationStatus =
+    statusStack.map(_.head)
+
+  def onCurrentStatus[E, A](
+      statusStack: Either[E, List[A]]
+  )(f: A => Either[E, A]): Either[E, List[A]] =
+    for {
+      statusStack <- statusStack
+      status <- f(statusStack.head)
+    } yield {
+      status +: statusStack.tail
+    }
 }
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
@@ -144,17 +144,11 @@ private[transaction] object ContractKeysValidation {
     */
   private[keys] final class KeyValidationState private[ContractKeysValidation] (
       private[keys] val activeStateKeys: Set[DamlStateKey],
-      private[keys] val submittedContractKeysToContractIds: Map[DamlContractKey, Option[
-        RawContractId
-      ]],
+      private[keys] val submittedContractKeysToContractIds: KeyConsistencyValidation.State,
   ) {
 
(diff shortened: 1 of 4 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -133,12 +129,26 @@ private[transaction] object ContractKeysValidation {
...
+    */
   private[keys] final class KeyValidationState private[ContractKeysValidation] (
       private[keys] val activeStateKeys: Set[DamlStateKey],
       private[keys] val submittedContractKeysToContractIds: Map[DamlContractKey, Option[
         RawContractId
       ]],
   ) {
+
+    /** Earlier active state keys are discarded.
+      *        submitted contract key mapping is merged biased towards earlier
```
miklos-da, line 149: The formatting of this comment is off.

miklos-da (commented):
Can you add some tests to `TransactionCommitterSpec`, please? See test case "validateContractKeys" in the previous suite, alternatively, feel free to create a new suite for `ContractKeysValidation`.

cocreature:
> Can you add some tests to `TransactionCommitterSpec`, please? See test case "validateContractKeys" in the previous suite, alternatively, feel free to create a new suite for `ContractKeysValidation`.

Perfect, not sure how I didn’t find that. Will add some tests :+1:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/KeyConsistencyValidation.scala
```diff
@@ -68,47 +67,42 @@ private[keys] object KeyConsistencyValidation {
...
+      case None => Right(submittedContractKeysToContractIds)
       case Some(submittedKeyWithMaintainers) =>
         val submittedDamlContractKey =
           Conversions.encodeContractKey(templateId, submittedKeyWithMaintainers.key)
-        val newKeyValidationState =
-          keyValidationState + ConsistencyKeyValidationState(
-            submittedContractKeysToContractIds = Map(
-              submittedDamlContractKey ->
-                targetContractId.map(_.coid)
+        val newSubmittedContractKeysToContractIds =
```
cocreature, line 77: @miklos-da this was just broken beforehand already. The `+` operator takes the second activeStateKeys set which means that here you always set it to an empty set. I added a testcase for this that fails on main (the two local key conflicts).

Rather than trying to fix this, I’ve just scraped the whole use of `+`. imho the new code is simpler and easier to get right but if you feel strongly about this, I can go back to the other solution. A fix would probably be for `+` to take the union instead. That doesn’t really make sense since a later state should always have a superset but given the way the code is setup here, that’s not the case.

cocreature:
@miklos-da added a bunch of tests and fixed a preexisting bug along the way.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
   }
 
   "validateContractKeys" should {
-    "return Inconsistent when a contract key resolves to a different contract ID than submitted by a participant" in {
-      def createNode(contractId: String): Create =
-        create(contractId, signatories = Seq("Alice"), keyAndMaintainer = Some(aKey -> "Alice"))
+    def createNode(contractId: String): Create =
```
fabiotudone-da, line 598: Since you're already here :) Perhaps this helps clarifying a bit:

```suggestion
    def newCreateNodeWithFixedKey(contractId: String): Create =
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+      if (inRollback) {
+        val rollback = builder.add(txBuilder.rollback())
+        builder.add(lookup, rollback)
+      } else {
+        builder.add(lookup)
+      }
+      builder.buildSubmitted()
+    }
+
+    val mykey = {
```
fabiotudone-da, line 619: If I understand correctly:

```suggestion
    val globalKeyConflictingWithLocalKey = {
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+          rejectionReason should startWith("InconsistentKeys")
       }
     }
 
+    "return DuplicateKeys when two local contracts conflict" in {
+      val builder = TransactionBuilder()
+      builder.add(createNode(s"#$freshContractId"))
+      builder.add(createNode(s"#$freshContractId"))
+      val transaction = builder.buildSubmitted()
+      val context = commitContextWithContractStateKeys(mykey -> None)
```
fabiotudone-da, line 657: Is this necessary?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+          )
+          val result = validate(context, transaction)
+          result shouldBe a[StepStop]
+          val rejectionReason =
+            getTransactionRejectionReason(result).getInconsistent.getDetails
+          rejectionReason should startWith("InconsistentKeys")
       }
     }
 
+    "return DuplicateKeys when two local contracts conflict" in {
```
fabiotudone-da, line 652: Isn't this checked by the engine?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+        getTransactionRejectionReason(result).getInconsistent.getDetails
+      rejectionReason should startWith("DuplicateKeys")
+    }
+
+    "return DuplicateKeys when a create in a rollback conflicts with a global key" in {
+      val builder = TransactionBuilder()
+      val rollback = builder.add(builder.rollback())
+      builder.add(createNode(s"#$freshContractId"), rollback)
+      val transaction = builder.buildSubmitted()
+      val context = commitContextWithContractStateKeys(mykey -> Some(s"#freshContractId"))
```
fabiotudone-da, line 670: Not that it really matters, but for consistency/readability:

```suggestion
      val context = commitContextWithContractStateKeys(mykey -> Some(s"#$freshContractId"))
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/KeyConsistencyValidation.scala
fabiotudone-da: I don't feel strongly about this, I agree `+` didn't really help the purpose.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: No it’s not, I’m trying to get us there but atm the engine does absolutely no duplicate key checks and this allows you to create duplicate keys in some cases that kvutils does not catch.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
-        create(contractId, signatories = Seq("Alice"), keyAndMaintainer = Some(aKey -> "Alice"))
+    def createNode(contractId: String): Create =
+      create(contractId, signatories = Seq("Alice"), keyAndMaintainer = Some(aKey -> "Alice"))
 
-      def newLookupByKeySubmittedTransaction(found: Boolean): SubmittedTransaction =
-        TransactionBuilder.justSubmitted(
-          txBuilder.lookupByKey(createNode(contractId = s"#$freshContractId"), found)
-        )
+    def freshContractId: String =
+      s"testContractId-${UUID.randomUUID().toString.take(10)}"
```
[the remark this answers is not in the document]
cocreature: You can only forward reference if you have no `val` in between and the `val` for the key broke it.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: Whether it conflicts with a local key really depends on the test. This is used in a bunch of of different ways across the tests. I’ve just called it `conflictingKey` now which seems vague enough to still be useful.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: Depends™. If there is no entry, kvutils failed to track it’s inputs correctly. I think the test still fails but that’s not what we want to test here. This entry means kvutils knows it’s an input and the key does not exist before.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -133,21 +133,45 @@ private[transaction] object ContractKeysValidation {
...
-        activeStateKeys = state.activeStateKeys,
-        submittedContractKeysToContractIds =
-          submittedContractKeysToContractIds ++ newContractKeyMappings,
+
+    def onSubmittedContractKeysToContractIds(
+        f: Map[DamlContractKey, Option[
+          RawContractId
+        ]] => Either[KeyValidationError, Map[DamlContractKey, Option[
+          RawContractId
+        ]]]
```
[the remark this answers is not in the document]
cocreature: good idea 793e4079849830c237abc4bcec29ad6c0d3bc382

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+    }
 
-      val lookupByKeyNotFound =
-        newLookupByKeySubmittedTransaction(found = false)
-      val lookupByKeyFound1 = newLookupByKeySubmittedTransaction(found = true)
-      val lookupByKeyFound2 = newLookupByKeySubmittedTransaction(found = true)
+    "return Inconsistent when a contract key resolves to a different contract ID than submitted by a participant" in {
+
+      val cases =
+        Seq((false, Some(s"#$freshContractId")), (true, Some(s"#$freshContractId")), (true, None))
```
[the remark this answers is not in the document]
cocreature: Added some names in ad3ccb2144f211c0acc34af1d7d02f00367d393d Is that what you had in mind or do you want the names somewhere else?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,125 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+          .flatMap { case (found, contractIdAtCommitter) =>
+            Seq(false, true).map(inRollback =>
+              newLookupByKeySubmittedTransaction(found, inRollback) -> contractIdAtCommitter
+            )
+          }
 
-      val cases = Table(
+      val casesTbl = Table(
```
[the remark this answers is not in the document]
cocreature: Sure, no strong feelings here ad3ccb2144f211c0acc34af1d7d02f00367d393d

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -66,27 +67,23 @@ private[transaction] object ContractKeysValidation {
...
+        exerciseBegin = (status, _, exerciseBeginNode) => {
+          val newStatus =
+            onCurrentStatus(status)(
+              checkNodeContractKey(exerciseBeginNode, contractKeysToContractIds, _)
+            )
+          (newStatus, true)
+        },
+        rollbackBegin = (status, _, _) =>
+          // Store state to restore after rollback
+          (status.map(stack => stack.head +: stack), true),
```
fabiotudone-da, line 82: But anyway, aren't rollback blocks flattened?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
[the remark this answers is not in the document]
fabiotudone-da: You could move up `conflictingKey` but it's very minor anyway. Nevermind.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -66,27 +67,23 @@ private[transaction] object ContractKeysValidation {
       transactionEntry: DamlTransactionEntrySummary,
   ): StepResult[DamlTransactionEntrySummary] = {
     val keysValidationOutcome = transactionEntry.transaction
-      .foldInExecutionOrder[KeyValidationStatus](
-        Right(
-          KeyValidationState(activeStateKeys = contractKeyDamlStateKeys)
-        )
+      .foldInExecutionOrder[Either[KeyValidationError, List[KeyValidationState]]](
```
fabiotudone-da, line 70: Maybe a `type KeyValidationStatusStack` could improve readability further?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +595,134 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+      val rollback = builder.add(builder.rollback())
+      builder.add(newCreateNodeWithFixedKey(s"#$freshContractId"), rollback)
+      builder.add(newCreateNodeWithFixedKey(s"#freshContractId"))
+      val transaction = builder.buildSubmitted()
+      val context = commitContextWithContractStateKeys(conflictingKey -> None)
+      val result = validate(context, transaction)
+      result shouldBe a[StepContinue[_]]
+    }
+
+    "return DuplicateKeys between local contracts even if second create is rolled back" in {
```
fabiotudone-da, line 698: Why this asymmetry though? What's the general principle for rollback validation w.r.t. keys here?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
cocreature: No, there is some level of normalization but in general we cannot flatten completely. Consider something like
rb - create
     - rb - create
      - exercise
If the exercise here is on the first create then it is valid. But if the exercise is on the second create then it has been rolled back and should fail.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: Let’s say we only have creates in a rollback. In that case, you can imagine that instead of having a rollback we archive everything at the end of the rollback.

That means we have the two situations:
1. Rollback create first, then another create
This becomes
cid <- create
archive cid
create
The contract has been archived so no duplicate key
2. Create first, then rollback create
cid <- create
cid2 <- create
archive cid2
We archive cid2 here but we already fail on the second create with a duplicate key error.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
cocreature: I don’t mind changing it if you prefer but this type is literally only used here. I’d rather know the actual type I’m folding over than introduce another indirection. That said, your code your choice so if you want the synonym I’ll happily add it.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
fabiotudone-da: Thanks for the explanation. I can think of other test cases though, e.g.:

1. cid <- create with key k
2. consume cid
3. rollback 2
4. cid2 <- create with key k - SHOULD FAIL

Aren't they worth testing as well?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
fabiotudone-da: I agree it's opinable; I'd rather jump and back than having to parse `]]]`, especially if the typedef tells me what I'm dealing with and so I'd have to do it probably only once in a while, but it's a very minor matter. Feel free to leave it as it is.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: Added the testcase in dd56491de514e356cdd1285b8cde46f617f27f82 I think there is a general question of how thorough you want to be wtr to testing here given that there were literally zero checks for duplicate keys before this PR but I like leaving things in a better state than they were before :slightly_smiling_face:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
cocreature: Added a synonym in 0af2beaa7cd6c5925b20f59105c083f676a4db6f

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
fabiotudone-da: I have only a partial view but I think that the lack of unit tests stems at least partially from a misunderstanding (or lack of understanding) about which component checks what.
I agree that it is beyond the scope of this PR to test the interaction with keys of constructs that predate rollback, but I think that tests about interactions of keys with rollback should instead be covered here as exhaustively as it makes sense (i.e., if something can possibly break individually, then it should have a dedicated unit test).
That is, unless we are going to very soon remove duplicated check logic in some way (e.g. b/c the engine is going to do it, or a shared lib for all to use will consolidate the logic).

cocreature: I do hope to do the engine-level checks very soon but regardless, I think the current state covers the cases I can think off (rolled-back create, rolled-back consuming choice, different orders). Do you have something else that you think is missing?

fabiotudone-da: If I understand correctly, only non-key-preserving actions need to be tested with rollback; consuming exercises are key-preserving, aren't they? If so, then testing archival should suffice. Another dimension is indeed global Vs. local; I am not sure if this can break individually but you've already tested it in other cases: how about also testing rolling back an archival of a conflicting global key?

cocreature: > consuming exercises are key-preserving, aren't they

I assume you mean non-consuming here? The archive is a consuming choice.

I added another test for a rolled-back archive of a global cid in 9db05395c906d41a50a8659a20e4fc5f99e46b8d

fabiotudone-da (approved):
Thanks @cocreature!

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -595,58 +596,165 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+      val result = validate(context, transaction)
+      result shouldBe a[StepStop]
+      val rejectionReason =
+        getTransactionRejectionReason(result).getInconsistent.getDetails
+      rejectionReason should startWith("DuplicateKeys")
+    }
+
+    "return InconsistentKeys on conflict local and global contracts even if global was archived in a rollback" in {
+      val builder = TransactionBuilder()
+      val globalCid = s"#freshContractId"
```
miklos-da, line 731:
```scala
   val globalCid = s"#$freshContractId"
```

PR #9479, opened 2021-04-22
Fix validation of rollback nodes

cocreature:
To my own embarassment there is no test in this PR. We don’t have any
direct tests for validation atm. We absolutely need some and I already
discussed with Remy that this is something we should focus on once
exceptions are done but for now I’ll stick to being sad.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
@@ -288,6 +288,21 @@ private final class Validation[Nid, Cid](implicit ECid: Equal[Cid]) {
                   keyIsReplayedBy(key1, key2) &&
                   result1 === result2 =>
               loop(rest1, rest2, stack)
+            case (
+                  Node.NodeRollback(
+                    children1,
+                    version1,
+                  ),
+                  Node.NodeRollback(
+                    children2,
+                    version2,
+                  ),
+                ) if version1 == version2 =>
+              loop(
+                children1.iterator.to(LazyList),
+                children2.iterator.to(LazyList),
+                stack,
+              )
             case _ =>
               Left(ReplayNodeMismatch(recorded, nid1, replayed, nid2))
           }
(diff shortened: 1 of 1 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -288,6 +288,21 @@ private final class Validation[Nid, Cid](implicit ECid: Equal[Cid]) {
...
+                  Node.NodeRollback(
+                    children2,
+                    version2,
+                  ),
+                ) if version1 == version2 =>
+              loop(
+                children1.iterator.to(LazyList),
+                children2.iterator.to(LazyList),
+                stack,
+              )
```
nickchapman-da, line 305: We have dropped `rest1` and `rest2` on the floor.
Don't we need to still process these?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -288,6 +288,21 @@ private final class Validation[Nid, Cid](implicit ECid: Equal[Cid]) {
...
             case _ =>
               Left(ReplayNodeMismatch(recorded, nid1, replayed, nid2))
```
nickchapman-da, line 307: I'm sad the scala compiler didn't discover statically that the case of `Rollback` nodes was missing. Initially I though the blame was our this catch-all...
```
            case _ =>
              Left(ReplayNodeMismatch(recorded, nid1, replayed, nid2))
```

And I wondered if being more precise would have provoked the desired unmatched case warnings when the new node type was added....

But no. There are no warnings even if the catch-all is removed. Scala is broken  :(

nickchapman-da (changes requested):
I think there is a bug,

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
cocreature: good point, I’ve pushed them to the stack like we do for exercise nodes 912b0f9e8e90b6f26ecfcc42455a1d5977d36016

nickchapman-da: :+1:

nickchapman-da (approved):

PR #9486, opened 2021-04-23
Support rollback nodes in ActiveStateManager

cocreature:
If some one can point me to existing tests for ActiveStateManager,
I’ll happily extend them I failed to find any.

I did test it against #9400 and it fixes most tests and the failing
ones fail in other parts so at that level it seems to work as
expected.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
@@ -266,8 +266,8 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
           exerciseEnd = (acc, _, _) => acc,
           rollbackEnd = (acc, _, _) => {
             val (beforeRollback, rest) = acc.rollbackStates match {
-              case Seq() => sys.error("IMPOSSIBLE: Rollback end but rollback stack is empty")
-              case head +: tail => (head, tail)
+              case Nil => sys.error("IMPOSSIBLE: Rollback end but rollback stack is empty")
+              case head :: tail => (head, tail)
             }
             // Discard archives in the rollback but mark contracts created in the rollback as archived.
             // This means that at the end of the traversal, we can use archivedIds to filter out
--- a/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
@@ -25,7 +25,14 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
     initialState: => ALS
 ) {
 
-  // State that is restored on rollback.
+  /** State that is restored on rollback.
+    * @param createdIds The set of local contract ids created in the transaction.
+    * @param archivedIds The set of contract ids
+    *  archived by the transaction or local contracts whose create has been rolled back.
+    *  In other words, contracts that we know won’t be active at the
+    *  end of the transaction.
+    * @param als The active ledger state
+    */
   private case class RollbackState(
       createdIds: Set[ContractId],
       archivedIds: Set[ContractId],
@@ -38,11 +45,12 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
   /** The accumulator state used while validating the transaction
     * and to record the required changes to the ledger state.
     * @param currentState The current state
+    * @param rollbackStates The stack of states at the beginning of a rollback node
+    *  so we can restore the previous state when leaving a rollback node.
+    *  The most recent rollback state comes first.
     * @param errs The list of errors produced during validation.
     * @param parties The set of parties that are used
     *  in the transaction and will be implicitly allocated
-    * @param archivedIds The set of contract ids
-    *  archived by the transaction or local contracts whose create has been rolled back.
     */
   private case class AddTransactionState(
       currentState: RollbackState,
--- a/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
@@ -84,7 +84,12 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
 
   private object AddTransactionState {
     def apply(acs: ALS): AddTransactionState =
-      AddTransactionState(RollbackState(Set.empty, Set.empty, Some(acs)), List(), Set(), Set.empty)
+      AddTransactionState(
+        RollbackState(Set.empty, Set.empty, Some(acs)),
+        List(),
+        Set.empty,
+        Set.empty,
+      )
   }
 
   /** A higher order function to update an abstract active ledger state (ALS) with the effects of the given transaction.
--- a/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
@@ -271,7 +271,10 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
             }
             // Discard archives in the rollback but mark contracts created in the rollback as archived.
             // This means that at the end of the traversal, we can use archivedIds to filter out
-            // both local and global contracts that are no longer active.
+            // both local and global contracts that are no longer active. This is required because
+            // we use it to filter out the divulged contracts to those that are active at the end of the
+            // transaction. Divulgence also affects contracts in a rollback so we need to filter
+            // those out.
             val newArchived =
               beforeRollback.archivedIds union (acc.currentState.createdIds diff beforeRollback.createdIds)
             acc.copy(
--- a/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
@@ -25,17 +25,40 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
     initialState: => ALS
 ) {
 
+  // State that is restored on rollback.
+  private case class RollbackState(
+      createdIds: Set[ContractId],
+      archivedIds: Set[ContractId],
+      als: Option[ALS],
+  ) {
+    def mapAcs(f: ALS => ALS): RollbackState =
+      copy(als = als.map(f))
+  }
+
+  /** The accumulator state used while validating the transaction
+    * and to record the required changes to the ledger state.
+    * @param currentState The current state
+    * @param errs The list of errors produced during validation.
+    * @param parties The set of parties that are used
+    *  in the transaction and will be implicitly allocated
+    * @param archivedIds The set of contract ids
+    *  archived by the transaction or local contracts whose create has been rolled back.
+    */
   private case class AddTransactionState(
-      acc: Option[ALS],
+      currentState: RollbackState,
+      rollbackStates: List[RollbackState],
       errs: Set[RejectionReason],
       parties: Set[Party],
-      archivedIds: Set[ContractId],
   ) {
 
-    def mapAcs(f: ALS => ALS): AddTransactionState = copy(acc = acc map f)
+    def mapAcs(f: ALS => ALS): AddTransactionState =
+      copy(currentState = currentState.mapAcs(f), rollbackStates = rollbackStates.map(_.mapAcs(f)))
 
     def result: Either[Set[RejectionReason], ALS] = {
-      acc match {
+      if (!rollbackStates.isEmpty) {
+        sys.error(s"IMPOSSIBLE finished transaction but rollback states is not empty")
+      }
+      currentState.als match {
         case None =>
           if (errs.isEmpty) {
             sys.error(s"IMPOSSIBLE no acc and no errors either!")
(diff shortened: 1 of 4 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -25,17 +25,40 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
...
+    * @param errs The list of errors produced during validation.
+    * @param parties The set of parties that are used
+    *  in the transaction and will be implicitly allocated
+    * @param archivedIds The set of contract ids
+    *  archived by the transaction or local contracts whose create has been rolled back.
+    */
   private case class AddTransactionState(
-      acc: Option[ALS],
+      currentState: RollbackState,
+      rollbackStates: List[RollbackState],
```
sofiafaro-da, line 49: What does the `rollbackStates` field represent? Is it a stack of rollback states? Might be worth documenting above.

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -53,7 +76,7 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
 
   private object AddTransactionState {
     def apply(acs: ALS): AddTransactionState =
-      AddTransactionState(Some(acs), Set(), Set.empty, Set.empty)
+      AddTransactionState(RollbackState(Set.empty, Set.empty, Some(acs)), List(), Set(), Set.empty)
```
sofiafaro-da, line 79: What's the difference between `Set()` and `Set.empty`?

cocreature: None, the former might be a tiny bit slower. Changed to `Set.empty` in ebe77bbc90c77749e29a9fad6b7f0077fc3c0713 (I just kept what was there before).

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
cocreature: Added some comments in 33c5e4f9f808e5821bb34d3b469f03aba52b61af

sofiafaro-da: Thanks!

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
sofiafaro-da: Thanks for the explanation! :)

nickchapman-da (commented):
I'm studying the details of how the ACS code worked before, so I have some chance of being able to give this PR a sensible review!

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -69,146 +105,184 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
...
-
-    val divulgedContractIds = divulgence -- st.archivedIds
+          },
+          rollbackBegin = (acc, _, _) => {
+            (acc.copy(rollbackStates = acc.currentState +: acc.rollbackStates), true)
+          },
+          leaf = (acc, nodeId, node) => handleLeaf(acc, nodeId, node),
+          exerciseEnd = (acc, _, _) => acc,
+          rollbackEnd = (acc, _, _) => {
+            val (beforeRollback, rest) = acc.rollbackStates match {
```
nickchapman-da, line 268: Prefer vanilla matching with `Nil` and `::` to avoid disabling exhaustive pattern match checking.
```
              case Nil => sys.error("IMPOSSIBLE: Rollback end but rollback stack is empty")
              case head :: tail => (head,tail)
```
Currently the `Seq()` case can be removed and the compiler won't complain. Scala, eh? :man_shrugging:

cocreature: b723f0518735980846e2b6d3749402541660bca8

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -69,146 +105,184 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
...
+              case Seq() => sys.error("IMPOSSIBLE: Rollback end but rollback stack is empty")
+              case head +: tail => (head, tail)
+            }
+            // Discard archives in the rollback but mark contracts created in the rollback as archived.
+            // This means that at the end of the traversal, we can use archivedIds to filter out
+            // both local and global contracts that are no longer active.
+            val newArchived =
+              beforeRollback.archivedIds union (acc.currentState.createdIds diff beforeRollback.createdIds)
+            acc.copy(
+              currentState = beforeRollback.copy(
```
nickchapman-da, line 278: Why doesn't `rollbackEnd` restore the `currentState` to exactly as it was when pushed?
Why is any effect allowed to persist?

cocreature: Because we need `archiveIds` to contain all contracts that are not active at the moment including archived contracts and rolled back creates. The only purpose of `archivedIds` is to filter divulged contracts to those that are still active at the end.

nickchapman-da: Isn't it the case that the ALS which contained the contracts created within the rollback, has already been dropped on the floor?

So what is it, that contains these contracts, which needs to be filtered?

cocreature: 5 lines below, line 285. divulgence includes both local and global contracts.

cocreature: Tried to make it a bit clearer in c1c6f3eebbb70fa442678421fd27671cc4f5eab6

nickchapman-da: Thanks.

So, to check I understand this correctly...

By _local_ contracts, you mean contracts created within a rollback?

The entry point to the `ActiveLedgerStateManager` code is a call to `addTransaction`, which is provided the `divulgence` (presumable already computed by an earlier transaction traversal, which includes local contracts created in a rollback node - why?).

So now `addTransaction` must re-track the _local_ contracts, just so it can remove them from the `divulgence`.

This seems very odd. Or am I still missing something?

cocreature: `local` = created within the transaction doesn’t matter if it was within a rollback or not and doesn’t matter if it was archived or not.
Divulgence is computed by `calculateBlindingInfo` which (as it should) also returns divulgence for local contracts (again regardless of whether they are created within a rollback or not).

`ActiveLedgerStateManager` wants divulgence for only those contracts that are active at the end of the transaction so it needs to trim 3 types of contracts:
1. Global (contracts created in a previous transaction) contracts that have been archived contracts.
2. Local contracts created outside of a rollback that have been archived in the same transaction.
3. Local contracts created in a rollback node (you can view them as having an implicit archive at the end of a rollback node).

The code here basically does the “implicit archive at the end of a rollback node”.

I agree it’s not super nice but I don’t really have a great idea for making it nicer without a pretty invasive change.

nickchapman-da: Thanks for the explanation. The code is perfectly _nice_!

I'm just trying to understand where my mental model of _anything in a rollback node never happened_ breaks down. And it seems to be:

- 1: Regarding divulgence. That did/does happen.
- 2: Contracts created within rollbacks did get created, but then we archived them at the end.

Is that a fair summary?

cocreature: I think there are two points here:
1. First the model should be "rollbacks undo writes but they preserve reads". So a fetch in a rollback for example still happened and that matters for example for divulgence. While that’s related I don’t think that’s quite what makes the difference here.
2. The second point is about visibility/privacy. Regardless of whether something is rolled back, we still project in the rollback node. So if there was a create in a rollback node I might still get to see that. That’s why `BlindingInfo` includes nodes under rollback nodes.

The second point here is what matters. Here we only care about visibility of active contracts rather than visibility of the full transaction tree so we need to filter divulgence down to divulgence of active contracts. As mentioned above there are 3 typos of contracts that might no longer be active at the end of the transaction one of them being a create under a rollback node.

I think there might be two ways of making this nicer:
1. Expose activeness in blindinginfo.
2. Expose the list of all active contracts in `ALS` so we can filter it down

Both of those seemed to be more invasive than necessary so I left them out of this PR at least and tried to preserve the previous structure.

nickchapman-da: Thanks again for more explanation. This is particular:
_"rollbacks undo writes but they preserve reads"._

nickchapman-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
cocreature: I’ve slightly simplified this after adding a helper method to `Transaction` which hopefully makes that less convoluted 3702149a568dbdfaec6bcd8e057eae80e1d261bb

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -69,146 +100,174 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
...
-                  )
-                }
+                (newState, true)
             }
-        }
-
-    val divulgedContractIds = divulgence -- st.archivedIds
+          },
+          rollbackBegin = (acc, _, _) => {
+            (acc.copy(rollbackStates = acc.currentState +: acc.rollbackStates), true)
```
rautenrieth-da, line 255: This saving of state checkpoints works if `ALS` is a value (like used in the in-memory sandbox classic [here](https://github.com/digital-asset/daml/blob/main/ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/InMemoryActiveLedgerState.scala#L21)).

If however it's an interface to some underlying mutable storage (like used in some ancient migration [here](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/db/migration/postgres/V2_1__Rebuild_Acs.scala#L303)), then you will actually keep modifying the "final" ACS after entering rollback nodes.

This is ok because the above mentioned migration will never run on transactions with rollback nodes.

I assume we still need to check the activeness and ledger time of contracts mentioned within rollback node subtrees? Otherwise we could simply skip the rollback nodes entirely.

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -69,146 +100,174 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
...
-                        acc = Some(acc.addContract(activeContract, Some(gk))),
-                        parties = parties.union(nodeParties),
-                      )
-                    }
+                  )
                 }
-              case ne: N.NodeExercises[_, ContractId] =>
+              } else {
+                // Otherwise, trust that the lookup was valid.
+                // The submitter being unknown means the transaction was submitted on a different participant,
```
rautenrieth-da, line 220: Note unrelated to this PR: this class is used in two places: 
- In-memory sandbox classic. In this case, there is only a single participant and we have full trust in the daml interpretation during command execution.
- In a very old migration of the index database. In this case, we *could* encounter transactions submitted through different participants. This use prevents us from simplifying or removing this class, and we should at some point try to get rid of this migration.

rautenrieth-da (commented):
The change looks good, left some comments.

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
cocreature: > I assume we still need to check the activeness and ledger time of contracts mentioned within rollback node subtrees? Otherwise we could simply skip the rollback nodes entirely.

Yes everything under a rollback node still needs to be valid but the writes resulting from it (i.e., creates and archives) will be rolled back so we do need to be rolled back.

You do make a valid point. Do you have an idea for how to make this better? Should I add a method to the interface to clone an ALS which can be trivially implemented for the in-memory version and can simply throw in the ancient migration (and hopefully we never will use it anywhere new).

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
cocreature: I’ve added a comment where this is used at the top of the class 51f73798e726066f8eb07c3934a0f3ea0c9b5af6

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
rautenrieth-da: :+1: I don't have a better idea, not without a major rewrite of this class.

rautenrieth-da (approved):

PR #9488, opened 2021-04-23
update GPG public key

garyverhaegen-da:
The existing public key is set to expire in May, so we've changed it.

> Note: this _should_ require no other change as the private key is unchanged (i.e. the new public key can be used to verify old signatures), but my understanding of GPG is somewhat limited so :shrug:.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #9489, opened 2021-04-23
Set supported jdbc driver names at compile time

cocreature:
This is mainly to unblock the work on Oracle support in the Ledger API
but I think it’s a sensible thing in general. For the Ledger API,
moving the dependency to the top-level is apparently rather
tricky. Because the SDK bundles everything into a single megajar,
Sandbox depending on the oracle library does also result in the JSON
API and the trigger service will also have the oracle library in scope
and will support Oracle in CE which they should not.

This PR simply hardcodes the list of supported drivers to address
that. Not pretty but does the job.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
@@ -36,6 +36,9 @@ object ContractDao {
     "oracle.jdbc.OracleDriver" -> SupportedJdbcDriver.Oracle,
   )
 
+  def supportedJdbcDriverNames(available: Set[String]): Set[String] =
+    supportedJdbcDrivers.keySet intersect available
+
   def apply(jdbcDriver: String, jdbcUrl: String, username: String, password: String)(implicit
       ec: ExecutionContext
   ): ContractDao = {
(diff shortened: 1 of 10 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
```diff
@@ -36,10 +36,6 @@ object ContractDao {
     "oracle.jdbc.OracleDriver" -> SupportedJdbcDriver.Oracle,
   )
 
-  lazy val supportedJdbcDriverNames = supportedJdbcDrivers.keySet filter { d =>
-    scala.util.Try(Class forName d).isSuccess
-  }
```
S11001001, line 41: Instead, this should be `supportedJdbcDrivers.keySet intersect JdbcDrivers.supportedJdbcDriverNames` with the change in http-json `Main` reverted.

cocreature: Why do you prefer that option? I’d rather move down the opposite direction and have the switch between CE/EE as far up the dependency tree as possible.

S11001001: I'm not moving that switch with this suggestion, merely making it so that adding `foo` to `JdbcDrivers` doesn't imply a promise to immediately support `foo` in json-api (which `--help` will advertise).

cocreature: Makes sense, fixed in 1febf666ea9fc151ddc49558bf9b16fbc032650a. I’ve slightly modified it such that the final switch is still in `Main`. That way we have a chance of getting to the point where we don’t have to compile `ContractDao` twice.

S11001001 (approved):

PR #5629, opened 2020-04-20
Speedy compiler: process packages in topological order

remyhaemmerle-da:
Extract the topological sort from the Recursion validation and use it in the compiler

This refactoring  prepares #5628

### Pull Request Checklist

- [X] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [X] Include appropriate tests
- [X] Set a descriptive title and thorough description
- [X] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9491, opened 2021-04-23
deprecate InsertOrdMap

S11001001:
Following up #9449, `InsertOrdMap` is no longer used anywhere, so we can deprecate it.

```rst
CHANGELOG_BEGIN
- [Scala bindings] ``InsertOrdMap`` is deprecated; it was never
  needed for the release version of ``GenMap``.  Use an ordinary
  ``Map``, or ``Primitive.GenMap`` for Scala codegen, instead.
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

remyhaemmerle-da:
`InsertOrdMap` was not specifically introduced for `GenMap`, it is way older (see [9aed2284](https://github.com/DACH-NY/daml/commit/9aed22842b92d59501e563729a32e19f4c58cf1e). 

I think it is currently not used, but I still believe it could be useful in some places were determinism would be needed and we do not have ordering.

remyhaemmerle-da:
Anyway, I think either `InsertOrdMap`  and `InsertOrdSet` should be deprecated together or none of them.

S11001001:
> `InsertOrdMap` was not specifically introduced for `GenMap`, it is way older (see [9aed2284](https://github.com/DACH-NY/daml/commit/9aed22842b92d59501e563729a32e19f4c58cf1e)).

Thanks for the context. In this case, I was trying to guess what an existing user (who would see the deprecation message) would likely be using `InsertOrdMap` for, so that the error message triggered by their dependency upgrade would reassure them as to the correct changes to make.

> ...but I still believe it could be useful in some places were determinism would be needed and we do not have ordering.

I agree, and it would be reasonable to undeprecate (possibly removed from the public API) if such a use case arises before it is deleted for being deprecated for a significant period of time.

> Anyway, I think either `InsertOrdMap` and `InsertOrdSet` should be deprecated together or none of them.

`InsertOrdSet` is currently used (since #5629); `InsertOrdMap` is not. I don't think we should keep unused code only because there is some used code that happens to share superficial naming similarities with it.

PR #9498, opened 2021-04-26
Switch from `@silent` to `@nowarn`.

SamirTalwar-DA:
This annotation is native to Scala 2.12.13+ and 2.13.2+. It replaces
most usages of `@silent`.

I had to get creative about a couple of use cases that didn't work.
Specifically:

  1.  Suppressing deprecation warnings works, but Scala 2.12 erroneously
      complains that the `@nowarn` is unnecessary.  I had to suppress
      this warning too with `-Ywarn-unused:-nowarn`.
  2.  I can't seem to suppress the warning, "The outer reference in this
      type test cannot be checked at run time." Instead, I have
      refactored the code to remove the warning.

We still need to use the silencer plugin to suppress some warnings about
unused imports (because of compatibility between Scala 2.12 and 2.13),
but this means we no longer need the library, and therefore it is not a
transitive dependency that downstream consumers need to worry about.

### Pull Request Checklist

trigger the build.


bazel-java-deps.bzl
```diff
@@ -41,7 +41,6 @@ def install_java_deps():
             "com.auth0:jwks-rsa:0.11.0",
             "com.chuusai:shapeless_{}:2.3.3".format(scala_major_version),
             "com.github.ben-manes.caffeine:caffeine:2.8.0",
-            "com.github.ghik:silencer-lib_{}:1.7.1".format(scala_version),
```
cocreature, line 44: Did you repin? I guess it won’t change much but I think I’d feel more confident if we at least bump the hash on any change to this file.

bazel_tools/scala.bzl
```diff
@@ -62,6 +62,7 @@ common_scalacopts = version_specific.get(scala_major_version, []) + [
     "-unchecked",
     # warn if using deprecated stuff
     "-deprecation",
+    "-Wconf:cat=deprecation:wv,cat=feature:wv,cat=optimizer:wv",
```
cocreature, line 65: Can we add a comment on what this does?

bazel_tools/scala.bzl
```diff
@@ -82,6 +83,7 @@ common_scalacopts = version_specific.get(scala_major_version, []) + [
     # Gives a warning for functions declared as returning Unit, but the body returns a value
     "-Ywarn-value-discard",
     "-Ywarn-unused:imports",
+    "-Ywarn-unused:-nowarn",
```
cocreature, line 86: What does this do?

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/ScenarioLoader.scala
```diff
@@ -116,12 +119,19 @@ private[sandbox] object ScenarioLoader {
...
     val candidateScenarios = getCandidateScenarios(packages, scenarioQualName)
     val (scenarioRef, scenarioDef) = identifyScenario(packages, scenario, candidateScenarios)
-    @com.github.ghik.silencer.silent("can be used only by sandbox classic.")
-    val scenarioLedger =
-      ScenarioRunner.getScenarioLedger(engine, scenarioRef, scenarioDef, transactionSeed)
+    val scenarioLedger = getScenarioLedger(engine, transactionSeed, scenarioRef, scenarioDef)
     (scenarioLedger, scenarioRef)
   }
 
+  //noinspection ScalaDeprecation
```
cocreature, line 126: What does this do?

cocreature (approved):
lgtm minus a few small comments, thanks!

bazel-java-deps.bzl
SamirTalwar-DA: I did. It's not a direct dependency but it's still a transitive dependency of the plugin, so the Maven pin files didn't change.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/ScenarioLoader.scala
SamirTalwar-DA: IntelliJ suppression I added by accident. I'll remove it. (IntelliJ should be able to understand `@nowarn`, but Scala support is a little lacking.)

bazel_tools/scala.bzl
SamirTalwar-DA: See the PR description. I'll add a comment here.

bazel_tools/scala.bzl
SamirTalwar-DA: I'm removing it because `compatibility` doesn't understand it. Do we use a different version of Scala there?

cocreature: > I'm removing it because `compatibility` doesn't understand it. Do we use a different version of Scala there?

Not deliberately but I wouldn’t be surprised if we ended up getting the two out of sync somehow.

SamirTalwar-DA: Just checked. We do, of course. I'll bump it in a separate PR.

This line was "nice to have" for more verbose warnings anyway; it's not necessary.

SamirTalwar-DA:
Blocked on #9502.

PR #9500, opened 2021-04-26
KVL-874 Add unit tests for Telemetry

hubert-da:
### Pull Request Checklist

trigger the build.


ledger/metrics/src/test/scala/com/daml/telemetry/TelemetryContextSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import io.opentelemetry.api.trace.Span
+import org.scalatest.Assertion
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+/** Other cases are covered by [[TelemetrySpec]] */
+class TelemetryContextSpec extends AnyWordSpecLike with Matchers {
+
+  "DefaultTelemetryContext.runInOpenTelemetryScope" should {
```
hubert-da, line 14: I tried to increase the coverage here but failed. If you have an idea on how to test it further, please let me know. I'm also not sure if this spec is not pointless now...

fabiotudone-da: This unit test only checks that the body is run but there are 2 more things to check for `runInOpenTelemetryScope` implementations: that they set the scope in the current context and that the scope is closed after the body has run, even if an exception is thrown.

fabiotudone-da: Also, `encodeMetadata` is currently not tested at all.

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetrySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+    "return an extracted context" in {
+      val span = tracerProvider
+        .get(InstrumentationName)
+        .spanBuilder(aSpanName)
+        .startSpan()
+      val metadata = DefaultTelemetryContext(span).encodeMetadata()
+
+      val context = DefaultTelemetry.contextFromMetadata(Some(metadata))
+
+      context.encodeMetadata().keySet contains "traceparent"
```
fabiotudone-da, line 59: Since you are asserting on the metadata, you can bind directly the metadata.

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetrySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  "runInSpan" should {
+    "create and finish a span" in {
+      DefaultTelemetry
+        .runInSpan(
+          aSpanName,
+          SpanKind.Internal,
+          anApplicationIdSpanAttribute,
+        ) { tc =>
```
fabiotudone-da, line 80:
```suggestion
        ) { telemetryContext =>
```

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetrySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  "runFutureInSpan" should {
+    "create and finish a span" in {
+      DefaultTelemetry
+        .runFutureInSpan(
+          aSpanName,
+          SpanKind.Internal,
+          anApplicationIdSpanAttribute,
+        ) { tc =>
```
fabiotudone-da, line 97:
```suggestion
        ) { telemetryContext =>
```

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetryContextSpec.scala
hubert-da: Yup, I know. The problem is how to capture these. The API doesn't expose too much.

fabiotudone-da: You could slightly change the API to optionally allow injecting a `Context` and then mock/fake it to detect setting a scope. You should also be able to easily mock a `Scope`.

hubert-da: I wanted to avoid that, as the problem is that we don't have a single method that exposes the `opentelemetry` API, so far.

fabiotudone-da: How about overloading just for the tests and restricting access control?

miklos-da: It should be enough to make the `Tracer` instance injectable into the `TelemetryContext` for testing. I.e., then you can use an in-memory exporter to check what spans have been created for the various test cases (just as you did for `Telemetry`). I would suggest focusing on testing `TelemetryContext` with OpenTelemetry components and not mocking any of the tracing related components, as our primary goal is to prove that OpenTelemetry picks up the created spans via the `TelemetryContext` abstraction.

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetrySpec.scala
```diff
@@ -0,0 +1,142 @@
...
+        .startSpan()
+      val metadata = DefaultTelemetryContext(span).encodeMetadata()
+
+      val context = DefaultTelemetry.contextFromMetadata(Some(metadata))
+
+      metadata.keySet contains "traceparent"
+      context.encodeMetadata() shouldBe metadata
+    }
+
+    "return a root context" in {
```
miklos-da, line 63:
```suggestion
    "return a root context if no metadata was provided" in {
```

ledger/metrics/src/test/scala/com/daml/telemetry/TelemetryContextSpec.scala
hubert-da: I find it a bad practice to add stuff just for test purposes to the production code, obviously. But if you are fine with me adding it, sure.

miklos-da: By allowing injecting the `Tracer` you are just removing dependency on a global object from `TelemetryContext` which is good design anyway.

hubert-da: Yup, initially I even did this. Then, I got inspired by some JUnit 5 extension ;) Let me make that change, thanks!

hubert-da: @fabiotudone-da @miklos-da, done. I didn't have access to the `Scope` but the coverage is much better now.

ledger/metrics/src/test/suite/scala/com/daml/telemetry/TelemetrySpec.scala
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+      val attributes = spanExporter.finishedSpanAttributes
+      attributes should contain(anApplicationIdSpanAttribute)
+      attributes should contain(aCommandIdSpanAttribute)
+    }
+  }
+
+  "runFutureInSpan" should {
+    "create and finish a span" in {
```
miklos-da, line 83: Can you also check the case when an exception is thrown in the body, please? I.e., the span should record the exception from the failed `Future`.

ledger/metrics/src/test/suite/scala/com/daml/telemetry/TelemetryContextSpec.scala
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
miklos-da, line 1: Can you add test case(s) for `runInNewSpan` as well, please?

miklos-da (approved):

ledger/metrics/src/test/suite/scala/com/daml/telemetry/TelemetrySpec.scala
hubert-da: Sure, will do in the next PR.

ledger/metrics/src/test/suite/scala/com/daml/telemetry/TelemetryContextSpec.scala
hubert-da: Sure, will do in the next PR.

PR #9503, opened 2021-04-26
Add a helper for getting all inactive contracts to Transaction

cocreature:
This comes up in a few places and rather than inlining the
implementation everywhere, I’d rather rely on this.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
+++ b/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
@@ -17,7 +17,6 @@ import com.daml.lf.transaction.Node.{
   NodeRollback,
 }
 import com.daml.lf.transaction.{
-  BlindingInfo,
   GenTransaction,
   NodeId,
   TransactionVersion,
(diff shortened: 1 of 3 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/ledger/BlindingTransaction.scala
```diff
@@ -112,9 +112,15 @@ object BlindingTransaction {
         processNode(s, initialParentExerciseWitnesses, nodeId)
       }
 
+    val inactiveContracts = tx.inactiveContracts
+
+    val activeDivulgence =
+      finalState.divulgences.filterKeys(coid => !inactiveContracts.contains(coid))
```
S11001001, line 118:
```suggestion
    val activeDivulgence =
      finalState.divulgences -- tx.inactiveContracts
```

cocreature:
I’ve slightly changed this to only add the helper to `Transaction` and leave `BlindingInfo` untoutched. The latter is serialized in a few places and evaluating whether we can afford serializing this in addition to the existing BlindingInfo needs more thought. This way we can use it in the places where we need it.

S11001001 (approved):

remyhaemmerle-da (commented):
Nice.

PR #9506, opened 2021-04-26
Support rollback nodes in TransactionIndexing

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
@@ -51,7 +51,6 @@ final class TransactionIndexingSpec extends AnyWordSpec with Matchers {
       val party = Party.assertFromString(partyStr)
       val bobStr = "Bob"
       val bob = Party.assertFromString(bobStr)
-//    val offset = LedgerOffset.Absolute(LedgerString.assertFromString("offset"))
 
       def create(builder: TransactionBuilder, id: String) =
         builder.create(
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
@@ -6,22 +6,27 @@ package com.daml.platform.store.dao.events
 import java.time.Instant
 
 import com.daml.ledger
+// import com.daml.ledger.api.domain.LedgerOffset
 import com.daml.ledger.participant.state.v1.Offset
+import com.daml.lf.data.ImmArray
+import com.daml.lf.ledger.BlindingTransaction
 import com.daml.lf.transaction.BlindingInfo
 import com.daml.lf.transaction.test.TransactionBuilder
+import com.daml.lf.value.{Value => V}
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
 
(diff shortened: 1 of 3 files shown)


ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
```diff
@@ -40,5 +45,125 @@ final class TransactionIndexingSpec extends AnyWordSpec with Matchers {
           .ContractWitnessesInfo(Set.empty, aDivulgence)
       )
     }
+
+    "filter out rollback nodes" in {
+      val partyStr = "Alice"
+      val party = Party.assertFromString(partyStr)
+      val bobStr = "Bob"
+      val bob = Party.assertFromString(bobStr)
+//    val offset = LedgerOffset.Absolute(LedgerString.assertFromString("offset"))
```
tudor-da, line 54: Can you remove this commented line?

tudor-da:
This implementation targets the legacy index schema, while we are already working on the new schema (in package `appendonlydao`. I am not sure if it should be you to implement it in the upcoming schema. But if we decide to port it later, we need to keep track of it. @mziolekda @nmarton-da what do you think?

ledger/participant-integration-api/src/test/suite/scala/platform/store/dao/events/TransactionIndexingSpec.scala
cocreature: 6dff0ee6d9b8156e816de569e1fd4e10ab5d1a7b

cocreature:
> This implementation targets the legacy index schema, while we are already working on the new schema (in package `appendonlydao`. I am not sure if it should be you to implement it in the upcoming schema. But if we decide to port it later, we need to keep track of it. @mziolekda @nmarton-da what do you think?

Where is the new schema code for this? Contrary to some other files, this only seems to exist once. We need exception support everywhere so this absolutely needs to be ported to the new schema. 

I care less about who has to do the work and more that we get it working so happy to take a shot or help someone if you can point me to the right place.

I think it probably makes sense to keep that for a separate PR though.

tudor-da:
> > This implementation targets the legacy index schema, while we are already working on the new schema (in package `appendonlydao`. I am not sure if it should be you to implement it in the upcoming schema. But if we decide to port it later, we need to keep track of it. @mziolekda @nmarton-da what do you think?
> 
> Where is the new schema code for this? Contrary to some other files, this only seems to exist once. We need exception support everywhere so this absolutely needs to be ported to the new schema.
> 
> I care less about who has to do the work and more that we get it working so happy to take a shot or help someone if you can point me to the right place.
> 
> I think it probably makes sense to keep that for a separate PR though.

Ok, now it is more clear. It seems the [new code](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/UpdateToDBDTOV1.scala#L140-L263) would be already ignoring Rollback nodes. So nothing is needed in that area, I believe. Thanks, @moritzkiefer-da

tudor-da (approved):

cocreature:
> Ok, now it is more clear. It seems the [new code](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/UpdateToDBDTOV1.scala#L140-L263) would be already ignoring Rollback nodes. So nothing is needed in that area, I believe. Thanks, @moritzkiefer-da

Looking at this, I’m not entirely sure that’s right.I think you filter out rollback nodes but you will still get a create under a rollback node. I’ll make sure to look into this and either fix it myself or reach out to you if I need help.

PR #9509, opened 2021-04-27
DPP-359 Add indexer flow level benchmark

rautenrieth-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
+++ b/ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
@@ -167,7 +167,7 @@ class IndexerBenchmark() {
   }
 
   private[this] def createReadService(
-      updates: Iterator[(Offset, Update)],
+      updates: Iterator[(Offset, Update)]
   ): ReadService = {
     val initialConditions = LedgerInitialConditions(
       IndexerBenchmark.LedgerId,
--- a/ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
+++ b/ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
@@ -41,7 +41,7 @@ class IndexerBenchmark() {
     * and functional tests.
     */
   def runWithEphemeralPostgres(
-      createUpdates: String => Future[() => Iterator[(Offset, Update)]],
+      createUpdates: Config => Future[Iterator[(Offset, Update)]],
       config: Config,
   ): Future[Unit] = {
     PostgresResource
@@ -53,7 +53,7 @@ class IndexerBenchmark() {
   }
 
   def run(
-      createUpdates: String => Future[() => Iterator[(Offset, Update)]],
+      createUpdates: Config => Future[Iterator[(Offset, Update)]],
       config: Config,
   ): Future[Unit] = {
     newLoggingContext { implicit loggingContext =>
@@ -69,10 +69,10 @@ class IndexerBenchmark() {
       val indexerEC = ExecutionContext.fromExecutor(indexerE)
 
       println("Generating state updates...")
-      val updates = Await.result(createUpdates(config.updateSource), Duration(10, "minute"))
+      val updates = Await.result(createUpdates(config), Duration(10, "minute"))
 
       println("Creating read service and indexer...")
-      val readService = createReadService(updates, config)
+      val readService = createReadService(updates)
       val indexerFactory = new JdbcIndexer.Factory(
         ServerRole.Indexer,
         config.indexerConfig,
@@ -167,8 +167,7 @@ class IndexerBenchmark() {
   }
 
   private[this] def createReadService(
-      updates: () => Iterator[(Offset, Update)],
-      config: Config,
+      updates: Iterator[(Offset, Update)],
   ): ReadService = {
     val initialConditions = LedgerInitialConditions(
       IndexerBenchmark.LedgerId,
@@ -185,18 +184,8 @@ class IndexerBenchmark() {
         Source.single(initialConditions)
       }
       override def stateUpdates(beginAfter: Option[Offset]): Source[(Offset, Update), NotUsed] = {
-        assert(beginAfter.isEmpty)
-        config.updateCount match {
-          case None =>
-            Source.fromIterator(updates)
-          case Some(updateCount) =>
-            Source
-              .cycle(() => {
-                println("(Re)starting the stream of updates")
-                updates()
-              }) // TODO append-only: cycling the exact same updates is probably not a good idea
-              .take(updateCount)
-        }
+        assert(beginAfter.isEmpty, s"beginAfter is $beginAfter")
+        Source.fromIterator(() => updates)
       }
       override def currentHealth(): HealthStatus = HealthStatus.healthy
     }
@@ -212,7 +201,7 @@ object IndexerBenchmark {
 
   def runAndExit(
       args: Array[String],
-      updates: String => Future[() => Iterator[(Offset, Update)]],
+      updates: Config => Future[Iterator[(Offset, Update)]],
   ): Unit = {
     val config: Config = Config.parse(args).getOrElse {
       sys.exit(1)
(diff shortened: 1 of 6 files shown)


ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
```diff
@@ -0,0 +1,227 @@
...
+    }
+    IndexerBenchmark.runAndExit(config, updates)
+  }
+
+  def runAndExit(
+      config: Config,
+      updates: String => Future[Iterator[(Offset, Update)]],
+  ): Unit = {
+    if (config.indexerConfig.jdbcUrl.isEmpty) {
+      new IndexerBenchmark().runWithEphemeralPostgres(updates, config)
```
nmarton-da, line 222: Why would we need ephemeral postgres support here?

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
```diff
@@ -0,0 +1,227 @@
...
+        _ <- config.metricsReporter.fold(Resource.unit)(reporter =>
+          ResourceOwner
+            .forCloseable(() => reporter.register(metrics.registry))
+            .map(_.start(config.metricsReportingInterval.getSeconds, TimeUnit.SECONDS))
+            .acquire()
+        )
+
+        _ = println("Setting up the index database...")
+        indexer <- Await
+          .result(indexerFactory.migrateSchema(false), Duration(100, "hour"))
```
nmarton-da, line 99: seconds maybe? ;)

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
```diff
@@ -0,0 +1,227 @@
...
+             |--------------------------------------------------------------------------------
+             |Indexer benchmark results
+             |--------------------------------------------------------------------------------
+             |
+             |Input:
+             |  source:   ${config.updateSource}
+             |  count:    ${config.updateCount}
+             |  jdbcUrl:  ${config.indexerConfig.jdbcUrl}
+             |
+             |Indexer parameters:
```
nmarton-da, line 123: We seem to be able to run this against any indexer implementation (which I like a lot), would be able somehow to render the indexer config with reflection, so we do not couple this output to a particular version of a particular implementation? (reflection maybe?)

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
```diff
@@ -0,0 +1,227 @@
...
+             |  inputMappingParallelism:   ${config.indexerConfig.inputMappingParallelism}
+             |  ingestionParallelism:      ${config.indexerConfig.ingestionParallelism}
+             |  submissionBatchSize:       ${config.indexerConfig.submissionBatchSize}
+             |  batchWithinMillis:         ${config.indexerConfig.batchWithinMillis}
+             |  tailingRateLimitPerSecond: ${config.indexerConfig.tailingRateLimitPerSecond}
+             |
+             |Result:
+             |  duration:    $duration
+             |  updates:     $updates
+             |  updates/sec: ${updates / duration}
```
nmarton-da, line 133: This number can be deceiving. As I understood the duration includes the whole export ingestion. With CHESS+ exports we have an initial phase (~100 updates), and the trade registration updates themselves. So this average TP will include for example all the heavy package uploads too.
With the CHESS+ trade registration flows is fairly easy to figure out the length of the prefix: we just have to have the number of total update in the export, and substract from it twice the amount of expected trade-registrations:
- for that to figure we need reporting of the total number of updates contained in the export in this tool
- configuration of this prefix for this tool
- a way to ingest first the prefix, and only start the counter/metrics/timer afterwards
(you can capture this behavior on the PoC branch)

I understand, that this would be some work to impement. Alternatively it would be useful to add per second/per 10 second metrics reporting to the logs, where the prefix ingestion can be omitted visually.

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
```diff
@@ -0,0 +1,227 @@
...
+            metrics.daml.parallelIndexer.inputMapping.batchSize.getSnapshot
+          )}
+             |  inputMapping.duration:      ${histogramToString(
+            metrics.daml.parallelIndexer.inputMapping.duration.getSnapshot
+          )}
+             |  inputMapping.duration.rate: ${metrics.daml.parallelIndexer.inputMapping.duration.getMeanRate}
+             |  ingestion.duration:         ${histogramToString(
+            metrics.daml.parallelIndexer.ingestion.duration.getSnapshot
+          )}
+             |  ingestion.duration.rate:    ${metrics.daml.parallelIndexer.ingestion.duration.getMeanRate}
```
nmarton-da, line 146: here as well: probably contains ingesting/decoding the CHESS+ packages as well

nmarton-da (commented):
nice and lightweight!
I think we need some improvements to have more qualitative results.
I am not sure that supporting ephemeral PG here is needed

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
rautenrieth-da: Added a comment to `runWithEphemeralPostgres`:
```
Using an uncontrolled local database does not give good performance results, but is useful for development and functional tests.
```

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
rautenrieth-da: Good point, I have tweaked all timeouts.

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
rautenrieth-da: I have added a simple output of the entire indexerConfig using the default case class rendering.

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
rautenrieth-da: Very good point. I have added a disclaimer to the console output that these numbers are not representative and users should inspect the metrics instead.

It's easy to dump metrics every 10 sec to the console using the `--metrics-reporter=console --metrics-reporting-interval=PT0M10S` CLI flags, or to a CSV file using `--metrics-reporter=csv:///PATH`. The output is very verbose (it includes _all_ metrics and not just the ones of interest) and not as nice as in your PoC benchmark, but the end goal is to have metrics appear in Grafana anyway.

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
nmarton-da: if it is just for testing, I am wondering whether we can move this into a test package.... but that is not seem to be easy, neither something we do in other case so fine by me :ok:

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
nmarton-da: I think that is not particularly useful because it wont have the names of the properties, just the property, so we would need to count positions in order to understand (not to mention if config changed between counting positions :) )
But I am also ok with leaving it as it is right now, and we improve on this as we need it for other indexers :ok:

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
nmarton-da: sounds great, let's see how that works in practice :ok:

ledger/indexer-benchmark/src/main/scala/ledger/indexerbenchmark/IndexerBenchmark.scala
nmarton-da: actually I take this back entirely: we anyway have to roll our config, and since we need to change that anyway as things change, to adapt this is not that big of an overhead

nmarton-da (commented):
LGTM, let's see it in  action! :rocket:

nmarton-da (approved):
Oh, forgot to approve this before...

PR #9510, opened 2021-04-27
Update LF spec transaction structure.

sofiafaro-da:
Just some small changes to take into account the new ledger model transaction structure (see #9396). This PR doesn't fix all the issues with the LF spec not matching the ledger model, but it tries to avoid widening that gap.

* Removed itr, iexercise which we don't need.
* Removed the exception type and value from the rollback node.
* We're not calling rollback nodes ledger actions anymore, so separate it out.

Part of #8020

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thank you!

PR #9511, opened 2021-04-27
Switch to an environment variable for enabling Oracle tests.

SamirTalwar-DA:
This allows developers who don't have access to Oracle to build the tests.

It also stops IntelliJ from complaining that the targets don't exist.

### Pull Request Checklist

trigger the build.


ledger-service/http-json-oracle/BUILD.bazel
```diff
@@ -60,4 +57,4 @@ da_scala_test(
         "//libs-scala/oracle-testing",
         "//libs-scala/ports",
     ],
-)
+) if oracle_testing else None
```
moritzkiefer-da, line 60: Can we instead set it to `manual`? testing that it builds still seems useful.

ledger/participant-integration-api/BUILD.bazel
```diff
@@ -295,11 +223,73 @@ da_scala_test_suite(
...
+        "@maven//:io_netty_netty_transport",
+        "@maven//:org_flywaydb_flyway_core",
+        "@maven//:org_mockito_mockito_core",
+        "@maven//:org_reactivestreams_reactive_streams",
+        "@maven//:org_slf4j_slf4j_api",
+    ],
 )
 
-da_scala_test_suite(
+# The "manual" test tag doesn't seem to work with `da_scala_test_suite`.
```
moritzkiefer-da, line 291: I’m confused. You switched from `da_scala_test_suite` but then you’re still not using `manual`. Can we do one or the other?

triggers/service/BUILD.bazel
```diff
@@ -239,19 +240,16 @@ da_scala_test_suite(
     deps = test_deps,
 )
 
-da_scala_test_suite(
+# The "manual" test tag doesn't seem to work with `da_scala_test_suite`.
```
moritzkiefer-da, line 243: same thing

ledger/participant-integration-api/BUILD.bazel
SamirTalwar-DA: Yeah, that was my original plan and then I switched it. I'll put it back.

cocreature: I’d prefer the manual option and fixing `da_scala_test_suite`.

SamirTalwar-DA: I did try to fix `da_scala_test_suite` but I can't find the problem. Can you see one?

I think it might be an issue in `scala_test_suite`, but I can't be sure.

SamirTalwar-DA: Switched to manual execution with the tag.

cocreature: hm, I thought it might not pass the tag along but doesn’t seem to be it. Fine to leave it out of this PR.

cocreature (approved):
Thanks!

triggers/service/BUILD.bazel
```diff
@@ -239,14 +240,12 @@ da_scala_test_suite(
...
+da_scala_test(
```
cocreature, line 244: there are a bunch of unused deps here. Not caused by your PR but `da_scala_test_suite` (deliberately) disables unused dep checks so now it blows up. Should be able to just kill them.

SamirTalwar-DA: Oh, missed those somehow. I purged the ones in participant-integration-api.

Not sure how `bazel build //...` didn't catch that.

SamirTalwar-DA: Seems I needed to run `DAML_ORACLE_TESTING=true bazel build //...`, even with the change to the manual tags. I guess they don't get built otherwise.

ledger/participant-integration-api/BUILD.bazel
aherrmann-da: Looking at the [implementation of `scala_test_suite`](https://github.com/bazelbuild/rules_scala/blob/b85d1225d0ddc9c376963eb0be86d9d546f25a4a/scala/private/rules/scala_test.bzl#L125-L144), it does forward `kwargs` to the individual `scala_test` targets, but it doesn't forward any tags to the final `test_suite` target. So, the final `test_suite` target will still be non-manual. I'd say that's a bug in rules_scala.

PR #9513, opened 2021-04-27
kvutils: Log a missing input state warning without the stack trace. [KVL-796]

SamirTalwar-DA:
Missing input state is the fault of the participant, not the ledger.
Logging it as a warning, with a stack trace, makes it look like the
ledger itself failed, when actually the participant failed to provide
the correct keys (potentially due to data corruption, but it could also
be malicious).

This drops the stack trace to make the error less scary. We still log at
WARN level because this sort of thing shouldn't happen very often, and
at the very least, the participant operator should be informed, and the
participant should be refreshed with the correct data somehow.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
@@ -276,7 +276,9 @@ private[kvutils] class TransactionCommitter(
             )
         } catch {
           case err: Err.MissingInputState =>
-            logger.warn("Exception during model conformance validation.", err)
+            logger.warn(
+              s"Model conformance validation failed due to a missing input state (most likely due to invalid state on the partcipant), correlationId=${transactionEntry.commandId}"
+            )
             reject(
               commitContext.recordTime,
               buildRejectionLogEntry(transactionEntry, RejectionReason.Disputed(err.getMessage)),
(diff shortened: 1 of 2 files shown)


miklos-da (approved):
This makes a lot of sense, thanks for the improvement, Samir!

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -276,7 +276,9 @@ private[kvutils] class TransactionCommitter(
             )
         } catch {
           case err: Err.MissingInputState =>
-            logger.warn("Exception during model conformance validation.", err)
+            logger.warn(
+              s"Model conformance validation failed due to a missing input state (most likely due to invalid state on the partcipant), correlationId=${transactionEntry.commandId}"
+            )
```
fabiotudone-da, line 281: Out of scope for this PR but we should probably use contextual logging in the committers.

fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
SamirTalwar-DA: Definitely. I'm thinking of making that change now.

SamirTalwar-DA: See https://github.com/digital-asset/daml/pull/9516.

PR #9516, opened 2021-04-27
kvutils: Use the contextualized logger in the committer.

SamirTalwar-DA:
This means we don't have to provide context explicitly on each log line; it'll get passed through.

We include extra logging context, extracted from the state at each step.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA:
Unfortunately this PR is mostly boilerplate changes due to passing logging contexts around explicitly. Perhaps we can change `Step` to be a trait? This will increase boilerplate at the declarations but remove it from the call sites.

fabiotudone-da:
> Unfortunately this PR is mostly boilerplate changes due to passing logging contexts around explicitly. Perhaps we can change `Step` to be a trait? This will increase boilerplate at the declarations but remove it from the call sites.

I think turning `Step` into a trait can't harm, quite the opposite. If you also see it reducing the diff, I an in favor of it.

SamirTalwar-DA:
I made the change but it enlarges the diff considerably. So I think we should do it separately.

fabiotudone-da (approved):
It looks much better now IMO, thanks!

PR #9517, opened 2021-04-27
First draw of an interface separating ingestion and storage

nmarton-da:
The intent of this change is to make the first step in this direction
in order to support dpp-336 work of sanbdox-classic integration by
injecting DAO functionality via this interface.
Level of quality is still pre-production, hence TODO comments.
Planned next step: move StorageBackend interface, and implementation
to platform/store, to it's final place.

* Introduce StorageBackend interface
* Decouple event-seq-id assignment logic from storage specific batching
* Pull out of batching step from input mapping, execution too
* Switch to stateless DAO functions
* Switch to DBDispatcher instead of custom JDBC Connection pool
* Introduce/adapt metrics
* Naturally extend configuration
* Move RunningBatch layer to ParallelIndexerFactory
* Remove dead code

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
+++ b/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
@@ -549,6 +549,8 @@ final class Metrics(val registry: MetricRegistry) {
     object parallelIndexer {
       private val Prefix: MetricName = daml.Prefix :+ "parallel_indexer"
 
+      val initialization = new DatabaseMetrics(registry, Prefix, "initialization")
+
       // Number of state updates persisted to the database
       // (after the effect of the corresponding Update is persisted into the database,
       // and before this effect is visible via moving the ledger end forward)
@@ -573,18 +575,32 @@ final class Metrics(val registry: MetricRegistry) {
         val batchSize: Histogram = registry.histogram(Prefix :+ "batch_size")
       }
 
-      // Ingestion stage
-      // Parallel ingestion of prepared data into the database
-      object ingestion {
-        private val Prefix: MetricName = parallelIndexer.Prefix :+ "ingestion"
+      // Batching stage
+      // Translating batch data objects to db-specific DTO batches
+      object batching {
+        private val Prefix: MetricName = parallelIndexer.Prefix :+ "batching"
 
         // Bundle of metrics coming from instrumentation of the underlying thread-pool
         val executor: MetricName = Prefix :+ "executor"
+
+        // The latency, which during an update element is residing in the batching-stage.
+        // Since batches are involved, this duration is divided by the batch size.
+        val duration: Timer = registry.timer(Prefix :+ "duration")
+      }
+
+      // Ingestion stage
+      // Parallel ingestion of prepared data into the database
+      object ingestion extends DatabaseMetrics(registry, Prefix, "ingestion") {
+        private val Prefix: MetricName = parallelIndexer.Prefix :+ "ingestion"
 
         // The latency, which during an update element is residing in the ingestion.
         // Since batches are involved, this duration is divided by the batch size.
         val duration: Timer = registry.timer(Prefix :+ "duration")
       }
+
+      // Tail ingestion stage
+      // The throttled update of ledger end parameters
+      val tailIngestion = new DatabaseMetrics(registry, Prefix, "tail_ingestion")
     }
 
     object services {
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
@@ -19,7 +19,7 @@ import com.daml.platform.ApiOffset.ApiOffsetConverter
 import com.daml.platform.common
 import com.daml.platform.common.MismatchException
 import com.daml.platform.configuration.ServerRole
-import com.daml.platform.indexer.parallel.ParallelIndexerFactory
+import com.daml.platform.indexer.parallel.{ParallelIndexerFactory, PostgresStorageBackend}
 import com.daml.platform.store.appendonlydao.events.{CompressionStrategy, LfValueTranslation}
 import com.daml.platform.store.dao.{JdbcLedgerDao, LedgerDao}
 import com.daml.platform.store.{DbType, FlywayMigrations, LfValueTranslationCache}
@@ -138,6 +138,7 @@ object JdbcIndexer {
         )
         indexer <- ParallelIndexerFactory(
           jdbcUrl = config.jdbcUrl,
+          storageBackend = PostgresStorageBackend, // TODO append-only: factory mechanism
           participantId = config.participantId,
           translation = new LfValueTranslation(
             cache = lfValueTranslationCache,
@@ -150,6 +151,7 @@ object JdbcIndexer {
             else CompressionStrategy.none(metrics),
           mat = materializer,
           inputMappingParallelism = config.inputMappingParallelism,
+          batchingParallelism = config.batchingParallelism,
           ingestionParallelism = config.ingestionParallelism,
           submissionBatchSize = config.submissionBatchSize,
           tailingRateLimitPerSecond = config.tailingRateLimitPerSecond,
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
@@ -3,6 +3,7 @@
 
 package com.daml.platform.indexer.parallel
 
+import java.sql.Connection
 import java.util.concurrent.TimeUnit
 
 import akka.NotUsed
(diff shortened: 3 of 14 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -3,109 +3,92 @@
...
+        Some(metrics.daml.parallelIndexer.batching.executor -> metrics.registry),
+      )
+      dbDispatcher <- DbDispatcher.owner( // TODO do we need to wire health status here somehow?
+        serverRole = ServerRole.Indexer,
+        jdbcUrl = jdbcUrl,
+        connectionPoolSize = ingestionParallelism + 1, // + 1 for the tailing ledger_end updates
+        connectionTimeout = FiniteDuration(
+          250,
+          "millis",
+        ), // TODO verify why would we need here a timeout, in this case (amount of parallelism aligned) it should not happen by design
```
mziolekda, line 63: Canton fork uses this instead:
```
30.seconds, // overridden in canton back to hikari default as the upstream 250ms setting is too low
```

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PostgresStorageBackend.scala
```diff
@@ -745,48 +733,50 @@ case class JDBCPostgresDAO(jdbcUrl: String) extends PostgresDAO with AutoCloseab
...
-
-    result
+    StorageBackend.Initialized(
+      lastOffset = offset,
+      lastEventSeqId = eventSeqId,
+    )
   }
 
-  private def queryLedgerAndAndEventSeqId(): (Option[Offset], Option[Long]) = {
+  private def queryLedgerAndAndEventSeqId(
```
mziolekda, line 777: Method name not new, but clearly wrong. Should be `queryLedgerEndAndEventSeqId`

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PostgresStorageBackend.scala
```diff
@@ -3,38 +3,18 @@
...
-    stmnt.execute("SET synchronous_commit TO OFF;")
-    c.commit()
-    stmnt.close()
-
-    c
-  }
-
-  private val preparedInsertEventsBatchDivulgence = connection.prepareStatement(
-    """
+  private val preparedInsertEventsBatchDivulgence: Connection => PreparedStatement =
```
mziolekda, line 15: I take your word for it that calling it repeatedly performs just as well as calling it once per session.

mziolekda (approved):
The changes seem uncontroversial, mostly code shuffling. I would only opt for changing the name of `RunningBatch` to just `Batch`

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: Which might be relevant for a high load scenario on the ReadDao side, but here currently I do not see any reason to have this not 0: the pool is exactly the same size how many users it has at the same time. (AFAIK this config boils down to Hikari pool config which defines timeout for getting the connection for a client)
I would keep the TODO, let's start digging into Hikari internals in epic #3

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/PostgresStorageBackend.scala
nmarton-da: well it will be slower per se, but let's hope it is not that much slower :crossed_fingers: 
we need the flow level benchmarks :pray:

ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
```diff
@@ -138,6 +138,7 @@ object JdbcIndexer {
         )
         indexer <- ParallelIndexerFactory(
           jdbcUrl = config.jdbcUrl,
+          storageBackend = PostgresStorageBackend, // TODO factory mechanism
```
rautenrieth-da, line 141:
```suggestion
          storageBackend = PostgresStorageBackend, // TODO append-only: factory mechanism
```

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -573,18 +575,30 @@ final class Metrics(val registry: MetricRegistry) {
...
+      // Parallel ingestion of prepared data into the database
+      object ingestion extends DatabaseMetrics(registry, Prefix, "ingestion") {
+        private val Prefix: MetricName = parallelIndexer.Prefix :+ "ingestion"
+
         // The latency, which during an update element is residing in the ingestion.
         // Since batches are involved, this duration is divided by the batch size.
         val duration: Timer = registry.timer(Prefix :+ "duration")
       }
+
+      val tailIngestion = new DatabaseMetrics(registry, Prefix, "tail_ingestion")
```
rautenrieth-da, line 601:
```suggestion
      // Tail ingestion stage
      // The throttled update of ledger end parameters
      val tailIngestion = new DatabaseMetrics(registry, Prefix, "tail_ingestion")
```

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -3,109 +3,92 @@
...
       )
-      postgresDaoPool <- asyncResourcePool(
-        () => JDBCPostgresDAO(jdbcUrl),
-        size = ingestionParallelism + 1,
-        Some(metrics.daml.parallelIndexer.ingestion.executor -> metrics.registry),
+      batcherExecutor <- asyncPool(
+        batchingParallelism,
+        Some(metrics.daml.parallelIndexer.batching.executor -> metrics.registry),
+      )
+      dbDispatcher <- DbDispatcher.owner( // TODO do we need to wire health status here somehow?
```
rautenrieth-da, line 56:
```suggestion
      dbDispatcher <- DbDispatcher.owner( // TODO append-only: do we need to wire health status here somehow?
```

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
rautenrieth-da, line 63:
```suggestion
        ), // TODO append-only: verify why would we need here a timeout, in this case (amount of parallelism aligned) it should not happen by design
```

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -115,16 +98,160 @@ object ParallelIndexerFactory {
...
       def subscribe(readService: ReadService): Future[Source[Unit, NotUsed]] =
-        postgresDaoPool
-          .execute(_.initialize)
-          .map { case (offset, eventSeqId) =>
-            ingest(eventSeqId.getOrElse(0L))(readService.stateUpdates(beginAfter = offset))
-          }(scala.concurrent.ExecutionContext.global)
+        dbDispatcher
+          .executeSql(metrics.daml.parallelIndexer.initialization)(
+            storageBackend.initialize
+          ) // TODO here is an implicit LoggingContext injected from the factory apply implicit param, is that good this way? shouldn't we create here something for the purpose?
```
rautenrieth-da, line 104: I don't understand this TODO. Do you see an issue with the logging context from `ParallelIndexerFactory.apply` being passed to `dbDispatcher.executeSql`?

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -115,16 +98,160 @@ object ParallelIndexerFactory {
...
+            ingest(initialized.lastEventSeqId.getOrElse(0L))(
+              readService.stateUpdates(beginAfter = initialized.lastOffset)
+            )
+          )(scala.concurrent.ExecutionContext.global)
 
       toIndexer(subscribe)(mat)
     }
   }
 
+  case class RunningBatch[T](
```
rautenrieth-da, line 115: It's more a "opaque batch with some metadata about the indexing pipeline" than `RunningBatch`, but I would not worry too much about the naming.

rautenrieth-da, line 115: This class has many properties relevant to different stages of the indexing pipeline. Not all properties are used in each stage, and it's not obvious what they mean. I think this class needs a good description of all members.

For example, what exactly is `lastSeqEventId`? It is assigned by the seqMapper, and seems to store the "maximum of the sequential event ids within this batch". Then the tailer takes the `lastSeqEventId` value of the last encountered batch, but that doesn't mean all sequential ids up to that point were written, right? The batches will not arrive at the tailer in the order of increasing sequential event ids.

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -115,16 +98,160 @@ object ParallelIndexerFactory {
...
+      averageStartTime = nowNanos,
+    )
+  }
+
+  def ingester[DB_BATCH](
+      ingestFunction: (Connection, DB_BATCH) => Unit,
+      dbDispatcher: DbDispatcher,
+      metrics: Metrics,
+  )(implicit
+      loggingContext: LoggingContext // TODO is that right so? shouldn't we prep something here?
```
rautenrieth-da, line 207: What would you like to prep? Add some info to the logging context, like the batch size?

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -115,16 +98,160 @@ object ParallelIndexerFactory {
...
+          TimeUnit.NANOSECONDS,
+        )
+        batch
+      }
+
+  def tailer[DB_BATCH]: (RunningBatch[DB_BATCH], RunningBatch[DB_BATCH]) => RunningBatch[DB_BATCH] =
+    (prev, curr) =>
+      RunningBatch[DB_BATCH](
+        lastOffset = curr.lastOffset,
+        lastSeqEventId = curr.lastSeqEventId,
```
rautenrieth-da, line 225: As noted above, is this correct? Can't `prev.lastSeqEventId` be greater than `curr.lastSeqEventId`?

rautenrieth-da (changes requested):
I like the direction, nice work 👍 

The only comment I would like to have addressed is the one regarding last event sequential IDs.

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: well, not, I am happy to remove this entirely if there is none. I was just concerned because executeSql needs this implicit param, so in theory we could do something like setting logging params in the context regarding whatnot before we hit the ingestion

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: @mziolekda suggested simply batch...I am ok with that as well until we find something better. So I am planning to refactor towards Batch for now, but as soons as you would have an enlightening moment of a good name for this please come forward :)

nmarton-da: good point, will add descriptions

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: I have no clue :) essentially this is the same what I was explaining above.

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
rautenrieth-da: I would only add values to the logging context if we think they might be useful for troubleshooting issues later. `batch.lastSeqEventId` is the closest to a unique ID of the current batch, which might be helpful for correlating log lines. You could consider adding the following line, but I don't think it's very important:

```
withEnrichedLoggingContext("lastSeqEventId" -> batch.lastSeqEventId) { implicit loggingContext =>
```

nmarton-da: What I had in mind, whether we already had something enriched already, which we were relying on...
The lastqlEventId can be actually the same for many batches, this is only for events. Maybe use the hex-offset?
I suggest turning this around: let's have needs from logging and then implement it here.
On log correlation: that is usually quite hard, due to batching. For example to track how a submission is travelling trough a system: for that we would need to do is to be able to correlate batches to all submissions which they relate to :thinking:

PR #9523, opened 2021-04-28
Move builtin exceptions to a stable package.

sofiafaro-da:
This PR moves the built-in exceptions (GeneralError, ArithmeticError, ContractError) to a stable package in daml-prim, in preparation for removing them from LF proper.

Part of #8020.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
+++ b/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
@@ -585,7 +585,7 @@ generateStablePackages lfVersion fp = do
         let dalfs = map (fp </>) $ concat
                 [ map ("daml-prim" </>)
                     [ "DA-Internal-Erased.dalf"
-                    , "DA-Internal-Error.dalf"
+                    , "DA-Internal-Exception-Types.dalf"
                     , "DA-Internal-PromotedText.dalf"
                     , "DA-Types.dalf"
                     , "GHC-Prim.dalf"
--- a/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
+++ b/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
@@ -585,6 +585,7 @@ generateStablePackages lfVersion fp = do
         let dalfs = map (fp </>) $ concat
                 [ map ("daml-prim" </>)
                     [ "DA-Internal-Erased.dalf"
+                    , "DA-Internal-Error.dalf"
                     , "DA-Internal-PromotedText.dalf"
                     , "DA-Types.dalf"
                     , "GHC-Prim.dalf"
--- a/compiler/damlc/daml-prim-src/DA/Internal/Error.daml
+++ b/compiler/damlc/daml-prim-src/DA/Internal/Error.daml
@@ -1,0 +1,30 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+#ifndef DAML_EXCEPTIONS
+
+-- | HIDE
+module DA.Internal.Error where
+
+import GHC.Types ()
+
+#else
+
+-- | MOVE DA.Exception
+module DA.Internal.Error where
+
+import GHC.Types (Text)
+
+-- | Exception raised by `error`.
+data GeneralError = GeneralError { message : Text }
+
+-- | Exception raised by an arithmetic operation, such as divide-by-zero or overflow.
+data ArithmeticError = ArithmeticError { message : Text }
+
+-- | Exception raised when a contract is invalid, i.e. fails the ensure clause.
+data ContractError = ContractError { message : Text }
+
+#endif
--- a/compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
@@ -21,8 +21,8 @@ module DA.Internal.Exception
     ) where
 
 import GHC.Types
+import DA.Internal.Exception.Types
 import DA.Internal.LF
-import DA.Internal.Error
 import DA.Internal.Prelude
 import DA.Internal.Record
 
@@ -94,6 +94,8 @@ instance HasToAnyException GeneralError where
 
 instance HasFromAnyException GeneralError where
     fromAnyException = handleBuiltinException "GeneralError" GeneralError (primitive @"EFromAnyException")
+        -- TODO https://github.com/digital-asset/daml/issues/8020
+        --   Remove the "handleBuiltinException" part if we decide we never want to throw a GeneralError in LF.
 
 --------------------------------------------------------------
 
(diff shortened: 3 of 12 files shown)


compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
```diff
@@ -585,6 +585,7 @@ generateStablePackages lfVersion fp = do
         let dalfs = map (fp </>) $ concat
                 [ map ("daml-prim" </>)
                     [ "DA-Internal-Erased.dalf"
+                    , "DA-Internal-Error.dalf"
```
cocreature, line 588: maybe worth considering moving this from `Error` to `Exception`. It seems a bit odd to have builtin exceptions defined in an `Error` module. But we can also postpone that until we finalize builtin exceptions.

compiler/damlc/daml-prim-src/DA/Internal/Error.daml
```diff
@@ -0,0 +1,30 @@
...
+import GHC.Types (Text)
+
+-- | Exception raised by `error`.
+data GeneralError = GeneralError { message : Text }
+
+-- | Exception raised by an arithmetic operation, such as divide-by-zero or overflow.
+data ArithmeticError = ArithmeticError { message : Text }
+
+-- | Exception raised when a contract is invalid, i.e. fails the ensure clause.
+data ContractError = ContractError { message : Text }
```
cocreature, line 28: maybe worth adding a todo to remove the message fields once we drop it in Daml-LF.

compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
```diff
@@ -137,4 +151,8 @@ instance ActionCatch Update where
 
 --------------------------------------------------------------
 
+-- | HIDE
+handleBuiltinException : Text -> (Text -> e) -> (AnyException -> Optional e) -> (AnyException -> Optional e)
+handleBuiltinException _ _ f = f -- magic @"handleBuiltinException"
```
cocreature, line 156: what’s up with the comment?

compiler/damlc/daml-prim-src/DA/Internal/Error.daml
cocreature: Actually you already implement it based on the `Text`-based stuff so I don’t understand why you need this.

compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
```diff
@@ -77,45 +79,57 @@ instance HasFromAnyException AnyException where
...
 instance HasMessage GeneralError where
-    message = primitive @"BEGeneralErrorMessage"
+    message (GeneralError m) = m
 
 instance HasToAnyException GeneralError where
     toAnyException = primitive @"EToAnyException"
 
 instance HasFromAnyException GeneralError where
-    fromAnyException = primitive @"EFromAnyException"
+    fromAnyException = handleBuiltinException "GeneralError" GeneralError (primitive @"EFromAnyException")
```
cocreature, line 96: Will we ever throw this? I thought we agreed that `error` is just going to bypass the LF primitive and throw the LF stdlib primitive.

compiler/damlc/stable-packages/src/GenerateStablePackage.hs
```diff
@@ -581,6 +583,54 @@ daInternalPromotedText = package version1_6 $ NM.singleton Module
       [ DefDataType Nothing ptextTyCon (IsSerializable False) [(mkTypeVar "t", KStar)] $ DataVariant []
       ]
 
+daInternalError :: Package
+daInternalError = Package
+    { packageLfVersion = versionDev
```
cocreature, line 588: can we use `featureMinVersion featureExceptions` instead?

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -50,7 +50,8 @@ darPackageIds fp = do
     Right dalfPkgIds  <- pure $ mapM (LFArchive.decodeArchivePackageId . BSL.toStrict) $ mainDalf : dalfDeps
     pure dalfPkgIds
 
-
+-- TODO https://github.com/digital-asset/daml/issues/8020
```
cocreature, line 53: hopefully the test should tell us that by failing but I guess the todo doesn’t do any harm.

cocreature (approved):
Thanks!

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
sofiafaro-da: We already have a `DA.Internal.Exception`. Maybe some other name? :-)

cocreature: `DA.Exception.Types` similar to the other `.Types` modules?

compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
sofiafaro-da: It's a placeholder for adding some rules in LFConversion for dealing with builtin exceptions (rather than handling them in daml).

cocreature: maybe add a comment?

compiler/damlc/daml-prim-src/DA/Internal/Error.daml
sofiafaro-da: I don't think we need to drop this field. We definitely want them for GeneralError, and it's simpler to have all these built-in execptions be the same than to have them be different. But we can also, I think, still support text messages in built-in exceptions, without having built-in exception types, since these can all carried by AnyException anyway. I think Rémy had said something like this at some point ... (Maybe we can talk about it next meeting.)

compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
sofiafaro-da: This is for catching, not throwing.

compiler/damlc/stable-packages/src/GenerateStablePackage.hs
sofiafaro-da: Ah, yeah :)

compiler/damlc/daml-stdlib-src/DA/Internal/Exception.daml
cocreature: Right but if we throw `GeneralError` directly in Daml-Stdlib and LF never throws this then we don’t need to go via `Text` here.

sofiafaro-da: Ah, true! I guess the question is do we ever want to throw a GeneralError in LF?

cocreature: my understanding is we don’t but we don’t have to solve this in this PR. Maybe just add a todo.

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
sofiafaro-da: I think we should probably refactor all of this (in a separate PR).
"numStablePackages" should be derived from a list of stable packages.

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
sofiafaro-da: `DA.Internal.Exception.Types` then :-)

cocreature: Is it internal though? Those types seem like they should be public API and they are by being serializable.

PR #9530, opened 2021-04-28
http-json-perf: querying larger ACSes with different data patterns

### LINKED ISSUE #9460: extend json-api perf tests to Oracle, and test 100k-contract ACS
As #7434 extended http-json-perf to PostgreSQL, we can do the same for Oracle, albeit in an EE-only context.

Additionally, we would like to run a matrix of perf-testing queries against a 1-million-contract ACS. We aren't too interested in the in-memory performance (we know it is linear, guaranteed), but observing performance characteristics of both PostgreSQL and Oracle under different query distributions would be valuable.

Some dimensions we might be interested in:

1. Low-to-high frequency of party matches (due to % presence in signatories/observers);
2. low-to-high frequency of party matches (due to querying by smaller-to-larger party-sets);
3. " of template ID matches;
4. " of payload value matches.

To explain the difference between (1) and (2): suppose parties {foo, bar, baz}. To alter the frequency for (1) we fix the party-set as {foo} and change the % of contracts that have the sole party foo in signatories/observers. To alter the frequency for (2) we fix the respective % of contracts as {10%, 40%, 50%} and can then reach 10, 50, 100% frequency by querying with party-sets {foo}, {foo, bar}, and {foo, bar, baz} respectively.

S11001001:
Advances #9460 with a 100k-contract ACS with 1% party presence frequency.

On local Oracle as of fb6f72c81d983eebe91b87f491238e4d414573c2 (no party indexing whatsoever) this takes 13s to copy whole ACS, 418ms/query. (99% takes 594ms, demonstrating the extra time for checking the payload over the extra 98000 contracts.)

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):

cocreature (approved):
Nice, thank you!

PR #9536, opened 2021-04-29
Fix TransactionCommitter.trimUnnecessaryNodes for rollback nodes

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
@@ -36,6 +36,7 @@ import com.daml.logging.{ContextualizedLogger, LoggingContext}
 import com.daml.metrics.Metrics
 import com.google.protobuf.{Timestamp => ProtoTimestamp}
 
+import scala.annotation.tailrec
 import scala.jdk.CollectionConverters._
 
 // The parameter inStaticTimeMode indicates that the ledger is running in static time mode.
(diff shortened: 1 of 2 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -387,9 +388,27 @@ private[kvutils] class TransactionCommitter(
...
-      }.toSet
+      val nodeMap: Map[String, TransactionOuterClass.Node] =
+        nodes.view.map(n => n.getNodeId -> n).toMap
+
+      @tailrec
+      def goNodesToKeep(todo: List[String], result: Set[String]): Set[String] = todo match {
+        case Nil => result
+        case head :: tail =>
+          import TransactionOuterClass.Node.NodeTypeCase
+          val node = nodeMap.get(head).getOrElse(throw Err.InvalidSubmission(s"Invalid transaction node id $head"))
```
cocreature, line 399: If there is a more appropriate error I should throw here let me know. Encoding/decoding didn’t really seem appropriate. InternalError also would work I guess.

SamirTalwar-DA (approved):
I approve, but I'm only somewhat qualified to do so.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -122,13 +129,16 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+    TransactionOuterClass.NodeRollback.newBuilder()
+
   private def createNodeBuilder = TransactionOuterClass.NodeCreate.newBuilder()
 
   private def lookupByKeyNodeBuilder =
     TransactionOuterClass.NodeLookupByKey.newBuilder()
 
   "trimUnnecessaryNodes" should {
-    "remove `Fetch` and `LookupByKey` nodes from transaction tree" in {
+    "remove `Fetch` `LookupByKey` and `Rollback` nodes from transaction tree" in {
```
SamirTalwar-DA, line 141:
```suggestion
    "remove `Fetch`, `LookupByKey`, and `Rollback` nodes from the transaction tree" in {
```

fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -387,9 +388,29 @@ private[kvutils] class TransactionCommitter(
...
+        nodes.view.map(n => n.getNodeId -> n).toMap
+
+      @tailrec
+      def goNodesToKeep(todo: List[String], result: Set[String]): Set[String] = todo match {
+        case Nil => result
+        case head :: tail =>
+          import TransactionOuterClass.Node.NodeTypeCase
+          val node = nodeMap
+            .get(head)
+            .getOrElse(throw Err.InvalidSubmission(s"Invalid transaction node id $head"))
```
fabiotudone-da, line 401: Can this really happen because of an invalid submission? It seems a rather impossible case to me as we build this map with the key being a piece of the node; we could perhaps have clashes if things are really broken. Perhaps I'd either throw an internal error or even let Scala throw if the element is absent.

cocreature: I don’t really know what validation happens up to this point so I don’t know for sure but I expect you’re right. I’ve switched it to `InternalError`.

fabiotudone-da: Thanks, this also increases our incentive to look into these cases, as we'll tend to assume it's a bug.

PR #9537, opened 2021-04-29
Share code between SBUFetchKey and SBULookupKey

cocreature:
Those do almost the same and the logic is growing increasingly complex
so I really want to share it. This also makes it easier to see the
things that are different.

Along the way I’ve also discovered a bug:
SBUFetchKey called crash where it should produce a proper error that
can be caught by submitMustFail. Not going to resolve that in this PR
though so I just added a TODO.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1174,7 +1174,7 @@ private[lf] object SBuiltin {
         args: util.ArrayList[SValue],
         machine: Machine,
         onLedger: OnLedger,
-    ): Unit = lookupKey(args, machine, onLedger)
+    ): Unit = handleKey(args, machine, onLedger, Lookup)
   }
 
   /** $insertLookup[T]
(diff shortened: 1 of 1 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1267,25 +1213,69 @@ private[lf] object SBuiltin {
...
+      override def handleKeyArchived(machine: Machine, key: GlobalKey): Unit = {
+        machine.returnValue = SV.None
+      }
+      override def cidToSValue(cid: V.ContractId) =
+        SOptional(Some(SContractId(cid)))
+      override def cidToSExpr(cid: V.ContractId) =
+        SBSome(SEImportValue(typ, V.ValueContractId(cid)))
+    }
+
+    protected final def lookupKey(
```
nickchapman-da, line 1253: suggest inline `lookupKey` and `fetchKey` into their sole respective callers, which can call `handleKey` directly.
(`handleKey` will need to become `protected final` instead of `private`.)

nickchapman-da (approved):
This is a really nice cleanup. Thanks.
I have just one small suggestion.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1267,25 +1213,55 @@ private[lf] object SBuiltin {
...
-    *   :: { key: key, maintainers: List Party }
-    *   -> ContractId T
-    */
-  final case class SBUFetchKey(templateId: TypeConName) extends OnLedgerBuiltin(1) {
+  private[speedy] sealed abstract class SBUKeyBuiltin(templateId: TypeConName)
+      extends OnLedgerBuiltin(1) {
     private[this] val typ = AstUtil.TContractId(Ast.TTyCon(templateId))
-    override protected final def execute(
+
+    sealed trait KeyOperation {
```
remyhaemmerle-da, line 1220:
```suggestion
    abstract class KeyOperation
```

remyhaemmerle-da (commented):
LGTM

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: Why is that better?

PR #9538, opened 2021-04-29
Cache global key inputs

cocreature:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1289,7 +1289,7 @@ private[lf] object SBuiltin {
           // Check if we have a cached global key result.
           onLedger.ptx.globalKeyInputs.get(gkey) match {
             case Some(optCid) =>
-              onLedger.ptx = onLedger.ptx.copy(keys = onLedger.ptx.keys + (gkey -> optCid))
+              onLedger.ptx = onLedger.ptx.copy(keys = onLedger.ptx.keys.updated(gkey, optCid))
               optCid match {
                 case Some(cid) =>
                   machine.returnValue = operation.cidToSValue(cid)
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1256,13 +1256,14 @@ private[lf] object SBuiltin {
         onLedger: OnLedger,
         operation: KeyOperation,
     ): Unit = {
+      import PartialTransaction.{KeyActive, KeyInactive}
       val keyWithMaintainers = extractKeyWithMaintainers(args.get(0))
       if (keyWithMaintainers.maintainers.isEmpty)
         throw DamlEFetchEmptyContractKeyMaintainers(templateId, keyWithMaintainers.key)
       val gkey = GlobalKey(templateId, keyWithMaintainers.key)
       // check if we find it locally
       onLedger.ptx.keys.get(gkey) match {
-        case Some(Some(coid)) if onLedger.ptx.localContracts.contains(coid) =>
+        case Some(KeyActive(coid)) if onLedger.ptx.localContracts.contains(coid) =>
           val cachedContract = onLedger.cachedContracts
             .get(coid)
             .getOrElse(crash(s"Local contract $coid not in cachedContracts"))
@@ -1281,9 +1282,9 @@ private[lf] object SBuiltin {
               },
             )
           )
-        case Some(Some(coid)) =>
+        case Some(KeyActive(coid)) =>
           machine.returnValue = operation.cidToSValue(coid)
-        case Some(None) =>
+        case Some(KeyInactive) =>
           operation.handleKeyArchived(machine, gkey)
         case None =>
           // Check if we have a cached global key result.
@@ -1291,9 +1292,9 @@ private[lf] object SBuiltin {
             case Some(optCid) =>
               onLedger.ptx = onLedger.ptx.copy(keys = onLedger.ptx.keys.updated(gkey, optCid))
               optCid match {
-                case Some(cid) =>
+                case KeyActive(cid) =>
                   machine.returnValue = operation.cidToSValue(cid)
-                case None =>
+                case KeyInactive =>
                   operation.handleKeyArchived(machine, gkey)
               }
             case None =>
@@ -1306,8 +1307,8 @@ private[lf] object SBuiltin {
                   {
                     case SKeyLookupResult.Found(cid) =>
                       onLedger.ptx = onLedger.ptx.copy(
-                        keys = onLedger.ptx.keys + (gkey -> Some(cid)),
-                        globalKeyInputs = onLedger.ptx.globalKeyInputs + (gkey -> Some(cid)),
+                        keys = onLedger.ptx.keys + (gkey -> KeyActive(cid)),
+                        globalKeyInputs = onLedger.ptx.globalKeyInputs + (gkey -> KeyActive(cid)),
                       )
                       // We have to check that the discriminator of cid does not conflict with a local ones
                       // however we cannot raise an exception in case of failure here.
@@ -1316,8 +1317,8 @@ private[lf] object SBuiltin {
                       true
                     case SKeyLookupResult.NotFound =>
                       onLedger.ptx = onLedger.ptx.copy(
-                        keys = onLedger.ptx.keys + (gkey -> None),
-                        globalKeyInputs = onLedger.ptx.globalKeyInputs + (gkey -> None),
+                        keys = onLedger.ptx.keys + (gkey -> KeyInactive),
+                        globalKeyInputs = onLedger.ptx.globalKeyInputs + (gkey -> KeyInactive),
                       )
                       operation.handleKeyNotFound(machine)
                     case SKeyLookupResult.NotVisible =>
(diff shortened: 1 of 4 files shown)


remyhaemmerle-da:
Could you rebase the PR ?

cocreature:
done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -338,7 +351,26 @@ private[lf] case class PartialTransaction(
...
-          Right((cid, ptx.copy(keys = ptx.keys.updated(ck, Some(cid)))))
+          val globalKeyInputs = keys.get(ck) match {
+            // We saw an archive or negative lookup, no constraint on global key inputs.
+            case Some(None) => ptx.globalKeyInputs
+            // We saw a create or successful lookup, no constraint
+            // on global key inputs.
+            // TODO (MK) In uck-mode this should be an error if there is no archive in between.
+            case Some(Some(_)) => ptx.globalKeyInputs
+            // No entry or archive => key must not be active
+            // at the beginning of the transaction.
```
remyhaemmerle-da, line 362: Why not factorizing the two cases ?

remyhaemmerle-da (commented):
Let me check if I understand properly. 

* `PartialTransaction#keys` will evolve non-monotonically during interpretation (i.e. it can be restore to a previous version when rolling back a part of the transaction)
*  `PartialTransaction#keys` will evolve monotonically keeping track of all query to the ledger we already did. (+ the extra special case to avoid querying archived transient contract).

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: Mainly because they’re going to be different when we add uniqueness checks and I think it’s useful to already keep them separate for documentation purposes.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1286,28 +1286,46 @@ private[lf] object SBuiltin {
...
-                  operation.handleKeyNotFound(machine)
-                case SKeyLookupResult.NotVisible =>
-                  machine.tryHandleSubmitMustFail()
-              },
-            )
-          )
+          // Check if we have a cached global key result.
+          onLedger.ptx.globalKeyInputs.get(gkey) match {
+            case Some(optCid) =>
+              onLedger.ptx = onLedger.ptx.copy(keys = onLedger.ptx.keys + (gkey -> optCid))
```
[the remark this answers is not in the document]
cocreature: Is there a difference between the two or do you just prefer the non-infix version?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: :+1:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1286,28 +1286,46 @@ private[lf] object SBuiltin {
...
+                  onLedger.committers,
+                  {
+                    case SKeyLookupResult.Found(cid) =>
+                      onLedger.ptx = onLedger.ptx.copy(
+                        keys = onLedger.ptx.keys + (gkey -> Some(cid)),
+                        globalKeyInputs = onLedger.ptx.globalKeyInputs + (gkey -> Some(cid)),
+                      )
+                      // We have to check that the discriminator of cid does not conflict with a local ones
+                      // however we cannot raise an exception in case of failure here.
+                      // We delegate to CtrlImportValue the task to check cid.
```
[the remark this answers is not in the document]
cocreature: How so? We still need to do that check and afaict we still do it in `CtrlImportValue`.

cocreature:
> Let me check if I understand properly.
> 
>     * `PartialTransaction#keys` will evolve non-monotonically during interpretation (i.e. it can be restore to a previous version when rolling back a part of the transaction)
> 
>     * `PartialTransaction#keys` will evolve monotonically keeping track of all query to the ledger we already did. (+ the extra special case to avoid querying archived transient contract).

Exactly, my new mantra is "rollbacks rollback reads not writes". That makes it pretty clear that global key reads should not be rolled back.

keys on the other hand tracks writes and reads. We could add some complicated logic to only rollback the "write" part of `keys` but I think that is much more confusing than just rolling back everything and adding a cache for the keys which is equivalent to only rolling back writes.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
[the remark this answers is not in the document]
remyhaemmerle-da: Not really your fault, but `CtrlImportValue` does not exist anymore.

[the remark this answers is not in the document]
cocreature: oh I see, renamed to `SEImportValue`

remyhaemmerle-da:
> Exactly, my new mantra is "rollbacks rollback reads not writes". 
Do you mean "rollbacks rollback writes not reads" ?

cocreature:
>  Do you mean "rollbacks rollback writes not reads" ?

duh of course :facepalm:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -212,11 +217,19 @@ private[lf] case class PartialTransaction(
     context: PartialTransaction.Context,
     aborted: Option[Tx.TransactionError],
     keys: Map[GlobalKey, Option[Value.ContractId]],
+    globalKeyInputs: Map[GlobalKey, Option[Value.ContractId]],
```
nickchapman-da, line 220: _keys on the other hand tracks writes and reads. We could add some complicated logic to only rollback the "write" part of keys but I think that is much more confusing than just rolling back everything and adding a cache for the keys which is equivalent to only rolling back writes._


Is there an abstraction which would capture the relationship between these two elements of the state: ?
```
    keys: Map[GlobalKey, Option[Value.ContractId]],
    globalKeyInputs: Map[GlobalKey, Option[Value.ContractId]],
```
So it's easier to be sure that they are being properly kept in sync.

And perhaps a more evocative name for 'keys' which makes it clearer the relationship it has with `globalkeyinputs`?

(Maybe this is a refactoring for a followup PR.)

cocreature: Not really sure what a good abstraction would be but if you have any ideas I’m happy to change it although preferably in a separate PR.

nickchapman-da: The code might be clearer if we use a custom scala type instead of `Option[Value.ContractId]`.
So we don't just have `None` to record the semantic info that a negative lookup has occurred.

Also the match statements we currently have like `case Some(Some(_)) =>` when we lookup in these maps would  be easier to read.

cocreature: good idea 3abd143265c527ad53658f075e496ca21f230dfa

nickchapman-da: For a given key, I'm trying to understand which of the 9 states is possible, and what it means.
So far I have... Is this correct?

```
keys        globalKeyInputs

no-entry    no-entry    -- key never seem
no-entry    None        -- negative lookup; rolled-back
no-entry    Some(_)     -- positive lookup; rolled-back

None        no-entry    -- IMPOSSIBLE
None        None        -- negative lookup
None        Some(_)     -- ??

Some(_)     no-entry    -- IMPOSSIBLE
Some(_)     None        -- ??
Some(c1     Some(c2)    -- positive lookup -- (c1==c2)?
```

nickchapman-da: (sorry, I see you already made a better type. But the question still stands)

remyhaemmerle-da (approved):
LGTM

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: keys        globalKeyInputs

no-entry    no-entry    -- key never seen
no-entry    None        -- negative lookup; rolled-back -- EDIT: this could also be rolled back
no-entry    Some(_)     -- positive lookup; rolled-back

None        no-entry    -- IMPOSSIBLE -- Edit: Not quite true, just a global archive of a key that has not been looked up by key before will produce this. This is a bug and there is a TODO in the tests. https://github.com/digital-asset/daml/pull/9546 fixes this
None        None        -- negative lookup -- EDIT: this can also be a create & archive in the same transaction
None        Some(_)     -- ?? -- EDIT: lookup of a global contract & archive

Some(_)     no-entry    -- IMPOSSIBLE
Some(_)     None        -- ?? -- EDIT: just a create or negative lookup & then a create
Some(c1)    Some(c2)    -- positive lookup -- (c1==c2)? -- EDIT: c1==c2 iff we did not see an archive of c2

nickchapman-da (approved):
Thanks for the changes. LGTM

PR #9539, opened 2021-04-29
Minor refactoring

nmarton-da:
* Add TODO comments to guide further work
* Add and wire up StorageBackend factory

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDBDTOV1.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDBDTOV1.scala
@@ -139,6 +139,11 @@ object UpdateToDBDTOV1 {
         )
 
       case u: Update.TransactionAccepted =>
+        // TODO add support for Daml Exceptions / Rollback nodes
+        //   Covering this functionality with unit test is important, since at the time of writing kvutils ledgers purge RollBack nodes already on WriteService, so conformance testing is impossible
+        //   Unit tests also need to cover the full semantic contract regarding fetch and lookup node removal as well
+        //   Hint for implementation: https://github.com/digital-asset/daml/pull/9506
+        //   Hint for conformance testing: if sandbox-classic append only integration uses this code, then ExceptionIT of the respective conformance test suit (append only + sandbox-classic) will cover the indexDB population cases.
         val blinding = u.blindingInfo.getOrElse(Blinding.blind(u.transaction))
         val preorderTraversal = u.transaction
           .fold(List.empty[(NodeId, Node)]) { case (xs, x) =>
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
@@ -21,7 +21,7 @@ import com.daml.platform.common.MismatchException
 import com.daml.platform.configuration.ServerRole
 import com.daml.platform.indexer.parallel.ParallelIndexerFactory
 import com.daml.platform.store.appendonlydao.events.{CompressionStrategy, LfValueTranslation}
-import com.daml.platform.store.backend.postgresql.PostgresStorageBackend
+import com.daml.platform.store.backend.StorageBackend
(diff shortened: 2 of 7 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDBDTOV1.scala
```diff
@@ -139,6 +139,11 @@ object UpdateToDBDTOV1 {
         )
 
       case u: Update.TransactionAccepted =>
+        // TODO add support for Daml Exceptions / Rollback nodes
+        //   Covering this functionality with unit test is important, since at the time of writing kvutils ledgers purge RollBack nodes already on WriteService, so conformance testing is impossible
```
rautenrieth-da, line 143: > kvutils ledgers purge RollBack nodes already on WriteService

Out of interest, how does the purging work? Reads below rollback nodes can lead to divulgence (if I understand @moritzkiefer-da correctly [here](https://github.com/digital-asset/daml/pull/9486#discussion_r620353948)), you can't just remove rollback nodes from a transaction.

If the kvutils purging operation removes create/exercise events below rollback nodes but preserves divulgence (or more generally, behaves correctly), can we apply the same operation to any incoming transaction? 

> conformance testing is impossible

I assume this means "conformance testing is impossible on kvutils ledgers in their current form", because below you note that sandbox-classic on append-only will be able to use conformance tests added in https://github.com/digital-asset/daml/pull/9400?

ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDBDTOV1.scala
```diff
@@ -139,6 +139,11 @@ object UpdateToDBDTOV1 {
         )
 
       case u: Update.TransactionAccepted =>
+        // TODO add support for Daml Exceptions / Rollback nodes
```
rautenrieth-da, line 142:
```suggestion
        // TODO append-only: add support for Daml Exceptions / Rollback nodes
```

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDBDTOV1.scala
cocreature: This is not actually specific to rollback nodes. KVUtils also drops fetch & lookup by key nodes which have the same issue (meaning they also have the same testing issue). The logic for this is in https://github.com/digital-asset/daml/blob/c85da6def052fa3d5a817789b66b3ebe3815c9e1/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala#L383.

The reason why this doesn’t cause issues around divulgence is that KVUtils calculates blinding info before trimming nodes and then passes it along with the transaction so while the transaction has lost the information it is still there and passed into the indexer somehow (not familiar with the details there myself).

As for sandbox classic, you are completely correct. That will result in this code path being exercised.

rautenrieth-da: > KVUtils calculates blinding info before trimming nodes and then passes it along with the transaction

I was originally thinking whether we can extract this transformation (something like `TransactionWithAllKindsOfNodes => (TransactionWithWriteNodesOnly, BlindingInfo)`?) from `kvutils` into `com.daml.lf.transaction`, so that this transformation can be applied to all transactions in the indexer without having to re-implement a rollback-node-aware traversal using `foldInExecutionOrder`.

Thanks for the explanation, I was mainly interested in whether this can be done, I did not have a look at whether it makes sense.

cocreature: Definitely in favor of trying to share this. We already have a kvutils utilties library in //daml-lf that erases rollbacks https://github.com/digital-asset/daml/blob/main/daml-lf/kv-transaction-support/src/main/scala/com/daml/lf/kv/TransactionNormalizer.scala. Having something more general that pairs this with blindinginfo seems sensible to me.

Issue #8773, opened 2021-02-08
Handle references to non-existent contract ids in Daml Script dumps


cocreature:
There are two cases where we might see a reference to a contract id that does not exist:

1. The contract was divulged. There isn’t anything we can do about this for now but ledger API changes are planned to allow us to fetch that contract. For now, we should just emit a reasonably useful error suggesting the cause. Another option could be to take the contract id as a parameter to the daml script dump and leave it to users to reconstruct it.
2. The contract id is just made up. As long as you don’t fetch the contract id, you can use an arbitrary string (with some formatting restrictions) in contract ids. This isn’t a useful case but we probably still want to handle it in some form at the very least by producing a reasonable error or maybe again by taking it as parameter.

Note that we cannot distinguish 1 and 2.

PR #9542, opened 2021-04-30
Separate Daml export creation from encoding

aherrmann-da:
- Separates collecting the information needed in a Daml export from its encoding to Daml script.
- Simplifies encoding by merging ACS and transaction tree actions before encoding.
- Checks for references to unknown contract ids in both ACS and transaction trees instead of only ACS.
- Adds test-cases for identifying known or unknown referenced contract ids.
- Fixes ACS topological sorting to not crash on references to unknown contract ids.

These changes are in preparation for addressing #8773.

### Pull Request Checklist

trigger the build.


daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
```diff
@@ -58,7 +58,7 @@ object TreeUtils {
     Graphs.topoSort(graph) match {
       case Left(cycle) =>
         throw new IllegalArgumentException(s"Encountered cyclic contract dependencies: $cycle")
-      case Right(sorted) => sorted.map(cid => acs.get(cid).get)
+      case Right(sorted) => sorted.collect(Function.unlift(cid => acs.get(cid)))
```
cocreature, line 61: We don’t actually expect to ever have `acs.get` return `None` right? No objection to the change just wondering. I sometimes tend to prefer a crash for unexpected cases to silently swallowing them but seems fine here.

cocreature (approved):
Nice cleanup, thank you!

daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
aherrmann-da: We do, if an ACS item has a reference to an unknown contract. E.g. the [`"unknown"` test-case](https://github.com/digital-asset/daml/pull/9542/files/65780dc798b7860e0d97c1a37cfc31d6a4ad8f00#diff-1fd7c179b7a83fed4577905906fd689510299a680d11464fc1a6aad1a60ceb3bR88) triggers this.

You're right that currently this doesn't occur with `daml ledger export` as invoked from the command-line because we guard against it with [this check](https://github.com/digital-asset/daml/pull/9542/files#diff-e376d2f8c626004f50a9317dc11a8944d4d9bc9eb00372ea4cd68628d5c5cfd6R131). However, the next step is to address https://github.com/digital-asset/daml/issues/8773 in which case we need to handle unknown contract ids here.

PR #9436, opened 2021-04-16
KVL-874 Add telemetry classes from the oem integration kit and use it for command submissions

hubert-da:
Moves telemetry classes from the oem-integration-kit repo and weaves the telemetry context from the command submission service to the ledger writer. It is then picked up by these changes: https://github.com/DACH-NY/daml-on-kv-oem-integration-kit/pull/1392

It is partially based on the PoC PR from @lucb-da: https://github.com/digital-asset/daml/pull/8284

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9543, opened 2021-04-30
KVL-919 Propagate trace context for package upload

hubert-da:
Introduces a breaking change (adding implicit `TelemetryContext`) to `WritePackagesService` which according to the discussion in the previous PR should be fine: #9436 (comment)

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/telemetry/SpanName.scala
+++ b/ledger/metrics/src/main/scala/com/daml/telemetry/SpanName.scala
@@ -1,0 +1,8 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.telemetry
+
+object SpanName {
+  val RunnerUploadDar: String = "daml.runner.upload-dar"
+}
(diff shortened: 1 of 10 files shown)


ledger/metrics/src/main/scala/com/daml/telemetry/SpanName.scala
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.telemetry
+
+object SpanName {
```
miklos-da, line 6: You may just inline this as it is only being used from two independent modules.

miklos-da (approved):

SamirTalwar-DA (approved):

ledger/metrics/src/main/scala/com/daml/telemetry/SpanName.scala
hubert-da: I'd like to gather all future values here, similarly to the oem integration kit.

miklos-da: In that case you might want to call this as `SpanNames`.

PR #9556, opened 2021-05-03
add Victor to app-runtime

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/release/rotation
+++ b/release/rotation
@@ -1,3 +1,4 @@
+U021DSF26BS victormueller-da
 UTUPD35SR aherrmann-da
 UPBBGF5MZ SamirTalwar-DA
 UEHLS0JUB cocreature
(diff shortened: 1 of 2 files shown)


release/rotation
```diff
@@ -1,3 +1,4 @@
+U021DSF26BS victormueller-da
```
cocreature, line 1: let’s give him a week or even start him at the bottom :slightly_smiling_face:

cocreature (approved):
Thanks!

release/rotation
garyverhaegen-da: @realvictorprm what do you think? Do you want to do the release this week (Wednesday), or would you rather have a bit more time? Either way, I'd be happy to walk you through the release process on Zoom for your first time.

PR #9559, opened 2021-05-03
Speed up daml-sdk-head Maven installation

cocreature:
This PR speeds up the Maven installation by doing 4 things:

1. Disable TS builds. This doesn’t make sense since we never install
them.
2. Add a --skip-sdk option that skips the SDK installation.
3. Add a --skip-jar-docs option that skips javadoc and more
importantly scaladoc.
4. Add a --scala-version=$VERSION option to only build against the
given version.

Unfortunately there are no tests and the CLI is a bit messy (you need
--scala-version=$VERSION and not --scala-version $VERSION).

Both are not great but this is a development only tool and it is bash
so my motivation for fixing it is rather low.

Given that this is growing in complexity a fair bit, I think we might
be better off as a long-term solution to move this towards at least
python or maybe even Haskell or Scala. But definitely nothing for this
PR.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/release/src/Options.hs
+++ b/release/src/Options.hs
@@ -6,6 +6,7 @@ module Options (
     parseOptions,
   ) where
 
+import qualified Data.Text as T
 import Options.Applicative
 
 import Types
@@ -18,6 +19,9 @@ data Options = Options
   { optsPerformUpload :: PerformUpload
   , optsReleaseDir :: FilePath
   , optsLocallyInstallJars :: Bool
+  , optIncludeTypescript :: IncludeTypescript
+  , optScalaVersions :: [T.Text]
+  , optIncludeDocs :: IncludeDocs
   } deriving (Eq, Show)
 
 optsParser :: Parser Options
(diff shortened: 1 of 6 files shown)


release/src/Options.hs
```diff
@@ -18,10 +20,20 @@ data Options = Options
...
+  , optIncludeDocs :: IncludeDocs
   } deriving (Eq, Show)
 
 optsParser :: Parser Options
 optsParser = Options
   <$> (PerformUpload <$> switch (long "upload" <> help "upload java/scala artifacts to Maven Central and typescript artifacts to the npm registry."))
   <*> option str (long "release-dir" <> help "specify full path to release directory")
   <*> switch (long "install-head-jars" <> help "install jars to ~/.m2")
+  <*> fmap (IncludeTypescript . not) (switch (long "no-ts" <> help "Do not build and upload typescript packages"))
+  <*> fmap (fromMaybe defaultScalaVersions) (optional (some (strOption (long "scala-version" <> help "Scala version to build JARs for"))))
```
aherrmann-da, line 34: Perhaps easier with `some ... <|> pure defaultScalaVersions`?

aherrmann-da (approved):
Nice! How much faster does it get?

release/src/Options.hs
cocreature: good idea, done

cocreature:
> Nice! How much faster does it get?

My testcase is modifying `com.daml.lf.data.ImmArray` after first building everything:

With this change and `daml-sdk-head --skip-sdk --skip-jar-docs --scala-version=2.13.5` (note that this is still slightly worse than it could be, the default Scala version is better):
~76s

Without this change and regular `daml-sdk-head`:
~18.45min (note the unit)

Not entirely sure I got the same number of cache hits but it definitely does match up with the numbers I would have guessed based on experience.

PR #9561, opened 2021-05-03
Dpp 336 sandbox classic on append only

nmarton-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
@@ -15,11 +15,6 @@ trait SequentialIndexer {
   def store(connection: Connection, offset: Offset, update: Option[Update]): Unit
 }
 
-object NoopSequentialIndexer extends SequentialIndexer {
-  override def store(connection: Connection, offset: Offset, update: Option[Update]): Unit =
-    throw new UnsupportedOperationException("NoopSequentialIndexer/store called. Not supported.")
-}
-
 case class SequentialIndexerImpl[DB_BATCH](
     storageBackend: StorageBackend[DB_BATCH],
     updateToDbDtos: Offset => Update => Iterator[DBDTOV1],
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
@@ -1,0 +1,77 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.indexer.sequential
+
+import java.sql.Connection
+
+import com.daml.ledger.participant.state.v1.{Offset, Update}
+import com.daml.platform.store.appendonlydao.JdbcLedgerDao
+import com.daml.platform.store.backend.{DBDTOV1, StorageBackend}
+
+import scala.util.chaining.scalaUtilChainingOps
+
+trait SequentialIndexer {
+  def store(connection: Connection, offset: Offset, update: Option[Update]): Unit
+}
+
+object NoopSequentialIndexer extends SequentialIndexer {
+  override def store(connection: Connection, offset: Offset, update: Option[Update]): Unit =
+    throw new UnsupportedOperationException("NoopSequentialIndexer/store called. Not supported.")
+}
+
+case class SequentialIndexerImpl[DB_BATCH](
+    storageBackend: StorageBackend[DB_BATCH],
+    updateToDbDtos: Offset => Update => Iterator[DBDTOV1],
+) extends SequentialIndexer {
+
+  private var lastEventSeqId: Long = _
+  private var lastEventSeqIdInitialized = false
+
+  private def lazyInit(connection: Connection): Unit =
+    if (!lastEventSeqIdInitialized) {
+      lastEventSeqId = storageBackend.ledgerEnd(connection).lastEventSeqId.getOrElse(0)
+      lastEventSeqIdInitialized = true
+    }
+
+  private def nextEventSeqId: Long = {
+    lastEventSeqId += 1
+    lastEventSeqId
+  }
+
+  private def adaptEventSeqIds(dbDtos: Iterator[DBDTOV1]): Vector[DBDTOV1] =
+    dbDtos.map { // TODO append-only: DRY, similar code as in the ParallelIndexerFactory
+      case e: DBDTOV1.EventCreate => e.copy(event_sequential_id = nextEventSeqId)
+      case e: DBDTOV1.EventDivulgence => e.copy(event_sequential_id = nextEventSeqId)
+      case e: DBDTOV1.EventExercise => e.copy(event_sequential_id = nextEventSeqId)
+      case notEvent => notEvent
+    }.toVector
+
+  override def store(connection: Connection, offset: Offset, update: Option[Update]): Unit =
+    synchronized {
+      lazyInit(connection)
+
+      val dbDtos = update
+        .map(updateToDbDtos(offset))
+        .map(adaptEventSeqIds)
+        .getOrElse(Vector.empty)
+
+      dbDtos
+        .pipe(storageBackend.batch)
+        .pipe(storageBackend.insertBatch(connection, _))
+
+      val configuration = dbDtos.reverseIterator
+        .collectFirst { // TODO append-only: DRY, same code as in the ParallelIndexerFactory
+          case c: DBDTOV1.ConfigurationEntry if c.typ == JdbcLedgerDao.acceptType => c.configuration
+        }
+
+      storageBackend.updateParams(
+        connection,
+        StorageBackend.Params(
+          ledgerEnd = offset,
+          eventSeqId = lastEventSeqId,
+          configuration = configuration,
+        ),
+      )
+    }
+}
--- a/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
@@ -12,19 +12,21 @@ import com.daml.logging.LoggingContext
 import com.daml.metrics.Metrics
 import com.daml.platform.ApiOffset
 import com.daml.platform.configuration.ServerRole
-import com.daml.platform.store.dao.JdbcLedgerDao
 import scalaz.Tag
 
 import scala.concurrent.{ExecutionContext, Future}
 
 object IndexMetadata {
 
-  def read(jdbcUrl: String)(implicit
+  def read(
+      jdbcUrl: String,
+      enableAppendOnlySchema: Boolean = false,
+  )(implicit
       resourceContext: ResourceContext,
       executionContext: ExecutionContext,
       loggingContext: LoggingContext,
   ): Future[IndexMetadata] =
-    ownDao(jdbcUrl).use { dao =>
+    ownDao(jdbcUrl, enableAppendOnlySchema).use { dao =>
       for {
         ledgerId <- dao.lookupLedgerId()
         participantId <- dao.lookupParticipantId()
@@ -32,20 +34,35 @@ object IndexMetadata {
       } yield metadata(ledgerId, participantId, ledgerEnd)
     }
 
-  private def ownDao(jdbcUrl: String)(implicit
+  private def ownDao(
+      jdbcUrl: String,
+      enableAppendOnlySchema: Boolean,
+  )(implicit
       executionContext: ExecutionContext,
       loggingContext: LoggingContext,
   ) =
-    JdbcLedgerDao.readOwner(
-      serverRole = ServerRole.ReadIndexMetadata,
-      jdbcUrl = jdbcUrl,
-      connectionPoolSize = 1,
-      eventsPageSize = 1000,
-      servicesExecutionContext = executionContext,
-      metrics = new Metrics(new MetricRegistry),
-      lfValueTranslationCache = LfValueTranslationCache.Cache.none,
-      enricher = None,
-    )
+    if (enableAppendOnlySchema)
+      com.daml.platform.store.appendonlydao.JdbcLedgerDao.readOwner(
+        serverRole = ServerRole.ReadIndexMetadata,
+        jdbcUrl = jdbcUrl,
+        connectionPoolSize = 1,
+        eventsPageSize = 1000,
+        servicesExecutionContext = executionContext,
+        metrics = new Metrics(new MetricRegistry),
+        lfValueTranslationCache = LfValueTranslationCache.Cache.none,
+        enricher = None,
+      )
+    else
+      com.daml.platform.store.dao.JdbcLedgerDao.readOwner(
+        serverRole = ServerRole.ReadIndexMetadata,
+        jdbcUrl = jdbcUrl,
+        connectionPoolSize = 1,
+        eventsPageSize = 1000,
+        servicesExecutionContext = executionContext,
+        metrics = new Metrics(new MetricRegistry),
+        lfValueTranslationCache = LfValueTranslationCache.Cache.none,
+        enricher = None,
+      )
 
   private val Empty = "<empty>"
 
--- a/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
@@ -52,8 +52,9 @@ object IndexMetadata {
         metrics = new Metrics(new MetricRegistry),
         lfValueTranslationCache = LfValueTranslationCache.Cache.none,
         enricher = None,
-        participantId =
-          "".asInstanceOf[v1.ParticipantId], // TODO fix ugly type cast // no participant id is available for the dump index meta path, also this property is not needed for the used the ReadDao
+        participantId = v1.ParticipantId.assertFromString(
+          "1"
+        ), // no participant id is available for the dump index meta path, also this property is not needed for the used the ReadDao
       )
     else
       com.daml.platform.store.dao.JdbcLedgerDao.readOwner(
--- a/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
@@ -6,6 +6,7 @@ package com.daml.platform.store
 import com.codahale.metrics.MetricRegistry
 import com.daml.buildinfo.BuildInfo
 import com.daml.ledger.api.domain.{LedgerId, ParticipantId}
+import com.daml.ledger.participant.state.v1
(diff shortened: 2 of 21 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -665,8 +863,49 @@ private class JdbcLedgerDao(
...
+      .executeSql(metrics.daml.index.db.storeTransactionDbMetrics) { implicit conn =>
+        sequentialIndexer.store(
+          conn,
+          offset,
+          validate(ledgerEffectiveTime, transaction, divulgedContracts) match {
+            case None =>
+              Some(
+                Update.TransactionAccepted(
+                  optSubmitterInfo = submitterInfo,
+                  transactionMeta = TransactionMeta(
```
mziolekda, line 878: Could you write a TODO to introduce a light-weight TransactionMeta for the read-path. This heavy one only makes sense in the write path. Using a type in such half-filled manner can lead to bad problems.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -792,6 +1059,10 @@ private[platform] object JdbcLedgerDao {
       validatePartyAllocation: Boolean = false,
       jdbcAsyncCommitMode: DbType.AsyncCommitMode = DbType.SynchronousCommit,
       enricher: Option[ValueEnricher],
+      sequentialIndexer: SequentialIndexer = NoopSequentialIndexer,
+      participantId: com.daml.ledger.participant.state.v1.ParticipantId = "".asInstanceOf[
```
mziolekda, line 1063: This can't be right to assign an empty participant. sandbox-classic takes a participant-id as a command line argument.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -13,13 +13,14 @@ trait StorageBackend[DB_BATCH] {
...
-  def initialize(connection: Connection): StorageBackend.Initialized
+  def initialize(connection: Connection): StorageBackend.LedgerEnd
+  def ledgerEnd(connection: Connection): StorageBackend.LedgerEnd
 }
 
 object StorageBackend {
   case class Params(ledgerEnd: Offset, eventSeqId: Long, configuration: Option[Array[Byte]])
 
-  case class Initialized(lastOffset: Option[Offset], lastEventSeqId: Option[Long])
+  case class LedgerEnd(lastOffset: Option[Offset], lastEventSeqId: Option[Long])
```
mziolekda, line 23:
```suggestion
  final case class LedgerEnd(lastOffset: Option[Offset], lastEventSeqId: Option[Long])
```
Same for Params above

ledger/sandbox-common/src/main/scala/platform/sandbox/cli/CommonCliBase.scala
```diff
@@ -303,6 +303,16 @@ class CommonCliBase(name: LedgerName) {
           s"The timeout used for requests by management services of the Ledger API. The default is set to ${SandboxConfig.DefaultManagementServiceTimeout.getSeconds} seconds."
         )
 
+      opt[Unit](
+        "enable-append-only-schema"
+      ) // TODO append-only: is this the right place to add this? Maybe too broad?
```
mziolekda, line 308: Looks like the right place

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -218,13 +230,111 @@ private class JdbcLedgerDao(
...
+              ),
+            )
+            PersistenceResponse.Ok
+          }).recover {
+            case NonFatal(e) if e.getMessage.contains(queries.DUPLICATE_KEY_ERROR) =>
+              logger.warn(
+                s"Ignoring duplicate party submission with ID ${partyDetails.party} for submissionId $submissionIdOpt"
+              )
+              conn.rollback(savepoint)
+              sequentialIndexer.store(conn, offsetStep.offset, None)
```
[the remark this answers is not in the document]
mziolekda: I take it back, we moved the offset update from the top. All good

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import java.sql.Connection
+
+import com.daml.ledger.participant.state.v1.{Offset, Update}
+import com.daml.platform.store.appendonlydao.JdbcLedgerDao
+import com.daml.platform.store.backend.{DBDTOV1, StorageBackend}
+
+import scala.util.chaining.scalaUtilChainingOps
+
+trait SequentialIndexer {
```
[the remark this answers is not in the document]
nmarton-da: why sealed? what is the benefit? Intention is to have something injected, not to declare a level in an ADT

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
```diff
@@ -0,0 +1,77 @@
...
+  def store(connection: Connection, offset: Offset, update: Option[Update]): Unit
+}
+
+object NoopSequentialIndexer extends SequentialIndexer {
+  override def store(connection: Connection, offset: Offset, update: Option[Update]): Unit =
+    throw new UnsupportedOperationException("NoopSequentialIndexer/store called. Not supported.")
+}
+
+case class SequentialIndexerImpl[DB_BATCH](
```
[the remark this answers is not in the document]
nmarton-da: why final? I have no intention of course to have this as a target for inheritance, but that is like usual in (functional) scala code. My point: why clutter if no benefit?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -218,13 +230,111 @@ private class JdbcLedgerDao(
       configuration: Configuration,
       rejectionReason: Option[String],
   )(implicit loggingContext: LoggingContext): Future[PersistenceResponse] =
-    throw new UnsupportedOperationException("not supported") // TODO add support
+    withEnrichedLoggingContext(Logging.submissionId(submissionId)) { implicit loggingContext =>
+      logger.info("Storing a configuration entry")
+      dbDispatcher.executeSql(
```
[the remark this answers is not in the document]
nmarton-da: in my understanding of the original code, the rollback there was only kicking in, if we have a duplication. with the change of the ingestion SQL just ignoring duplication, we wont have the rollback mechanism kicking in ever

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
```diff
@@ -0,0 +1,77 @@
...
+    storageBackend: StorageBackend[DB_BATCH],
+    updateToDbDtos: Offset => Update => Iterator[DBDTOV1],
+) extends SequentialIndexer {
+
+  private var lastEventSeqId: Long = _
```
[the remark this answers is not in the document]
nmarton-da: yes it would be, but on the other hand it would obfuscate the code because everywhere we would need to do a .get or a pattern match from the lastEventSeqId. not sure which is better.
if we wouldn't have to do lazy init, it would be much nicer.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: Well I would not implement an ADT instead the TransactionMeta, because that would obfuscate accessor code all along the lines. First idea comes in mind: just encapsulating this "light" nature with a well placed factory function :thinking: 
Anyhow I think it is good to keep in mind that all this is temporary until removal of sandbox-classic, so it would be great to minimize change on code which is permanent in this sense.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: the sequentialIndexer and the participantId is only used for the sandbox-classic integration, so I thought is it a good idea to limit the dependancy this way.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
nmarton-da: Maybe I missing here something? Why put final here and not everywhere else?
Scala is broad and you could do OO in scala yes, but where I come from I wouldn't ever do extend from a class, there are so many nicer functional ways to compose. It was some time ago I was working with hardcore OO language like java, but if I remember correctly even there is the saying to prefer composition to inheritance. So I sort of assumed in scala it is also ok to follow the same principle (even if not abandoning the whole OO like entirely ;) ), which would make all these final-s an unnecessary clutter to me.
Especially in case classes inheritance can be tricky with the hashcode equals contract, so AFAIK it is an antipattern.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: let's do that to prevent confusion in future

mziolekda (approved):

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -218,13 +231,111 @@ private class JdbcLedgerDao(
...
+
+        val update = finalRejectionReason match {
+          case None =>
+            Update.ConfigurationChanged(
+              recordTime = Time.Timestamp.assertFromInstant(recordedAt),
+              submissionId =
+                submissionId.asInstanceOf[SubmissionId], // TODO append-only: FIXME type cast
+              participantId = null.asInstanceOf[
+                v1.ParticipantId
+              ], // TODO append-only: FIXME type cast // not used for DBDTO generation
```
rautenrieth-da, line 267: (unrelated to this PR) Which begs the question, why does the field exist and why is it not indexed?

@mziolekda Do you know what the purpose of [ConfigurationChanged.participantId](https://github.com/digital-asset/daml/blob/main/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala#L34) is? This information is not stored in the index database, and is not used in the ledger API.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -218,13 +231,111 @@ private class JdbcLedgerDao(
...
+                  displayName = partyDetails.displayName.orNull,
+                  participantId = participantId,
+                  recordTime = Time.Timestamp.assertFromInstant(recordTime),
+                  submissionId = submissionIdOpt,
+                )
+              ),
+            )
+            PersistenceResponse.Ok
+          }).recover {
+            case NonFatal(e) if e.getMessage.contains(queries.DUPLICATE_KEY_ERROR) =>
```
rautenrieth-da, line 313: In `storeConfigurationEntry`, you have removed the savepoint-rollback-on-conflict in favor of an idempotent insert. Can we do the same here?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -332,7 +335,8 @@ object PostgresStorageBackend extends StorageBackend[RawDBBatchPostgreSQLV1] {
       |   typ_in,
       |   configuration_in,
       |   rejection_reason_in
-      | );
+      | )
+      | ON CONFLICT DO NOTHING;
```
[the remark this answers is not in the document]
rautenrieth-da: AFAICS, by moving the conflict detection from `JdbcLedgerDao` to here, you have lost the ability to log warnings for duplicate entries. How about the following changes:
- The above line changes to `preparedStatement.executeUpdate()` (which returns the number of rows affected)
- This method returns the number of rows affected
- The line `rawDBBatchPostgreSQLV1.configurationEntriesBatch.foreach` checks whether the number of affected rows equals the number of configuration entries in the batch. If it differs, it emits a warning.

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
rautenrieth-da, line 14: Will this class also be used for sandbox-on-x?

rautenrieth-da (approved):
Looks good, left a few minor comments.

The only thing that I found confusing is `SequentialIndexer`:
- It lives in the `indexer` package, but is not related to the `Indexer` trait.
- It is used by JdbcLedgerDao, but the DAO is a lower-level interface and should not know about an "indexer".

I think it's still fine because the use of `SequentialIndexer` in `JdbcLedgerDao` will be removed after sandbox classic is gone.

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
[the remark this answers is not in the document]
nmarton-da: I am happy to change.
Do you have a suggestion?
Do you think we should change the placement of it as well?

ledger/participant-integration-api/src/main/scala/platform/store/IndexMetadata.scala
```diff
@@ -6,46 +6,66 @@ package com.daml.platform.store
...
+        serverRole = ServerRole.ReadIndexMetadata,
+        jdbcUrl = jdbcUrl,
+        connectionPoolSize = 1,
+        eventsPageSize = 1000,
+        servicesExecutionContext = executionContext,
+        metrics = new Metrics(new MetricRegistry),
+        lfValueTranslationCache = LfValueTranslationCache.Cache.none,
+        enricher = None,
+        participantId =
+          "".asInstanceOf[v1.ParticipantId], // TODO fix ugly type cast // no participant id is available for the dump index meta path, also this property is not needed for the used the ReadDao
```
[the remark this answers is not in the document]
nmarton-da: ahhh, thanks for that, I will do this everywhere!
(we still need the comment at the end of the line though)

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -13,13 +13,14 @@ trait StorageBackend[DB_BATCH] {
   def batch(dbDtos: Vector[DBDTOV1]): DB_BATCH
   def insertBatch(connection: Connection, batch: DB_BATCH): Unit
   def updateParams(connection: Connection, params: StorageBackend.Params): Unit
-  def initialize(connection: Connection): StorageBackend.Initialized
+  def initialize(connection: Connection): StorageBackend.LedgerEnd
+  def ledgerEnd(connection: Connection): StorageBackend.LedgerEnd
```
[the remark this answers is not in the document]
nmarton-da: good point, I will add a todo to beef up scaladoc for this trait

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: I think it is not really possible to have here a duplication if there is conflict checking in the ledger this regard. I guess the only reason for this is in sandbox-classic: perhaps a race condition issuing the same config change command? :man_shrugging: 

Anyhow I am not entirely sure we need those warnings, and/or it is worth to make these changes just to have those. Let's discuss this offline. If there is a use case I am not against, but I would like to not take unclear complications if we don't have to.
Also we have to be aware that the original indexer code was not caring at all about duplication-s here, and the motivation of this change is to support sandbox-classic. Therefore I think we need to understand whether there is an issue of duplicates in non-sandbox-classic situations as well, because that might changes how we implement this.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -753,9 +999,33 @@ private[platform] object JdbcLedgerDao {
       lfValueTranslationCache,
       validatePartyAllocation,
       enricher = enricher,
+      participantId = participantId,
+      sequentialIndexer = sequentialIndexer(dbType, participantId, lfValueTranslationCache, metrics),
     ).map(new MeteredLedgerDao(_, metrics))
   }
 
+  private def sequentialIndexer(
```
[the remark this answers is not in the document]
nmarton-da: thanks for that, I agree. Normally I would wanted to have limit dependencies caused by not even used functionality, but the factories anyway have everything already except the participantId, which we changed after discussion with Martino, so until we have segregated JdbcLedgerDao, I agree this is simpler than the noop :+1:

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: We most certainly can, but that is also a gray area to me: let's try to understand the whole picture in a discussion jointly with the configuration changes.

ledger/participant-integration-api/src/main/scala/platform/indexer/sequential/SequentialIndexer.scala
nmarton-da: AFAIK not: with sandbox-on-x we do not need sequential inserts, we still want to do parallel ingestion if possible.
sandbox-on-x will have some groovy conflict checking logic otherwise to take care of conflicts.

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: As discussed with @rautenrieth-da we keep the implementation as is, and we open an implementation to handling duplicate submissions.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: as discussed we agreed to not change this to keep implementation as close as possible to the original. in that effort we agreed to use this approach here with the configuration change as well (which also gives us back the warning logs as well)

PR #9567, opened 2021-05-04
Proposal for command dedup changes to the ledger API

andreaslochbihler-da:
Proposal for extending the ledger API to implement the [new proposal for command deduplication](https://docs.google.com/document/d/1R0yAD2buzkHjPiWAHm-1FtX8acQofZLRGA4KDZ4NI2w/edit) (internal document). The extension aims at being backwards compatible on the wire.

**Comments for reviewers**
* Error reporting rules are summarized in `error.proto` and should probably be looked at first. This focuses on errors related to command deduplication. There is no attempt to clean-up the existing errors or document them better. Some known cleanup areas are marked as `TODO(error cleanup)`.
* Open discussion points are marked with `TODO(v2)`. Please comment if you think you can contribute something there.
* Ledger API users: Does this API make sense to you? Can the ledger API clients such as the JSON API server be adapted to work with this API, and extend the command deduplication guarantees to JSON API clients?
* Ledger API server / Daml ledger implementors: Do you foresee any problems implementing this API on top of the Read/WriteService API? The corresponding changes to the Read/WriteService API are sketched in #9638.

Out of scope:
* Updating the documentation on `docs.daml.com` (PR #9817).

### Pull Request Checklist

trigger the build.


ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -82,21 +92,67 @@ message SubmitAndWaitRequest {
...
+  // The completion offset for the submission that triggered this response.
+  // This offset can be used as the submission rank for subsequent command submissions.
+  // Optional
+  // TODO Make it required for responses with status code `OK` once all Daml ledgers support it.
+  string completion_offset = 1;
+
+  // Whether this response is a definite answer for the deduplication and rank guarantees,
+  // as defined in ``commands.proto``.
+  // Always true if the status code is `OK`.
+  bool definite_answer = 2;
```
andreaslochbihler-da, line 105: We want to return these two fields especially when the status code is not OK (so that the application gets hold of a higher offset for the next submission rank / can make a decision to not resubmit). I'm not sure whether this is the idiomatic way to do it in gRPC.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,37 @@ message Commands {
...
+    // The clock for this purpose is typically not fully synchronized with the application’s clock or record time,
+    // and message and processing delays may further increase the difference in interpretation.
+    google.protobuf.Timestamp deduplication_start = 15;
+
+    // Specifies the length of the deduplication period.
+    // The participant or Daml ledger converts the duration into a timestamp `deduplication_start` since
+    // by subtracting the duration from its local clock at some point during processing the submission.
+    //
+    // Must be non-negative.
+    google.protobuf.Duration deduplication_time = 9;
```
andreaslochbihler-da, line 90: The semantics of this field changes from extending into the future to looking into the past.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +32,37 @@ message Completion {
...
+  // The submission rank of the submission this completion refers to, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional
+  // TODO: Can be made required once all Daml ledgers return the application ID
+  string submission_rank = 5;
+
+  // Whether this completion is a definite answer for the deduplication and rank guarantees,
+  // as defined in ``commands.proto``.
+  // Completions with status code `OK` are always definite.
+  bool definite_answer = 8;
```
rautenrieth-da, line 64: This field alone is not enough to enable the application to react correctly to non-successful commands. Some rejections are final (e.g., a command that is not well authorised), some are retryable (e.g., contention on contract key), some are retryable in slightly different ways (e.g., ledger time model failure). Wouldn't it be better to standardise the error codes we use, or come up with our own error type?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +32,37 @@ message Completion {
...
+  // TODO: Can be made required once all Daml ledgers return the application ID
+  string application_id = 6;
+
+  // The set of parties on whose behalf the commands were executed.
+  // Contains the union of `party` and `act_as` from ``commands.proto``.
+  // The order of the parties need not be the same as in the submission.
+  // Each element must be a valid PartyIdString (as described in ``value.proto``).
+  // Optional
+  // TODO: Can be made required once all Daml ledgers return the application ID
+  repeated string act_as = 7;
```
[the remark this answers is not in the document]
andreaslochbihler-da: Yes. An application can submit with different `act_as` and command deduplication takes `act_as` into account.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: I agree that this information is not enough to determine an appropriate reaction. I do not intend to solve the error code problem as part of the command deduplication discussion though. I therefore propose to add this field to decouple the discussions. Even if we standardize on some error code scheme, whether something is a definite answer or not is IMO cross-cutting through the error hierarchy. I don't think it's a good idea to structure the hierarchy along the definite answer property.

Regarding error codes: IMO plain gRPC error codes are not enough to convey enough information about appropriate reactions of the applications. There are competing proposals circulating at the moment, like [gRPC's richer error model](https://cloud.google.com/apis/design/errors#error_model) and rolling our own system of error codes (https://github.com/DACH-NY/canton/issues/5853 gives an entry point into the Canton discussion).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
meiersi-da: I'd suggest to allow the participant to distinguish old and new semantics. This way it can provide proper log and error messages that inform the user about deprecation and defaulting. Moreover, users can choose the point at which they switch.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,37 @@ message Commands {
...
+  //
+  // The participant or Daml ledger may reject the submission with OUT_OF_RANGE
+  // if the deduplication period starts in the future (e.g., by specifying an offset that is higher than the completion end).
+  // It may reject the submission with NOT_FOUND
+  // if the deduplication period starts too early (e.g., by specifying an offset before the completion begin).
+  //
+  // If omitted, the participant chooses the start of the deduplication period at its discretion.
   // Optional
-  google.protobuf.Duration deduplication_time = 9;
+  oneof deduplication {
```
meiersi-da, line 73: Consider introducing a `DeduplicationMethod` message that provides space for parameters shared over the different ways of specifying the parameters. For example, whether the validity checks for the `deduplication_offset` should be skipped or not.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +122,37 @@ message Commands {
...
+  // Applications can rank their submissions with the same change ID.
+  // This allows the participants and Daml ledger to detect when one submission has been overtaken by a later resubmission.
+  // The participant and Daml ledger provide the following rank guarantee:
+  // Let there be two submissions for the same change ID with submission ranks `rank1 <= rank2` and submission IDs `sid1` and `sid2`.
+  // Suppose that the submission `sid2` has completed with a definite answer on the completion stream at offset `off2`.
+  // If there is no completion for the submission `sid1` before `off2`, then the submission `sid1` will result at most
+  // in a cancellation (it might be that the submission was lost and no result will ever reach the application).
+  //
+  // The submission rank is a completion offset.
+  // It must be no earlier than the offset that defines the start of the deduplication period.
```
meiersi-da, line 145:
```suggestion
  // It must be no earlier than the offset that defines the start of the deduplication period.
  // Otherwise, the submission is rejected with INVALID_ARGUMENT.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +32,37 @@ message Completion {
...
+  // The order of the parties need not be the same as in the submission.
+  // Each element must be a valid PartyIdString (as described in ``value.proto``).
+  // Optional
+  // TODO(v2): Can be made required once all Daml ledgers return the application ID
+  repeated string act_as = 7;
+
+  // The submission ID this completion refers to, as described in ``commands.proto``.
+  // Optional
+  // TODO(v2): Can be made required once all Daml ledgers return the application ID
+  // TODO(v2): Do we want to enforce the format of a UUID here?
```
meiersi-da, line 52: What do we gain by enforcing a UUID format?

I'd suggest we only limit the length, and otherwise treat it as an opaque string.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +32,37 @@ message Completion {
...
+  string submission_id = 4;
+
+  // The submission rank of the submission this completion refers to, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional
+  // TODO(v2): Can be made required once all Daml ledgers return the application ID
+  string submission_rank = 5;
```
meiersi-da, line 59: @andreaslochbihler-da : can you remind me again why this information is required?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
meiersi-da: I agree that we don't want to solve the full error problem here.

However, I would expect this PR to yield a protocol that enables applications to properly handle duplicates. AFAIR, an application requires some information (at least the `completion_offset`) of the completion that caused a `DUPLICATE` error. How do you propose handling this?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -82,21 +92,69 @@ message SubmitAndWaitRequest {
...
   // Must be a valid LedgerString (as described in ``value.proto``).
   // Required
+  // TODO(v2) Why is this field required? In ``completion.proto``, the `transaction_id` is optional.
   string transaction_id = 1;
+
+  // The completion offset for the submission that triggered this response
+  // This offset can be used as the submission rank for subsequent command submissions.
+  // Optional
+  // TODO(v2) Make it required for responses with status code `OK` once all Daml ledgers support it.
+  string completion_offset = 2;
```
meiersi-da, line 121: In case of a `DUPLICATE` completion for submission `sid2` that was caused by the submission `sid1`. Do we get the completion offset for `sid2` or the one for `sid1` here?

meiersi-da (commented):
Thank you @andreaslochbihler-da . This is complex work. I've done a first pass, but will require at least another one to wrap my head around it.

andreaslochbihler-da:
On how to communicate error details over gRPC: gRPC's [rich error model](https://cloud.google.com/apis/design/errors#error_model) used by Google Cloud isn't supported by [grpc-web](https://github.com/grpc/grpc-web/issues/736) nor the [Scala bindings](https://grpc.io/docs/guides/error/). Node.js doesn't support it [server-side](https://github.com/grpc/grpc-node/issues/1338) but there is a library for deserializing the details [on the client](https://github.com/stackpath/node-grpc-error-details).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: I agree with the advantages of a separate field. The downside is that all applications must explicitly be migrated to the new API, which is different from what we had discussed in the architecture meeting. The obvious compromise is accumulating some tech debt, namely having both fields, deprecating the old one at some point and removing it only after the deprecation cycle. Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: I don't yet see the use case where the application wants to skip the validity checks. If those checks are skipped, the deduplication guarantee doesn't apply. Then why does it set a deduplication offset in the first place (this is anyway optional).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: No operational benefits. It just conveys better the semantic meaning that this should be universally unique.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: Best practice of including all information that is relevant for expressing properties. For example:
* If you want to audit the completion stream of a participant, then you can check the rank guarantee without having to keep a lookup table from submission UUIDs to ranks.
* Some ledgers cannot enforce that the submission UUID is indeed unique because the application chooses the UUID. In those cases, one could not even build such a sensible lookup table.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
andreaslochbihler-da: This is the completion offset that triggered this response. The details of the other submission are in the error details. I'll separately sketch how I imagine to pass error details back to the application.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
```diff
@@ -51,6 +51,10 @@ message GetActiveContractsRequest {
   // Optional
   TraceContext trace_context = 1000;
 
+  // Whether synchronous API-level errors include the error details as part of the error message string,
+  // as explained in `errors.proto`.
```
meiersi-da, line 55: typo: the file is `error.proto`

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package com.daml.ledger.api.v1;
+
+// Rules on error reporting over the Ledger API
+//
+// 1. We use gRPC status codes to report synchronous success and failures
+//    coming from the underlying gRPC protocol and at the API level.
```
meiersi-da, line 11:
```suggestion
//    coming from the underlying gRPC protocol or the API-level request processing.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+//    which themselves include a `google.rpc.Status` message to indicate the status.
+//    The status code `OK` denotes success.
+//
+// 2. Error reporting includes relevant error details in the `details` field of the status message,
+//    following the gRPC rich error model as described at https://cloud.google.com/apis/design/errors#error_model.
+//    Error details use the `google.rpc.ErrorInfo` message defined at
+//    https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112.
+//    Its `reason` field includes an error code.
+//    Further details on the error and information for subsequent API usage may be included in `metadata`
+//    The RPC documentation may document further metadata of the error that applications should
```
meiersi-da, line 22: trailing text

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+//    Asynchronous API-level errors are reported over a completion stream like the command completion stream,
+//    which themselves include a `google.rpc.Status` message to indicate the status.
+//    The status code `OK` denotes success.
+//
+// 2. Error reporting includes relevant error details in the `details` field of the status message,
+//    following the gRPC rich error model as described at https://cloud.google.com/apis/design/errors#error_model.
+//    Error details use the `google.rpc.ErrorInfo` message defined at
+//    https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112.
+//    Its `reason` field includes an error code.
+//    Further details on the error and information for subsequent API usage may be included in `metadata`
```
meiersi-da, line 21:
```suggestion
//    Further details on the error and information for subsequent API calls may be included in the `metadata` field.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+// 1. We use gRPC status codes to report synchronous success and failures
+//    coming from the underlying gRPC protocol and at the API level.
+//    Asynchronous API-level errors are reported over a completion stream like the command completion stream,
+//    which themselves include a `google.rpc.Status` message to indicate the status.
+//    The status code `OK` denotes success.
+//
+// 2. Error reporting includes relevant error details in the `details` field of the status message,
+//    following the gRPC rich error model as described at https://cloud.google.com/apis/design/errors#error_model.
+//    Error details use the `google.rpc.ErrorInfo` message defined at
+//    https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112.
```
meiersi-da, line 19: @andreaslochbihler-da : is the plan to adopt an error payloads table as in https://cloud.google.com/apis/design/errors#error_payloads, but for now we only use `ErrorInfo`?

meiersi-da: Do you have plans to adopt https://github.com/googleapis/googleapis/blob/599fed858da3a3fdd59a2530046ee506928da25b/google/rpc/error_details.proto#L40 for the cases where clients can retry?

It might though not be sufficiently expressive for our retry cases.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,28 @@
...
+//    Its `reason` field includes an error code.
+//    Further details on the error and information for subsequent API usage may be included in `metadata`
+//    The RPC documentation may document further metadata of the error that applications should
+//
+// 3. RPC requests can specify that the API-level error details of the RPC response shall be JSON-encoded in the error message
```
meiersi-da, line 24: If you JSON encode the RPC response in the message, do you encode the `google.rpc.Status` object or just its `details` field?

I'm asking as in the latter case the caller wouldn't have access to a human-readable message.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: Let's leave it as a ledger string.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
andreaslochbihler-da: Fixed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: 1. For now, I'd go with `ErrorInfo` only. I find it more important to document somehow what error details we provide rather than creating sub-divisions like in the linked table. We can create such a table once we've built up some experience with our API.

2. `RetryInfo` is worth looking at when we actually deal with plain retryable errors. For command dedup, only few error situations are "retry as is", so I don't think it is worth to think about this for the command dedup work.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -24,12 +24,38 @@ service CommandService {
...
+  //     If the submission led to a completion event, the error details' metadata key ``completion_offset``
+  //     contains the corresponding completion offset. This offset can be used as a submission rank in later calls.
+  //     If the completion event at the offset is a definite answer, the metadata key ''definite_answer``
+  //     is set to ``true``.
   // - ``RESOURCE_EXHAUSTED``: if the number of in-flight commands reached the maximum (if a limit is configured)
   // - ``UNAVAILABLE``: if the participant is not yet ready to submit commands or if the service has been shut down.
-  // - ``ABORTED``: if a contract key is missing or duplicated due to for example contention on resources
-  rpc SubmitAndWait (SubmitAndWaitRequest) returns (google.protobuf.Empty);
+  // - ``ABORTED`` if one of the following holds:
+  //   - A contract key is missing or duplicated due to for example contention on resources.
```
meiersi-da, line 44:
```suggestion
  //   - A lock for serializability checking could not be acquired during commit time, e.g., due to contention.
```

meiersi-da: The contract key example seems too specific, and is also confusing wrt the the `INVALID_ARGUMENT` that I'd expect to be raised in case a `NoSuchKey` constraint is violated during commit time.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -24,12 +24,38 @@ service CommandService {
...
+  // - ``NOT_FOUND``: if the request does not include a valid ledger id or the submission rank or
+  //   deduplication period are too old.
+  // - ``INVALID_ARGUMENT`` if one of the following holds:
+  //   - The payload is malformed or is missing required fields.
+  //   - A party involved in the transaction is not known on the ledger.
+  //   - The interpreted command is inconsistent with the state of the ledger
+  //     If the submission led to a completion event, the error details' metadata key ``completion_offset``
+  //     contains the corresponding completion offset. This offset can be used as a submission rank in later calls.
+  //     If the completion event at the offset is a definite answer, the metadata key ''definite_answer``
+  //     is set to ``true``.
```
meiersi-da, line 39: I'd move this to a `FAILED_PRECONDITION` error, as it is the state of the system that makes this argument invalid.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
```diff
@@ -37,11 +37,17 @@ service CommandSubmissionService {
...
   // - ``UNAUTHENTICATED``: if the request does not include a valid access token
   // - ``PERMISSION_DENIED``: if the claims in the token are insufficient to perform a given operation
-  // - ``NOT_FOUND``: if the request does not include a valid ledger id
+  // - ``NOT_FOUND``: if the request does not include a valid ledger id or the submission rank or
+  //   deduplication period are too old.
   // - ``INVALID_ARGUMENT``: if the payload is malformed or is missing required fields
   // - ``UNAVAILABLE``: if the participant is not yet ready to submit commands or if the service has been shut down.
   // - ``RESOURCE_EXHAUSTED``: if the participant or the ledger is overloaded. Clients should back off exponentially and retry.
-  // - ``ABORTED``: if a contract key is missing or duplicated due to for example contention on resources
+  // - ``ABORTED``: if a contract key is missing or duplicated due to for example contention on resources or
```
meiersi-da, line 43: @andreaslochbihler-da : is the contract-key case really a synchronous error? I don't have a good model of where and how that happens.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
```diff
@@ -37,11 +37,17 @@ service CommandSubmissionService {
...
+  //   if there is another submission in flight for the same change ID.
+  // - ``FAILED_PRECONDITION`` if there is another submission with a higher submission rank.
```
meiersi-da, line 45: @andreaslochbihler-da : this is a definite rejection, that I'd expect to be communicated via a completion. Is that so?

meiersi-da: This way the caller is guaranteed to get a new completion offset to use a the submission rank.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
```diff
@@ -37,11 +37,17 @@ service CommandSubmissionService {
...
+  // - ``ALREADY_EXISTS`` if there is a definite completion for the same deduplication key with at least the same submission rank
```
meiersi-da, line 46: This seems to mix violations from deduplication with violations from the rank guarantee. It would make sense to allow reporting deduplication violations synchronously. For rank violations, the async route would be better as it generates a new completion offset.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,38 @@ message Commands {
...
   repeated string read_as = 13;
+
+  // A unique identifier to distinguish completions for different submissions with the same change ID.
+  // Typically a random UUID. Applications are expected to use a different UUID for each retry of a submission
+  // with the same change ID.
+  //
+  // If omitted, the participant or Daml ledger may set a UUID of their choice.
+  // Optional
+  //
+  // TODO(v2) Do we want to enforce the format of a UUID here?
```
meiersi-da, line 138: No, but add the constraint that it is a LedgerString

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,38 @@ message Commands {
...
+  string submission_id = 17;
+
+  // Applications can rank their submissions with the same change ID.
+  // This allows the participants and Daml ledger to detect when one submission has been overtaken by a later resubmission.
+  // The participant and Daml ledger provide the following rank guarantee:
+  // Let there be two submissions for the same change ID with submission ranks `rank1 <= rank2` and submission IDs `sid1` and `sid2`.
+  // Suppose that the submission `sid2` has completed with a definite answer on the completion stream at offset `off2`.
+  // If there is no completion for the submission `sid1` before `off2`, then the submission `sid1` will result at most
+  // in a cancellation (it might be that the submission was lost and no result will ever reach the application).
```
meiersi-da, line 147: 'cancellation' is not defined in the context of these Ledger API gRPC services. Consider referring to the `ALREADY_EXISTS` error.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -21,15 +21,45 @@ service CommandService {
 
   // Submits a single composite command and waits for its result.
   // Propagates the gRPC error of failed submissions including Daml interpretation errors.
+  // Error details are communicated using https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
+  // embedded within https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto
+  //
```
meiersi-da, line 26:
```suggestion
  // See `error.proto` for documentation on how error details are communicated using Google's
  // rich error model.
  //
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/ledger_configuration_service.proto
```diff
@@ -50,8 +55,8 @@ message LedgerConfiguration {
   reserved 1; // was min_ttl
   reserved 2; // was max_ttl
 
-  // The maximum value for the ``deduplication_time`` parameter of command submissions
-  // (as described in ``commands.proto``). This defines the maximum time window during which
-  // commands can be deduplicated.
+  // A lower bound on the length of the deduplication periods that can be used for command submissions
+  // (as described in ``commands.proto``). The deduplication period is measured on a local clock
+  // of the participant or Daml ledger.
   google.protobuf.Duration max_deduplication_time = 3;
```
meiersi-da, line 61: @andreaslochbihler-da : I'm unsure about this parameter. I'd have expect that parameter to remain an upper bound on the time-based deduplication options that ledgers MAY enforce. I'll need to do some more digging wrt KV ledgers and their deduplication implementation to understand what we need here. If you already have a good mental model for it, I'd appreciate you sharing it.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -22,6 +22,11 @@ message GetLedgerApiVersionRequest {
   // Must be a valid LedgerString (as described in ``value.proto``).
   // Required
   string ledger_id = 1;
+
+  // Whether synchronous API-level errors include the error details as part of the error message string,
+  // as explained in `error.proto`.
+  // Optional
+  bool error_details_in_message = 1001;
```
meiersi-da, line 29: @andreaslochbihler-da : the changes wrt `error_details_in_message` are cluttering the API surface. I'm wondering whether these wouldn't be better communicated via an extra header being set. That would also allow adding that parameter transparently by middle-ware.

I'd suggest revisiting this choice, and potentially factoring it out of this PR.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -23,6 +23,11 @@ message Completion {
   // Identifies the exact type of the error.
   // For example, malformed or double spend transactions will result in a ``INVALID_ARGUMENT`` status.
   // Transactions with invalid time time windows (which may be valid at a later date) will result in an ``ABORTED`` error.
+  //
```
meiersi-da, line 26: This comment is slightly confusing. I'd expect it to talk about both success and error.

meiersi-da (commented):
@andreaslochbihler-da : I've completed the second round review. Thanks for all this good work. Comments inline.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
```diff
@@ -21,10 +21,8 @@ option csharp_namespace = "Com.Daml.Ledger.Api.V1";
 //
 // Commands may fail in 2 distinct manners:
 //
-// 1. Failure communicated in the gRPC error of the submission.
-// 2. Failure communicated in a Completion.
-//
-// Only successfully submitted commands may produce a completion event.
```
andreaslochbihler-da, line 27: @meiersi-da Why did you remove this line?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
andreaslochbihler-da: That's IMO part of cleaning up our error story. It has little to do with command deduplication as such. We currently do report such a case as `INVALID_ARGUMENT`.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
andreaslochbihler-da: It can be synchronous or asynchronous, depending on where the check fails (during submission on the participant or during conflict detection on the ledger).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
andreaslochbihler-da: If there's a conflict on ranks, we agreed to make that a cancellation. We should have here the same text as for the command service, namely that the error details contain the conflicting rank.

We should find a way to avoid copy-pasting all this stuff.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: Rephrased into `definite answer`. `ALREADY_EXISTS` is rather orthogonal to whether the completion is a definite answer.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: Agreed. We should find a single place for documenting errors. Maybe `error.proto`?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
andreaslochbihler-da: Add a `// TODO(error clean-up)` to error.proto

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_submission_service.proto
andreaslochbihler-da: Double-check condition `at least the same submission rank`

ledger-api/grpc-definitions/com/daml/ledger/api/v1/ledger_configuration_service.proto
andreaslochbihler-da: rename like in write service

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
andreaslochbihler-da: Rewritten in error.proto and added a TODO

ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
```diff
@@ -50,7 +50,7 @@ message GetActiveContractsRequest {
   // This field is a future extension point and is currently not supported.
   // Optional
   TraceContext trace_context = 1000;
-
+
```
meiersi-da, line 53:
```suggestion
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,38 @@ message Commands {
...
+  string submission_id = 17;
+
+  // Applications can rank their submissions with the same change ID.
+  // This allows the participants and Daml ledger to detect when one submission has been overtaken by a later resubmission.
+  // The participant and Daml ledger provide the following rank guarantee:
+  // Let there be two submissions for the same change ID with submission ranks `rank1 <= rank2` and submission IDs `sid1` and `sid2`.
+  // Suppose that the submission `sid2` has completed with a definite answer on the completion stream at offset `off2`.
+  // If there is no completion for the submission `sid1` before `off2`, then the submission `sid1` will not result
+  // in a definite answer or successful completion. So if a result ever reaches the application,
+  // the result will be a rejection that is not marked as a definite answer.
```
meiersi-da, line 147:
```suggestion
  // the result will be a rejection that is not marked as a definite answer. The application
  // can therefore assume that the change in submission `sid1` will not be applied to the 
  // Daml ledger.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,126 @@
...
+//   that the request succeeded nor that it was rejected.
+//
+// Error codes for command submissions
+// ***********************************
+//
+// TODO(v2) Define test cases for these errors and enforce these error rules via the ledger API test tool
+//
+// This applies to the services in ``command_service.proto`` and ``command_submission_service.proto``.
+// The common error codes from above also apply here.
+// The status codes in command completions as defined in ``completion.proto`` follow the same rules.
```
meiersi-da, line 78: Code vs google.rpc.Status message

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,126 @@
...
+//
+// Some Daml ledger implementations may produce the same error codes in further situations.
+// See the documentation of the specific Daml ledger implementation.
+//
+// If the submission led to a completion event, the following error details are provided:
+// 1. If the error is reported outside of the completion event (e.g., via the command service),
+//    the metadata key ``completion_offset`` contains the corresponding completion offset.
+//    This offset can be used as a submission rank in later calls.
+// 2. If the completion event at the offset is a definite answer,
```
meiersi-da, line 87: @andreaslochbihler-da: the notion of a definite answer is actually not defined in the Ledger API. It would probably make sense to do that in `commands.proto`

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,38 @@ message Commands {
...
+  // It must be no earlier than the offset that defines the start of the deduplication period.
+  // Otherwise, the submission is rejected with INVALID_ARGUMENT.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  //
+  // The participant or Daml ledger may reject the submission with ``OUT_OF_RANGE``
+  // if the submission rank offset is higher than the completion end.
+  // It may reject the submission with NOT_FOUND
+  // if the submission rank is too low (e.g., lower than the the completion begin).
+  //
+  // If omitted, the participant defaults the submission rank to the completion end.
```
meiersi-da, line 152: I've added some changes, but we should consider of whether we want to remove these side-constraints completely; and assume people will discover them either by reading `error.proto` or by encountering the error and then acting on it.

meiersi-da (commented):
@andreaslochbihler-da : another round of review done. Also see the commit with some inline changes.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da:
```suggestion
// The status message in command completions as defined in ``completion.proto`` follow the same rules.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,126 @@
...
+//    the metadata key ''definite_answer`` is set to ``true``.
+//
+// ``UNAVAILABLE``:
+// - The participant is not yet ready to submit commands or if the service has been shut down.
+//
+// ``RESOURCE_EXHAUSTED``:
+// - The number of in-flight commands reached the maximum (if a limit is configured).
+//
+// ``NOT_FOUND``:
+// - The submission rank or deduplication period are too old.
```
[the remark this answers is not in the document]
andreaslochbihler-da: The participant may say `too old` even though it's not the completion begin, e.g., on kv ledgers without pruning where we strictly use the max deduplication timeout instead of offsets.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,38 @@ message Commands {
...
+  // If omitted, the participant or Daml ledger may set a value of their choice.
+  // Optional
+  string submission_id = 17;
+
+  // Applications can rank their submissions with the same change ID.
+  // This allows the participants and Daml ledger to detect when one submission has been overtaken by a later resubmission.
+  // The participant and Daml ledger provide the following rank guarantee:
+  // Let there be two submissions for the same change ID with submission ranks `rank1 <= rank2` and submission IDs `sid1` and `sid2`.
+  // Suppose that the submission `sid2` has completed with a definite answer on the completion stream at offset `off2`.
+  // If there is no completion for the submission `sid1` before `off2`, then the submission `sid1` will not result
```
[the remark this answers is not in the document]
andreaslochbihler-da: The majority of usages is two backticks; made this uniform now.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/command_service.proto
```diff
@@ -17,86 +17,75 @@ option csharp_namespace = "Com.Daml.Ledger.Api.V1";
...
   // This field is a future extension point and is currently not supported.
   // Optional
   TraceContext trace_context = 1000;
+}
 
+message SubmitAndWaitResponse {
+  // The completion offset for the submission that triggered this response.
+  // This offset can be used as the submission rank for subsequent command submissions.
+  // Optional
+  string completion_offset = 1;
```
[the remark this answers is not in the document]
andreaslochbihler-da: After a crash, an application might not know whether a submission is the first one for the change ID. We had come up with the idea of sending an initial transaction over the ledger to obtain a fresh offset. If you want to do that via the command service, we need to report the offset somewhere. And this is the place.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: Given that "definite answer" is a property of a completion, I'd rather define it in `completion.proto` and add a reference here and in `commands.proto`.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,42 @@ message Commands {
...
-  // the same command ID will be deduplicated.
-  // Duplicate commands submitted before the end of this window return an ALREADY_EXISTS error.
+  // Specifies the deduplication period for the change ID.
+  // Deduplication provides the following deduplication guarantee:
+  // If the completion for this submission is a definite answer (as defined in ``completion.proto``), say at the completion offset ``off``,
+  // then there is no accepting completion for the same change ID between the start of the deduplication period and ``off``.
+  //
+  // The participant or Daml ledger may reject the submission with ``OUT_OF_RANGE``
+  // if the deduplication period starts in the future (e.g., by specifying an offset that is higher than the completion end).
+  // It may reject the submission with NOT_FOUND
```
andreaslochbihler-da, line 68:
```suggestion
  // It may reject the submission with ``NOT_FOUND``
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +127,36 @@ message Commands {
...
+  // Applications can rank their submissions with the same change ID.
+  // This allows the participants and Daml ledger to detect when one submission has been overtaken
+  // by a later resubmission.
+  //
+  // The participant and Daml ledger provide the following rank guarantee:
+  // Let there be two submissions for the same change ID with submission ranks ``rank1 <= rank2``
+  // and submission IDs ``sid1`` and ``sid2``. Suppose that the submission `sid2` has completed
+  // with a definite answer (as defined in ``completion.proto``) on the completion stream at offset ``off2``.
+  // If there is no completion for the submission ``sid1`` before ``off2``, then the submission ``sid1`` will not result
+  // in a definite answer or successful completion. So if a result ever reaches the application,
```
andreaslochbihler-da, line 149:
```suggestion
  // in a definite answer completion. So if a result ever reaches the application,
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,42 @@ message Commands {
...
-  // Duplicate commands submitted before the end of this window return an ALREADY_EXISTS error.
+  // Specifies the deduplication period for the change ID.
+  // Deduplication provides the following deduplication guarantee:
+  // If the completion for this submission is a definite answer (as defined in ``completion.proto``), say at the completion offset ``off``,
+  // then there is no accepting completion for the same change ID between the start of the deduplication period and ``off``.
+  //
+  // The participant or Daml ledger may reject the submission with ``OUT_OF_RANGE``
+  // if the deduplication period starts in the future (e.g., by specifying an offset that is higher than the completion end).
+  // It may reject the submission with NOT_FOUND
+  // if the deduplication period starts too early (e.g., by specifying an offset before the completion begin).
```
andreaslochbihler-da, line 69:
```suggestion
  // The submission will be rejected if the deduplication period starts in the future
  // (e.g., by specifying an offset that is higher than the completion end)
  // or too early (e.g., by specifying an offset before the completion begin).
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: There's now a write-up of how to communicate error details in `error.proto`.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,130 @@
...
+
+syntax = "proto3";
+
+package com.daml.ledger.api.v1;
+
+// Rules on error reporting over the Ledger API
+// ********************************************
+//
+// 1. We use gRPC status codes to report synchronous success and failures
+//    coming from the underlying gRPC protocol or the API-level request processing.
```
cocreature, line 12: Do you have a definition of API-level? I think I know what you mean but I have trouble pinning it down precisely in a way that gets at the difference between the CommandService and the CommandSubmissionService.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,130 @@
...
+//    e.g. because some middleware drops them, it can obtain the details by parsing the message.
+//    This flag does not affect asynchronous errors reported as part of completions.
+//
+//    TODO(v2) Evaluate whether we can implement this as a gRPC interceptor that listens for the header
+//     and converts the rich error details into JSON. Otherwise this might not work for
+//     errors coming from the underlying gRPC protocol.
+//     If this doesn't work, we can alternatively add a flag to all RPC requests.
+//     For transparency, consider encoding the whole ``google.rpc.Status message`` in the message.
+//
+//    TODO(v2) We shoule use the same JSON encoding here as what we get from our JSON API server.
```
cocreature, line 37:
```suggestion
//    TODO(v2) We should use the same JSON encoding here as what we get from our JSON API server.
```

cocreature, line 37: Not quite sure what you mean here. We don’t have structured errors on the JSON API atm. Errors are an http status code + a list of string error messages.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,130 @@
...
+// ``NOT_FOUND``:
+// - The submission rank or deduplication period are too old.
+//
+// ``OUT_OF_RANGE``
+// - The deduplication period starts in the future (e.g., the specified offset is higher than the completion end).
+// - The submission rank is after the completion end.
+//
+// ``INVALID_ARGUMENT``:
+// - A party involved in the transaction is not known on the ledger.
+// - The interpreted command is inconsistent with the state of the ledger.
```
cocreature, line 106: I think this needs to be defined more precisely. Probably this PR is not the right place for doing that though.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
```diff
@@ -0,0 +1,130 @@
...
+//
+// 2. Error reporting includes relevant error details in the ``details`` field of the status message,
+//    following the gRPC rich error model as described at https://cloud.google.com/apis/design/errors#error_model.
+//    Error details use the `google.rpc.ErrorInfo` message defined at
+//    https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L112.
+//    Its ``reason`` field includes an alphanumeric error code.
+//    Further details on the error and information for subsequent API calls may be included in the ``metadata`` field.
+//    The Ledger API RPC documentation may document further metadata of the error.
+//
+// 3. RPC requests can specify that the API-level error details of the RPC response shall be JSON-encoded in the error message
```
cocreature, line 25: What data points are you basing the need for this feature for? I’m aware of http trailers causing issues (or more generally) but you’re not avoiding trailersly. You’re just a bit more robust against things meddling with the trailer. Do we have a client that uses something like this? Or is there a common middleware that does this?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: The idea is that API-level errors are those that our implementation of the gRPC endpoint produces, in contrast to errors coming from the underlying gRPC library or even some transport layer. It has nothing to do with command service vs. command submission service. The distinction is made only because we can't really influence how those libraries produce their errors.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: I'm not aware of a client that requested this. I've just seen that [grpc-web](https://github.com/grpc/grpc-web) throws away the error details. And some gRPC libraries don't support the error details out of the box; the Java version is fine, but the native Scala bindings don't.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: That's why there's the `TODO(error cleanup)` in the next line. As the PR description says, this needs to be worked out in detail, but not here.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
andreaslochbihler-da: With the new proposal, applications will need to obtain some information from the errors that they can use for retries or further submissions (mostly offsets). So what does the JSON API server do if it gets a gRPC error with rich error details? How do you pass this information on to the client?

cocreature: Ah I misread this as suggesting to match the existing error format. Matching how we represent the rich errors seems very sensible :+1: I’m hoping to find the time today to write up a proposal for this.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
cocreature: Thanks for the explanation, makes sense

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
cocreature: Makes sense, afaict grpc-web does actually expose it, it just seems awkward to use it, e.g., https://github.com/shumbo/grpc-web-error-details. I assume by Scala you mean akka-grpc?

Assuming we can fix our own Scala uses, we could probably get away with not having this given that we only support the Java bindings externally and add it later if required but probably simpler to just add it directly.

andreaslochbihler-da: With Scala, I meant what `scalapb` generates. Errors are reported as `StatusRuntimeException` exceptions and if you convert them back into a Status object with `io.grpc.Status#fromThrowable`, the details are gone.

cocreature:
For reference, here are the issues for the corresponding changes on the JSON API:
1. error handling https://github.com/digital-asset/daml/issues/9834
2. dedup https://github.com/digital-asset/daml/issues/9835

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,41 @@ message Commands {
...
+    string deduplication_offset = 14;
+
+    // Specifies a point in time at which the deduplication shall start.
+    // The participant or Daml ledger converts this timestamp into a completion offset using its local clock.
+    // The clock for this purpose is typically not fully synchronized with the application’s clock or record time,
+    // and message and processing delays may further increase the difference in interpretation.
+    google.protobuf.Timestamp deduplication_start = 15;
+
+    // Specifies the length of the deduplication period.
+    // The participant or Daml ledger converts the duration into a timestamp ``deduplication_start`` since
```
meiersi-da, line 90:
```suggestion
    // The participant or Daml ledger converts the duration into a timestamp ``deduplication_start``
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/error.proto
meiersi-da: If we could avoid adding this feature that would likely save us engineering effort.

So we could also consider doing the PoC's without this feature available, and checking the Dev UX before adding it.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +58,41 @@ message Commands {
...
+    // by subtracting the duration from its local clock at some point during processing the submission.
+    //
+    // Must be non-negative.
+    google.protobuf.Duration deduplication_since = 16;
```
meiersi-da, line 94:
```suggestion
    google.protobuf.Duration deduplication_duration = 16;
```
I noticed that it is easier to talk about duration and explain that they are applied backwards than talking about a "deduplication-since" value or the like.

andreaslochbihler-da: `deduplication_since` isn't a good name. In the Read/WriteService API, I've used the term `deduplication_window` for this, but `deduplication_duration` also sounds reasonable. Changed to `duration` for now. Please shout if you think `window` is better.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
andreaslochbihler-da: Changed.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
meiersi-da: Thanks. The write-up for KV ledgers worked well with duration.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +36,51 @@ message Completion {
...
+  // The submission ID this completion refers to, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional for historic completions where this data is not available.
+  string submission_id = 4;
+
+  // The submission rank of the submission this completion refers to, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional for historic completions where this data is not available.
+  string submission_rank = 5;
+
```
meiersi-da, line 60:
```suggestion
   // Record time of this completion.
   google.protobuf.Timestamp record_time = 6;
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -32,6 +36,51 @@ message Completion {
...
+  // The ledger may convert the deduplication period into other descriptions and extend the period
+  // in implementation-specified ways.
+  // Used to audit the deduplication guarantee described in ``commands.proto``.
+  //
+  // Optional; the deduplication guarantee applies even if the completion omits the deduplication period.
+  oneof deduplication {
+    // Specifies the start of the deduplication period by a completion stream offset.
+    //
+    // Must be a valid LedgerString (as described in ``value.proto``).
+    string deduplication_offset = 8;
```
meiersi-da, line 71: @andreaslochbihler-da : is the gap in field numbers intentional or due to a copy-past case?

Also: did you consider defining a DeduplicationPeriod message and reusing that here and in the `SubmitterInfo`?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: Isn't the record time communicated as part of the Checkpoint that contains the completions?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: Field numbers 6 and 7 are already assigned further up (`applicationId` and `actAs`). But given that we're introducing them as part of this PR anyway, it makes sense to assign them consecutively. Changed.

andreaslochbihler-da: I didn't introduce a separate `DeduplicationPeriod` message for symmetry with `Commands`, where we decided that such a message would break backwards compatibility.

meiersi-da: Thanks for reminding me of the backwards-compatibility issue. I forgot about it. Makes sense.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: I've never understood why we distinguish between completion events and checkpoints. My current understanding is that we emit a checkpoint for every completion, but I may well be wrong.

andreaslochbihler-da:
Command Dedup API has been released with Daml 2.0, albeit without submission ranks.

PR #9568, opened 2021-05-04
Accept a comma-separated list of parties for daml ledger export

aherrmann-da:
Addresses https://github.com/digital-asset/daml/pull/9446/files/9de406ba6858f37c3c40baf3c6a3cf64d4ef36c4#r616379478

- Users can specify multiple parties using either `--party Alice,Bob` or `--party Alice --party Bob`.
- Adds a test-suite for command-line flag parsing.
- The test-suite highlighted that the `--end` flag was not implemented correctly. This PR fixes that as well.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/export/src/main/scala/com/daml/script/export/Config.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/Config.scala
@@ -11,7 +11,7 @@ import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
 final case class Config(
     ledgerHost: String,
     ledgerPort: Int,
-    parties: List[String],
+    parties: Seq[String],
     start: LedgerOffset,
     end: LedgerOffset,
     exportType: Option[ExportType],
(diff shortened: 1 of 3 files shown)


daml-script/export/src/main/scala/com/daml/script/export/Config.scala
```diff
@@ -63,7 +66,7 @@ object Config {
       )
     opt[String]("end")
       .optional()
-      .action((x, c) => c.copy(start = parseLedgerOffset(x)))
+      .action((x, c) => c.copy(end = parseLedgerOffset(x)))
```
cocreature, line 69: :facepalm:

daml-script/export/src/main/scala/com/daml/script/export/Config.scala
```diff
@@ -53,8 +53,11 @@ object Config {
     opt[String]("party")
       .required()
       .unbounded()
-      .action((x, c) => c.copy(parties = x :: c.parties))
-      .text("Export ledger state as seen by these parties.")
+      .action((x, c) => c.copy(parties = x.split(",").toList ++ c.parties))
```
cocreature, line 56: afaik you don’t have to split this yourself, you can specify a `Seq[String]` and scopt will do it for you. see https://github.com/digital-asset/daml/blob/2935b6fe84c5e9abd10e25849c838577249f5e77/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/Cli.scala#L146 for example

cocreature (approved):
Nice, thank you!

daml-script/export/src/main/scala/com/daml/script/export/Config.scala
aherrmann-da: Nice, thanks for the pointer!

PR #9572, opened 2021-05-04
LF: SBUKeyBuiltin clean up.

remyhaemmerle-da:
This PR some minor clean up for #9537 nd #9538.
    
In particular it:
    
- Refactor the class KeyOperation outside SBUKeyBuiltin

- Renames and refactor methods `handleKeyFound`, `handleKeyArchived`,
     `cidToSValue` and `cidToSExpr` to gave them a more obvious meaning.
    
- Replaces `immutable.Map#+` by `immutable.Map#updated` (cosmetic)
    
- Replaces `trait` by `abstract class` (cosmetic)

- Replaces 'get(???).getOrElse(???)' by `getOrElse(???, ???)` (cosmetic)

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1170,11 +1170,16 @@ private[lf] object SBuiltin {
     *   -> Maybe (ContractId T)
     */
   final case class SBULookupKey(templateId: TypeConName) extends SBUKeyBuiltin(templateId) {
-    override protected final def execute(
-        args: util.ArrayList[SValue],
-        machine: Machine,
-        onLedger: OnLedger,
-    ): Unit = handleKey(args, machine, onLedger, Lookup)
+    override protected def handleNewKeyNotFound(machine: Machine): Boolean = {
+      machine.returnValue = SValue.SValue.None
+      true
+    }
+    override protected def handleInactiveKey(machine: Machine, key: GlobalKey) =
+      machine.returnValue = SValue.SValue.None
+    override protected def handleActiveKey(machine: Machine, cid: V.ContractId) =
+      machine.returnValue = SOptional(Some(SContractId(cid)))
+    override protected def handleNewKeyFound(machine: Machine, cid: V.ContractId): Unit =
+      machine.ctrl = SBSome(importCid(cid))
   }
 
   /** $insertLookup[T]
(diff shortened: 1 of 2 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1215,48 +1220,32 @@ private[lf] object SBuiltin {
...
-      def cidToSValue(cid: V.ContractId): SValue
-      def cidToSExpr(cid: V.ContractId): SExpr
-    }
-    final case object Fetch extends KeyOperation {
-      override def handleKeyNotFound(machine: Machine): Boolean =
-        machine.tryHandleSubmitMustFail()
-      override def handleKeyArchived(machine: Machine, key: GlobalKey): Unit =
-        // TODO (MK) Produce a proper error here.
-        crash(s"Could not find key $key")
-      override def cidToSValue(cid: V.ContractId) =
```
remyhaemmerle-da, line 1234: The important of this method is not the returned type, but the fact it does not check for freshness.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1215,48 +1220,32 @@ private[lf] object SBuiltin {
...
-        SContractId(cid)
-      override def cidToSExpr(cid: V.ContractId) =
```
remyhaemmerle-da, line 1236: The importand of this methiod is not the return type, but the fact it check for cid freshness.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1265,38 +1254,30 @@ private[lf] object SBuiltin {
...
+            .getOrElse(coid, crash(s"Local contract $coid not in cachedContracts"))
           val stakeholders = cachedContract.signatories union cachedContract.observers
           throw SpeedyHungry(
             SResultNeedLocalKeyVisible(
               stakeholders,
               onLedger.committers,
               {
                 case SVisibleByKey.Visible =>
-                  machine.returnValue = operation.cidToSValue(coid)
+                  handleActiveKey(machine, coid)
```
cocreature, line 1265: not a huge fan of this, you are factoring less than we did before.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1265,38 +1254,30 @@ private[lf] object SBuiltin {
...
                   )
               },
             )
           )
-        case Some(KeyActive(coid)) =>
-          machine.returnValue = operation.cidToSValue(coid)
-        case Some(KeyInactive) =>
-          operation.handleKeyArchived(machine, gkey)
+        case Some(keyMapping) =>
+          handleKnownKey(machine, gkey, keyMapping)
```
cocreature, line 1274: same here, there is less factoring than before.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1215,48 +1220,32 @@ private[lf] object SBuiltin {
...
-        true
-      }
-      override def handleKeyArchived(machine: Machine, key: GlobalKey): Unit = {
-        machine.returnValue = SV.None
+    private[this] val typ = AstUtil.TContractId(Ast.TTyCon(templateId))
+    final protected def importCid(cid: V.ContractId): SEImportValue =
+      SEImportValue(typ, V.ValueContractId(cid))
+
+    // Callback from the engine returned NotFound
+    protected def handleNewKeyNotFound(machine: Machine): Boolean
```
cocreature, line 1230:
```suggestion
    protected def handleGlobalKeyNotFound(machine: Machine): Boolean
```
maybe a bit clearer?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1215,48 +1220,32 @@ private[lf] object SBuiltin {
...
+    // We already saw this key, but it was archived
```
cocreature, line 1231: We also get this for a negative lookup I believe not just an archive.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1215,48 +1220,32 @@ private[lf] object SBuiltin {
...
+    protected def handleInactiveKey(machine: Machine, gkey: GlobalKey): Unit
+    // We already saw this key and is still active
+    protected def handleActiveKey(machine: Machine, cid: V.ContractId): Unit
+    // CallBack from the engine returned a new Cid
+    protected def handleNewKeyFound(machine: Machine, cid: V.ContractId): Unit
```
cocreature, line 1236: I’d place this next to the corresponding `handleNewKeyNotFound`.

cocreature (commented):
I’m open to merging this but I’d like to be convinced first that this is an improvement: Before it was trivial to see what actually depends on the operation: Everything that called `operation` in some form.
Now I have to check which method call is `private` which is `protected` and for some of the protected calls we’re factoring some of the shared code into those methods so we lose the guarantee that they behave the same for those parts.

remyhaemmerle-da:
> I’m open to merging this but I’d like to be convinced first that this is an improvement: Before it was trivial to see what actually depends on the operation: Everything that called `operation` in some form.
> Now I have to check which method call is `private` which is `protected` and for some of the protected calls we’re factoring some of the shared code into those methods so we lose the guarantee that they behave the same for those parts.

Sure, I have put the PR as draft to discuss advantage/inconvenient. 
My approach is more traditional Scala/object oriented. 

Let me think a bit about your arguments.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: I let it here to see the correspondence with `cidToSExpr`. I will move it before merging, if we merge.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: Negative lookup are handled by `handleKeyNotFound`, aren't they ? 
As far I can tell, the original `handleKeyArchived` appeared only twice, each time under a pattern matching with `KeyInactive`, which seems to indicate it is use only for inactive key.

cocreature: The first time yes but then `keys` is updated and the next time, you will run via `handleInactiveKey`.

remyhaemmerle-da:
@cocreature do you prefer this version ? 

> not a huge fan of this, you are factoring less than we did before.
I believe using only method of the form `handleXXX: Unit` make the API more consistent and more obvious to understand.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1213,90 +1201,96 @@ private[lf] object SBuiltin {
...
+        keyMapping: PartialTransaction.KeyMapping,
+    ): Unit =
+      keyMapping match {
+        case PartialTransaction.KeyActive(cid) => handleActiveKey(machine, cid)
+        case PartialTransaction.KeyInactive => handleInactiveKey(machine, gkey)
+      }
+  }
+
+  private[this] object KeyOperation {
+    final class Fetch(override val templateId: TypeConName) extends KeyOperation {
```
cocreature, line 1231: why does the operation need to store the template id?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1213,90 +1201,96 @@ private[lf] object SBuiltin {
     }
   }
 
-  private[speedy] sealed abstract class SBUKeyBuiltin(templateId: TypeConName)
-      extends OnLedgerBuiltin(1) {
+  private[this] abstract class KeyOperation {
+    val templateId: TypeConName
     private[this] val typ = AstUtil.TContractId(Ast.TTyCon(templateId))
```
cocreature, line 1206: where is this used?

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: in `cidImport`, just 2 lines under.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: `SEImport` used it in the case of `handleInputKeyFound`, required the type. 
The type is actually not used in the case of single contract ID, we could use a dummy (wrong !) type instead but I do not really like that.

PR #9573, opened 2021-05-04
KVL-921 Expose opentelemetry Context from Telemetry and TelemetryContext

hubert-da:
Add changes to make integration easier for consumers that are using the Open Telemetry API directly.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
+++ b/ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
@@ -6,6 +6,7 @@ package com.daml.telemetry
 import java.util.{Map => jMap}
 
 import io.opentelemetry.api.trace.{Span, Tracer}
+import io.opentelemetry.context.Context
 
 import scala.concurrent.Future
 
@@ -29,6 +30,11 @@ abstract class Telemetry(protected val tracer: Tracer) {
     * @see [[com.daml.telemetry.TelemetryContext.encodeMetadata()]]
     */
   def contextFromMetadata(metadata: Option[jMap[String, String]]): TelemetryContext
+
+  /** TODO
+    * @param context
+    */
+  def contextFromOpenTelemetryContext(context: Context): TelemetryContext
 
   /** Creates the first span of a new trace, and runs the computation inside it.
     *
@@ -85,6 +91,11 @@ abstract class DefaultTelemetry(override protected val tracer: Tracer) extends T
     }
   }
 
+  override def contextFromOpenTelemetryContext(context: Context): TelemetryContext =
+    Option(Span.fromContextOrNull(context))
+      .map(span => DefaultTelemetryContext(tracer, span))
+      .getOrElse(rootContext)
+
   override protected def rootContext: TelemetryContext = RootDefaultTelemetryContext(tracer)
 }
 
@@ -103,4 +114,7 @@ object NoOpTelemetry extends Telemetry(Tracer.getDefault) {
     NoOpTelemetryContext
 
   override protected def rootContext: TelemetryContext = NoOpTelemetryContext
+
+  override def contextFromOpenTelemetryContext(context: Context): TelemetryContext =
+    NoOpTelemetryContext
 }
--- a/ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
+++ b/ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
@@ -31,8 +31,11 @@ abstract class Telemetry(protected val tracer: Tracer) {
     */
   def contextFromMetadata(metadata: Option[jMap[String, String]]): TelemetryContext
 
-  /** TODO
-    * @param context
+  /** Allows to transform an Open Telemetry context into a telemetry context.
+    * Makes integration easier for consumers that are using the Open Telemetry API directly.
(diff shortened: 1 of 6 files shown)


ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
```diff
@@ -85,6 +91,11 @@ abstract class DefaultTelemetry(override protected val tracer: Tracer) extends T
     }
   }
 
+  override def contextFromOpenTelemetryContext(context: Context): TelemetryContext =
```
danilofaria, line 94: so how am i supposed to call this?

hubert-da: @danilofaria I would go with `DefaultTelemetry.contextFromOpenTelemetryContext(rawContext)`.

ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
```diff
@@ -30,6 +31,11 @@ abstract class Telemetry(protected val tracer: Tracer) {
     */
   def contextFromMetadata(metadata: Option[jMap[String, String]]): TelemetryContext
 
+  /** TODO
```
danilofaria, line 34: are you planning to write this doc in this PR?

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
```diff
@@ -77,6 +77,11 @@ trait TelemetryContext {
     */
   def encodeMetadata(): jMap[String, String]
 
+  /** TODO
```
danilofaria, line 80: same thing here

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
```diff
@@ -154,8 +159,10 @@ protected class DefaultTelemetryContext(protected val tracer: Tracer, protected
 
   override def encodeMetadata(): jMap[String, String] = {
     import scala.jdk.CollectionConverters._
-    Tracing.encodeTraceMetadata(Context.current.`with`(span)).asJava
+    Tracing.encodeTraceMetadata(openTelemetryContext).asJava
   }
+
+  override def openTelemetryContext: Context = Context.current.`with`(span)
```
danilofaria, line 165: what's the reasoning for using Context.current instead of Context.root?

ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
hubert-da: Yes, it's a draft and according to the description, I am going to add it.

ledger/metrics/src/main/scala/com/daml/telemetry/TelemetryContext.scala
hubert-da: There are a few reasons:
- according to the docs 
>It should generally not be required to use the root {@link Context} directly - instead, use {@link Context#current()} to operate on the current {@link Context}. Only use this method if you are absolutely sure you need to disregard the current {@link Context} - this almost always is only a workaround hiding an underlying context propagation issue.
- also as you can see above I'm just refactoring and preserving the current behavior
- if someone would like to use it within `Telemetry.runInOpenTelemetryScope`, it might not work as expected, as the current context would be different

ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
```diff
@@ -30,6 +31,14 @@ abstract class Telemetry(protected val tracer: Tracer) {
     */
   def contextFromMetadata(metadata: Option[jMap[String, String]]): TelemetryContext
 
+  /** Allows to transform an Open Telemetry context into a telemetry context.
+    * Makes integration easier for consumers that are using the Open Telemetry API directly.
+    * Consider using other factory methods.
+    *
+    * @param context raw Open Telemetry context
+    */
+  def contextFromOpenTelemetryContext(context: Context): TelemetryContext
```
fabiotudone-da, line 40: This is a bit shorter and it works too IMO, as the parameter type/name already contains the info:

```suggestion
  def contextFromOpenTelemetry(context: Context): TelemetryContext
```

hubert-da: @fabiotudone-da I wanted to keep it consistent with the previous ones like `contextFromGrpcThreadLocalContext`. Does it make sense or shall I rename it?

fabiotudone-da: Yes sure, it's just a minor thing.

fabiotudone-da (approved):

miklos-da (approved):
LGTM

ledger/metrics/src/main/scala/com/daml/telemetry/Telemetry.scala
miklos-da: Another option is to name this explicitly `adaptOpenTelemetryContext` (to avoid having to repeat 'context') and leave the other one constructing a context from gRPC thread-local context as-is.

PR #9587, opened 2021-05-05
LF: extends LF command with Fetch and Lookup 

remyhaemmerle-da:
This PR includes additionnaly :

- Adding ApiCommand to distinguish between generic command (that are
  accepted by the engine) and command that are accepted by the ledger
  API. 

- Reimplement Canton's reinterpret method using ~commands~ instead of
  node.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


remyhaemmerle-da:
cc @oliverse-da

oliverse-da:
Hi @remyhaemmerle-da , thank you.

I would like to test this with canton, ideally before you merge. Please let me know when you are happy with the changes, so I can try them via `daml-sdk-head`

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
```diff
@@ -134,58 +128,64 @@ private[lf] final class CommandPreprocessor(compiledPackages: CompiledPackages)
...
+            choiceArgument,
+          ) =>
+        unsafePreprocessCreateAndExercise(
+          templateId,
+          createArgument,
+          choiceId,
+          choiceArgument,
+        )
+      case command.Fetch(templateId, coid) =>
+        (speedy.Command.Fetch(templateId, SValue.SContractId(coid)), Set.empty)
```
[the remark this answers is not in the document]
cocreature: so I thought about this a bit more after realizing that we already had this before:

I still think this is wrong but I can’t come up with a way to exploit this since if you send a `Fetch`, you cannot create a contract so you will never have any local contracts so the freshness check breaks.

However, this only works somewhat by accident. If we had a version of `reinterpret` that accepts multiple root nodes. I could first run a fetch and then a create and you wouldn’t run the freshness check.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
```diff
@@ -134,58 +128,64 @@ private[lf] final class CommandPreprocessor(compiledPackages: CompiledPackages)
...
+          val (speedyCmd, newCids) = unsafePreprocessCommand(cmd)
+          go(rest, processed :+ speedyCmd, acc | newCids)
         case FrontStack() =>
-          processed.toImmArray
+          (processed.toImmArray, acc)
       }
     }
 
-    go(FrontStack(cmds), BackStack.empty) -> cids
+    go(FrontStack(cmds), BackStack.empty, Set.empty)
```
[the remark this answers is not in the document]
remyhaemmerle-da: before `cids` was modify imperatively and go return only the list of commands. 
I change to a purer approach where go return the list of commands together with the list of global CIDs. 
The `Set.empty` is the initialization of the CIds accumulator.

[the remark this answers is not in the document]
cocreature: oh I see, clearly didn’t take a close enough look

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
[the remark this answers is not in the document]
remyhaemmerle-da: changed.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
```diff
@@ -134,58 +128,64 @@ private[lf] final class CommandPreprocessor(compiledPackages: CompiledPackages)
     speedy.Command.LookupByKey(templateId, key)
   }
 
+  private[preprocessing] def unsafePreprocessCommand(
```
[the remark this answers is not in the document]
remyhaemmerle-da: done.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/command/Command.scala
```diff
@@ -73,8 +93,8 @@ final case class CreateAndExerciseCommand(
   *    interpretation will take this instant
   *  @param commandsReference id passed only for error reporting
   */
-case class Commands(
-    commands: ImmArray[Command],
+final case class Commands(
+    commands: ImmArray[ApiCommand],
```
remyhaemmerle-da, line 97: @gerolf-da be aware I change the API type here.

gerolf-da: Thanks for the ping. LGTM.

cocreature (approved):
Nice, thank you!

oliverse-da:
Ok, I will give this a try in a canton branch - let's identify the changes on the canton side, so we don't have an issue upgrading to the first 1.14 snapshot ;-)

cocreature:
Test failure just needs a fix in the test afaict, https://github.com/digital-asset/daml/blob/26a53d84698277c12b3e4c6458f65fcb229fad00/ledger/participant-integration-api/src/test/suite/scala/platform/apiserver/execution/StoreBackedCommandExecutorSpec.scala#L58

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -132,19 +132,19 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
     */
   def reinterpret(
       submitters: Set[Party],
```
oliverse-da, line 134: Canton currently uses GenActionNode.requiredAuthorizers as the submitters which we get as a side effect of building nodes which let us use the logic in `NodeInfo.requiredAuthorizers`. Do you see a way that we can avoid duplicating the `NodeInfo.requiredAuthorizers` mapping?:

- create: submitters := signatories
- exercise and fetch: submitters := actingParties
- lookup by key: submitters := maintainers

oliverse-da:
> Ok, I will give this a try in a canton branch - let's identify the changes on the canton side, so we don't have an issue upgrading to the first daml 1.14 snapshot release ;-)

Looking at the interface change it looks straightforward to generate `lf.Command`s instead of nodes. Unfortunately, I cannot actually test this yet, as it appears as if `daml-sdk-head` is broken with:

```
/Users/oliver/workspace/daml/dev-env/bin/daml-sdk-head: line 224: SCALA_VERSIONS[@]: unbound variable
FAILED TO INSTALL JARS!
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
cocreature: @oliverse-da I’m not following. `requiredAuthorizers` just picks the value from the node. If you can currently construct the node, why can’t you pass in the value you use when building the node as the `submitters`?

We can’t compute the authorizers from the command without LF interpretation and imho we shouldn’t do that here. The submitters should really be stored separately.

oliverse-da: we can continue contructing the node - I was only considering the option of canton turning the canton `ActionDescription` directly into `lf.command.Command` rather than first cooking up a `GenActionNode`.

oliverse-da:
With the tinyish changes on the canton side, we seem to be okay: https://github.com/DACH-NY/canton/pull/6024 , so good to merge from my side.

oliverse-da (approved):

remyhaemmerle-da:
After sleep over the PR, I think it is cleaner to distinguish `Command` from `Action` (and get rid of the involve inheritance hierarchy I introduce in the first commit):
- `CreateAndExercise` is not an action 
- `Fetch`s and `Lookup` are not a (Ledger API) command.

Does it looks OK for you @moritzkiefer-da ?

oliverse-da:
> After sleepping over the PR, I think it is cleaner to distinguish `Command` from `Action` (and get rid of the involve inheritance hierarchy I introduce in the first commit):
> 
> * `CreateAndExercise` is not an action
> * `Fetch`es and `Lookup` are not a (Ledger API) command.
> 
> See [e0cc578](https://github.com/digital-asset/daml/commit/e0cc578a130d8d18568d395233d86af73ebfb961)
> 
> Does it looks OK for you @cocreature ?

@remyhaemmerle-da , also please note @andreaslochbihler-da comment at https://github.com/DACH-NY/canton/pull/6024#discussion_r627397088

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/Action.scala
```diff
@@ -0,0 +1,80 @@
...
+package com.daml.lf
+
+import com.daml.lf.data.Ref._
+import com.daml.lf.value.Value
+
+sealed abstract class Action extends Product with Serializable {
+  val templateId: TypeConName
+}
+
+object Action {
```
cocreature, line 13: I find this a bit confusing. To me, a command is something I’m requesting to happen vs an action is something that has happened and is recorded in a transaction. That might sound like a small distinction but it is an important one. E.g., an action always has informees. A command might result in something that crashes while evaluating the signatories or for another reason and does not have informees.

So I’d probably keep the commands name.

remyhaemmerle-da: I am more than happy to discuss this point. 

I took the name `Action` from canton `ActionDescription` whish include in addition of my `Action` the node seed (for create and exercise) and the LF version. I could rename my `Action` to `ActionDescription` and include the seed. Note sure if we really need the version wish is used only to rebuild the Node. 

I would be happy to get opinion of @andreaslochbihler-da.

cocreature: Rather than looking at Canton’s implementation let’s look at the ledger model: In there an action always has informees. This is not the case for a command, it can fail to produce a list of signatories, stakeholders or for a numerous other reasons.

I’d prefer to stick to `Command` as being something that we feed into speedy to produce a transaction whereas an action is a Node in a (partial) transaction which therefore always has informees and other stuff.

remyhaemmerle-da: OK. I reverted the last commit.

cocreature (approved):
Thank you!

PR #9588, opened 2021-05-05
Ledger API Server: Named threadpools

gerolf-da:
Some cached threadpools weren't given names, meaning at runtime there
are a bunch of pool-x-thread-y threads. This makes it hard to understand
which threads are being used for what.

The following pool names were introduced:

append-only indexer: input-mapping-pool, batching-pool
ProgramResource: program-resource-pool
kvutils PackageCommitter: package-preload-executor

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
@@ -342,6 +342,7 @@ final private[kvutils] class PackageCommitter(
     Executors.newSingleThreadExecutor { (runnable: Runnable) =>
       val t = new Thread(runnable)
       t.setDaemon(true)
+      t.setName("package-preload-executor")
       t
     }
 
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
@@ -340,9 +340,8 @@ final private[kvutils] class PackageCommitter(
 
   private lazy val preloadExecutor =
     Executors.newSingleThreadExecutor { (runnable: Runnable) =>
-      val t = new Thread(runnable)
+      val t = new Thread(runnable, "package-preload-executor")
       t.setDaemon(true)
-      t.setName("package-preload-executor")
       t
     }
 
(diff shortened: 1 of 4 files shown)


mziolekda (approved):
Beutiful

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
```diff
@@ -342,6 +342,7 @@ final private[kvutils] class PackageCommitter(
     Executors.newSingleThreadExecutor { (runnable: Runnable) =>
       val t = new Thread(runnable)
```
miklos-da, line 343: You can simply [use the constructor](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#Thread-java.lang.Runnable-java.lang.String-) for this:
```suggestion
      val t = new Thread(runnable, "package-preload-executor")
```

miklos-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
gerolf-da: Fixed in 7384e15092046d667e0635194de6110af2a0fa5a.

nmarton-da (approved):
Great! Thanks :bow:

SamirTalwar-DA (approved):

libs-scala/resources/src/main/scala/com/digitalasset/resources/ProgramResource.scala
```diff
@@ -20,7 +21,12 @@ final class ProgramResource[Context: HasExecutionContext, T](
 ) {
   private val logger = ContextualizedLogger.get(getClass)
 
-  private val executorService = Executors.newCachedThreadPool()
+  private val executorService = {
+    val counter = new AtomicLong(0L)
+    Executors.newCachedThreadPool((runnable: Runnable) =>
+      new Thread(runnable, s"program-resource-pool-${counter.incrementAndGet()}")
+    )
```
SamirTalwar-DA, line 28: Why not use a `ThreadFactoryBuilder` here?

gerolf-da: Because I would have to add the dependency to google-commons to the `libs-scala/resources` library. And considering that this is a small utility library, I'd rather keep the dependency surface small. But maybe this actually asks for extracting out [NamedThreadFactory](https://github.com/digital-asset/daml/blob/main/runtime-components/non-repudiation-resources/src/main/scala/com/daml/nonrepudiation/resources/HikariTransactorResourceOwner.scala#L30) into a common place.

SamirTalwar-DA: Perhaps leave it for now and let's see if we use this a second time. Didn't realize the builder was part of Guava.

fabiotudone-da (approved):

PR #9592, opened 2021-05-06
Refactor out setGlobalLogLevel into ContextualLogging

realvictorprm:
CHANGELOG_BEGIN

- [Ledger HTTP Json Service] Logging now also tells service name if log level was changed.

CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
+++ b/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
@@ -1,0 +1,37 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.cliopts
+
+import ch.qos.logback.classic.{Level => LogLevel}
+
+object Logging {
+
+  private val KnownLogLevels = Set("ERROR", "WARN", "INFO", "DEBUG", "TRACE")
+
+  private implicit val scoptLogLevel: scopt.Read[LogLevel] = scopt.Read.reads { level =>
+    Either
+      .cond(
+        KnownLogLevels.contains(level.toUpperCase),
+        LogLevel.toLevel(level.toUpperCase),
+        s"Unrecognized logging level $level",
+      )
+      .getOrElse(throw new java.lang.IllegalArgumentException(s"Unknown logging level $level"))
+  }
+
+  /** Parse in the cli option for the logging level.
+    */
+  def loggingLevelParse[C](
+      parser: scopt.OptionParser[C]
+  )(logLevel: Setter[C, Option[LogLevel]]): Unit = {
+    import parser.opt
+
+    opt[LogLevel]("log-level")
+      .optional()
+      .action((level, c) => logLevel(_ => Some(level), c))
+      .text(
+        s"Default logging level to use. Available values are ${KnownLogLevels.mkString(", ")}. Defaults to INFO."
+      )
+    ()
+  }
+}
--- a/ledger-service/cli-opts/BUILD.bazel
+++ b/ledger-service/cli-opts/BUILD.bazel
@@ -19,6 +19,9 @@ da_scala_library(
     visibility = ["//visibility:public"],
     deps = [
         "//ledger/ledger-api-common",
+        "@maven//:ch_qos_logback_logback_classic",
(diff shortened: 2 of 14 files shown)


ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -0,0 +1,38 @@
...
+    import parser.opt
+
+    val _ =
+      opt[LogLevel]("log-level")
+        .optional()
+        .action((level, c) => logLevel(_ => Some(level), c))
+        .text(
+          s"Default logging level to use. Available values are ${KnownLogLevels.mkString(", ")}. Defaults to INFO."
+        )
+    ()
```
realvictorprm, line 36: @S11001001 did apply the changes you requested here

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package com.daml.cliopts
+
+import ch.qos.logback.classic.{Level => LogLevel}
+
+object Logging {
+
+  private val KnownLogLevels = Set("ERROR", "WARN", "INFO", "DEBUG", "TRACE")
+
+  implicit val scoptLogLevel: scopt.Read[LogLevel] = scopt.Read.reads { level =>
```
cocreature, line 12:
```suggestion
  private implicit val scoptLogLevel: scopt.Read[LogLevel] = scopt.Read.reads { level =>
```

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/GlobalLogLevel.scala
```diff
@@ -1,20 +1,20 @@
 // Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package com.daml.platform.sandbox
+package com.daml.logging
```
cocreature, line 4: I’d probably move this In the CLI opts library instead. You really want to use this together with the CLI parser whereas the contextualized logger can be used perfectly fine without either of these and the other way around, you don’t need the contextualized logger for this to be useful.

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  /** Parse in the cli option for the logging level.
+    */
+  def loggingLevelParse[C](
+      parser: scopt.OptionParser[C]
+  )(logLevel: Setter[C, Option[LogLevel]]): Unit = {
+    import parser.opt
+
+    val _ =
+      opt[LogLevel]("log-level")
```
cocreature, line 30:
```suggestion
    opt[LogLevel]("log-level")
```
While I’m usually a fan of `NonUnitStatement` warning and avoiding those here, `scopt` is one of those libraries where it’s just pointless to try since they are clearly not designed for this so I’d rather be consistent and ignore it everywhere instead of making this the odd one out.

cocreature (approved):
Nice, thank you!

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/GlobalLogLevel.scala
victormueller-da: Valid point, will do that

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
realvictorprm: I was required to do that there if I don't want to suppress the warning.

realvictorprm: ok bit weird how it works but it seems like placing the `()` at the end of the method is enough for scalac not complaining about discarding non unit results.

Anyways, resolved.

S11001001: > placing the `()` at the end of the method is enough for scalac not complaining about discarding non unit results

This suppresses the `value-discard` warning built into scalac, but not `NonUnitStatements`, which is turned on for `http-json-*` but not `cli-opts`.

PR #9597, opened 2021-05-06
Drop todo for conversion of exception primitives

cocreature:
While there might be nicer ways to do this, the current way is
consistent what we do for example with `create` which seems pretty
comparable and the check that `ty1` is an exception type happens in
the LF typechecker (again consistent with how we check for create that
the type is a template type) so this seems fine.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Thanks :-)

PR #9604, opened 2021-05-06
check whether collection.compat is unused when compiling for Scala 2.12

S11001001:
Instead of always suppressing warnings for `collection.compat._`, we should only do it for Scala 2.13.

We also reduce boilerplate by automatically adding this option when both silencer_plugin and collection-compat are present.  There are a few other possible approaches, such as adding silencer_plugin when collection-compat is present, or having a special `collection_compat = True` option to `da_scala_library` et al; this one just happens to be the smallest change that eliminates the flags.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice cleanup, thank you!

stefanobaghino-da (approved):

PR #9607, opened 2021-05-06
Add duplicate contract key checks to Speedy

cocreature:
This depends on #9610 and #9608 to get through CI but for review you should be able to ignore that.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -9,6 +9,7 @@ import com.daml.lf.data.{BackStack, ImmArray, Ref, Time}
 import com.daml.lf.language.Ast
 import com.daml.lf.ledger.{Authorize, FailedAuthorization}
 import com.daml.lf.transaction.{
+  ContractKeyUniquenessMode,
   GenTransaction,
   GlobalKey,
   Node,
@@ -23,6 +24,12 @@ import scala.annotation.nowarn
 import scala.collection.immutable.HashMap
 
 private[lf] object PartialTransaction {
+
+  sealed trait KeyConflict extends Product with Serializable
+  object KeyConflict {
+    final case object None extends KeyConflict
+    final case object Duplicate extends KeyConflict
+  }
 
   type NodeIdx = Value.NodeIdx
   type Node = Node.GenNode[NodeId, Value.ContractId]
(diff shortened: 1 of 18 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -354,26 +364,67 @@ private[lf] case class PartialTransaction(
...
+                case None => (ptx.globalKeyInputs.updated(ck, KeyInactive), KeyConflict.None)
+                // This is case 1.4.
+                case Some(KeyInactive) =>
+                  (ptx.globalKeyInputs.updated(ck, KeyInactive), KeyConflict.None)
+                // This is case 1.5
+                case Some(KeyActive(_)) =>
+                  (ptx.globalKeyInputs.updated(ck, KeyInactive), KeyConflict.Duplicate)
+              }
+
+          }
```
remyhaemmerle-da, line 413: I am wondering if it would not be simpler to split computation of `conflict` from the update of `globalKeyInputs` with something like that: 

```suggestion
          val conflict =  keys.get(ck).orElse(ptx.globalKeyInputs(ck) match {
             case Some(KeyActive(_)) => KeyConflict.Duplicate
             case Some(Inactive) | None => KeyConflict.None
          }
          val globalKeyInputs = keys.get(ck) match {
            case None => ptx.globalKeyInputs.updated(ck, KeyInactive)
            case Some(_) => ptx.globalKeyInputs
          }
```

cocreature: I’ve tried that first but it seemed more confusing and easier to get out of sync. Enumerating all the possible cases explicitly and doing so only once seems like an easy way to not mess it up.

cocreature: In particular, this way there is a very clear correspondence between the pattern matches and the doc comment which explains the details and the reasoning. At least to me that correspondence is pretty useful.

remyhaemmerle-da (approved):
LGTM.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: Switched over to your suggestion (with some minor tweaks & fixes) which also fixed a bug. Thanks for the suggestion and for catching this!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -766,6 +766,26 @@ object Transaction {
...
+    * if both contracts are used in the transaction in by-key operations
+    * meaning lookup, fetch or exercise-by-key or local creates.
+    *
+    * Two notable cases that will never produce duplicate key errors
+    * is a standalone create or a create and a fetch (but not fetch-by-key)
+    * with the same key.
+    *
+    * For ledger implementors this means that (for contract key uniqueness)
+    * it is sufficient to only look at the inputs and the outputs of the
+    * transaction whlie leaving all internal checks within the transaction
```
miklos-da, line 782: `transaction, while leaving ...`

PR #9608, opened 2021-05-06
Unwind partial transaction on mustFail

cocreature:
This PR fixes a very confusing and subtle bug in the handling of
submitMustFail in scenarios:

submitMustFail did not properly unwind the transaction. Afaict, this
only matters for contract id derivation in this context.
It also only matters for errors during interpretation not errors that
happen when committing to the ledger (global duplicate key errors).

Specifically what happens is that we end up leaking the exercise
context that we were in when we hit the failure into the next
submission. I’m not sure if this actually breaks anything (if it does,
doesn’t really matter, it’s scenarios) but it is definitely not what
should be happening.

I hit this when adding uck checks to the engine since that turns some
commit errors into interpretation errors and changed contract ids in
tests that really shouldn’t have changed.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


nickchapman-da (approved):
This looks entirely sensible.

PR #9618, opened 2021-05-07
Fix #7170

### LINKED ISSUE #7170: Review and extend logging in the JSON API
As part of making the JSON API more usable in a production setup we should review what and how we currently log and extend it as needed. A few things that come to mind:

1. Failed command and possibly also succeeding commands at some level of detail. Probably at least including the command id.
2. Token decoding failures (not validation failures, that is just a failed command to the JSON API).
3. Verify that we do not log secrets (primarily JWTs)

I’m sure there are many more things but I expect most of this can also be extended over time.

realvictorprm:
### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -51,45 +56,80 @@ class Endpoints(
...
+    case req @ HttpRequest(GET, Uri.Path("/livez"), _, _, _) =>
+      mkRequestLoggingContext(req)(implicit lc => {
+        logger.trace("Incoming Request")
+        Future.successful(HttpResponse(status = StatusCodes.OK))
+      })
+    case req @ HttpRequest(GET, Uri.Path("/readyz"), _, _, _) =>
+      mkRequestLoggingContext(req)(implicit lc => {
+        logger.trace("Incoming Request")
+        healthService.ready().map(_.toHttpResponse)
+      })
```
realvictorprm, line 126: I know, many duplicates here. However I decided to do this so in the parent code calling this code no concrete request needs to be handed down as otherwise this would result in composing the functions for each incoming request (at least if I understood the code in `HttpService` correctly).

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
```diff
@@ -56,19 +57,21 @@ private class ContractsFetch(
...
     getCreatesAndArchivesSince: LedgerClientJwt.GetCreatesAndArchivesSince,
     getTermination: LedgerClientJwt.GetTermination,
-)(implicit dblog: doobie.LogHandler, sjd: SupportedJdbcDriver)
-    extends StrictLogging {
+)(implicit dblog: doobie.LogHandler, sjd: SupportedJdbcDriver) {
 
   import ContractsFetch._
   import sjd.retrySqlStates
 
+  private[this] def logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 65:
```suggestion
  private[this] val logger = ContextualizedLogger.get(getClass)
```
Any reason to not make it a `val`?

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -308,8 +339,14 @@ class ContractsService(
       parties: OneAnd[Set, domain.Party],
       templateIds: Set[domain.TemplateId.RequiredPkg],
       queryParams: InMemoryQuery,
+  )(implicit
+      lc: LoggingContextOf[CorrelationID with RequestID]
   ): Source[Error \/ domain.ActiveContract[LfValue], NotUsed] = {
 
+    logger.debug(
+      s"Searching in memory, parties: $parties, templateIds: $templateIds, queryParms: $queryParams"
+    )
```
stefanobaghino-da, line 348: Should parties and template identifiers be part of the logging context, rather than logged in the message? Putting the query parameters could also be reasonable but a bit challenging, because the logging context right now doesn't allow to nest objects.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -51,45 +52,64 @@ class Endpoints(
     encoder: DomainJsonEncoder,
     decoder: DomainJsonDecoder,
     maxTimeToCollectRequest: FiniteDuration = FiniteDuration(5, "seconds"),
-)(implicit ec: ExecutionContext, mat: Materializer)
-    extends StrictLogging {
+)(implicit ec: ExecutionContext, mat: Materializer) {
+
+  private[this] def logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 57:
```suggestion
  private[this] val logger = ContextualizedLogger.get(getClass)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -4,46 +4,51 @@
...
+import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-object Main extends StrictLogging {
+object Main {
+
+  private[this] def logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 28:
```suggestion
  private[this] val logger = ContextualizedLogger.get(getClass)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -8,17 +8,19 @@ import com.daml.lf.data.Ref
...
 import scalaz._
 
 import scala.collection.compat._
 import scala.concurrent.{ExecutionContext, Future}
 
-private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadPackageStore)
-    extends StrictLogging {
+private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadPackageStore) {
+
+  private[this] def logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 23:
```suggestion
  private[this] val logger = ContextualizedLogger.get(getClass)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebsocketEndpoints.scala
```diff
@@ -48,45 +49,53 @@ object WebsocketEndpoints {
 class WebsocketEndpoints(
     decodeJwt: ValidateJwt,
     webSocketService: WebSocketService,
-) extends StrictLogging {
+) {
 
   import WebsocketEndpoints._
 
-  lazy val transactionWebSocket: PartialFunction[HttpRequest, Future[HttpResponse]] = {
+  private[this] def logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 56:
```suggestion
  private[this] val logger = ContextualizedLogger.get(getClass)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Logging.scala
```diff
@@ -0,0 +1,28 @@
...
+  abstract case class RequestID()
+
+  def correlationIdLogCtx[Z](fn: LoggingContextOf[CorrelationID] => Z): Z =
+    newLoggingContext(label[CorrelationID], Map("CorrelationID" -> UUID.randomUUID().toString))(fn)
+
+  def extendWithRequestIdLogCtx[Z](
+      fn: LoggingContextOf[CorrelationID with RequestID] => Z
+  )(implicit lc: LoggingContextOf[CorrelationID]): Z =
+    withEnrichedLoggingContext(label[RequestID], Map("RequestID" -> UUID.randomUUID().toString))
+      .run(fn)
```
stefanobaghino-da, line 23: Other components in the stack use `snake_case` in the logging context, we might want to stick to that.
```suggestion
  def correlationIdLogCtx[Z](fn: LoggingContextOf[CorrelationID] => Z): Z =
    newLoggingContext(label[CorrelationID], Map("correlation_id" -> UUID.randomUUID().toString))(fn)

  def extendWithRequestIdLogCtx[Z](
      fn: LoggingContextOf[CorrelationID with RequestID] => Z
  )(implicit lc: LoggingContextOf[CorrelationID]): Z =
    withEnrichedLoggingContext(label[RequestID], Map("request_id" -> UUID.randomUUID().toString))
      .run(fn)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -4,46 +4,51 @@
...
-      ContractDao.supportedJdbcDriverNames(JdbcDrivers.availableJdbcDriverNames),
-    ) match {
-      case Some(config) =>
-        main(config)
-      case None =>
-        // error is printed out by scopt
-        sys.exit(ErrorCodes.InvalidUsage)
-    }
+  def main(args: Array[String]): Unit = {
+    correlationIdLogCtx(implicit lc =>
```
stefanobaghino-da, line 37: What is exactly the correlation identifier? Does it identify exactly an individual HTTP JSON API server process?

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
realvictorprm: This is exactly what I was wondering too a couple hours ago. Thanks for picking this one up, you're right!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: Yes it should, it's not using the process id but a randomly generated UUID which I hope fulfills the requirement in 99% of the time.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
realvictorprm: Well this one was kind of a suggestion what could be logged additionally.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -454,6 +497,8 @@ object ContractsService {
         override def findByContractId(
             ctx: SearchContext[Option, Option],
             contractId: domain.ContractId,
+        )(implicit
+            lc: LoggingContextOf[CorrelationID with RequestID]
```
S11001001, line 501: Something to think about when declaring types of your `lc` implicit parameters: is there something essential about `findByContractId` that says "this must be called in the scope of a request"? If not, the intersection should not mention `RequestID`, and the variable should be `LoggingContextOf[Any]` in the extreme. A counterexample would be the functions in `Endpoints`, for which I _would_ intrinsically expect a `RequestID` to be present.

In general, I would expect that the deeper into the callstack you are, the wider the type parameter of `LoggingContextOf` is; that is the pattern that implicit resolution supports without you having to do anything special. So at the outset of a request you are more likely to see `LoggingContextOf[CorrelationID with RequestID]`, and ultimate utility-only functions take only `LoggingContextOf[Any]`, if anything.

S11001001: Triggers provide an example of this; `LoggingContextOf[Config with Trigger]` is requested by _trigger service_ methods, but it calls out to the lower-level trigger _runner_, which only requests `LoggingContextOf[Trigger]` (because service configuration is not a relevant concept to the runner).

realvictorprm: You're right about this. I didn't think about using Any yet although I already thought that it is not practical to enforce `CorrelationID with RequestID` everywhere.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -51,45 +52,64 @@ class Endpoints(
...
+            extendWithRequestIdLogCtx(implicit lc =>
+      downloadPackage(req, packageId))
     // format: on
     case req @ HttpRequest(POST, Uri.Path("/v1/packages"), _, _, _) =>
-      httpResponse(uploadDarFile(req))
+      extendWithRequestIdLogCtx(implicit lc => httpResponse(uploadDarFile(req)))
     case HttpRequest(GET, Uri.Path("/livez"), _, _, _) =>
-      Future.successful(HttpResponse(status = StatusCodes.OK))
+      extendWithRequestIdLogCtx(implicit lc => {
+        logger.trace("Incoming Request")
```
cocreature, line 99: this log is not very useful. We emit the exact same log message for both `livez` and `ready` which is both weirdly specific (why not for any other request) and not specific enough (we don’t know which of the two).

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Logging.scala
```diff
@@ -0,0 +1,28 @@
...
+package com.daml.http.util
+
+import com.daml.logging.LoggingContextOf
+import com.daml.logging.LoggingContextOf.{label, newLoggingContext, withEnrichedLoggingContext}
+
+import java.util.UUID
+
+object Logging {
+
+  abstract case class CorrelationID()
```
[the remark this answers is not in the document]
realvictorprm: For me the names were quite selfdescribing, however if they're not I can add documentation and if necessary also rename them to something which fits better :thinking:

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: part of the logs I added here are just there so we have at least one in there but with means to extend it after this PR. I probably should add for these a couple TODO's.

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -68,7 +70,11 @@ object HttpService extends StrictLogging {
       mat: Materializer,
       aesf: ExecutionSequencerFactory,
       ec: ExecutionContext,
+      lc: LoggingContextOf[CorrelationID with RequestID],
   ): Future[Error \/ ServerBinding] = {
+
+    logger.info("HTTP Server pre-startup (contains own RequestID for all calls)")
```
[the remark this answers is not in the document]
realvictorprm: Thanks to @S11001001 I've got a solution to that problem. I'll remove the requestid at places where it's not absolutely necessary.

daml-script/test/src/com/digitalasset/daml/lf/engine/script/test/JsonApiIt.scala
```diff
@@ -139,6 +140,9 @@ trait JsonApiFixture
         channel <- GrpcClientResource.owner(server.port)
         httpService <- new ResourceOwner[ServerBinding] {
           override def acquire()(implicit context: ResourceContext): Resource[ServerBinding] = {
+            implicit val lc: LoggingContextOf[CorrelationID with RequestID] = newFullCtx(
```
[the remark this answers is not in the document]
realvictorprm: I wasn't sure yet what role it plays in the test. Looking into this now more.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
S11001001: You can eliminate the duplication without changing the type of `all` or the semantics you have defined.

```scala
import scalaz.std.partialFunction._, scalaz.syntax.arrow._

    lazy val all: PartialFunction[HttpRequest, Future[HttpResponse]] = {
      val dispatch: PartialFunction[HttpRequest, LoggingContextOf[...] => Future[HttpResponse]] = {
          case req @ HttpRequest(POST, Uri.Path("/v1/create"), _, _, _) =>
            implicit lc => httpResponse(create(req))
         // other cases
      }
      (dispatch &&& { case req => req }) andThen {
        case (lcFhr, req) => mkRequestLoggingContext(req)(lcFhr)
      }
    }
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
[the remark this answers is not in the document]
realvictorprm: that one fixed

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -92,9 +97,11 @@ object Main extends StrictLogging {
...
 
     val serviceF: Future[HttpService.Error \/ ServerBinding] =
-      HttpService.start(
-        startSettings = config,
-        contractDao = contractDao,
+      extendWithRequestIdLogCtx(implicit lc =>
+        HttpService.start(
+          startSettings = config,
+          contractDao = contractDao,
+        )
```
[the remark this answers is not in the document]
realvictorprm: RequestID's are now not mandatory for the startup anymore, only the `CorrelationID`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Logging.scala
```diff
@@ -0,0 +1,28 @@
...
+  def correlationIdLogCtx[Z](fn: LoggingContextOf[CorrelationID] => Z): Z =
+    newLoggingContext(label[CorrelationID], Map("correlation_id" -> UUID.randomUUID().toString))(fn)
+
+  def extendWithRequestIdLogCtx[Z](
+      fn: LoggingContextOf[CorrelationID with RequestID] => Z
+  )(implicit lc: LoggingContextOf[CorrelationID]): Z =
+    withEnrichedLoggingContext(label[RequestID], Map("request_id" -> UUID.randomUUID().toString))
+      .run(fn)
+
+  def newFullCtx[Z](fn: LoggingContextOf[CorrelationID with RequestID] => Z): Z =
```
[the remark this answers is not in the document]
realvictorprm: removed that helper function because it was not needed anymore

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Logging.scala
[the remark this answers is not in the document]
realvictorprm: Added extensive documentation

daml-script/test/src/com/digitalasset/daml/lf/engine/script/test/JsonApiIt.scala
```diff
@@ -139,6 +140,9 @@ trait JsonApiFixture
         channel <- GrpcClientResource.owner(server.port)
         httpService <- new ResourceOwner[ServerBinding] {
           override def acquire()(implicit context: ResourceContext): Resource[ServerBinding] = {
+            implicit val lc: LoggingContextOf[CorrelationID] = correlationIdLogCtx(
+              identity(_)
+            )
```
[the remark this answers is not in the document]
realvictorprm: done

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -176,12 +191,16 @@ class ContractsService(
...
+  )(implicit
+      lc: LoggingContextOf[CorrelationID with Any]
   ): SearchResult[Error \/ domain.ActiveContract[LfValue]] =
     retrieveAll(jwt, jwtPayload.parties)
 
   def retrieveAll(
       jwt: Jwt,
       parties: OneAnd[Set, domain.Party],
+  )(implicit
+      lc: LoggingContextOf[CorrelationID with Any]
```
cocreature, line 203: Isn’t that equivalent to just `CorrelationID`?

realvictorprm: yep you're right. I'll remove that crap, learned something again about Scala.

PR #9619, opened 2021-05-07
fix flaky test in RecoveringIndexer

mziolekda:
Move the log "Restarted Indexer Server" up to ascertain that this log action happens before re-subscription action and race is avoided.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/RecoveringIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/RecoveringIndexer.scala
@@ -83,12 +83,10 @@ private[indexer] final class RecoveringIndexer(
             logger.info("Restarting Indexer Server")
             val newSubscription = subscribe()
             if (subscription.compareAndSet(oldSubscription, newSubscription)) {
-              resubscribeOnFailure(
-                newSubscription, {
-                  updateHealthStatus(HealthStatus.healthy)
-                  logger.info("Restarted Indexer Server")
-                },
-              )
+              resubscribeOnFailure(newSubscription) {
+                updateHealthStatus(HealthStatus.healthy)
+                logger.info("Restarted Indexer Server")
+              }
               Future.unit
             } else { // we must have stopped the server during the restart
               logger.info("Indexer Server was stopped; cancelling the restart")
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/RecoveringIndexer.scala
```diff
@@ -79,10 +79,11 @@ private[indexer] final class RecoveringIndexer(
             logger.info("Restarting Indexer Server")
             val newSubscription = subscribe()
             if (subscription.compareAndSet(oldSubscription, newSubscription)) {
-              resubscribeOnFailure(newSubscription)
               newSubscription.asFuture.map { _ =>
                 logger.info("Restarted Indexer Server")
               }
+              resubscribeOnFailure(newSubscription)
```
SamirTalwar-DA, line 85: These two things will run in parallel, so I don't think this will help flakiness.

SamirTalwar-DA: Only if you use `parasitic` (or `DirectExecutionContext`) everywhere!

ledger/participant-integration-api/src/main/scala/platform/indexer/RecoveringIndexer.scala
```diff
@@ -97,9 +95,13 @@ private[indexer] final class RecoveringIndexer(
         }
       } yield ()
 
-    def resubscribeOnFailure(currentSubscription: Resource[IndexFeedHandle]): Unit =
+    def resubscribeOnFailure(
+        currentSubscription: Resource[IndexFeedHandle],
+        actOnSuccess: => Unit,
+    ): Unit =
```
SamirTalwar-DA, line 101: I would prefer if this used a separate parameter list for the function and was called with `{ … }` to show that it's a lazily-evaluated function.

```suggestion
    def resubscribeOnFailure(
        currentSubscription: Resource[IndexFeedHandle]
    )(actOnSuccess: => Unit): Unit =
```

SamirTalwar-DA (approved):

PR #9622, opened 2021-05-07
Patch perf reference Bazel derivation

aherrmann-da:
The patch that it downloads via Nix was taken from a GH PR instead of a commit such that the hash is not fully stable. This adds a patch to download the relevant patch directly from a GitHub commit.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/perf/BazelExclusiveSandboxing.patch
+++ b/ci/cron/perf/BazelExclusiveSandboxing.patch
@@ -1,0 +1,15 @@
+diff --git a/nix/nixpkgs.nix b/nix/nixpkgs.nix
+index 8c7eb189783..e0b7e1e5534 100644
+--- a/nix/nixpkgs.nix
++++ b/nix/nixpkgs.nix
+@@ -38,8 +38,8 @@ let
+         # for the only change that actually affects the code in this patch. The rest is tests
+         # and/or documentation.
+         (pkgs.fetchurl {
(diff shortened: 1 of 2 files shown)


aherrmann-da:
Test per run succeeded: https://dev.azure.com/digitalasset/daml/_build/results?buildId=76709&view=logs&j=c064121a-8b6f-58e0-c599-66888bd1070e&t=73a501b2-8e0c-50c2-3927-aa6d10783055

garyverhaegen-da:
Are we confident this will solve the issue? As far as I can tell the PR itself has had no recent changes at all, and the new hash is the one we get from the PR as well at this point, so I'm wondering if this may be GitHub changing something in the way they format their diffs rather than being linked to the PR vs. commit question.

I don't mean to block this by any means as it does update the hash, I'm just not sure the change in URL achieves any more future-proofing than _just_ updating the hash would.

cocreature:
> Are we confident this will solve the issue? As far as I can tell the PR itself has had no recent changes at all, so I'm wondering if this may be GitHub changing something in the way they format their diffs rather than being linked to the PR vs. commit question.

I don’t know but at worst it’s as bad as the current state. Given that this is the first time after a few months that we hit this, I think this is a sensibel solution. If it happens again, I’d either drop the patch completely or drop the whole build of the old commit.

PR #9624, opened 2021-05-08
sandbox-on-x: Support `beginAfter` in state updates.

SamirTalwar-DA:
Because @kamil-da needs it.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
+++ b/ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
@@ -115,17 +115,16 @@ case class ReadWriteServiceBridge(
     //   no beginAfter supported
     //   neither multiple subscriptions
     //   neither bootstrapping the bridge from indexer persistence
-    assert(
-      beginAfter.isEmpty,
-      "Re-subscribing not supported. Only supported to subscribe once, and from inception.",
-    )
     synchronized {
       if (stateUpdatesWasCalledAlready)
         throw new IllegalStateException("not allowed to call this twice")
       else stateUpdatesWasCalledAlready = true
     }
     logger.info("Indexer subscribed to state updates.")
-    queueSource
+    beginAfter match {
+      case None => queueSource
+      case Some(offset) => queueSource.dropWhile(_._1 <= offset)
+    }
   }
 
   val (queue: BoundedSourceQueue[Submission], queueSource: Source[(Offset, Update), NotUsed]) =
--- a/ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
+++ b/ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
@@ -112,9 +112,9 @@ case class ReadWriteServiceBridge(
   var stateUpdatesWasCalledAlready = false
   override def stateUpdates(beginAfter: Option[Offset]): Source[(Offset, Update), NotUsed] = {
     // TODO for PoC purposes:
-    //   no beginAfter supported
-    //   neither multiple subscriptions
-    //   neither bootstrapping the bridge from indexer persistence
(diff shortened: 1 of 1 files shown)


ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
```diff
@@ -115,17 +115,16 @@ case class ReadWriteServiceBridge(
...
-      "Re-subscribing not supported. Only supported to subscribe once, and from inception.",
-    )
     synchronized {
       if (stateUpdatesWasCalledAlready)
         throw new IllegalStateException("not allowed to call this twice")
       else stateUpdatesWasCalledAlready = true
     }
     logger.info("Indexer subscribed to state updates.")
-    queueSource
+    beginAfter match {
```
nmarton-da, line 124: what is this good for...can you describe in which scenario this would help?

nmarton-da (commented):
not sure this is a way forward, except it helps on some use case temporaliry

ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
SamirTalwar-DA: I suggest you ask @kamil-da. I'm just the implementor. :upside_down_face:

kamil-da: This change allows to run the `sandbox-on-x` on a database created with a Postgres export.
I needed that for testing the `ledger-api-bench-tool`. I would like to quickly prepare a running ledger with a significant amount of transactions. To do that I use existing ledger database exports.

SamirTalwar-DA: I'd also like to add that I don't think the previous implementation (of not conforming to the API) necessarily had a good justification. So I'm not sure why this would be _worse_.

kamil-da: Thank you @nmarton-da for providing the context. 👍 
For my current work it would be sufficient just to use the patch applied locally but if you foresee value in having this change in the codebase, then I would go with the second option that @nmarton-da described - an explicit configuration parameter allowing starts with a non-empty database.

mziolekda: I agree with @nmarton-da opinion that a filter sends a misleading signal. I would be fine with suggestion 1 and a disclaimer in the comments. This code is only there to facilitate our own testing at the moment.

ledger/sandbox-on-x/src/main/scala/com/daml/ledger/sandbox/ReadWriteServiceBridge.scala
```diff
@@ -115,17 +115,16 @@ case class ReadWriteServiceBridge(
     //   no beginAfter supported
     //   neither multiple subscriptions
     //   neither bootstrapping the bridge from indexer persistence
```
SamirTalwar-DA, line 117:
```suggestion
    //   This method may only be called once, either with `beginAfter` set or unset.
    //   A second call will result in an error unless the server is restarted.
    //   Bootstrapping the bridge from indexer persistence is not supported either.
```

SamirTalwar-DA: @nmarton-da, how's this?

nmarton-da: first part works, the last part is actually works now, so I would do `Bootstrapping the bridge from indexer persistence is supported.`

tudor-da (approved):
LGTM

PR #9627, opened 2021-05-10
Remove version field from rollback node.

nickchapman-da:
Remove version field from rollback node
- fix encode/decode
- adapt testcases etc

part of #8020 

Review tip for  `TransactionCoder.scala` -- hide whitespace diffs

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
@@ -109,7 +109,7 @@ final class ValueEnricher(engine: Engine) {
 
   def enrichNode[Nid](node: GenNode[Nid, ContractId]): Result[GenNode[Nid, ContractId]] =
     node match {
-      case rb @ Node.NodeRollback(_) => // reconsider if fields added
+      case rb @ Node.NodeRollback(_) =>
         ResultDone(rb)
       case create: Node.NodeCreate[ContractId] =>
         for {
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
@@ -109,7 +109,7 @@ final class ValueEnricher(engine: Engine) {
 
   def enrichNode[Nid](node: GenNode[Nid, ContractId]): Result[GenNode[Nid, ContractId]] =
     node match {
-      case rb @ Node.NodeRollback(_, _) => // reconsider if fields added
+      case rb @ Node.NodeRollback(_) => // reconsider if fields added
         ResultDone(rb)
       case create: Node.NodeCreate[ContractId] =>
         for {
--- a/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
+++ b/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
@@ -359,12 +359,11 @@ object ValueGenerators {
   /** Makes rollback node with some random child IDs. */
   val danglingRefRollbackNodeGen: Gen[NodeRollback[NodeId]] = {
     for {
-      version <- transactionVersionGen()
       children <- Gen
         .listOf(Arbitrary.arbInt.arbitrary)
         .map(_.map(NodeId(_)))
         .map(ImmArray(_))
-    } yield NodeRollback(children, version)
+    } yield NodeRollback(children)
   }
 
   /** Makes exercise nodes with some random child IDs. */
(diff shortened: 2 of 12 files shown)


daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
```diff
@@ -427,7 +426,8 @@ object ValueGenerators {
   /** Makes nodes with the problems listed under `malformedCreateNodeGen`, and
     * `malformedGenTransaction` should they be incorporated into a transaction.
     */
-  val danglingRefGenNode: Gen[(NodeId, Tx.Node)] = {
+  val danglingRefGenNode //TODO: FIXME: this generator never produces rollback nodes
```
cocreature, line 429: Please add a reference to the issue so we can grep for this.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
```diff
@@ -20,11 +20,15 @@ object Node {
...
       with Serializable
       with CidContainer[GenNode[Nid, Cid]] {
-    def version: TransactionVersion
+
+    private[lf] def updateVersion(version: TransactionVersion): GenNode[Nid, Cid]
 
     def foreach2(fNid: Nid => Unit, fCid: Cid => Unit): Unit
 
-    private[lf] def updateVersion(version: TransactionVersion): GenNode[Nid, Cid]
+    def optVersion: Option[TransactionVersion] = this match {
```
cocreature, line 28: I’m slightly split about this. It makes a few call sites a bit more convenient but I think it might be preferable to be explicit and instead match on the node type in those cases. wdyt @nickchapman-da @remyhaemmerle-da?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
```diff
@@ -365,17 +371,10 @@ object Node {
...
-      // Note: if we add values, we must extend the function which checks they are serializable
-      children: ImmArray[Nid],
-      // For the sake of consistency between types with a version field, keep this field the last.
-      override val version: TransactionVersion,
+      children: ImmArray[Nid]
   ) extends GenNode[Nid, Nothing] {
 
-    override private[lf] def updateVersion(version: TransactionVersion): NodeRollback[Nid] =
-      copy(version = version)
+    private[lf] def updateVersion(version: TransactionVersion): NodeRollback[Nid] = this
```
cocreature, line 377: Similar to above, this seems convenient but really updating the version of a rollback node doesn’t make sense so maybe only having this method on action nodes is preferable?

cocreature (approved):
Nice, thank you!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
nickchapman-da: funnily enough, I just made a commit that moves slightly in the opposite direction.
I reinstated `Node.version` (removing `Node.optVersion`) 
```
    /** minimum version for a transaction containing this node */
    def version: TransactionVersion
```
With implementation for `NodeRollback`
```
    def version: TransactionVersion = TransactionVersion.minExceptions
```

This cleans up and avoid duplicating logic at multiple call sites.

I think this is nicer. But I can remove it if you hate it!

cocreature: I’d prefer to stick to the previous solution or to drop it like I suggested above. The rollback version is really different. E.g., consider some code which wants to validate that the children are never newer than their parent. This does not work with this implementation (well it works now since there is nothing newer than `minExceptions` but it will fail later) so I’d rather have the caller make a deliberate choice on how they want to handle this case than silently do the wrong thing.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
```diff
@@ -109,7 +109,7 @@ final class ValueEnricher(engine: Engine) {
 
   def enrichNode[Nid](node: GenNode[Nid, ContractId]): Result[GenNode[Nid, ContractId]] =
     node match {
-      case rb @ Node.NodeRollback(_, _) => // reconsider if fields added
+      case rb @ Node.NodeRollback(_) => // reconsider if fields added
```
remyhaemmerle-da, line 112: I think we can drop the comment.

remyhaemmerle-da (approved):
LGTM

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
nickchapman-da: I reverted this change. I kept the solution with `optVersion`

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Node.scala
nickchapman-da: replaced this method with a function at the call site

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
nickchapman-da: yup. dropped this _not very helpful_ comment everywhere.

PR #9628, opened 2021-05-10
Simplify kvutils contract keys validation [KVL-911]

cocreature:
### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -72,25 +72,18 @@ private[transaction] object ContractKeysValidation {
...
-        leaf = (status, _, leafNode) =>
-          onCurrentStatus(status)(checkNodeContractKey(leafNode, contractKeysToContractIds, _)),
-        exerciseEnd = (accum, _, _) => accum,
-        rollbackEnd = (status, _, _) => status.map(stack => stack.tail),
-      )
+    val transaction = transactionEntry.transaction
+
+    val keysValidationOutcome = for {
+      keyInputs <- transaction.contractKeyInputs.left.map(_ => Duplicate)
+      _ <- keyInputs.toList.traverse_ { case (key, optCid) =>
```
cocreature, line 79: Didn’t really seem worth factoring this out given that this is the only thing this function does but happy to factor it out in some form if you prefer.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -696,7 +696,7 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
       rejectionReason should startWith("DuplicateKeys")
     }
 
-    "return DuplicateKeys when a create in a rollback conflicts with a global key" in {
+    "return InconsistentKeys when a create in a rollback conflicts with a global key" in {
```
cocreature, line 699: Please verify that this is what you expect. I think the new behavior is correct:
The global key was not active during submission but it is active now so we get `InconsistentKeys`.

The new code makes this very explicit: `Transaction.keyInputs` checks that the transaction has no internal conflicts which should already have been an error during submission. Anything else is an inconsistency.

fabiotudone-da (approved):

cocreature:
@fabiotudone-da I rebased this on top of https://github.com/digital-asset/daml/pull/9683 which allows us to preserve the previous behavior. As discussed in Slack, I think we might want to revisit this more comprehensively since at least parts of the current behavior seem more of an artifact of the current implementation rather than something intentionally designed but we’ll do that separately. Factoring out the logic for these checks should actually also make these changes easier and more consistent across ledgers.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
cocreature: just to close the loop on this: I reverted to the previous behavior now.

fabiotudone-da (approved):

PR #9631, opened 2021-05-10
Add Transaction.contractKeyInputs to inputs of a transaction

cocreature:
You can see an example of using this in https://github.com/digital-asset/daml/pull/9628/commits/5560c11185ef029ff8375020784485c0d4d0bb42.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
@@ -586,11 +586,11 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
           case lookup: Node.NodeLookupByKey[Value.ContractId] =>
             handleLookup(lookup)
         }
-      def beginRollback(): State =
+      def beginRollback: State =
         copy(
           rollbackStack = keys :: rollbackStack
         )
-      def endRollback(): State =
+      def endRollback: State =
         copy(
           keys = rollbackStack.head,
           rollbackStack = rollbackStack.tail,
@@ -602,8 +602,8 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
       )(
         exerciseBegin = (acc, _, exe) => (acc.flatMap(_.handleExercise(exe)), true),
         exerciseEnd = (acc, _, _) => acc,
-        rollbackBegin = (acc, _, _) => (acc.map(_.beginRollback()), true),
-        rollbackEnd = (acc, _, _) => acc.map(_.endRollback()),
+        rollbackBegin = (acc, _, _) => (acc.map(_.beginRollback), true),
+        rollbackEnd = (acc, _, _) => acc.map(_.endRollback),
         leaf = (acc, _, leaf) => acc.flatMap(_.handleLeaf(leaf)),
       )
       .map(_.keyInputs)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -483,6 +483,132 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
...
+      def handleLeaf(leaf: Node.LeafOnlyActionNode[Value.ContractId]): Either[GlobalKey, State] =
+        leaf match {
+          case create: Node.NodeCreate[Value.ContractId] =>
+            handleCreate(create)
+          case fetch: Node.NodeFetch[Value.ContractId] =>
+            assertKeyMapping(fetch.templateId, fetch.coid, fetch.key)
+          case lookup: Node.NodeLookupByKey[Value.ContractId] =>
+            handleLookup(lookup)
+        }
+      def beginRollback(): State =
```
remyhaemmerle-da, line 589:
```suggestion
      def beginRollback: State =
```

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -483,6 +483,132 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
...
+        copy(
+          rollbackStack = keys :: rollbackStack
+        )
+      def endRollback(): State =
```
remyhaemmerle-da, line 593:
```suggestion
      def endRollback: State =
```

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -483,6 +483,132 @@ sealed abstract class HasTxNodes[Nid, +Cid] {
     }
   }
 
+  /** Return the expected contract key inputs (i.e. the state before the transaction)
+    * for this transaction or an error if the transaction contains a
+    * duplicate key error or has an inconsistent mapping for a key.
+    *
+    * Because we do not preserve byKey flags across transaction serialization
+    * this method will consider all operations with keys for conflicts
+    * rather than just by-key operations.
```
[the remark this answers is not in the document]
remyhaemmerle-da: I am  a bit uncomfortable with the "confusing semantics" argument when dealing with backward compatibility:
"Of course it is confusing, we are dealing with backward compatibility"

The question am I interested by is:  won't a ledger implementer that does not have to deal with version older than minByKey` tempted to implement that for byKey node only, hence defeating the purpose of this generic method ?  

Not saying we should go this way, just trying to challenge you through process.  
If you persist please modify the comment with something like: 
```suggestion
    * Because transaction serialization for LF <= 1.13 does not preserve byKey flags,
    * this method will consider all operations with keys for conflicts
    * rather than just by-key operations.
```

remyhaemmerle-da (approved):
This seems good.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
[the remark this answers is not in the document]
cocreature: > The question am I interested by is: won't a ledger implementer that does not have to deal with version older than minByKey` tempted to implement that for byKey node only, hence defeating the purpose of this generic method ?

I don’t really see why they would. This method is perfectly fine even for newer versions and they should be able to use it safely. At worst, they get some extra checks or inputs that might not be required.

I think what we might want to consider is to have a separate method that _only_ works for LF > 1.13 and does implement the `byKey` semantics. The nice thing about this is that at that point we will never get `Left` since the checks match up with Speedy.

What I’d like to avoid is mixing the semantics. That gets pretty confusing and like an easy way to introduce bugs.

At worst, you might end up

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
cocreature: Just to improve my scala knowledge. Is there a good reason why you prefer this? In Scala 2.13, there is a difference between the two and the call below will fail unless I remove the parentheses (and the other way around).

cocreature: Thanks!

PR #9633, opened 2021-05-10
Fix sandbox classic package upload

rautenrieth-da:
Previously, the copying of initial packages was broken, as `storePackageEntry` did not insert any packages if the `optEntry` parameter was missing.

This was found by temporarily making the append-only schema the default one and then running all sadbonx classic tests.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
@@ -668,44 +668,42 @@ private class JdbcLedgerDao(
         // This is again not an issue since all callers of this method always use the same value for all source descriptions.
         val update = optEntry match {
           case None =>
-            if (packages.nonEmpty) {
-              // Calling storePackageEntry() without providing a PackageLedgerEntry is used to copy initial packages,
-              // or in the case where the submission ID is unknown (package was submitted through a different participant).
-              Some(
-                Update.PublicPackageUpload(
-                  archives = packages.view.map(_._1).toList,
-                  sourceDescription = packages.head._2.sourceDescription,
-                  recordTime = Time.Timestamp.assertFromInstant(packages.head._2.knownSince),
-                  submissionId =
-                    None, // If the submission ID is missing, this update will not insert a row in the package_entries table
-                )
-              )
-            } else {
-              None
-            }
+            // Calling storePackageEntry() without providing a PackageLedgerEntry is used to copy initial packages,
+            // or in the case where the submission ID is unknown (package was submitted through a different participant).
+            Update.PublicPackageUpload(
+              archives = packages.view.map(_._1).toList,
+              sourceDescription = packages.headOption.flatMap(
+                _._2.sourceDescription
+              ),
+              recordTime = Time.Timestamp.assertFromInstant(
+                packages.headOption
+                  .map(
+                    _._2.knownSince
+                  )
+                  .getOrElse(Instant.EPOCH)
+              ),
+              submissionId =
+                None, // If the submission ID is missing, this update will not insert a row in the package_entries table
+            )
 
           case Some(PackageLedgerEntry.PackageUploadAccepted(submissionId, recordTime)) =>
-            Some(
-              Update.PublicPackageUpload(
-                archives = packages.view.map(_._1).toList,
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -655,22 +655,54 @@ private class JdbcLedgerDao(
...
+        // This is not an issue since all callers of this method always use the same value for all knownSince and recordTime times.
+        //
+        // Note on sourceDescription:
+        // - In the input, each package can have its own source description (see PackageDetails.sourceDescription)
+        // - In the intermediate value, there is only one source description for all packages (see PublicPackageUpload.sourceDescription)
+        // - In the database schema, each package can have its own source description (see packages.source_description)
+        // This is again not an issue since all callers of this method always use the same value for all source descriptions.
+        val update = optEntry match {
+          case None =>
+            if (packages.nonEmpty) {
```
nmarton-da, line 671: this if condition is not needed: in UpdateToDBDTOV1 there will be an empty iterator produced as soon as submissionId is empty and packages is empty.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -655,22 +655,54 @@ private class JdbcLedgerDao(
...
+        // - There are two different time values in the database schema: packages.known_since and package_entries.recorded_at
+        // This is not an issue since all callers of this method always use the same value for all knownSince and recordTime times.
+        //
+        // Note on sourceDescription:
+        // - In the input, each package can have its own source description (see PackageDetails.sourceDescription)
+        // - In the intermediate value, there is only one source description for all packages (see PublicPackageUpload.sourceDescription)
+        // - In the database schema, each package can have its own source description (see packages.source_description)
+        // This is again not an issue since all callers of this method always use the same value for all source descriptions.
+        val update = optEntry match {
+          case None =>
```
nmarton-da, line 670: we seem to have no test case for this in conformance tests. do you think we have in the Jdbc test suits? shall we have a TODO to add one?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: also if the if condition is dropped all the updates will be Some(...) so we can drop the Some(...) from everywhere and do it when we use the update in sequentialIndexer.store

nmarton-da (changes requested):
Great found, @rautenrieth-da !!! Kudos! How did you figure?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
rautenrieth-da: We do cover this in [JdbcLedgerDaoPackagesSpec](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPackagesSpec.scala#L104). These tests are not run with the append-only schema though.

I'm not 100% sure, but I don't think it's possible to cover this in the conformance test.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
rautenrieth-da: > this if condition is not needed

It's needed for the code below, because it uses `packages.head`, and I didn't want to rely too much on the fact that the packages will never be empty. I'll try to rewrite this whole block.

nmarton-da: oh yeah, right!

nmarton-da (approved):
quite involved piece with dependencies all around, seems like this is on the bleeding edge of this type of integration.
I think it is good enough for sandbox-classic.
The fix itself looks good :ok_hand: 
Thanks for spotting this early, @rautenrieth-da :raised_hands:

PR #9635, opened 2021-05-11
Upgrade rules-nodes to version 3.5.0

cocreature:
No particularly strong reason for doing that. I just like staying up2date.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature:
Looks like this was introduced by https://github.com/bazelbuild/rules_nodejs/pull/2435/files.

aherrmann-da (approved):

bazel_tools/rules_nodejs_windows_mkdir.patch
```diff
@@ -0,0 +1,31 @@
...
+-            ["mkdir", "-p", dirname],
++            args,
+             quiet = repository_ctx.attr.quiet,
+         )
+         if result.return_code:
+@@ -344,12 +345,13 @@ def _copy_file(repository_ctx, f):
+ 
+     # copy the file; don't use the repository_ctx.template trick with empty substitution as this
+     # does not copy over binary files properly
++    cp_args = ["cp", "-f", repository_ctx.path(f), to] if not is_windows_os(repository_ctx) else ["cmd", "/c", "copy /y {} {}".format("\\".join(str(repository_ctx.path(f)).split("/")), "\\".join(to_segments))]
```
cocreature, line 20: @aherrmann-da do you have a better idea for handling this? `ctx.file` doesn’t work since you either mess up executable bits or break binary files. Just `cp` doesn’t work because it might not be in PATH and this is a repository rule so we cannot get it from the posix toolchain. Just `cmd /c copy` does not work because it does not like forward slashes.
So in the end I ended up with this but it seems terrible.

aherrmann-da: Unfortunately, I don't really have a better idea. As you describe all these other methods have their respective limitations.

Perhaps [`string.replace`](https://docs.bazel.build/versions/master/skylark/lib/string.html#replace) is also a bit less noisy than the split/join, but, I have no strong preference there.

PR #9637, opened 2021-05-11
Support append-only schema in Sandbox migration tests

cocreature:
Note that these tests are switched off for now until Robert has
implemented the migration.

To avoid having to stuff too much in one PR the plan is to merge this
one first and then switch the flag once the migrations work.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


rautenrieth-da (approved):
I have manually tested that `//:migration-stable-append-only` fails if there is an error in the data migration of the append-only schema, but succeeds if the error is fixed.

PR #9638, opened 2021-05-11
sketch Read/WriteService API v2

andreaslochbihler-da:
Proposal for the `Read/WriteService` API that implements the [new proposal for command deduplication](https://docs.google.com/document/d/1R0yAD2buzkHjPiWAHm-1FtX8acQofZLRGA4KDZ4NI2w/edit) (internal document).

The corresponding changes to the Ledger API are sketched in #9567 and should be looked at before/simultaneously, in particular w.r.t. error reporting.

**Comments for reviewers**
* The changes here have been applied directly against the v1 API to ease understanding them. Once we have them well-understood, they will be rolled out in a backwards compatible manner, likely by introducing new `participant.state.api.v2` classes to the side of the existing unchanged `participant.state.v1` classes.
* Open discussion points are marked as `TODO (v2)`. Please comment if you think you can contribute something.
* Daml ledger: Do you foresee any problems with your ledger implementing the Read/WriteService API with the stated properties?
* Ledger API server: Do you foresee any problems with implementing the updated ledger API on top of this Read/WriteService API?

**Out of scope for this PR**
* Improving the reporting of existing errors

### Pull Request Checklist

trigger the build.


ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/DeduplicationPeriod.scala
```diff
@@ -0,0 +1,34 @@
...
+    */
+  case class DeduplicationTimepoint(since: Instant) extends DeduplicationPeriod
+
+  /** The length of the deduplication window, which ends when the [[WriteService]] or underlying Daml ledger processes
+    * the command submission. The window is measured on some unspecified clock on the participant or the Daml ledger.
+    *
+    * @throws java.lang.IllegalArgumentException if the `duration` is negative
+    */
+  case class DeduplicationWindow(duration: Duration) extends DeduplicationPeriod {
+    require(!duration.isNegative, s"The deduplication window must not be negative: $duration")
```
meiersi-da, line 32: @andreaslochbihler-da : this exception would likely be raised by the request parsing code in the API server. If we do this, will it be translated to the righ gRPC status code?

I have a hunch that the control flow would be easier to understand if parsing was total and the validation of the `DeduplicationPeriod` checked this constraint.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
```diff
@@ -0,0 +1,175 @@
...
+    *   A [[Update.CommandRejected]] completion does not trigger deduplication and implementations SHOULD
+    *   process such resubmissions normally, subject to the submission rank guarantee listed below.
+    *
+    *   TODO (SM): we would like to weaken this requirement to allow multiple
+    *   [[Update.TransactionAccepted]] updates provided
+    *   the transactions are sub-transactions of each other. Thereby enabling
+    *   the after-the-fact communication of extra details about a transaction
+    *   in case a party is newly hosted at a participant.
+    *   See https://github.com/digital-asset/daml/issues/430
+    *
```
meiersi-da, line 96:
```suggestion
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
```diff
@@ -0,0 +1,175 @@
...
+    * follows.
+    *
+    * - *unique offsets*: for any update `u1` with offset `uo` in `s1` and any
+    *   update `u2` with the same offset `uo` in `se2` it holds that `u1 == u2`.
+    *   This means that offsets can never be reused. Together with
+    *   *strictly increasing [[Offset]]* this also implies that the order of
+    *   elements present in both `s1` and `s2` cannot change.
+    *
+    * - *persistent updates*: any update other than
+    *   [[Update.CommandRejected]] in `s2` must also be present in `s1`.
```
meiersi-da, line 151: This whole specification is too strong; and for example not satisfied by `sandbox-on-x`, which uses an in-memory bridge. What we need to specify is that the streams are resumable in the following sense: the single stream properties are also satisfied when consuming updates from offsets `o1` to `o3`, and then restarting from `o2` while ingoring all `Update`s in the range `[o2, o3]` received on the first stream.

Do you see this the same @andreaslochbihler-da ?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
```diff
@@ -0,0 +1,175 @@
...
+    * - an accepted transaction with identifier `tid` evidenced to both participants `p1` and `p2`
+    *   in their state update streams after the [[Update.PartyAddedToParticipant]] updates for
+    *   `A`, respectively `B`.
+    * The projections of `tx1` and `tx2` to the nodes visible to both `A` and `B` is the same.
+    *
+    * Note that the transaction `tx1` associated to `tid` on `p1` is not required to be the same as
+    * the transaction `tx2` associated to `tid` on `p2`, as these two participants do not necessarily
+    * host the same parties; and some implementations ensure data segregation on the ledger. Requiring
+    * only the projections to sets of parties to be equal leaves just enough leeway for this
+    * data segregation.
```
meiersi-da, line 172:
```suggestion
    * data segregation.
    *
    * Note further that the offsets of the transactions might not agree, as these offsets are participant-local.
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmissionResult.scala
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  /** The submission has failed with a synchronous error.
+    *
+    * Asynchronous errors are reported via the command completion stream as a [[Update.CommandRejected]]
+    *
+    * TODO(v2) I don't think there's much point in fleshing out a detailed error model at this point
+    *  because the error conditions will change as we implement new features. To support error refinement in
+    *  the [[WriteService]] clients, we may later flesh out a more detailed error model or rely on conventions in
+    *  how to report refinable error conditions
```
meiersi-da, line 24: I agree.

meiersi-da (commented):
I've only gone through half-way, but won't have more time today. Here's the current state of the review.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
```diff
@@ -0,0 +1,60 @@
...
+  * [[com.daml.ledger.participant.state.v2]] package.  Where possible
+  * we will evolve them in a backwards compatible fashion, so that a simple
+  * recompile suffices to upgrade to a new version. Where that is not
+  * possible, we plan to introduce new version of this API in a separate
+  * package and maintain it side-by-side with the existing version if
+  * possible. There can therefore potentially be multiple versions of
+  * participant state APIs at the same time. We plan to deprecate and drop old
+  * versions on separate and appropriate timelines.
+  *
+  * TODO(v2) We only list the classes and types that differ from v1.
```
meiersi-da, line 48: TODO: explain the relation to the v1 interface

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
```diff
@@ -0,0 +1,155 @@
...
+    with WritePartyService
+    with WriteConfigService
+    with WriteParticipantPruningService
+    with ReportsHealth {
+
+  /** Submit a transaction for acceptance to the ledger.
+    *
+    * TODO(v2) Why do we need the requirement below on no IO? Canton doesn't respect it at the moment.
+    *  For command deduplication, we will need to do some IO inside the write service,
+    *  and want to report the outcome synchronously whenever possible.
```
meiersi-da, line 51: @andreaslochbihler-da : we can drop that requirement. It's a left-over from a time where the return result was not wrapped in a `CompletionStage`

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmitterInfo.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package com.daml.ledger.participant.state.v2
+
+import com.daml.ledger.participant.state.v1.{ApplicationId, CommandId, Offset, Party}
+
+/** Information provided by the submitter of changes submitted to the ledger.
+  *
+  * Note that this is used for party-originating changes only. They are
+  * usually issued via the Ledger API.
+  *
```
meiersi-da, line 12:
```suggestion
/** Information provided by the submitter of changes submitted to the ledger.
  * 
  * Note that the submitter is the process that acts as a client of the Ledger API server and not a specific party.
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmitterInfo.scala
```diff
@@ -0,0 +1,50 @@
...
+  *   submitted the command. This is used for monitoring and to allow DAML
+  *   applications subscribe to their own submissions only.
+  *
+  * @param commandId a submitter-provided identifier to identify an intended ledger change
+  *   within all the submissions by the same parties and application.
+  *
+  * @param deduplicationPeriod The deduplication period for the command submission.
+  *   If not given explicitly as an offset, the [[WriteService]] is responsible for
+  *   determining a suitable completion offset where deduplication starts according
+  *   to the [[ReadService]]'s deduplication guarantee.
```
meiersi-da, line 25:
```suggestion
  *   Used for the deduplication guarantee described in the [[ReadService.stateUpdates]]
```

KV ledgers will ultimately translate the given information to record_time timestamps. So, its probably better to not be too specific here.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmitterInfo.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /** The ID for the ledger change */
+  def changeId: ChangeId = ChangeId(applicationId, commandId, actAs = actAs.toSet)
+}
+
+/** Identifier for ledger changes used by command deduplication
+  *
+  * @see ReadService.stateUpdates for the command deduplication guarantee
+  */
+case class ChangeId(applicationId: ApplicationId, commandId: CommandId, actAs: Set[Party])
```
meiersi-da, line 50: @andreaslochbihler-da : we've had all sorts of problems from hash-based implementations of `Set`s. Consider changing this to `SortedSet` so we are explicit about requiring a stable ordering.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
```diff
@@ -0,0 +1,227 @@
...
+  /** Signal that the current [[Configuration]] has changed. */
+  final case class ConfigurationChanged(
+      recordTime: Timestamp,
+      submissionId: AdminSubmissionId,
+      participantId: ParticipantId,
+      newConfiguration: Configuration,
+  ) extends Update {
+    override def description: String =
+      s"Configuration change '$submissionId' from participant '$participantId' accepted with configuration: $newConfiguration"
+  }
```
meiersi-da, line 46: @andreaslochbihler-da : I'm thinking about how to implement the Ledger API server that handles both v1 and v2 ReadService backends, and it looks to me that a construction where a `v2.Update` is either a `v1.Update` or a `v2.TransactionAccepted` or a `v2.CommandRejected` would be a simple option with well-defined code-reuse properties. 

The constraint that a v2.ReadService does not generate v1.TransactionAccepted messages could then be enforced at runtime.

Once we have all ledgers updated to v2, we can then do the trivial refactoring of inlining the right v1 classes into v2 and enforce the structure at the type level.

What do you think?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
```diff
@@ -0,0 +1,155 @@
...
+    * described in [[ReadService.stateUpdates]].
+    *
+    * @param submitterInfo the information provided by the submitter for correlating this submission
+    *                      with its rejection or cancellation on the associated [[ReadService]].
+    * @param reason The rejection reason to be included in the [[Update.CommandRejected]]
+    * @param telemetryContext Implicit context for tracing.
+    */
+  def rejectSubmission(
+      submitterInfo: SubmitterInfo,
+      reason: com.google.rpc.Status
```
meiersi-da, line 153: This seems like the right approach. Do you expect the error details to be serialized in the IndexDB?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
```diff
@@ -0,0 +1,60 @@
...
+  */
+package object v2 {
+
+  /** Identifiers used to correlate admin submission with results. */
+  val AdminSubmissionId: Ref.LedgerString.type = Ref.LedgerString
+  type AdminSubmissionId = Ref.LedgerString
+
+  /** Identifier for command submissions. */
+  // TODO(v2) Should we restrict this to UUIDs?
```
meiersi-da, line 57: I'd suggest to restrict the length to 54 characters, which allows for base64url encoded 33 byte values (e.g. a SHA256 output prefixed by a one byte tag) .

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
```diff
@@ -0,0 +1,60 @@
...
+  * package and maintain it side-by-side with the existing version if
+  * possible. There can therefore potentially be multiple versions of
+  * participant state APIs at the same time. We plan to deprecate and drop old
+  * versions on separate and appropriate timelines.
+  *
+  * TODO(v2) We only list the classes and types that differ from v1.
+  */
+package object v2 {
+
+  /** Identifiers used to correlate admin submission with results. */
```
meiersi-da, line 52: Apply the same length restriction as for SubmissionIds

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
```diff
@@ -0,0 +1,175 @@
...
+    * need to be persisted. Before explaining them in detail we provide
+    * intuition.
+    *
+    * All [[Update]]s other than [[Update.CommandRejected]] must
+    * always be persisted by the backends implementing the [[ReadService]].
+    * For rejections, the situation is more nuanced, as we want to provide
+    * the backends with additional implementation leeway.
+    *
+    * TODO(v2) Do we actually exploit this freedom anywhere? With later [[Update.CommandRejected]]s referring
+    *  to earlier completion offsets (e.g., ALREADY_EXISTS), this becomes tricky to specify.
```
meiersi-da, line 124: I'd suggest we shorten this text to only cover resumability as outlined below.

meiersi-da (commented):
@andreaslochbihler-da : thanks for creating this draft. The direction looks good to me.

I've added some comments for simplification.

I superficially read the comments, but not yet made an in-depth pass to polish them, as I wasn't sure whether you already pursue the goal of having polished comments on this PR. Should I make such a pass?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/DeduplicationPeriod.scala
andreaslochbihler-da: With defensive programming, the parsing code should responsible for checking this precondition before invoking the constructor. We can add a suitable parsing method that does that for the companion object.  (That's how we're doing it in Canton.) The benefit of having the `require` here is that all users of the class can rely on the duration being non-negative.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
andreaslochbihler-da: Removed.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
andreaslochbihler-da: The whole resumption requirements are taken over literally from what we currently have in `v1.ReadService`. So are you saying that our own implemenations do not conform to the requirements we currently have? Let's discuss this in person.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ReadService.scala
andreaslochbihler-da: Added.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmitterInfo.scala
andreaslochbihler-da: I have taken that over from the `v0.Update` definition. Should we change it there, too? I'd rather rename the class than to provide this clarification because `actAs`'s documentation also uses the word "submit" and they are Daml parties.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/SubmitterInfo.scala
andreaslochbihler-da: :+1:

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
andreaslochbihler-da: Let's wait until we have an overview of how many Updates remain the same and how much in the Ledger API server needs to change. I don't have a feeling that we'd keep a lot of code the same. If that's the case, then it might make sense to factor out a super-trait for the case classes that remain the same.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
andreaslochbihler-da: Great. I cut it down to "thread-safe".

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
andreaslochbihler-da: Yes. See my proposal in the Ledger API design PR #9567.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
andreaslochbihler-da: FYI: In Canton, we're using a multibyte prefix for hashes.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
andreaslochbihler-da: We've settled on LedgerString (255 characters).

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/package.scala
andreaslochbihler-da: As discussed, let's stick with LedgerString for now.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/ReadService.scala
```diff
@@ -70,7 +70,8 @@ trait ReadService extends ReportsHealth {
     * - *time skew*: given a [[Update.TransactionAccepted]] with an associated
     *   ledger time `lt_tx` and a record time `rt_tx`, it holds that
     *   `rt_TX - minSkew <= lt_TX <= rt_TX + maxSkew`, where `minSkew` and `maxSkew`
-    *   are parameters specified in the ledger [[com.daml.ledger.participant.state.v1.TimeModel]].
+    *   are parameters specified in the ledger [[com.daml.ledger.participant.state.v1.TimeModel]]
+    *   of the last [[Update.ConfigurationChanged]] before the [[Update.TransactionAccepted]].
```
andreaslochbihler-da, line 74: FYI: Longer-term, I suspect that `ConfigurationChanged` will have an `effectiveAt` field that is after its record time.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/DeduplicationPeriod.scala
meiersi-da: Thank you for the explanation. That makes sense.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/ReadService.scala
meiersi-da: Agreed. I would though not expect the note in the text to say that. Would you?

andreaslochbihler-da: No. That's why I put an `FYI` into the comment.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
```diff
@@ -3,35 +3,44 @@
...
+  * @param submissionRank The rank of the submission among all submissions with the same change ID.
+  *   Used for the submission rank guarantee described in the [[ReadService.stateUpdates]].
   */
 final case class SubmitterInfo(
     actAs: List[Party],
     applicationId: ApplicationId,
     commandId: CommandId,
-    deduplicateUntil: Instant,
-)
+    deduplicationPeriod: DeduplicationPeriod,
```
meiersi-da, line 33: DeduplicationPeriod is not defined in the Scala classes. Do you expect this to be the class generated from the .proto definition?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Configuration.scala
```diff
@@ -11,15 +11,18 @@ import scala.util.Try
...
-  * @param maxDeduplicationTime The maximum time window during which commands can be deduplicated.
+  * @param timeModel The time model of the ledger.
+  * @param deduplicationPeriodLengthGuarantee A lower bound on the length of the deduplication period for command submissions
+  *                                           that the WriteService provides.
   */
 final case class Configuration(
     generation: Long,
     timeModel: TimeModel,
-    maxDeduplicationTime: Duration,
+    deduplicationPeriodLengthGuarantee: Duration,
```
meiersi-da, line 21: The interaction between this field and the `deduplicationPeriod` on the `SubmitterInfo` is not clear to me. Can you clarify?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmissionResult.scala
```diff
@@ -14,18 +14,11 @@ object SubmissionResult {
...
-
-  /** Submission is not supported */
-  case object NotSupported extends SubmissionResult {
-    override val description: String = "Submission is not supported"
-  }
-
-  /** Submission ended up with internal error */
-  final case class InternalError(reason: String) extends SubmissionResult {
-    override val description: String = s"Failed with an internal error, reason=$reason"
+  /** The submission has failed with a synchronous error.
```
meiersi-da, line 17:
```suggestion
  /** The submission has failed with a synchronous error.
  * 
  * Please see the documentation on the `CommandSubmissionService` to choose appropriate errors.
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
```diff
@@ -3,35 +3,44 @@
...
+
+  /** The ID for the ledger change */
+  def changeId: ChangeId = ChangeId(applicationId, commandId, actAs = actAs.toSet)
+}
+
+/** Identifier for ledger changes used by command deduplication
+  *
+  * @see ReadService.stateUpdates for the command deduplication guarantee
+  */
+case class ChangeId(applicationId: ApplicationId, commandId: CommandId, actAs: Set[Party])
```
meiersi-da, line 46:
```suggestion
// TODO v2: change to a `SortedSet` if possible to ensure a canonical representation of the ChangeId
case class ChangeId(applicationId: ApplicationId, commandId: CommandId, actAs: Set[Party])
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -155,9 +154,13 @@ object Update {
     *
     * @param optSubmitterInfo:
     *   The information provided by the submitter of the command that
-    *   created this transaction. It must be provided if the submitter is
-    *   hosted at this participant. It can be elided otherwise. This allows
-    *   ledgers to implement a fine-grained privacy model.
+    *   created this transaction. It must be provided if this participant
+    *   hosts one of the submitters and shall output a completion event
```
meiersi-da, line 158:
```suggestion
    *   hosts one of the [[SubmitterInfo.actAs]] parties and shall output a completion event
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,16 +200,19 @@ object Update {
...
   final case class CommandRejected(
       recordTime: Timestamp,
       submitterInfo: SubmitterInfo,
-      reason: RejectionReason,
+      reason: com.google.rpc.status.Status,
+      cancelled: Boolean,
   ) extends Update {
     override def description: String = {
-      s"Reject command ${submitterInfo.commandId}: $reason"
+      s"Reject command ${submitterInfo.commandId}: ${reason.message}"
```
meiersi-da, line 215:
```suggestion
      s"Reject command ${submitterInfo.commandId} ${if (cancelled) "(cancelled)"}: ${reason.message}"
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,16 +200,19 @@ object Update {
 
   /** Signal that a command submitted via [[WriteService]] was rejected.
     *
-    * See the different [[RejectionReason]] for why a command can be
-    * rejected.
+    * @param cancelled If false, the [[ReadService]]'s deduplication and submission rank guarantees
+    *   apply to this rejection. The participant state implementations should
+    *   strive to set this flag to false as often as possible so that applications
+    *   get better guarantees.
```
meiersi-da, line 206: TODO: add commentary for all parameters, and in particular, refer to the guidance on the error messages given via the Ledger API

meiersi-da (commented):
@andreaslochbihler-da : I've completed a thorough review. Let's work towards resolving my comments either async here or in today's sync meeting.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Configuration.scala
andreaslochbihler-da: It plays the same role as `maxDeduplicationTime` in the old model. If the write service reports a guaranteed deduplication period length of 1 day, then it shouldn't complain about the deduplication period being too long if it receives a command that uses a period of up to one 1 (subject to the usual clock skew provisions).

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmissionResult.scala
andreaslochbihler-da: We can add that now, but I do not expect that the documentation on the `CommandSubmissionService` will ever be exhaustive, as different ledger implementations will generate different kinds errors. So your suggestion is too strong. How about the following:
```suggestion
  /** The submission has failed with a synchronous error.
    *
    * See the documentation on the `CommandSubmissionService` for how to report common errors.
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
andreaslochbihler-da: `SortedSet` looks like an implementation concern of kvutils. We hardly use `SortedSet` in Canton, so I'm hesitant to enforce it here. All I care about is the equality of `ChangeId`s, which we need for the specification of the `ReadService`. How about making the fields `private`?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,16 +200,19 @@ object Update {
...
     */
   final case class CommandRejected(
       recordTime: Timestamp,
       submitterInfo: SubmitterInfo,
-      reason: RejectionReason,
+      reason: com.google.rpc.status.Status,
+      cancelled: Boolean,
```
andreaslochbihler-da, line 212: We probably don't want `cancelled` be a field of the class here, given that the ledger API now models this as a `definite_answer` key in the error details of `reason`. I'll make it derived.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: Changed.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: Added a `TODO (v2)`.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
meiersi-da: Making them private makes this class almost unusable. Alternatively, we could add a comment for implementations that they should ensure canonical serialization.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,17 +202,29 @@ object Update {
...
     *
-    * See the different [[RejectionReason]] for why a command can be
-    * rejected.
+    * TODO(v2) add commentary for all parameters, and in particular, refer to the guidance on the error messages given via the Ledger API
     */
   final case class CommandRejected(
       recordTime: Timestamp,
       submitterInfo: SubmitterInfo,
-      reason: RejectionReason,
+      reason: com.google.rpc.status.Status,
```
meiersi-da, line 210: @andreaslochbihler-da : for `KVUtils` (and likely also for Daml on Corda), it would be great to have the help of the `IndexDB` for resolving `(submissionId, submissionRank)` tuples to their completion offset. With the parallel indexer, this actually means that one needs the help of the `Indexer`, as there is an unspecified number of completions that the indexer is in process of persisting; and lookups need to be resolved against them as well as historic completions.

My hunch is that delegating that work explicitly to the indexer in a typed fashion (i.e., a Sum-type here) would lead to a cleaner design, than trying to wire it up via method calls from the `ReadService` implementation to the `Indexer`

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Configuration.scala
andreaslochbihler-da: Leaving comment as FYI.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
andreaslochbihler-da: TODO: add comments
Should offer stable hash for actAs that we can store in the DB. Maybe even use the hash for equality?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: I'll give it a try.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
andreaslochbihler-da: Let's keep this idea in the comment.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
andreaslochbihler-da: Added the interface.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,17 +202,54 @@ object Update {
...
+      * strive to set this flag to false as often as possible so that applications
+      * get better guarantees.
+      */
+    lazy val cancelled: Boolean = !reason.details.exists { any =>
+        if (any.is(com.google.rpc.ErrorInfo.class)) {
+          Try(any.unpack(com.google.rpc.ErrorInfo.class)).exists( errorInfo =>
+            errorInfo.getMetadataOrDefault("definite_answer", "false") == "true"
+          )
+        } else false
+      }
```
andreaslochbihler-da, line 231: This doesn't work as is because we don't yet have a completed reason. Move this method into the template.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/ReadService.scala
```diff
@@ -63,75 +64,65 @@ trait ReadService extends ReportsHealth {
     *   meta-data.
     *   Note that the ledger time of unrelated updates is not necessarily monotonically
     *   increasing.
+    *   The creating transaction need not have a [[Update.TransactionAccepted]] even on this participant
+    *   if the participant does not host a stakeholder of the contract, e.g., in the case of divulgence.
     *
     * - *time skew*: given a [[Update.TransactionAccepted]] with an associated
     *   ledger time `lt_tx` and a record time `rt_tx`, it holds that
     *   `rt_TX - minSkew <= lt_TX <= rt_TX + maxSkew`, where `minSkew` and `maxSkew`
-    *   are parameters specified in the ledger [[TimeModel]].
```
meiersi-da, line 70: Note for reviewers: a good chunk of the changes on the guarantees of `ReadService.stateUpdates` is cleanup that is in principle independent of this PR. It's probably easiest to view the file as a whole and check that commentary makes sense to you.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,17 +202,54 @@ object Update {
...
-      s"Reject command ${submitterInfo.commandId}: $reason"
+      s"Reject command ${submitterInfo.commandId}${if (cancelled) " (cancelled)"}: ${reason.message}"
     }
+
+    /** If false, the [[ReadService]]'s deduplication and submission rank guarantees
+      * apply to this rejection. The participant state implementations should
+      * strive to set this flag to false as often as possible so that applications
+      * get better guarantees.
+      */
+    lazy val cancelled: Boolean = !reason.details.exists { any =>
```
meiersi-da, line 225: @andreaslochbihler-da : it seems this would be easier to understand if we defined the positive variant of this: `definite_answer`. I know we've talked about cancelled in the proposal, but it seems that your initial approach of talking about definite answers might work better for the API we have.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,17 +202,54 @@ object Update {
...
+    class FinalReason(val statusCode: com.google.rpc.status.Status) extends RejectionReasonTemplate
+
+    /** The indexer shall fill in a completion offset for the completion that corresponds to the `submissionId`,
+      * by calling `provideCompletionOffset` with the completion offset. If no completion offset
+      * for the `submissionId` can be provided, [[scala.None$]] can be used instead,
+      * which may lead to less informative errors.
+      */
+    class NeedCompletionOffsetForSubmissionId(
+        val submissionId: SubmissionId,
+        val provideCompletionOffset: Option[Offset] => RejectionReasonTemplate,
```
meiersi-da, line 252:
```suggestion
        val provideCompletionOffset: Option[Offset] => FinalReason,
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -197,17 +202,54 @@ object Update {
...
+    /** The status code for the command rejection. */
+    class FinalReason(val statusCode: com.google.rpc.status.Status) extends RejectionReasonTemplate
+
+    /** The indexer shall fill in a completion offset for the completion that corresponds to the `submissionId`,
+      * by calling `provideCompletionOffset` with the completion offset. If no completion offset
+      * for the `submissionId` can be provided, [[scala.None$]] can be used instead,
+      * which may lead to less informative errors.
+      */
+    class NeedCompletionOffsetForSubmissionId(
+        val submissionId: SubmissionId,
```
meiersi-da, line 251:
```suggestion
        val submissionId: SubmissionId,
        val submissionRank: SubmissionRank,
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
```diff
@@ -95,12 +95,35 @@ trait WriteService
...
       transactionMeta: TransactionMeta,
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
   )(implicit telemetryContext: TelemetryContext): CompletionStage[SubmissionResult]
+
+  /** Record the rejection of a command submission on the ledger.
+    *
+    * This method must be thread-safe.
+    *
+    * The result is communicated asynchronously via a [[ReadService]] implementation backed by the same participant
```
meiersi-da, line 110:
```suggestion
    * This method is used by the Ledger API server to deliver interpretation failures as definitive answers to the
    * Ledger API client via completions; thereby allowing the Ledger API client to safely retry the command if desired.
    * The result is communicated asynchronously via a [[ReadService]] implementation backed by the same participant
```

meiersi-da (commented):
@andreaslochbihler-da : another review round done. I've left some inline comments that need addressing.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
andreaslochbihler-da: Added roughly. Note that the point of recording is that the application can decide to **not** retry. Retrying itself would always be safe thanks to the deduplication guarantees.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: Code changed completely anyway.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: Added.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
andreaslochbihler-da: Switched to `definite answer`.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
```diff
@@ -3,35 +3,71 @@
...
+  /** The ID for the ledger change */
+  val changeId: ChangeId = new ChangeId(applicationId, commandId, actAs.toSet)
+}
+
+/** Identifier for ledger changes used by command deduplication
+  *
+  * @see ReadService.stateUpdates for the command deduplication guarantee
+  */
+class ChangeId(
+    private val applicationId: ApplicationId,
```
meiersi-da, line 49: @andreaslochbihler-da : so far we didn't use the `applicationId` as part of the change-id. Adding it now without binding it to a version (e.g., the transaction version of the submitted transaction), will break backwards compatibility.

How important is that change? Should we just leave the `applicationId` out of the changeId?

andreaslochbihler-da: Our current documentation is inconsistent whether the application ID should be part of the change ID or not (https://github.com/digital-asset/daml/blob/main/ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto#L38 says that application ID should be part of it whereas https://docs.daml.com/app-dev/app-arch.html#command-deduplication doesn't mention the application ID).

As part of command dedup cleanup, we should make up our mind whether we want to provide cross-application command deduplication in the long run. If not, then now is a good time to add the application ID to the change ID.

Not providing command dedup across application IDs provides the following benefits:
* Several applications cannot interfere with each other if they happen to choose the same command ID.
* We may want to offer different `max_deduplication_time`s for different applications (high-throughput applications need higher SLAs for long dedup times).

So ultimately this comes down to a product decision. During the design discussion, @bame-da confirmed that he doesn't see a use case for command dedup across applications.

As for backwards compatibility, I don't think we need versioning. When we switch a Daml ledger to the new scheme, we just have to check that there is no deduplication entry for the old format (i.e., without application ID), but we don't need to create one in the old format (provided participants of one application switch over to the format in one step). Since the lifetime of command dedup is bounded, we can eventually completely get rid of this additional check.

meiersi-da: I'm all for including the `applicationId` in the `changeId`.

We do have both KV production ledgers as well as daml-on-sql production ledgers. So, we need to devise a way to roll-out these changes in a way that guarantees deduplication across upgrades. Your approach of checking the duplication keys for both the `(actAs, commandId)` and `(actAs, applicationId, commandId)` sounds good. We can then remove the check for `(actAs, commandId)` in a later release (assuming deduplication is not required for multiple weeks).

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
```diff
@@ -3,35 +3,71 @@
...
+  * @param submissionRank The rank of the submission among all submissions with the same change ID.
+  *   Used for the submission rank guarantee described in the [[ReadService.stateUpdates]].
   */
 final case class SubmitterInfo(
     actAs: List[Party],
     applicationId: ApplicationId,
     commandId: CommandId,
-    deduplicateUntil: Instant,
-)
+    deduplicationPeriod: DeduplicationPeriod,
```
meiersi-da, line 35: @andreaslochbihler-da : the `SubmitterInfo` is also used on `TransactionUpdated`. Would you expect this field to be populated there?

I currently would not, which hints at moving the `deduplicationPeriod` into either the `submitTransaction` method itself or into its own case class.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
```diff
@@ -95,12 +95,39 @@ trait WriteService
     * @param estimatedInterpretationCost Estimated cost of interpretation that may be used for
     *                                    handling submitted transactions differently.
     * @param telemetryContext            Implicit context for tracing.
-    * @return an async result of a SubmissionResult
     */
   def submitTransaction(
       submitterInfo: SubmitterInfo,
       transactionMeta: TransactionMeta,
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
```
meiersi-da, line 103: @andreaslochbihler-da : KV implementations will need to check the requested `SubmitterInfo.deduplicationPeriod` against the `maxDeduplicationPeriod` provided in the `LedgerConfiguration`. 

Would you expect that to be provided by the caller or read on-demand by the KV implementations?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
andreaslochbihler-da: I don't have a strong opinion on that. `deduplicateUntil` was already there and I didn't see a reason to change this now, as I don't know whether the indexer makes use of this field or there are plans to do so. If there are none, then we can pass this as another parameter to `submitTransaction` and `rejectSubmission` instead of making it part of `SubmitterInfo`.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
andreaslochbihler-da: The KV implementations anyway need some hook into persistence (on the indexer) for filling in completion offsets, so they can get the current ledger configuration from there, possibly with an in-memory cache to avoid frequent reads from the DB. Other Read/WriteService implementations may not need the current `LedgerConfiguration`, so it would feel strange to add such a parameter to the Read/WriteService.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
```diff
@@ -95,12 +95,39 @@ trait WriteService
...
+    *
+    * Recorded rejections fall under the deduplication and submission rank guarantees
+    * described in [[ReadService.stateUpdates]].
+    *
+    * @param submitterInfo the information provided by the submitter for correlating this submission
+    *                      with its rejection on the associated [[ReadService]].
+    * @param reason The rejection reason to be included in the [[Update.CommandRejected]]
+    * @param telemetryContext Implicit context for tracing.
+    */
+  def rejectSubmission(
```
meiersi-da, line 129: @andreaslochbihler-da : we need the `submissionTime` available here to convert deduplicationPeriods. We might want to add that timestamp to the `SubmitterInfo`; it's though then duplicated in the `TransactionMeta`.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
meiersi-da: I thought we'd delegate the filling of completion offsets to the indexer due to the complex interactions with parallel, batched ingestion.

I agree though that we should have clear rules of what is passed as a parameter on a `WriteService` method. I'd suggest: the Ledger API server passes explicit immutable references for all mutable data read from the IndexDB.

This way `WriteService` implementations can use the same version of that data that was used by the Ledger API server.

So in this case, we should check whether the Ledger API server uses the `LedgerConfiguration` during interpretation or not. If it does, it should pass it on. If not, the implementation should look it up.

meiersi-da: @andreaslochbihler-da : as per our discussion, let's add the `LedgerConfiguration` that was used to interpret the command and determine it's ledger effective time.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
```diff
@@ -95,12 +95,39 @@ trait WriteService
...
+      submitterInfo: SubmitterInfo,
```
meiersi-da, line 130:
```suggestion
      submitterInfo: SubmitterInfo,
      submissionTime : Timestamp,
      submissionConfig: LedgerConfig,
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
meiersi-da: Let's remove `deduplicationPeriod` from here, and add it directly to the `WriteService` call

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
meiersi-da: Group into `SubmissionContext` case class and add to `submitTransactoin` as well

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
andreaslochbihler-da: Added a `SubmissionContext` parameter for such context information.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/WriteService.scala
andreaslochbihler-da: Added it as a separate parameter. Let's discuss what's the best place for this parameter. For `submitTransaction`, the `submissionTime` is used for interpreting / validating the transaction, so `TransactionMeta` is actually the right place.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/SubmitterInfo.scala
andreaslochbihler-da: I'm not yet convinced. Let's discuss this.

andreaslochbihler-da: As discussed, let's leave it in `SubmitterInfo` now that we've also got `CompletionInfo`.

andreaslochbihler-da:
V2 API has by now been implemented, albeit without submission ranks.

PR #9640, opened 2021-05-11
Block concurrent SDK installations.

### LINKED ISSUE #9612: Guard against concurrent SDK installations
Currently, the assistant will fall over if you try to install the same SDK version in parallel.

The most sensible solution that I see here seems to be to write out a lock file before modifying anything in `~/.daml` and deleting it afterwards.

There are two options I see for what a concurrent invocation can do:
1. Error out with a clean error.
2. Block until the first one is finished.

I think 2 probably leads to a nicer UX. E.g., consider a build system which does parallel `daml build` invocations which end up installing the same SDK version. It doesn’t really make sense to fail a build because of that. Arguably the build system should enforce that the SDK is already installed but we still want to provide a nice UX if users don’t do that.

sofiafaro-da:
Fixes #9612

changelog_begin

- [Daml Assistant] The assistant will now avoid
  installing SDK versions concurrently, waiting
  for the previous installation to finish before
  starting the next installation (if still
  necessary). This fixes a bug where the SDK
  would become corrupted because two
  installations were started concurrently.

changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice work! A test would be great but this looks rather painful to test so perhaps more sensible to not add one.

oliverse-da:
Hi @sofiafaro-da , quick question regarding the "if still necessary" aspect:

> - [Daml Assistant] The assistant will now avoid
>   installing SDK versions concurrently, waiting
>   for the previous installation to finish before
>   starting the next installation (if still
>   necessary). 

On our nightly performance runner server, @matthiasS-da has observed that we may perform multiple downloads. Does that look possible / plausible?

sofiafaro-da:
@oliverse-da Are these multiple installs for the same SDK version? I would expect multiple downloads only if there are different SDK versions being installed, or if an install fails for some reason.

I wouldn't expect multiple *concurrent* downloads ever, unless you're using something other than the daml assistant to start the install (e.g. the get.daml.com script).

oliverse-da:
Thanks @sofiafaro-da - as discussed with @matthiasS-da - if this happens again, we will produce a proper report. He mentioned he had an intermittent internet connection, so retries are a possibilities. I also didn't think they could be parallel, but perhaps serial downloads (e.g. retries).

Issue #7622, opened 2020-10-09
LF: Mark nodes that correspond to "by key" operations.


remyhaemmerle-da:
This issue is concerned with: 
- marking `Exercise` and `Fetch` nodes in a transaction with a `byKey` flag to indicate if those nodes where created by a `ExerciseByKey` and `FetchByKey` respectively. (done)
- add the flag in the next transaction serialization format + transaction spec (done #9632 and #9642) 
- change validation to take into account the flag for newer version of LF/transaction. (done #9641)

remyhaemmerle-da:
cc @hurryabit @sofiafaro-da

hurryabit:
I assume the only way you can exercise by key is via the Ledger API since there's no primitive to do so in DAML-LF. Should we change that and add an actual `exerciseByKey` primitive to the next version of DAML-LF? This might improve transaction sizes since we won't have a `fetchByKey` node followed by an `exercise` node anymore. cc @andreaslochbihler-da

andreaslochbihler-da:
We already have such a flag in the transaction metadata that's output by DAML engine. I'm happy if we can move this flag into the transaction tree itself. That saves in Canton a few consistency checks between the metadata and the transaction tree. I'm also in favor of adding the flag to exercise nodes in the tree. If the flag is set, this then corresponds to a `exercisebyKey` action just like a plain `Fetch` node with the flag represents a `fetchByKey` action.

Regarding `exerciseByKey`, we're re-experiencing the same story as for ordinary `fetch + exercise`. A while ago, every DAML `exercise` was compiled into `fetch + exercise` because the original semantics of `exercise` didn't check activeness. If we can remove the redundancy now for `byKey` as well, I'm for it.

hurryabit:
@remyhaemmerle-da Given that @andreaslochbihler-da would be interested in us producing an `exercise` node with `byKey = true` for every `exerciseByKey` in the DAML code and it would probably make things clearer, shall we add such an `exerciseByKey` primitive to DAML-LF?

remyhaemmerle-da:
> shall we add such an `exerciseByKey` primitive to DAML-LF?
I am fine with that. I do not expect this to be difficult.

hurryabit:
> I am fine with that. I do not expect this to be difficult.

Cool. Let's do it then while we're at cleaning up things in this regard. Would DAML-LF 1.9 be a realistic target?

cocreature:
fixed!

PR #9641, opened 2021-05-11
Check byKey in transaction validation

cocreature:
No tests because validation has no tests in general :(

part of #7622

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
@@ -6,7 +6,7 @@ package engine
 package preprocessing
 
 import com.daml.lf.data.{BackStack, ImmArray}
-import com.daml.lf.transaction.{GenTransaction, Node, NodeId}
+import com.daml.lf.transaction.{GenTransaction, Node, NodeId, TransactionVersion}
 import com.daml.lf.value.Value
 import com.daml.lf.value.Value.ContractId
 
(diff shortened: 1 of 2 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -227,7 +240,7 @@ private final class Validation[Nid, Cid](implicit ECid: Equal[Cid]) {
                     signatories2,
                     stakeholders2,
                     key2,
-                    byKey2 @ _,
+                    byKey2,
```
remyhaemmerle-da, line 243: Good catch. I wondering who write the original code ;).

remyhaemmerle-da (approved):
LGTM. Thanks.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -237,7 +250,8 @@ private final class Validation[Nid, Cid](implicit ECid: Equal[Cid]) {
                   (actingParties1.isEmpty || actingParties1 == actingParties2) &&
                   signatories1 == signatories2 &&
                   stakeholders1 == stakeholders2 &&
-                  (key1.isEmpty || keyIsReplayedBy(key1, key2)) =>
+                  (key1.isEmpty || keyIsReplayedBy(key1, key2)) &&
```
remyhaemmerle-da, line 253:
```suggestion
                  (key1. isEmpty || byKeyIsReplayedBy(byKey, byKey2)) &&
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/TransactionPreprocessor.scala
```diff
@@ -95,12 +95,23 @@ private[preprocessing] final class TransactionPreprocessor(
...
             case exe: Node.NodeExercises[_, Cid] =>
-              val (cmd, newCids) = commandPreprocessor.unsafePreprocessExercise(
-                exe.templateId,
-                exe.targetCoid,
-                exe.choiceId,
-                exe.chosenValue,
-              )
+              import scala.Ordering.Implicits.infixOrderingOps
+              val (cmd, newCids) = exe.key match {
+                case Some(key) if exe.byKey && exe.version >= TransactionVersion.minByKey =>
```
cocreature, line 100: @remyhaemmerle-da can you please check if seems sensible? This was (understandably) necessary to fix the conformance tests.

cocreature: > I would prefer we do check for version, i.e. the transaction coder ensures already that exe.byKey is defined only if exe.version >= TransactionVersion.minByKey

That seems fragile. Now it matters if I first roundtrip via serialization & deserialization or whether I get a transaction directly out of the engine and pass it here. I don’t really see a usecase for the second case but I still feel slightly uneasy about breaking it.

remyhaemmerle-da: The validation was original designed to work only on deserialized transaction. I think  we should continue to assume this as: 
- it simplify reasoning about validation as transaction are somehow normalized by the serialization/deserialization process. 
- I do not see use case were we may want to validate non serialized transaction
- The way we are distributing the versioning logic through the code seems to me very fragile, I really want to limit it to some very limited places. You introduce in this PR tow new places.  
- why changing something that has proven to work.

remyhaemmerle-da: I would be fine with writing a transaction (version aware) normalize equivalent to the serialization-deserialization.

remyhaemmerle-da: I was indeed thinking to apply it inside the finish. For canton I see several solution:
 - We add a config flag to normalize or not the `byKey` flag
 - We include the `byKey` flag info in the `Transaction.MetaData`
 - We force Canton to use only LF 1.14 (wish could be somehow possible as there is no Canton production).

PR #9651, opened 2021-05-12
Drop version check from TransactionPreprocessor

cocreature:
This is only supposed to be run on nodes that have been normalized via
de(serialization) so this is just useless shotgun validation.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractDiscriminatorFreshnessCheckSpec.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractDiscriminatorFreshnessCheckSpec.scala
@@ -354,14 +354,16 @@ class ContractDiscriminatorFreshnessCheckSpec
         pcs = _ => None,
         keys = _ => None,
       )
-      engine.replay(
-        submitters = Set(alice),
-        tx,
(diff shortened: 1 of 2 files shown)


remyhaemmerle-da (approved):
LGTM.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractDiscriminatorFreshnessCheckSpec.scala
```diff
@@ -354,14 +354,16 @@ class ContractDiscriminatorFreshnessCheckSpec
...
       )
-      engine.replay(
-        submitters = Set(alice),
-        tx,
-        ledgerEffectiveTime = txMeta.submissionTime,
-        participantId = participant,
-        submissionTime = txMeta.submissionTime,
-        submissionSeed = txMeta.submissionSeed.get,
-      ) shouldBe a[ResultDone[_]]
+      engine
```
cocreature, line 357: @remyhaemmerle-da side note: This is a case where we do not serialize/deserialize and probably tons of other tests as well. So definitely in favor of doing something like your suggestion of normalizing in `finish` (while figuring out what we do with canton & scenarios).

PR #9653, opened 2021-05-12
LF: add Arithmetic builtin exceptions

remyhaemmerle-da:
part of #8020

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
```diff
@@ -299,11 +299,42 @@ object SExpr {
...
+        actuals.add(v)
+        i += 1
+      }
+      builtin.executePure(actuals) match {
+        case Right(value) =>
+          machine.pushEnv(value) //use pushEnv not env.add so instrumentation is updated
+          machine.ctrl = body
+        case Left(message) =>
+          machine.traceLog.add("ArithmeticError:" + message, machine.lastLocation)
+          unwindToHandler(machine, SBuiltinException.Arithmetic)
```
nickchapman-da, line 325: Did we decide it's ok to unconditionally throw a catchable exception?
i.e  LF which predates `minException` will now throw catchable exceptions.

nickchapman-da (approved):
LGTM

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
remyhaemmerle-da: For what I remember, we said it was OK as long we do not create node with `version < minException` under a RollBack node

cocreature: Should be fine. It is slightly unfortunate that this means that just looking a function doesn’t tell me whether I can catch the exceptions it throws or not but given that we cannot avoid the dynamic nature in general, I think we don’t gain anything by handling this part differently.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
```diff
@@ -299,11 +299,42 @@ object SExpr {
...
+        val v = arg.lookupValue(machine)
+        actuals.add(v)
+        i += 1
+      }
+      builtin.executePure(actuals) match {
+        case Right(value) =>
+          machine.pushEnv(value) //use pushEnv not env.add so instrumentation is updated
+          machine.ctrl = body
+        case Left(message) =>
+          machine.traceLog.add("ArithmeticError:" + message, machine.lastLocation)
```
cocreature, line 324: I feel slightly uneasy about abusing the tracelog for this and mixing user logs with internal logs produced by speedy. Fine for this PR but I think we should separate the two.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -193,6 +193,26 @@ private[speedy] sealed abstract class SBuiltinPure(val arity1: Int) extends SBui
...
+
+  override private[speedy] def execute(
+      args: util.ArrayList[SValue],
+      machine: Machine,
+  ): Unit =
+    executePure(args) match {
+      case Right(value) =>
+        machine.returnValue = value
+      case Left(message) =>
+        machine.traceLog.add("ArithmeticError:" + message, machine.lastLocation)
```
cocreature, line 206: same here, I’m not sure relying on the tracelog for this is a good idea.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
cocreature: Looking at the log above, it also appears that you log the same error twice. Once here once in `SEBuiltinArithmetic`.

cocreature (approved):
lgtm with a few comments. It is slightly hard to see what is going on since it mixes a lot of refactoring with the actual changes. Let’s please try to separate those out in the future.

cocreature (approved):
Nice thank you!

I think there are two integration-type tests that we should have but those can be added in a separate PR:
1. An uncaught arithmetic error in particular in the scenario service. Ledger API is probably less interesting here.
2. A caught arithmetic error from Daml just to make sure it is pushed through all elements of the stack.

PR #9654, opened 2021-05-12
Throw on internal errors instead of setting ptx to aborted

cocreature:
The main purpose of aborted is to be able to get out a ptx in the
scenario service. However, for internal errors that makes no
sense. Users should never encounter them. And the use of `aborted`
here has swallowed these errors silently in a few places before which
is clearly bad.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -573,8 +573,7 @@ private[lf] case class PartialTransaction(
           nodes = nodes.updated(nodeId, exerciseNode),
           actionNodeSeeds = actionNodeSeeds :+ (nodeId -> ec.actionNodeSeed),
         )
-      case _ =>
-        noteAbort(Tx.NonExerciseContext)
+      case _ => throw new RuntimeException("endExercises called in non-exercise context")
     }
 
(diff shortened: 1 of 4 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -573,8 +573,7 @@ private[lf] case class PartialTransaction(
           nodes = nodes.updated(nodeId, exerciseNode),
           actionNodeSeeds = actionNodeSeeds :+ (nodeId -> ec.actionNodeSeed),
         )
-      case _ =>
-        noteAbort(Tx.NonExerciseContext)
+      case _ => throw new RuntimeException("endExercises called in non-exercise context")
```
cocreature, line 576: @remyhaemmerle-da do you think we should throw something else here? `RuntimeException` is what the two other cases internal errors in `PartialTransaction` throw atm so this seems consistent. `crash` is not used at all here.

But happy to switch to something else or define a new exception if you prefer.

remyhaemmerle-da: I am fine with `RuntimeException`.

remyhaemmerle-da (approved):
This makes a lot of sens. Thanks.

nickchapman-da (approved):
That's nicer

PR #9656, opened 2021-05-12
skip Windows tests on release

garyverhaegen-da:
Caching doesn't seem to work very well here. On a release, we build an old commit, which has already been tested twice (once as a commit on `main`, once as part of the release PR).

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -59,4 +59,6 @@ bazel build `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/build_exec
 
 bazel shutdown
 
-bazel test `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/test_execution_windows.log //...
+if ($env:SKIP_TESTS = "False") {
+    bazel test `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/test_execution_windows.log //...
+}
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Fair warning: I have no idea how either YAML or PowerShell programming works. 🤷

build.ps1
```diff
@@ -59,4 +59,6 @@ bazel build `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/build_exec
 
 bazel shutdown
 
-bazel test `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/test_execution_windows.log //...
+if ($env:SKIP_TESTS = "False") {
```
cocreature, line 62: Maybe worth adding a comment here or in the yaml file documenting that
1. We only do this for release builds where the PR (should) run exactly the same tests as the build on main.
2. The reason why release builds don’t benefit from caching on Windows is because we have to cache per working directory and different pipelines run in different working directories.

cocreature (approved):
lgtm assuming it works :slightly_smiling_face:

build.ps1
garyverhaegen-da: Added comments on the `ci/build-windows.yml` side; I think it makes more sense to have the comments near the decision logic.

PR #9659, opened 2021-05-12
Add a data-dependencies test for user-defined exceptions

sofiafaro-da:
Part of #8020

This only tests that (catching & throwing) user-defined exceptions compiles as expected across data-dependencies.

Once they're supported, we should also add a test involving throwing built-in exceptions (arithmetic & contract errors) in the data-dependency and catching them in main.


compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -1103,6 +1103,66 @@ tests Tools{damlc,repl,validate,davlDar,oldProjDar} = testGroup "Data Dependenci
             ]
         withCurrentDirectory (tmpDir </> "proj") $
             callProcessSilent damlc ["build"]
+
```
cocreature, line 1106: sort of related: we probably have to take another stab at optimizing those tests. I’ve seen a few timeouts recently.

cocreature (approved):
Thanks!

PR #9671, opened 2021-05-12
restore natural join and avoid table aliasing in participant contracts queries

dasormeter:
A regression was found in daml on sql that causes high load on postgres.
This was due to what becomes a cross join when syntax `select ... from participant_contracts pc, participant_contract_witnesses pcw` is used without a column on which to join on.

The following action was taken to fix the regression.  This approach allows for cross-platform syntax which works equivalently on Postgres, Oracle and H2

- Use join for active contracts queries (equivalent to inner join) specifying the column to match on
- Explicitly specify the table from which the contract_id column should be referenced
- Remove references to other columns with table alias as they are not ambiguous and the alias is therefore unnecessary

We have avoided the use of natural join as the column referencing required for natural join is contradictory between H2 and Oracle

- H2 treats natural join as an equivalent to an inner join with an implied `on` on the matching column(s) between the tables, but still requires you to specifically disambiguate the source table for the column in your select and where clauses
- Oracle will use a natural join to collapse common columns between the joined tables, but prevents you from referencing these columns with their source table alias (as it internally takes care of collapsing to a disambiguated 3rd joined table)

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsReader.scala
@@ -57,9 +57,9 @@ private[dao] sealed class ContractsReader(
 
         dispatcher.executeSql(metrics.daml.index.db.lookupContractByKeyDbMetrics) {
           implicit connection =>
-            SQL"""select pc.contract_id from #$contractsTable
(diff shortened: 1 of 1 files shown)


tudor-da (approved):
LGTM, performance tests look good as well

gerolf-da (approved):
Thanks for the quick fix, @dasormeter!

PR #9674, opened 2021-05-12
Split daml test-script code into a separate library

cocreature:
Currently we have the following dependency chain:

1. The compiler depends on the scenario service for daml test
2. The scenario service depends on Daml Script
3. Daml Script depends on the Sandbox code only for daml test-script

The last one can easily be split. The scenario service does not care
about this code.

This means that now if we change ledger code, at least not all
compiler tests are going to rerun.

Verified that this actually breaks the dependency fully via

```
bazel query 'somepath(//compiler/damlc/tests:packaging, //ledger/participant-integration-api/...)'
```

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-assistant/daml-sdk/util.bzl
+++ b/daml-assistant/daml-sdk/util.bzl
@@ -3,7 +3,8 @@
 
 def deps(edition):
     return [
-        "//daml-script/runner:script-runner-lib-{}".format(edition),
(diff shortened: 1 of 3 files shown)


dasormeter (approved):
LGTM

PR #9679, opened 2021-05-13
Never create a rollback node containing pre-exception descendants.

nickchapman-da:
Never create a rollback node containing descendant nodes with a version  older than `minExceptions`.
Instead throw an (old-style) uncatchable exception.

changelog_begin
changelog_end

part of #8020

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -6,8 +6,8 @@ package speedy
 
 import com.daml.lf.data.Ref.{ChoiceName, Location, Party, TypeConName}
 import com.daml.lf.data.{BackStack, ImmArray, Ref, Time}
-import com.daml.lf.language.Ast
 import com.daml.lf.ledger.{Authorize, FailedAuthorization}
+import com.daml.lf.speedy.SError.DamlEUnhandledException
 import com.daml.lf.transaction.{
   ContractKeyUniquenessMode,
   GenTransaction,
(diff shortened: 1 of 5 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -647,22 +646,49 @@ private[lf] case class PartialTransaction(
...
+        }
         copy(
           context = info.parent.addRollbackChild(info.nodeId, context.nextActionChildIdx),
           nodes = nodes.updated(info.nodeId, rollbackNode),
         ).resetActiveState(info.beginState)
       case _ => throw new RuntimeException("rollbackTry called in non-catch context")
     }
+  }
+
+  private def oldestContainedVersion(top: Node): TxVersion = {
```
cocreature, line 669: This is somewhat inefficient. You are retraversing everything under the rollback node. We could instead keep track of the minimum below each subtree as we build up the transaction and avoid retraversing anything.
Fine for this PR but I think we probably wan to revisit this later.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ExceptionTest.scala
```diff
@@ -536,11 +541,81 @@ class ExceptionTest extends AnyWordSpec with Matchers with TableDrivenPropertyCh
     }
   }
 
+  "rollback of creates" should {
```
cocreature, line 544: Maybe throw some other node types in there as well, at least an exercise.

cocreature (approved):
Thanks!

PR #9683, opened 2021-05-14
Differentiate between negative inputs & errors in contractKeyInputs

cocreature:
For better or for worse, kvutils validation insists on treating a
negative input from a create different to a negative key lookup (and
to make things more annoying the first one will never blow up during
submission only during validation).

While, nobody seems to argue all that strongly that the current errors
are very sensible, we agreed to do it in two steps:

1. Expose enough information in `contractKeyInputs` so that we can use
   it in kvutils while preserving the current error semantics.
2. Revisit contract key error handling across ledgers which is
   currently an inconsistent mess.

changelog_begin
changelog_end

.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


remyhaemmerle-da (approved):
LGTM

PR #9687, opened 2021-05-14
Lift constraint tuples up to a type synonym in data-dependencies.

### LINKED ISSUE #9663: Data-dependencies do not preserve constraint tuple constraints.
Given a constraint tuple constraint on a function like this, in a data-dependency:

```Haskell
module A

type MyConstraintTuple t = (Show t, Eq t)

constraintTupleFn : MyConstraintTuple t => t -> Text
constraintTupleFn x = show x
```

And a simple attempt to use it in the main project:

```Haskell
module B where
import A

useConstraintTupleFn : Text
useConstraintTupleFn = constraintTupleFn (10 : Int)
```

We get LF typechecker errors:

```
error type checking value B.useConstraintTupleFn:
  type mismatch:
  * expected type:
  <s_1 :
  <m_showsPrec : Unit -> Int64 -> Int64 -> Text -> Text;
  m_show : Unit -> Int64 -> Text;
  m_showList : Unit -> List Int64 -> Text -> Text>;
  s_2 :
  <m_== : Unit -> Int64 -> Int64 -> Bool;
  m_/= : Unit -> Int64 -> Int64 -> Bool>>
  * found type:
  <m_showsPrec : Unit -> Int64 -> Int64 -> Text -> Text;
  m_show : Unit -> Int64 -> Text;
  m_showList : Unit -> List Int64 -> Text -> Text>
  * expression:
  fb8198cecd7ab5bab0439c7f22f1c760651c7c60d5778b589d72aaa43693f2f9:GHC.Show:$fShowInt
```

Indicating that the stub being generated by data-dependencies is completely wrong.

sofiafaro-da:
Fixes #9663 and adds a regression test.

GHC was crushing the constraint tuples into regular constraints, so the function caller no longer matched the original LF API. By lifting the constraint tuple into a type synonym, we eliminate this effect.

Part of the issue here is that we're losing the original constraint tuple type synonym during LFConversion. I'll open a separate issue for that, since I think it's also worth fixing that going forward.

(Preserving the type synonym would solve this issue for newly compiled packages. By contrast, the fix in this PR works for both new and old packages.)

changelog_begin

- [Daml Compiler] Fixed a data-dependencies bug where functions in a data-dependency that used a constraint tuple constraint (e.g. `Template t`) could not be directly invoked.

changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

PR #9688, opened 2021-05-14
Add race condition tests for exceptions

cocreature:
This PR addresses
https://github.com/digital-asset/daml/pull/9400#pullrequestreview-634770251
and adds tests that match RaceConditionTests but to the read side in a
rollback (we cannot do writes in rollbacks, they are rolled back :)).

The tests are as close as possible to the other race condition tests
to ease maintenance and reduce confusion.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionRaceConditionIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionRaceConditionIT.scala
@@ -10,11 +10,8 @@ import com.daml.ledger.api.testtool.infrastructure.participant.ParticipantTestCo
 import com.daml.ledger.api.testtool.infrastructure.RaceConditionTests._
 import com.daml.ledger.client.binding.Primitive
 import com.daml.ledger.test.semantic.ExceptionRaceTests._
-// import com.daml.timer.Delayed
 
-// import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
-// import scala.util.Success
 
 final class ExceptionRaceConditionIT extends LedgerTestSuite {
   raceConditionTest(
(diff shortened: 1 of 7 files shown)


ledger/sandbox-classic/BUILD.bazel
```diff
@@ -407,9 +407,6 @@ server_conformance_test(
         "--contract-id-seeding=testing-weak",
     ],
     servers = SERVERS,
-    test_tool_args = [
```
cocreature, line 410: the test suite is enabled by default. Specifying the includes manually would potentially speed it up since it only runs those tests but I think it’s too fragile and makes it easy to miss newly added test suites that should be included here so I dropped it.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ExceptionRaceConditionIT.scala
```diff
@@ -0,0 +1,178 @@
...
+package com.daml.ledger.api.testtool.suites
+
+import com.daml.ledger.api.testtool.infrastructure.Allocation._
+import com.daml.ledger.api.testtool.infrastructure.Assertions._
+import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
+import com.daml.ledger.api.testtool.infrastructure.participant.ParticipantTestContext
+import com.daml.ledger.api.testtool.infrastructure.RaceConditionTests._
+import com.daml.ledger.client.binding.Primitive
+import com.daml.ledger.test.semantic.ExceptionRaceTests._
+// import com.daml.timer.Delayed
```
tudor-da, line 13: Commented line

cocreature: duh c3510e62fcc0f1a7114c4447b125a4ee94857d7c

ledger/test-common/src/main/daml/semantic/ExceptionRaceTests.daml
```diff
@@ -0,0 +1,116 @@
...
+          catch
+            E -> pure ()
+
+template ExerciseWrapper
+  with
+    owner : Party
+  where
+    signatory owner
+    controller owner can
+      nonconsuming ExerciseWrapper_ExerciseConsumingRollback : ()
```
tudor-da, line 98: `nonconsuming`?

cocreature: No the `consuming` vs `nonconsuming` here refers to the exercise in the body of this choice. For this one it’s `Archive` which is consuming. For the one below it’s `ContractWithKey_Exercise` which is non-consuming.

tudor-da: Ah right! Thanks for the clarification

tudor-da (approved):
LGTM

PR #9693, opened 2021-05-14
Dpp 391 todo cleanup store backend

nmarton-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
@@ -5,11 +5,16 @@ package com.daml.platform.store.backend
 
 import java.time.Instant
 
-trait DBDTOV1
+import com.daml.scalautil.NeverEqualsOverride
+
+sealed trait DBDTOV1
+    extends NeverEqualsOverride
+    with Product
+    with Serializable // to aid type inference for case class implementors
 
 object DBDTOV1 {
 
-  case class EventDivulgence(
+  final case class EventDivulgence(
       event_offset: Option[String],
       command_id: Option[String],
       workflow_id: Option[String],
@@ -23,7 +28,7 @@ object DBDTOV1 {
       event_sequential_id: Long,
   ) extends DBDTOV1
 
-  case class EventCreate(
+  final case class EventCreate(
       event_offset: Option[String],
       transaction_id: Option[String],
       ledger_effective_time: Option[Instant],
@@ -48,7 +53,7 @@ object DBDTOV1 {
       event_sequential_id: Long,
   ) extends DBDTOV1
 
-  case class EventExercise(
+  final case class EventExercise(
       consuming: Boolean,
       event_offset: Option[String],
       transaction_id: Option[String],
@@ -75,67 +80,63 @@ object DBDTOV1 {
       event_sequential_id: Long,
   ) extends DBDTOV1
 
-  // TODO append-only: wartremover complained about having Array-s in case classes. I would prefer case classes. can we work that somehow around? Similarly in other DTO cases...
-  // TODO append-only: there are some options:
-  //   - mixing in SomeArrayEquals if we need array equality for some reason: would be proper if we move SomeArrayEquals out from speedy codebase to scalalib first.
-  //   - spawning somewhere something like trait NeverEqualsOverride { override equals(o: Object): Boolean = false }, and mixing in these classes
-  class ConfigurationEntry(
-      val ledger_offset: String,
-      val recorded_at: Instant,
-      val submission_id: String,
-      val typ: String,
-      val configuration: Array[Byte],
-      val rejection_reason: Option[String],
+  final case class ConfigurationEntry(
+      ledger_offset: String,
+      recorded_at: Instant,
+      submission_id: String,
+      typ: String,
+      configuration: Array[Byte],
+      rejection_reason: Option[String],
   ) extends DBDTOV1
 
-  class PackageEntry(
-      val ledger_offset: String,
-      val recorded_at: Instant,
-      val submission_id: Option[String],
-      val typ: String,
-      val rejection_reason: Option[String],
+  final case class PackageEntry(
+      ledger_offset: String,
+      recorded_at: Instant,
+      submission_id: Option[String],
+      typ: String,
+      rejection_reason: Option[String],
   ) extends DBDTOV1
 
-  class Package(
-      val package_id: String,
-      val upload_id: String,
-      val source_description: Option[String],
-      val size: Long,
-      val known_since: Instant,
-      val ledger_offset: String,
-      val _package: Array[Byte],
+  final case class Package(
+      package_id: String,
+      upload_id: String,
+      source_description: Option[String],
+      size: Long,
+      known_since: Instant,
+      ledger_offset: String,
+      _package: Array[Byte],
   ) extends DBDTOV1
 
-  class PartyEntry(
-      val ledger_offset: String,
-      val recorded_at: Instant,
-      val submission_id: Option[String],
-      val party: Option[String],
-      val display_name: Option[String],
-      val typ: String,
-      val rejection_reason: Option[String],
-      val is_local: Option[Boolean],
+  final case class PartyEntry(
+      ledger_offset: String,
+      recorded_at: Instant,
+      submission_id: Option[String],
+      party: Option[String],
+      display_name: Option[String],
+      typ: String,
+      rejection_reason: Option[String],
+      is_local: Option[Boolean],
   ) extends DBDTOV1
 
-  class Party(
-      val party: String,
-      val display_name: Option[String],
-      val explicit: Boolean,
-      val ledger_offset: Option[String],
-      val is_local: Boolean,
+  final case class Party(
+      party: String,
+      display_name: Option[String],
+      explicit: Boolean,
+      ledger_offset: Option[String],
+      is_local: Boolean,
   ) extends DBDTOV1
 
-  class CommandCompletion(
-      val completion_offset: String,
-      val record_time: Instant,
-      val application_id: String,
-      val submitters: Set[String],
-      val command_id: String,
-      val transaction_id: Option[String],
-      val status_code: Option[Int],
-      val status_message: Option[String],
+  final case class CommandCompletion(
+      completion_offset: String,
+      record_time: Instant,
+      application_id: String,
+      submitters: Set[String],
+      command_id: String,
+      transaction_id: Option[String],
+      status_code: Option[Int],
+      status_message: Option[String],
   ) extends DBDTOV1
 
-  class CommandDeduplication(val deduplication_key: String) extends DBDTOV1
+  final case class CommandDeduplication(deduplication_key: String) extends DBDTOV1
 
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
 import com.daml.platform.store.DbType
 import com.daml.platform.store.backend.postgresql.PostgresStorageBackend
 
-// TODO append-only: add detailed scaladoc
+/** Encapsulates the interface which hides database technology specific implementations for parallel ingestion.
+  *
+  * @tparam DB_BATCH Since parallel ingestion comes also with batching, this implementation specific type allows separation of the CPU intensive batching operation from the pure IO intensive insertBatch operation.
+  */
 trait StorageBackend[DB_BATCH] {
+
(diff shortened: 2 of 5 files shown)


nmarton-da:
Note for reviewers: best approach might be go commit by commit, separate commits are compiling and holding self-contained part of the work.
In the separate commit descriptions I tried to leave sufficient justifications if TODOs were just removed.
Some commits are closely related to independent discussion as:
@S11001001 @stefanobaghino-da https://github.com/digital-asset/daml/pull/9693/commits/e4a19afb3afe3050832800c1f9a4bb03e4939144
@meiersi-da https://github.com/digital-asset/daml/pull/9693/commits/9de24ec6dab30e2a336ed4ba46aac88d710e2128

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/NeverEqualsOverride.scala
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.scalautil
+
+/** This trait is useful to be mixed in if:
+  * - equality semantics for the base type is not an issue
+  * - class has non-comparable typed fields
+  *
+  * Practical usage: mixing in in DTO ADT top trait, which has case class implementors having Array fields: disables equality semantics, and prevents Wartremover warnings.
```
S11001001, line 10: For arrays specifically, I would use `SomeArrayEquals` for this (this can be promoted from the Speedy library to `scala-utils`), along with its tests.

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/NeverEqualsOverride.scala
```diff
@@ -0,0 +1,14 @@
...
+  */
+trait NeverEqualsOverride {
+  override def equals(o: Any): Boolean = false
```
S11001001, line 13: It might be better to simulate `java.lang.Object`'s behavior here.

```suggestion
trait NeverEqualsOverride {
  override final def equals(o: Any): Boolean = o match {
    case ar: AnyRef => this eq ar
    case _ => false
  }
  override final def hashCode = System.identityHashCode(this)
```

S11001001: If you find that disagreeable, then alternatively

```suggestion
trait NeverEqualsOverride extends Equals {
  override final def canEqual(o: Any) = false
  override final def equals(o: Any): Boolean = false
```

S11001001: If you're really committed to this, you should make `equals` throw an exception. 👻

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/NeverEqualsOverride.scala
nmarton-da: I would completely agree if the purpose would be to support equality of arrays. Here the purpose is different: we explicitly do not care about equality (hence the name)

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
```diff
@@ -75,67 +80,63 @@ object DBDTOV1 {
...
-      val ledger_offset: String,
-      val _package: Array[Byte],
+  final case class Package(
+      package_id: String,
+      upload_id: String,
+      source_description: Option[String],
+      size: Long,
+      known_since: Instant,
+      ledger_offset: String,
+      _package: Array[Byte],
```
tudor-da, line 107: why with `_`?

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
```diff
@@ -75,67 +80,63 @@ object DBDTOV1 {
...
-      val is_local: Option[Boolean],
+  final case class PartyEntry(
+      ledger_offset: String,
+      recorded_at: Instant,
+      submission_id: Option[String],
+      party: Option[String],
+      display_name: Option[String],
+      typ: String,
+      rejection_reason: Option[String],
+      is_local: Option[Boolean],
```
tudor-da, line 118: Can you use camelCase?

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
```diff
@@ -5,11 +5,16 @@ package com.daml.platform.store.backend
 
 import java.time.Instant
 
-trait DBDTOV1
+import com.daml.scalautil.NeverEqualsOverride
+
+sealed trait DBDTOV1
```
tudor-da, line 10: What does the V1 at the end signify? Can we drop it?

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/NeverEqualsOverride.scala
tudor-da: Isn't this rather dangerous for future usage? What if someone comes and starts implementing tests wrongly relying on the case class contract?

tudor-da (approved):
LGTM, but I'm looking forward to understanding some points

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/NeverEqualsOverride.scala
nmarton-da: well, the test would fail I suppose? :) or the same what would happen if you would not have wartremover and would want to compare two case classes with fields without equality on the contract...
This guy is more about the wartremover warnings than case classes/contracts in general.
Personally I would rather go for this relaxation, so we can reap the benefit and cut the noise, but happy to change if you see fit.

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
nmarton-da: yes, and also we will, FYI: not part of this review

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
nmarton-da: original intent was: to signal correspondence to db table columns, and make visual comparison easier.
I understand this violates scala code-style principles, for me the benefits weigh more than the drawbacks, and you are the first to complain :stuck_out_tongue: 
FYI: this is merely a refactoring, not part of this review

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
nmarton-da: try it without :trollface: :)
I think because it looks better then ``package`` :thinking: but that is also debatable :)
FYI: not part of this review

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
mziolekda: I second that question. Could we introduce camelCase in the grand renaming PR?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+/** Encapsulates the interface which hides database technology specific implementations for parallel ingestion.
+  *
+  * @tparam DB_BATCH Since parallel ingestion comes also with batching, this implementation specific type allows separation of the CPU intensive batching operation from the pure IO intensive insertBatch operation.
+  */
 trait StorageBackend[DB_BATCH] {
+
+  /** The CPU intensive batching operation hides the batching logic, and the mapping to the database specific representation of the inserted data.
+    * This should be pure CPU logic without IO.
+    *
+    * @param dbDtos is a collection of DBDTOV1 which from the batch is formed
```
mziolekda, line 21:
```suggestion
    * @param dbDtos is a collection of DBDTOV1 from which the batch is formed
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * @return the database-specific batch DTO, which can be inserted via insertBatch
+    */
   def batch(dbDtos: Vector[DBDTOV1]): DB_BATCH
+
+  /** Using a JDBC connection, a batch will be inserted into the database.
+    * No significant CPU load, mostly blocking JDBC communication with the database backend.
+    *
+    * @param connection to be used for inserting a batch
```
mziolekda, line 29:
```suggestion
    * @param connection to be used when inserting the batch
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * @param batch to be inserted
+    */
   def insertBatch(connection: Connection, batch: DB_BATCH): Unit
+
+  /** This method is used to update the parameters table: setting the new observable ledger-end, and other parameters.
+    * No significant CPU load, mostly blocking JDBC communication with the database backend.
+    *
+    * @param connection to be used for upating the parameters table
```
mziolekda, line 37:
```suggestion
    * @param connection to be used when updating the parameters table
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * @param params the actual parameters
+    */
   def updateParams(connection: Connection, params: StorageBackend.Params): Unit
+
+  /** Custom initialization code before the start of ingestion.
```
mziolekda, line 42:
```suggestion
  /** Custom initialization code before the start of an ingestion.
  * This method is responsible for the recovery after a possibly non-graceful stop of previous indexing.
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * This method is responsible to recover after a possibly non-graceful stop of previous indexing.
```
mziolekda, line 43:
```suggestion
    * This method is responsible for the recovery after a possibly non-graceful stop of previous indexing.
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * No significant CPU load, mostly blocking JDBC communication with the database backend.
+    *
+    * @param connection to be used for initialization
```
mziolekda, line 46:
```suggestion
    * @param connection to be used when initializing
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    * @return the actual LedgerEnd, which should be the basis for further indexing.
```
mziolekda, line 47: Is there a different form of ledgerEnd other than the "actual"? If not, this adjective can be dropped

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -9,12 +9,51 @@ import com.daml.ledger.participant.state.v1.Offset
...
+    */
   def initialize(connection: Connection): StorageBackend.LedgerEnd
+
+  /** Query the actual ledger-end, read from the parameters table.
```
mziolekda, line 51:
```suggestion
  /** Query the actual ledgerEnd, read from the parameters table.
```
Same comment as above above about the "actual"

ledger/participant-integration-api/src/main/scala/platform/store/backend/DBDTOV1.scala
tudor-da: Ah reserved word. But still, first thing that comes to mind is an internal `var` that should not be exposed. Looks confusing. Maybe `packagePayload` would help

mziolekda (approved):
Looks good thanks

PR #9698, opened 2021-05-17
Drop sanitize, add MagicHash in data-dependencies.

sofiafaro-da:
The only thing `sanitize` does is mangle the names of identifiers and
types that end in `#`. That can't be a good thing. This PR removes
sanitize and enables MagicHash in data-dependencies, meaning names can
end in `#` without causing issues. (MagicHash also enables unboxed
literals, but that doesn't matter here.)

In practice, sanitize was only affecting `GHC.Prim.Void#`, which we
define because GHC sometimes references it. AFAIK there's no other
use of `#` in the standard library that we don't rewrite away.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

PR #9701, opened 2021-05-17
Fix client_server_build/test with port file

aherrmann-da:
The default runner created the port-file for sandbox to write it's port into. (Relevant with dynamic port, i.e. `--port 0`). However, sandbox will not overwrite an existing port-file. Sandbox silently ignored this error and left the port-file untouched.

This PR
- Changes the `client_server_*` runner to create a temporary directory instead of a file and lets sandbox create a fresh port file within that directory.
- Changes sandbox such that it fails if the port-file cannot be written to, instead of silently ignoring the error.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
@@ -30,10 +30,12 @@ import com.daml.platform.packages.InMemoryPackageStore
 import com.daml.platform.services.time.TimeProviderType
 import com.daml.platform.store.LfValueTranslationCache
 import com.daml.ports.{Port, PortFiles}
+import com.daml.resources.FutureResourceOwner
 import io.grpc.{BindableService, ServerInterceptor}
+import scalaz.{-\/, \/-}
 
 import scala.collection.immutable
-import scala.concurrent.ExecutionContextExecutor
+import scala.concurrent.{ExecutionContextExecutor, Future}
 
 // Main entry point to start an index server that also hosts the ledger API.
 // See v2.ReferenceServer on how it is used.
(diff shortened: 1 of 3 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
```diff
@@ -165,8 +167,15 @@ final class StandaloneApiServer(
       .fold({ case (err, file) => sys.error(s"Could not load package $file: $err") }, identity)
   }
 
-  private def writePortFile(port: Port): Unit =
-    config.portFile.foreach { path =>
-      PortFiles.write(path, port)
+  private def writePortFile(port: Port): Future[Unit] = {
```
cocreature, line 170: Does the change to return a future matter or is the only crucial change to throw an exception? No objection to keep it just trying to understand the bug.

cocreature (approved):
Thanks!

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
aherrmann-da: The only crucial part is throwing the exception. The future is just there to make this fit into the resource owner for-block on the calling side. It seemed cleaner to me to integrate it into that block rather than to have a line in the yield part throw an exception.

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
```diff
@@ -121,8 +123,8 @@ final class StandaloneApiServer(
         servicesExecutionContext,
         metrics,
       )
+      _ <- new FutureResourceOwner[ResourceContext, Unit](() => writePortFile(apiServer.port))
```
SamirTalwar-DA, line 126: This could be:

```scala
      _ <- ResourceOwner.forFuture(() => writePortFile(apiServer.port))
```

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
SamirTalwar-DA: You could use a `Try` and then use `ResourceOwner.forTry` on the other side.

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
aherrmann-da: @SamirTalwar-DA Thank you, that's a good suggestion. I've implemented the `forTry` approach [here](https://github.com/digital-asset/daml/pull/9720).

PR #9709, opened 2021-05-17
Split DA.Internal.Exception.Types into three packages.

sofiafaro-da:
I think this is a good idea because it means the user can disambiguate between these modules in daml. All three types define a `message` field, so if the programmer wants to refer to the field for one of these types specifically, it's nice to have a canonical way to disambiguate them (by module name).

Also the module names show up in the error message for uncaught exceptions. These names are a bit nicer than "DA.Internal.Exception.Types".

Decoupling these types just sounds like a good idea going forward.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


compiler/damlc/daml-prim-src/DA/Exception/GeneralError.daml
```diff
@@ -0,0 +1,24 @@
...
+import GHC.Types ()
+
+#else
+
+-- | MOVE DA.Exception
+module DA.Exception.GeneralError where
+
+import GHC.Types (Text)
+
+-- | Exception raised by `error`.
```
cocreature, line 21: we should probably also revisit the name of this soonish. I think `GeneralError` sounds a bit too general (no pun intended) for something only thrown by calls to `error`.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandServiceIT.scala
```diff
@@ -301,7 +301,7 @@ final class CommandServiceIT extends LedgerTestSuite {
         Status.Code.INVALID_ARGUMENT,
         Some(
           Pattern.compile(
-            "Command interpretation error in LF-DAMLe: Interpretation error: Error: (User abort: Assertion failed\\.|unhandled exception: DA.Internal.Exception.Types:GeneralError@6ad87e6c\\{ message = \"Assertion failed\" \\}\\.) Details: Last location: \\[[^\\]]*\\], partial transaction: root node"
+            "Command interpretation error in LF-DAMLe: Interpretation error: Error: (User abort: Assertion failed\\.|unhandled exception: DA.Exception.GeneralError:GeneralError@[0-9a-f]*\\{ message = \"Assertion failed\" \\}\\.) Details: Last location: \\[[^\\]]*\\], partial transaction: root node"
```
cocreature, line 304: maybe be more general here and just drop the match on the exception type? I don’t think that’s really useful in this context.

cocreature (approved):
Thank you!

compiler/damlc/daml-prim-src/DA/Exception/GeneralError.daml
sofiafaro-da: What do you suggest? Maybe `Abort` or something, since it's also raised by `abort`?

sofiafaro-da: Let's discuss during the meeting today :)

sofiafaro-da:
> > All three types define a `message` field, so if the programmer wants to refer to the field for one of these types specifically, it's nice to have a canonical way to disambiguate them (by module name).
> 
> Don’t we have record dot syntax for disambiguating here?
> (not at all opposed to the change regardless, reviewing now)

It's really minor, but:

I think the only real issue is if you try to use a field selector function `message` from one of these types. The field selector also overlaps with the `message` method from the `HasMessage` typeclass, so there's no reason to use the `message` field selectors (and we don't reexport them in `DA.Exception`). But if for some reason you want to refer to a specific field selector function for one of these types, it would be impossible to disambiguate without having separate modules. 

The second issue is if you happen to have these `message` functions in scope (from the same module), the error messages you get from GHC are a bit confusing.

PR #9712, opened 2021-05-17
Normalize rollback nodes when a transaction is constructed

nickchapman-da:
Done:
- Implemented normalization rules as new code pass in `NormalizeRollbacks`
- Fairly extensive tests in `NormalizeRollbacksSpec`
- ensure pre-order traversal of node-ids is in increasing order; and starts from 0, fixes some existing tests which depend on this
- make normalization pass stack-safe (CPS & trampolines)
- enable normalization pass in `PartialTransaction`

part of #8020

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
@@ -1,0 +1,25 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf
+package data
+
+import scala.annotation.tailrec
+
+// Trampolines for stack-safety
+object Trampoline {
+
+  sealed trait Trampoline[A] {
+
+    @tailrec
+    final def bounce: A = this match {
+      case Land(a) => a
+      case Bounce(continue) => continue().bounce
+    }
+
+  }
+
+  final case class Land[A](a: A) extends Trampoline[A]
+  final case class Bounce[A](func: () => Trampoline[A]) extends Trampoline[A]
+
+}
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
@@ -7,9 +7,9 @@ package data
 import scala.annotation.tailrec
 
 // Trampolines for stack-safety
-object Trampoline {
+private[lf] object Trampoline {
 
-  sealed trait Trampoline[A] {
+  private[lf] sealed trait Trampoline[A] {
 
     @tailrec
     final def bounce: A = this match {
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
@@ -13,12 +13,12 @@ import scala.annotation.tailrec
 
 private[lf] object NormalizeRollbacks {
 
-  type Nid = NodeId
-  type Cid = Value.ContractId
-  type TX = GenTransaction[Nid, Cid]
-  type Node = GenNode[Nid, Cid]
-  type LeafNode = LeafOnlyActionNode[Cid]
-  type ExeNode = NodeExercises[Nid, Cid]
+  private[this] type Nid = NodeId
+  private[this] type Cid = Value.ContractId
+  private[this] type TX = GenTransaction[Nid, Cid]
+  private[this] type Node = GenNode[Nid, Cid]
+  private[this] type LeafNode = LeafOnlyActionNode[Cid]
+  private[this] type ExeNode = NodeExercises[Nid, Cid]
 
   // Normalize a transaction so rollback nodes satisfy the normalization rules.
   // see `makeRoll` below
@@ -92,7 +92,7 @@ private[lf] object NormalizeRollbacks {
 
   //   rule #2/#3 overlap: ROLL [ ROLL [ xs… ] ] -> ROLL [ xs… ]
 
-  private def makeRoll[R](norms: List[Norm])(k: List[Norm] => Tramp[R]): Tramp[R] = {
+  private[this] def makeRoll[R](norms: List[Norm])(k: List[Norm] => Tramp[R]): Tramp[R] = {
     caseNorms(norms) match {
       case Case.Empty =>
         // normalization rule #1
@@ -134,7 +134,7 @@ private[lf] object NormalizeRollbacks {
 
   // There is no connection between the ids in the original and normalized transaction.
 
-  private case class State(index: Int, nodeMap: Map[Nid, Node]) {
+  private[this] case class State(index: Int, nodeMap: Map[Nid, Node]) {
 
     def next[R](k: (State, Nid) => Tramp[R]): Tramp[R] = {
       k(State(index + 1, nodeMap), NodeId(index))
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
@@ -57,8 +57,9 @@ private[lf] object NormalizeRollbacks {
           }
         }
         val norms = traverseNids(rootsOriginal.toList)
-        val (finalState, roots) = pushNorms(initialState, norms)
-        GenTransaction(finalState.nodeMap, ImmArray(roots))
+        pushNorms(initialState, norms) { (finalState, roots) =>
+          GenTransaction(finalState.nodeMap, ImmArray(roots))
+        }
     }
   }
 
@@ -68,58 +69,73 @@ private[lf] object NormalizeRollbacks {
   // There is no connection between the ids in the original and normalized transaction.
 
   private case class State(index: Int, nodeMap: Map[Nid, Node]) {
-    def next: (State, Nid) = (State(index + 1, nodeMap), NodeId(index))
-    def push(nid: Nid, node: Node): (State, Nid) =
-      (State(index, nodeMap = nodeMap + (nid -> node)), nid)
+
+    def next[R](k: (State, Nid) => R): R = {
+      k(State(index + 1, nodeMap), NodeId(index))
+    }
+
+    def push[R](nid: Nid, node: Node)(k: (State, Nid) => R): R = {
+      k(State(index, nodeMap = nodeMap + (nid -> node)), nid)
+    }
+
   }
 
   private val initialState = State(100, Map.empty)
 
   // The `push*` functions convert the Canonical types to the tx being collected in State.
 
-  private def pushAct(s0: State, x: Norm.Act): (State, Nid) = {
-    val (s, me) = s0.next
-    x match {
-      case Norm.Leaf(node) =>
-        s.push(me, node)
-      case Norm.Exe(exe, subs) =>
-        val (s1, children) = pushNorms(s, subs)
-        val node = exe.copy(children = ImmArray(children))
-        s1.push(me, node)
+  private def pushAct[R](s: State, x: Norm.Act)(k: (State, Nid) => R): R = {
+    s.next { (s, me) =>
+      x match {
+        case Norm.Leaf(node) =>
+          s.push(me, node)(k)
+        case Norm.Exe(exe, subs) =>
+          pushNorms(s, subs) { (s, children) =>
+            val node = exe.copy(children = ImmArray(children))
+            s.push(me, node)(k)
+          }
+      }
     }
   }
 
-  private def pushRoll(s0: State, x: Norm.Roll): (State, Nid) = {
-    val (s, me) = s0.next
-    x match {
-      case Norm.Roll1(act) =>
-        val (s1, child) = pushAct(s, act)
-        val node = NodeRollback(children = ImmArray(List(child)))
-        s1.push(me, node)
-      case Norm.Roll2(h, m, t) =>
-        val (s1, hh) = pushAct(s, h)
-        val (s2, mm) = pushNorms(s1, m)
-        val (s3, tt) = pushAct(s2, t)
-        val children = List(hh) ++ mm ++ List(tt)
-        val node = NodeRollback(children = ImmArray(children))
-        s3.push(me, node)
+  private def pushRoll[R](s: State, x: Norm.Roll)(k: (State, Nid) => R): R = {
+    s.next { (s, me) =>
+      x match {
+        case Norm.Roll1(act) =>
+          pushAct(s, act) { (s, child) =>
+            val node = NodeRollback(children = ImmArray(List(child)))
+            s.push(me, node)(k)
+          }
+        case Norm.Roll2(h, m, t) =>
+          pushAct(s, h) { (s, hh) =>
+            pushNorms(s, m) { (s, mm) =>
+              pushAct(s, t) { (s, tt) =>
+                val children = List(hh) ++ mm ++ List(tt)
+                val node = NodeRollback(children = ImmArray(children))
+                s.push(me, node)(k)
+              }
+            }
+          }
+      }
     }
   }
 
-  private def pushNorm(s: State, x: Norm): (State, Nid) = {
+  private def pushNorm[R](s: State, x: Norm)(k: (State, Nid) => R): R = {
     x match {
-      case act: Norm.Act => pushAct(s, act)
-      case roll: Norm.Roll => pushRoll(s, roll)
+      case act: Norm.Act => pushAct(s, act)(k)
+      case roll: Norm.Roll => pushRoll(s, roll)(k)
     }
   }
 
-  private def pushNorms(s: State, xs: List[Norm]): (State, List[Nid]) = {
+  private def pushNorms[R](s: State, xs: List[Norm])(k: (State, List[Nid]) => R): R = {
     xs match {
-      case Nil => (s, Nil)
+      case Nil => k(s, Nil)
       case x :: xs =>
-        val (s1, y) = pushNorm(s, x)
-        val (s2, ys) = pushNorms(s1, xs)
-        (s2, y :: ys)
+        pushNorm(s, x) { (s, y) =>
+          pushNorms(s, xs) { (s, ys) =>
+            k(s, y :: ys)
+          }
+        }
     }
   }
 
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
@@ -1,0 +1,188 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf
+package speedy
+
+import com.daml.lf.transaction.{NodeId, GenTransaction}
+import com.daml.lf.transaction.Node.{
+  GenNode,
+  GenActionNode,
+  NodeRollback,
+  NodeExercises,
+  LeafOnlyActionNode,
+}
+import com.daml.lf.value.Value
+import com.daml.lf.data.ImmArray
+
+private[lf] object NormalizeRollbacks {
+
+  type Nid = NodeId
+  type Cid = Value.ContractId
+  type TX = GenTransaction[Nid, Cid]
+  type Node = GenNode[Nid, Cid]
+  type ActNode = GenActionNode[Nid, Cid]
+
+  // Normalize a transaction so rollback nodes satisfy the normalization rules.
+  // see `makeRoll` below
+
+  def normalizeTx(txOriginal: TX): TX = {
+
+    // Here we generate fresh node-ids for the normalized transaction.
+    val ids = Iterator.from(0).map(NodeId(_)) // perhaps we could reuse orig nids
+
+    // There is no connection between the ids in the pre and post transaction.
+    // Semantically this is fine: The ids are just of tie the tree structure together.
+    // But perhaps for ease of debugging we might consider trying to reuse existing ids.
+
+    // We accumulate (imperatively!) a nodes-map for the normalized transaction
+    var nodes: Map[Nid, Node] = Map.empty
+
+    def pushNode(node: Node): Nid = {
+      val nodeId = ids.next()
+      nodes += (nodeId -> node)
+      nodeId
+    }
+
+    def pushNodes(nodes: List[Node]): List[Nid] = nodes.map(pushNode)
+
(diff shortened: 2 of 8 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+          // normalization rule #2 or #3
+          List(roll)
+
+        case Case.Single(act: Norm.Act) =>
+          // no rule
+          List(Norm.Roll1(act))
+
+        case Case.Multi(h: Norm.Roll, m, t) =>
+          // normalization rule #2
+          h :: makeRoll(m ++ List(t))
```
sofiafaro-da, line 99: This recursion here should also be made stack-safe.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+class NormalizeRollbacksSpec extends AnyWordSpec with Matchers with Inside {
+
+  import NormalizeRollbackSpec._
+
+  // TODO: https://github.com/digital-asset/daml/issues/8020
+  // Below we test a set of hand-constructed testcases. Hopefully we got all edge cases!
+  // But for more confidence we might consider adding scalacheck based testing.
+
+  // TODO: https://github.com/digital-asset/daml/issues/8020
+  // We should test that the `meaning` of a transaction is preserved by normalization.
```
sofiafaro-da, line 28: Good idea. "Meaning" here would be the transaction tree among actions (i.e. after removing/flattening all rollback nodes), as well as the consistency order and the set of rolled back actions. That's a lot, but we definitely want to guarantee we're preserving these.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  // normalization rules #2 and #3
+  test("mixed-rollback-nesting-A")(
+    Top(R(R(c1), c2, R(c3))),
+    Top(R(c1), R(c2, c3)),
+  )
+  test("mixed-rollback-nesting-B")(
+    Top(R(R(c1), c2, c3, R(c4))),
+    Top(R(c1), R(c2, c3, c4)),
+  )
+
```
sofiafaro-da, line 158: These tests look great! I would also add some more odd mixed rollback cases:

```
R(c1, R(R(c2), c3)) --> R(c1, R(c2), c3)
```

and 

```
R(R(c1, R(c2)), c3) --> R(c1, c2), R(c3)
```

sofiafaro-da (commented):
This normalization algorithm looks right, and the unit tests are extensive. Thank you!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+
+    import Canonical.{Norm, Case, caseNorms}
+    // The normalization phase works by constructing intermediate values which are
+    // `Canonical` in the sense that only properly normalized nodes can be represented.
+    // Although this doesn't ensure correctness, one class of bugs is avoided.
+
+    // The `force*` functions move from the world of Norms into standard transactions.
+    // This is also where node-ids for the resulting normalized transaction are generated.
+
+    def forceRoll(x: Norm.Roll): Node = {
```
cocreature, line 57: what exactly are we forcing here? I don’t understand the naming csheme.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+      final case object Empty extends Case
+      final case class Single(n: Norm) extends Case
+      final case class Multi(h: Norm, m: List[Norm], t: Norm) extends Case
+    }
+
+    def caseNorms(xs: List[Norm]): Case = {
+      xs.length match {
+        case 0 => Case.Empty
+        case 1 => Case.Single(xs(0))
+        case n => Case.Multi(xs(0), xs.slice(1, n - 1), xs(n - 1))
```
cocreature, line 184: Maybe worth using a `Vector` or an `ImmArray` instead of slicing & calling length on a single-linked list? @remyhaemmerle-da probably has a better idea what makes sense here.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+    // The List is necessary because the rules can:
+    // (1) drop nodes; (2) combine nodes (3) lift nodes from a lower level to a higher level.
+
+    // TODO: https://github.com/digital-asset/daml/issues/8020
+    // make this traversal code stack-safe
+
+    txOriginal match {
+      case GenTransaction(nodesOriginal, rootsOriginal) =>
+        def traverseNids(nids: List[Nid]): List[Norm] = nids.flatMap(traverseNid)
+        def traverseNid(nid: Nid): List[Norm] = traverseNode(nodesOriginal(nid))
```
cocreature, line 130: this doesn’t look stacksafe afaict.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+
+        def traverseNode(node: Node): List[Norm] = {
+          node match {
+
+            case NodeRollback(children) =>
+              makeRoll(traverseNids(children.toList))
+
+            case ex: NodeExercises[_, _] =>
+              val norms = traverseNids(ex.children.toList)
+              val children = pushNodes(forceNorms(norms))
```
cocreature, line 140: From what I can tell this might be where you’re breaking the node id order. Now children of an exercise will have a lower node id than their parent. As mentioned on slack, I think we need to preserve node ids = index in preorder at least for Daml Studio and at that point it’s probably easiest to preserve it everywhere (and it’s also useful for debugging over the ledger API).

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,188 @@
...
+    // `Canonical` in the sense that only properly normalized nodes can be represented.
+    // Although this doesn't ensure correctness, one class of bugs is avoided.
+
+    // The `force*` functions move from the world of Norms into standard transactions.
+    // This is also where node-ids for the resulting normalized transaction are generated.
+
+    def forceRoll(x: Norm.Roll): Node = {
+      x match {
+        case Norm.Roll1(act) =>
+          val child = pushNode(act.node)
```
cocreature, line 60: the logic around when we call `forceRoll` definitely looks like it doesn’t result in `pushNode` calls in prefix order either.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
```diff
@@ -0,0 +1,291 @@
...
+  // Shape: description of a transaction, with conversions to and from a real tx
+  sealed trait Shape
+  object Shape {
+
+    final case class Top(xs: List[Shape])
+    final case class Create(x: Long) extends Shape
+    final case class Exercise(x: List[Shape]) extends Shape
+    final case class Rollback(x: List[Shape]) extends Shape
+
+    def toTransaction(top: Top): TX = {
```
cocreature, line 212: this looks like we can probably simplify the code a fair bit by using the transaction builder.

cocreature (commented):
Nice work!
I think there are two main points we need to address before merging:
1. Node ids should be indices in preorder traversal. While this is not required for a valid transaction, this is very useful for users and we shouldn’t break it here.
2. We should fix the code to be stacksafe

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: done

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
nickchapman-da: added this, thanks

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: renamed `force*` to `push*` and updated explanatory comment

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: everything should be stack safe now

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: the node ordering is fixed now, and also tested

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: fixed

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
nickchapman-da: I looked at the transaction builder, and it look really awkward to use in this case.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
remyhaemmerle-da: calling length on List is probably a good idea. 
Why not using standard List pattern matching ?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+      final case object Empty extends Case
+      final case class Single(n: Norm) extends Case
+      final case class Multi(h: Norm, m: List[Norm], t: Norm) extends Case
+    }
+
+    def caseNorms(xs: List[Norm]): Case = {
+      xs.length match {
+        case 0 => Case.Empty
+        case 1 => Case.Single(xs(0))
+        case n => Case.Multi(xs(0), xs.slice(1, n - 1), xs(n - 1))
```
remyhaemmerle-da, line 252:
```suggestion
      xs match {
        case Nil => Case.Empty
        case x0 :: Nil => Case.Single(h)
        case x0 :: x1 :: tail  => Case.Multi(x0, List(x1), tail)
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+      def run: A = {
+        this match {
+          case Land(a) => a
+          case Bounce(func) => func().run
+        }
+      }
+    }
+    final case class Land[A](a: A) extends Tramp[A]
+    final case class Bounce[A](func: () => Tramp[A]) extends Tramp[A]
+  }
```
remyhaemmerle-da, line 270: Could we factorize the Trampoline of `com.daml.lf.speedy.Anf` ? 
I will put it in `com.daml.lf.data`

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+import scala.annotation.tailrec
+
+private[lf] object NormalizeRollbacks {
+
+  type Nid = NodeId
+  type Cid = Value.ContractId
+  type TX = GenTransaction[Nid, Cid]
+  type Node = GenNode[Nid, Cid]
+  type LeafNode = LeafOnlyActionNode[Cid]
+  type ExeNode = NodeExercises[Nid, Cid]
```
remyhaemmerle-da, line 21: Why this is not private ? 

```suggestion
  private[this] type Nid = NodeId
  private[this]  type Cid = Value.ContractId
  private[this] type TX = GenTransaction[Nid, Cid]
  private[this] type Node = GenNode[Nid, Cid]
  private[this] type LeafNode = LeafOnlyActionNode[Cid]
  private[this] type ExeNode = NodeExercises[Nid, Cid]
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  // makeRoll: encodes the normalization transformation rules:
+  //   rule #1: ROLL [ ] -> ε
+  //   rule #2: ROLL [ ROLL [ xs… ] , ys… ] -> ROLL [ xs… ] , ROLL [ ys… ]
+  //   rule #3: ROLL [ xs… , ROLL [ ys… ] ] ->  ROLL [ xs… , ys… ]
+
+  //   rule #2/#3 overlap: ROLL [ ROLL [ xs… ] ] -> ROLL [ xs… ]
+
+  private def makeRoll[R](norms: List[Norm])(k: List[Norm] => Tramp[R]): Tramp[R] = {
```
remyhaemmerle-da, line 95:
```suggestion
  private[this] def makeRoll[R](norms: List[Norm])(k: List[Norm] => Tramp[R]): Tramp[R] = {
```
almost only cosmetic

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+      case Case.Multi(h: Norm.Act, m, t: Norm.Act) =>
+        // no rule
+        k(List(Norm.Roll2(h, m, t)))
+    }
+  }
+
+  private def pushIntoRoll(a1: Norm.Act, xs2: List[Norm], t: Norm.Roll): Norm.Roll = {
+    t match {
+      case Norm.Roll1(a3) => Norm.Roll2(a1, xs2, a3)
+      case Norm.Roll2(a3, xs4, a5) => Norm.Roll2(a1, xs2 ++ List(a3) ++ xs4, a5)
```
remyhaemmerle-da, line 128:
```suggestion
      case Norm.Roll2(a3, xs4, a5) => Norm.Roll2(a1, xs2 ++ a3 +: xs4, a5)
```
Anyway is concatenating lists a good idea ? 
Either add a comment to explain why this is not linear, either you use better data structure (e.g. Frontstack, Backstack, Vector ...)

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+    }
+  }
+
+  // State manages a counter for node-id generation, and accumulates the nodes-map for
+  // the normalized transaction
+
+  // There is no connection between the ids in the original and normalized transaction.
+
+  private case class State(index: Int, nodeMap: Map[Nid, Node]) {
```
remyhaemmerle-da, line 137:
```suggestion
  private[this] case class State(index: Int, nodeMap: Map[Nid, Node]) {
```
:)

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
remyhaemmerle-da: Just a bit curious about what is awkward.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
cocreature: That’s not the same. Before we had first, middle, last. Now you have first, second, rest.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  private def pushNorm[R](s: State, x: Norm)(k: (State, Nid) => Tramp[R]): Tramp[R] = {
+    x match {
+      case act: Norm.Act => pushAct(s, act)(k)
+      case roll: Norm.Roll => pushRoll(s, roll)(k)
+    }
+  }
+
+  private def pushNorms[R](s: State, xs: List[Norm])(
```
cocreature, line 201: I find the trampolined CPS style here pretty hard to follow. I hacked up a version of this that doesn’t use cps or trampolines and at least for me seems much easier to follow (ignore variable names and so on) https://github.com/digital-asset/daml/commit/2efd36a6f4efe69edbc2f48fd347cbd9b5ac916b. wdyt @remyhaemmerle-da @nickchapman-da? We can probably do something very similar for the first phase as well.

nickchapman-da: I think the nice thing about this approach is that it's a systematic transformation of the original code.
See:
https://github.com/digital-asset/daml/pull/9712/commits/4511623e529c9c3d022538f3bf078ce6916d214b
https://github.com/digital-asset/daml/pull/9712/commits/e520a935e6943023d0f165b450f71c7f63077335

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
remyhaemmerle-da: indeed.
We should then use `ImmArray` or `Vector`.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: done

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/NormalizeRollbacksSpec.scala
nickchapman-da: Mainly that I am building the transaction bottom up from a declarative description.
But the transaction builder seems to assume that I am building top-down. Create nodes, and then mutably attach children.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: We agreed to proceed with the trampoline approach, so long as Remy is happy. Which he is!

remyhaemmerle-da: I am fine with both approaches.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/NormalizeRollbacks.scala
nickchapman-da: switched to Vector to address this concern

remyhaemmerle-da (approved):
LGTM

cocreature (approved):
Thanks!

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package com.daml.lf
+package data
+
+import scala.annotation.tailrec
+
+// Trampolines for stack-safety
+object Trampoline {
+
+  sealed trait Trampoline[A] {
```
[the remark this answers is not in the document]
nickchapman-da: done

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Trampoline.scala
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf
+package data
+
+import scala.annotation.tailrec
+
+// Trampolines for stack-safety
+object Trampoline {
```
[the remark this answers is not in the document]
nickchapman-da: done

PR #9713, opened 2021-05-17
Upgrade protobuf-java and scalapb. [KVL-938]

SamirTalwar-DA:
Much of the complexity here is because I decided to generate ScalaPB classes without `unknownFields`.

`unknownFields` is very annoying because it means that extractors need an extra `_` for no good reason. It also breaks Circe's automatic derivation.

Including unknown fields is the default behavior in Protocol Buffers for Java now, and we'll probably have to reckon with it at some point, but let's kick that can down the road a little.

This didn't work for `com.google.rpc.Status`, probably because ScalaPB does something special for the messages defined in core Protocol Buffers and gRPC, so I had to add a single `_` to cope with that.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA:
At least on my machine, this fails to build with some errors that do not look at all related.

@S11001001, any idea why upgrading ScalaPB would affect Circe magic? 

```
ERROR: extractor/BUILD.bazel:75:17: scala //extractor:extractor failed: (Exit 1): scalac failed: error executing command
...
Execution platform: @io_tweag_rules_nixpkgs//nixpkgs/platforms:host
extractor/src/main/scala/com/digitalasset/extractor/writers/postgresql/StateHandler.scala:46: error: could not find implicit value for parameter encodeA1: io.circe.Encoder[com.daml.extractor.writers.postgresql.StateHandler.StartUpParameters]
  )(s => (s.ledgerId, s.startUpParameters, s.multiTableState, s.witnessedPackages))
   ^
extractor/src/main/scala/com/digitalasset/extractor/writers/postgresql/StateHandler.scala:53: error: could not find implicit value for parameter decodeA1: io.circe.Decoder[com.daml.extractor.writers.postgresql.StateHandler.StartUpParameters]
  )(Status.apply)
   ^
2 errors
Build failed
```

S11001001:
If scalapb makes different structures, anything different could need different implicits.

SamirTalwar-DA:
Oh, I didn't realize we were using generated structures here. (After staring at it for a while, too.)

Any objection to swapping `LedgerOffset` out with the one from _ledger-api-domain_?

S11001001:
> Any objection to swapping `LedgerOffset` out with the one from _ledger-api-domain_?

Not from me.

SamirTalwar-DA:
Decided to generate the ScalaPB classes without `unknownFields` instead. This was more complicated but probably worthwhile.

bazel_tools/scala_version.bzl
```diff
@@ -35,6 +35,9 @@ scala_artifacts = {
     },
 }
 
+scalapb_version = "0.11.3"
+scalapb_protoc_version = "0.9.2"
```
cocreature, line 39: Is there a good reason why these should be in here rather than in `bazel_tools/scala.bzl`? It doesn’t seem like they depend on the scala version.

bazel_tools/scalapb/com/daml/package.proto
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+import "scalapb/scalapb.proto";
+
+package com.daml;
+
+option (scalapb.options) = {
```
cocreature, line 10: not sure if using protobuf to configure your protobuf library is clever or annoying :slightly_smiling_face:

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/LedgerOffsetValidator.scala
```diff
@@ -32,12 +32,12 @@ object LedgerOffsetValidator {
       ledgerOffset: LedgerOffset,
       fieldName: String,
   ): Either[StatusRuntimeException, domain.LedgerOffset] = {
-    ledgerOffset match {
-      case LedgerOffset(LedgerOffset.Value.Absolute(value)) =>
+    ledgerOffset.value match {
```
cocreature, line 35: Is this just cleanup or does the old code no longer work?

cocreature (approved):
Nice, thank you!

fabiotudone-da (commented):
LGTM but it's a very technical change in areas I am not knowledgeable enough in (Bazel, `scala-protoc-plugins`, ScalaPB itself) so I'll leave the approval to someone more versed.

bazel_tools/scalapb/com/daml/package.proto
SamirTalwar-DA: Do _not_ get me started. :upside_down_face:

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/LedgerOffsetValidator.scala
SamirTalwar-DA: Cleanup; I reverted the other changes in this file (adding `_` to discard unknown fields everywhere) but thought I'd leave this one alone, because I find it easier to read.

cocreature: fine with me, just wanted to check

bazel_tools/scala_version.bzl
SamirTalwar-DA: It's used in the `WORKSPACE`. Can we move it and still use it there?

cocreature: I don’t see why we shouldn’t be able to do that.

SamirTalwar-DA: Got some cycles, but moving them into their own file (_scalapb.bzl_) works well.

bazel_tools/scalapb/com/daml/package.proto
S11001001: next-level dogfooding

S11001001 (approved):

hubert-da (commented):
Same as Fabio, LGTM

bazel_tools/scalapb/com/daml/package.proto
```diff
@@ -0,0 +1,13 @@
...
+  scope: PACKAGE
+  preserve_unknown_fields: false
```
miklos-da, line 12: :+1:

PR #9714, opened 2021-05-17
Upgrade gatling to version 3.5.1

cocreature:
Ths is required to get a version of Gatling that supports Scala
2.13 (and only that because they do not cross build).

Unfortunately the upgrade is a bit more annoying than I was hoping for:

Our custom gatling utils rely on parsing the simulation log. This is
an internal file format with zero documentation or stabilityt
guarantees and as expected it has changed in incompatible ways during
the upgrade.

Rather than trying to reverse engineer and adapt to changes everytime
we upgrade, this PR switches us to a slightly more supported codepath
by parsing the `stats.json` and `assertions.json` produced by the
highcharts stuff. Afaict this is also what for example the Jenkins
integration relies on so while it’s not completely public API it seems
like the best option I could find.

There are a few pieces of information we can’t get out of those
files. Specifically:

1. maxUsers: we only ever need one users anyway so not really relevant.
2. start, duration, end: no idea why we would want those. we want per
   request metrics not the total duration.
3. geometric mean: slightly annoying, but avg & stdev should be good
enough™.
4. The scenario name: Not really an issue but if it is, we can
   disambiguate by changing request names.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature:
I manually ran the JSON API perf tests to make sure they work with this.

S11001001 (approved):

libs-scala/gatling-utils/src/main/scala/com/daml/gatling/stats/SimulationLog.scala
```diff
@@ -1,30 +1,23 @@
 // Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-// Copyright (c) 2019, Digital Asset (Switzerland) GmbH and/or its affiliates.
-// All rights reserved.
```
stefanobaghino-da, line 5: Ah, yes, a ledger of licenses.

stefanobaghino-da (approved):
🙇🏻

PR #9725, opened 2021-05-18
Add cli option & system property to enable json only logging for http json service

realvictorprm:
changelog_begin
- [Ledger HTTP Json Service] Logging output can now be in JSON either via providing the cli option `--log-encoder json` or via setting the env var `LOG_FORMAT_JSON=true`

changelog_end

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -25,14 +25,27 @@ import scala.util.{Failure, Success, Try}
 
 object Main {
 
-  private[this] val logger = ContextualizedLogger.get(getClass)
+  private[this] lazy val logger = ContextualizedLogger.get(getClass)
```
stefanobaghino-da, line 28: Why making this lazy? If it's not used it can simply be removed, if it's used what's to gain from deferring its creation?

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LeveledLogger.scala
```diff
@@ -82,10 +92,11 @@ private[logging] sealed abstract class LeveledLogger {
   protected def log(msg: String, t: Throwable): Unit
   protected def log(m: Marker, msg: String, t: Throwable): Unit
   protected def log(fmt: String, arg: AnyRef): Unit
+  protected def log(m: Marker, msg: String): Unit
 
   final def apply(msg: => String)(implicit loggingContext: LoggingContext): Unit =
     if (isEnabled)
-      loggingContext.ifEmpty(log(msg))(log(s"$msg (context: {})", _))
+      loggingContext.ifEmpty(log(msg))(log(_, msg))
```
stefanobaghino-da, line 99: This would mean that the logging context is completely removed unless you are using JSON logging. There are usages where logging a meaningful message relies on the context being printed along with the message for it to be meaningful (in those cases the message only prints what even happened, e.g. something along the lines of "Request received", but the data about the request is in the context -- this is a deliberate choice to avoid the same information to appear in both the message and the context). I think it's important to find a way to print the context along with the message is a non-structured encoder is chosen.

stefanobaghino-da (changes requested):
Basically good, but I think it's quite important to allow users to have the context printed out if they choose a non-structured logger.

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -7,6 +7,28 @@ import ch.qos.logback.classic.{Level => LogLevel}
...
+    try {
+      val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+      val configurator = new JoranConfigurator
+      configurator.setContext(context)
+      context.reset()
+      configurator.doConfigure(stream)
+    } catch {
+      case je: JoranException =>
+        // Fallback to System.out because the logger won't work in any way anymore.
+        System.out.println(s"reconfigured failed using url $pathToLogbackFileInJarFile: $je")
```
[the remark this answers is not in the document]
realvictorprm: Very good point, I forgot that it exists :joy: Has been a while since I last used it

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: Good catch, it's not necessary anymore. In a prior version I had to defer it's creation so the configuration for logback is proceed late enough such that the changed system properties are used.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LeveledLogger.scala
realvictorprm: with the adjusted logback.xml file this should be now fine.

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -7,6 +7,29 @@ import ch.qos.logback.classic.{Level => LogLevel}
...
+    val stream = clazz.getClassLoader.getResource(pathToLogbackFileInJarFile).openStream()
+    try {
+      val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+      val configurator = new JoranConfigurator
+      configurator.setContext(context)
+      context.reset()
+      configurator.doConfigure(stream)
+    } catch {
+      case je: JoranException =>
+        // Fallback to System.out because the logger won't work in any way anymore.
```
cocreature, line 25:
```suggestion
        // Fallback to System.err.println because the logger won't work in any way anymore.
```

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -7,6 +7,29 @@ import ch.qos.logback.classic.{Level => LogLevel}
 
 object Logging {
 
+  def reconfigure(clazz: Class[_], pathToLogbackFileInJarFile: String): Unit = {
+    // Try reconfiguring the library
+    import ch.qos.logback.core.joran.spi.JoranException
+    import ch.qos.logback.classic.LoggerContext
+    import ch.qos.logback.classic.joran.JoranConfigurator
+    import org.slf4j.LoggerFactory
+    val stream = clazz.getClassLoader.getResource(pathToLogbackFileInJarFile).openStream()
```
cocreature, line 16: You should be able to throw `Using`/try with resources at this.

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -34,4 +57,24 @@ object Logging {
       )
     ()
   }
+
+  def useJsonLogEncoderParse[C](
+      parser: scopt.OptionParser[C]
+  )(useJsonLogEncoder: Setter[C, Boolean]): Unit = {
```
cocreature, line 63: Maybe introduce a separate type for the supported logging formats instead of using a boolean? That makes it easier to not swap the meaning of the boolean accidentally.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -33,6 +33,19 @@ object Main {
...
+  def adjustAndReloadLoggingOptions(
+      config: Config
+  ): Unit = {
+    // If the system property was explicitly set before application startup and the config option was provided,
+    // the prior value will be overridden here.
+    if (config.useJsonLogEncoder) {
+      Logging.setUseJsonLogEncoderSystemProp()
+      Logging.reconfigure(getClass, "logback.xml")
+    }
+    // Not related to the content of the logback.xml file.
```
cocreature, line 45: not sure what this comment means

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
realvictorprm: done

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
realvictorprm: can you elaborate?

cocreature:
```
diff --git a/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala b/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
index 844fe711dc..d42b85a070 100644
--- a/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
+++ b/ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
@@ -4,6 +4,7 @@
 package com.daml.cliopts

 import ch.qos.logback.classic.{Level => LogLevel}
+import scala.util.Using

 object Logging {

@@ -13,20 +14,20 @@ object Logging {
     import ch.qos.logback.classic.LoggerContext
     import ch.qos.logback.classic.joran.JoranConfigurator
     import org.slf4j.LoggerFactory
-    val stream = clazz.getClassLoader.getResource(pathToLogbackFileInJarFile).openStream()
-    try {
-      val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
-      val configurator = new JoranConfigurator
-      configurator.setContext(context)
-      context.reset()
-      configurator.doConfigure(stream)
-    } catch {
-      case je: JoranException =>
-        // Fallback to System.err.println because the logger won't work in any way anymore.
-        System.err.println(s"reconfigured failed using url $pathToLogbackFileInJarFile: $je")
-        je.printStackTrace(System.err)
-    } finally {
-      stream.close()
+    Using.resource(clazz.getClassLoader.getResource(pathToLogbackFileInJarFile).openStream()) {
+      stream =>
+        try {
+          val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+          val configurator = new JoranConfigurator
+          configurator.setContext(context)
+          context.reset()
+          configurator.doConfigure(stream)
+        } catch {
+          case je: JoranException =>
+            // Fallback to System.err.println because the logger won't work in any way anymore.
+            System.err.println(s"reconfigured failed using url $pathToLogbackFileInJarFile: $je")
+            je.printStackTrace(System.err)
+        }
     }
   }
```
Definitely subjective but I like it because it makes it very explicit that you’re not leaking resources.

realvictorprm: I was looking for something which ensures that I'm not leaking resources, thanks a lot!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: I edited the comment, I hope it is now more clear :thinking:

ledger-service/cli-opts/BUILD.bazel
```diff
@@ -16,7 +16,15 @@ da_scala_library(
...
     scalacopts = lf_scalacopts,
     tags = ["maven_coordinates=com.daml:http-json-cli-opts:__VERSION__"],
+    versioned_scala_deps = {
+        "2.12": [
+            "@maven//:org_scala_lang_modules_scala_collection_compat",
+        ],
+    },
     visibility = ["//visibility:public"],
+    runtime_deps = [
+        "@maven//:org_codehaus_janino_janino",
```
cocreature, line 26: have you checked that this work via `daml-sdk-head`? We bundle everything into a single JAR there to save size and we might also need to add it to the runtime deps of that jar.

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -19,9 +54,26 @@ object Logging {
       .getOrElse(throw new java.lang.IllegalArgumentException(s"Unknown logging level $level"))
   }
 
+  @SuppressWarnings(Array("org.wartremover.warts.All"))
+  private implicit val scoptLogEncoder: scopt.Read[LogEncoder] =
+    scopt.Read.reads { encoder =>
+      Either
+        .cond(
+          KnownLogEncoders.contains(encoder.toLowerCase()),
```
cocreature, line 62: maybe just make the pattern match exhaustive and handle the case where it’s not one of the two strings there instead of having a separate check upfront?

ledger-service/cli-opts/src/main/scala/cliopts/Logging.scala
```diff
@@ -19,9 +54,26 @@ object Logging {
...
+          encoder.toLowerCase match {
+            case "plain" => LogEncoder.Plain
+            case "json" => LogEncoder.Json
+          },
+          s"Unrecognized logging encoder $encoder",
+        )
+        .getOrElse(
```
cocreature, line 69: if you drop the message in the `Left` you might as well use an `Option above or just throw directly in the pattern match.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
```diff
@@ -4,17 +4,19 @@
...
 import scala.util.Try
 
 class OptionParser(getEnvVar: String => Option[String], supportedJdbcDriverNames: Set[String])
-    extends scopt.OptionParser[Config]("http-json-binary")
-    with StrictLogging {
+    extends scopt.OptionParser[Config]("http-json-binary") {
+
+  // We need a deferring logger because we do not know
+  // whether the user expects json or plain logging output.
+  val logger: DeferringLogger = DeferringLogger(getClass)
```
cocreature, line 19: this seems like a lot of complexity and slightly weird UX (by delaying logs) just to get the right format for logs from the config parser. I’m not sure we actually log anything useful here apart from maybe some deprecation warnings here and there which we could just move out of this.

I would suggest to drop this and just accept that for this PR logs will have the wrong format and move logs out of option parsing in a separate PR to fix that properly.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
```diff
@@ -47,6 +49,11 @@ class OptionParser(getEnvVar: String => Option[String], supportedJdbcDriverNames
 
   override def renderingMode: RenderingMode = RenderingMode.OneColumn
 
+  override def displayToOut(msg: String): Unit = logger.info(msg)
+  override def displayToErr(msg: String): Unit = logger.error(msg)
+  override def reportError(msg: String): Unit = logger.error(msg)
+  override def reportWarning(msg: String): Unit = logger.warn(msg)
```
cocreature, line 55: Not sure CLI output needs to go via the logger. Structured logging is really mainly useful for long-running stuff not when you messed up CLI flags and the thing didn’t even start.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
realvictorprm: It definitely is a bit weird to defer the logs. I aimed to be consistent here, however if you think this is unnecessary then I can drop the DeferringLogger.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
realvictorprm: Will remove this too then

realvictorprm: And this is what I talked about yesterday with you in Slack :D

ledger-service/http-json/src/main/resources/logback.xml
```diff
@@ -1,10 +1,18 @@
...
-        </encoder>
+        <if condition='isDefined("LOG_FORMAT_JSON")'>
+            <then>
+                <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+            </then>
+            <else>
+                <!-- encoders are assigned the type
+                    ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+                <encoder>
+                    <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg %replace(, context: %marker){', context: $', ''} %n</pattern>
```
stefanobaghino-da, line 12: This solution is fine but please note that you need to apply it to all `logback.xml` files that are used by default in all the components that use the contextualized logger, otherwise the logging context will be lost there.

stefanobaghino-da (changes requested):

ledger-service/http-json/src/main/resources/logback.xml
realvictorprm: Great, will apply it to the other files then :)

ledger-service/cli-opts/BUILD.bazel
realvictorprm: Got no problems

ledger-service/http-json/src/main/resources/logback.xml
realvictorprm: OK did that, hope I caught all projects which are affected.

cocreature (approved):
lgtm, thank you!

realvictorprm:
OK the CI hates me I think but let's see after another run :thinking:

PR #9727, opened 2021-05-18
Dpp 384 todo cleanup postgres storage backend

nmarton-da:
### Pull Request Checklist

trigger the build.


nmarton-da:
Note for reviewers: best approach might be go commit by commit, separate commits are compiling and holding self-contained part of the work.
In the separate commit descriptions I tried to leave sufficient justifications if TODOs were just removed.

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V51__drop_configuration_from_parameters_table.sha256
```diff
@@ -0,0 +1 @@
+6326123dd12a9d3652369df4ff68e7479716307b95960211aefa08fc8a6ce057
```
rautenrieth-da, line 1: Note to self: this file will have to be renamed to V101 or merged with the V100 migration after https://github.com/digital-asset/daml/pull/9481 is merged.

rautenrieth-da (approved):
Nice work removing the last configuration from the parameters table. Did you investigate whether it has any performance impact? Either a positive one because the heavily used parameters table is now smaller, or a negative one because the latest config lookup is now more expensive.

nmarton-da:
> Nice work removing the last configuration from the parameters table. Did you investigate whether it has any performance impact? Either a positive one because the heavily used parameters table is now smaller, or a negative one because the latest config lookup is now more expensive.

Good point @rautenrieth-da , I did not. I do not expect any negative effect on the ingestion side (less data to insert). So what regression can happen is on the reads. I guess an LR run with some comparison with a baseline is in order, but there I might be a bit out of my depth. Let's discuss this on the next daily.

PR #9733, opened 2021-05-18
Parallelize the data-dependencies test.

sofiafaro-da:
It runs about twice as fast on my laptop (240s -> 120s) when allowed to
run tests in parallel. I'm not sure this will work or make a huge
difference when it comes to timeouts in CI, but it's worth a shot.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/src/DA/Test/DataDependencies.hs
+++ b/compiler/damlc/tests/src/DA/Test/DataDependencies.hs
@@ -15,7 +15,6 @@ import Data.List (sort)
 import qualified Data.NameMap as NM
 import Module (unitIdString)
 import System.Directory.Extra
-import System.Environment.Blank
 import System.FilePath
 import System.Info.Extra
 import System.IO.Extra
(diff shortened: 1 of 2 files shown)


compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -27,7 +26,6 @@ import SdkVersion
 
 main :: IO ()
 main = do
-    setEnv "TASTY_NUM_THREADS" "1" True
```
cocreature, line 30: given that you report a 2x speedup how much do we gain by setting this to 2 and 4? We should also change the resources Bazel allocates for this either if we don’t fix the number of CPUs here by marking it exclusive or via `    tags = ["cpu:2"],`.

sofiafaro-da: I just tried this now on my laptop:

- Setting TASTY_NUM_THREADS to 1: 229.0s
- Setting TASTY_NUM_THREADS to 2: 142.3s
- Setting TASTY_NUM_THREADS to 3: 103.9s
- Setting TASTY_NUM_THREADS to 4: 96.5s
- Not setting TASTY_NUM_THREADS: 134.5s

I'm going to set it to 3. Should I change anything else (e.g. in Bazel)?

cocreature: sounds good, then let’s add `cpu:3` to the tags to make sure we don’t schedule too much.

cocreature (approved):
Thanks!

PR #9742, opened 2021-05-19
ghc-lib: update

robin-da:
### Pull Request Checklist

trigger the build.


compiler/damlc/tests/daml-test-files/RecordDotUpdates.daml
```diff
@@ -50,3 +50,15 @@ y = Outer {inner = x, text = "world", int = 21}
 y' = let text = "world" in let int = 12 in y {text, inner.text = text, int, inner.int = int}
 
 z = let Outer {..} = y in Outer {..}
+
+data One = One with two : Two
```
cocreature, line 54: maybe also a good idea to add a test that uses whatever the magic name is we use on the GHC side given how much trouble we had to go through to fix that :)

compiler/damlc/tests/daml-test-files/RecordUpdate.daml
```diff
@@ -3,8 +3,8 @@
 
 -- @INFO range=43:18-43:42; Evaluate
 -- @INFO range=43:25-43:41; Use const
--- @QUERY-LF .modules[] | .values[] | select(.name_with_type | lf::get_value_name($pkg) == ["p_1_0"]) | .expr.rec_upd | (lf::get_field($pkg) == "x") and (.record.val | lf::get_value_name($pkg) == ["origin"]) and (.update.prim_lit.int64 == "1")
--- @QUERY-LF .modules[] | .values[] | select(.name_with_type | lf::get_value_name($pkg) == ["p_1_2"]) | .expr.rec_upd | (lf::get_field($pkg) == "y") and (.record.rec_upd | (lf::get_field($pkg) == "x") and (.record.val | lf::get_value_name($pkg) == ["origin"]) and (.update.prim_lit.int64 == "1")) and (.update.prim_lit.int64 == "2")
+-- @QUERY-LF .modules[] | .values[] | select(.name_with_type | lf::get_value_name($pkg) == ["p_1_0"]) | .expr.let | (.body.rec_upd | (lf::get_field($pkg) == "x") and (.update.prim_lit.int64 == "1")) and (.bindings[].bound.val | (lf::get_value_name($pkg) == ["origin"]))
```
cocreature, line 6: I’m confused. All of this operates on let bindings but your PR switches to case bindings?

compiler/damlc/tests/daml-test-files/RecordUpdate.daml
```diff
@@ -61,10 +61,12 @@ data Point = Point with
 origin = Point with x = 0; y = 0
 
 -- NOTE(MH): We expect this to become
--- `ERecUpd _ "x" (EVal ["origin"]) (EBuiltin (BEInt64 1))`
+-- NOTE(drsk): The ELet binding was added with RecordDotUpdates for better type inference.
+-- `ELet (Binding ("arg", _) (EVal ["origin"])) (ERecUpd _ "x" (EVal ["arg"]) (EBuiltin (BEInt64 1)))`
```
cocreature, line 65: I’m slightly worried about this. The `let` for non-nested updates does(’t help with inference. It just slows things down. Can we easily detect a singleton field list in GHC and not generate the case there?

garyverhaegen-da:
Please update reference to a merged-into-da-master-8.8.1 commit before merging this (i.e. merge https://github.com/digital-asset/ghc/pull/65 first & refer to resulting commit).

compiler/damlc/tests/daml-test-files/RecordUpdate.daml
robin-da: Yes! Ghc turns the single case binding back into a let binding ...

robin-da:
@garyverhaegen-da yes, we'll merge the ghc PR when we're happy with the results here, also the windows deps need to be pinned before merging.

compiler/damlc/tests/daml-test-files/RecordDotUpdates.daml
robin-da: That's a good idea. The magic name is 'arg' now.

compiler/damlc/tests/daml-test-files/RecordUpdate.daml
robin-da: I can think of two things: Either traverse the updates and check whether we have nested ones or move the Let/Case bindings to he update expression instead outside of the record update and only add it for nested updates. But that would potentially mean many let/case bindings.

robin-da: This has been changed now in the ghc fork to only add the case/let expression for nested updates.

compiler/damlc/tests/daml-test-files/RecordDotUpdates.daml
```diff
@@ -50,3 +50,22 @@ y = Outer {inner = x, text = "world", int = 21}
...
+q : One
+q = def {two.val = 42}
+
+-- NOTE (drsk) This is here to check that we don't colide with the local binding of the `arg`
+-- variable introduced by ghc when adding a preceding let/case expression in the record dot update.
+arg : One
+arg = One with two = Two 32
+
+arg' : One
+arg' = arg {two.val = 42}
```
cocreature, line 71: with the new code, I think a test like this would also be good
```
module Test where

data X = X with
  a : Int
  c : Y

data Y = Y with
  r : Int

f :  Int -> X -> X
f arg x = x with
  a = 1
  c.r = arg
```
luckily this does indeed work correctly, so just needs a test.

cocreature (approved):
Nice work, thanks!

compiler/damlc/tests/daml-test-files/RecordDotUpdates.daml
robin-da: @cocreature do you want to check the generated daml-lf code for this?

cocreature: shouldn’t be necessary. If we mess it up it’s a type error. The `arg` we bind in the `case` is of type `X`. The `arg` in the function is of type `int`.

robin-da: agreed, updated the test.

PR #9743, opened 2021-05-19
more frequent Windows cache cleanup

garyverhaegen-da:
We'd like this to run more than once an hour, and Azure doesn't support that as a cron.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -2,6 +2,38 @@
 # SPDX-License-Identifier: Apache-2.0
 
 jobs:
+
+- job: fix_bazel_cache
+  timeoutInMinutes: 120
+  pool:
+    name: 'ubuntu_20_04'
+    demands: assignment -equals default
+  steps:
+    - checkout: self
+    - bash: ci/dev-env-install.sh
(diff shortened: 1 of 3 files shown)


ci/build.yml
```diff
@@ -2,6 +2,38 @@
...
+        set -euo pipefail
+        eval "$(dev-env/bin/dade assist)"
+
+        bazel build //ci/cron:cron
+
+        key=$(mktemp)
+        cleanup="rm -rf $key ~/.config/gcloud"
+        trap "$cleanup" EXIT
+        echo "$GCRED" > $key
+        gcloud auth activate-service-account --key-file=$key
```
cocreature, line 28: do we have delete permissions on PRs?

cocreature (approved):
Nice, thank you!

ci/build.yml
garyverhaegen-da: Good question, let me check... Yes, we do. Bazel needs it in general to manage the cache.

PR #9748, opened 2021-05-19
Fix flakiness in hot reload tests

cocreature:
I started looking into failures in the integration tests and this
seems by far the most prevalent one (maybe even the only one but that
needs more investigation). The issue here is that sometimes waiting
40s isn’t enough. On the other hand, sometimes we wait unnecessarily
long which is also bad.

The fix in this PR is to wait until we see the "reload complete"
message which should be exactly the right amount of time to wait.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Thank you!

PR #9750, opened 2021-05-19
Make hikari connection pool timeout configurable

dasormeter:
Make hikari connection pool timeout configurable (default remains at 250.millis as it was previously)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -1143,7 +1143,7 @@ private[platform] object JdbcLedgerDao {
         serverRole,
         jdbcUrl,
         connectionPoolSize,
-        250.millis,
```
nmarton-da, line 1146: 250 millis is the minimum setting, hikari does not allow lower, if we need to increase this in some deployment scenarios that means that the query load + max connection pool size + this timeout combination is insufficient, (and I guess: increasing the connection pool size is not desirable).
So how about making this also a configuration parameter instead, so we can play around with this parameters per deployment?

dasormeter: alright in the next commit extracted this value out to something configurable -- it does lead to a lot of repetition of the configuration in multiple places (we may be able to do further defaulting). https://github.com/digital-asset/daml/pull/9750/commits/c193e875b62cfaffca5242a20aec28242b07e4e8

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/ParticipantConfig.scala
```diff
@@ -32,6 +35,7 @@ object ParticipantConfig {
     s"jdbc:h2:mem:$participantId;db_close_delay=-1;db_close_on_exit=false"
 
   val DefaultManagementServiceTimeout: Duration = Duration.ofMinutes(2)
+  val DefaultApiServerDatabaseConnectionTimeout: Duration = Duration.ofMinutes(2)
```
nmarton-da, line 38: 250 millis perhaps?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: wow, that exploded! sorry for that and thanks for doing it :raised_hands:

nmarton-da (approved):
Looks excellent to me :bow: 
Made one comment on a two minutes dbtimeout (probably a cpypastetypo)

PR #9732, opened 2021-05-18
Add golden test for example Daml ledger export

aherrmann-da:
Addresses part of https://github.com/digital-asset/daml/pull/9446#discussion_r615906418

- Adds an example Daml ledger export to the docs tree to be used as an example in the Daml ledger export documentation.
    Incorporating it in the actual documentation is left for a follow-up PR.
- Adds a Bazel build target to generate a Daml ledger export using `client_server_build`.
- Adds a golden test that compares the example export in the docs to the one generated by the build target.
- Some minor changes in support of the above:
    - Expose an overload of the Daml script runner's main function that takes a config object
    - Extend `client_server_build` to support multiple output files.
    - Extend `sh_inline_test` to support toolchains for make variable substitution.

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #9756, opened 2021-05-20
Example export

aherrmann-da:
Pretty print the generated `daml.yaml` file as suggested in https://github.com/digital-asset/daml/pull/9732#discussion_r634998661.

Adds the generated `daml.yaml` to the golden test. In this case we need to normalize the data-dependencies by replacing the path to the output directory by a placeholder and by replacing the package-id hashes by a placeholder before comparison to the golden file, so that it doesn't have to be updated whenever a stdlib package-id changes.

Sorts the list of data-dependencies in the generated `daml.yaml` file to improve reproducibility. 

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice, thank you!

aherrmann-da:
Windows was failing with
```
java.nio.file.FileAlreadyExistsException: bazel-out\x64_windows-opt\bin\daml-script\export\example-export\deps
```
because the ledger export was generated directly into the Bazel execroot. This is problematic on Windows due to the lack of sandboxing where undeclared outputs may persist across builds and then cause conflicts like the above with later builds.
To avoid this I've changed the code to generated the ledger export into a temp-dir and then only move the exected outputs to the execroot.

PR #9758, opened 2021-05-20
prepare for EV Windows signing

garyverhaegen-da:
Setting up a non-disruptive way to test out EV signing of our Windows artifacts.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
looks very sensible for testing, thanks for looking into this!

PR #9764, opened 2021-05-20
participant-state: rejection reason abstractions

daravep:
Before this change, any write-service was restricted to the predefined
set of rejection reasons which couldn't capture the complex canton
rejection reasons.
In order to open up the path for a redesign, I've migrated the current
rejection reasons into a V0 version and opened the abstract class for
customized implementations.
In addition, I added another SubmissionResult which will support synchronous error responses.

CHANGELOG_BEGIN

- Participant-state: Introduced versioned form of rejection reason

CHANGELOG_END

There is a related PR https://github.com/DACH-NY/canton/pull/6177 that is going to leverage these changes, such that we can work out the V1 error responses.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
@@ -42,6 +42,7 @@ trait ErrorFactories {
   def unimplemented(description: String): StatusRuntimeException =
     grpcError(Status.UNIMPLEMENTED.withDescription(description))
 
+  // permission denied is intentionally without description to ensure we don't leak security relevant information by accident
   def permissionDenied(): StatusRuntimeException =
     grpcError(Status.PERMISSION_DENIED)
 
--- a/ledger/participant-state/kvutils/tools/BUILD.bazel
+++ b/ledger/participant-state/kvutils/tools/BUILD.bazel
@@ -63,7 +63,6 @@ da_scala_test(
         "@maven//:org_scalaz_scalaz_core",
     ],
     deps = [
-        "@maven//:org_mockito_mockito_core",
         "//daml-lf/archive:daml_lf_dev_archive_proto_java",
         "//daml-lf/data",
         "//daml-lf/transaction",
@@ -75,6 +74,7 @@ da_scala_test(
         "//ledger/participant-state/kvutils",
         "//ledger/participant-state/kvutils/tools",
         "@maven//:io_grpc_grpc_api",
+        "@maven//:org_mockito_mockito_core",
     ] + (["@maven//:com_google_protobuf_protobuf_java"] if scala_major_version == "2.13" else []),
 )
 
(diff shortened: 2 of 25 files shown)


daravep:
@gerolf-da and @mziolekda know about it and @meiersi-da is aware of the direction. Generally the idea is that I'm trying to systematically document all possible Canton rejections and then find the correct split of rejections shared among implementations and the ones that are specific to a ledger. Additionally, we want to include some information on the error context and error-retryability.

gerolf-da (approved):
Thanks, @daravep.
Nitpick: I'm not a huge fan of the `V0` suffix, but it's not the end of the world.

meiersi-da:
Thanks @daravep. I agree with the direction, and trust that you are sure about this change being necessary.

This PR touches the `//ledger/participant-state-api/` directory that is currently under heavy refactoring by @nmarton-da and @rautenrieth-da in @mziolekda 's team. I've seen that you rename all files, which is good.

 To avoid any complications, I'd suggest to wait for an approval by @mziolekda so that this change does not stomp on his teams work.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -42,6 +42,7 @@ trait ErrorFactories {
   def unimplemented(description: String): StatusRuntimeException =
     grpcError(Status.UNIMPLEMENTED.withDescription(description))
 
+  // permission denied is intentionally without description to ensure we don't leak security relevant information by accident
```
rautenrieth-da, line 45: This is exactly the reason why the description was left out initially.

Some context: what the ledger API authorization does is only compare the content of the request (e.g., the command object) with the content of the access token, both of which are provided by the user. Ledger API authorization does not depend on any on-ledger data.

The main concern should be about leaking cryptographic secrets related to the access token (such as the correct signature or the keys used for checking the signature). If the validation/decoding of the access token fails, we only store a generic [Claims.Unauthenticated](https://github.com/digital-asset/daml/blob/main/ledger/ledger-api-auth/src/main/scala/com/digitalasset/ledger/api/auth/Claims.scala#L52) object without any details. The code responsible for sending out the gRPC error message does not have access to the original access token.

Adding details for why permission was denied should therefore be possible.

daravep: Yeah, so @soren-da suggested to rather just leave it as is. We don't have any test ensuring that we do not leak security-relevant information through that rejection. So if we want to respond with meaningful errors, then we also need to add the whole test harness to ensure this remains secure even after refactorings. As this is out-of-scope for now, I guess that it will have to be picked up at a later point in time.

ledger/participant-state/kvutils/tools/BUILD.bazel
```diff
@@ -74,6 +74,7 @@ da_scala_test(
         "//ledger/participant-state",
         "//ledger/participant-state/kvutils",
         "//ledger/participant-state/kvutils/tools",
+        "@maven//:io_grpc_grpc_api",
```
mziolekda, line 77: Can we please sort the deps?

mziolekda (approved):
Looks good

ledger/participant-state/kvutils/tools/BUILD.bazel
daravep: done, hopefully in the way you are preferring it.

PR #9765, opened 2021-05-20
Use munit for diffs in ledger-api-test-tool

cocreature:
As everyone knows that has worked on the ledger API test tool, diffx
makes compilation very very slow (>150s for TransactionServiceIT).

As some people (myself included) know, trying to get diffx to not be
slow is also a giant nightmare (have fun debugging mutually recursive
implicits derived by Scala macros)

So this PR takes a different approach:

Drop diffx completely and replace it by munit. This leads to slightly
different diffs but I think they’re close enough and importantly munit
involves absolutely no macros for those diffs so compilation takes 5s
for TransactionServiceIT.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature:
Comparison of diffs:
diffx



munit
Screenshot from 2021-05-20 19-10-57

The main difference I see is that we get false positives on ArraySeq vs vector. Doesn’t seem that bad imho.

kamil-da (approved):
The speed-up is awesome!

tudor-da (approved):
Nice one!

stefanobaghino-da (approved):
Now all we have to do is replace Scalatest with MUnit everywhere. :trollface:

PR #9772, opened 2021-05-21
fix signing machine

garyverhaegen-da:
Turns out PowerShell is not Bash. Who knew? :shrug:

CHANGELOG_BEGIN
CHANGELOG_END


garyverhaegen-da:
Note: I've already deployed this on the assumption that I could piggyback on the authorization given for #9758.

cocreature (approved):
Not that I understand powershell but looks plausible

PR #9773, opened 2021-05-21
test_windows_signing: add quotes

garyverhaegen-da:
Otherwise Bash interprets the baskslashes and it's not a path anymore.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #9775, opened 2021-05-21
[In-memory fan-out] EventsBuffer implementation

tudor-da:
Implements an ordered-by-offset ring buffer wrapper necessary for the in-memory fan-out mechanism for Ledger API serving. The implementation is generic in order to allow particularization for both TransactionsBuffer and CommandCompletionsBuffer.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
+++ b/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
@@ -636,6 +636,14 @@ final class Metrics(val registry: MetricRegistry) {
         val deduplicateCommand: Timer = registry.timer(Prefix :+ "deduplicate_command")
         val stopDeduplicateCommand: Timer = registry.timer(Prefix :+ "stop_deduplicating_command")
         val prune: Timer = registry.timer(Prefix :+ "prune")
+
+        object streamsBuffer {
+          private val Prefix: MetricName = index.Prefix :+ "streams_buffer"
+
+          def push(qualifier: String): Timer = registry.timer(Prefix :+ qualifier :+ "push")
+          def slice(qualifier: String): Timer = registry.timer(Prefix :+ qualifier :+ "slice")
+          def prune(qualifier: String): Timer = registry.timer(Prefix :+ qualifier :+ "prune")
+        }
       }
 
       object read {
--- a/ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
@@ -1,0 +1,174 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.cache
+
+import com.daml.metrics.{Metrics, Timed}
+import com.daml.platform.store.cache.EventsBuffer.{
+  BufferStateRef,
+  RequestOffBufferBounds,
+  SearchableByVector,
+  UnorderedException,
+}
+
+import scala.annotation.tailrec
+import scala.collection.Searching.{Found, InsertionPoint, SearchResult}
+import scala.math.Ordering
+import scala.math.Ordering.Implicits.infixOrderingOps
+
+/** An ordered-by-offset ring buffer.
+  *
+  * The buffer allows appending only elements with strictly increasing offsets.
+  *
+  * @param maxBufferSize The maximum buffer size.
+  * @param metrics The DAML metrics.
+  * @param bufferQualifier The qualifier used for metrics tag specialization.
+  * @param isRangeEndMarker Identifies if an element [[E]] should be treated
+  *                         as a range end marker, in which case the element would be treated
+  *                         as a buffer range end updater and not appended to the actual buffer.
+  * @tparam O The offset type.
+  * @tparam E The entry buffer type.
+  */
+private[cache] final class EventsBuffer[O: Ordering, E](
+    maxBufferSize: Int,
+    metrics: Metrics,
+    bufferQualifier: String,
+    isRangeEndMarker: E => Boolean,
+) {
+  @volatile private var _bufferStateRef = BufferStateRef[O, E]()
+
+  private val pushTimer = metrics.daml.services.index.streamsBuffer.push(bufferQualifier)
+  private val sliceTimer = metrics.daml.services.index.streamsBuffer.slice(bufferQualifier)
+  private val pruneTimer = metrics.daml.services.index.streamsBuffer.prune(bufferQualifier)
+
+  /** Appends a new event to the buffer.
+    *
+    * Starts evicting from the tail when `maxBufferSize` is reached.
+    *
+    * @param offset The event offset.
+    *              Must be higher than the last appended entry's offset, with the exception
+    *              of the range end marker, which can have an offset equal to the last appended element.
+    * @param entry The buffer entry.
+    */
+  def push(offset: O, entry: E): Unit =
+    Timed.value(
+      pushTimer,
+      synchronized {
+        _bufferStateRef.rangeEnd.foreach { lastOffset =>
+          // Ensure vector grows with strictly monotonic offsets.
+          // Only specially-designated range end markers are allowed
+          // to have offsets equal to the buffer range end.
+          if (lastOffset > offset || (lastOffset == offset && !isRangeEndMarker(entry))) {
+            throw UnorderedException(lastOffset, offset)
+          }
+        }
+
+        var auxBufferVector = _bufferStateRef.vector
+
+        // The range end markers are not appended to the buffer
+        if (!isRangeEndMarker(entry)) {
+          if (auxBufferVector.size == maxBufferSize) {
+            auxBufferVector = auxBufferVector.drop(1)
+          }
+          auxBufferVector = auxBufferVector :+ offset -> entry
+        }
+
+        // Update the buffer reference
+        _bufferStateRef = BufferStateRef(auxBufferVector, Some(offset))
+      },
+    )
+
+  /** Returns a slice of events from the buffer.
+    *
+    * Throws an exception if requested with `endInclusive` higher than the highest offset in the buffer.
+    *
+    * @param startExclusive The start exclusive bound of the requested range.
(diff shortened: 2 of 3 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
```diff
@@ -0,0 +1,167 @@
...
+    *              Must be higher than the last appended entry's offset, with the exception
+    *              of the range end marker, which can have an offset equal to the last appended element.
+    * @param entry The buffer entry.
+    */
+  def push(offset: O, entry: E): Unit =
+    Timed.value(
+      pushTimer,
+      synchronized {
+        if (bufferRef.size == maxBufferSize)
+          bufferRef = bufferRef.drop(1)
```
kamil-da, line 60: This is probably a minor thing but in case when we have a full buffer and a range-end marker is pushed, a single buffer element will be dropped but no new element will be appended. Is that intentional?

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
```diff
@@ -0,0 +1,167 @@
...
+    * @return The series of events as an ordered vector satisfying the input bounds.
+    */
+  def slice(startExclusive: O, endInclusive: O): Vector[(O, E)] =
+    Timed.value(
+      sliceTimer, {
+        val buffer = bufferRef
+        if (buffer.isEmpty) {
+          Vector.empty
+        } else if (bufferRangeEnd.exists(_ < endInclusive)) {
+          throw RequestOffBufferBounds(buffer.last._1, endInclusive)
```
kamil-da, line 96: Shouldn't the condition on `endInclusive` not being greater than `bufferRangeEnd` be before returning an empty buffer to have consistent behavior?
E.g. after pruning the whole buffer and asking for an out-of-range end offset we won't get an exception but an empty buffer.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+        buffer.slice(0, LastOffset) shouldBe Vector(8 -> 16, 13 -> 26)
+      }
+    }
+
+    "element inserted" should {
+      "prune inclusive" in withBuffer() { buffer =>
+        buffer.prune(6)
+        buffer.slice(0, LastOffset) shouldBe Vector(8 -> 16, 13 -> 26)
+      }
+    }
```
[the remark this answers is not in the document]
tudor-da: Yes, it is basically asserting the end inclusive behavior: whether it lands on a vacant offset or not

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
tudor-da: Good catch, thanks! Fortunately, this is harmless. I'll commit a fix

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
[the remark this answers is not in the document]
kamil-da: How about changing `element inserted` to `element not present`? I find it a bit counter-intuitive because after reading the test case header I went for looking where element `6` was inserted :)

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
tudor-da: You are right. For the sake of consistency, they should be inverted.

kamil-da (approved):
👏

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+            val expected = ((idx + 1) to (100 + idx)).map(idx => idx -> idx)
+            buffer.push(1000 + idx, 1000 + idx)
+            buffer.slice(idx, 100 + idx) should contain theSameElementsInOrderAs expected
+        }
+      }
+    }
+  }
+
+  "prune" when {
+    "element found" should {
```
meiersi-da, line 107: I could not find a test that would test that the binary-search implementation works correctly for arrays of size 0, 1, 2, which I'd consider boundary conditions for it.

Consider either testing that indirectly via `prune` here, or also copying the unit tests from the source you've copied the binary-search implementation.

meiersi-da (changes requested):
@tudor-da : thanks for the ping this morning. There might be a data race lurking in the code. See the comments inline.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  "getEvents" when {
+    "called with valid range" should {
+      val bufferElements = increasinglySparse(10).map(offset => offset -> offset * 2)
+
+      "return the correct series" in withBuffer(10, bufferElements) { buffer =>
+        val rangesGen = for {
+          startExclusive <- Gen.choose(0, bufferElements.last._1)
+          endInclusive <- Gen.choose(startExclusive, bufferElements.last._1)
+        } yield (startExclusive, endInclusive)
```
mziolekda, line 73: I am not sure how exhaustive those generators are, I would be very interested in a pinpointed test that explicitly hits the buffer at its extremes i.e. returns the full buffer

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
```diff
@@ -0,0 +1,166 @@
...
+    *
+    * @param startExclusive The start exclusive bound of the requested range.
+    * @param endInclusive The end inclusive bound of the requested range.
+    * @return The series of events as an ordered vector satisfying the input bounds.
+    */
+  def slice(startExclusive: O, endInclusive: O): Vector[(O, E)] =
+    Timed.value(
+      sliceTimer, {
+        val buffer = bufferRef
+        if (bufferRangeEnd.exists(_ < endInclusive)) {
```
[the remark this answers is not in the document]
tudor-da: Yes, you are right. This method was initially synchronized, but decided that it is not needed and forgot to addthe volatile to the field. Thanks!

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
```diff
@@ -0,0 +1,166 @@
...
+          // Ensure vector grows with strictly monotonic offsets.
+          // Only specially-designated range end markers are allowed
+          // to have offsets equal to the buffer range end.
+          if (lastOffset > offset || (lastOffset == offset && !isRangeEndMarker(entry))) {
+            throw UnorderedException(lastOffset, offset)
+          }
+        }
+
+        // Update the buffer range end
+        bufferRangeEnd = Some(offset)
```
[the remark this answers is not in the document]
tudor-da: Yep are right. Between two changes, I dropped all memory syncs on `bufferRangeEnd`. Will update.. 
The read-write conflict can not happen in the scope of this design (the write is always ahead of the read by design), but I agree that it should be defensively implemented in here as well.

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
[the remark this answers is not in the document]
meiersi-da: Note that a volatile field on its own is not enough to avoid violating the invariant that `buffer` contains all elements up to `bufferRangeEnd`. See https://github.com/digital-asset/daml/pull/9775#discussion_r640421507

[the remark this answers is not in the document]
tudor-da: Thanks for spotting the issues. As mentioned above, the invariant was not necessary to hold in this design since the buffer is only updated _ahead_ of the read requests. But you are right, @meiersi-da that we can easily make this invariant hold without external assumptions.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
tudor-da: Fine with that.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
tudor-da: Good idea, why not: Wrote a small suite for that binary search method

ledger/participant-integration-api/src/main/scala/platform/store/cache/EventsBuffer.scala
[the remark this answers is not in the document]
tudor-da: Done. Created a `BufferStateRef` wrapper case class for the two values holding the state, so it can be easily addressed and updated atomically.

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/EventsBufferSpec.scala
tudor-da: Done.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -636,6 +636,14 @@ final class Metrics(val registry: MetricRegistry) {
         val deduplicateCommand: Timer = registry.timer(Prefix :+ "deduplicate_command")
         val stopDeduplicateCommand: Timer = registry.timer(Prefix :+ "stop_deduplicating_command")
         val prune: Timer = registry.timer(Prefix :+ "prune")
+
+        object buffer {
```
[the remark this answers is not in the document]
tudor-da: Done

meiersi-da (approved):
Thanks for fixing all the comments @tudor-da ! Much appreciated!

PR #9776, opened 2021-05-21
test_windows_signing: install for u

garyverhaegen-da:
Turns out "`--global`" means "for this user".

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -59,7 +59,8 @@ jobs:
       artifactName: test-signing
       targetPath: $(Build.StagingDirectory)/test-signing
   - bash: |
-      azuresigntool sign \
+      "/C/Program Files/dotnet/dotnet.exe" tool install --global AzureSignTool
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -59,7 +59,8 @@ jobs:
       artifactName: test-signing
       targetPath: $(Build.StagingDirectory)/test-signing
   - bash: |
-      azuresigntool sign \
+      "/C/Program Files/dotnet/dotnet.exe" tool install --global AzureSignTool
+      /c/Users/u/.dotnet/tools/azuresigntool sign \
```
cocreature, line 63:
```suggestion
      /C/Users/u/.dotnet/tools/azuresigntool sign \
```
doesn’t matter but let’s be consistent between the two lines at least

cocreature (approved):
lgtm thanks!

azure-pipelines.yml
garyverhaegen-da: 👍

PR #9777, opened 2021-05-21
DPP-406 split append only migration

rautenrieth-da:
This PR splits the append-only migration into the following scripts:

- `100.0`: Create new tables and fill them with data
- `100.1`: Drop old tables and create a replacement view
- `100.2`: Create indices (now run after migrating data, which should be faster)
- `100.3`: Vacuum analyze

None of the SQL statements have changed in this PR, it should have no effect on the correctness.


ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V100_1__Append_only_cleanup.sql
```diff
@@ -0,0 +1,186 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
nmarton-da, line 1: this migration is quite lightweight compared to the others, but makes sense from logical composition point of view

nmarton-da (approved):
LGTM
I would have separated more the heavy migration, but if that is already sufficient, I am fine with it

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V100_1__Append_only_cleanup.sql
rautenrieth-da: It is indeed. It's less about reporting progress but more about having the possibility to compare old and new data before the old data is deleted.

PR #9778, opened 2021-05-21
Add ledger API test tool tests for rollback projections

cocreature:
This adds 3 tests for projections under rollback nodes.

The first one is relatively clear hopefully and tests divulgence.

The other two are a bit more intricate. For both of those we can also
not test too much via the ledger API since we don’t actually get
access to rollback nodes. However, it still seems useful to at least
exercise those code paths and make sure they don’t do anything
horribbly wrong.

The second test tests the normalization rules from
https://github.com/digital-asset/daml/blob/main/docs/source/concepts/ledger-model/ledger-exceptions.rst#privacy

The last one tests a more complex structure with deeply nested
rollback nodes and different informees.

See https://docs.google.com/document/d/1wsk0WLDnaQUWB1bPmKbNM_s8n55Dr1-9BptHt1RBq-s/edit#heading=h.ioizcquu71zw (internal doc) for a presentation of those test cases that might be easier to understand than the mix of Scala & Daml here.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/test-common/src/main/daml/semantic/Exceptions.daml
+++ b/ledger/test-common/src/main/daml/semantic/Exceptions.daml
@@ -258,7 +258,7 @@ template ExceptionTester
         pbcInformer : ContractId Informer
       controller p
       do -- Projection to B:
-         --   Normalization rule #3 folds R3 intüo R1
+         --   Normalization rule #3 folds R3 into R1
          --   R2 and R3 must sit under R1 for key checks to pass
          -- Projection to C:
          --   Normalization rule #2 pulls R2 out of R1.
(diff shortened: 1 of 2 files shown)


ledger/test-common/src/main/daml/semantic/Exceptions.daml
```diff
@@ -130,24 +217,114 @@ template ExceptionTester
...
+    nonconsuming choice ProjectionNormalization : ()
+      with
+        b : Party
+        keyDelegate : ContractId WithKeyDelegate
+        pbInformer : ContractId Informer
+        pcInformer : ContractId Informer
+        pbcInformer : ContractId Informer
+      controller p
+      do -- Projection to B:
+         --   Normalization rule #3 folds R3 intüo R1
```
sofiafaro-da, line 261:
```suggestion
         --   Normalization rule #3 folds R3 into R1
```

sofiafaro-da (approved):
This looks great, thank you!

PR #9779, opened 2021-05-21
fix azuresigntool invocation

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,7 +71,7 @@ jobs:
         --timestamp-rfc3161 "http://timestamp.digicert.com" \
         --file-digest sha256 \
         --verbose \
-        $(Build.StagingDirectory)/test-signing/installer.exe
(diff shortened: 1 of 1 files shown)


ci/prs.yml
```diff
@@ -11,6 +11,51 @@ pr:
       - release/*
 
 jobs:
+
+- job: test_windows_signing
```
cocreature, line 15: I assume you’re still debugging this which is why it is in the PRs build?

garyverhaegen-da: Yes. Currently, on main, it looks like it's not running at all. Trying to debug this on main is not only super inconvenient, it's also a bit dangerous. By running this on a PR, I can print out whatever I want as the secrets are just not available to the pipeline to start with.

cocreature (approved):
Thanks!

PR #9780, opened 2021-05-21
debug signtool on main

garyverhaegen-da:
I don't know what's going on. I was hoping I could try to debug from
PRs, but that doesn't seem to work.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -59,19 +59,26 @@ jobs:
       artifactName: test-signing
       targetPath: $(Build.StagingDirectory)/test-signing
   - bash: |
+      ls -l '$(Build.StagingDirectory)'\\test-signing\\installer.exe
       "/C/Program Files/dotnet/dotnet.exe" tool install --global AzureSignTool
+      ls -l /C/Users/u/.dotnet/tools/azuresigntool
       /C/Users/u/.dotnet/tools/azuresigntool sign \
         --azure-key-vault-url "$AZURE_KEY_VAULT_URL" \
         --azure-key-vault-client-id "$AZURE_CLIENT_ID" \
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -59,7 +59,9 @@ jobs:
       artifactName: test-signing
       targetPath: $(Build.StagingDirectory)/test-signing
   - bash: |
+      ls -l '$(Build.StagingDirectory)'\\test-signing\\installer.exe
       "/C/Program Files/dotnet/dotnet.exe" tool install --global AzureSignTool
+      ls -l /C/Users/u/.dotnet/tools/azuresigntool
```
cocreature, line 64: looking at this I wonder if we might be screwing yourself over with path conversion somewhere. I vaguely recall that bash’s automatic path conversion caused very weird issues in the past in a few places and the easiest solution was to just use powershell for those tasks.

cocreature (approved):

azure-pipelines.yml
garyverhaegen-da: There is a bad interaction between Bash escaping and YAML escaping: lines like
```bash
$(Build.StagingDirectory)/test.exe
```
will get mangled because Azure YAML substitution goes first, so Bash sees
```bash
C:\Path\With\Backslashes/test.exe
```
and it interprets the backslashes as escaping characters, yielding
```bash
C:PathWithBackslashes/test.exe
```
which does not exist. Adding single quotes around the YAML substitution makes it look a bit weird from a Bash perspective (single quotes meaning no substitution) but make this work properly as the YAML substitution ignores the single quotes, but the single quotes do tell Bash to not interpret the backslashes as escape characters.

If you were thinking of another issue, then I'm not (yet) aware of it.

PR #9784, opened 2021-05-24
Update LF spec for exceptions.

sofiafaro-da:
The changes here are:

* Removing the built-in exception types
* Changing the result in the operational semantics to include both an "exception thrown" case and a "fatal error" case.
* Cleaning up the semantics of create/exercise/etc and try/catch.

Not included in this PR is anything to do with built-in arithmetic exceptions. There's room to add it in the future (as a value of AnyException type), but I would do it in a follow-up PR.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/spec/daml-lf-1.rst
```diff
@@ -2192,20 +2150,11 @@ types that satisfies the following rules::
...
-  ——————————————————————————————————————————————————— TypeOrderGeneralErrorArithmeticError
-    'GeneralError' <ₜ 'ArithmeticError'
-
-  ——————————————————————————————————————————————————— TypeOrderArithmeticErrorContractError
-    'ArithmeticError' <ₜ 'ContractError'
-
-  ——————————————————————————————————————————————————— TypeOrderContractErrorBigNumeric
-    'ContractError' <ₜ 'BigNumeric'
+  ——————————————————————————————————————————————————— TypeOrderAnyExceptionBigNumeric
+    'AnyException' <ₜ 'BigNumeric'
```
sofiafaro-da, line 2157: @remyhaemmerle-da
Are AnyException values comparable (in the value order)?
Is `AnyException` ordered as a type (in the type order)?

cocreature: https://github.com/digital-asset/daml/blob/fb6f72c81d983eebe91b87f491238e4d414573c2/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/TypeOrdering.scala#L92 for the type order. Afaict, it’s not orderable by value atm see https://github.com/digital-asset/daml/blob/main/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/svalue/Ordering.scala but I think it should be. I don’t see a compelling reason why `Any` is orderable but `AnyException` is not.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3253,65 +3191,74 @@ as described by the ledger model::
...
-     u₁ ‖ S₀  ⇓ᵤ  (Err @τ₁ v₁, tr₁)
-     e₂[x ↦ 'to_any_exception' @τ₁ v₁]  ⇓  Ok ('Some' @σ u₂)
+     u₁ ‖ S₀  ⇓ᵤ  (Err (Throw v₁), tr₁)
+     e₂[x ↦ v₁]  ⇓  Ok ('Some' @σ u₂)
      u2 ‖ S₀  ⇓ᵤ  (Ok v₂, tr₂) ‖ S₂
-   —————————————————————————————————————————————————————————————————————— EvUpdTryCatchErr2_OkHandle_Ok
+   —————————————————————————————————————————————————————————————————————— EvUpdTryCatchThrow2_OkHandle_Ok
      'try' @τ e₁ 'catch' x. e₂ ‖ S₀
        ⇓ᵤ
      (Ok v₂, ('rollback' tr₁) ⋅ tr₂) ‖ S₂
```
cocreature, line 3229: I guess technically we need to insert normalization here or alternatively describe it as a postprocessing step somewhere.

cocreature: Similarly for the other cases below.

cocreature (approved):
Nice, thank you!

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: We have a minor issue here.
Either we make it comparable, and we have to revive `ArithmeticError` either we wait we have the LF standard library :cry:

remyhaemmerle-da (approved):
LGTM.

daml-lf/spec/daml-lf-1.rst
cocreature: We only need it in the spec though not exposed right?

remyhaemmerle-da: As soon you make ArithmeticError (wrapped in a AnyException) and an user Exception (wrapped in a AnyException) comparable and fix the order, I consider it as exposed.

remyhaemmerle-da: In other word, once you make it comparable, you have to keep it in the speedy code.

cocreature: Ah I see because moving it to a standard library would change the order and unless we go for a magic package id (like 0000 or ffff…) we have no way to make sure that ArithmeticError comes always before or always after everything else. In that case, I think we’re better off not adding ordering for now.

sofiafaro-da: 👍 ok, so AnyException values are not comparable for now.


Also I just noticed / found that `AnyException` as a type is ordered after `RoundingMode` (in the TypeRep order), so I'll update the spec.

daml-lf/spec/daml-lf-1.rst
sofiafaro-da: I think we should keep it as a separate step, since that's closer to the implementation.

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: We cannot really make `AnyException` comparable in the future, if we do not do it now.  
Comparison is not version aware, and I think it is pretty impossible to make it version aware.

PR #9785, opened 2021-05-24
`ledger-api-bench-tool`: max consumption delay SLO [DPP-400]

kamil-da:
### Changes:
- new CLI parameter for stream configuration: `max-delay` for specifying max allowed consumption delay
- generic mechanism for service-level objectives
- returning the exit code `1` for stream runs with violated SLOs or with other errors

### Pull Request Checklist

trigger the build.


ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/LedgerApiBenchTool.scala
```diff
@@ -59,31 +69,46 @@ object LedgerApiBenchTool {
...
+                    )(system)
                   transactionService.transactionTrees(streamConfig, observer)
                 }
           }
         }
-        .map(_ => ())
+        .transform {
+          case Success(results) =>
+            if (results.contains(MetricsManager.Message.MetricsResult.ObjectivesViolated))
+              Failure(new RuntimeException("Metrics objectives violated."))
```
tudor-da, line 107: `violated` makes me thing about some criticality. How about `not met`?

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -7,37 +7,51 @@ import com.google.protobuf.timestamp.Timestamp
...
 }
 
 object Metric {
+  trait MetricValue {
+    def formatted: List[String]
+  }
+
+  trait ServiceLevelObjective[MetricValueType <: MetricValue] {
+    def isViolatedBy(metricValue: MetricValueType): Boolean
+    def moreViolatingOf(first: MetricValueType, second: MetricValueType): MetricValueType
```
tudor-da, line 35: I can't infer the actual purpose of this method. A better name or/and description might help here.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -141,13 +193,43 @@ object Metric {
...
 
-    def empty[T](recordTimeFunction: T => Seq[Timestamp], clock: Clock): DelayMetric[T] =
-      DelayMetric(recordTimeFunction, clock)
+    sealed trait DelayObjective extends ServiceLevelObjective[Value]
+    object DelayObjective {
+      final case class MaxDelay(maxDelaySeconds: Long) extends DelayObjective {
+        override def isViolatedBy(metricValue: Value): Boolean =
+          metricValue.meanDelaySeconds.exists(_ > maxDelaySeconds)
+
+        override def moreViolatingOf(first: Value, second: Value): Value =
```
tudor-da, line 219: This method is a member of `MaxDelay`, but it doesn't really have anything to do with it, does it? (e.g. it doesn't take into account the object's context  - `maxDelaySeconds`

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -141,13 +193,43 @@ object Metric {
...
+      )
+
     final case class Value(meanDelaySeconds: Option[Long]) extends MetricValue {
       override def formatted: List[String] =
         List(s"mean delay: ${meanDelaySeconds.getOrElse("-")} [s]")
     }
 
-    def empty[T](recordTimeFunction: T => Seq[Timestamp], clock: Clock): DelayMetric[T] =
-      DelayMetric(recordTimeFunction, clock)
+    sealed trait DelayObjective extends ServiceLevelObjective[Value]
```
tudor-da, line 213: Are we going to have another `DelayObjective` aside from `MaxDelay`?

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
tudor-da: I see this is the max function over `MetricValueType`s. Can you extract and use it as such
e.g. 
```
implicit val metricsOrdering: Ordering[MetricValueType] = Ordering.by{...}
```

tudor-da: usage example from your test
```
      forAll(cases) { (first, second, expected) =>
        Ordering[MaxDelay].max(first, second) shouldBe expected
      }
```

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -122,13 +141,46 @@ object Metric {
...
+        newValue: DelayMetric.Value
+    ): Map[DelayMetric.DelayObjective, Option[DelayMetric.Value]] = {
+      objectives
+        .map { case (objective, currentViolatingValue) =>
+          // verify if the new value violates objective's requirements
+          if (objective.isViolatedBy(newValue)) {
+            currentViolatingValue match {
+              case None =>
+                // if the new value violates objective's requirements and there is no other violating value,
+                // record the current value
```
tudor-da, line 154:
```suggestion
                // record the new value
```

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -122,13 +141,46 @@ object Metric {
...
+                objective -> Some(newValue)
+              case Some(currentValue) =>
+                // if the new value violates objective's requirements and there is already a value that violates
+                // requirements, record the extreme value of the two
```
tudor-da, line 158:
```suggestion
                // requirements, record the maximum value of the two
```

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
kamil-da: It's sort of max function but in a context of an objective. I want to be able to get the most objective-violating value, so the method is to get a comparison between values. 
My thinking was that we could have an objective for max value, and another on min value, so different comparison functions would be required in different contexts.
Now I think such use case is useless, because in reality we will have only objectives asserting that performance is greater than a threshold and never in the other way.

kamil-da: I will implement `Ordering` instances as you suggest 👍

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
kamil-da: I've explained my motivation in the comment above.
I'll move out the logic out of the objective 👍

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
kamil-da: I don't expect any. The `DelayObjective` trait was added to set the `SeviceLevelObjective[T]` type parameter to a specific type but I'm not attached to this solution :)

tudor-da: But you can specify the type as well in one step no?
`final case class MaxDelay(maxDelaySeconds: Long) extends ServiceLevelObjective[Value]`

kamil-da: You're right. It's a relic of the past commits :) I'll change this.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/MetricsManager.scala
```diff
@@ -13,9 +13,14 @@ object MetricsManager {
 
   sealed trait Message[T]
```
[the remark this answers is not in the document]
kamil-da: I need the type parameter to be able to abstract over stream element type in metrics. Metrics expose `def onNext(value: T)` method which is in turn used by the `MetricsManager[T]`.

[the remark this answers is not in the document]
tudor-da: But you don't use the type downstream, it's unchecked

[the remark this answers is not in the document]
kamil-da: Do you propose instead of having `Message.NewValue[T](value: T)` to have `Message.NewValue(value: Any)`?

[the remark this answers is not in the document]
tudor-da: Example https://github.com/digital-asset/daml/commit/56f006043abced201c3685e7d1c987ee5eac183a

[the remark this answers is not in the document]
tudor-da: >Do you propose instead of having Message.NewValue[T](value: T) to have Message.NewValue(value: Any)?

No, you can have the type parameter where it is needed: only for `NewValue[T]`. The other case classes don't need it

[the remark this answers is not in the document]
kamil-da: You're right! Thank you :)

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/LedgerApiBenchTool.scala
kamil-da: Are you fine with changing just error message and leaving changing variables/classes names for a follow-up PR?

tudor-da: Sure

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
kamil-da: I've implemented `Ordering[MaxDelay.Value]` and renamed `def moreViolatingOf()` to `def maxViolatingValue()`.
On seconds thoughts I feel that having this method in `ServiceLevelObjective[T]` trait makes concerns separated more cleanly among classes. A `Metric` knows how to compute metric values based on elements flowing in the stream while a `ServiceLevelObjective` knows which value is not allowed and which value is "better"/"worse" than another.
However, if you think this is unintuitive I can change it to be easier to use.

kamil-da:
@tudor-da Thank you for the review!
I've made some changes to meet your suggestions and left some comments as well.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
```diff
@@ -141,13 +194,47 @@ object Metric {
...
+          case (None, None) => 0
+        }
+      }
+    }
+
+    final case class MaxDelay(maxDelaySeconds: Long) extends ServiceLevelObjective[Value] {
+      override def isViolatedBy(metricValue: Value): Boolean =
+        metricValue.meanDelaySeconds.exists(_ > maxDelaySeconds)
+
+      override def maxViolatingValue(first: Value, second: Value): Value =
```
tudor-da, line 232: Why do you need this method still? Can't you inline it where it is used since it doesn't use anyway anything from the `MaxDelay` context?

ledger/ledger-api-bench-tool/src/test/suite/scala/com/daml/ledger/api/benchtool/metrics/DelayMetricSpec.scala
```diff
@@ -11,6 +11,8 @@ import org.scalatest.wordspec.AnyWordSpec
 import java.time.{Clock, Duration, Instant, ZoneId}
 import scala.util.Random
 
+import scala.language.existentials
```
tudor-da, line 14: Are you using existential types?

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
tudor-da: I hate being too opinionated on random stuff, but here I think this method should follow the typeclass pattern. It is just a `Ordering[MaxDelay.Value]` that can be instantiated in e.g. `MaxDelay` companion object and brought into scope wherever you need to perform the `max` or other ordering ops. Having it as a prototype of `ServiceLevelObjective` gives me a false sense of cohesion: it is not using any context like its siblings `isViolatedBy` and `formatted`, whereas you only restrict its usage if you already have a `ServiceLevelObjective` object.

tudor-da (approved):
There are still two open comments but minor. Nice job 👍

ledger/ledger-api-bench-tool/src/test/suite/scala/com/daml/ledger/api/benchtool/metrics/DelayMetricSpec.scala
kamil-da: Yes, specifically when calling `.violatedObjectives` method.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/Metric.scala
kamil-da: It's just a matter of separation on responsibilities. I cannot convince myself to hardcode `Ordering[Type].max(a,b)` inside a metric because we could potentially have a metric where one objective was for finding a max value, and another for finding a min value.
But as this is only a potential problem that might not even occur, I will inline this to make it more clean just as you suggest.
Thanks!

PR #9790, opened 2021-05-25
LF: Drop Builtin Exceptions completly

remyhaemmerle-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
+++ b/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
@@ -3,7 +3,7 @@
 
 -- @SINCE-LF-FEATURE DAML_EXCEPTIONS
 -- @ERROR range=130:1-130:23; Unhandled exception: ExceptionSemantics:E
--- @ERROR range=145:1-145:25; Unhandled exception: ArithmeticError while evaluating (DIV_INT64 1 0).
+-- @ERROR range=145:1-145:25; ArithmeticError while evaluating (DIV_INT64 1 0).
 -- @ERROR range=182:1-182:11; Attempt to exercise a consumed contract
 module ExceptionSemantics where
 
--- a/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
+++ b/compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
@@ -3,7 +3,7 @@
 
 -- @SINCE-LF-FEATURE DAML_EXCEPTIONS
 -- @ERROR range=130:1-130:23; Unhandled exception: ExceptionSemantics:E
--- @ERROR range=145:1-145:25; ArithmeticError while evaluating (DIV_INT64 1 0).
+-- @ERROR range=145:1-145:25; Unhandled exception: DA.Exception.ArithmeticError:ArithmeticError@f1cf1ff41057ce327248684089b106d0a1f27c2f092d30f663c919addf173981 with message = "ArithmeticError while evaluating (DIV_INT64 1 0)."
 -- @ERROR range=182:1-182:11; Attempt to exercise a consumed contract
 module ExceptionSemantics where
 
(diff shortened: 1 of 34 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -176,15 +176,14 @@ private[speedy] sealed abstract class SBuiltin(val arity: Int) {
         )
     }
 
-  final protected def getSException(args: util.ArrayList[SValue], i: Int): SException =
+  final protected def getSException(args: util.ArrayList[SValue], i: Int): SAnyException =
```
sofiafaro-da, line 179: Might be worth renaming this function `getSAnyException`

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -207,13 +203,29 @@ object SValue {
...
+      val array = new util.ArrayList[SValue](1)
+      array.add(
+        SText(s"ArithmeticError while evaluating ($builtinName ${args.iterator.mkString(" ")}).")
+      )
+      SAnyException(typ, SRecord(tyCon, fields, array))
+    }
+    // Assumes excep is properly typed
+    def unapply(excep: SAnyException): Option[SValue] =
+      excep match {
+        case SAnyException(`typ`, SRecord(_, _, args)) =>
```
sofiafaro-da, line 224: Scala Q: Is this matching against the `typ` value?

compiler/damlc/tests/daml-test-files/ExceptionSemantics.daml
```diff
@@ -3,7 +3,7 @@
 
 -- @SINCE-LF-FEATURE DAML_EXCEPTIONS
 -- @ERROR range=130:1-130:23; Unhandled exception: ExceptionSemantics:E
--- @ERROR range=145:1-145:25; Unhandled exception: ArithmeticError while evaluating (DIV_INT64 1 0).
+-- @ERROR range=145:1-145:25; ArithmeticError while evaluating (DIV_INT64 1 0).
```
cocreature, line 6: Can we keep the full error here to make sure this gets propagated as an `Unhandled exception`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
cocreature: Yes

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -207,13 +203,29 @@ object SValue {
...
-  final case class SArithmeticError(
-      builtinName: String,
-      args: ImmArray[String],
-  ) extends SException {
-    def message = s"ArithmeticError while evaluating ($builtinName ${args.iterator.mkString(" ")})."
+  final case class SAnyException(ty: Type, value: SValue) extends SValue
+
+  object SArithmeticError {
+    val tyCon: Ref.TypeConName = Ref.Identifier.assertFromString(
+      "f1cf1ff41057ce327248684089b106d0a1f27c2f092d30f663c919addf173981:DA.Exception.ArithmeticError:ArithmeticError"
```
cocreature, line 210: maybe add a comment that this should be kept in sync with the generator or the tests.

cocreature (approved):
Nice! The one issue in this approach that I could come up with after thinking about it for a while is that if we start exposing unhandled exceptions on the ledger API we will produce values for which the ledger might not have a Dalf. I don’t expect that to be that big of an issue in practice but it is a bit annoying.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
sofiafaro-da: Thanks for the explanation :-)

cocreature (approved):
Thanks!

PR #9791, opened 2021-05-25
upgrade to Scalaz 7.3.3

S11001001:
Significant changes:

* scalaz/scalaz#1469
* scalaz/scalaz#1479
* scalaz/scalaz#1687
* scalaz/scalaz#1305

* [x] merge #9997

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/SortedLookupList.scala
```diff
@@ -8,7 +8,8 @@ import ScalazEqual.{equalBy, orderBy}
 import scalaz.{Applicative, Equal, Order, Traverse}
 import scalaz.std.tuple._
 import scalaz.std.string._
-import scalaz.syntax.traverse._
+import scalaz.syntax.functor0._
```
cocreature, line 11: What exactly is the difference between `functor0` and `functor`? (and same for traverse)

cocreature (approved):
thanks

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/SortedLookupList.scala
realvictorprm: So far I understand from [here](https://github.com/scalaz/scalaz/pull/1479) this seems to be the `functor` only import instead of getting transitive dependencies of a functor too.

ledger-service/http-json-ledger-client/src/main/scala/com/daml/http/LedgerClientBase.scala
```diff
@@ -82,9 +82,9 @@ trait LedgerClientBase extends StrictLogging {
       aesf: ExecutionSequencerFactory,
   ): Future[LedgerClientBase.Error \/ DamlLedgerClient] =
     buildLedgerClient(ledgerHost, ledgerPort, clientConfig, nonRepudiationConfig)
-      .map(_.right)
+      .map(\/.r[LedgerClientBase.Error](_))
       .recover { case NonFatal(e) =>
-        \/.left(
+        \/.l[DamlLedgerClient](
```
realvictorprm, line 87: Can I somewhere read up why this was decided? I like types but explicitly typing everywhere while I have type inference is not my strongest preference.

realvictorprm: Note, take my question with a grain of salt, as my knowledge about this code here is pretty limited :grimacing:

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -445,7 +446,7 @@ class WebSocketService(
       .map { case (oeso, ejv) =>
         for {
           offPrefix <- oeso.sequence
-          jv <- ejv
+          jv <- ejv.widenLeft[Error]
```
realvictorprm, line 449: I'm also wondering whats up with the widening here :see_no_evil:

realvictorprm (approved):
I'm not that familiar with the code, however I didn't see anything which looks weird to me. Thus, looks good to me :)

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/SortedLookupList.scala
S11001001: Exactly.

ledger-service/http-json-ledger-client/src/main/scala/com/daml/http/LedgerClientBase.scala
S11001001: Start here https://github.com/digital-asset/daml/pull/9997#discussion_r651998142

Note that this is not necessary if you have context ([expected type](https://discuss.daml.com/t/scala-inference-rules-of-thumb/2242#type-parameters-can-be-inferred-from-expected-type-5), which goes backwards into lambdas). I simply added this where I didn't have context.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
S11001001: You know, it might be worth adding another implicit to `widenLeft` et al to ensure that it isn't a no-op, i.e. it will fail to compile if the source type already equals the goal type. What do you think?

cocreature:
@S11001001 do you want to merge this before it bitrots too much?

S11001001:
@cocreature aye; I wanted #9997 to "settle" a bit first, for whatever it's worth.

realvictorprm:
We should not let this age too long, it's already getting out of sync :sweat: In this sense, ping :smile:

cocreature:
@S11001001 any plans to revive & merge this?

S11001001:
@cocreature Next time I'm not immediately busy with other things.

garyverhaegen-da:
@S11001001 Is this still on the table?

stephencompall-DA:
This might never happen; serves as a starting point if it does.

PR #9792, opened 2021-05-25
[In memory fan-out] Transaction log updates stream implementation

tudor-da:
### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -246,6 +249,102 @@ private[appendonlydao] final class TransactionsReader(
...
+        size = outputStreamBufferSize,
+      )
+      .concat(endMarker)
+  }
+
+  private def splitRange(
+      startExclusive: Long,
+      endInclusive: Long,
+      numberOfChunks: Int,
+  ): Seq[EventsRange[Long]] =
```
kamil-da, line 309: A few unit tests for this method would be great.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionLogUpdatesReader.scala
```diff
@@ -0,0 +1,274 @@
...
+    override def getMessage: String = s"Missing mandatory field $field"
+  }
+
+  final case class InvalidEventKind(eventKind: Int) extends RuntimeException {
+    override def getMessage: String =
+      s"Invalid event kind: $eventKind"
+  }
+
+  private object EventKind {
+    val Create = 10
```
kamil-da, line 236: There are a few classes/methods common for both `TransactionLogUpdatesReader` and `ContractStateEventsReader`, so maybe we could extract some common parts out of them into a single place. What do you think? Do you prefer to keep both readers isolated?

tudor-da: I would have gone for abstraction, but this one is actually going to replace the `ContractStateEventsReader`

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: Yeah, why not, good practice. Added tests

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionLogUpdatesReader.scala
kamil-da: Ok, thanks!

kamil-da (approved):

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -353,6 +353,11 @@ final class Metrics(val registry: MetricRegistry) {
       val uploadPackages: Timer = registry.timer(Prefix :+ "upload_packages")
       val publishConfiguration: Timer = registry.timer(Prefix :+ "publish_configuration")
 
+      val decodeTransactionLogUpdate: Timer =
+        registry.timer(Prefix :+ "decode_transaction_log_update")
```
meiersi-da, line 357: Consider `transaction_log_update_decode` to improve the logical grouping when lexicographically sorting metric names. .

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
```diff
@@ -390,4 +384,21 @@ private[dao] final class TransactionsReader(
     }
     mat
   }
+
+  override def getContractStateEvents(startExclusive: (Offset, Long), endInclusive: (Offset, Long))(
+      implicit loggingContext: LoggingContext
+  ): Source[((Offset, Long), ContractStateEvent), NotUsed] =
+    throw new UnsupportedOperationException(
+      "The operation is not supported in the current version."
```
meiersi-da, line 392: The reason for not being supported is that it is not yet implemented isn't it?

The message currently just repeats the name of the exception. Consider a message like "getContractStateEvents: not yet implemented."

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
```diff
@@ -390,4 +384,21 @@ private[dao] final class TransactionsReader(
...
+    )
+
+  override def getTransactionLogUpdates(
+      startExclusive: (Offset, Long),
+      endInclusive: (Offset, Long),
+  )(implicit
+      loggingContext: LoggingContext
+  ): Source[((Offset, Long), TransactionLogUpdate), NotUsed] =
+    throw new UnsupportedOperationException(
+      "The operation is not supported in the current version."
```
meiersi-da, line 402: Same here.

ledger/participant-integration-api/src/main/scala/platform/store/interfaces/TransactionLogUpdate.scala
```diff
@@ -0,0 +1,86 @@
...
+trait TransactionLogUpdate extends Product with Serializable
+
+object TransactionLogUpdate {
+  final case class Transaction(
+      transactionId: String,
+      commandId: String,
+      workflowId: String,
+      effectiveAt: Instant,
+      offset: Offset,
+      lastEventSequentialId: EventSequentialId,
```
meiersi-da, line 25: Why do we need this? Couldn't it be computed on-demand?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -446,3 +523,35 @@ private[appendonlydao] final class TransactionsReader(
     mat
   }
 }
+
+private[appendonlydao] object TransactionsReader {
+  private[appendonlydao] def splitRange(
```
meiersi-da, line 528: What is this method used for?

What can a caller expect?

ledger/participant-integration-api/src/main/scala/platform/store/interfaces/TransactionLogUpdate.scala
```diff
@@ -0,0 +1,86 @@
...
+import com.daml.ledger.participant.state.v1.Offset
+import com.daml.lf.data.Ref.IdString
+import com.daml.lf.ledger.EventId
+import com.daml.platform.store.appendonlydao.events
+import com.daml.platform.store.appendonlydao.events.{ContractId, Identifier}
+import com.daml.platform.store.cache.MutableCacheBackedContractStore.EventSequentialId
+
+trait TransactionLogUpdate extends Product with Serializable
+
+object TransactionLogUpdate {
```
meiersi-da, line 18: I'd expect some commentary here, as to what the purpose of the classes within this `object` are. Otherwise, readers can only infer it from how it is currently being used, which seems like a recipe for keeping accidental dependencies intact; and adding new ones.

meiersi-da (commented):
@tudor-da : consider adding the commentary on the PR as comments on the code in the PR to persist it in time. 

Otherwise, LGTM.

ledger/participant-integration-api/src/main/scala/platform/store/interfaces/TransactionLogUpdate.scala
tudor-da: Yes we can. Done

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -246,6 +251,78 @@ private[appendonlydao] final class TransactionsReader(
...
+      endInclusive -> TransactionLogUpdate.LedgerEndMarker(
+        eventOffset = endInclusive._1,
+        eventSequentialId = endInclusive._2,
+      )
+    )
+
+    val eventsSource = Source
+      .fromIterator(() =>
+        TransactionsReader
+          .splitRange(
```
mziolekda, line 270: when split range throws an exception we do no recover

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -446,3 +523,46 @@ private[appendonlydao] final class TransactionsReader(
...
+    numberOfChunks match {
+      case 1 =>
+        Seq(EventsRange(startExclusive, endInclusive))
+
+      case invalid if invalid < 1 =>
+        throw new IllegalArgumentException(
+          s"You can only split a range in a strictly positive number of chunks ($numberOfChunks)"
+        )
+
+      case _ if (rangeSize / numberOfChunks.toLong) < minChunkSize.toLong =>
```
mziolekda, line 554: This condition seems to lead to unintended results e.g. rangeSize = 900, numberOfChunks = 100 will result in a single large range of 900

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: When it throws, it's actually unrecoverable, not much we can do about it. The error condition is actually reachable only via a programming error, so far-fetched and unrecoverable

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
tudor-da: Good point, done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsReader.scala
tudor-da: Done

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: Done, added a comment

ledger/participant-integration-api/src/main/scala/platform/store/interfaces/TransactionLogUpdate.scala
tudor-da: Added docs

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: Furthermore, the consumers can also retry, which we actually do for `getContractStateEvents` and we will do with the consumer of this as well. Actually the good point here is to somehow take down the entire process since it can never recover. What do you think?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: But I guess you're right. Let's make it more defensive and adapt if somehow the parameters are wrongly set.

mziolekda (commented):
Found two problems that need addressing

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
mziolekda: The solution is simple:
effectiveNumChunks = ceil(rangeSize/minChunkSize.toLong)

tudor-da: Done

mziolekda (approved):

PR #9793, opened 2021-05-25
DPP-393 Activate JdbcLedgerDao suite for the append-only schema

rautenrieth-da:
This PR:
- fixes a few minor issues in the JdbcLedgerDao test suite and the append-only schema
- enables the JdbcLedgerDao test suite for the append-only schema

The PR is best reviewed commit by commit.


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
```diff
@@ -50,34 +50,31 @@ private[events] object ContractsTable extends PostCommitValidationData {
       def lookup(id: ContractId): Option[Option[Instant]] =
         SQL"""
   WITH archival_event AS (
-         SELECT participant_events.*
-           FROM participant_events, parameters
+         SELECT participant_events_consuming_exercise.*
```
nmarton-da, line 53: Is this an independent change, or somehow it is needed for the tests to pass?
(if it is independent I would suggest not doing it in here...especially so that in the current consistent approach on the db read side using the view might give us leverage)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -983,6 +988,19 @@ private class JdbcLedgerDao(
       }
   }
 
+  // The old indexer (com.daml.platform.indexer.JdbcIndexer) uses IncrementalOffsetStep,
+  // and we have tests in JdbcLedgerDao*Spec checking that this is handled correctly.
+  // The append-only schema and the parallel ingestion indexer doesn't use this class.
+  // TODO append-only: Remove IncrementalOffsetStep along with this method and all corresponding tests.
```
nmarton-da, line 994: Do we have a reason to keep the OffsetStep at all? If not I would change this TODO, to let it be the TODO for that.
(It seems to me the changes here changing the LedgerDao methods Offset -> OffsetStep needs to be reverted upon cleanup...was that your intention?)

nmarton-da (commented):
This approach (using same test suit for both append only and mutating schemas) really needed some not-that-nice-hammering, but it is a good tradeoff altogether compared to duplicating the test suite :+1: 

DPP-393 is also about activating postgre-appendonly tests for sandbox-classic. That seems to be missing here. Do you plan to add that in a later PR?

(I was not reviewing the divulgence vs. maximum ledger time test changes)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
rautenrieth-da: Good point, not only can we remove `IncrementalOffsetStep`, we can remove the whole `OffsetStep` trait. I'll change the TODO.

The changes around `Offset -> OffsetStep` were only required so that the append-only JdbcLedgerDao can use the existing `JdbcLedgerDao*Spec` tests.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ContractsTable.scala
rautenrieth-da: It is independent, I'll remove it from this PR.

> using the view might give us leverage

Is there any advantage to using the view? Using the partitions directly seems strictly better to me (more safe and more performant).

tudor-da (approved):
LGTM

PR #9802, opened 2021-05-26
Parallelize daml-assistant integration tests.

sofiafaro-da:
This mainly involves avoiding changing the working directory or the environment while inside tests.

AFAIK the daml start tests can't be parallelized without starting up a sandbox instance each time, which seems bad. These tests could be separated from the rest...

The magic number "2" was picked via experimentation:

```
1 threads -> 306 s
2 threads -> 199 s
3 threads -> 198 s
4 threads -> 195 s
```

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-assistant/integration-tests/BUILD.bazel
+++ b/daml-assistant/integration-tests/BUILD.bazel
@@ -106,7 +106,6 @@ da_haskell_library(
 da_haskell_test(
     name = "integration-tests",
     timeout = "long",
-    tags = ["cpu:2"]
     srcs = [
         "src/DA/Daml/Assistant/FreePort.hs",
         "src/DA/Daml/Assistant/IntegrationTests.hs",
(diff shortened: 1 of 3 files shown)


cocreature:
> AFAIK the daml start tests can't be parallelized without starting up a sandbox instance each time, which seems bad. These tests could be separated from the rest...

I think at least parts of it should be parallelizable by creating different parties per test. I think the one thing where this doesn’t apply is the reload test.

cocreature (approved):
Nice, thank you!

sofiafaro-da:
> I think at least parts of it should be parallelizable by creating different parties per test. I think the one thing where this doesn’t apply is the reload test.

Good point. That sounds like it might be worth exploring in a future PR.

cocreature:
> Good point. I think that might be worth exploring in a future PR.

Definitely something for a separate PR and probably not super urgent either, just wanted to mention the idea.

PR #9804, opened 2021-05-26
LF: SBAnyExceptionMessage queries for unknown packages.

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1575,15 +1575,30 @@ private[lf] object SBuiltin {
 
   /** $any-exception-message :: AnyException -> Text */
   final case object SBAnyExceptionMessage extends SBuiltin(1) {
+    private def exceptionMessage(tyCon: TypeConName, value: SValue, machine: Machine) =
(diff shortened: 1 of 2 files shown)


cocreature (commented):
This seems pretty bad. This means that the change to drop ArithmeticError wasn’t as safe as we assumed. Now you can build a type correct LF program that fails with a dependency error at runtime :/ cc @sofiafaro-da 
Not quite sure if this is really a route we want to go down.

remyhaemmerle-da:
> This seems pretty bad. This means that the change to drop ArithmeticError wasn’t as safe as we assumed. Now you can build a type correct LF program that fails with a dependency error at runtime :/ cc @sofiafaro-da
> Not quite sure if this is really a route we want to go down.

You are right about `ArithmeticError`, let thinks a bit more about PR #9790.
However, we should anyway add this PR to the repo. The builtin should not assume the package is loaded.

sofiafaro-da:
> > This seems pretty bad. This means that the change to drop ArithmeticError wasn’t as safe as we assumed. Now you can build a type correct LF program that fails with a dependency error at runtime :/ cc @sofiafaro-da
> > Not quite sure if this is really a route we want to go down.
> 
> You are right about `ArithmeticError`, let think a bit more about PR #9790.
> However, we should anyway add this PR to the repo. The builtin should not assume the package is loaded.

It's not very elegant, but `ANY_EXCEPTION_MESSAGE` could special case ArithmeticError, so you always get the message field even when the package isn't loaded.

PR #9807, opened 2021-05-26
Add reference docs for exceptions

cocreature:
Not that these are reference-style docs so they are deliberately brief
and don’t focus on usecases. I’ll add a section to the Daml intro for
that in a separate PR.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/daml/reference/exceptions.rst
+++ b/docs/source/daml/reference/exceptions.rst
@@ -1,0 +1,89 @@
+.. Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
+.. _exceptions:
+
+Exceptions
+##########
+
+Exceptions are a Daml feature which allows provides a way to handle
+certain errors that arise during interpretation instead of aborting
+the transaction and rollback changes that happened before the
+error.
+
+There are two types of errors:
+
+Builtin Errors
+..............
+
+.. list-table::
+   :header-rows: 1
+
+   * - Exception type
+     - Thrown on
+   * - ``GeneralError``
+     - Calls to ``error`` and ``abort``
+   * - ``ArithmeticError``
+     - Arithmetic errors like overflows and division by zero
+   * - ``PreconditionFailed``
+     - ``ensure`` statements that return ``False``
+   * - ``AssertionFailed``
+     - Failed ``assert`` calls (or other functions from ``DA.Assert``)
+
+Note that other errors cannot be handled via exceptions, e.g., an
+exercise on an inactive contract will still result in a transaction
+abort.
+
+User-Defined Exceptions
+.......................
+
+Users can define their own exception types which can be thrown and
+caught. The definition looks similar to templates and as for
+templates, the definition will produce a record type of the given name
+as well as instances to make that type throwable and catchable.
+
+In addition to the record fields, exceptions also need to define a
+``message`` function.
+
+.. literalinclude:: ../code-snippets-dev/Exceptions.daml
+   :language: daml
+   :start-after: -- BEGIN_EXCEPTION
+   :end-before: -- END_EXCEPTION
+
+Throwing Exceptions
+...................
+
+There are two ways to throw exceptions:
+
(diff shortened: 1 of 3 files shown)


docs/source/daml/reference/exceptions.rst
```diff
@@ -0,0 +1,89 @@
+.. Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+.. SPDX-License-Identifier: Apache-2.0
+
+.. _exceptions:
+
+Exceptions
+##########
+
+Exceptions are a Daml feature which allows provides a way to handle
```
sofiafaro-da, line 9:
```suggestion
Exceptions are a Daml feature which provides a way to handle
```

docs/source/daml/reference/exceptions.rst
```diff
@@ -0,0 +1,89 @@
...
+certain errors that arise during interpretation instead of aborting
+the transaction and rollback changes that happened before the
```
sofiafaro-da, line 11: The "and rollback changes" phrase sounds like it is still part of the instead clause.
Not sure what the best way to disambiguate this is, but here is a suggestion:

```suggestion
the transaction, and to roll back the state changes that lead to the
```

docs/source/daml/reference/exceptions.rst
```diff
@@ -0,0 +1,89 @@
...
+exercise on an inactive contract will still result in a transaction
+abort.
+
+User-Defined Exceptions
+.......................
+
+Users can define their own exception types which can be thrown and
+caught. The definition looks similar to templates and as for
+templates, the definition will produce a record type of the given name
+as well as instances to make that type throwable and catchable.
```
sofiafaro-da, line 43:
```suggestion
caught. The definition looks similar to templates, and just like with
templates, the definition produces a record type of the given name
as well as instances to make that type throwable and catchable.
```

docs/source/daml/reference/exceptions.rst
```diff
@@ -0,0 +1,89 @@
...
+...................
+
+There are two ways to throw exceptions:
+
+1. Inside of an ``Action`` like ``Update`` or ``Script`` you can use
+   ``throw`` from ``DA.Exception``. This works for any ``Action`` that
+   is an instance of ``ActionThrow``.
+2. Outside of ``ActionThrow`` you can throw exceptions using ``throwPure``.
+
+If both are an option, it is geenrally preferable to use ``throw``
```
sofiafaro-da, line 63:
```suggestion
If both are an option, it is generally preferable to use ``throw``
```

docs/source/daml/reference/exceptions.rst
```diff
@@ -0,0 +1,89 @@
...
+...................
+
+Exceptions are caught in try-catch blocks similar to those found in
+languages like Java. The ``try`` block defines the scope within which
+errors should be handled while the ``catch`` clauses defines which
+types of errors are handled and how the program should continue. If
+one of the catch clauses applies, all writes (creates, consuming
+exercises) in the ``try`` are rolled back. Reads on the other hand
+(e.g., fetches) are still preserved and validated so even in a ``try``
+block a ``fetch`` can never fetch an inactive contract.
```
sofiafaro-da, line 77: I think this paragraph may give the wrong impression. E.g. if you do 

```
   try
       cid <- create T
       fetch cid
       throw E
   catch
       E -> ...
```

It's not true that the "create was rolled back but the fetch still happens", because the fetch here only makes sense in the context of the create.

I'm not sure what's the best way to rephrase this. Maybe something like "All ledger changes since the beginning of the try block are rolled back, but all actions in the try block are still validated. For example, a fetch can never fetch an inactive contract, and a create cannot happen without authorization from all signatories" ... or something.

sofiafaro-da (approved):
Great work! I have some suggestions below.

docs/source/daml/reference/exceptions.rst
cocreature: good points. I rewrote it to make it more explicit. wdyt?

sofiafaro-da: Looks great! Thank you :-)

PR #9808, opened 2021-05-26
`ledger-api-bench-tool` - min consumption speed SLO [DPP-401]

kamil-da:
### Changes
- new service level objective for minimum consumption speed, configurable with `min-consumption-speed` parameter
- redefined definition of the consumption speed metric to be more meaningful

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/ConsumptionSpeedMetric.scala
+++ b/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/ConsumptionSpeedMetric.scala
@@ -11,11 +11,16 @@ import java.time.Instant
 final case class ConsumptionSpeedMetric[T](
     periodMillis: Long,
     recordTimeFunction: T => Seq[Timestamp],
+    objectives: Map[ServiceLevelObjective[ConsumptionSpeedMetric.Value], Option[
+      ConsumptionSpeedMetric.Value
+    ]],
     firstRecordTime: Option[Instant] = None,
     lastRecordTime: Option[Instant] = None,
 ) extends Metric[T] {
+  import ConsumptionSpeedMetric._
 
-  override type Value = ConsumptionSpeedMetric.Value
+  override type V = Value
+  override type Objective = ServiceLevelObjective[Value]
 
   override def onNext(value: T): ConsumptionSpeedMetric[T] = {
     val recordTimes = recordTimeFunction(value)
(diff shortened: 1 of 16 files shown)


ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/ConsumptionSpeedMetric.scala
```diff
@@ -11,57 +11,120 @@ import java.time.Instant
...
+  override type Objective = ServiceLevelObjective[Value]
 
   override def onNext(value: T): ConsumptionSpeedMetric[T] = {
     val recordTimes = recordTimeFunction(value)
-    val updatedFirstRecordTime =
-      firstRecordTime match {
+    val newPreviousLatest =
+      previousLatest match {
         case None =>
           recordTimes.headOption.map { recordTime =>
```
tudor-da, line 30: I didn't check where this is coming from, but is `recordTimes` sorted descending or does it grow like a stack? It's a bit confusing to see the head value named as `..latest`

kamil-da: From what I understand streams served by a ledger have ascending record times.
I've redefined the `ConsumptionSpeecMetric` a bit. Previously we computed a periodic speed like:
`periodicSpeed = (lastRecordTimeInPeriod - firstRecordTimeInPeriod) / periodDuration`. This definition was problematic in case of a single element in a period or no elements at all.
Currently the definition is: `periodicSpeed = (previousLastConsideredElement - lastConsideredElement) / periodDuration`. The `previousLastConsideredElement` may be in the current period (for the first period) or in the most recent previous period with elements.
I'll update the readme file with detailed description of this.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/ConsumptionSpeedMetric.scala
```diff
@@ -4,64 +4,117 @@
...
+        case (objective, value) if value.isDefined => objective -> value.get
+      }
 
   private def periodicSpeed: Option[Double] =
-    (firstRecordTime, lastRecordTime) match {
-      case (Some(first), Some(last)) =>
-        Some((last.toEpochMilli - first.toEpochMilli) * 1.0 / periodMillis)
+    (previousLatest, currentPeriodLatest) match {
+      case (Some(previous), Some(current)) =>
+        Some((current.toEpochMilli - previous.toEpochMilli).toDouble / periodMillis)
```
mziolekda, line 68: There is no guarantee that the periodMillis is being maintained by the thread scheduler. It is safer to assume it is not and rely on a current time, to facilitate the unit testing you may want to supply current time as argument of periodicValue(periodDuration: Duration)

mziolekda (approved):
It looks good. I have made a suggestion not to rely on periodValue being called by the thread scheduler in a timely fashion. I fully realize that such change is big enough to go into another PR. I would be fine with that.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/metrics/ConsumptionSpeedMetric.scala
kamil-da: Good catch! I'll fix this metric and others that may rely on the constant period duration.

PR #9809, opened 2021-05-26
LF: fix parser

remyhaemmerle-da:
fix the parsing of package ID.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/parser/src/main/scala/com/digitalasset/daml/lf/testing/parser/ExprParser.scala
+++ b/daml-lf/parser/src/main/scala/com/digitalasset/daml/lf/testing/parser/ExprParser.scala
@@ -18,16 +18,18 @@ private[parser] class ExprParser[P](parserParameters: ParserParameters[P]) {
   import typeParser._
 
   lazy val expr0: Parser[Expr] =
-    literal ^^ EPrimLit |
+    // expressions starting with fullIdentifier should come before litterals
+    eRecCon |
+      eRecProj |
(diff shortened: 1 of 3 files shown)


daml-lf/parser/src/main/scala/com/digitalasset/daml/lf/testing/parser/ExprParser.scala
```diff
@@ -18,16 +18,18 @@ private[parser] class ExprParser[P](parserParameters: ParserParameters[P]) {
   import typeParser._
 
   lazy val expr0: Parser[Expr] =
-    literal ^^ EPrimLit |
+    // expressions starting with fullIdentifier should come before litterals
+    eRecCon |
```
cocreature, line 22: Was the problem that we parsed it as a literal because that comes first here?

remyhaemmerle-da: Yes. 
PartyId and PackageId have the same syntax.

cocreature: thanks for the explanation!

cocreature (approved):
Thanksk

nickchapman-da (approved):
Thanks for this fix Remy

PR #9815, opened 2021-05-27
Track oldChild in PartialTransaction context

nickchapman-da:
Track in the `PartialTransaction.Context` if we have seen any _old_ children (predating `TxVersion.minException`).
And so avoid potential quadratic blowup of `smallestContainedVersion`.
Also make a small cleanup of the `Context` code, to avoid code duplication when creating a new _empty_ context.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -673,7 +673,8 @@ private[lf] case class PartialTransaction(
         // But we do that in a later normalization phase, not here.
         val rollbackNode = Node.NodeRollback(context.children.toImmArray)
         copy(
-          context = info.parent.addRollbackChild(info.nodeId, context.minChildVersion, context.nextActionChildIdx),
+          context = info.parent
+            .addRollbackChild(info.nodeId, context.minChildVersion, context.nextActionChildIdx),
           nodes = nodes.updated(info.nodeId, rollbackNode),
         ).resetActiveState(info.beginState)
       case _ => throw new RuntimeException("rollbackTry called in non-catch context")
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -70,15 +70,16 @@ private[lf] object PartialTransaction {
     */
   final case class Context(
       info: ContextInfo,
-      minChildVersion: TxVersion,
+      minChildVersion: TxVersion, // tracks the minimum version of any child within `children`
       children: BackStack[NodeId],
       nextActionChildIdx: Int,
   ) {
+    // when we add a child node we must pass the minimum-version contained in that child
     def addActionChild(child: NodeId, version: TxVersion): Context = {
       Context(info, minChildVersion min version, children :+ child, nextActionChildIdx + 1)
     }
-    def addRollbackChild(child: NodeId, nextActionChildIdx: Int): Context =
-      Context(info, minChildVersion, children :+ child, nextActionChildIdx)
+    def addRollbackChild(child: NodeId, version: TxVersion, nextActionChildIdx: Int): Context =
+      Context(info, minChildVersion min version, children :+ child, nextActionChildIdx)
     // This function may be costly, it must be call at most once for each node.
     def nextActionChildSeed: crypto.Hash = info.actionChildSeed(nextActionChildIdx)
   }
@@ -86,6 +87,7 @@ private[lf] object PartialTransaction {
   object Context {
 
     def apply(info: ContextInfo): Context =
+      // An empty context, with no children; minChildVersion is set to the max-int.
       Context(info, TxVersion.VDev, BackStack.empty, 0)
 
     def apply(initialSeeds: InitialSeeding): Context =
@@ -661,6 +663,7 @@ private[lf] case class PartialTransaction(
     * i.e. a exception was thrown inside the context, and the catch associated to the try context did handle it.
     */
   def rollbackTry(ex: SValue.SAny): PartialTransaction = {
+    // we must never create a rollback containing a node with a version pre-dating exceptions
     if (context.minChildVersion < TxVersion.minExceptions) {
       throw DamlEUnhandledException(ex)
     }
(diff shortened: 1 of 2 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -69,27 +69,34 @@ private[lf] object PartialTransaction {
     */
   final case class Context(
       info: ContextInfo,
+      oldChild: Boolean, //one or more children predate exceptions
```
cocreature, line 72: How about tracking `minChildVersion` instead? I tend to prefer moving to a boolean only at the end and keep as much information in between.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -69,27 +69,34 @@ private[lf] object PartialTransaction {
...
       children: BackStack[NodeId],
       nextActionChildIdx: Int,
   ) {
-    def addActionChild(child: NodeId): Context =
-      Context(info, children :+ child, nextActionChildIdx + 1)
+    def addActionChild(child: NodeId, version: TxVersion): Context = {
+      import scala.Ordering.Implicits.infixOrderingOps
+      val oldChild2 = oldChild || (version < TxVersion.minExceptions)
```
cocreature, line 78: I think I’m missing something. Consider the following
1. rollback node
1.1 exercise node version 1.1.dev
1.1.1 create node version 1.1.dev

The exercise node will have the flag set but I don’t see where we propagate it to the rollback node. What am I missing?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: I had it this way to start with, but switched to the bool because it seemed tracking the precise version was more than required, and anyway a rollback node doesn't have a version as such... but you do know if you every created a rollback node then none of its children could be too old.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: Yes, There is a bug if we have an old-create inside a new-exercise inside a rollback.
(I think this is the example you mean)

We definitely need more testcase for these more subtle examples.
(Now we have Rem'y parser fix I will try to write them)

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: But I'll change to tracking the version if you think this is clearer.

remyhaemmerle-da:
part f #8020

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: code fixed now; still need tests.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -69,27 +70,32 @@ private[lf] object PartialTransaction {
...
       nextActionChildIdx: Int,
   ) {
-    def addActionChild(child: NodeId): Context =
-      Context(info, children :+ child, nextActionChildIdx + 1)
+    def addActionChild(child: NodeId, version: TxVersion): Context = {
+      Context(info, minChildVersion min version, children :+ child, nextActionChildIdx + 1)
+    }
     def addRollbackChild(child: NodeId, nextActionChildIdx: Int): Context =
-      Context(info, children :+ child, nextActionChildIdx)
+      Context(info, minChildVersion, children :+ child, nextActionChildIdx)
```
cocreature, line 81: can we propagate the version from the rollback node? I know it’s not needed since we ensure that they are not older but imho it makes the semantics of what `minChildVersion` records much clearer.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -69,27 +70,32 @@ private[lf] object PartialTransaction {
     */
   final case class Context(
       info: ContextInfo,
+      minChildVersion: TxVersion,
```
cocreature, line 73: Some documentation here might be good. I think there are two cases in particular that are important:
1. Childrens of rollback nodes, see my comment below.
2. No children at all -> 1.dev. We could also use an `Option` to make that explicit but I don’t feel strongly here either way so your choice.

cocreature (approved):
Nice! Let’s please add a testcase for the create in exercise case that was broken initially before merging.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: tests added

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: ok

nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: added some comments to help explain what is going on

nickchapman-da:
Think we are done now. All comments are addressed. Adding auto-merge

cocreature:
> Think we are done now. All comments are addressed. Adding auto-merge

:shipit: (you need to format first though :slightly_smiling_face:)

nickchapman-da:
> > Think we are done now. All comments are addressed. Adding auto-merge
> 
> :shipit: (you need to format first though )

grrr, that's annoying. the one time I forget!

remyhaemmerle-da (approved):
Nice.

PR #9816, opened 2021-05-27
Ledger model section on transaction normalization.

sofiafaro-da:
The main reason to bring normalization out as a separate section (rather have it be part of the projection section) is to be able to say that all transactions must be normalized, and that Daml code only produces normalized transactions. This also makes the section on projection much simpler (and projection serves as a practical example for normalization).

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -122,6 +122,84 @@ And the third:
 
 As you can see, in each of these continuities, no contract was consumed twice.
 
+Transaction Normalization
++++++++++++++++++++++++++
+
+The same "before-after" relation can be represented in more than one way using
+rollback nodes. For example, the following three transactions have the same
+"before-after" relation among their ledger actions (A and B):
```
andreaslochbihler-da, line 130: Elsewhere in the ledger model, we use upper-case letters `A` and `B` to represent parties. Actions are typically called `act` with indices.
```suggestion
"before-after" relation among their ledger actions (`act1` and `act2`):
```

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -122,6 +122,84 @@ And the third:
...
+**Transaction normalization** is the process by which equivalent transactions
+are converted into the same transaction. In the case above, all three
+transactions become the transaction in the middle when normalized.
+
+.. https://lucid.app/lucidchart/8a5a09a1-5473-4abf-a72a-57bf03b56794/edit
+.. image:: ./images/exception-normalization-2.svg
+   :align: center
+   :width: 80%
+
+To normalize a transaction, we apply three rules across the whole transaction:
```
andreaslochbihler-da, line 164: Minor: Do you want to walk the reader through the normalization of the diagrams? It's not plain obvious how to get from the left to the middle (and there are actually two ways to get there: use "pull out the start" or "pull out the end" and then remove the empty remaining rollback node.

andreaslochbihler-da (approved):
Looks good. Left a few small comments.

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -122,6 +122,84 @@ And the third:
...
+   :align: center
+   :width: 80%
+
+Because of this, these three transactions are equivalent.
+More generally, two transactions are equivalent if:
+
+- The transactions are the same when you ignore all rollback nodes. That is,
+  if you remove every rollback node and absorb its children into its parent,
+  then two transactions are the same. Equivalently, the transactions have
+  the same ledger actions with the same prefix order and subaction relation.
```
cocreature, line 143: do you mean prefix order or the same preorder traversal?

cocreature (approved):
Nice, thank you!

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: Ah, preorder traversal makes more sense :-)

docs/source/concepts/ledger-model/ledger-exceptions.rst
sofiafaro-da: Good idea. I changed the example a bit to eliminate ambiguity and to showcase both rules 2 and 3.

PR #9817, opened 2021-05-27
Documentation for command deduplication v2

andreaslochbihler-da:
Draft of the new documentation on command deduplication API described in #9567.

The scenarios in `app-arch.rst` are very much "this is how we imagine that this could be used". So these scenarios should merely give an impression of the expected UX around deduplication. Once we have built a PoC of the new command dedup feature, we should migrate some simple application and double-check that this really works. A tutorial on how to use command dedup would probably be better than such a long list of rules.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/app-dev/app-arch.rst
+++ b/docs/source/app-dev/app-arch.rst
@@ -187,7 +187,7 @@ We further assume that the application persists its intended ledger changes and 
 It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
 The application keeps track of the in-flight command submissions via their submission IDs, but need not persist them.
 
-#. After starting up, the application submits a dummy command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
+#. After starting up, the application submits a no-op command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
    This should generate a definite-answer completion with a completion offset ``off_start``.
    The application can use the command service for the submission and extract the completion offset from the successful response or the error details of a rejection.   
    If the submission fails without generating a completion offset, the application should repeat this step until it receives such a fresh completion offset.
@@ -212,8 +212,8 @@ The application keeps track of the in-flight command submissions via their submi
    - ``NOT_FOUND`` for the deduplication start:
      The specified deduplication period is too long.
      The application should find a shorter deduplication time that is still safe.
-     Alternatively, the application can resubmit the change ID with a dummy command instead of the intended ledger change;
-     when it receives a definite answer for the dummy and there are no other submissions with a higher submission rank,
+     Alternatively, the application can resubmit the change ID with a no-op command instead of the intended ledger change;
+     when it receives a definite answer for the no-op command submission and there are no other submissions with a higher submission rank,
      it can conclude that the actual ledger change will not be applied.
 
    - ``OUT_OF_RANGE`` for the deduplication start:
@@ -221,8 +221,9 @@ The application keeps track of the in-flight command submissions via their submi
      Resubmit with an earlier deduplication start.
 
    - ``OUT_OF_RANGE`` for the submission rank:
-     This should happen only when the application uses a completion offset obtained from one participant for a submission on another participant.
+     This should happen only when the application uses a completion offset obtained from one participant (instance) for a submission on another participant.
      The application should not mix completion offsets from different participants.
+     If this happens during fail-over from one participant instance to another, the application should retry.
 
    - ``ALREADY_EXISTS`` with reason ``DEDUPLICATION``:
      The error details contain the offset and submission id of the earlier completion.
--- a/docs/source/app-dev/app-arch.rst
+++ b/docs/source/app-dev/app-arch.rst
@@ -133,22 +133,132 @@ Command deduplication
 
 The interaction of a Daml application with the ledger is inherently asynchronous: applications send commands to the ledger, and some time later they see the effect of that command on the ledger.
 
-There are several things that can fail during this time window: the application can crash, the participant node can crash, messages can be lost on the network, or the ledger may be just slow to respond due to a high load.
+There are several things that can fail during this time window: the application can crash, the participant node can crash, messages can be lost on the network, messages can overtake each other, or the ledger may be just slow to respond due to a high load.
 
-If you want to make sure that a command is not executed twice, your application needs to robustly handle all the various failure scenarios.
-Daml ledgers provide a mechanism for :ref:`command deduplication <command-submission-service-deduplication>` to help deal with this problem.
+If you want to make sure that a command is applied to the ledger at most, your application needs to robustly handle all the failure scenarios.
+Daml ledgers provide a mechanism for :ref:`command deduplication and submission ranking <command-submission-service-deduplication>` to help deal with this problem.
 
-For each command application provide a command ID and an optional parameter that specifies the deduplication time. If the latter parameter is not specified in the command submission itself, the ledger will fall back to using the configured maximum deduplication time.
-The ledger will then guarantee that commands for the same submitting party and command ID will be ignored within the deduplication time window.
+An application can provide for each command the following parameters:
 
-To use command deduplication, you should:
+- A :ref:`command ID <com.daml.ledger.api.v1.Commands.command_id>`.
+  The command ID, the application ID, and the set of submitting parties together form the change ID,
+  which uniquely identifies the intended ledger change.
+- An optional :ref:`submission ID <com.daml.ledger.api.v1.Commands.submission_id>` to uniquely identify the submission among all (re)submissions with the same change ID.
+- An optional :ref:`submission rank <com.daml.ledger.api.v1.Commands.submission_rank>` to rank the submissions with the same change ID.
+- An optional :ref:`deduplication period <com.daml.ledger.api.v1.Commands.deduplication>` during which command submissions with the same change ID are deduplicated.
 
-- Use generous values for the deduplication time. It should be large enough such that you can assume the command was permanently lost if the deduplication time has passed and you still don’t observe any effect of the command on the ledger (i.e. you don't see a transaction with the command ID via the :ref:`transaction service <transaction-service>`).
-- Make sure you set command IDs deterministically, that is to say: the "same" command must use the same command ID. This is useful for the recovery procedure after an application crash/restart, in which the application inspects the state of the ledger (e.g. via the :ref:`Active contracts service <active-contract-service>`) and sends commands to the ledger. When using deterministic command IDs, any commands that had been sent before the application restart will be discarded by the ledger to avoid duplicate submissions.
-- If you are not sure whether a command was submitted successfully, just resubmit it. If the new command was submitted within the deduplication time window, the duplicate submission will safely be ignored. If the deduplication time window has passed, you can assume the command was lost or rejected and a new submission is justified.
+The change ID, submission ID, and the submission rank are included in the completion event to correlate the event to the command submission.
+Their guarantees are described in the :ref:`Ledger API Services <command-submission-service-deduplication>` section.
+We now look at how applications should set these parameters in different scenarios.
 
 
-For more details on command deduplication, see the :ref:`Ledger API Services <command-submission-service-deduplication>` documentation.
+No deduplication needed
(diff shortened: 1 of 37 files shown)


docs/source/app-dev/services.rst
```diff
@@ -68,20 +73,64 @@ For full details, see :ref:`the proto documentation for the service <com.daml.le
...
+#. Applications can specify the start of the :ref:`deduplication period <com.daml.ledger.api.v1.Commands.deduplication>` for each command submission.
+   If this parameter is not set, the participant chooses the start at its discretion.
+   
+#. **Deduplication guarantee:**
+   A command submission is considered a duplicate submission if another command with the same change ID is already in flight or there is an accepting command completion since the start of the deduplication period with the same change ID.
+   Ledgers are free to extend the deduplication period beyond the start of the deduplication period.
+   Commands with rejected completions do not trigger deduplication.
+
+#. A duplicate command submission leads to an error as an RPC response or in a completion with the status codes ``ABORTED`` for in-flight conflicts and ``ALREADY_EXISTS`` for already accepted commands.
+   The error details contain information about the conflicting command submission.
```
meiersi-da, line 94: Do we have a place where we explain the error reporting approach employed by the Ledger API? Should we refer to the docs generated from the gRPC definition of `error.proto`?

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+* The downtime or failover time of the application system, the particpant, and the ledger.
+
+Commands may be duplicated and finality violated if the actual skews and delays exceed the bounds.
+Finality may not be acheiveable if unhealthy conditions persist for longer than the :ref:`max deduplication time <com.daml.ledger.api.v1.LedgerConfiguration.max_deduplication_time>`.
+
+We further assume that the application persists its intended ledger changes and whether it has decided to not retry a change.
+It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
+The application keeps track of the in-flight command submissions via their submission IDs, but need not persist them.
+
+#. After starting up, the application submits a dummy command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
```
meiersi-da, line 190: Consider using 'noop' command instead of 'dummy'.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+     Alternatively, the application can resubmit the change ID with a dummy command instead of the intended ledger change;
+     when it receives a definite answer for the dummy and there are no other submissions with a higher submission rank,
+     it can conclude that the actual ledger change will not be applied.
+
+   - ``OUT_OF_RANGE`` for the deduplication start:
+     The deduplication start lies in the future.
+     Resubmit with an earlier deduplication start.
+
+   - ``OUT_OF_RANGE`` for the submission rank:
+     This should happen only when the application uses a completion offset obtained from one participant for a submission on another participant.
```
meiersi-da, line 224: This can also happen due to the eventual consistency of two participant-instances synchronized via a centralized Committer.

meiersi-da: The application should retry there.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+  
+* The message delay on the network from the application to the participant and the latency for processing a submission.
+
+* The downtime or failover time of the application system, the particpant, and the ledger.
+
+Commands may be duplicated and finality violated if the actual skews and delays exceed the bounds.
+Finality may not be acheiveable if unhealthy conditions persist for longer than the :ref:`max deduplication time <com.daml.ledger.api.v1.LedgerConfiguration.max_deduplication_time>`.
+
+We further assume that the application persists its intended ledger changes and whether it has decided to not retry a change.
+It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
```
meiersi-da, line 187: I cannot easily associate a mental model with this sentence. Would it be possible to give a more concrete example of the kind of watermark that you'd persist? For example, by concretizing the example with the scenario of ingesting from Kafka?

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: Changed.

docs/source/app-dev/services.rst
andreaslochbihler-da: I guess that we will flesh out `error.proto` into a proper `.rst` file and add it to the documentation webpage.

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: Rephrased.

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: That's a good suggestion for a tutorial example. We'll need some hands-on examples with working code.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+* The downtime or failover time of the application system, the particpant, and the ledger.
+
+Commands may be duplicated and finality violated if the actual skews and delays exceed the bounds.
+Finality may not be achieveable if unhealthy conditions persist for longer than the :ref:`max deduplication time <com.daml.ledger.api.v1.LedgerConfiguration.max_deduplication_time>`.
+
+We further assume that the application persists its intended ledger changes and whether it has decided to not retry a change.
+It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
+The application keeps track of the in-flight command submissions via their submission IDs, but need not persist them.
+
+#. After starting up, the application submits a dummy command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
```
cocreature, line 190: I’m slightly sad that this cannot actually be an empty transaction because we reject those. I guess createAndExercise or a nonconsuming choice will have to do.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+   This should generate a definite-answer completion with a completion offset ``off_start``.
+   The application can use the command service for the submission and extract the completion offset from the successful response or the error details of a rejection.
```
cocreature, line 192:
```suggestion
   The application can use the synchronous command service for the submission and extract the completion offset from the successful response or the error details of a rejection.   
```
ime nobody can remember whether the "command service" or the "command submission service" is the synchronous one so spelling it out explicitly is usually helpful imho.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+We further assume that the application persists its intended ledger changes and whether it has decided to not retry a change.
+It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
+The application keeps track of the in-flight command submissions via their submission IDs, but need not persist them.
+
+#. After starting up, the application submits a dummy command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
+   This should generate a definite-answer completion with a completion offset ``off_start``.
+   The application can use the command service for the submission and extract the completion offset from the successful response or the error details of a rejection.   
+   If the submission fails without generating a completion offset, the application should repeat this step until it receives such a fresh completion offset.
+
+#. All subsequent command submission should use a submission rank that is between ``off_start`` and the current completion end.
```
cocreature, line 195: It sounds like you’re just saying that `off_running` is always the latest observed completion offset. Is that correct? If so, I think it might be worth spelling this out a bit more explicit.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+     when it receives a definite answer for the dummy and there are no other submissions with a higher submission rank,
+     it can conclude that the actual ledger change will not be applied.
+
+   - ``OUT_OF_RANGE`` for the deduplication start:
+     The deduplication start lies in the future.
+     Resubmit with an earlier deduplication start.
+
+   - ``OUT_OF_RANGE`` for the submission rank:
+     This should happen only when the application uses a completion offset obtained from one participant for a submission on another participant.
+     The application should not mix completion offsets from different participants.
```
cocreature, line 225: Even in HA setups or are you only referring to different logical participants?

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: Is there a strong reason why we reject empty transactions?

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: Good idea.

docs/source/app-dev/app-arch.rst
```diff
@@ -133,22 +133,132 @@ Command deduplication
...
+It also maintains a watermark of the changes that have either been applied or rejected and will not be retried and the time when it has advanced the watermark.
+The application keeps track of the in-flight command submissions via their submission IDs, but need not persist them.
+
+#. After starting up, the application submits a dummy command with a fresh command ID and submission ID (e.g., a random UUID), but no submission rank or deduplication period.
+   This should generate a definite-answer completion with a completion offset ``off_start``.
+   The application can use the command service for the submission and extract the completion offset from the successful response or the error details of a rejection.   
+   If the submission fails without generating a completion offset, the application should repeat this step until it receives such a fresh completion offset.
+
+#. All subsequent command submission should use a submission rank that is between ``off_start`` and the current completion end.
+   To that end, we assume that the application maintains a running completion offset ``off_running``.
```
andreaslochbihler-da, line 196:
```suggestion
   To that end, we assume that the application maintains a running completion offset ``off_running``, which is always the latest observed completion offset.
```

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: Yep, that's the idea in this scenario. Added a relative clause to explain that.

docs/source/app-dev/app-arch.rst
cocreature: None that I’m aware of. I think @stefanobaghino-da added the restriction to disallow empty commands so maybe he knows.

docs/source/app-dev/app-arch.rst
andreaslochbihler-da: We have various HA setups (VMBC's logical participants, Canton's replicated participant) with different failover stories and transparency of the replication. So we probably can't have rules here that work for every ledger (as some ledgers don't even provide stable offsets across participants). I've added a sentence for the failover scenario.

bame-da:
I started reading writing some detailed comments, but I think we need to address the high-level structure of this first. 

In short: I'm having trouble reading these docs, and that's despite me having reviewed the design docs. I think any advanced or complex feature like this needs two pieces of documentation:

1. A tutorial style guide that leads the reader into the topic, conveys a solid (but not fully detailed) mental model of what's going on, and explains practically how to use the thing.
2. Reference documentation that gives you the full spec.

The current approach mixes the two too much. The docs on the app-arch page don't lead me into the topic or provide me with a mental model of what's happening. Eg it just says "there's a submission rank" without giving me any indication of what that thing is.
It's also too abstract, talking about maintaining watermarks without really saying what that watermark is. Yes, as per the specs I could use a submission rank that is not the latest offset I've observed, but I don't see the value in that precision.

I would suggest doing the following:

In AppArch (probably on its own dedicated page) _explain_ command deduplication. Target that page at a novice. Provide a decent mental model - you even had some pictures in the past. Make the implementation guidance as concrete as possible, maybe using psudo-code. Describe roughly the guarantees you get.

Move all the detail - the precise semantics, the error codes, etc into the Ledger API docs. Make that complete, but keep it as concise and precise as possible. Wherever you leave our detail int he explainer, point into those reference docs.

cocreature:
For reference, here are the issues for the corresponding changes on the JSON API:
1. error handling https://github.com/digital-asset/daml/issues/9834
2. dedup https://github.com/digital-asset/daml/issues/9835

andreaslochbihler-da:
Superseded by #11928.

PR #9821, opened 2021-05-27
KVL-961 Add trace context propagation in CommandClient.trackCommandsUnbounded

hubert-da:
CHANGELOG_BEGIN
 - [Ledger API Client] Introduce implicit telemetry context for tracing
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
@@ -23,15 +23,16 @@ import com.daml.ledger.api.validation.CommandsValidator
 import com.daml.ledger.client.LedgerClient
 import com.daml.ledger.client.configuration.CommandClientConfiguration
 import com.daml.ledger.client.services.commands.CommandTrackerFlow.Materialized
+import com.daml.telemetry.{NoOpTelemetryContext, TelemetryContext}
 import com.daml.util.Ctx
 import com.daml.util.akkastreams.MaxInFlight
 import com.google.protobuf.duration.Duration
 import com.google.protobuf.empty.Empty
 import org.slf4j.{Logger, LoggerFactory}
+import scalaz.syntax.tag._
 
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 import scala.util.Try
-import scalaz.syntax.tag._
 
 /** Enables easy access to command services and high level operations on top of them.
   *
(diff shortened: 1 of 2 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
```diff
@@ -113,7 +120,8 @@ final class CommandClient(
     * @param parties Commands that have a submitting party which is not part of this collection will fail the stream.
     */
   def trackCommandsUnbounded[Context](parties: Seq[String], token: Option[String] = None)(implicit
-      ec: ExecutionContext
+      ec: ExecutionContext,
+      telemetryContext: TelemetryContext,
```
danilofaria, line 124: the way you have here will not work. we don't want to provide the telemetry context at the moment of calling `trackCommandsUnbounded`. We want to have a different one for each `Ctx[Context, SubmitRequest]` passed into the `Flow`, that's why it makes more sense to have it as part of the `Ctx` object

danilofaria: you could change Ctx to be like this
```Scala
final case class Ctx[+Context, +Value](context: Context, value: Value, telemetryContext: TelemetryContext = NoOpTelemetryContext)
```
This won't break existing code and will allow passing the context around. The only missing thing would be to actually propagate that context further at the appropriate places

hubert-da: @danilofaria I dropped another draft [PR](https://github.com/digital-asset/daml/pull/9833) with the changes you suggested. There are a few drawbacks of this approach, though:
- it doesn't break the existing code only partially, at least the pattern matches have to be modified
- I cannot use the `TelemetryContext`, as the Akka Streams' `Outlet` requires the `Ctx` to be serializable. The `TelemetryContext` is not, as it contains the `Tracer`, for example
- I prefer to avoid default arguments, which is minor

Please let me know if we cannot use the changes from here for sure, as each gRPC call will be traced anyway (the client interceptor will create sub-spans).

There is also one more thing to tackle. I did not touch the `CommandCompletionService`. I assume you'd like to trace it, too.

cc @miklos-da @lucb-da

danilofaria: i confirm the changes from this PR are not usable since we need a new context per submission and we want to provide one from the parent context. i will look at your other PR. i dont think think you need to trace the CommandCompletionService since that doesnt touch the context we're passing to the flow

hubert-da: Thanks, rejecting this one then.

PR #9824, opened 2021-05-27
`ledger-api-bench-tool`: TLS support

kamil-da:
### Changes
Added TLS support configurable in the same way as in the `ledger-api-test-tool`.
Tested manually.

```
CHANGELOG_BEGIN
- [Integration Kit] - TLS support for the ledger-api-bench-tool
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/Cli.scala
+++ b/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/Cli.scala
@@ -3,64 +3,73 @@
 
 package com.daml.ledger.api.benchtool
 
+import com.daml.ledger.api.tls.TlsConfigurationCli
 import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
 import com.daml.ledger.api.v1.value.Identifier
-import scopt.{OParser, Read}
+import scopt.{OptionParser, Read}
 
 import scala.concurrent.duration.FiniteDuration
 import scala.util.{Failure, Success, Try}
 
 object Cli {
-  def config(args: Array[String]): Option[Config] =
-    OParser.parse(parser, args, Config.Default)
-
-  private val parser = {
-    val builder = OParser.builder[Config]
+  private val ProgramName: String = "ledger-api-bench-tool"
+  private val parser: OptionParser[Config] = new OptionParser[Config](ProgramName) {
     import Reads._
-    import builder._
-    OParser.sequence(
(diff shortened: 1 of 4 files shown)


ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/Cli.scala
```diff
@@ -58,6 +60,20 @@ object Cli {
         .action { case (size, config) =>
           config.copy(concurrency = config.concurrency.copy(maxPoolSize = size))
         },
+      opt[String]("pem")
```
cocreature, line 63: you might want to use https://github.com/digital-asset/daml/blob/main/ledger-service/cli-opts/src/main/scala/ledger/api/tls/TlsConfigurationCli.scala instead of reimplementing this once more.

kamil-da: Thanks - I wasn't aware of that util.
However, the `TlsConfigurationCli` uses object-oriented scopt DSL and the `ledger-api-bench-tool` uses functional DSL (reference: https://github.com/scopt/scopt). I've implemented functional version in the `TlsConfigurationCli` to be reused in the future.

stefanobaghino-da: I would prefer if we could not have two implementations of the same helper that we need to keep in-sync. Throughout out codebase we are using the object-oriented scopt DSL. I understand it could be not as pleasing but I'd prefer if we could stick to one DSL to avoid having parts of our codebase using one approach and parts the other. If we want to revisit this choice and use the functional DSL we can talk about it, but it should be a codebase-wise effort.

My recommendation is to have a single implementation of this, using the object-oriented scopt DSL, possibly placed in a common directory (perhaps `/libs-scala`).

cocreature: fwiw afaik the main reason we are using the OO DSL is that until fairly recently (beginning of the Scala 2.13) work we were on a version of scopt that only had that. I’m very open to switching.

kamil-da: I've changed the `ledger-api-bench-tool` to use OO-style parser.
Thanks for comments!

PR #9830, opened 2021-05-28
Fix typeclass detection in data-dependencies

### LINKED ISSUE #9689: Preserve constraint synonyms during compilation.
Constraint synonyms are not compiled into LF during LF conversion.

This means that you can't import them in data-dependencies. For example, if we have a data-dependency

```Haskell
module A where
type MyConstraint t = (Show t, Additive t)
```

And its user:

```Haskell
module B where
import A
foo : MyConstraint t => t -> ()
foo = const ()
```

Then the `MyConstraint` is an undefined reference in `B`, because the definition was not preserved. 

IMO, it would be better to compile the constraint synonym into LF, as a type synonym. The only tricky thing is distinguishing these from typeclass definitions, in data-dependencies, since the underlying LF type looks the same. We can probably do that with a `MetadataEncoding` like we already do for functional dependencies, minimal pragmas, and overlap modes.

In addition, we should use these synonyms directly when converting types (instead of expanding them as we currently do), because of #9663.

sofiafaro-da:
Fixes #9689 which was actually caused by misinterpreting a constraint
synonym as an empty typeclass.

This PR fixes and deduplicates the logic for detecting typeclass
definitions in data-dependencies. It also tries to avoid lifting
constraint tuples unnecessarily (since constraint synonym
definitions are already lifted). And for any constraint tuples
that are lifted, it picks a more unique name.

The PR adds a regression test, and a test for nested constraint tuples.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
@@ -613,6 +613,9 @@ getTypeClassFields :: LF.Type -> Maybe [(LF.FieldName, LF.Type)]
 getTypeClassFields = \case
     LF.TStruct fields | all isTypeClassField fields -> Just fields
     LF.TUnit -> Just []
+        -- Type classes with no fields are translated to TUnit
+        -- since LF structs need to have a non-zero number of
+        -- fields.
     _ -> Nothing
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Nice, thank you!

PR #9831, opened 2021-05-28
Collect build-event-stream

aherrmann-da:
Collect Bazel metrics and copy the data to a GCS bucket.

- Collects Bazel metrics and build metadata:
  - [build event stream](https://docs.bazel.build/versions/main/build-event-protocol.html) as JSON dump from the `bazel build` and `bazel test` steps in `build.sh` and `build.ps1`
  - [JSON profile](https://docs.bazel.build/versions/main/skylark/performance.html#performance-profiling) from the `bazel build` and `bazel test` steps in `build.sh` and `build.ps1`
  - Metadata about the current job
- Uploads this data to the GCS bucket `gs://daml-data/bazel-metrics` under a directory of the form `TIMESTAMP-JOBNAME-JOBID-JOBATTEMPT` 
- Installs `jq` in Windows dev-env

The uploaded data ranges between 20M to 50M in size per CI Job.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -65,11 +65,23 @@ jobs:
   dependsOn:
     - da_ghc_lib
     - check_for_release
+    - git_sha
   variables:
-    release_sha: $[ dependencies.check_for_release.outputs['out.release_sha'] ]
-    release_tag: $[ coalesce(dependencies.check_for_release.outputs['out.release_tag'], '0.0.0') ]
-    trigger_sha: $[ dependencies.check_for_release.outputs['out.trigger_sha'] ]
-    is_release: $[ dependencies.check_for_release.outputs['out.is_release'] ]
+    - name: release_sha
+      value: $[ dependencies.check_for_release.outputs['out.release_sha'] ]
+    - name: release_tag
+      value: $[ coalesce(dependencies.check_for_release.outputs['out.release_tag'], '0.0.0') ]
+    - name: trigger_sha
+      value: $[ dependencies.check_for_release.outputs['out.trigger_sha'] ]
+    - name: is_release
+      value: $[ dependencies.check_for_release.outputs['out.is_release'] ]
+    - name: git_branch_sha
+      value: $[ dependencies.git_sha.outputs['out.branch'] ]
+    - name: git_main_sha
+      value: $[ dependencies.git_sha.outputs['out.main'] ]
+    - name: git_fork_point_sha
+      value: $[ dependencies.git_sha.outputs['out.fork_point'] ]
+    - template: job-variables.yml
   timeoutInMinutes: 360
   pool:
     name: 'ubuntu_20_04'
@@ -88,6 +100,7 @@ jobs:
         release_tag: $(release_tag)
         name: 'linux'
         is_release: variables.is_release
+    - template: upload-bazel-metrics.yml
     - bash: |
         set -euo pipefail
         eval "$(./dev-env/bin/dade-assist)"
@@ -114,11 +127,23 @@ jobs:
   dependsOn:
     - da_ghc_lib
     - check_for_release
+    - git_sha
   variables:
-    release_sha: $[ dependencies.check_for_release.outputs['out.release_sha'] ]
-    release_tag: $[ coalesce(dependencies.check_for_release.outputs['out.release_tag'], '0.0.0') ]
-    trigger_sha: $[ dependencies.check_for_release.outputs['out.trigger_sha'] ]
-    is_release: $[ eq(dependencies.check_for_release.outputs['out.is_release'], 'true') ]
+    - name: release_sha
+      value: $[ dependencies.check_for_release.outputs['out.release_sha'] ]
+    - name: release_tag
+      value: $[ coalesce(dependencies.check_for_release.outputs['out.release_tag'], '0.0.0') ]
+    - name: trigger_sha
+      value: $[ dependencies.check_for_release.outputs['out.trigger_sha'] ]
+    - name: is_release
+      value: $[ eq(dependencies.check_for_release.outputs['out.is_release'], 'true') ]
+    - name: git_branch_sha
+      value: $[ dependencies.git_sha.outputs['out.branch'] ]
+    - name: git_main_sha
+      value: $[ dependencies.git_sha.outputs['out.main'] ]
+    - name: git_fork_point_sha
+      value: $[ dependencies.git_sha.outputs['out.fork_point'] ]
+    - template: job-variables.yml
   timeoutInMinutes: 360
   pool:
     name: 'ubuntu_20_04'
@@ -138,6 +163,7 @@ jobs:
         name: 'linux-scala-2.12'
         is_release: variables.is_release
         scala_version: "2.12.13"
+    - template: upload-bazel-metrics.yml
     - template: tell-slack-failed.yml
       parameters:
         trigger_sha: '$(trigger_sha)'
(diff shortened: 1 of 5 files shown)


ci/upload-bazel-metrics.yml
```diff
@@ -0,0 +1,54 @@
...
+        system_job_display_name: "$(System.JobDisplayName)",
+        system_job_id: "$(System.JobId)",
+        system_job_name: "$(System.JobName)",
+        system_pullRequest_pullRequestId: "$(system.pullRequest.pullRequestId)",
+        system_pullRequest_pullRequestNumber: "$(system.pullRequest.pullRequestNumber)",
+        system_pullRequest_mergedAt: "$(system.pullRequest.mergedAt)",
+        system_pullRequest_sourceBranch: "$(system.pullRequest.sourceBranch)",
+        system_pullRequest_targetBranch: "$(system.pullRequest.targetBranch)",
+        system_pullRequest_sourceRepositoryUri: "$(system.pullRequest.sourceRepositoryUri)",
+        system_pullRequest_sourceCommitId: "$(system.pullRequest.sourceCommitId)",
```
garyverhaegen-da, line 40: I'm not sure what's the best approach here, but I should point out this will not work as expected once merged into main. Specifically, the `$()` syntax evaluation at the Azure level only happens if the variables are defined, and the `System.PullRequest.*` ones are not defined on runs that are not PRs (like main runs, or the manual run I just did on this branch). This results in the following JSON object:
```json
{
  "timestamp": "2021-06-07T09:20:08.0000000",
  "id": "2021-06-07T09:20:08.0000000-0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb-1",
  "agent_id": "17473",
  "agent_job_name": "Linux",
  "agent_machine_name": "vsts-agent-ubuntu-20-04-gl00",
  "agent_name": "vsts-agent-ubuntu-20-04-gl00",
  "agent_os": "Linux",
  "agent_os_architecture": "X64",
  "build_build_id": "79607",
  "build_build_number": "20210607.4",
  "build_definition_name": "PRs",
  "build_source_branch": "refs/heads/ci-monitor",
  "build_source_branch_name": "ci-monitor",
  "build_source_version": "8478803b5c3fe6832d8e7c1aa2d16c217866ea98",
  "system_job_attempt": "1",
  "system_job_display_name": "Linux",
  "system_job_id": "0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb",
  "system_job_name": "__default",
  "system_pullRequest_pullRequestId": "$(system.pullRequest.pullRequestId)",
  "system_pullRequest_pullRequestNumber": "$(system.pullRequest.pullRequestNumber)",
  "system_pullRequest_mergedAt": "$(system.pullRequest.mergedAt)",
  "system_pullRequest_sourceBranch": "$(system.pullRequest.sourceBranch)",
  "system_pullRequest_targetBranch": "$(system.pullRequest.targetBranch)",
  "system_pullRequest_sourceRepositoryUri": "$(system.pullRequest.sourceRepositoryUri)",
  "system_pullRequest_sourceCommitId": "$(system.pullRequest.sourceCommitId)",
  "git_branch": "",
  "git_commit": "8478803b5c3fe6832d8e7c1aa2d16c217866ea98",
  "git_tree": "791117d6ca03ae798e72f52721fc8eb35623bf13"
}
```

I _think_ the (non-secret) [Azure variables _are_ available as env vars by default](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#environment-variables), using a simple all-caps, replace-dots-with-underscores naming scheme, so it's worth trying `${SYSTEM_PULLREQUEST_PULLREQUESTID:-}` etc.

aherrmann-da: Thanks! I'll try the env-vars.

ci/upload-bazel-metrics.yml
```diff
@@ -0,0 +1,54 @@
...
+        git_branch: "'"$(git branch --show-current)"'",
+        git_commit: "'"$(git rev-parse HEAD)"'",
+        git_tree: "'"$(git rev-parse HEAD:./)"'",
```
garyverhaegen-da, line 43: I'm unsure about the branch name, but HEAD on a PR build is going to reference the merge commit, which we typically don't keep anywhere. It may be worth also reporting the commits output by the `git_sha` job (i.e. the branch commit and the fork point).

.dadew
```diff
@@ -7,6 +7,7 @@
         "cacert",
         "curl-7.73.0",
         "java-openjdk-8u201",
+        "jq",
```
cocreature, line 10: we have jq as part of msys already. Can we use that instead of installing it twice?

ci/job-variables.yml
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+variables:
```
cocreature, line 4: There seem to be a lot of other shared variables, can we throw those in here as well?

ci/build.yml
```diff
@@ -185,13 +224,26 @@ jobs:
...
+    - name: skip_tests
+      value: $[ and(eq(variables.is_release, 'true'),
                        eq(variables['Build.SourceBranchName'], 'main')) ]
+    - name: git_branch_sha
+      value: $[ dependencies.git_sha.outputs['out.branch'] ]
+    - name: git_main_sha
+      value: $[ dependencies.git_sha.outputs['out.main'] ]
+    - name: git_fork_point_sha
+      value: $[ dependencies.git_sha.outputs['out.fork_point'] ]
+    - template: job-variables.yml
```
cocreature, line 246: How does this work? I would have thought this just gets inlined and via have
```
variables:
  - name: release_sha
    value: …
  - variables:
    - name: pipeline_timestamp
      value: …
```
Are nested variable blocks a thing?

ci/upload-bazel-metrics.yml
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+steps:
+  - template: bash-lib.yml
+    parameters:
+      var_name: bash_lib
+  - bash: |
```
cocreature, line 8: Answering my own question: this takes less than a minute so not an issue for CI times.

cocreature (approved):
Nice work!

.dadew
aherrmann-da: Oh, missed that. I'll try to re-use it.

ci/build.yml
aherrmann-da: If taken that syntax from the docs [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#variable-reuse). As I understand it `template` has slightly different meanings in different contexts in that it's not just a string substitution.

ci/job-variables.yml
aherrmann-da: You mean outside of the variables added by this PR? Yes, I guess we could add those here as well. I'll look into that in a separate PR. Feels a bit out of scope for this one.

cocreature: I was thinking of things like `    - name: git_branch_sha
      value: $[ dependencies.git_sha.outputs['out.branch'] ]` which seem new in a few places

ci/build.yml
cocreature: lovely thanks for the explanation

garyverhaegen-da: > As I understand it `template` has slightly different meanings in different contexts in that it's not just a string substitution.

AFAICT it's always a string substitution; the surprising part about it is _when_ that substitution happens compared to other substitutions (`$()`, `$[[]]`, etc.).

aherrmann-da: @garyverhaegen-da I meant that the contents of `job-variables.yml` are not just pasted in as in
```
variables:
  - name: release_sha
    value: …
  - variables:
    - name: pipeline_timestamp
      value: …
```

But, instead, it expects a `variables` map-item in the template file and inserts these variables as in 
```
variables:
  - name: release_sha
    value: …
  - name: pipeline_timestamp
    value: …
```

garyverhaegen-da: Oh, right. Yes, there are different types of templates. Details [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#insert-a-template); generally speaking, if your template contains multiple jobs, you start it with `jobs:` and you can use it in `jobs:`; if it has multiple steps, you start it with `steps:` and you can use it in `steps:`, etc.

ci/job-variables.yml
aherrmann-da: Oh, I see. Yes, moving these into the template seems to work.

.dadew
aherrmann-da: Done.

PR #9833, opened 2021-05-28
Add trace context propagation in CommandClient.trackCommandsUnbounded [KVL-961]

hubert-da:
### Pull Request Checklist

trigger the build.


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandSubmissionFlow.scala
```diff
@@ -21,16 +23,20 @@ object CommandSubmissionFlow {
...
+        DefaultTelemetry
+          .contextFromMetadata(Some(telemetryMetadata.asJava))
+          .runInOpenTelemetryScope {
+            submit(request)
+              .transform { tryResponse =>
+                Success(
+                  Ctx(
+                    context,
+                    tryResponse,
+                  )
```
danilofaria, line 36:
```suggestion
                  telemetryMetadata)
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandSubmissionFlow.scala
```diff
@@ -21,16 +23,20 @@ object CommandSubmissionFlow {
...
-          .transform { tryResponse =>
-            Success(
-              Ctx(
-                context,
-                tryResponse,
-              )
-            )
-          }(DirectExecutionContext)
+      .mapAsyncUnordered(maxInFlight) { case Ctx(context, request, telemetryMetadata) =>
+        DefaultTelemetry
```
danilofaria, line 27: there is a big issue here that you're forcing the usage of the global tracer. it would be much nicer if the Ctx included the telemetry context object itself with the tracer so i could have my own tracer passed in with the Ctx.
I know you mentioned something about the need for Ctx to be serializable and Tracer not being serializable, but are you sure this is necessary? i dont see anywhere enforcing this

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandSubmissionFlow.scala
hubert-da: good catch

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandSubmissionFlow.scala
hubert-da: > there is a big issue here that you're forcing the usage of the global tracer

It is not an issue for us in other places, however you are right, this is what we discussed on the call a few days ago, the `Telemetry` also has to be provided externally in your case. Will change this.

>I know you mentioned something about the need for Ctx to be serializable and Tracer not being serializable, but are you sure this is necessary? i dont see anywhere enforcing this

Check the Akka Stream's `Outlet`, it's type parameter requires things to be serializable, I cannot use non-serializable objects in the `Flow`.

danilofaria: why are you making a connection between Ctx and akka's Outlet? 
if Ctx should be serializable, how do you guarantee that Ctx.context (which could be of an arbitrary type) is serializable?

hubert-da: I had changed it the way we wanted and it stopped compiling. It was complaining about one of the classes. Let me check once again.

hubert-da: I couldn't reproduce the compilation error today, so I must have done something wrong, sorry. It's good that you've been persuading this, thanks 😉 

Because there is the `TelemetryContext` in `Ctx` you should be able to extend the `DefaultTelemetry` with your tracer. Please let me know if it's fine for you or you'd like me to expose another implementation.

Now, I'm going to add some tests.

hubert-da:
@miklos-da do you think we can make such a change? It at least breaks pattern matches on `Ctx`.

fabiotudone-da (commented):
LGTM but leaving approval to the most directly interested parties.

miklos-da:
> @miklos-da do you think we can make such a change? It at least breaks pattern matches on `Ctx`.

Looks good to me.

miklos-da (approved):

danilofaria (approved):

PR #9838, opened 2021-05-28
Switch to Ubuntu 20.04 nodes for ghc lib

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


azure-pipelines.yml
```diff
@@ -49,7 +49,8 @@ jobs:
                  eq(dependencies.check_for_release.outputs['out.is_release'], 'true'),
                  eq(variables['Build.SourceBranchName'], 'main'))
   pool:
-    vmImage: "Ubuntu-16.04"
+    name: ubuntu_20_04
+    demands: assignment -equals default
```
garyverhaegen-da, line 53: We don't set assignments on Azure-hosted VMs, so it may fail to find any machine to run this.

garyverhaegen-da: 🤦 you're giving up on Azure nodes.

cocreature: this is not an azure hosted vm. I don’t really see a reason why this should run on one so I tried to switch to our own vms at the same time.

garyverhaegen-da (approved):

ci/da-ghc-lib/compile.yml
```diff
@@ -7,11 +7,12 @@ parameters:
 jobs:
 - job: ${{ parameters.final_job_name }}
   pool:
-    vmImage: 'Ubuntu 16.04'
+    name: ubuntu_20_04
+    demands: assignment -equals default
   variables:
-    ghc-lib-sha: '77fcc42eef2256ba9bb61a7e442a65bedb19a35d'
+    ghc-lib-sha: '42e5c306dcfbc84b83336fdd531023e93bfcc5b2'
```
cocreature, line 13: this is just the latest `master`

ci/da-ghc-lib/compile.yml
```diff
@@ -7,11 +7,12 @@ parameters:
...
     base-sha: '9c787d4d24f2b515934c8503ee2bbd7cfac4da20'
-    patches: '323ed5bbd2af0fc8bad7debb1c2ca2a050206bde 833ca63be2ab14871874ccb6974921e8952802e9'
+    patches: '25a18b7a5b3ea4dfadbcd9fd55410c9fbc2b8e28 833ca63be2ab14871874ccb6974921e8952802e9'
```
cocreature, line 15: See https://github.com/digital-asset/ghc/pull/66 will bump once that is approved & merged

azure-pipelines.yml
cocreature: I reverted this part for now. imho we should try to switch those as well but I’d like to spend at least some time going over what we do here exactly and convincing myself that this is not going to fail on the next release and the ghc-lib stuff has turned out to be more complex than expected so doing it separately seems sensible.

ci/da-ghc-lib/compile.yml
```diff
@@ -41,15 +42,20 @@ jobs:
...
                        CI.hs \
                        -- \
                          --da \
                          --merge-base-sha=$(base-sha) \
                          $(for p in $(patches); do echo -n "--patch $p "; done) \
                          --gen-flavor=$(flavor) \
-            | tee log
+            | tee log"
+      export NIX_PATH=nixpkgs=https://github.com/NixOS/nixpkgs/archive/cc855b133100a956d4587079e27c5704cbc8342b.tar.gz
+      nix-shell -p stack -p gmp -p autoconf269 -p automake -p coreutils -p perl -p m4 -p 'haskell.packages.ghc8104.ghcWithPackages(_: [])' 'haskell.packages.ghc865Binary.ghcWithPackages(_: [])' --command "$STACK_CMD"
```
aherrmann-da, line 58: Why do we have three different GHC versions involved here? Maybe worth adding a comment.
```
flavor: 'ghc-8.8.1'
-p 'haskell.packages.ghc8104.ghcWithPackages(_: [])' 'haskell.packages.ghc865Binary.ghcWithPackages(_: [])'
```

cocreature: added a comment. 8.10.4 is what we use to build & run ghc-lib-gen. 8.6.5 is what we use to build our GHC fork based on 8.8.1.

garyverhaegen-da (approved):
Thanks!

ci/da-ghc-lib/compile.yml
aherrmann-da: Thanks!

aherrmann-da (approved):

PR #9839, opened 2021-05-28
Uniform configuration of the command service across all ledgers

mziolekda:
Modify command line options of the ledgers based on the kvutils/app library:
- Remove max-commands-in-flight sub-option from the participant group.
- Add command service configuration options in line with sandbox and daml-on-sql: max-commands-in-flight, max-parallel-submissions, input-buffer-size

### DIFF
--- a/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
+++ b/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
@@ -26,6 +26,7 @@ final case class Config[Extra](
     mode: Mode,
     ledgerId: String,
     archiveFiles: Seq[Path],
+    commandConfig: CommandConfiguration,
     tlsConfig: Option[TlsConfiguration],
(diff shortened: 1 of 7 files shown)


ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -333,13 +327,40 @@ object Config {
...
+          )
+
+        opt[Int]("input-buffer-size")
+          .optional()
+          .action((value, config) =>
+            config.copy(commandConfig = config.commandConfig.copy(inputBufferSize = value))
+          )
+          .text(
+            "The maximum number of commands waiting to be submitted for each party. Overflowing this threshold will cause back-pressure, signaled by a RESOURCE_EXHAUSTED error code. Default is 512."
+          )
```
stefanobaghino-da, line 355: If this is defined somewhere else, it would probably be wise to factor it out in a common shared packages (`//libs-scala` probably being a good place it).

stefanobaghino-da (approved):

PR #9847, opened 2021-05-31
Add per-query offsets to HTTP-JSON API

stefanobaghino-da:
changelog_begin
[HTTP/JSON API] The streaming query endpoint now accepts to override the
offset for each query using the `query_offset` field. You can read more
about it on the documentation.
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/json-api/index.rst
+++ b/docs/source/json-api/index.rst
@@ -1301,7 +1301,7 @@ Moreover, an ``offset`` may be specified alongside the query itself::
         {"templateIds": ["Iou:Iou"], "offset": "5609"}
     ]
 
-Per-query ``offset``s take precedence over the ``offset`` message. The
+Per-query ``offset`` s take precedence over the ``offset`` message. The
 ``offset`` message is still applied to queries that don't specify one.
 
 The output is a series of JSON documents, each ``payload`` formatted
@@ -1345,8 +1345,24 @@ off an initial "loading" indicator::
         "offset": "2"
     }
 
-Events in the following "live" data may include ``events`` that precede
-this ``offset`` if an earlier per-query ``offset`` was specified.
+.. note::
+
+    Events in the following "live" data may include ``events`` that precede
+    this ``offset`` if an earlier per-query ``offset`` was specified.
+
+    This has been done with the intent of allowing to use per-query ``offset`` s to
+    efficiently use a single connection to multiplex various requests. To give an
+    example of how this would work, let's say that there are two contract templates,
+    ``A`` and ``B`` . Your application first queries for ``A`` s without specifying
+    an offset. Then some client-side interaction requires the application to do the
+    same for ``B`` s. The application can save the latest observed offset for the
+    previous query, which let's say is ``42``, and issue a new request that queries
+    for all ``B`` s without specifying an offset and all ``A`` s from ``42``. While
+    this happens on the client, a few more ``A`` s and ``B`` s are created and the
+    new request is issued once the latest offset is ``47``. The response to this
+    will contain a message with all active ``B`` s, followed by the message reporting
+    the offset ``47``, followed by a stream of live updates that contains new ``A`` s
+    starting from ``42`` and new ``B`` s starting from ``47`` .
 
 To keep the stream alive, you'll occasionally see messages like this,
 which can be safely ignored if you do not need to capture the last seen ledger offset::
--- a/docs/source/json-api/index.rst
+++ b/docs/source/json-api/index.rst
@@ -1284,12 +1284,25 @@ different sets of template IDs::
         {"templateIds": ["Iou:Iou"]}
     ]
 
+Queries have two ways to specify an offset.
+
 An optional ``offset`` returned by a prior query (see output examples
 below) may be specified *before* the above, as a separate body.  It must
 be a string, and if specified, the stream will begin immediately *after*
 the response body that included that offset::
 
     {"offset": "5609"}
+
+Moreover, a ``queryOffset`` may be specified alongside the query itself::
+
+    [
+        {"templateIds": ["Iou:Iou"], "query": {"amount": {"%lte": 50}}},
+        {"templateIds": ["Iou:Iou"], "query": {"amount": {"%gt": 50}}},
+        {"templateIds": ["Iou:Iou"], "queryOffset": "5609"}
+    ]
+
+If both an ``offset`` and a ``queryOffset`` are specified, ``queryOffset``
+takes precedence.
 
 The output is a series of JSON documents, each ``payload`` formatted
 according to :doc:`lf-value-specification`::
@@ -1331,6 +1344,9 @@ off an initial "loading" indicator::
         "events": [],
         "offset": "2"
     }
+
+Events in the following "live" data may include ``events`` that precede
+this ``offset`` if an earlier ``queryOffset`` was specified.
 
 To keep the stream alive, you'll occasionally see messages like this,
 which can be safely ignored if you do not need to capture the last seen ledger offset::
--- a/docs/source/json-api/index.rst
+++ b/docs/source/json-api/index.rst
@@ -1293,16 +1293,16 @@ the response body that included that offset::
 
     {"offset": "5609"}
 
-Moreover, a ``queryOffset`` may be specified alongside the query itself::
+Moreover, an ``offset`` may be specified alongside the query itself::
 
     [
         {"templateIds": ["Iou:Iou"], "query": {"amount": {"%lte": 50}}},
         {"templateIds": ["Iou:Iou"], "query": {"amount": {"%gt": 50}}},
-        {"templateIds": ["Iou:Iou"], "queryOffset": "5609"}
+        {"templateIds": ["Iou:Iou"], "offset": "5609"}
     ]
 
-If both an ``offset`` and a ``queryOffset`` are specified, ``queryOffset``
-takes precedence.
+Per-query ``offset``s take precedence over the ``offset`` message. The
+``offset`` message is still applied to queries that don't specify one.
 
 The output is a series of JSON documents, each ``payload`` formatted
 according to :doc:`lf-value-specification`::
@@ -1346,7 +1346,7 @@ off an initial "loading" indicator::
     }
 
 Events in the following "live" data may include ``events`` that precede
-this ``offset`` if an earlier ``queryOffset`` was specified.
+this ``offset`` if an earlier per-query ``offset`` was specified.
 
 To keep the stream alive, you'll occasionally see messages like this,
 which can be safely ignored if you do not need to capture the last seen ledger offset::
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -747,6 +747,28 @@ abstract class AbstractWebsocketServiceIntegrationTest
     go(createCount)
   }
 
+  private[this] def readAcsAt(until: domain.Offset): Consume.FCC[JsValue, Map[String, JsValue]] = {
+    val dslSyntax = Consume.syntax[JsValue]
+    import dslSyntax._
+    def go(
+        acs: Map[String, JsValue],
+        latest: Option[domain.Offset],
+    ): Consume.FCC[JsValue, Map[String, JsValue]] =
+      if (latest.fold(false)(domain.Offset.ordering.greaterThanOrEqual(_, until))) {
+        point(acs)
+      } else {
+        for {
+          ContractDelta(creates, deletes, offset) <- readOne
+          _ = println("#" * 80)
+          _ = println(offset)
+          _ = println(creates)
+          _ = println(deletes)
+          next <- go(acs ++ creates -- deletes.map(_.contractId.unwrap), offset)
+        } yield next
+      }
+    go(acs = Map.empty, latest = None)
+  }
+
   "fetch should should return an error if empty list of (templateId, key) pairs is passed" in withHttpService {
     (uri, _, _) =>
       singleClientFetchStream(jwt, uri, "[]")
@@ -914,6 +936,111 @@ abstract class AbstractWebsocketServiceIntegrationTest
       } yield inside(result) { case ShouldHaveEnded(_, 2, _) =>
         succeed
       }
+  }
+
+  "Per-query offsets should override offsets issued on a per-query basis" in withHttpService {
+    (uri, _, _) =>
+      val dslSyntax = Consume.syntax[JsValue]
+      import dslSyntax._
+      import spray.json._
+      def createIouCommand(currency: String): String =
+        s"""{
+           |  "templateId": "Iou:Iou",
+           |  "payload": {
+           |    "observers": [],
+           |    "issuer": "Alice",
+           |    "amount": "999.99",
+           |    "currency": "$currency",
+           |    "owner": "Alice"
+           |  }
+           |}""".stripMargin
+      def createIou(currency: String): Future[Assertion] =
+        HttpServiceTestFixture
+          .postJsonStringRequest(
+            uri.withPath(Uri.Path("/v1/create")),
+            createIouCommand(currency),
+            headersWithAuth,
+          )
+          .map(_._1 shouldBe a[StatusCodes.Success])
+      def contractsQuery(currency: String): String =
+        s"""{"templateIds":["Iou:Iou"], "query":{"currency":"$currency"}}"""
+      def contractsQueryWithOffset(offset: domain.Offset, currency: String): String =
+        s"""{"templateIds":["Iou:Iou"], "query":{"currency":"$currency"}, "query_offset":"${offset.unwrap}"}"""
+      def contracts(currency: String, offset: Option[domain.Offset]): String =
+        offset.fold(contractsQuery(currency))(contractsQueryWithOffset(_, currency))
+      def ledgerEnd(expectedContracts: Int): Future[domain.Offset] = {
+        def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[domain.Offset]] =
+          Consume.interpret(
+            for {
+              _ <- readAcsN(expectedContracts)
+              ContractDelta(Vector(), _, Some(offset)) <- readOne
+              _ = killSwitch.shutdown()
+              _ <- drain
+            } yield offset
+          )
+        val (killSwitch, source) =
+          singleClientQueryStream(jwt, uri, """{"templateIds":["Iou:Iou"]}""")
+            .viaMat(KillSwitches.single)(Keep.right)
+            .preMaterialize()
+        source.via(parseResp).runWith(go(killSwitch))
+      }
+      def test(
+          end: domain.Offset,
+          eurFrom: Option[domain.Offset],
+          usdFrom: Option[domain.Offset],
+          expected: Map[String, Int],
+      ): Future[Assertion] = {
+        def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[Assertion]] = {
+          Consume.interpret(
+            for {
+              contracts <- readAcsAt(end)
+              result = contracts
+                .map(_._2.asJsObject.fields("currency").asInstanceOf[JsString].value)
+                .groupBy(identity)
+                .view
+                .mapValues(_.size)
+                .toMap
+              ContractDelta(Vector(), _, _) <- readOne
+              _ = println("*" * 80)
+              _ = println(contracts)
+              _ = println(result)
+              _ = println("*" * 80)
+              _ = killSwitch.shutdown()
+              _ <- drain
+            } yield result shouldEqual expected
+          )
+        }
+        val (killSwitch, source) =
+          singleClientQueryStream(
+            jwt,
+            uri,
+            Seq(contracts("EUR", eurFrom), contracts("USD", usdFrom)).mkString("[", ",", "]"),
+          )
+            .viaMat(KillSwitches.single)(Keep.right)
+            .preMaterialize()
+        source.via(parseResp).runWith(go(killSwitch))
+      }
+
+      for {
+        _ <- createIou("EUR")
+        _ <- createIou("USD")
+        offset1 <- ledgerEnd(2)
+        _ <- createIou("EUR")
+        _ <- createIou("USD")
+        offset2 <- ledgerEnd(4)
+        _ <- createIou("EUR")
+        _ <- createIou("USD")
+        offset3 <- ledgerEnd(6)
+        result <- test(
+          end = offset3,
+          eurFrom = Some(offset1),
+          usdFrom = Some(offset2),
+          expected = Map(
+            "EUR" -> 2,
+            "USD" -> 1,
+          ),
+        )
+      } yield result
   }
 
   "ContractKeyStreamRequest" - {
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -1088,7 +1088,7 @@ abstract class AbstractWebsocketServiceIntegrationTest
         usdFrom = Some(offset2),
         expected = Map(
           "EUR" -> 3,
-          "USD" -> 1,
+          "USD" -> 3, // FIXME This should be 1
         ),
       )
     } yield succeed
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -1032,6 +1032,28 @@ abstract class AbstractWebsocketServiceIntegrationTest
       _ <- createIou("USD")
       offset3 <- ledgerEnd(6)
       _ <- test(
+        clue = "No offsets",
+        end = offset3,
+        queryFrom = None,
+        eurFrom = None,
+        usdFrom = None,
+        expected = Map(
+          "EUR" -> 3,
+          "USD" -> 3,
+        ),
+      )
+      _ <- test(
+        clue = "Offset message only",
+        end = offset3,
+        queryFrom = Some(offset2),
+        eurFrom = None,
+        usdFrom = None,
+        expected = Map(
+          "EUR" -> 1,
+          "USD" -> 1,
+        ),
+      )
+      _ <- test(
         clue = "Per-query offsets only",
         end = offset3,
         queryFrom = None,
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -747,7 +747,9 @@ abstract class AbstractWebsocketServiceIntegrationTest
     go(createCount)
   }
 
-  private[this] def readAcsAt(until: domain.Offset): Consume.FCC[JsValue, Map[String, JsValue]] = {
+  private[this] def activeContractsAt(
+      until: domain.Offset
+  ): Consume.FCC[JsValue, Map[String, JsValue]] = {
     val dslSyntax = Consume.syntax[JsValue]
     import dslSyntax._
     def go(
@@ -759,10 +761,6 @@ abstract class AbstractWebsocketServiceIntegrationTest
       } else {
         for {
           ContractDelta(creates, deletes, offset) <- readOne
-          _ = println("#" * 80)
-          _ = println(offset)
-          _ = println(creates)
-          _ = println(deletes)
           next <- go(acs ++ creates -- deletes.map(_.contractId.unwrap), offset)
         } yield next
       }
@@ -994,16 +992,12 @@ abstract class AbstractWebsocketServiceIntegrationTest
       def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[Assertion]] = {
         Consume.interpret(
           for {
-            contracts <- readAcsAt(end)
+            contracts <- activeContractsAt(end)
             result = contracts
               .map(_._2.asJsObject.fields("currency").asInstanceOf[JsString].value)
               .groupBy(identity)
               .map { case (k, vs) => k -> vs.size }
             ContractDelta(Vector(), _, _) <- readOne
-            _ = println("*" * 80)
-            _ = println(contracts)
-            _ = println(result)
-            _ = println("*" * 80)
             _ = killSwitch.shutdown()
             _ <- drain
           } yield withClue(clue) { result shouldEqual expected }
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -752,11 +752,13 @@ abstract class AbstractWebsocketServiceIntegrationTest
   ): Consume.FCC[JsValue, Map[String, JsValue]] = {
     val dslSyntax = Consume.syntax[JsValue]
     import dslSyntax._
+    import domain.Offset.ordering
+    import scalaz.syntax.order._
     def go(
         acs: Map[String, JsValue],
         latest: Option[domain.Offset],
     ): Consume.FCC[JsValue, Map[String, JsValue]] =
-      if (latest.fold(false)(domain.Offset.ordering.greaterThanOrEqual(_, until))) {
+      if (latest.fold(false)(_ >= until)) {
         point(acs)
       } else {
         for {
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -986,6 +986,7 @@ abstract class AbstractWebsocketServiceIntegrationTest
       }
       def test(
           end: domain.Offset,
+          queryFrom: Option[domain.Offset],
           eurFrom: Option[domain.Offset],
           usdFrom: Option[domain.Offset],
           expected: Map[String, Int],
@@ -1013,6 +1014,7 @@ abstract class AbstractWebsocketServiceIntegrationTest
             jwt,
             uri,
             Seq(contracts("EUR", eurFrom), contracts("USD", usdFrom)).mkString("[", ",", "]"),
+            queryFrom,
           )
             .viaMat(KillSwitches.single)(Keep.right)
             .preMaterialize()
@@ -1029,8 +1031,9 @@ abstract class AbstractWebsocketServiceIntegrationTest
         _ <- createIou("EUR")
         _ <- createIou("USD")
         offset3 <- ledgerEnd(6)
-        result <- test(
+        _ <- test(
           end = offset3,
+          queryFrom = None,
           eurFrom = Some(offset1),
           usdFrom = Some(offset2),
           expected = Map(
@@ -1038,7 +1041,29 @@ abstract class AbstractWebsocketServiceIntegrationTest
             "USD" -> 1,
           ),
         )
-      } yield result
+        // FIXME Fails, for some reason the ACS is fetched but it shouldn't, investigating
+//        _ <- test(
+//          end = offset3,
+//          queryFrom = Some(offset2),
+//          eurFrom = None,
+//          usdFrom = Some(offset1),
+//          expected = Map(
+//            "EUR" -> 1,
+//            "USD" -> 2,
+//          ),
+//        )
+        // FIXME Fails, the ACS blocks must be filtered
+//        _ <- test(
+//          end = offset3,
+//          queryFrom = None,
+//          eurFrom = None,
+//          usdFrom = Some(offset2),
+//          expected = Map(
+//            "EUR" -> 3,
+//            "USD" -> 1,
+//          ),
+//        )
+      } yield succeed
   }
 
   "ContractKeyStreamRequest" - {
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -1088,7 +1088,7 @@ abstract class AbstractWebsocketServiceIntegrationTest
         usdFrom = Some(offset2),
         expected = Map(
           "EUR" -> 3,
-          "USD" -> 3, // FIXME This should be 1
+          "USD" -> 1,
         ),
       )
     } yield succeed
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -997,9 +997,7 @@ abstract class AbstractWebsocketServiceIntegrationTest
               result = contracts
                 .map(_._2.asJsObject.fields("currency").asInstanceOf[JsString].value)
                 .groupBy(identity)
-                .view
-                .mapValues(_.size)
-                .toMap
+                .map { case (k, vs) => k -> vs.size }
               ContractDelta(Vector(), _, _) <- readOne
               _ = println("*" * 80)
               _ = println(contracts)
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -1041,17 +1041,16 @@ abstract class AbstractWebsocketServiceIntegrationTest
             "USD" -> 1,
           ),
         )
-        // FIXME Fails, for some reason the ACS is fetched but it shouldn't, investigating
-//        _ <- test(
-//          end = offset3,
-//          queryFrom = Some(offset2),
-//          eurFrom = None,
-//          usdFrom = Some(offset1),
-//          expected = Map(
-//            "EUR" -> 1,
-//            "USD" -> 2,
-//          ),
-//        )
+        _ <- test(
+          end = offset3,
+          queryFrom = Some(offset2),
+          eurFrom = None,
+          usdFrom = Some(offset1),
+          expected = Map(
+            "EUR" -> 1,
+            "USD" -> 2,
+          ),
+        )
         // FIXME Fails, the ACS blocks must be filtered
 //        _ <- test(
 //          end = offset3,
--- a/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
+++ b/ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
@@ -938,13 +938,12 @@ abstract class AbstractWebsocketServiceIntegrationTest
       }
   }
 
-  "Per-query offsets should override offsets issued on a per-query basis" in withHttpService {
-    (uri, _, _) =>
-      val dslSyntax = Consume.syntax[JsValue]
-      import dslSyntax._
-      import spray.json._
-      def createIouCommand(currency: String): String =
-        s"""{
+  "Per-query offsets should work as expected" in withHttpService { (uri, _, _) =>
+    val dslSyntax = Consume.syntax[JsValue]
+    import dslSyntax._
+    import spray.json._
+    def createIouCommand(currency: String): String =
+      s"""{
            |  "templateId": "Iou:Iou",
            |  "payload": {
            |    "observers": [],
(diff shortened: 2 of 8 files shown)


ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +747,28 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+    ): Consume.FCC[JsValue, Map[String, JsValue]] =
+      if (latest.fold(false)(domain.Offset.ordering.greaterThanOrEqual(_, until))) {
+        point(acs)
+      } else {
+        for {
+          ContractDelta(creates, deletes, offset) <- readOne
+          _ = println("#" * 80)
+          _ = println(offset)
+          _ = println(creates)
+          _ = println(deletes)
```
stefanobaghino-da, line 765: Debug prints, will be removed before moving this out of draft.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -916,6 +938,111 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+                .map(_._2.asJsObject.fields("currency").asInstanceOf[JsString].value)
+                .groupBy(identity)
+                .view
+                .mapValues(_.size)
+                .toMap
+              ContractDelta(Vector(), _, _) <- readOne
+              _ = println("*" * 80)
+              _ = println(contracts)
+              _ = println(result)
+              _ = println("*" * 80)
```
stefanobaghino-da, line 1007: Debug prints, will be removed before moving this out of draft.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -250,6 +263,18 @@ object WebSocketService {
...
+      override def requestOffset(request: SearchForeverRequest): Option[StartingOffset] =
+        request.queries.toList.view
+          .map(_.offset)
+          .fold(none[domain.Offset]) {
+            case (Some(o1), Some(o2)) => Some(domain.Offset.ordering.min(o1, o2))
+            case (o1, None) if o1.isDefined => o1
+            case (None, o2) if o2.isDefined => o2
+            case _ => None
+          }
+          .map(StartingOffset.apply)
```
stefanobaghino-da, line 276: I suspect there is a more principled way to do this, looking forward to suggestions.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -481,6 +511,16 @@ class WebSocketService(
...
 
+    val requestOffset = Q.requestOffset(request)
+
+    val startingOffset =
+      (offPrefix, requestOffset) match {
+        case (Some(a), Some(b)) => Some(domain.StartingOffset.ordering.min(a, b))
+        case (a, None) if a.isDefined => a
+        case (None, b) if b.isDefined => b
+        case _ => None
+      }
```
stefanobaghino-da, line 522: Some duplication with the `requestOffset` method implementation itself. I'd like to improve how the comparison is done and factor it out. Suggestions?

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
```diff
@@ -238,33 +239,56 @@ object JsonProtocol extends DefaultJsonProtocol with ExtraFormats {
...
     }
   }
 
+  implicit val GetActiveContractsRequestFormat: RootJsonReader[domain.GetActiveContractsRequest] =
+    requestJsonReader(validExtraFields = Set.empty)((tids, query, _) =>
+      domain.GetActiveContractsRequest(tids, query)
+    )
+
+  implicit val SearchForeverQueryFormat: RootJsonReader[domain.SearchForeverQuery] = {
+    val OffsetKey = "query_offset"
```
stefanobaghino-da, line 281: Initially this was just `offset` but clashing with the generic `offset` message was difficult to manage. If there are suggestions on how to cleanly manage it I'm happy to change this back to just `offset`.

stefanobaghino-da (commented):
On top of what I put in my early self-review above, the following points will be addressed before moving this out of draft:
- [ ] More tests, in particular checking different combinations of offsets in and out of each query
- [ ] Documentation

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/ae9d8f7d2dda39f946f5bdac6dc9eb1f0b28b02b

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/ae9d8f7d2dda39f946f5bdac6dc9eb1f0b28b02b

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/6a832849983f8d68bdbc48b8ede4b12bd84f16be

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/6a832849983f8d68bdbc48b8ede4b12bd84f16be

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -519,7 +612,7 @@ class WebSocketService(
                 shiftedPrefix,
                 Terminates.Never,
               )
-              .via(convertFilterContracts(fn))
+              .via(convertFilterContracts(txsPred.fn))
```
S11001001, line 615: Two issues I see here, though I cannot quite say what is the best way to fix them:

1. `shiftedPrefix` is determined by `prefilter`, and if an ACS update was made, it is the offset that the ACS update reached (hence, "shifted" from the original offset passed into the function). It comes from `fetchAndPersist`.  As you see from the following, an ACS-and-txns stream is the concatenation of the prefiltered (ACS + live-begin) and liveFiltered (txn) parts. So, when an ACS update occurs, you are *starting* the live-filtered portion at whatever offset the ACS update ended at, which will likely be *after* whatever offsets were passed in.
2. If a query has an ACS portion, its function-predicate is *still* required during the liveFiltered portion, but your partition in `splitRequest` excludes the function-predicates for any query with an ACS portion from the predicate passed on line 615 to `convertFilterContracts`, *unless* there happen to be zero ACS-less queries.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -250,6 +275,43 @@ object WebSocketService {
...
+      private implicit val stringOrder = scalaz.Order.fromScalaOrdering[String]
+
+      override def startingOffset(
+          prefix: Option[domain.StartingOffset],
+          request: SearchForeverRequest,
+      ): Option[domain.StartingOffset] =
+        request.queries
+          .map(_.offset)
+          .minimumBy(_.fold("")(_.unwrap))
+          .get // Safe on NonEmptyList
```
S11001001, line 313: Pull in `foldable1` syntax and use `minimumBy1` instead.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +747,26 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+  private[this] def activeContractsAt(
+      until: domain.Offset
+  ): Consume.FCC[JsValue, Map[String, JsValue]] = {
+    val dslSyntax = Consume.syntax[JsValue]
+    import dslSyntax._
+    def go(
+        acs: Map[String, JsValue],
+        latest: Option[domain.Offset],
+    ): Consume.FCC[JsValue, Map[String, JsValue]] =
+      if (latest.fold(false)(domain.Offset.ordering.greaterThanOrEqual(_, until))) {
```
S11001001, line 759: `scalaz.syntax.order` supplies `>=`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -171,6 +172,20 @@ object WebSocketService {
...
 
     def renderCreatedMetadata(p: Positive): Map[String, JsValue]
+
+    /** The first item of the tuple is the request to be issued to the
+      * ActiveContractsService, the second to the TransactionService
+      */
+    def splitRequest(
+        prefix: Option[domain.StartingOffset],
+        request: A,
+    ): (Option[A], Option[A])
```
S11001001, line 182: If you think it would behoove your implementation to have *different* types for the first and second `A`s here, then by all means, add a type member (just as I did with `Positive`) to represent one or the other.

Additionally, I think you want `\&/` here, because it appears that `(None, None)` is None-sense.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -198,17 +213,27 @@ object WebSocketService {
...
             (resolved, unresolved, q transform ((_, p) => NonEmptyList((p, ix))))
           }
-        def fn(a: domain.ActiveContract[LfV]): Option[Positive] =
+        def fn(a: domain.ActiveContract[LfV], o: Option[domain.Offset]): Option[Positive] = {
           q.get(a.templateId).flatMap { preds =>
-            preds.collect(Function unlift { case ((_, p), ix) => p(a.payload) option ix })
+            preds.collect(Function unlift { case ((_, p), ix) =>
+              val matchesPredicate = p(a.payload)
+              val matchesOffset = indexedOffsets(ix).flatMap(q => o.map((_, q))).fold(true) {
+                case (o, q) => domain.Offset.ordering.greaterThan(o, q)
```
S11001001, line 230: Same with `>`

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/7670eadebd735de629897ad66bdada2fb876d455

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/72a613ae5738e8e41d89462c79c75b9f92642236

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: > If you think it would behoove your implementation to have different types for the first and second As here, then by all means, add a type member (just as I did with Positive) to represent one or the other.

Yeah, I was a bit unsure of how to manage different types for the `query` and `fetch` endpoints, probably the type parameter is the way to go, I'll try this today, thanks.

> Additionally, I think you want \&/ here, because it appears that (None, None) is None-sense.

Thanks, I'll have a look at this if I end up needing it.

stefanobaghino-da: Working on a different approach, reverted this one with https://github.com/digital-asset/daml/pull/9847/commits/0a3f85f25b2b85b9b8112c96bf8b9d27f8de0fe6

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -402,6 +403,20 @@ class ContractsService(
...
+      parties: OneAnd[Set, lar.Party],
+      templateIds: List[domain.TemplateId.RequiredPkg],
+  ): Source[ContractStreamStep.LAV1, NotUsed] = {
+    val txnFilter = util.Transactions.transactionFilterFor(parties, templateIds)
+    getActiveContracts(jwt, txnFilter, true)
+      .map { case GetActiveContractsResponse(offset, _, activeContracts, _) =>
+        if (activeContracts.nonEmpty) Acs(activeContracts.toVector)
+        else LiveBegin(AbsoluteBookmark(domain.Offset(offset)))
+      }
+      .wireTap(step => println(s",,, ACS $step"))
```
stefanobaghino-da, line 417: Remove debug print.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -431,10 +446,12 @@ class ContractsService(
...
-    contractsAndBoundary mapMaterializedValue { fob =>
-      fob.foreach(a => logger.debug(s"contracts fetch completed at: ${a.toString}"))
-      NotUsed
-    }
+    contractsAndBoundary
+      .mapMaterializedValue { fob =>
+        fob.foreach(a => logger.debug(s"contracts fetch completed at: ${a.toString}"))
+        NotUsed
+      }
+      .wireTap(step => println(s"^^^ TXS $step"))
```
stefanobaghino-da, line 454: Remove debug print.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -476,69 +555,131 @@ class WebSocketService(
...
+          mdContracts <- predicate.dbQuery(parties, dao)
+        } yield {
+          val acs =
+            if (mdContracts.nonEmpty) {
+              Source.single(StepAndErrors(Seq.empty, ContractStreamStep.Acs(mdContracts)))
+            } else {
+              Source.empty
+            }
+          val liveMarker = liveBegin(bookmark.map(_.toDomain))
+          (acs ++ liveMarker).wireTap(step => println(s"--- DBS $step"))
```
stefanobaghino-da, line 578: Remove debug print.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -476,69 +555,131 @@ class WebSocketService(
...
+            )
+        })
         .mapMaterializedValue { _: Future[NotUsed] =>
           NotUsed
         }
         .via(removePhantomArchives(remove = Q.removePhantomArchives(request)))
         .map(_.mapPos(Q.renderCreatedMetadata).render)
         .prepend(reportUnresolvedTemplateIds(unresolved))
         .map(jsv => \/-(wsMessage(jsv)))
+        .wireTap(msg => println(s"%%%%%% $msg"))
```
stefanobaghino-da, line 682: Remove debug print.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/3c9e0f2e4e7e7e68619ab187e0fe7c97432fa427

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/96e5cb5132f5c6459bbec6c198cdbec6f1f177a1

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/987353abb9949a224fe8871d7217d9ab00641bf7

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/e0901285a6549a0d2a381dfb2d6358df118a402c

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
```diff
@@ -427,7 +427,7 @@ private[http] object ContractsFetch {
   /** Split a series of ACS responses into two channels: one with contracts, the
     * other with a single result, the last offset.
     */
-  private[this] def acsAndBoundary
+  private[http] def acsAndBoundary
```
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/ec980dfc09f227e4e179a62bcf173a63e58f5eee

stefanobaghino-da (commented):
I think there is a lot of room for improvement here:

1. I believe this is mandatory to be worked on before merging: the tests are _very_ slow, adding ~2 minutes of running time. I believe I'm doing something wrong in the tests but I don't know what and I'd like to know what to do to fix it.
2. Probably doable as part of a separate PR: I'm not sure the `StreamPredicate` abstraction is particularly fit for this anymore as it is, it could make sense to re-work that a bit.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -431,10 +445,11 @@ class ContractsService(
...
     )
-    contractsAndBoundary mapMaterializedValue { fob =>
-      fob.foreach(a => logger.debug(s"contracts fetch completed at: ${a.toString}"))
-      NotUsed
-    }
+    contractsAndBoundary
+      .mapMaterializedValue { fob =>
+        fob.foreach(a => logger.debug(s"contracts fetch completed at: ${a.toString}"))
+        NotUsed
+      }
```
stefanobaghino-da, line 452: Unnecessary change, revert.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/0d925938042007fac733a2929c0f500362603604

stefanobaghino-da (commented):
Furthermore, the documentation is still missing. Working on it.

cocreature (commented):
will finish review later but looks great so far!

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
stefanobaghino-da: The difference between what?

cocreature: I think what I was trying to say is "difficulty" not "difference". I blame lack of sleep.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -199,17 +215,29 @@ object WebSocketService {
...
           }
-        def fn(a: domain.ActiveContract[LfV]): Option[Positive] =
+        def fn(a: domain.ActiveContract[LfV], o: Option[domain.Offset]): Option[Positive] = {
+          import domain.Offset.ordering
+          import scalaz.syntax.order._
           q.get(a.templateId).flatMap { preds =>
-            preds.collect(Function unlift { case ((_, p), ix) => p(a.payload) option ix })
+            preds.collect(Function unlift { case ((_, p), ix) =>
+              val matchesPredicate = p(a.payload)
+              val matchesOffset = indexedOffsets(ix).flatMap(q => o.map((_, q))).fold(true) {
```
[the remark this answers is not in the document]
stefanobaghino-da: I think this returns an `Option[Boolean]`, right? I need a `Boolean` that is `true` if the offset is missing. Would you prefer me to create a separate method for this?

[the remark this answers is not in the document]
cocreature: oh you’re right yeah I think a separate method would be helpful.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -476,62 +555,123 @@ class WebSocketService(
...
+                    throw new IllegalStateException("Transaction in ACS")
+                }
+              }, {
+                // Get the earliest available offset from where to start from
+                val liveStartingOffset = Q.liveStartingOffset(offPrefix, request)
+                val StreamPredicate(_, _, fn, _) =
+                  Q.predicate(request, resolveTemplateId, lookupType)
+
+                Future.successful {
+                  contractsService
```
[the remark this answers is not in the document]
stefanobaghino-da: I'm relatively sure we always fetched only the ACS from the database, isn't that the case?

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -476,62 +555,123 @@ class WebSocketService(
...
+                    // can filter out anything from liveStartingOffset to the query-specific offset
+                    val acsEnd = offset.toOption.map(domain.StartingOffset(_))
+                    val shiftedRequest = Q.adjustRequest(acsEnd, request)
+
+                    // Get the earliest available offset from where to start from
+                    val liveStartingOffset = Q.liveStartingOffset(acsEnd, shiftedRequest)
+                    val StreamPredicate(_, _, fn, _) =
+                      Q.predicate(shiftedRequest, resolveTemplateId, lookupType)
+
+                    Source.single(liveBegin) ++ contractsService
```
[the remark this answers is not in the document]
stefanobaghino-da: Not necessarily, if the query offset is earlier than the end of the ACS we need to start from there.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -172,6 +172,22 @@ object WebSocketService {
...
+        maybePrefix: Option[domain.StartingOffset],
+        request: A,
+    ): Option[A]
+
+    def adjustRequest(
+        prefix: Option[domain.StartingOffset],
+        request: A,
+    ): A
+
+    def liveStartingOffset(
```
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/5e695f854f6153e81bbd8f359fa5f1e6b4509479

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -199,17 +215,29 @@ object WebSocketService {
         import scalaz.syntax.foldable._
         import util.Collections._
 
+        val indexedOffsets = request.queries.map(_.offset).toVector
```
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/ebc8889a911fd4497d1ee03cad02099c30924e4c

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
```diff
@@ -238,33 +239,58 @@ object JsonProtocol extends DefaultJsonProtocol with ExtraFormats {
...
         query: Option[Map[String, JsValue]],
     )
-    val validKeys = Set("templateIds", "query")
-    implicit val primitive: JsonReader[GACR] = jsonFormat2(GACR.apply)
+    val validKeys = Set("templateIds", "query") ++ validExtraFields
+    implicit val primitive: JsonReader[BaseRequest] = jsonFormat2(BaseRequest.apply)
     jsv => {
-      val GACR(tids, q) = jsv.convertTo[GACR]
+      val BaseRequest(tids, query) = jsv.convertTo[BaseRequest]
       val extras = jsv.asJsObject.fields.keySet diff validKeys
```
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/e7a79a749ce3a4996bd9678022c4182ddc4508c0

I used the term "unsupported" which is already used in the error message.

ledger-service/http-json/src/main/scala/com/digitalasset/http/domain.scala
```diff
@@ -210,6 +215,8 @@ object domain {
 
     implicit val semigroup: Semigroup[Offset] = Tag.unsubst(Semigroup[Offset @@ Tags.LastVal])
     implicit val ordering: Order[Offset] = Order.orderBy[Offset, String](Offset.unwrap(_))
+    implicit val orderingOptional: Order[Option[Offset]] =
```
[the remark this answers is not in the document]
stefanobaghino-da: Ah, that's exactly what I was looking for, thanks.

Addressed by https://github.com/digital-asset/daml/pull/9847/commits/cfbada101d60adc61a1fe89991ed90fc1fe17b5f

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
stefanobaghino-da: If a query contains the field `offset` the HTTP-JSON API thinks it's a late or duplicate offset message and it discards it with a message.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -251,6 +279,40 @@ object WebSocketService {
...
+        )
+
+      import scalaz.syntax.foldable1._
+      import domain.Offset.orderingOptional
+
+      override def liveStartingOffset(
+          prefix: Option[domain.StartingOffset],
+          request: SearchForeverRequest,
+      ): Option[domain.StartingOffset] =
+        request.queries
```
cocreature, line 322: do you assume that the request has already been adjusted? probably worth documenting

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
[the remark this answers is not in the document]
cocreature: Oh I see, this is not the case where we have no offset but it’s the case where we made no acs request because everything had an offset. What is slightly confusing is that `insertDeleteStepSource` will fetch the ACS again if you pass `None`. Afaict we will never actually pass `None` here (if `acsRequest` is None`, `liveStartOffset` must be `Some`) so maybe add an assertion and/or a comment to make that explict.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/b35dd470cd0cacb6999e4c4a883b7e685d4ad97e

stefanobaghino-da:
> I think there is a lot of room for improvement here:
> 
> 1. I believe this is mandatory to be worked on before merging: the tests are _very_ slow, adding ~2 minutes of running time. I believe I'm doing something wrong in the tests but I don't know what and I'd like to know what to do to fix it.
> 2. Probably doable as part of a separate PR: I'm not sure the `StreamPredicate` abstraction is particularly fit for this anymore as it is, it could make sense to re-work that a bit.

@S11001001 Can you please provide some feedback on the first point?

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/893c5536383a64fda2d9ff54b0187f15af0c9707

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/bf1c88e6ae28ce0be6a3da245d44ada31c0c8eb4

cocreature:
@stefanobaghino-da does the slowdown only apply to the new tests, i.e., if we drop those do we get back to something comparable to what we had before, or are tests slowed down across the board?

stefanobaghino-da:
> @stefanobaghino-da does the slowdown only apply to the new tests, i.e., if we drop those do we get back to something comparable to what we had before, or are tests slowed down across the board?

The tests in general are reasonably slow, i.e. they're actually quite fast for integration tests. The new ones I introduced are very slow and I would like to improve that but I have no idea how.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -916,6 +953,173 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+        Consume.interpret(
+          for {
+            acs <- readAcsN(expectedAcsSize)
+            contracts <- updateAcs(acs, expectLivenessMarker, end)
+            result = contracts
+              .map(_._2.asJsObject.fields("currency").asInstanceOf[JsString].value)
+              .groupBy(identity)
+              .map { case (k, vs) => k -> vs.size }
+            // If this fails with a match error, there is something else lingering in the stream
+            ContractDelta(Vector(), _, _) <- readOne
```
cocreature, line 1021:
```suggestion
            ContractDelta(Vector(), Vector(), Some(_)) <- readOne
```
afaict you are reading the heartbeat here (as made explicit by my suggestion). We send heartbeats every 5 seconds. With 6 tests, this gives you roughly a 30s delay at worst. Probably not the only source of slowness but maybe one.

Easy fix is to just drop this but it does unfortunately mean that you are more risky to drop something on the floor. Alternatively, you can increase the heartbeating frequency to reduce the wait.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -916,6 +953,173 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+      def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[domain.Offset]] =
+        Consume.interpret(
+          for {
+            _ <- readAcsN(expectedContracts)
+            ContractDelta(Vector(), _, Some(offset)) <- readOne
+            _ = killSwitch.shutdown()
+            _ <- drain
+          } yield offset
+        )
+      val (killSwitch, source) =
```
cocreature, line 995: you are probably paying some overhead for establishing separate streams for each call to `getLedgerEnd`. What I’d probably try is to hardcode the offsets temporarily and see if `getLedgerEnd` plays a significant role in the slowdown or note.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +748,42 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+      JsObject("events" -> JsArray.empty, "offset" -> JsNumber(0)).asInstanceOf[JsValue]
+    val dslSyntax = Consume.syntax[JsValue]
+    import dslSyntax._
+    import domain.Offset.ordering
+    import scalaz.syntax.order._
+    def go(
+        acs: Map[String, JsValue],
+        latest: Option[domain.Offset],
+    ): Consume.FCC[JsValue, Map[String, JsValue]] =
+      if (latest.fold(false)(_ >= until)) {
```
cocreature, line 768: I think I’m confused. If `latest` is actually greater than `until`, we don’t want to include this do we? It looks to me like we might be reading the heartbeat again which adds more 5s delays:
1. line 774 reads something with offset `until`
2. We get here, we read again in line 770, at this point we are reading the heartbeat and things work only by accident because it is empty.

Am I missing something here?

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Removing this did cut 30 seconds off. If we drop something the test should come back negative, so that should be ok. Thanks!

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +748,42 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+          ContractDelta(creates, deletes, _) <- readOne
+        } yield acs ++ creates -- deletes.map(_.contractId.unwrap)
+      } else {
+        for {
+          ContractDelta(creates, deletes, offset) <- readOne
+          next <- go(acs ++ creates -- deletes.map(_.contractId.unwrap), offset)
+        } yield next
+      }
+    for {
+      // Ignoring the live marker, w/ per-query offsets it can come in with an offset greater than `until`
```
cocreature, line 779: I think it would be nice to have some tests for what we return in the liveness markers. Primarily interested in test case 1 so feel free to omit the others.
1. All offsets come before the ledger end (or rather what the ACS service returns which is technically not guaranteed to be ledger end but that’s what it does atm).
2. All offsets come after the ACS end. You have to "guess" offsets in this case which is a bit odd but not too hard. If the ACS service actually does not return ledger end this matters a bit more.
3. Some mix where a few offsets are before and a few are after.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: I was looking at that and measured the gap between retrieving the ACS and getting the ledger end and it's consistently ~300 micros.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Mmmh, I think you're right. I did this to not miss the last object, but probably using a dummy create as you suggested initially is easier, let me try that.

cocreature: I forgot what the case actually was where we can actually miss something. If we read a non-heartbeat message and the offset is >= until we can stop. The only case where we might want to read a heartbeat is if `latest` is `None` but we already do that in the second case.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -916,6 +953,173 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+      s"""{"templateIds":["Iou:Iou"], "query":{"currency":"$currency"}}"""
+    def contractsQueryWithOffset(offset: domain.Offset, currency: String): String =
+      s"""{"templateIds":["Iou:Iou"], "query":{"currency":"$currency"}, "query_offset":"${offset.unwrap}"}"""
+    def contracts(currency: String, offset: Option[domain.Offset]): String =
+      offset.fold(contractsQuery(currency))(contractsQueryWithOffset(_, currency))
+    def ledgerEnd(expectedContracts: Int): Future[domain.Offset] = {
+      def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[domain.Offset]] =
+        Consume.interpret(
+          for {
+            _ <- readAcsN(expectedContracts)
```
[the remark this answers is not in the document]
stefanobaghino-da: After some playing around with `updateAcs` it'll probably be better to do this as part of this PR to make the tests run in a reasonable amount of time.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: I'll try to find a way to speed up our tests by not reading the heartbeat at the end and include these cases as well.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/ab2535a7f091464d915222749bebac2fd0b091d6

I switched from using offsets to asking for the expected number of events. I also renamed `ledgerEnd` to `acsEnd`, which is more accurate.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: After https://github.com/digital-asset/daml/pull/9847/commits/ab2535a7f091464d915222749bebac2fd0b091d6 tests are down to run in ~1s each, which is reasonable. Will work on this additional test cases next.

stefanobaghino-da: > All offsets come before the ledger end (or rather what the ACS service returns which is technically not guaranteed to be ledger end but that’s what it does atm).

This is covered by the existing test.

> All offsets come after the ACS end.

_and_

> Some mix where a few offsets are before and a few are after.

What would this test represent? It looks to me like this is not particularly relevant as offsets are opaque to clients and need to be observed before being used (which is a documented property of offsets). I can probably come up with a test for this but I'm not entirely sure of its usefulness.

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
```diff
@@ -238,33 +239,58 @@ object JsonProtocol extends DefaultJsonProtocol with ExtraFormats {
...
     }
   }
 
+  implicit val GetActiveContractsRequestFormat: RootJsonReader[domain.GetActiveContractsRequest] =
+    requestJsonReader(validExtraFields = Set.empty)((tids, query, _) =>
+      domain.GetActiveContractsRequest(tids, query)
+    )
+
+  implicit val SearchForeverQueryFormat: RootJsonReader[domain.SearchForeverQuery] = {
+    val OffsetKey = "queryOffset"
```
cocreature, line 283: Given that once we set this, we’re stuck with this ~forever~ a very long time could you expand on your earlier comment that calling this `offset` results in messages being reinterpreted? I don’t understand why we cannot disambiguate between the two requests.

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
```diff
@@ -238,33 +239,58 @@ object JsonProtocol extends DefaultJsonProtocol with ExtraFormats {
     *  1. template IDs are required
     *  2. query key may be absent
     *  3. special error if you appear to have queried outside 'query'
+    *
+    *  This provides an (almost) consistent behavior when reading the 'templateIds' and
+    *  'query' fields. Further extra fields may be added by concrete implementations.
     */
-  implicit val GetActiveContractsRequestFormat: RootJsonReader[domain.GetActiveContractsRequest] = {
-    case class GACR(
+  private def requestJsonReader[Request](validExtraFields: Set[String])(
```
cocreature, line 246: probably not for this PR but this seems like something that should have unit tests.

cocreature (approved):
Nice work! Thanks a lot for pushng this through

docs/source/json-api/index.rst
```diff
@@ -1332,6 +1345,9 @@ off an initial "loading" indicator::
         "offset": "2"
     }
 
+Events in the following "live" data may include ``events`` that precede
```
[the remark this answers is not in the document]
stefanobaghino-da: > Is there a good usecase for this behavior?

Let's say that I have two contract templates, `First` and `Second`. I start querying for `First` on the client. Then I want to query for `Second`. In order to not open a second connection, I save the latest offset (let's say it's `42`) that I got for `First` and create a new query asking for all `First` from `42` and all `Second`s starting from the ACS. While this was happening on the client, a new `First` is created at `45` and a new `Second` is created at `47`. When I retrieve the contracts with per-query offsets, I will get the ACS for `Second` ending at `47`, then the `First` created at `45`, followed by the rest of the active stream.

Does this example make sense?

Originally this paragraph was much longer, but I thought it was a bit of a wall of text and wanted to keep it short, readable and simple.

If the example makes sense, though, I can include it in the documentation. What do you think?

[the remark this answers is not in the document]
cocreature: Oh I see, I think I misread the text. makes sense :+1:

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +748,32 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+        acs: Map[String, JsValue],
+        missingEvents: Int,
+    ): Consume.FCC[JsValue, Map[String, JsValue]] =
+      if (missingEvents <= 0) {
+        point(acs)
+      } else {
+        for {
+          ContractDelta(creates, deletes, _) <- readOne
+          newAcs = acs ++ creates -- deletes.map(_.contractId.unwrap)
+          events = creates.size - deletes.size
```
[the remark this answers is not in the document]
stefanobaghino-da: > It’s not the total number of events since deletes (which maybe should be called archives) are clearly also events.

The intent was actually this, I simply put a `-` there but I meant `+`. It works simply because I only have creates in tests.

> which maybe should be called archives

In principle I agree, I assumed that the type parameter `D` in the various domain objects representing events in the JSON API was meant to mean `Delete` and wanted to go with that, but thinking better about it could have been picked because it came after `C` in the alphabet (which I assume stands for `Create`). Changing this,

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -916,6 +943,171 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
+        eurFrom: Option[domain.Offset],
+        usdFrom: Option[domain.Offset],
+        expected: Map[String, Int],
+    ): Future[Assertion] = {
+      val fakeLiveMarker: JsValue = JsObject("events" -> JsArray.empty, "offset" -> JsNumber(0))
+      def go(killSwitch: UniqueKillSwitch): Sink[JsValue, Future[Assertion]] = {
+        Consume.interpret(
+          for {
+            acs <- readAcsN(expectedAcsSize)
+            _ <- if (acs.nonEmpty) readOne else point(fakeLiveMarker)
```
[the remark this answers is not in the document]
stefanobaghino-da: If I don't read the marker but it's there, wouldn't I cause an error down the road as the marker is read when reading the following messages? Or is there a way to read and discard that I'm missing?

docs/source/json-api/index.rst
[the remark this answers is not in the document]
stefanobaghino-da: Should I add the practical example or keep it short?

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
stefanobaghino-da: @cocreature I'll mark this as resolved, feel free to reopen if you think there's a good case for these tests to come in.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -747,6 +748,32 @@ abstract class AbstractWebsocketServiceIntegrationTest
     go(createCount)
   }
 
+  /** Updates the ACS retrieved with [[readAcsN]] with the given number of events
+    * The caller is in charge of reading the live marker if that is expected
+    */
+  private[this] def updateAcs(
+      acs: Vector[(String, JsValue)],
```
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/581c22b7f87cf619333944221c1f5b15e9b14b26

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/178fd6ccee57b84e434cbc5df72edc5ee1f07407

docs/source/json-api/index.rst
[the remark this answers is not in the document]
cocreature: I’d probably add it. Easy enough to skip by if users don’t care but if I need a few seconds to come up with one it’s probably also useful for others.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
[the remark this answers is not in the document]
cocreature: I’m not suggesting to not read the marker. I’m suggesting to replace `point(fakeLiveMarker)` by `point(())`.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
cocreature: Fine to not add it for now at least. The case I was thinking off is where the ACS service is lagging behind which afaik the docs explicitly state that it can. Our implementation doesn’t do that atm so you can’t hit it but I’m slightly worried this might change and we forget to fix this.

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/JsonProtocol.scala
stefanobaghino-da: This used to make some earlier tests fail. I thought about it a bit and what the current code does is just check for the presence of `offset` in the object root and reject this with a mistake if an `offset` message was passed already. Thinking better about it, the per-query `offset`s are always in the array that specifies the queries, not in the root, so I have no idea why this failed initially. Probably something else was broken at that point. Changing this to `offset`.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/7575a67b643248a31d037dc60fe9d8d65539bb44

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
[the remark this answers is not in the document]
stefanobaghino-da: I assumed that would have not compiled but if it type-checks that's fine with me. 🙂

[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/9f80f40a9f647a1a376de4dde63b573d541b75ae

docs/source/json-api/index.rst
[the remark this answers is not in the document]
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/9847/commits/58853091ac6f22cb2c46f79b55c20e87ee0456e9 (alongside a few fixes to formatting -- a space is needed after the end of the fixed-width formatting markup to not mess it up).

Bad:
```
``offset``s
```

Good:
```
``offset`` s
```

PR #9848, opened 2021-05-31
Switch Scala version Bazel config

stefanobaghino-da:
changelog_begin
changelog_end

This allows to easily test with Scala 2.12.x locally. The `scala_2_13` Bazel
configuration is no longer needed becase now Scala 2.13.x is the default.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -374,9 +374,9 @@ jobs:
         CMP="$(semver compare "$RELEASE_TAG" '1.11.0-snapshot.20210212.6300.0.ad161d7f')"
 
         if [[ $CMP == '1' || "$RELEASE_TAG" == '0.0.0' ]]; then
-            setvar scala_2_13 true
+            setvar scala_2_12 false
         else
-            setvar scala_2_13 false
+            setvar scala_2_12 true
         fi
       name: out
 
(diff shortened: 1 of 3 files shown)


ci/build.yml
```diff
@@ -374,9 +374,9 @@ jobs:
         CMP="$(semver compare "$RELEASE_TAG" '1.11.0-snapshot.20210212.6300.0.ad161d7f')"
 
         if [[ $CMP == '1' || "$RELEASE_TAG" == '0.0.0' ]]; then
-            setvar scala_2_13 true
+            setvar scala_2_12 false
         else
-            setvar scala_2_13 false
+            setvar scala_2_12 true
```
stefanobaghino-da, line 379: I'm not 100% sure that this makes sense but that's why we have code reviews. 🙂

cocreature: I don’t think it makes sense. This variable was used to hack around the fact that the Scala 2.13 job does not work for older commits. We don’t have this issue for the Scala 2.12 job we have now. If anything, we should drop this completely or just leave it as it was.

cocreature (approved):
lgtm ignoring the change to build.yml

aherrmann-da (approved):

ci/build.yml
stefanobaghino-da: Addressed in https://github.com/digital-asset/daml/pull/9848/commits/93c77f19f5f074c7f97284baa9d47c26cf5e853b

PR #9850, opened 2021-05-31
Fix the notion of transaction equivalence in the ledger model for exceptions.

sofiafaro-da:
@andreaslochbihler-da found that our normalization rules don't normalize rollbacks across exercises, which our notion of transaction equivalence expects them to. For example, the transactions

```
rollback [ exercise [ rollback [ create ]]]
```

and

```
rollback [ exercise [ create ] ]
```

are considered equivalent under the old definition, but do not normalize to each other.

This PR is one possible solution, where we keep our current normalization rules but change the definition of transaction equivalence. In particular, instead of just requiring that the set of rolled back actions to be the same, we require that the set of "rollback children" be the same, where a "rollback child" is an action that has a rollback node as its direct parent. This prevents normalization across exercises because that would eliminate some rollback children.

The idea here is that the set of "rollback childern" encodes the set of "rolled back actions" relative to each action's subtree, unlike before where we only looked at the set of "rolled back actions" relative to the whole transaction. The set of "rollback children" encodes all that information, but with a simple definition.

A completely different solution is to introduce new normalization rules that work across exercise nodes. Unfortunately these rules would be non-local, and that seems very difficult to do correctly & efficiently.

Follow up to #9816, part of #8020.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/concepts/ledger-model/ledger-exceptions.rst
+++ b/docs/source/concepts/ledger-model/ledger-exceptions.rst
@@ -144,13 +144,13 @@ More generally, two transactions are equivalent if:
 
 - The transactions have the same "before-after" relation between their actions.
 
-- The transactions have the same set of "rollback roots".
-  A rollback root is an action whose parent is a rollback node.
+- The transactions have the same set of "rollback children".
+  A "rollback child" is an action whose direct parent is a rollback node.
 
 For all three transactions above, the "transaction tree ignoring rollbacks"
 consists only of top-level actions (`act1`, `act2`, and `act3`), the
 "before-after" relation only says that `act2` comes before `act3`,
-and all three actions are rollback roots. Thus all three transactions
+and all three actions are rollback children. Thus all three transactions
 are equivalent.
 
 **Transaction normalization** is the process by which equivalent transactions
--- a/docs/source/concepts/ledger-model/ledger-exceptions.rst
+++ b/docs/source/concepts/ledger-model/ledger-exceptions.rst
@@ -144,13 +144,14 @@ More generally, two transactions are equivalent if:
 
 - The transactions have the same "before-after" relation between their actions.
 
-- The transactions have the same set of rolled back actions. That is, they have
-  the same set of actions directly or indirectly under a rollback node.
+- The transactions have the same set of "rollback roots".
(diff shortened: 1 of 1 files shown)


docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -144,13 +144,14 @@ More generally, two transactions are equivalent if:
 
 - The transactions have the same "before-after" relation between their actions.
 
-- The transactions have the same set of rolled back actions. That is, they have
-  the same set of actions directly or indirectly under a rollback node.
+- The transactions have the same set of "rollback roots".
+  A rollback root is an action whose parent is a rollback node.
```
cocreature, line 148:
```suggestion
  A rollback root is an action whose parent is not a rollback node.
```
Otherwise I don’t understand this.

cocreature: Ah no, I think I get it now nvm.

docs/source/concepts/ledger-model/ledger-exceptions.rst
```diff
@@ -144,13 +144,14 @@ More generally, two transactions are equivalent if:
 
 - The transactions have the same "before-after" relation between their actions.
 
-- The transactions have the same set of rolled back actions. That is, they have
-  the same set of actions directly or indirectly under a rollback node.
+- The transactions have the same set of "rollback roots".
```
cocreature, line 147: The name seems quite confusing. A rollback root is not a rollback node. It’s the child of a rollback node. Maybe “rollback children” (but not descendants) would be clearer?

sofiafaro-da: Ah, yeah. I wasn't sure about the name. "Rollback child"/"rollback children" works.

sofiafaro-da: Changed. How's this?

cocreature (approved):
Nice, thank you!

sofiafaro-da:
Merging since this reflects the current normalization rules more accurately.

PR #9854, opened 2021-05-31
Introduce metrics in the http-json service.

realvictorprm:
This implements simple request-time metrics (for the http-json service).

Looking into existing code of the code base I think my implementation is a bit controversial, however I decided to mostly not use existing code within the code base, because the abstractions I used are not tied to Future (thus more flexible) and from an external library (cats.effect) which means less code to maintain. Especially the flexibility for the implementation was important to me, because I would like to reduce the usage of Future if possible and use an abstraction which is referential transparent and cancelable (my implementation is in this regard not perfect either but better less Future than more).

Besides these points, the implementation tries to make sure all resources are released properly as the introduction of an additional resource forced me to take care of scaleable resource management with good working shutdown management. This is the reason for the great refactoring of the main method.

Lastly, I'm aware that we did not decide on how much Cats we like to use here, however to me it was more important to produce well working and maintainable code first.

changelog_begin
[HTTP/JSON API]
- metrics reporting can now be chosen via the command line option --metrics-reporter, valid options are console, csv://, graphite:// and prometheus://
- metrics reporting interval can also now be chosen via a command line option --metrics-reporting-interval
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -40,6 +40,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 import scala.util.control.NonFatal
 import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
+import com.daml.metrics.{Metrics, Timed}
 
 class Endpoints(
     allowNonHttps: Boolean,
@@ -63,7 +64,8 @@ class Endpoints(
   import Uri.Path._
 
   def all(implicit
-      lc: LoggingContextOf[InstanceUUID]
+      lc: LoggingContextOf[InstanceUUID],
+      metrics: Metrics,
   ): PartialFunction[HttpRequest, Future[HttpResponse]] = {
     val dispatch: PartialFunction[HttpRequest, LoggingContextOf[
       InstanceUUID with RequestID
@@ -103,16 +105,16 @@ class Endpoints(
     }
     import scalaz.std.partialFunction._, scalaz.syntax.arrow._
     (dispatch &&& { case r => r }) andThen { case (lcFhr, req) =>
-      extendWithRequestIdLogCtx(implicit lc =>
-        // TODO: Refactor this somehow into an own function
-        for {
-          _ <- Future.unit
-          _ = logger.trace(s"Incoming request on ${req.uri}")
-          t0 = System.nanoTime()
-          res <- lcFhr(lc)
-          _ = logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
-        } yield res
-      )
+      extendWithRequestIdLogCtx(implicit lc => {
+        val t0 = System.nanoTime
+        logger.trace(s"Incoming request on ${req.uri}")
+        Timed
+          .future(metrics.registry.timer("daml.ledger_service.http_json.requests_timer"), lcFhr(lc))
+          .map(res => {
+            logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
+            res
+          })
+      })
     }
   }
 
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
@@ -34,6 +34,7 @@ import com.daml.ledger.client.{LedgerClient => DamlLedgerClient}
 import com.daml.ledger.service.LedgerReader
 import com.daml.ledger.service.LedgerReader.PackageStore
 import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
+import com.daml.metrics.Metrics
 import com.daml.ports.{Port, PortFiles}
 import com.daml.scalautil.Statement.discard
 import scalaz.Scalaz._
@@ -71,6 +72,7 @@ object HttpService {
       aesf: ExecutionSequencerFactory,
       ec: ExecutionContext,
       lc: LoggingContextOf[InstanceUUID],
+      metrics: Metrics,
   ): Future[Error \/ ServerBinding] = {
 
     logger.info("HTTP Server pre-startup")
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
@@ -7,12 +7,16 @@ import java.nio.file.Path
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http.ServerBinding
 import akka.stream.Materializer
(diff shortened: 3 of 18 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -103,16 +105,16 @@ class Endpoints(
...
-          t0 = System.nanoTime()
-          res <- lcFhr(lc)
-          _ = logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
-        } yield res
-      )
+      extendWithRequestIdLogCtx(implicit lc => {
+        val t0 = System.nanoTime
+        logger.trace(s"Incoming request on ${req.uri}")
+        Timed
+          .future(metrics.registry.timer("daml.ledger_service.http_json.requests_timer"), lcFhr(lc))
```
tudor-da, line 112: We usually put all the metrics definitions in our global `com.daml.metrics.Metrics` definition registry. Can you do the same here?

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -71,6 +72,7 @@ object HttpService {
       aesf: ExecutionSequencerFactory,
       ec: ExecutionContext,
       lc: LoggingContextOf[InstanceUUID],
+      metrics: Metrics,
```
tudor-da, line 75: Is this one needed?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -50,6 +55,35 @@ object Main {
...
+      F: Applicative[F],
+      globalConcurrentEffect: ConcurrentEffect[F],
+      globalTimer: Timer[F],
+      loggingCtx: LoggingContextOf[InstanceUUID],
+  ): F[A] = {
+    import cats.FlatMap.ops.toAllFlatMapOps
+    for {
+      _ <- F.pure(()): F[Unit]
+      shutdownHookCalled <- Deferred[F, Unit](globalConcurrentEffect)
+      didProgramAlreadyEnd <- Deferred[F, Unit](globalConcurrentEffect)
```
tudor-da, line 71: Why is the same effect deferred twice?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -50,6 +55,35 @@ object Main {
...
+      shutdownHookTimeout: FiniteDuration = 10.seconds,
+  )(implicit
+      F: Applicative[F],
+      globalConcurrentEffect: ConcurrentEffect[F],
+      globalTimer: Timer[F],
+      loggingCtx: LoggingContextOf[InstanceUUID],
+  ): F[A] = {
+    import cats.FlatMap.ops.toAllFlatMapOps
+    for {
+      _ <- F.pure(()): F[Unit]
```
tudor-da, line 69: Can't we avoid this redundant application here?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -50,6 +55,35 @@ object Main {
...
+          s"Shutdown hook will wait $shutdownHookTimeout for the application to exit gracefully otherwise it will forcefully shutdown"
+        )
+        globalConcurrentEffect
+          .toIO(Concurrent.timeoutTo(didProgramAlreadyEnd.get, shutdownHookTimeout, F.unit))
+          .unsafeRunSync()
+      }
+      result <- program(shutdownHookCalled)
+      _ <- didProgramAlreadyEnd.complete(())
+    } yield result
+  }
```
tudor-da, line 85: Wouldn't this be simpler without Cats Effect? I'm far from expert in this lib, but it seems extremely hard to follow what's going on

cocreature (changes requested):
Let’s please not entangle switching to cats-effect with adding metrics. Those are two unrelated topics and trying to solve both at the same time is just confusing and makes the PR hard to review.

As for switching to cats-effect, ignoring whether we want to do this in general, I think here you are making the situation worse. `akka-http` forces us into futures everywhere. You are only using cats-effect to startup the http-server (the startup is pretty uninteresting from a metrics pov which backs up my earlier point that we shouldn’t tie the two together) only to then handover to akka-http for the actual program which does not use cats-effect at all. The startup logic seems to get significantly more complex and you’re introducing an extra concept a reader of this code has to understand.

If we really want to use cats-effect more aggressively, I think we probably need to switch http4s to actually be able to use it in the interesting logic which is a huge undertaking that we should thoroughly discuss upfront and not sneak in via metrics changes.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -86,68 +120,160 @@ object Main {
...
-            logFailure("Shutdown error", fa)
-            terminate()
-          }
-      }
-    }
+          (contractDao, config.jdbcConfig) match {
+            case (Some(dao), Some(c)) if c.createSchema =>
+              logger.info("Creating DB schema...")
+              import dao.{jdbcDriver, logHandler}
+              Try(dao.transact(ContractDao.initialize).unsafeRunSync()) match {
```
tudor-da, line 175: Is `dao.transact` supposed to throw if it is returning an `IO`? 
On another note, can't you thread it in the `IO` including the error handling instead of `..unsafeRunSync()) match{ .. IO(ErrorCodes.Ok)` ?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: ah thanks for the heads up! this is the thing I wanted to do but forgot :)

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: I don't think this is going to get that much simpler if I get rid of cats effect here.

realvictorprm:
I'll retry implementing this

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: This is less of something being deferred, it is more about a flag which I use to concurrently check if something is done or not done.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
tudor-da: You might be right, I am not an expert as I mentioned.

PR #9858, opened 2021-05-31
[In-memory fan-out] BuffersUpdater implementation

tudor-da:
In the context of the in-memory fan-out implementation, this PR introduces the `BufferUpdater` that populates the mutable contract state cache and a newly-introduced `TransactionsBuffer` (to be used for serving `getTransactions` and `getTransactionTrees`):
* The `TransactionsBuffer` is instantiated (as a `EventsBuffer`) in `ReadOnlySqlLedgerWithMutableCache`.
* Implementations and tests for the `BuffersUpdater` which updates:
  * The `MutableCacheBackedContractStore`.
  * `TransactionsBuffer`.
* Introduced `buffered-ledger-api-streams-unsafe` CLI flag for enabling in-memory fan-out flow.
* Enabled the `buffered-ledger-api-streams-unsafe` in ledger-on-sql conformance test for append-only schema.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -548,19 +548,19 @@ private[appendonlydao] object TransactionsReader {
           s"You can only split a range in a strictly positive number of chunks ($numberOfChunks)"
         )
 
-      case _ if numberOfChunks == 1 || minChunkSize >= rangeSize =>
+      case _ if numberOfChunks == 1 || minChunkSize > (rangeSize / 2L + 1L) =>
```
tudor-da, line 551: Not related to this PR, but this condition needed tightening

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -548,19 +548,19 @@ private[appendonlydao] object TransactionsReader {
           s"You can only split a range in a strictly positive number of chunks ($numberOfChunks)"
         )
 
-      case _ if numberOfChunks == 1 || minChunkSize >= rangeSize =>
+      case _ if numberOfChunks == 1 || minChunkSize > (rangeSize / 2L) =>
```
mziolekda, line 551: Do we need this case actually? Methinks `case _` achieves exactly the same outcome

ledger/participant-integration-api/src/main/scala/platform/index/BuffersUpdater.scala
```diff
@@ -0,0 +1,168 @@
...
+    Offset.beforeBegin -> EventSequentialId.beforeBegin
+  )
+
+  private val (transactionLogUpdatesKillSwitch, transactionLogUpdatesDone) =
+    RestartSource
+      .withBackoff(
+        RestartSettings(
+          minBackoff = minBackoffStreamRestart,
+          maxBackoff = 10.seconds,
+          randomFactor = 0.2,
```
mziolekda, line 64: randomFactor helps when there are multiple consumers hitting the resource at the same time. I don't think it is the case here.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
tudor-da: Not quite. For `rangeSize == 100` and `minChunkSize = 51`, `case _` throws. I've seen it in action :trollface:

ledger/participant-integration-api/src/test/suite/scala/platform/index/BuffersUpdaterSpec.scala
```diff
@@ -0,0 +1,302 @@
...
+        exerciseResult = null,
+      )
+      val transaction = TransactionLogUpdate.Transaction(
+        transactionId = "some-tx-id",
+        commandId = "some-cmd-id",
+        workflowId = "some-workflow-id",
+        effectiveAt = Instant.EPOCH,
+        offset = Offset.beforeBegin,
+        events = Vector(
+          TransactionLogUpdate.CreatedEvent(
```
mziolekda, line 235: Don't you want to factor it out to a val to increase readability, same as in consumingExercise?

ledger/participant-integration-api/src/main/scala/platform/index/BuffersUpdater.scala
tudor-da: True. Changed to `0.0`

ledger/participant-integration-api/src/test/suite/scala/platform/index/BuffersUpdaterSpec.scala
tudor-da: Done

ledger/participant-integration-api/src/main/scala/platform/index/BuffersUpdater.scala
```diff
@@ -0,0 +1,168 @@
...
+  * @param minBackoffStreamRestart Minimum back-off before restarting the transaction log updates stream.
+  * @param mat The Akka materializer.
+  * @param loggingContext The logging context.
+  * @param executionContext The execution context.
+  */
+private[index] class BuffersUpdater(
+    subscribeToTransactionLogUpdates: SubscribeToTransactionLogUpdates,
+    updateTransactionsBuffer: (Offset, TransactionLogUpdate) => Unit,
+    toContractStateEvents: TransactionLogUpdate => Iterator[ContractStateEvent],
+    updateMutableCache: ContractStateEvent => Unit,
```
meiersi-da, line 46: @tudor-da : is there a particular reason that this `BufferUpdater` takes these three individual elements instead of just a `updateBuffers :: (Offset, TransactionLogUpdate) => Unit` argument?

In the latter case, the caller could take care of doing what it needs doing for each `TransactionLogUpdate`; i.e., we'd gain more context independence.

meiersi-da: It might well be that this file is just a glue code file, and it makes sense to have the methods in here. Even in that case though you might consider factoring out the pure code that updates the multiple caches, as this way the reader knows that these calls always happen together in the same way.

ledger/participant-integration-api/src/test/suite/scala/platform/index/BuffersUpdaterSpec.scala
```diff
@@ -0,0 +1,303 @@
...
+        subscribeToTransactionLogUpdates = sourceSubscriptionFixture,
+        updateTransactionsBuffer = updateTransactionsBufferMock,
+        toContractStateEvents = Map.empty,
+        updateMutableCache = _ => (),
+        minBackoffStreamRestart = 1.millis,
+        shutdownCodeCapture.set,
+      )(materializer, loggingContext, scala.concurrent.ExecutionContext.global)
+
+      eventually {
+        shutdownCodeCapture.get() shouldBe 1
```
meiersi-da, line 166: I assume the exception and/or the shutdown is also logged at ERROR level, is it?

meiersi-da: Could / should the test assert that as well?

meiersi-da (commented):
Thanks for the PR @tudor-da; and the diligence in creating it!

I have made a review pass, and did not find any big problems, but have to admit that I don't have a good mental model to check how all the streams are setup, and that they will work fine together. There I'm relying on the tests you've written.

ledger/participant-integration-api/src/main/scala/platform/index/BuffersUpdater.scala
tudor-da: `updateBuffers :: (Offset, TransactionLogUpdate) => Unit` is the actual essence of this class. Factoring it out fully would leave this class an empty shell.

tudor-da: > It might well be that this file is just a glue code file, and it makes sense to have the methods in here. Even in that case though you might consider factoring out the pure code that updates the multiple caches, as this way the reader knows that these calls always happen together in the same way.

Sensible suggestion, thanks. I have extracted the logic that updates the buffers from `BuffersUpdater` and passed it as a parameter. That logic is assembled in the `BuffersUpdater$` companion object.

ledger/participant-integration-api/src/test/suite/scala/platform/index/BuffersUpdaterSpec.scala
tudor-da: Asserted the logging 👍

ledger/participant-integration-api/src/main/scala/platform/index/BuffersUpdater.scala
meiersi-da: :+1:

PR #9864, opened 2021-06-01
Special case _tryCatch @Update in LFConversion.

sofiafaro-da:
This is in line with the special cases for (>>=), pure, etc

Part of #8020

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
+++ b/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
@@ -1143,6 +1143,16 @@ convertExpr env0 e = do
             EVal semi' <- convertExpr env semi
             let bind' = EVal semi'{qualObject = mkVal ">>="}
             pure $ mkEApps bind' [TyArg monad', TmArg dict', TyArg t1', TyArg t2', TmArg x', TmArg (ETmLam (mkVar "_", t1') y')]
+    go env tryCatch@(VarIn DA_Internal_Exception "_tryCatch") allArgs@(LType monad : LExpr _dict : LType t : LExpr x : LExpr (Lam b y)) = do
+        monad' <- convertType env monad
+        case monad' of
+            TBuiltin BTUpdate -> do
+                t' <- convertType env t
+                x' <- convertExpr env x
+                y' <- convertExpr env y
+                pure (UTryCatch t' (ETmApp x' EUnit) (convVar b) y')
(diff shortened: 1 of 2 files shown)


compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1143,6 +1143,16 @@ convertExpr env0 e = do
             EVal semi' <- convertExpr env semi
             let bind' = EVal semi'{qualObject = mkVal ">>="}
             pure $ mkEApps bind' [TyArg monad', TmArg dict', TyArg t1', TyArg t2', TmArg x', TmArg (ETmLam (mkVar "_", t1') y')]
+    go env tryCatch@(VarIn DA_Internal_Exception "_tryCatch") allArgs@(LType monad : LExpr _dict : LType t : LExpr x : LExpr (Lam b y) : args) = do
```
cocreature, line 1146: Is there an intuition for what we special case here and what we don’t?
It seems like it’s all monomorphized typeclass methods are special cased? If so I think `throw` should also fall under this.
I’m not quite sure about `create` and friends. It seems like they should also fall under this but afaict we don’t special case them.
Or am I just searching for a pattern that does not exist and the truth is that we do it more or less randomly?

cocreature (approved):
Nice, thanks!

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
sofiafaro-da: I don't think there's a consistent pattern... :'-(

My intuition is that for something simple like create/throw/etc the simplifier eliminates the indirection, but for `UBind` and other expressions that introduce binders in the LF, or have an unusual form, it's going to struggle.

sofiafaro-da: Maybe the pattern is closer to "does this have special control flow"?  (And "pure" is special cased to make the output of LFConversion more readable.)

cocreature: Probably better to just admit that we don’t have a pattern and try to either introduce one or just accept that it’s a case by case decision. At least I can stop searching for a pattern now :slightly_smiling_face:

PR #9867, opened 2021-06-01
Daml-lf suffixCid ignores V0 contract ids rather than erroring

oliverse-da:
- canton currently ignores v0 contracts upon suffixing as those may still appear in value parameters. Ignoring rather erroring on such contract ids will make it possible for canton to adopt the lf-based suffixer
- corda appears to work with v1 contract ids, assuming that suffixing always results in a Right:
  https://github.com/DACH-NY/daml-on-corda/blob/main/engine/src/main/scala/com/digitalasset/platform/corda/engine/VaultBackedDamlState.scala#L101

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/CidContainer.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/CidContainer.scala
@@ -33,7 +33,7 @@ object CidMapper {
   type NoCidChecker[-A1, +A2] = CidChecker[A1, A2, Nothing]
 
   type CidSuffixer[-A1, +A2] =
-    CidMapper[A1, A2, Value.ContractId, Value.ContractId.V1]
+    CidMapper[A1, A2, Value.ContractId, Value.ContractId]
 
   @SuppressWarnings(Array("org.wartremover.warts.Any"))
   private val _trivialMapper: CidMapper[Any, Any, Nothing, Any] =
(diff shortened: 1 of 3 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/CidContainer.scala
```diff
@@ -74,10 +74,7 @@ trait CidContainer[+A] {
     suffixer.traverse[String] {
       case Value.ContractId.V1(discriminator, Bytes.Empty) =>
         Value.ContractId.V1.build(discriminator, f(discriminator))
-      case acoid @ Value.ContractId.V1(_, _) =>
-        Right(acoid)
-      case acoid @ Value.ContractId.V0(_) =>
```
oliverse-da, line 79: Fyi @simonmaxen-da @danielporterda @vivd-da - it looks like in the corda code base you also use `suffixCid` but assume that you are working with V1 contract ids only. From what I could tell (see description of this PR) this should not affect daml-on-corda.

simonmaxen-da: That is correct.  We don't have an existing install base so all our ContractIds will only ever be V1 is my understanding.

oliverse-da: Thank you for confirming, Sir!

cocreature (approved):
Thanksk

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/value/CidContainer.scala
```diff
@@ -74,10 +74,7 @@ trait CidContainer[+A] {
...
-        Left(s"expect a Contract ID V1, found $acoid")
+      case acoid => Right(acoid)
```
remyhaemmerle-da, line 77: (minor) I prefer to keep the pattern exhaustiveness check.  

```suggestion
      case acoid: Value.ContractId.V0 => Right(acoid)
```

remyhaemmerle-da (approved):
LGTM

PR #9870, opened 2021-06-01
Make data-deps exception tests version-aware.

sofiafaro-da:
This PR changes the exception data-dependencies tests to do a cross-version import of exceptions (once we release exceptions in preview).

It also removes the last exceptions TODO from the Haskell side, other than changing `featureMinVersion` once we start releasing excepions.

Part of #8020

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/src/DA/Test/DataDependencies.hs
+++ b/compiler/damlc/tests/src/DA/Test/DataDependencies.hs
@@ -1211,7 +1211,7 @@ tests Tools{damlc,repl,validate,davlDar,oldProjDar} = testGroup "Data Dependenci
             [ "build"
             , "--project-root", tmpDir </> "lib"
             , "-o", tmpDir </> "lib" </> "lib.dar"
-            , "--target", LF.renderVersion (LF.featureMinVersion LF.featureDamlExceptions) ]
+            , "--target", LF.renderVersion (LF.featureMinVersion LF.featureExceptions) ]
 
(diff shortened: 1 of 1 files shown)


compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -1183,8 +1183,6 @@ tests Tools{damlc,repl,validate,davlDar,oldProjDar} = testGroup "Data Dependenci
             ]
         callProcessSilent damlc [ "build", "--project-root", tmpDir </> "proj" ]
 
-    -- TODO https://github.com/digital-asset/daml/issues/8020
-    --   Replace with a simpleImportTest once exceptions are stable.
```
cocreature, line 1187: Should we allow specifying versions in `simpleImportTest`? Afaict that’s the only difference here.

cocreature: Ah I guess we also run `damlc test` so not quite.

cocreature (approved):
Thanks!

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
sofiafaro-da: We probably should refactor these tests though.

PR #9873, opened 2021-06-01
Metrics for streamed elements

mziolekda:
Add metrics counting the elements in transaction, transaction-tree, completion and acs streams

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiActiveContractsService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiActiveContractsService.scala
@@ -34,8 +34,6 @@ private[apiserver] final class ApiActiveContractsService private (
 
   private val logger = ContextualizedLogger.get(this.getClass)
 
-  private val acsCounter = metrics.daml.lapi.streams.acs
-
   override protected def getActiveContractsSource(
       request: GetActiveContractsRequest
   ): Source[GetActiveContractsResponse, NotUsed] =
@@ -46,10 +44,7 @@ private[apiserver] final class ApiActiveContractsService private (
           .validate(request.getFilter)
           .fold(Source.failed, backend.getActiveContracts(_, request.verbose))
           .via(logger.logErrorsOnStream)
-          .map(item => {
-            acsCounter.inc()
-            item
-          })
+          .via(StreamMetrics.countElements(metrics.daml.lapi.streams.acs))
     }
 
   override def bindService(): ServerServiceDefinition =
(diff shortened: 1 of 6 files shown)


gerolf-da (approved):
Thanks, @mziolekda!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiActiveContractsService.scala
```diff
@@ -42,6 +46,10 @@ private[apiserver] final class ApiActiveContractsService private (
           .validate(request.getFilter)
           .fold(Source.failed, backend.getActiveContracts(_, request.verbose))
           .via(logger.logErrorsOnStream)
+          .map(item => {
```
tudor-da, line 49: Counting for all consumers might lead to metrics with ambiguous metrics when we have multiple consumers with varying behaviors in the application. Are you ok with this limitation or should we rather particularize the metric per consumer somehow (e.g. use the traceId)?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -115,6 +125,10 @@ private[apiserver] final class ApiTransactionService private (
         )
         .via(logger.debugStream(transactionTreesLoggable))
         .via(logger.logErrorsOnStream)
+        .map(item => {
```
tudor-da, line 128: I would extract a generic/extension method and instrument all sources with it for conciseness

tudor-da: e.g. 
```
  implicit class InstrumentedCountingSource[E, N](s: Source[E, N]) {
    def counter(counter: Counter): Source[E, N] =
      s.map(item => {
        counter.inc()
        item
      })
  }
```
or a simple object method if extension methods don't appeal to you. What do you think?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiActiveContractsService.scala
mziolekda: You cannot expand the metrics namespace without serious consequences for the metrics gathering systems. Wavefront does not support well a dynamically growing metrics namespace (datadog also does not), we are also reaching the limit of what grafana is willing to accept. 
The solution would be to add labels, in dropwizard terms they are called tags. They will be supported in version 5.0 of dropwizard, which has been i n the making for the past 3 years.

tudor-da (approved):
LGTM!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiActiveContractsService.scala
tudor-da: You're right that it's not ideal to have dynamic metrics namespace, but we already do have them in a lot of places. If this solution serves the need now, let's leave it like this and keep metric specialization in mind for future iterations.

PR #9875, opened 2021-06-01
Dpp 392 todo cleanup postgre batch

nmarton-da:
### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGSchema.scala
```diff
@@ -0,0 +1,142 @@
...
+    "ledger_offset" -> PGString(_.ledger_offset),
+    "recorded_at" -> PGTimestamp(_.recorded_at),
+    "submission_id" -> PGString(_.submission_id.orNull),
+    "typ" -> PGString(_.typ),
+    "rejection_reason" -> PGString(_.rejection_reason.orNull),
+  )
+
+  val packages: PGTable[DBDTOV1.Package] = PGTable(
+    tableName = "packages",
+    insertSuffix = "on conflict (package_id) do nothing",
```
tudor-da, line 100: This insert conflict action clause is not a property of a table but rather of a statement. I see this is also the only place where it is non-empty. I'd remove it from `PGTable` if it doesn't complicate code in other places much

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGTable.scala
```diff
@@ -0,0 +1,49 @@
...
+package com.daml.platform.store.backend.postgresql
+
+import java.sql.PreparedStatement
+
+case class PGTable[FROM](
+    tableName: String,
+    fields: Vector[(String, PGField[FROM, _, _])],
+    insertSuffix: String = "",
+) {
+  val insertStatement: String = {
```
tudor-da, line 13: Where is this used?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGTable.scala
```diff
@@ -0,0 +1,49 @@
...
+       |   unnest($unnestFields)
+       | as t($inputFields)
+       | $insertSuffix
+       |""".stripMargin
+  }
+
+  def prepareData(in: Vector[FROM]): Array[Array[_]] =
+    fields.view.map(_._2.toArray(in)).toArray
+
+  def setupData(data: Array[Array[_]], preparedStatement: PreparedStatement): Unit =
```
tudor-da, line 40: Where is this used?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresDbBatch.scala
```diff
@@ -0,0 +1,49 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.backend.postgresql
+
+import com.daml.platform.store.backend.DBDTOV1
+import com.daml.scalautil.NeverEqualsOverride
+
+import scala.reflect.ClassTag
+
+case class PostgresDbBatch(
```
tudor-da, line 11: Can you make these private to `store`?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -11,706 +11,53 @@ import com.daml.platform.store.backend.{DBDTOV1, StorageBackend}
...
-        batch.upload_id,
-        batch.source_description,
-        batch.size,
-        batch.known_since,
-        batch.ledger_offset,
-        batch._package,
-      )
-    )
+    def executeTable(pgTable: PGTable[_], data: Array[Array[_]]): Unit =
+      if (data(0).length > 0) execute(pgTable.insertStatement, pgTable.setupData(data, _))
```
tudor-da, line 40: Why isn't `data(0)` throwing?

tudor-da: Or is `data(0)` some sort of header that's always present?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGTable.scala
tudor-da: Ah, nevermind. Intellij was playing tricks with me

mziolekda (approved):
Looks good!

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGSchema.scala
nmarton-da: I agree in general, but here in this DSL PGTable's responsibility is to:
* provide method to create the insert SQL for preparedstatements (val insertStatement)
* provide the method to assemble batch arrays (def prepareData)
* provide method to set up data preparedStatement data from built batch arrays (def setupData)

We could of course walk some more lengths to have these responsibilities separated somehow, currently it seemed to me the right level of abstraction, but please hit me with ideas if you can think something better!

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresDbBatch.scala
nmarton-da: yeah, good idea, I'll seal off visibility for the entire package

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
nmarton-da: the meaning of this two dimensional array: 
* 1st dimension (outer array): table columns
* 2nd dimension: to-be-inserted data for those table columns (this will be unnest-ed)

so data(0) means: the array of data to be inserted for the first column
note: if data(0) would be out of bonds, that would mean we have a table setup without columns
note2: all 2nd dimension arrays are of the same length (transposed structure)

:thinking: actually I will add a comment to explain why it is safe, good point

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresDbBatch.scala
nmarton-da: actually since the existential type on the factory this can be also private[postgresql]

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGField.scala
```diff
@@ -0,0 +1,98 @@
...
+import java.time.format.DateTimeFormatter
+
+import scala.reflect.ClassTag
+
+/** Type parameters TO and CONVERTED need to comply to the following contract:
+  * - either both of them nullable (AnyRef)
+  * - or neither of them nullable (CONVERTED AnyVal, TO AnyVal or AnyRef which will be never null)
+  * This is important, since absence is automatically signaled with null references (and we still would want to support primitives)
+  */
+private[postgresql] sealed abstract class PGField[FROM, TO, CONVERTED](implicit
```
rautenrieth-da, line 17: I took me a while to understand the type parameters here. Is the following correct?
- `FROM`: an arbitrary type. The database field value can somehow be extracted from these objects.
- `TO`: a type describing the content of the database field
- `CONVERTED`: the actual SQL type of the database field

For example, a field holding a list of parties would use `TO = List[String]` (because this is how you would store a list of parties in Scala) and `FROM = String` (because what we actually store in the SQL schema is the concatenation of all party strings).

A small description might be useful.

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -785,7 +135,6 @@ object PostgresStorageBackend extends StorageBackend[RawDBBatchPostgreSQLV1] {
       |FROM package_entries
       |WHERE ledger_offset > ?;
       |
-      |-- TODO append-only: we do not have currently an index to support efficiently this operation. either add or make sure it is okay to do full table scans here
```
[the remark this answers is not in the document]
nmarton-da: yes, I think it is okay to not add indexes just for initialization purposes for a relatively small table.
But you are right, let's get some more broad approval here!
@meiersi-da @gerolf-da @mziolekda do you agree?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGField.scala
nmarton-da: yeah unfortunately all these 3 types needed for the purpose, since we want to have one part of the conversion in the type, and another in the definition.

I will beef up the scaladoc in this area

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGTable.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import java.sql.PreparedStatement
+
+private[postgresql] case class PGTable[FROM](
+    tableName: String,
+    fields: Vector[(String, PGField[FROM, _, _])],
+    insertSuffix: String = "",
+) {
+  val insertStatement: String = {
+    def comaSeparatedOf(extractor: ((String, PGField[FROM, _, _])) => String): String =
```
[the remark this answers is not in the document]
nmarton-da: yepp, typo, thanks

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGSchema.scala
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.backend.postgresql
+
+import com.daml.platform.store.backend.DBDTOV1
+
+private[postgresql] object PGSchema {
+  val eventsDivulgence: PGTable[DBDTOV1.EventDivulgence] = PGTable("participant_events_divulgence")(
+    "event_offset" -> PGString(_.event_offset.orNull),
```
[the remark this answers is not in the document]
nmarton-da: I like this idea a lot: keeping the extractor part scala-esc, implementation simpler on the tradeoff to have some prolifiration in the PG types :+1:

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGField.scala
```diff
@@ -0,0 +1,98 @@
...
+  def selectFieldExpression(inputFieldName: String): String = inputFieldName
+
+  final def toArray(input: Vector[FROM]): Array[CONVERTED] =
+    input.view
+      .map(extract)
+      .map {
+        case null =>
+          null.asInstanceOf[CONVERTED] // this is safe if clients comply with the contract above
+        case notNull => convert(notNull)
+      }
```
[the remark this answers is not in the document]
nmarton-da: good point, I will rewamp this part :+1:

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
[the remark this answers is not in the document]
meiersi-da: Let's compare the cost and risk profile of the two options:
1. add index: low implementation cost and risk (one SQL statement, one manual test to check query plan), low runtime cost as indices work well in DB's
2. no index: no implementation cost and risk, medium operational risk, as a customer might discover slow startup on a ledger having ingested lots of packages.

It seems to me that Option 1 offers the better trade-offs, and follow the mantra of prudent engineering and optimizes for a smooth operational experience of our customers.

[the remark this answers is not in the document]
nmarton-da: hm :thinking: so already multiple opinions, that is fine, created task to mediate, capture and implement a decision here for the six cases: https://digitalasset.atlassian.net/browse/DPP-428

PR #9881, opened 2021-06-02
Fix Windows test skipping

cocreature:
= doesn’t seem to work. -ceq is case insensitive comparison. I think
-eq would also work here but I see no point in worrying about case.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -59,6 +59,6 @@ bazel build `-`-experimental_execution_log_file ${ARTIFACT_DIRS}/logs/build_exec
 
 bazel shutdown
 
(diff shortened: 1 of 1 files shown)


garyverhaegen-da (approved):
Does that mean we've been skipping all Windows tests all the time ever since that got merged? 🤔

cocreature:
> Does that mean we've been skipping all Windows tests all the time ever since that got merged? thinking

Nope, it seems to always evaluate to `false` luckily rather than `true`

PR #9888, opened 2021-06-02
Introduce metrics in the http-json service

realvictorprm:
Just adds metrics to the http json service.

Should we track the metrics for the websocket endpoint with the same timer as we do track the http endpoint? I seperated them as they are different protocols but I'm happy to change it if one thinks it's better to combine this.

changelog_begin
[HTTP/JSON API]
- metrics reporting can now be chosen via the command line option --metrics-reporter, valid options are console, csv://, graphite:// and prometheus://
- metrics reporting interval can also now be chosen via a command line option --metrics-reporting-interval
changelog_end

### Pull Request Checklist

trigger the build.


ledger-service/http-json/BUILD.bazel
```diff
@@ -65,11 +65,18 @@ hj_scalacopts = lf_scalacopts + [
             "//ledger-service/utils",
             "//ledger/ledger-api-auth",
             "//ledger/ledger-api-common",
+            "//ledger/ledger-resources",
+            "//ledger/sandbox-common:sandbox-common-{}".format(edition),
```
cocreature, line 69: how did we end up pulling in `sandbox-common`? That seems off

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -90,10 +92,10 @@ class Endpoints(
       case req @ HttpRequest(GET, Uri.Path("/v1/packages"), _, _, _) =>
         (implicit lc => httpResponse(listPackages(req)))
       // format: off
```
cocreature, line 94: What happens if we just turn on formatting here?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -103,16 +105,16 @@ class Endpoints(
...
-          _ = logger.trace(s"Incoming request on ${req.uri}")
-          t0 = System.nanoTime()
-          res <- lcFhr(lc)
-          _ = logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
-        } yield res
-      )
+      extendWithRequestIdLogCtx(implicit lc => {
+        val t0 = System.nanoTime
+        logger.trace(s"Incoming request on ${req.uri}")
+        Timed
```
cocreature, line 111: I think this is perfectly fine to setup the infrastructure in this PR but in terms of general metrics, I think we need to be more specific. There is no meaningful correlation between times on queries, package downloads or command submissions which is usually an indicator that we need more specific metrics. I think we can address all of that in a separate PR provided we hide the option to avoid having backwards compat issues around this one.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -18,12 +18,17 @@ import scalaz.std.option._
 import scalaz.syntax.show._
 import com.daml.cliopts.{GlobalLogLevel, Logging}
 import com.daml.http.util.Logging.{InstanceUUID, instanceUUIDLogCtx}
+import com.daml.ledger.resources.ResourceContext
 import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
 
+import com.daml.platform.sandbox.metrics.MetricsReporting
```
cocreature, line 24: Ah I guess that’s what pulls in `sandbox-common`. Maybe we should move that to a separate package at some point. Doesn’t seem specific to sandbox.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -113,14 +128,20 @@ object Main {
...
+          startSettings = config,
+          contractDao = contractDao,
+        )
       )
+    }
 
     discard {
       sys.addShutdownHook {
+        metricsResource
+          .release()
```
cocreature, line 143: In line 149 we call `terminate` which also releases the resource. Are we releasing it twice now?

cocreature (approved):
Nice work!
As mentioned in the comments, I think this sets up the right infrastructure so we can next focus on the actual metrics. A few that come to mind that I think are useful:
1. Concurrent requests: This should probably be split into 3 or maybe 4 metrics: concurrent command submissions, concurrent queries, concurrent websocket connections and maybe one for party & package allocation.
2. Total duration of requests. This is what you already added. I think we should at least split that in command submissions and queries and lastly party/package management. Not quite sure what we actually measure for websocket connections in this PR. The total duration there doesn’t seem all that interesting given that they run forever until the client closes the connection.  
3. More fine grained durations. A few examples:
    1. Validation & conversion of the incoming JSON values.
    2. For command submissions the duration of the request to the ledger.
    3. Conversion on the outgoing side.
    4. Database operations at least split into reads & writes but possibly more fine-grained.

This is not intended to be an exhaustive list and I expect we’ll expand it over time but maybe a good starting point.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
victormueller-da: Now we're speaking of it, I think yes.
I'm not particularly fond of the resource freeing here but without different handling of the shutdown hook this needs to stay like it is.

realvictorprm: (oh what have I done, now I'm here twice :facepalm:)

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -682,5 +682,12 @@ final class Metrics(val registry: MetricRegistry) {
         val prune: Timer = registry.timer(Prefix :+ "prune")
       }
     }
+
+    object http_json_api {
+      private val Prefix: MetricName = daml.Prefix :+ "http_json_api"
+
+      val httpRequest: Timer = registry.timer(Prefix :+ "http_request")
+      val websocketRequest: Timer = registry.timer(Prefix :+ "websocket_request")
+    }
```
stefanobaghino-da, line 691: The metrics have been collected here to make it easier to track and document them for the participant server.

For separate components (like the HTTP-JSON API) we should follow the same approach but use a different file.

For example, [this](https://github.com/digital-asset/daml/blob/2fa250f06acb3e1a037db94037dcce262556080b/runtime-components/non-repudiation/src/main/scala/com/daml/nonrepudiation/Metrics.scala) is the file collecting together the metrics for the non-repudiation middleware.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
victormueller-da: let's try out :boom:

victormueller-da: This:
```
      case req @ HttpRequest(
            GET,
            Uri(
              _,
              _,
              Slash(Segment("v1", Slash(Segment("packages", Slash(Segment(packageId, Empty)))))),
              _,
              _,
            ),
            _,
            _,
            _,
          ) => (
        implicit lc => downloadPackage(req, packageId)
      )
```
Ugly formatting intensifies in my opinion :joy:

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def metricsReporterParse[C](parser: scopt.OptionParser[C])(
+      metricsReporter: Setter[C, Option[MetricsReporter]],
+      metricsReportingInterval: Setter[C, FiniteDuration],
+  ): Unit = {
+    import parser.opt
+
+    opt[MetricsReporter]("metrics-reporter")
+      .action((reporter, config) => metricsReporter(_ => Some(reporter), config))
+      .optional()
```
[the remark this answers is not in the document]
victormueller-da: yeah sounds reasonable

ledger-service/cli-opts/BUILD.bazel
```diff
@@ -27,7 +27,7 @@ da_scala_library(
     ],
     deps = [
         "//ledger/ledger-api-common",
-        "//ledger/participant-integration-api",
```
stefanobaghino-da: This looks very good already.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -91,11 +91,23 @@ class Endpoints(
...
+              Slash(Segment("v1", Slash(Segment("packages", Slash(Segment(packageId, Empty)))))),
+              _,
+              _,
+            ),
+            _,
+            _,
+            _,
+          ) => (
+        implicit lc => downloadPackage(req, packageId)
+      )
```
stefanobaghino-da: Unfortunately the formatter cannot handle this while keeping this readable. I would recommend to revert this and maybe add a comment to let readers know why.

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporter.scala
```diff
@@ -94,12 +93,12 @@ object MetricsReporter {
...
+        throw new RuntimeException(cliHint + " " + exception.getMessage)
     }
 
   val cliHint: String =
     """Must be one of "console", "csv:///PATH", "graphite://HOST[:PORT][/METRIC_PREFIX]", or "prometheus://HOST[:PORT]"."""
 
-  private def invalidRead: InvalidConfigException =
-    new InvalidConfigException(cliHint)
+  private def invalidRead =
+    new RuntimeException(cliHint)
```
stefanobaghino-da, line 103: I guess you had to do this to avoid grabbing a dependency. Maybe have a look if it could make sense to move `InvalidConfigException` here.

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporting.scala
```diff
@@ -45,7 +42,7 @@ final class MetricsReporting(
         .map(_.start())
       _ <- extraMetricsReporter.fold(Resource.unit) { reporter =>
         acquire(reporter.register(registry))
-          .map(_.start(extraMetricsReportingInterval.getSeconds, TimeUnit.SECONDS))
+          .map(_.start(extraMetricsReportingInterval.toNanos, TimeUnit.NANOSECONDS))
```
stefanobaghino-da, line 45: Nanoseconds are _way_ too fine-grained. Seconds is sufficient.

ledger/sandbox-common/src/test/lib/scala/platform/sandbox/cli/CommonCliSpecBase.scala
```diff
@@ -6,21 +6,20 @@ package com.daml.platform.sandbox.cli
...
+import com.daml.metrics.MetricsReporter
+import com.daml.metrics.MetricsReporter.{Graphite, Prometheus}
 import com.daml.platform.sandbox.cli.CommonCliSpecBase._
 import com.daml.platform.sandbox.config.SandboxConfig
 import com.daml.platform.services.time.TimeProviderType
 import com.daml.ports.Port
 import org.scalatest.Assertion
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
+import scala.concurrent.duration._
```
stefanobaghino-da, line 22:
```suggestion
import scala.concurrent.duration.DurationInt
```

Not necessary but I usually prefer to not bring in scope things I don't use, mostly to prevent unwanted implicit conversions to be brought in scope.

stefanobaghino-da (commented):
The latest commit is a welcome improvement. 🙂

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporting.scala
realvictorprm: I agree that it is quite fine grained but is it so bad to have it that like this? Seconds are in my opinion not fine grained enough, would down to Milliseconds be fine? Then the durations can at least be a bit finer than seconds (e.g. 2.5 seconds).

cocreature: This is not the duration though its the reporting interval afaict?

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporter.scala
realvictorprm: I already looked into that and adding a whole exception hierarchy into the metrics project did not seem right.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: did revert that

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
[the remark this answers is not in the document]
realvictorprm: done

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: Will address this with a seperate PR

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporting.scala
realvictorprm: correct, here the input from the cli is handed over to the reporter as interval.

cocreature: Then I don’t understand why you’re worried about subsecond granularity. Of course the indvidual metrics should be able to handle durations that are smaller than a second but reporting more frequently than once per second doesn’t seem useful.

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporter.scala
stefanobaghino-da: Cool, then please don't drop the cause (it made sense to drop it originally because `InvalidConfigException` extends `NoStackTrace`). Alternatively you can simply not re-throw.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
realvictorprm: ok adressed, hope it is fine.

ledger/metrics/src/main/scala/com/daml/metrics/ParticipantMetrics.scala
```diff
@@ -0,0 +1,686 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.metrics
+
+import java.time.Instant
+
+import com.codahale.metrics.MetricRegistry.MetricSupplier
+import com.codahale.metrics._
+
+final class ParticipantMetrics(override val registry: MetricRegistry) extends Metrics(registry) {
```
S11001001, line 11: It would have been a significant sanity improvement to leave this named `Metrics` and place it in the `participant` subpackage, *or* to have every import do `ParticipantMetrics => Metrics` and leave all the existing references alone. We aren't working with Java.

S11001001: ...which would avoid, e.g., the conflict in ParallelIndexerFactory as of 70cca4b.

realvictorprm: Didn't though of just putting it into a different package, thanks for the idea @S11001001 :)!

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
```diff
@@ -0,0 +1,41 @@
...
+    import parser.opt
+
+    def hideIfRequested[A](opt: OptionDef[A, C]): Unit = {
+      if (hide) {
+        opt.hidden()
+        ()
+      }
+    }
+
+    val optMr =
```
miklos-da, line 27: Please use meaningful variable names and not abbreviations.

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
```diff
@@ -0,0 +1,41 @@
...
+import scopt.OptionDef
+
+import scala.concurrent.duration.FiniteDuration
+
+object Metrics {
+
+  def metricsReporterParse[C](parser: scopt.OptionParser[C])(
+      metricsReporter: Setter[C, Option[MetricsReporter]],
+      metricsReportingInterval: Setter[C, FiniteDuration],
+      hide: Boolean = false,
```
miklos-da, line 16: Consider removing this flag and the accompanying logic as this is never called with `hide = false`.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -3,684 +3,6 @@
```
miklos-da, line 3: Please revert changes to this file and add the metrics you need either here or to a separate class/module. Not all of these metrics are 'participant metrics', and such refactoring should be done in a separate PR. As a bonus, you won't have a PR with 150+ files changed :)

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporter.scala
realvictorprm: OK, I propagated the exception into the RuntimeException with the existing msg. I hope this is sufficient

miklos-da (changes requested):

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: As proposed, will do this in a seperate PR.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
realvictorprm: I now have confliciting messages about how the metrics are handled.

@stefanobaghino-da told me to move the metrics for the json api out into a seperate class/file.

Because I wanted to reuse the MetricsReporting class I abstracted the metrics definition and moved the content in here into a seperate class/file.

I can revert all this but this will result in code duplication so far I see the situation right now.

PS: I'm not that fond of refactoring 150 files either but sometimes refactoring results in that and so PR's with 150+ files changed will exist.

realvictorprm:
I will start over another time

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
cocreature: Refactoring is fine, mixing it up with actual changes is usually less good of an idea. You make your life harder (since it’s usually harder to get those changes merged) and that of reviewers. E.g., this PR adds metrics to the JSON API, if we want to shuffle something else around for that, make a PR which does nothing but that, get that merged and then follow up with the actual change.

For changes to `Metrics.scala`, I would suggest to keep that file either completely untouched and only add the JSON API metrics in a separate file or add the JSON API metrics in there addressing @stefanobaghino-da’s comment separately.

cocreature:
> I will start over another time
I don’t think there is any need for that! Let’s have a chat about how we can move this forward.

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
realvictorprm: This logic exists to hide the option in the http json api.

miklos-da: Thanks for the clarification, I seem to have missed that.

PR #9892, opened 2021-06-02
Include PackageMetadata in LF interface

cocreature:
I want to use this in codegens but more generally, I think it also
fits well into the scope of the iface library in that it’s only
metadata and relatively stable.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/EnvironmentInterface.scala
+++ b/daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/EnvironmentInterface.scala
@@ -5,26 +5,34 @@ package com.daml.lf
 package iface
 
 import com.daml.lf.archive.Dar
-import data.Ref.Identifier
+import data.Ref.{Identifier, PackageId}
 
 import scala.collection.immutable.Map
 import scalaz.syntax.std.map._
 import scalaz.Semigroup
 
 /** The combination of multiple [[Interface]]s, such as from a dar. */
-final case class EnvironmentInterface(typeDecls: Map[Identifier, InterfaceType])
+final case class EnvironmentInterface(
+    metadata: Map[PackageId, PackageMetadata],
+    typeDecls: Map[Identifier, InterfaceType],
+)
 
 object EnvironmentInterface {
   def fromReaderInterfaces(i: Interface, o: Interface*): EnvironmentInterface =
-    EnvironmentInterface((i +: o).iterator.flatMap { case Interface(packageId, typeDecls) =>
-      typeDecls mapKeys (Identifier(packageId, _))
-    }.toMap)
+    (i +: o).foldLeft(EnvironmentInterface(Map.empty, Map.empty)) {
+      case (acc, Interface(packageId, optMetadata, typeDecls)) =>
+        acc.copy(
+          metadata =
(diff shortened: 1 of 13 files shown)


daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/Interface.scala
```diff
@@ -42,12 +42,22 @@ object InterfaceType {
   }
 }
 
+// Duplicate of the one in com.daml.lf.language to separate Ast and Iface
```
cocreature, line 45: If someone tells me that it’s fine to just expose parts of the `Ast`, I’m also perfectly happy with that. I don’t know what our policy here is.

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/EnvironmentInterface.scala
```diff
@@ -5,26 +5,34 @@ package com.daml.lf
...
-      typeDecls mapKeys (Identifier(packageId, _))
-    }.toMap)
+    (i +: o).foldLeft(EnvironmentInterface(Map.empty, Map.empty)) {
+      case (acc, Interface(packageId, optMetadata, typeDecls)) =>
+        acc.copy(
+          metadata =
+            optMetadata.fold(acc.metadata)(metadata => acc.metadata.updated(packageId, metadata)),
+          typeDecls = acc.typeDecls ++ typeDecls.mapKeys(Identifier(packageId, _)),
+        )
+    }
```
S11001001, line 29: I'd rather iterate over `i +: o` twice.

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/Interface.scala
S11001001: I think this is fine. Maybe we'll add a type parameter later :trollface:

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/EnvironmentInterface.scala
cocreature: why? Doing it in one go seems simple enough and should be faster

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/Interface.scala
cocreature: everything gets better if you add another type parameter.

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/EnvironmentInterface.scala
cocreature: oh didn’t realize updating maps was so much more expensive. changed 9dd1a3f91d29a03567eec4f4746632a70d37faf6

S11001001: I'm not generally opposed to updating maps iteratively, if it's clearer, but when I do, I try to use `scalaz.==>>` instead, which is guaranteed to behave like `Data.Map.Strict`.

S11001001 (approved):

PR #9896, opened 2021-06-02
DPP-387 Remove migration todos

rautenrieth-da:
This PR removes two TODOs as `wontfix`, and addresses a third one by reordering table columns to avoid unnecessary padding.


nmarton-da (approved):
Looks good!

PR #9909, opened 2021-06-03
Use simpler tree representation for transactions when possible

nickchapman-da:
WIP/Prototype

This work was initially driven for perf reasons. And there is about a 1% improvement. But it turns out that some tricky and fragile code in `PartialTransaction` which managed and used `NodeId` has been removed.

Use simpler tree representation for transactions when possible:
- add simpler tree representation for transactions (`TxTree`, `TxTreeTop`)
- have the rollback normalizer take/consume this rep
- support conversion from this simpler rep to the standard tx rep ( `toStandardTransaction`)

Change speedy `PartialTransaction` to construct this simpler rep:
- remove all node-id tracking
- ensure collection of `actionNodeSeeds` still works
- remove use of `NodeId` in some daml exceptions
- All tests `bazel test daml-lf/...` are working

TODO:
- some work still required to adapt `scenario-service`

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/bench.sh
+++ b/bench.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+wi=5
+i=25
+
+whatBranch() {
+    git rev-parse --abbrev-ref HEAD
+}
+
+bench() {
+    git checkout $1
+    bazel run //daml-lf/scenario-interpreter:scenario-perf -- -f 1 -i $i -wi $wi
+}
+
+run() {
+    bench $1 2>&1 | grep 'Average]' | xargs echo $1
+}
+
+here=$(whatBranch)
+base=main
+
+run $here
+run $here
+
+run $base
+run $base
+
+run $here
+run $here
+
+run $base
+run $base
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -13,6 +13,7 @@ import com.daml.lf.speedy.Speedy.Machine
 import com.daml.lf.speedy.SResult._
 import com.daml.lf.transaction.{SubmittedTransaction, Transaction => Tx}
 import com.daml.lf.transaction.Node._
+import com.daml.lf.transaction.NodeId
 import com.daml.lf.value.Value
 import java.nio.file.Files
 
@@ -301,8 +302,8 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
         case SResultError(err) =>
           return ResultError(
             Error(
-              s"Interpretation error: ${Pretty.prettyError(err, onLedger.ptx).render(80)}",
-              s"Last location: ${Pretty.prettyLoc(machine.lastLocation).render(80)}, partial transaction: ${onLedger.ptx.nodesToString}",
+              s"Interpretation error: ${Pretty.prettyError(err).render(80)}",
+              s"Last location: ${Pretty.prettyLoc(machine.lastLocation).render(80)}", //, partial transaction: ${onLedger.ptx.nodesToString}", //NICK
             )
           )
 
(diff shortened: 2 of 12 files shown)


bench.sh
```diff
@@ -0,0 +1,36 @@
...
+run $here
+
+run $base
+run $base
+
+run $here
+run $here
+
+run $base
+run $base
```
remyhaemmerle-da, line 36: I think this file should not be in the root directory.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -370,7 +371,12 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
           submissionTime = onLedger.ptx.submissionTime,
           usedPackages = Set.empty,
           dependsOnTime = onLedger.dependsOnTime,
-          nodeSeeds = onLedger.ptx.actionNodeSeeds.toImmArray,
+          //nodeSeeds = onLedger.ptx.xactionNodeSeeds.toImmArray //NICK
+          nodeSeeds = ImmArray(
+            onLedger.ptx.actionNodeSeeds.toImmArray.toList.zipWithIndex.map { case (v, i) =>
+              (NodeId(i), v)
+            }
```
remyhaemmerle-da, line 378:
```suggestion
nodeSeeds = onLedger.ptx.actionNodeSeeds.toImmArray.toSeq.zipWithIndex.map { case (v, i) => (NodeId(i), v) }.toImmArray
```

The method `toSeq` on `ImmArray` returns you a  wrapper around ImmArray with `Seq` API. So you avoid the conversion from `ImmArray` to `List` and back.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
```diff
@@ -78,14 +62,14 @@ private[lf] object Pretty {
         text("Update failed due to fetch of an inactive contract") & prettyContractId(coid) &
           char('(') + (prettyTypeConName(tid)) + text(").") /
           text(s"The contract had been consumed in sub-transaction #$consumedBy:") +
```
remyhaemmerle-da, line 64: You have change quite a bit the type of `consumedBy`, so I think you have change quite a bit the output string.
This is probably not tested :weary:

remyhaemmerle-da: We need test for that, since it used in the scenario service and then in the IDE.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
```diff
@@ -78,14 +62,14 @@ private[lf] object Pretty {
...
-          (ptx.nodes.get(consumedBy) match {
+          (consumedBy match {
             // FIXME(JM): How should we show this? If the node pointed to by consumedBy
             // is not in nodes, it must not be done yet, hence the contract is recursively
             // exercised.
```
remyhaemmerle-da, line 68: > FIXME(JM): How should we show this?
When an exercise is opened but not closed.
> If the node pointed to by consumedBy is not in nodes, it must not be done yet, hence the contract is recursively exercised.

No, if the body of an exercise fails, the Node Id of the the exercise was not added to `ptx.node`.
On the other hand a "recursive exercised" can never happen, because the contract is consumed at the beginning of an exercise before its body is evaluated.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
```diff
@@ -78,14 +62,14 @@ private[lf] object Pretty {
...
             case None =>
               (line + text("Recursive exercise of ") + prettyTypeConName(tid)).nested(4)
```
remyhaemmerle-da, line 70: This is non-nonsensical, it should be changed, probably in a other PR thought.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SError.scala
```diff
@@ -67,7 +67,7 @@ object SError {
   final case class DamlELocalContractNotActive(
       coid: ContractId,
       templateId: TypeConName,
-      consumedBy: NodeId,
+      consumedBy: Option[PartialTransaction.Tree],
```
remyhaemmerle-da, line 70: The Option here is a bit ugly.
I think the info carries by the domain of the map was before not very useful.
In fact I am pretty sure this is useful only in the pretty printer to "explain" why a transaction failed.

I believe the Map can be replaced by a Set
```suggestion
      consumed: Set[Value.ContractId],
```
the you can recompute which node consumed the contract directly from the partially transaction when you need it (you just look for the first consuming node that exercised the contract)

remyhaemmerle-da (commented):
This looks reasonable.
There are issues with the Error pretty printer so.

PR #9920, opened 2021-06-08
skip fix_bazel_cache on forks

garyverhaegen-da:
Secrets are not available to forks, so they can't delete things from the
cache.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
looks very plausible but also it’s azure so who knows, thanks!

PR #9923, opened 2021-06-08
Add metrics to the http json service

realvictorprm:
### Pull Request Checklist

Just adds metrics to the http json service.

Also moves out some metrics related stuff to the Metrics project so the http json service does not depend on unrelated services.

changelog_begin

[HTTP/JSON API]
- metrics reporting can now be chosen via the command line option --metrics-reporter (currently hidden), valid options are console, csv://, graphite:// and prometheus://
- metrics reporting interval can also now be chosen via a command line option --metrics-reporting-interval (currently hidden)

changelog_end

Minimised version of try 2.

trigger the build.


ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -682,5 +682,12 @@ final class Metrics(val registry: MetricRegistry) {
         val prune: Timer = registry.timer(Prefix :+ "prune")
       }
     }
+
+    object http_json_api {
```
miklos-da, line 686: Please don't use underscore in class or object names.
```suggestion
    object httpJsonApi {
```

ledger/sandbox-common/src/main/scala/platform/sandbox/cli/CommonCliBase.scala
```diff
@@ -309,6 +300,11 @@ class CommonCliBase(name: LedgerName) {
           s"By default compression is off, this switch enables it. This has only effect for append-only ingestion." // TODO append-only: fix description
         )
 
+      com.daml.cliopts.Metrics.metricsReporterParse(this)(
+        (f, c) => c.copy(metricsReporter = f(c.metricsReporter)),
```
miklos-da, line 304: To be consistent with the style already established, please don't use abbreviations:
```suggestion
        (setter, config) => config.copy(metricsReporter = setter(config.metricsReporter)),
```

miklos-da (changes requested):
Thanks for greatly simplifying the original PR #9888! Just requesting minor changes, otherwise LGTM.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
realvictorprm: Adressed with 169de2eb1b310000636dac82e4e4e72da476044c

ledger/metrics/src/main/scala/com/daml/metrics/MetricsReporter.scala
```diff
@@ -89,17 +89,17 @@ object MetricsReporter {
...
+  val cliHint: String =
+    """Must be one of "console", "csv:///PATH", "graphite://HOST[:PORT][/METRIC_PREFIX]", or "prometheus://HOST[:PORT]"."""
+
   def parseUri(value: String): URI =
     try {
       new URI(value)
     } catch {
       case NonFatal(exception) =>
-        throw new InvalidConfigException(cliHint + " " + exception.getMessage)
+        throw new RuntimeException(cliHint, exception)
```
cocreature, line 100: Seems like a reasonable change, nobody should be catching this anyway.

cocreature (approved):
Nice, thank you!

miklos-da (approved):

ledger/sandbox-common/src/test/lib/scala/platform/sandbox/cli/CommonCliSpecBase.scala
```diff
@@ -268,8 +268,8 @@ abstract class CommonCliSpecBase(
 
     "parse the metrics reporting interval when given" in {
       checkOption(
-        Array("--metrics-reporting-interval", "PT1M30S"),
-        _.copy(metricsReportingInterval = Duration.ofSeconds(90)),
+        Array("--metrics-reporting-interval", "1.5m"),
```
realvictorprm, line 271: @cocreature Is this cli option change OK? This is due to the change from Java Duration to scala duration

cocreature: This seems like a breaking change. Can we continue accepting a Java duration here? Or alternatively accept both preferring a Java duration. We can still convert to a Scala duration at the very beginning if that helps.

cocreature: You should be able to use a custom reader that tries one first and if that fails tries the other.

realvictorprm: Yeah just wrote one, I will make another PR as I forgot to remove the label here.

PR #9926, opened 2021-06-08
Allow two different time formats as input for the metrics reporting interval

realvictorprm:
and accordingly revert to using the old test for the CommonCliSpecBase in sandbox-common.

changelog_begin

- for applications which support the --metrics-reporter-interval cli option, these now support both the java and scala duration string format (e.g. PT1M30S and 1.5min)

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
+++ b/ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
@@ -6,9 +6,28 @@ package com.daml.cliopts
 import com.daml.metrics.MetricsReporter
 import scopt.OptionDef
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{Duration, FiniteDuration, NANOSECONDS}
+import scala.util.Try
 
 object Metrics {
+  private case class DurationFormat(unwrap: FiniteDuration)
+
+  private implicit val scoptDurationFormat: scopt.Read[DurationFormat] = scopt.Read.reads {
+    duration =>
+      Try {
+        Duration(duration)
+      }.orElse(Try {
+        Duration.fromNanos(
+          java.time.Duration.parse(duration).toNanos
+        )
+      }).flatMap(duration =>
+        Try {
+          if (!duration.isFinite)
+            throw new RuntimeException(s"Input duration $duration is not finite")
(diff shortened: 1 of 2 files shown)


ledger/sandbox-common/src/test/lib/scala/platform/sandbox/cli/CommonCliSpecBase.scala
```diff
@@ -268,7 +268,7 @@ abstract class CommonCliSpecBase(
 
     "parse the metrics reporting interval when given" in {
       checkOption(
-        Array("--metrics-reporting-interval", "1.5m"),
+        Array("--metrics-reporting-interval", "PT1M30S"),
```
cocreature, line 271: Probably a good idea to add a test for both formats.

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
```diff
@@ -6,9 +6,28 @@ package com.daml.cliopts
...
 object Metrics {
+  private case class DurationFormat(unwrap: FiniteDuration)
+
+  private implicit val scoptDurationFormat: scopt.Read[DurationFormat] = scopt.Read.reads {
+    duration =>
+      Try {
+        Duration(duration)
+      }.orElse(Try {
+        Duration.fromNanos(
+          java.time.Duration.parse(duration).toNanos
```
cocreature, line 21: Maybe add a comment that this is for backwards compat. I’d also be slightly more comfortable trying the java version first. I think there is no collision where something parses as both but if there is we want to prefer Java.

ledger-service/cli-opts/src/main/scala/cliopts/Metrics.scala
```diff
@@ -6,9 +6,28 @@ package com.daml.cliopts
...
+        )
+      }).flatMap(duration =>
+        Try {
+          if (!duration.isFinite)
+            throw new RuntimeException(s"Input duration $duration is not finite")
```
cocreature, line 26:
```suggestion
            throw new IllegalArgumentException(s"Input duration $duration is not finite")
```
seems to match what scopt throws itself

cocreature (approved):
Thanks for the quick fix!

PR #9929, opened 2021-06-08
Handle module-prefixes in Java codegen

cocreature:
changelog_begin

- [Java Codegen] The Java codegen will now pick up the
  `module-prefixes` field from `daml.yaml` which can be used to handle
  module name collisions between different DALFs.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/CodegenConfigReader.scala
+++ b/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/CodegenConfigReader.scala
@@ -7,8 +7,9 @@ import java.io.File
 import java.nio.file.{Path, Paths}
 
 import ch.qos.logback.classic.Level
+import com.daml.lf.data.Ref.{PackageName, PackageVersion}
 import com.daml.assistant.config._
-import io.circe.ACursor
+import io.circe.{ACursor, KeyDecoder}
 
 import scala.util.Try
 
@@ -30,6 +31,7 @@ object CodegenConfigReader {
     for {
       dar <- darPath(sdkConf)
       packagePrefix <- packagePrefix(sdkConf, dest)
+      modulePrefixes <- modulePrefixes(sdkConf)
       outputDirectory <- outputDirectory(sdkConf, dest)
       decoderPkgAndClass <- decoderPkgAndClass(sdkConf, dest)
       verbosity <- verbosity(sdkConf, dest): Result[Option[Int]]
@@ -37,6 +39,7 @@ object CodegenConfigReader {
       root <- root(sdkConf, dest): Result[Option[List[String]]]
     } yield Conf(
       darFiles = Map(dar -> packagePrefix),
+      modulePrefixes = modulePrefixes,
       outputDirectory = outputDirectory,
       decoderPkgAndClass = decoderPkgAndClass,
       verbosity = logLevel,
@@ -71,6 +74,14 @@ object CodegenConfigReader {
     codegen(sdkConf, mode)
       .downField("package-prefix")
       .as[Option[String]]
+      .left
+      .map(configParseError)
+
+  private def modulePrefixes(sdkConf: ProjectConfig): Result[Map[PackageReference, String]] =
+    sdkConf.content.hcursor
+      .downField("module-prefixes")
+      .as[Option[Map[PackageReference, String]]]
+      .map(_.getOrElse(Map.empty))
       .left
       .map(configParseError)
 
@@ -145,4 +156,20 @@ object CodegenConfigReader {
 
   private def resultR[A](a: A): Result[A] =
     Right(a): Result[A]
+
+  implicit val decodePackageReference: KeyDecoder[PackageReference] =
+    new KeyDecoder[PackageReference] {
+      // TODO (MK) For now we only allow name-vesion pairs to match the compiler. Once the compiler
+      // accepts package ids we can allow for those here as well.
+      final def apply(key: String): Option[PackageReference] = nameVersion(key)
+
+      private def nameVersion(key: String): Option[PackageReference] = key.split("-") match {
+        case Array(name, version) =>
+          for {
+            name <- PackageName.fromString(name).toOption
+            version <- PackageVersion.fromString(version).toOption
+          } yield PackageReference.NameVersion(name, version)
+        case _ => None
+      }
+    }
 }
--- a/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/CodegenConfigReader.scala
+++ b/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/CodegenConfigReader.scala
@@ -159,7 +159,8 @@ object CodegenConfigReader {
 
   implicit val decodePackageReference: KeyDecoder[PackageReference] =
     new KeyDecoder[PackageReference] {
-      // TODO (MK) For now we only allow name-vesion pairs to match the compiler. Once the compiler
+      // TODO (MK) https://github.com/digital-asset/daml/issues/9934
+      // For now we only allow name-vesion pairs to match the compiler. Once the compiler
       // accepts package ids we can allow for those here as well.
       final def apply(key: String): Option[PackageReference] = nameVersion(key)
 
--- a/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/Conf.scala
+++ b/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/Conf.scala
@@ -13,7 +13,8 @@ import scopt.{OptionParser, Read}
 sealed trait PackageReference extends Product with Serializable
 
 object PackageReference {
-  // TODO (MK): We probably want to allow package id references here
+  // TODO (MK) https://github.com/digital-asset/daml/issues/9934
+  // We probably want to allow package id references here
   // but this needs to be supported in damlc first.
   final case class NameVersion(name: PackageName, version: PackageVersion)
       extends PackageReference {
--- a/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/Conf.scala
+++ b/language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/Conf.scala
@@ -7,7 +7,20 @@ import java.nio.file.{Path, Paths}
 
 import ch.qos.logback.classic.Level
 import com.daml.buildinfo.BuildInfo
+import com.daml.lf.data.Ref.{PackageName, PackageVersion}
 import scopt.{OptionParser, Read}
+
+sealed trait PackageReference extends Product with Serializable
+
+object PackageReference {
+  // TODO (MK): We probably want to allow package id references here
+  // but this needs to be supported in damlc first.
+  final case class NameVersion(name: PackageName, version: PackageVersion)
+      extends PackageReference {
+    override final def toString(): String =
+      s"$name-$version"
+  }
+}
 
 /** @param darFiles The [[Set]] of Daml-LF [[Path]]s to convert into code. It MUST contain
   *                        all the Daml-LF packages dependencies.
@@ -17,6 +30,7 @@ import scopt.{OptionParser, Read}
 final case class Conf(
     darFiles: Map[Path, Option[String]] = Map(),
     outputDirectory: Path,
+    modulePrefixes: Map[PackageReference, String] = Map.empty,
     decoderPkgAndClass: Option[(String, String)] = None,
     verbosity: Level = Level.ERROR,
     roots: List[String] = Nil,
--- a/language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
+++ b/language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
@@ -171,11 +171,9 @@ object CodeGenRunner extends StrictLogging {
     }.toList
     allModules.groupBy(_._1).foreach { case (m, grouped) =>
       if (grouped.length > 1) {
-        val pkgIds = grouped.map(_._2)
+        val pkgIds = grouped.view.map(_._2).mkString(", ")
         throw new IllegalArgumentException(
-          s"""Duplicate module $m found in multiple packages ${pkgIds.mkString(
-            ", "
-          )}. To resolve the conflict rename the modules in one of the packages via `module-prefixes`."""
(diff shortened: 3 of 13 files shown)


language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/Conf.scala
```diff
@@ -7,8 +7,21 @@ import java.nio.file.{Path, Paths}
...
 import com.daml.buildinfo.BuildInfo
+import com.daml.lf.data.Ref.{PackageName, PackageVersion}
 import scopt.{OptionParser, Read}
 
+sealed trait PackageReference extends Product with Serializable
+
+object PackageReference {
+  // TODO (MK): We probably want to allow package id references here
+  // but this needs to be supported in damlc first.
+  final case class NameVersion(name: PackageName, version: PackageVersion)
```
cocreature, line 18: The indirection here might be a bit overkill. I started having both ids and name/version pairs and dropped the former once I realized that this breaks because the compiler won’t pick it up. Given that I expect we want to add it I kept the indirection but also happy to remove it if someone prefers.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,79 @@ object CodeGenRunner extends StrictLogging {
...
+        }
+        k -> prefix
+      }.toMap
+    resolvedPrefixes
+  }
+
+  /** Verify that no two module names collide when the given
+    * prefixes are applied.
+    */
+  def detectModuleCollisions(
```
cocreature, line 161: I tried to somewhat fit in with the surrounding code which doesn’t seem to be keen on using `Either` for error handling but more than happy to switch to that if someone feels that the inconsistency is better.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
...
+    allModules.groupBy(_._1).foreach { case (m, grouped) =>
+      if (grouped.length > 1) {
+        val pkgIds = grouped.map(_._2)
+        throw new IllegalArgumentException(
+          s"""Duplicate module $m found in multiple packages ${pkgIds.mkString(
+            ", "
+          )}. To resolve the conflict rename the modules in one of the packages via `module-prefixes`."""
+        )
+      }
+    }
```
stefanobaghino-da, line 181:
```suggestion
    for ((m, grouped) <- allModules.groupBy(_._1) if grouped.length > 1) {
      val pkgIds = grouped.valuesIterator.mkString(", ")
      throw new IllegalArgumentException(
        s"Duplicate module $m found in multiple packages $pkgIds. To resolve the conflict rename the modules in one of the packages via `module-prefixes`."
      )
    }
```
Using the `for`-comprehension is probably a matter of taste but I'd recommend to not have interpolations that require curly braces in very long strings, the formatting because a bit broken and makes it less readable.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
...
+      interfaces: Seq[Interface],
+  ): Unit = {
+    val allModules: List[(String, PackageId)] = interfaces.view.flatMap { interface =>
+      val modules = interface.typeDecls.keys.view.map(name => name.module).toSet
+      modules.map(m =>
+        pkgPrefixes
+          .get(interface.packageId)
+          .fold(m.toString)(prefix => s"$prefix.$m") -> interface.packageId
+      )
+    }.toList
```
stefanobaghino-da, line 171:
```suggestion
    val allModules: Seq[(String, PackageId)] =
      for {
        interface <- interfaces
        modules = interface.typeDecls.keySet.map(_.module)
        module <- modules
        maybePrefix = pkgPrefixes.get(interface.packageId)
        prefixedName = maybePrefix.fold(module.toString)(prefix => s"$prefix.$module")
      } yield prefixedName -> interface.packageId
```
* Doesn't go through a view but for the number of items we expect here it's probably fine.
* Uses `.keySet` instead of `.keys ... .toSet`.
* Breaks down the logic in a couple of lines, should be more readable

Mostly a matter of taste I believe, feel free to use it or not.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -118,15 +190,18 @@ object CodeGenRunner extends StrictLogging {
       s"Start processing packageIds '${interfaces.map(_.packageId).mkString(", ")}' in directory '${conf.outputDirectory}'"
     )
 
+    val prefixes = resolvePackagePrefixes(pkgPrefixes, conf.modulePrefixes, interfaces)
+    detectModuleCollisions(prefixes, interfaces)
+
     // TODO (mp): pre-processing and escaping
```
stefanobaghino-da, line 196: The reason why I don't like initials beside `TODO`s is that `git blame` works better and initials tend to become irrelevant over time.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
     logger.warn(s"Finish writing file '$outputFile'")
   }
 
+  /** Given the package prefixes specified per DAR and the module-prefixes specified in
+    * daml.yaml, produce the combined prefixes per package id.
+    */
+  def resolvePackagePrefixes(
```
stefanobaghino-da, line 115: Would it make sense to restrict the visibility of this to either `private` or `private[codegen]` if you need to test this?

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
...
+        }
+        k -> prefix
+      }.toMap
+    resolvedPrefixes
+  }
+
+  /** Verify that no two module names collide when the given
+    * prefixes are applied.
+    */
+  def detectModuleCollisions(
```
stefanobaghino-da, line 160: Would it make sense to restrict the visibility of this to either `private` or `private[codegen]` if you need to test this?

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
...
+          throw new IllegalArgumentException(
+            s"""No package $nameVersion found, available packages: ${metadata.keys.mkString(
+              ", "
+            )}"""
+          ),
+        )
+    }
+    val resolvedModulePrefixes: Map[PackageId, String] = modulePrefixes.view.map { case (k, v) =>
+      resolveRef(k) -> v.toLowerCase
+    }.toMap
```
stefanobaghino-da, line 140:
```suggestion
    val resolvedModulePrefixes: Map[PackageId, String] = modulePrefixes.map { case (k, v) =>
      resolveRef(k) -> v.toLowerCase
    }
```
No need to go through a view for a single transformation where the input and output types are the same, right?

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
```diff
@@ -109,6 +109,78 @@ object CodeGenRunner extends StrictLogging {
...
+            throw new RuntimeException(
+              "Internal error: key in pkgPrefixes and resolvedModulePrefixes could not be found in either of them"
+            )
+          case (Some(a), None) => a.stripSuffix(".")
+          case (None, Some(b)) => b.stripSuffix(".")
+          case (Some(a), Some(b)) => s"""${a.stripSuffix(".")}.${b.stripSuffix(".")}"""
+        }
+        k -> prefix
+      }.toMap
+    resolvedPrefixes
```
stefanobaghino-da, line 154:
```suggestion
    (pkgPrefixes.keySet union resolvedModulePrefixes.keySet).view.map { k =>
      val prefix = (pkgPrefixes.get(k), resolvedModulePrefixes.get(k)) match {
        case (None, None) =>
          throw new RuntimeException(
            "Internal error: key in pkgPrefixes and resolvedModulePrefixes could not be found in either of them"
          )
        case (Some(a), None) => a.stripSuffix(".")
        case (None, Some(b)) => b.stripSuffix(".")
        case (Some(a), Some(b)) => s"""${a.stripSuffix(".")}.${b.stripSuffix(".")}"""
      }
      k -> prefix
    }.toMap
```
Right?

S11001001 (approved):

language-support/codegen-common/src/main/scala/com/digitalasset/daml/lf/codegen/conf/CodegenConfigReader.scala
```diff
@@ -145,4 +156,20 @@ object CodegenConfigReader {
 
   private def resultR[A](a: A): Result[A] =
     Right(a): Result[A]
+
+  implicit val decodePackageReference: KeyDecoder[PackageReference] =
+    new KeyDecoder[PackageReference] {
+      // TODO (MK) For now we only allow name-vesion pairs to match the compiler. Once the compiler
+      // accepts package ids we can allow for those here as well.
```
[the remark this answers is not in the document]
cocreature: Good point, created a ticket and referenced it in 7442a2a18c898f99ef604837bfb3ccfeb19fdfcf

language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
```diff
@@ -54,4 +64,91 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
...
+    ).toMap
+    val interface1 = interface(pkg1, None)
+    val interface2 = interface(pkg2, Some(PackageMetadata(name2, version)))
+    val interface3 = interface(pkg3, Some(PackageMetadata(name3, version)))
+    CodeGenRunner.resolvePackagePrefixes(
+      pkgPrefixes,
+      modulePrefixes,
+      Seq(interface1, interface2, interface3),
+    ) shouldBe
+      Seq(pkg1 -> "com.pkg1", pkg2 -> "com.pkg2.a.b", pkg3 -> "c.d").toMap
```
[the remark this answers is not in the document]
cocreature: You’re not missing anything, I was just being stupid 7f830c0060eeeff2e4a0d8ffd0df44b1346d854d

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
cocreature: 55e100a94f1e004ffc66fd72b0a3cd6d5b18caa0 I tend to avoid for comprehensions for this type of stuff since I read @S11001001’s https://discuss.daml.com/t/pattern-always-adds-withfilter-scala-ritual/1463

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
cocreature: I like it, thanks!

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
cocreature: I find them useful pretty frequently. Yes `git blame` works but it’s an extra step, it’s annoying to track it across refactorings and so on.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
cocreature: 3e2c71c427b43d2959d28fb80ad3767991749daf

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
cocreature: My kingdom for a compiler with proper fusion so I don’t have to worry about this 497eacbc0ff1bda4b15401d3ad7856185b551eae

language-support/java/codegen/src/test/scala/com/digitalasset/daml/lf/codegen/CodeGenRunnerTests.scala
```diff
@@ -54,4 +64,91 @@ class CodeGenRunnerTests extends AnyFlatSpec with Matchers with BazelRunfiles {
...
+        )
+        .toMap,
+    )
+  }
+
+  it should "succeed if there are no collisions" in {
+    CodeGenRunner.detectModuleCollisions(
+      Map.empty,
+      Seq(interface("pkg1", "A", "A.B"), interface("pkg2", "B", "A.B.C")),
+    ) shouldBe (())
```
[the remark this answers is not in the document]
cocreature: switched to asserts in 38a004012523a5d21b6e9593aaedf76f214acfc5. And yes I want it to blow up if we start returning something other than `()`.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
stefanobaghino-da: Good to know, thanks.

language-support/java/codegen/src/main/scala/com/digitalasset/daml/lf/codegen/CodeGenRunner.scala
stefanobaghino-da: There is one but I heard you complaining about segfaults there. :trollface:

PR #9931, opened 2021-06-08
DPP-390 configurable event decoding parallelism

rautenrieth-da:
This PR adds a CLI parameter for the events decoding parallelism.


kamil-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -62,9 +64,6 @@ private[appendonlydao] final class TransactionsReader(
   // This significantly improves the performance of the transaction service.
   private val outputStreamBufferSize = 128
 
-  // TODO: make this parameter configurable
-  private val ContractStateEventsStreamParallelismLevel = 4
-
   private val TransactionEventsFetchParallelism = 8
```
tudor-da, line 67: Very noble PR. Can you please reuse the config parameter for this value (`TransactionEventsFetchParallelism`) as well?
Context: These two are never to be used concurrently (i.e. it depends on the in-memory fan-out toggle which one is used), but they both can leverage the CLI config param

rautenrieth-da: Those two have different default values and a slightly different meaning, I would be careful with reusing the same config parameter.

Are you sure you want this parameters default value to be linked to the default value of the other parameter?

tudor-da: Yeah, the different default values are random. 
1. wrt to the meaning, they both specify the parallelism used in the fetching/decoding with the only difference being that `getContractStateEvents` doesn't use parallel fetch but only parallel decode. 
2. `getContractStateEvents` which uses `ContractStateEventsStreamParallelismLevel` is interchangeable in usage with `getTransactionLogUpdates` which uses `TransactionEventsFetchParallelism`: they both are internal streams meant for populating the LAPI caches/buffers. When in-memory fan-out is enabled, the later used. Without in-memory fan-out `getContractStateEvents` is used

rautenrieth-da: @kamil-da Are you ok with changing the default of `ContractStateEventsStreamParallelismLevel` to 8?

kamil-da: Yes, I think it's ok.

rautenrieth-da: @kamil-da @tudor-da I have updated the PR according to the above discussion.

ledger/sandbox-common/src/main/scala/platform/sandbox/cli/CommonCliBase.scala
```diff
@@ -213,6 +213,15 @@ class CommonCliBase(name: LedgerName) {
         )
         .action((eventsPageSize, config) => config.copy(eventsPageSize = eventsPageSize))
 
+      opt[Int]("events-decoding-parallelism")
```
tudor-da:
```suggestion
      opt[Int]("buffers-prefetching-parallelism")
```
Let's be more specific, since not all streaming queries use parallel fetch/decode

ledger/sandbox-common/src/main/scala/platform/sandbox/cli/CommonCliBase.scala
```diff
@@ -213,6 +213,15 @@ class CommonCliBase(name: LedgerName) {
...
+        .optional()
+        .text(
+          s"Number of events decoded in parallel. Default is ${SandboxConfig.DefaultEventsDecodingParallelism}."
```
tudor-da:
```suggestion
          s"Number of events fetched/decoded in parallel for populating the internal buffers. Default is ${SandboxConfig.DefaultEventsDecodingParallelism}."
```

tudor-da (approved):
Thanks for picking this up @rautenrieth-da, very welcome change. Left minor comments regarding the scope of this CLI config

kamil-da (approved):

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -63,6 +64,7 @@ object Config {
       participants = Vector.empty,
       maxInboundMessageSize = DefaultMaxInboundMessageSize,
       eventsPageSize = IndexConfiguration.DefaultEventsPageSize,
+      eventsProcessingParallelism = IndexConfiguration.DefaultEventsProcessingParallelism,
```
miklos-da, line 67: This case class has reached 20 fields this way. Can we group these arguments, please?
For example, the following could be clearly separate groups:
* feature flags
* caches
* metrics
* engine

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -62,11 +64,6 @@ private[appendonlydao] final class TransactionsReader(
   // This significantly improves the performance of the transaction service.
   private val outputStreamBufferSize = 128
 
-  // TODO: make this parameter configurable
```
stefanobaghino-da, line 65: Mission accomplished.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPostCommitValidationSpec.scala
```diff
@@ -20,9 +20,10 @@ import scala.concurrent.duration.DurationInt
 private[dao] trait JdbcLedgerDaoPostCommitValidationSpec extends LoneElement {
   this: AsyncFlatSpec with Matchers with JdbcLedgerDaoSuite =>
 
-  override protected def daoOwner(eventsPageSize: Int)(implicit
+  override protected def daoOwner(eventsPageSize: Int, eventsProcessingParallelism: Int)(implicit
       loggingContext: LoggingContext
-  ): ResourceOwner[LedgerDao] =
+  ): ResourceOwner[LedgerDao] = {
+    val _ = eventsProcessingParallelism // Silence unused param warning
```
stefanobaghino-da, line 26: Weird, I'm relatively sure unused parameters don't cause a warning on `override`s.

PR #9939, opened 2021-06-09
Make PartialTransaction be an engine-internal detail

nickchapman-da:
Goal: Make `PartialTransaction` be an engine-internal detail, and not a user facing concept.
- This opens up the chance to simplify the representation of transactions (https://github.com/digital-asset/daml/pull/9909)

Changes here:
- Stop `scenario-service` reporting _partial transaction_ info.
- Adapt a couple of tests.
- Update doc

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/daml/daml-studio.rst
+++ b/docs/source/daml/daml-studio.rst
@@ -221,8 +221,6 @@ Stack trace
    location is the first entry in the list.
 Ledger time
    The ledger time at which the error occurred.
-Partial transaction
-   The transaction that is being constructed, but not yet committed to the ledger.
 Committed transaction
    Transactions that were successfully committed to the ledger prior to the error.
 Trace
@@ -243,8 +241,6 @@ The ``abort``, ``assert`` and ``debug`` inbuilt functions can be used in updates
       Aborted:  stop
 
     Ledger time: 1970-01-01T00:00:00Z
-
-    Partial transaction:
 
     Trace:
       "hello, world!"
@@ -270,13 +266,6 @@ in the contract, but not authorizing the create:
           failed due to a missing authorization from 'Bob'
 
     Ledger time: 1970-01-01T00:00:00Z
-
-    Partial transaction:
-      Sub-transactions:
-         #0
-         └─> create CreateAuthFailure:Example
-             with
-               party1 = 'Alice'; party2 = 'Bob'
 
 To create the "Example" contract one would need to bring both parties to
 authorize the creation via a choice, for example 'Alice' could create a contract
@@ -304,15 +293,6 @@ choice 'Consume' of which he is not a controller
           failed due to a missing authorization from 'Alice'
 
     Ledger time: 1970-01-01T00:00:00Z
-
-    Partial transaction:
-      Sub-transactions:
-         #0
-         └─> fetch #0:0 (ExerciseAuthFailure:Example)
-
-         #1
-         └─> 'Alice' exercises Consume on #0:0 (ExerciseAuthFailure:Example)
-                     with
 
     Committed transactions:
       TX #0 1970-01-01T00:00:00Z (unknown source)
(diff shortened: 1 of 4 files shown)


docs/source/daml/daml-studio.rst
```diff
@@ -305,15 +294,6 @@ choice 'Consume' of which he is not a controller
 
     Ledger time: 1970-01-01T00:00:00Z
 
-    Partial transaction:
```
cocreature, line 308: They are super helpful for debugging, let’s please not drop this.

nickchapman-da: So you really want a partial transaction as a user facing concept?

cocreature: Yes, seeing the transaction that got built up to the point where your code failed is much more useful than a stack trace. And users already have to understand transactions so understanding a partial transaction adds very little mental overhead.
I’m open to decoupling the user-facing partial transaciton from the internal one though but I’m not sure we gain much from that.

nickchapman-da: My feeling is that _node-id_s are just an internal implementation detail.
But the way that the scenario-service makes use partial-transaction makes it really hard to change. 
It's conceivable to have a concept of partial-transaction without node-ids
But I thought it would be way simpler to just strip them out entirely.

nickchapman-da: I'm open to keeping a user facing partial-transaction if we can decouple it from Node-Ids.

nickchapman-da: But is a partial transaction user facing anywhere other than in the scenario service?
(and that is semi deprecated, no?)

remyhaemmerle-da: _Node ID_ could be "consider" as an internal implementation detail. (I put consider between "." because ledger implementer heavily really on it)

However _event ID_ is a **central** concept of the ledger API.  Currently the event ID is very close to the Node Id so we somehow  use interchangeably in the scenario service.

On the other hand I think that `PartialTransaction` is an internal implementation detail, I pretty we can now use the standard `GenTransaction` even in case of failure, hence simplify greatly the pretty printer. 

Before having the exception mechanism in place it would have been very complicate to produce a such a transaction in case of failure, but now we just have to pop *all* the continuations with something like :

```
  while (machine.kontDepth() != 0)  
    machine.popKont() match {
          case handler: KTryCatchHandler =>
              machine.withOnLedger("unwindToHandler/KCloseExercise") { onLedger =>
                onLedger.ptx = onLedger.ptx.abortTry
              }
          case _: KCloseExercise =>
            machine.withOnLedger("unwindToHandler/KCloseExercise") { onLedger =>
              onLedger.ptx = onLedger.ptx.abortExercises
            }
          case _ =>
        }
  machin.ptx.finish 
```

remyhaemmerle-da: It goes without saying that canton need access to the failing interpretation -- We even promise to implement it pretty soon. 
We definitively do not want to expose canton to `PartialTransaction`.

remyhaemmerle-da: > So I’d like to keep exposing partial transactions 
if you speak of the `PartialTransaction` type, I think we should use proper `GenTransaction` instead
if you speak of a transaction not properly "finished", I agree, but I think we could/should  represent if using `GenTransaction` type. 
> and I’d like to keep exposing cross-references between nodes and I’d like to keep sensible contract ids.
I agree.

cocreature: I don’t care about the type we expose, I only care about exposing an incomplete transaction in some form. If GenTransaciton works that’s perfectly fine. That said, I think at the moment, the difference between partial transaciton is pretty small so less fuzzed about that.

nickchapman-da:
Thanks for the comments above.

I made a new PR to expose an external-facing definition for `IncompleteTransaction`, which is build on top of `GenTransaction`. See: #9952

I will close this PR.

PR #9940, opened 2021-06-09
Log call to submitAndWaitRequest with the command id provided in the log ctx

### LINKED ISSUE #9918: Log correlation between the JSON API and the Ledger
Currently there is no nice way to correlate logs of the JSON API with logs from the Ledger. We should include the command id in the JSON API logs to enable this and maybe also the trace context.

realvictorprm:
this fixes #9918

The trace ctx is being done in a seperate pr.

changelog_begin

- [HTTP-JSON] Calls which trigger a submitAndWaitRequest are logged with the command id provided in the log ctx

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
@@ -162,7 +162,7 @@ class CommandService(
     )
       .run { implicit lc =>
         logger.info(
-          s"Submitting $commandKind command and synchronously waiting for completion"
+          s"Submitting $commandKind command"
         )
         Commands.submitAndWaitRequest(
           jwtPayload.ledgerId,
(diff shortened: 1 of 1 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
```diff
@@ -150,18 +151,27 @@ class CommandService(
...
-      commandId,
-      jwtPayload.actAs,
-      jwtPayload.readAs,
-      command,
+    withEnrichedLoggingContext(
+      label[lar.CommandId],
+      Map("command_id" -> commandId.toString),
     )
+      .run { implicit lc =>
+        logger.info(
```
cocreature, line 163: We already log every request, e.g, `    logger.trace("sending exercise command to ledger")`.
I think what would be more helpful is to pass in the type of request (exercise, create, …) in here and then log something more specific than "doing a submit and wait request".

ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
```diff
@@ -150,18 +151,28 @@ class CommandService(
...
+          s"Submitting $commandKind command and synchronously waiting for completion"
```
cocreature, line 165:
```suggestion
          s"Submitting $commandKind command"
```
The JSON API only submits things synchronously so no point in adding that.

cocreature (approved):
Nice, thank you!

PR #9945, opened 2021-06-09
Proxy ledger API health endpoint in JSON API health endpoint

cocreature:
This is a bit more useful than just querying ledger end.

changelog_begin

- [JSON API] The healthcheck endpoint on the JSON API now proxies the
  health check endpoint on the underlying ledger. Previously it only
  queried for ledger end to determine ledger health.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/HealthService.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/HealthService.scala
@@ -4,37 +4,49 @@
 package com.daml.http
 
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, HttpResponse, StatusCodes}
+import io.grpc.health.v1.health.HealthCheckResponse
 import scalaz.Scalaz._
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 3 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/HealthService.scala
```diff
@@ -4,37 +4,49 @@
...
 import scalaz.Scalaz._
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.Try
+import scala.util.{Failure, Success}
 import scala.util.control.NonFatal
 
 final class HealthService(
-    getLedgerEnd: HealthService.GetLedgerEnd,
+    getLedgerEnd: HealthService.GetHealthCheckResponse,
```
SamirTalwar-DA, line 15: Probably rename this?

cocreature: good point c12384a11cd0ad4a0f447f896f3e9bb5d3425e54

S11001001 (approved):

PR #9948, opened 2021-06-10
Resolve contract keys conflicts in disclosed contracts

mziolekda:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
@@ -23,6 +23,14 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
   private def deleteContract(contractId: ContractId): Vector[NamedParameter] =
     Vector[NamedParameter]("contract_id" -> contractId)
 
+  // This query blanks out contract keys for all previous contracts irrespective of how they were created
+  // a) for disclosed contracts
+  // b) for contracts where parties hosted by this participant are stakeholders
+  // The contracts covered by case b) are an empty set, considering the guarantees of the committer that
+  // all transactions communicated via the update stream adhere to the ledger model i.e. all conflict on
+  // contract keys have been resolved at this point. Such contracts must have been archived already
+  // either in a previous or current transaction. What we are left with effectively are contracts covered
+  // by case a).
   private val nullifyPastKeysQuery =
     s"update participant_contracts set create_key_hash = null where create_key_hash = {create_key_hash}"
 
@@ -41,7 +49,16 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
   }
 
   protected def buildNullifyPastKeys(info: TransactionIndexing.ContractsInfo): Option[BatchSql] = {
-    val nullifyPastKey = info.netKeyNullifies.iterator.map(nullifyPastKeys).toSeq
+    val nullifyPastKey = info.netCreates
+      .flatMap(create =>
+        create.key
+          .map(convertLfValueKey(create.templateId, _))
+          .map(_.hash.bytes.toByteArray)
+          .toList
+      )
+      .iterator
+      .map(nullifyPastKeys)
+      .toSeq
     batch(nullifyPastKeysQuery, nullifyPastKey)
   }
 
@@ -82,7 +99,7 @@ private[events] object ContractsTable {
   final case class Executables(
       deleteContracts: Option[BatchSql],
       insertContracts: Executable,
-      keyNullifies: Option[BatchSql],
+      nullifyPastKeys: Option[BatchSql],
   )
 
   def apply(dbType: DbType): ContractsTable =
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
@@ -23,10 +23,10 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
   private def deleteContract(contractId: ContractId): Vector[NamedParameter] =
     Vector[NamedParameter]("contract_id" -> contractId)
 
-  private val deleteContractByKeyQuery =
-    s"delete from participant_contracts where create_key_hash = {create_key_hash}"
+  private val nullifyPastKeysQuery =
+    s"update participant_contracts set create_key_hash = null where create_key_hash = {create_key_hash}"
 
-  private def deleteContractByKey(contractKeyHash: Array[Byte]): Vector[NamedParameter] =
+  private def nullifyPastKeys(contractKeyHash: Array[Byte]): Vector[NamedParameter] =
     Vector[NamedParameter]("create_key_hash" -> contractKeyHash)
 
   def toExecutables(
@@ -40,9 +40,9 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
     batch(deleteContractQuery, deletes)
   }
 
-  protected def buildDeletesByKey(info: TransactionIndexing.ContractsInfo): Option[BatchSql] = {
-    val deletesByKey = info.netArchivesByKey.iterator.map(deleteContractByKey).toSeq
-    batch(deleteContractByKeyQuery, deletesByKey)
+  protected def buildNullifyPastKeys(info: TransactionIndexing.ContractsInfo): Option[BatchSql] = {
+    val nullifyPastKey = info.netKeyNullifies.iterator.map(nullifyPastKeys).toSeq
+    batch(nullifyPastKeysQuery, nullifyPastKey)
   }
 
   override final def lookupContractKeyGlobally(
@@ -79,7 +79,7 @@ private[events] object ContractsTable {
     def execute()(implicit connection: Connection): Unit
   }
 
-  final case class Executables(deleteContracts: Option[BatchSql], insertContracts: Executable, deleteContractsByKey: Option[BatchSql])
+  final case class Executables(deleteContracts: Option[BatchSql], insertContracts: Executable, keyNullifies: Option[BatchSql])
 
   def apply(dbType: DbType): ContractsTable =
     dbType match {
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
@@ -23,6 +23,12 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
   private def deleteContract(contractId: ContractId): Vector[NamedParameter] =
     Vector[NamedParameter]("contract_id" -> contractId)
 
+  private val deleteContractByKeyQuery =
+    s"delete from participant_contracts where create_key_hash = {create_key_hash}"
+
+  private def deleteContractByKey(contractKeyHash: Array[Byte]): Vector[NamedParameter] =
+    Vector[NamedParameter]("create_key_hash" -> contractKeyHash)
+
   def toExecutables(
       info: TransactionIndexing.ContractsInfo,
       tx: TransactionIndexing.TransactionInfo,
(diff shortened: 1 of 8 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTableH2.scala
```diff
@@ -27,6 +27,7 @@ object ContractsTableH2 extends ContractsTable {
   ): ContractsTable.Executables = ContractsTable.Executables(
     deleteContracts = buildDeletes(info),
     insertContracts = buildInserts(tx, info, serialized),
+    buildNullifyPastKeys(info),
```
gerolf-da, line 30:
```suggestion
    keyNullifies = buildNullifyPastKeys(info),
```

gerolf-da (approved):
Thanks, @mziolekda. I'd like somebody else from the team to also review this before merging.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
```diff
@@ -23,6 +23,12 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
   private def deleteContract(contractId: ContractId): Vector[NamedParameter] =
     Vector[NamedParameter]("contract_id" -> contractId)
 
+  private val nullifyPastKeysQuery =
```
cocreature, line 26: I think this could benefit from some comments somewhere on why this is a safe thing to do.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
```diff
@@ -23,6 +23,12 @@ private[events] abstract class ContractsTable extends PostCommitValidationData {
...
+    s"update participant_contracts set create_key_hash = null where create_key_hash = {create_key_hash}"
```
oliverse-da, line 27: Is there a possibility that nulling out key hashes may break multi-key semantics in canton? cc @andreaslochbihler-da

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
```diff
@@ -209,6 +209,12 @@ object TransactionIndexing {
       val allContractIds = allCreatedContractIds.union(archived)
       val netCreates = created.filterNot(c => archived(c.coid))
       val netArchives = archived.filterNot(allCreatedContractIds)
+      val netKeyNullifies = netCreates.flatMap(create =>
```
cocreature, line 212: This seems slightly off. I think it makes more sense to serialize later like we do for everything else afaict.

cocreature (approved):
Nice work, thanks!

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
meiersi-da: Yes, I'm pretty sure that this change breaks multi-key semantics, as there it does not hold that seeing a create with a key means that all previous creates with that key must have been archived.

That said: this change is for the legacy IndexDB schema only. In the append-only schema it is not required.

cocreature: We don’t make any promises though which contract you can fetch in multi-key semantics. So it’s not entirely clear to me that this actually breaks anything.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
```diff
@@ -68,7 +79,11 @@ private[events] object ContractsTable {
     def execute()(implicit connection: Connection): Unit
   }
 
-  final case class Executables(deleteContracts: Option[BatchSql], insertContracts: Executable)
+  final case class Executables(
+      deleteContracts: Option[BatchSql],
+      insertContracts: Executable,
+      keyNullifies: Option[BatchSql],
```
meiersi-da, line 85:
```suggestion
      nullifyPastKeys: Option[BatchSql],
```
consistency with the other fields and the naming in other parts of the code

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
meiersi-da, line 212:
```suggestion
      val netCreateKeys = netCreates.flatMap(create =>
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsWriter.scala
```diff
@@ -49,6 +49,10 @@ private[platform] object TransactionsWriter {
         Timed.value(deleteContractsBatch, deleteContracts.execute())
       }
 
+      for (keyNullifies <- contractsTableExecutables.keyNullifies) {
+        keyNullifies.execute()
```
meiersi-da, line 53: Metrics are missing

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
oliverse-da: But in the following scenario:

1. Create contract1 with key key1
2. Create contract2 with key key1 (nulls out contract1's key1 reference)
3. Archive contract2
4. FetchByKey(key1) would fail to look up key1

The canton tests still run on the legacy index. If we had an indexer configuration to opt out the nulling out, we could use this in canton.

meiersi-da (commented):
Thanks for creating this patch @mziolekda . It looks functionally correct.

I've added a bunch of comments for simple code cleanup, but don't consider them must-haves for the patch, if we're really short on time.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
andreaslochbihler-da: Indeed, this changes the behavior. I agree with @cocreature that this is probably still within the guarantees for "fetchSome" semantics. But the UX will degrade with this change even on a single non-UCK domains. Example topology: Alice (key maintainer) on P1, Bob on P2.

1. Alice creates a contract with key and observer Bob. Bob concurrently creates another contract with the same key. Due to concurrency, both transactions are accepted. 
2. Bob archives his contract.
3. Alice tries to lookup the key and finds nothing, even though her contract is still active (and reported as active by the ACS service).

I haven't tried it out but I suspect that this scenario used to work in Canton.

cocreature: I think it’s somewhat debatable if UX degrades or not. If you do step 2 and 3 in a single transaction you get `Nothing` at the moment, if you do it over multiple transactions you get a key (assuming your suspicion is correct). It’s not clear to me whether changing the multi-transaction behavior to match how things work within a single transaction is really worse.
More generally, I think this gets to my general point that multi-key semantics are ill-defined at the moment and I think worrying about UX without even being clear on what exactly the current semantics are is not all that productive.

andreaslochbihler-da: How would you be able to merge steps 2 and 3? They are submitted over distinct participants.

cocreature: I can run a submission as Alice that gets authority from Bob somehow to archive the contract. And more generally, I think we shouldn’t make the behavior here topology dependent.

tudor-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
andreaslochbihler-da: > And more generally, I think we shouldn’t make the behavior here topology dependent.

I agree that the *allowed* behaviors should not depend on the topology. But the *observed* behavior does depend on it, because different topologies lead to different partial knowledge.

daravep: @cocreature @bame-da @meiersi-da @mziolekda @gerolf-da  @vivd-da @sormeter 

I've discussed this with @oliverse-da and @andreaslochbihler-da.

1. This patch will very much likely break current Canton on old schema running with non-UCK mode.
2. We acknowledge the urgency of this and ask you to not hesitate to proceed with his on the old schema. 
3. We assume that Canton GA and DDF4O for HKEX will be on the parallel indexer schema where this is not an issue.
4. Until we have fully migrated to the new indexer schema, we will mark all upcoming Canton releases as "DEFECT IN CERTAIN CASES" (June, July).
5. We will just disable any test that fails as a result of this change for now.
6. @oliverse-da will first finish his work on adding DAML exceptions to Canton before enabling the parallel indexer schema as default.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTableH2.scala
mziolekda: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
mziolekda: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
mziolekda: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
mziolekda: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionsWriter.scala
mziolekda: Done

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
cocreature: I think you’re missing a crucial point here: Canton GA and DDF4O run in UCK mode and have to. NON-UCK mode has zero data continuity guarantees since we have not specd multi-key semantics in any way that we can keep stable in speedy and you fall over if they change.

daravep: @cocreature there are practical and theoretical issues. we'll have to pin down the semantics pretty soon once we following up with the multi-domain MVP GA release, but that is nothing that we need to do in Q2. however, the issue that we face right now is that this change is going to break quite a few of our existing tests. so the above path is to deal with these breaks. and data continuity is something we didn't promise anyway as of today.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala
mziolekda: Done

rautenrieth-da:
Not necessary for this PR, but I think it would be nice if this situation was covered in the JdbcLedgerDao test suite. These tests use the low-level interface to access the index database, and can simulate things like state updates coming from a privacy-aware distributed ledger.

[JdbcLedgerDaoContractsSpec.scala](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoContractsSpec.scala#L237) would be a good place to add such a test.

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/ContractsTable.scala
oliverse-da: Canton follow-up tracked in https://github.com/DACH-NY/canton/issues/6493

PR #9949, opened 2021-06-10
LF: Fix specification of ArithmeticError

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

part of #8020

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/spec/daml-lf-1.rst
+++ b/daml-lf/spec/daml-lf-1.rst
@@ -3871,11 +3871,11 @@ BigNumeric functions
 
   - ``'ROUNDING_UP'`` : Round away from zero
 
-  - ``'ROUNDING_DOWN'`` : Rounds towards zero
+  - ``'ROUNDING_DOWN'`` : Round towards zero
 
-  - ``'ROUNDING_CEILING'`` : Rounds towards positive infinity.
+  - ``'ROUNDING_CEILING'`` : Round towards positive infinity.
 
-  - ``'ROUNDING_FLOOR'`` : Rounds towards negative infinity
+  - ``'ROUNDING_FLOOR'`` : Round towards negative infinity
 
   - ``'ROUNDING_HALF_UP'`` : Round towards the nearest neighbor unless
     both neighbors are equidistant, in which case round away from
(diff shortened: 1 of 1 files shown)


daml-lf/spec/daml-lf-1.rst
```diff
@@ -3364,6 +3351,58 @@ This section lists the built-in functions supported by Daml-LF 1.
...
+~~~~~~~~~~~~~~~~
+
+Some builtin functions can throw non-fatal exceptions, i.e. exceptions
+catchable by the ``TryCatch`` update expression. Those exceptions are
+not built in the language but are standard exceptions defined in user
+land. The builtin functions from an engine compliant with the current
+specification should be able to produce and handle (notably the
+``ANY_EXCEPTION_MESSAGE`` builtin function) such exceptions without
+the packages where those exception are defined to be explicitly
+loaded.  Any other usage like on the exception payload, like
```
cocreature, line 3364:
```suggestion
``ANY_EXCEPTION_MESSAGE`` builtin function) such exceptions even if the package they are defined in has not been loaded.
```

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3364,6 +3351,58 @@ This section lists the built-in functions supported by Daml-LF 1.
...
+      exception ArithmeticError ↦ {
+         'message' λ x : DA.Exception.ArithmeticError:ArithmeticError.
+            DA.Exception.ArithmeticError:ArithmeticError { message } x
+      } ;
+   }
+
+The package can be produced in a stable way by Daml SDK 1.14 or latter
+with the command::
+
+  bazel build //compiler/damlc/stable-packages:stable-packages
```
cocreature, line 3392: not sure that makes sense in a spec.

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3846,8 +3889,8 @@ BigNumeric functions
     unless both neighbors are equidistant, in which case round towards
     the even neighbor.
 
-  - ``'ROUNDING_UNNECESSARY'`` : Throw `ArithmeticError` if the exact result cannot be
-    represented.
+  - ``'ROUNDING_UNNECESSARY'`` : Throws ``ArithmeticError`` if the
```
cocreature, line 3892: that change seems off, `ROUNDING_UNNECSSARY` does not throw. It causes the bulitin you pass it in to throw.

cocreature (approved):
Thanks!

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: I am not sure what is the issue here. 
I just added
- a "s" at the end of "Throw" 
- two "`" 
- and a newline

cocreature: With the `s` it sounds  to me like `ROUNDING_UNNECESSARY` is a primitive that throws rather than an argument which causes something else to throw. But that’s mostly gut feel so maybe my language intuition is just off here.

remyhaemmerle-da: OK. I revert back and remove the 's'. I also removing from some of the "Round" there were in the enumeration.

PR #9950, opened 2021-06-10
Proxy ledger API health endpoint in JSON API health endpoint (#9945)

cocreature:
* Proxy ledger API health endpoint in JSON API health endpoint

This is a bit more useful than just querying ledger end.

changelog_begin

- [JSON API] The healthcheck endpoint on the JSON API now proxies the
  health check endpoint on the underlying ledger. Previously it only
  queried for ledger end to determine ledger health.

changelog_end

* naming things is hard

changelog_begin
changelog_end

* Fix Scala 2.12

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/failure/scala/http/FailureTests.scala
+++ b/ledger-service/http-json/src/failure/scala/http/FailureTests.scala
@@ -63,7 +63,7 @@ final class FailureTests
         (status, out) <- getRequestEncoded(uri.withPath(Uri.Path("/readyz")))
         _ = status shouldBe StatusCodes.ServiceUnavailable
         _ = out shouldBe
-          """[-] ledger failed
+          """[-] ledger failed (io.grpc.StatusRuntimeException: UNAVAILABLE: io exception)
             |[+] database ok
             |readyz check failed
             |""".stripMargin.replace("\r\n", "\n")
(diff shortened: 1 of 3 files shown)


ledger-service/http-json/src/failure/scala/http/FailureTests.scala
```diff
@@ -63,7 +63,7 @@ final class FailureTests
         (status, out) <- getRequestEncoded(uri.withPath(Uri.Path("/readyz")))
         _ = status shouldBe StatusCodes.ServiceUnavailable
         _ = out shouldBe
-          """[-] ledger failed
+          """[-] ledger failed (io.grpc.StatusRuntimeException: UNAVAILABLE: io exception)
```
stefanobaghino-da, line 66: I'm not sure we should expose this level of detail, `UNAVAILABLE` and the exception message is probably enough, without the specific type, but we don't need to do it here.

stefanobaghino-da (approved):

ledger-service/http-json/src/failure/scala/http/FailureTests.scala
cocreature: You’re assuming that we know that we can only get `StatusRuntimeException`s. I’m not sure that’s true. I think we can fail on lower levels of the stack as well.

stefanobaghino-da: It's possible, but that's why I was suggesting to keep the message in. But it's not really important. :)

PR #9952, opened 2021-06-10
Define IncompleteTransaction for use by scenario-service

nickchapman-da:
Define `IncompleteTransaction`, as the external representation for an incomplete transaction.

- Use this in `scenario-service` instead of `PartialTransaction`, without changing the behaviour.
- Goal is for `PartialTransaction` to be an engine-internal detail.
- This opens up the chance to simplify the representation of transactions and there incremental construction (#9909)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -407,12 +407,12 @@ final class Conversions(
       .addAllNodes(tx.nodes.map(convertNode).asJava)
       .addAllRoots(tx.roots.toList.map(convertTxNodeId).asJava)
 
-    incomplete.exerciseContextMaybe.foreach { ctx =>
+    incomplete.exerciseContextMaybe.foreach { exe =>
       val ecBuilder = proto.ExerciseContext.newBuilder
-        .setTargetId(mkContractRef(ctx.targetCoid, ctx.templateId))
-        .setChoiceId(ctx.choiceId)
-        .setChosenValue(convertValue(ctx.chosenValue))
-      ctx.optLocation.map(loc => ecBuilder.setExerciseLocation(convertLocation(loc)))
+        .setTargetId(mkContractRef(exe.targetCoid, exe.templateId))
+        .setChoiceId(exe.choiceId)
+        .setChosenValue(convertValue(exe.chosenValue))
+      exe.optLocation.map(loc => ecBuilder.setExerciseLocation(convertLocation(loc)))
       builder.setExerciseContext(ecBuilder.build)
     }
     builder.build
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -6,18 +6,17 @@ package com.daml.lf.scenario
 import com.daml.lf.data.{ImmArray, Numeric, Ref}
 import com.daml.lf.ledger.EventId
 import com.daml.lf.scenario.api.{v1 => proto}
-import com.daml.lf.speedy.{SError, SValue, PartialTransaction => SPartialTransaction, TraceLog}
-import com.daml.lf.transaction.{GlobalKey, Node => N, NodeId}
+import com.daml.lf.speedy.{SError, SValue, TraceLog}
+import com.daml.lf.transaction.{GlobalKey, IncompleteTransaction, Node => N, NodeId}
 import com.daml.lf.ledger._
 import com.daml.lf.value.{Value => V}
 
-import scala.annotation.tailrec
 import scala.jdk.CollectionConverters._
 
 final class Conversions(
     homePackageId: Ref.PackageId,
     ledger: ScenarioLedger,
-    ptx: SPartialTransaction,
+    incomplete: IncompleteTransaction,
     traceLog: TraceLog,
     commitLocation: Option[Ref.Location],
     stackTrace: ImmArray[Ref.Location],
@@ -30,7 +29,7 @@ final class Conversions(
 
   // The ledger data will not contain information from the partial transaction at this point.
   // We need the mapping for converting error message so we manually add it here.
-  private val ptxCoidToNodeId = ptx.nodes
+  private val ptxCoidToNodeId = incomplete.transaction.nodes
     .collect { case (nodeId, node: N.NodeCreate[V.ContractId]) =>
       node.coid -> ledger.ptxEventId(nodeId)
     }
@@ -77,7 +76,7 @@ final class Conversions(
     builder.addAllStackTrace(stackTrace.map(convertLocation).toSeq.asJava)
 
     builder.setPartialTransaction(
-      convertPartialTransaction(ptx)
+      convertPartialTransaction(incomplete)
     )
 
     err match {
@@ -401,26 +400,16 @@ final class Conversions(
       .build
   }
 
-  def convertPartialTransaction(ptx: SPartialTransaction): proto.PartialTransaction = {
+  def convertPartialTransaction(incomplete: IncompleteTransaction): proto.PartialTransaction = {
+    val tx = incomplete.transaction
+
     val builder = proto.PartialTransaction.newBuilder
-      .addAllNodes(ptx.nodes.map(convertNode).asJava)
-      .addAllRoots(
-        ptx.context.children.toImmArray.toSeq.sortBy(_.index).map(convertTxNodeId).asJava
-      )
+      .addAllNodes(tx.nodes.map(convertNode).asJava)
+      .addAllRoots(tx.roots.toList.map(convertTxNodeId).asJava)
 
-    @tailrec
-    def unwindToExercise(
-        contextInfo: SPartialTransaction.ContextInfo
-    ): Option[SPartialTransaction.ExercisesContextInfo] = contextInfo match {
-      case ctx: SPartialTransaction.ExercisesContextInfo => Some(ctx)
-      case ctx: SPartialTransaction.TryContextInfo =>
-        unwindToExercise(ctx.parent.info)
-      case _: SPartialTransaction.RootContextInfo => None
-    }
-
-    unwindToExercise(ptx.context.info).foreach { ctx =>
+    incomplete.exerciseContextMaybe.foreach { ctx =>
       val ecBuilder = proto.ExerciseContext.newBuilder
-        .setTargetId(mkContractRef(ctx.targetId, ctx.templateId))
+        .setTargetId(mkContractRef(ctx.targetCoid, ctx.templateId))
         .setChoiceId(ctx.choiceId)
         .setChosenValue(convertValue(ctx.chosenValue))
       ctx.optLocation.map(loc => ecBuilder.setExerciseLocation(convertLocation(loc)))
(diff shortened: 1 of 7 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -35,6 +37,14 @@ private[lf] object PartialTransaction {
   type Node = Node.GenNode[NodeId, Value.ContractId]
   type LeafNode = Node.LeafOnlyActionNode[Value.ContractId]
 
+  private type TX = GenTransaction[NodeId, Value.ContractId]
+  private type ExerciseNode = Node.NodeExercises[NodeId, Value.ContractId]
+
+  private final case class ConcreteIncompleteTx(
```
cocreature, line 43:
```suggestion
  private final case class IncompleteTxImpl(
```

cocreature: doesn’t matter much but this is common in our codebase and in Scala in general so I don’t see a reason to diverge from that naming scheme.

cocreature (approved):
looks sensible, thanks!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/IncompleteTransaction.scala
```diff
@@ -0,0 +1,18 @@
...
+package com.daml.lf
+package transaction
+
+import com.daml.lf.value.Value
+
+trait IncompleteTransaction {
+
+  type Nid = NodeId
+  type Cid = Value.ContractId
+  type TX = GenTransaction[Nid, Cid]
```
remyhaemmerle-da, line 13:
```suggestion
  type TX = Transaction.Transaction
```

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/IncompleteTransaction.scala
```diff
@@ -0,0 +1,18 @@
...
+  type ExerciseNode = Node.NodeExercises[Nid, Cid]
+
+  def transaction: TX
+  def exerciseContextMaybe: Option[ExerciseNode]
+}
```
remyhaemmerle-da, line 18: I think  it will be convenient that an `IncompleteTransaction` is just a normal transaction like we have done for `SubmittedTransaction` and `CommitedTransaction`. 

You could add in package.scala

```
  val IncompleteTransaction = DiscriminatedSubtype[VersionedTransaction[NodeId, ContractId]]
  type IncompleteTransaction = IncompleteTransaction.T
```

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -401,26 +400,16 @@ final class Conversions(
...
-        contextInfo: SPartialTransaction.ContextInfo
-    ): Option[SPartialTransaction.ExercisesContextInfo] = contextInfo match {
-      case ctx: SPartialTransaction.ExercisesContextInfo => Some(ctx)
-      case ctx: SPartialTransaction.TryContextInfo =>
-        unwindToExercise(ctx.parent.info)
-      case _: SPartialTransaction.RootContextInfo => None
-    }
+    val builder = proto.PartialTransaction.newBuilder
+      .addAllNodes(tx.nodes.map(convertNode).asJava)
+      .addAllRoots(tx.roots.toList.map(convertTxNodeId).asJava)
```
remyhaemmerle-da, line 408: `incomplete.transaction` use only the node until the last failing exercise, it should from the root.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -401,26 +400,16 @@ final class Conversions(
...
 
-    unwindToExercise(ptx.context.info).foreach { ctx =>
+    incomplete.exerciseContextMaybe.foreach { ctx =>
```
remyhaemmerle-da, line 410:
```suggestion
    val maybeFailingExercise = 
      tx.fold(Option.empty[NodeExercises[NodeId, ContractId]]) {
      case (_, (_, exe: NodeExercises[NodeId, ContractId])) if exe.exerciseResult.isEmpty =>
        Some(exe)
      case (acc, _) => acc
    }

    maybeFailingExercise.foreach { exe =>
```

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -401,26 +400,16 @@ final class Conversions(
...
       val ecBuilder = proto.ExerciseContext.newBuilder
-        .setTargetId(mkContractRef(ctx.targetId, ctx.templateId))
+        .setTargetId(mkContractRef(ctx.targetCoid, ctx.templateId))
         .setChoiceId(ctx.choiceId)
         .setChosenValue(convertValue(ctx.chosenValue))
```
remyhaemmerle-da, line 414:
```suggestion
        .setTargetId(mkContractRef(exe.targetCoid, exe.templateId))
        .setChoiceId(exe.choiceId)
        .setChosenValue(convertValue(exe.chosenValue))
```

remyhaemmerle-da (commented):
Look a very good start. There is two issues thought:
- The IncompletTransaction should start from the RootContext and not from the last ExericeContext (which correspond to the inner failling exercise)
- I will not expose the `maybeExerciseContext` field which is redundant and only usefull for the Scenario service.

cocreature:
>     * The IncompletTransaction should start from the RootContext and not from the last ExericeContext (which correspond to the inner failling exercise)

I don’t understand what you mean by that. Where is the “start” in your setting.

>     * I will not expose the `maybeExerciseContext` field which is redundant and only usefull for the Scenario service.

How else do you get it in the scenario service?

remyhaemmerle-da:
You can have a look in `def convertPartialTransaction(ptx: SPartialTransaction)` before all the closed nodes where included, now only the node under the last failing exercise are included.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -333,6 +343,28 @@ private[lf] case class PartialTransaction(
...
+      case ctx: PartialTransaction.ExercisesContextInfo => Some(ctx)
+      case ctx: PartialTransaction.TryContextInfo =>
+        unwindToExercise(ctx.parent.info)
+      case _: PartialTransaction.RootContextInfo => None
+    }
+
+    ConcreteIncompleteTx(
+      GenTransaction(
+        nodes,
+        ImmArray(context.children.toImmArray.toSeq.sortBy(_.index)),
```
cocreature, line 362: oh I see here is the bug @remyhaemmerle-da mentioned: This should be the root nodes of the partial transaction not the children of the context.

remyhaemmerle-da:
Of course we can change the type `IncompleteTransaction` in another PR, but : 
- it should be made `lf` private until we remove from it the `PartialTransaction.Context`
- we should output all the nodes in `convertPartialTransaction`

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: We still need to "unwind" completely the partial transaction.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/IncompleteTransaction.scala
remyhaemmerle-da: this is a suggestion.

remyhaemmerle-da:
By the way, we need a test that catch the regression bug introduced by this PR.

remyhaemmerle-da:
> By the way, we need a test that catch the regression bug introduced by this PR.

In fact the scenario client does not seem to use the nodes, so there is not really anything to test.

remyhaemmerle-da (approved):
There is in fact no change of behavior in the current state of the PR. 
I am fine to merge it as it is now.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -30,7 +29,7 @@ final class Conversions(
 
   // The ledger data will not contain information from the partial transaction at this point.
   // We need the mapping for converting error message so we manually add it here.
-  private val ptxCoidToNodeId = ptx.nodes
+  private val ptxCoidToNodeId = incomplete.transaction.nodes
```
cocreature, line 32: @remyhaemmerle-da @nickchapman-da we use the nodes here at least. I guess we don’t use the roots so it works?  But that means we rely on having nodes not reachable by the roots which seems extremely confusing so I think fixing this should happen before merging this PR.

nickchapman-da: This PR is a semantic no-op
Probably best to make the fix in a separate PR. To avoid confusing things.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
nickchapman-da: done

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/IncompleteTransaction.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: done

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
cocreature: I don’t think it is. You’re transaction has different roots than the partial transaction we had before We get lucky because we don’t use the roots but I don’t think that’s a sensible state. You’re effectively building an invalid transaction with unreachable nodes while relying on those nodes being there.

nickchapman-da: Are you sure?
The roots before were: `ptx.context.children.toImmArray.toSeq.sortBy(_.index)`
And they still are now.

nickchapman-da: Offline we discuss and agree that this PR did not introduce any bug/change. But the original code from scenario-service which is now moved to `PartialTransaction.finishIncomplete` is somehow not correct and should be looked at in the future.

PR #9954, opened 2021-06-10
Skip Scala_2_12 job on releases

cocreature:
This currently breaks older releases because they require a different
Scala 2.12 version. It also adds zero value for a release that
defaults to Scala 2.12 and it adds basically no value for a release
that defaults to Scala 2.13 (see comment for details).

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -104,6 +104,13 @@ jobs:
     - template: report-end.yml
 
 - job: Linux_scala_2_12
+  # We disable this for release builds. The actual release happens via the regular Linux builds
+  # and running tests on two scala versions doesn’t add any value if the tests already
+  # passed on the respective PRs that added the changes. The only difference here
+  # is the release version which should not affect things differently depending on the
+  # Scala version.
+  condition: and(succeeded(),
+                 eq(dependencies.check_for_release.outputs['out.is_release'], 'false'))
   dependsOn:
     - da_ghc_lib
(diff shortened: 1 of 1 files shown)


ci/build.yml
```diff
@@ -104,16 +104,20 @@ jobs:
     - template: report-end.yml
 
 - job: Linux_scala_2_12
+  # We disable this for release builds. The actual release happens via the regular Linux builds
+  # and running tests on two scala versions doesn’t add any value if the tests already
+  # passed on the respective PRs that added the changes. The only difference here
+  # is the release version which should not affect things differently depending on the
+  # Scala version.
+  condition: and(succeeded(),
```
cocreature, line 112: @garyverhaegen-da iirc this is fine and skipping this doesn’t break jobs like collect_build_data that depend on it. Can you confirm this?

cocreature: Not sure if I can refer to the variable directly below. This seemed to match what we do in other places.

garyverhaegen-da: As far as I understand, yes, you should be able to access dependencies in `condition`, and that's implicitly what things like `succeeded` do. I usually prefer to write the `condition` block after the `dependsOn` one for that reason, but I don't think Azure cares about ordering here.

garyverhaegen-da (approved):

PR #9970, opened 2021-06-11
A `ledger-api-test-tool` test case for divulging a contract with key twice on multiple participants [DPP-433]

kamil-da:
### Scenario
```
1. allocate partyA on participantA
2. allocate partyB on participantB
3. create a (helper : WithKeyDivulgenceHelper) contract with partyB being a signatory
4. create a (withKey1 : WithKey) contract with partyA being an owner
5. archive the withKey1 by exercising a choice on the helper - divulgence to partyB
6. create another contract (withKey2 : WithKey) with partyA being an owner
7. archive the withKey2 by exercising a choice on the helper - divulgence to partyB
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
@@ -6,16 +6,7 @@ package com.daml.ledger.api.testtool.suites
 import com.daml.ledger.api.testtool.infrastructure.Allocation._
 import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
 import com.daml.ledger.api.testtool.infrastructure.Synchronize.{synchronize, waitForContract}
-import com.daml.ledger.test.model.Test.Divulgence2._
-import com.daml.ledger.test.model.Test.Proposal._
-import com.daml.ledger.test.model.Test.{
-  Asset,
-  Divulgence1,
-  Divulgence2,
-  Proposal,
-  WithKey,
-  WithKeyDivulgenceHelper,
-}
+import com.daml.ledger.test.model.Test
 import scalaz.Tag
 
 final class DivulgenceIT extends LedgerTestSuite {
@@ -24,6 +15,7 @@ final class DivulgenceIT extends LedgerTestSuite {
     "Divulged contracts should not be exposed by the transaction service",
     allocate(TwoParties),
   )(implicit ec => { case Participants(Participant(ledger, alice, bob)) =>
+    import Test.{Divulgence1, Divulgence2}
     for {
       divulgence1 <- ledger.create(alice, Divulgence1(alice))
       divulgence2 <- ledger.create(bob, Divulgence2(bob, alice))
@@ -164,6 +156,7 @@ final class DivulgenceIT extends LedgerTestSuite {
     "Divulged contracts should not be exposed by the active contract service",
     allocate(TwoParties),
   )(implicit ec => { case Participants(Participant(ledger, alice, bob)) =>
+    import Test.{Divulgence1, Divulgence2}
     for {
       divulgence1 <- ledger.create(alice, Divulgence1(alice))
       divulgence2 <- ledger.create(bob, Divulgence2(bob, alice))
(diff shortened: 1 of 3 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
```diff
@@ -223,4 +230,37 @@ final class DivulgenceIT extends LedgerTestSuite {
...
+        WithKeyDivulgenceHelper(divulgedTo = partyB, withKeyOwner = partyA),
+      )
+
+      _ <- synchronize(alpha, beta)
+
+      // Create a WithKey contract owned by the partyA
+      withKey1 <- alpha.create(partyA, WithKey(partyA))
+
+      // Archive the withKey1 contract
+      _ <- alpha.exercise(partyA, helper.exerciseWithKeyDivulgenceHelper_Archive(_, withKey1))
```
cocreature, line 252: I’m not quite sure this is testing what you want to test. My understanding is we want to test the following (for simplicity assume each party has their own participant so party = participant).
1. A creates a contract where they are the only stakeholder.
2. A divulges that contract to B.
3. A archives that contract (without B knowing about the archive).
4. A creates another contract with the same key.
5. A divulges that contract to B.

You mixed the divulgence with the archival which means that even if B didn’t track keys for divulged contracts properly, it would never see a state where both contracts are archived.

kamil-da: Do you mean that combining divulgence and archiving contract in the same exercise will work differently than doing this in two steps?

cocreature: Yes in my scenario B does not see the archive. That was exactly the bug around disclosure: B saw two contracts without an interleaving archive.
In your example, B does see the archive so it never has two contracts active at the same time.

kamil-da: You're right! Thank you for catching this one.

kamil-da: I've updated the test case. Please review when you have time.

cocreature: It seems like you are still divulging the archive. I think what you want is something like the following:
```
_ <- alpha.exercise(partyA, withKey1.exerciseArchive(_))
```

kamil-da: 🤦  Right. A fix incoming.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/DivulgenceIT.scala
```diff
@@ -223,4 +230,42 @@ final class DivulgenceIT extends LedgerTestSuite {
...
+      // Create a WithKey contract owned by the partyA
+      withKey1 <- alpha.create(partyA, WithKey(partyA))
+
+      // Divulge the withKey1 contract
+      _ <- alpha.exercise(partyA, helper.exerciseWithKeyDivulgenceHelper_Fetch(_, withKey1))
+
+      _ <- synchronize(alpha, beta)
+
+      // Archive the withKey1 contract
+      _ <- alpha.exercise(partyA, withKey1.exerciseWithKey_Archive(_))
```
cocreature, line 257: any reason not to use the autogenerated `Archive` here, i.e., `exerciseArchive(_)`?

kamil-da: I haven't noticed the autogenerated one. Thank you!

kamil-da: I've changed the test case to use the autogenerated `Archive` and due to this I needed to restructure imports to avoid implicits ambiguity.

cocreature (approved):
Thanks!

PR #9972, opened 2021-06-11
Add proper error handling for missing contract keys

cocreature:
These are clearly not internal errors so we should not be calling
`crash` here. Canton in fact already started string matching on the
crash message which is definitely not what we want.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
@@ -88,7 +88,7 @@ object Error {
     }
 
     final case class ContractKeyNotFound(key: GlobalKey) extends Error {
-      override def msg = s"Contract with key $key could not be found"
+      override def msg = s"dependency error: couldn't find key: $key"
     }
 
     /** See com.daml.lf.transaction.Transaction.DuplicateContractKey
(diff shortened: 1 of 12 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
```diff
@@ -45,6 +45,11 @@ final case class ContractNotFound(ci: ContractId) extends Error {
   override def detailMsg: String = msg
 }
 
+final case class ContractKeyNotFound(key: GlobalKey) extends Error {
```
cocreature, line 48: @daravep fyi, this also covers a case that your string matching misses I believe (see the two different tests we have).

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
```diff
@@ -45,6 +45,11 @@ final case class ContractNotFound(ci: ContractId) extends Error {
...
+  override def msg = s"dependency error: couldn't find key: $key"
```
cocreature, line 49: I kept the error message the same to avoid unnecessarily breaking things like the ledger API test tool.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
daravep: Thanks a lot! My string matching is a "first step", reverse engineering the possible errors using the ledger API conformance tests. @remyhaemmerle-da is rewriting the lf-errors into case classes so we don't have to match the strings again. Please feel free to correct and fix the errors and change the strings.

remyhaemmerle-da (approved):
LGTM.

PR #9973, opened 2021-06-11
update Daml Hub auth in create-daml-app

garyverhaegen-da:
Current documentation says the token is passed as a cookie. This is a little bit more secure than passing it as a URL param, as it used to be. Note that while it is no longer documented behaviour, the token is actually still passed as a param (in addition to the cookie), so existing users are not broken (yet). Still, better if the code we generate for them matches our current docs.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/templates/create-daml-app/ui/src/components/LoginScreen.tsx.template
+++ b/templates/create-daml-app/ui/src/components/LoginScreen.tsx.template
@@ -45,16 +45,14 @@ const LoginScreen: React.FC<Props> = ({onLogin}) => {
 
   useEffect(() => {
     const url = new URL(window.location.toString());
-    const token = url.searchParams.get('token');
-    if (token === null) {
-      return;
-    }
     const party = url.searchParams.get('party');
     if (party === null) {
-      throw Error("When 'token' is passed via URL, 'party' must be passed too.");
+      return;
     }
     url.search = '';
     window.history.replaceState(window.history.state, '', url.toString());
+    const tokenCookiePair = document.cookie.split('; ').find(row => row.startsWith('DAMLHUB_LEDGER_ACCESS_TOKEN')) || '';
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Note: I've actually gone through the whole `daml-sdk-head` -> `daml-head new` -> `daml build && ...` -> deploy to Daml Hub process to double-check that this works.

templates/create-daml-app/ui/src/components/LoginScreen.tsx.template
```diff
@@ -45,16 +45,14 @@ const LoginScreen: React.FC<Props> = ({onLogin}) => {
...
-      return;
-    }
     const party = url.searchParams.get('party');
     if (party === null) {
-      throw Error("When 'token' is passed via URL, 'party' must be passed too.");
+      return;
     }
     url.search = '';
     window.history.replaceState(window.history.state, '', url.toString());
+    const tokenCookiePair = document.cookie.split('; ').find(row => row.startsWith('DAMLHUB_LEDGER_ACCESS_TOKEN')) || '';
```
cocreature, line 54: maybe consider something like https://stackoverflow.com/a/25490531/3877993 which is probably a bit more reliable (e.g., this would match on `DAMLHUB_LEDGER_ACCESS_TOKEN2` afaict).

cocreature (approved):
Thank you!

templates/create-daml-app/ui/src/components/LoginScreen.tsx.template
garyverhaegen-da: Updated & tested. @da-tanabe You guys may want to update the docs to match.

da-tanabe: cc @alexmatson-da @sammy-da

Issue #9974, opened 2021-06-11
LF: Structure Errors.


remyhaemmerle-da:
TODOS: 

- [x] On `ContractNotFound`

```
            // We crash here because:
            //  1. You cannot construct a cid yourself in scenarios or
            //     daml script
            //  2. Contract id fetch failures because a contract was
            //     archived or what not are turned into more specific
            //     errors so we never produce ContractNotFound
```

- [x] On `WronglyTypedContract` #10145

```
            //   (MK) This isn’t actually true. You can easily
            //   coerceContractId your way into this error. Not for
            //   this PR though.
```

- [x] On `Pretty` #10470

```
    //  (MK) I’m not convinced the partial transaction here provides any real value.
    //  Afaict the only thing we use it for is to produce slightly different error
    //  if a contract is inactive due to a recursive exercise. We don’t do that in
    //  the scenario service however and I’ve never heard a request for it so I
    //  have a hard time believing that it is really useful here.
```

- [x] Check the error send by IdeLedgerClient. See if we can use only scenario.Error and not SError. (See #10075)

- [x]  Add an test for wrongly type contract with daml script 
      See  #10091 and [WronglyTypedContractId.daml](https://gist.github.com/cocreature/a9a2f55031413f50c7eef5d5bda7cb17)

- [x] try to give more context about missing package, for instance changing the type of MissingPackage to something like:
```MissingPackage(packageIds: Set[PackageId], cultprit: Set[Ref.Identifier])``` (See #10314 and #10418)

**Comment by remyhaemmerle-da** [comment]
cc @daravep

**Comment by stefanobaghino-da** [comment]
On `ContractNotFound`

```
            // We crash here because:
            //  1. You cannot construct a cid yourself in scenarios or
            //     daml script
            //  2. Contract id fetch failures because a contract was
            //     archived or what not are turned into more specific
            //     errors so we never produce ContractNotFound
```

On `WronglyTypedContract`

```
            //   (MK) This isn’t actually true. You can easily
            //   coerceContractId your way into this error. Not for
            //   this PR though.
```

PR #9980, opened 2021-06-14
LF: Structure package loading errors

remyhaemmerle-da:
part of #9974.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -228,7 +228,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
             case Some(pkg) =>
               compiledPackages.addPackage(pkgId, pkg).flatMap(_ => loadPackages(rest))
             case None =>
-              ResultError(Error.Package.Generic(s"package $pkgId not found"))
+              ResultError(Error.Package.MissingPackage(pkgId))
           },
         )
       case Nil =>
@@ -237,7 +237,8 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
 
   @inline
   private[lf] def runSafely[X](
-      handleMissingDependencies: => Result[Unit]
+      methodName: String,
+      handleMissingDependencies: => Result[Unit],
   )(run: => Result[X]): Result[X] = {
     def start: Result[X] =
       try {
@@ -246,7 +247,12 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
         case speedy.Compiler.PackageNotFound(_) =>
           handleMissingDependencies.flatMap(_ => start)
         case speedy.Compiler.CompilationError(error) =>
-          ResultError(Error.Package.Generic(s"CompilationError: $error"))
+          ResultError(
+            Error.Package.Internal(
+              methodName,
+              s"CompilationError: $error",
+            )
+          )
       }
     start
   }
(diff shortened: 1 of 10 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -428,31 +435,17 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
...
-        Either.cond(
-          missingDeps.isEmpty,
-          (),
-          Error.Package.Generic(
-            s"The set of packages ${pkgIds.mkString("{'", "', '", "'}")} is not self consistent, the missing dependencies are ${missingDeps
-              .mkString("{'", "', '", "'}")}."
-          ),
-        )
-      }
+      unknownPackages = pkgIds.filterNot(pkgs.isDefinedAt)
```
cocreature, line 445: Looking at the usages of this function I’m not entirely sure the API here makes sense. We pass in a `Set[PackageId]` and a `Map[PackageId, Package]` and then assert that the set contains the key set of the map. We should be able to just accept the map only, eliminate this error and afaict all callsites would work well with that. Probably something for a separate PR though.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
```diff
@@ -54,7 +54,7 @@ private[engine] final class Preprocessor(compiledPackages: MutableCompiledPackag
                 )
               } yield r
             case None =>
-              ResultError(Error.Package.Generic(s"Couldn't find package $pkgId"))
+              ResultError(Error.Package.MissingPackage(pkgId))
```
cocreature, line 57: I’m a bit worried that there seem no tests for any of the other errors.  Am I missing something? At least self consistency seems worth testing somewhere™.

cocreature (approved):
LGTM with a few comments, thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
cocreature: https://github.com/digital-asset/daml/pull/9986

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -246,7 +247,12 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
         case speedy.Compiler.PackageNotFound(_) =>
           handleMissingDependencies.flatMap(_ => start)
         case speedy.Compiler.CompilationError(error) =>
-          ResultError(Error.Package.Generic(s"CompilationError: $error"))
+          ResultError(
+            Error.Package.Internal(
+              methodName,
+              s"CompilationError: $error",
```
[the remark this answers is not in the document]
remyhaemmerle-da: In fact, this is not a `Package.Error`, I move it to `Preprocessing.Error`.
Good catch,

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
remyhaemmerle-da: I added tests and spotted/fixed a bug in 2ff96785562ac41f2793aff38ec889b6075595a0

PR #9986, opened 2021-06-14
Drop pkgIds argument from Engine.validatePackages

cocreature:
Doesn’t do anything useful, we just use it to validate that it matches
the keys of pkgs.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
@@ -188,18 +188,10 @@ final private[kvutils] class PackageCommitter(
   ): Either[String, Map[Ref.PackageId, Ast.Package]] =
     metrics.daml.kvutils.committer.packageUpload.decodeTimer.time { () =>
       type Result = Either[List[String], Map[Ref.PackageId, Ast.Package]]
-      val knownPackages = engine.compiledPackages().packageIds.toSet[String]
-
       archives
         .foldLeft[Result](Right(Map.empty)) { (acc, arch) =>
           try {
-            if (knownPackages(arch.getHash)) {
-              // If the package is already known by the engine, we don't decode it but still verify its hash.
-              lf.archive.Reader.HashChecker.decodeArchive(arch)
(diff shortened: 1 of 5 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
```diff
@@ -188,18 +188,10 @@ final private[kvutils] class PackageCommitter(
...
         .foldLeft[Result](Right(Map.empty)) { (acc, arch) =>
           try {
-            if (knownPackages(arch.getHash)) {
-              // If the package is already known by the engine, we don't decode it but still verify its hash.
-              lf.archive.Reader.HashChecker.decodeArchive(arch)
-              acc
-            } else {
-              acc.map(_ + lf.archive.Decode.decodeArchive(arch))
-            }
+            acc.map(_ + lf.archive.Decode.decodeArchive(arch))
```
cocreature, line 194: This is slightly worse in terms of performance. However, I think this is a cost worth paying for the decrease in complexity and the extra safety (much harder to skip package validation accidentally) for two reasons:
1. We really only pay decoding costs/validation here. The package is still not stored to the ledger. Deduplication happens later.
2. Package uploads are extremely rare. We are talking about days if not weeks between uploads so this is absolutely not performance critical.

remyhaemmerle-da (commented):
Thanks.

fabiotudone-da (approved):
LGTM but I'd wait for at least another KV review before merging, if you're not in a hurry.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
remyhaemmerle-da: and it actually fix a bug catched by the test I add in 61da6c6c113f32f3c7aaa86eada5dfdd44e8256a

SamirTalwar-DA (approved):
Seems sensible; no complaints. Thanks for catching the bug!

miklos-da (approved):

PR #9993, opened 2021-06-14
Support reinterpretation for failing commands

nickchapman-da:
Support reinterpretation for failing commands

- `Engine.reinterpret` calls new compile method
- `Compiler.compileCommandForReinterpretation`
- which wraps `catchEverything` around the compiled command, so a rollback node is produced.
- test that uncatchable errors remains failures

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -9,7 +9,7 @@ import com.daml.lf.data._
 import com.daml.lf.data.Ref.{PackageId, ParticipantId, Party}
 import com.daml.lf.language.Ast._
 import com.daml.lf.speedy.{InitialSeeding, PartialTransaction, Pretty, SError, SExpr}
-import com.daml.lf.speedy.Speedy.{Machine, Validating}
+import com.daml.lf.speedy.Speedy.Machine
 import com.daml.lf.speedy.SResult._
 import com.daml.lf.transaction.{SubmittedTransaction, Transaction => Tx}
 import com.daml.lf.transaction.Node._
@@ -96,7 +96,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       .preprocessCommands(cmds.commands)
       .flatMap { case (processedCmds, globalCids) =>
         interpretCommands(
-          validating = Validating.Off,
+          validating = false,
           submitters = submitters,
           commands = processedCmds,
           ledgerTime = cmds.ledgerEffectiveTime,
@@ -144,7 +144,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       sexpr = compiledPackages.compiler.unsafeCompileForReinterpretation(speedyCommand)
       // reinterpret is never used for submission, only for validation.
       result <- interpretExpression(
-        validating = Validating.On,
+        validating = true,
         submitters = submitters,
         sexpr = sexpr,
         ledgerTime = ledgerEffectiveTime,
@@ -167,7 +167,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       commandsWithCids <- preprocessor.translateTransactionRoots(tx.transaction)
       (commands, globalCids) = commandsWithCids
       result <- interpretCommands(
-        validating = Validating.On,
+        validating = true,
         submitters = submitters,
         commands = commands,
         ledgerTime = ledgerEffectiveTime,
@@ -259,7 +259,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
 
   // command-list compilation, followed by interpretation
   private[engine] def interpretCommands(
-      validating: Validating,
+      validating: Boolean,
       submitters: Set[Party],
       commands: ImmArray[speedy.Command],
       ledgerTime: Time.Timestamp,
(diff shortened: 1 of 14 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -1520,6 +1525,24 @@ private[lf] final class Compiler(
 
   private val SEUpdatePureUnit = unaryFunction(_ => SEValue.Unit)
 
+  private val handleEverything: SExpr = SBSome(SEUpdatePureUnit)
```
remyhaemmerle-da, line 1528:
```suggestion
  private[this] val handleEverything: SExpr = SBSome(SEUpdatePureUnit)
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -141,11 +141,12 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
     for {
       commandWithCids <- preprocessor.preprocessCommand(command)
       (speedyCommand, globalCids) = commandWithCids
+      sexpr = compiledPackages.compiler.unsafeCompileForReinterpretation(speedyCommand)
```
remyhaemmerle-da, line 144: Maybe we can extend the comment to explain what's happening when the command fails with an unhanded exception ?

remyhaemmerle-da:
cc @oliverse-da @andreaslochbihler-da

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ReinterpretTest.scala
```diff
@@ -0,0 +1,179 @@
...
+import com.daml.lf.command._
+import com.daml.lf.transaction.test.TransactionBuilder.assertAsVersionedValue
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.EitherValues
+import org.scalatest.wordspec.AnyWordSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.language.implicitConversions
+
+class ReinterpretTest
```
cocreature, line 26: Can we get some tests for the edge cases here, I’m particularly interested in uncatchable errors due to
1. version checks
2. because they are uncatchable in general, e.g. failed DAML-LF ensure, error call, ….

cocreature (approved):
lgtm provided we extend the tests before merging, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
nickchapman-da: done

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
nickchapman-da: done

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ReinterpretTest.scala
nickchapman-da: failure ensure -- done
version check -- still to do
error call -- do we have an _uncatchable_ error primitive?

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ReinterpretTest.scala
```diff
@@ -0,0 +1,203 @@
...
+              (Some[Ref.Name]("v"), ValueInt64(42)), // ok (< 50)
+            ),
+          ),
+        )
+      }
+      val tx = reinterpretCommand(theCommand)
+      Shape.ofTransaction(tx.transaction) shouldBe Top(Create())
+    }
+
+    "not be a rollback for a create which fails an ensure" in {
```
cocreature, line 151: This is not the case I was trying to get at. I mean the following:
1. Exercise C
2. In that exercise create a contract with an ensure that returns `False` (so not one produced by the compiler, those throw so they are catchable).

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ReinterpretTest.scala
cocreature: Yes, Daml-LF’s error primitive is uncatchable. Daml’s `error` function does not use that primitive so you have to hand-write that to test it.
I think testing one uncatchable error is sufficient so either failed ensure (see my comment below, I think your test isn’t quite testing the right case), error call, ordering failure, ….

nickchapman-da: Removed the previous useless ensure test.
And added a test which provokes an uncatchable error by creating a broken `Ord` instance (as we discussed),

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ReinterpretTest.scala
nickchapman-da: didn't do this, as we discussed

nickchapman-da:
Added tests that uncatchable errors remain as failures for reinterpretation.

nickchapman-da:
We are ready to merge now

cocreature (approved):
Thanks for the tests!

PR #9995, opened 2021-06-14
[JSON-API] Add information from the jwtpayload to the logging context

### LINKED ISSUE #9982: Include token information in JSON API logging context
The token a user supplies provides us with a lot of useful informaton that we should include in logs. Specifically:
1. `actAs`
2. `readAs`
3. `applicationId`
4. less useful but no harm in including it: `ledgerId`

realvictorprm:
This fixes #9982.

Only problematic thing with my changes is that IntelliJ now complains a bit more although it compiles fine:
image

However, I think this is the right direction but I'm not sure if that is already enough, what do you think @cocreature?

changelog_begin

- [JSON API] For applicable requests actAs, readAs, applicationId & ledgerId are included in the log context

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -123,7 +123,7 @@ class Endpoints(
     }
   }
 
-  def withJwtPayloadLoggingContext[T[_], A](jwtPayload: JwtPayloadG[T])(
+  def withJwtPayloadLoggingContext[A](jwtPayload: JwtPayloadG)(
       fn: LoggingContextOf[JwtPayloadTag with InstanceUUID with RequestID] => A
   )(implicit lc: LoggingContextOf[InstanceUUID with RequestID]): A =
     withEnrichedLoggingContext(
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -21,7 +21,7 @@ import com.daml.lf
 import com.daml.http.ContractsService.SearchResult
 import com.daml.http.EndpointsCompanion._
 import com.daml.scalautil.Statement.discard
-import com.daml.http.domain.{JwtPayload, JwtWritePayload, TemplateId}
+import com.daml.http.domain.{JwtPayload, JwtPayloadG, JwtPayloadTag, JwtWritePayload, TemplateId}
 import com.daml.http.json._
 import com.daml.http.util.Collections.toNonEmptySet
 import com.daml.http.util.FutureUtil.{either, eitherT}
@@ -123,11 +123,11 @@ class Endpoints(
     }
   }
 
-  def withJwtPayloadLoggingContext[A](jwtPayload: JwtPayload)(
-      fn: LoggingContextOf[JwtPayload with InstanceUUID with RequestID] => A
+  def withJwtPayloadLoggingContext[T[_], A](jwtPayload: JwtPayloadG[T])(
+      fn: LoggingContextOf[JwtPayloadTag with InstanceUUID with RequestID] => A
   )(implicit lc: LoggingContextOf[InstanceUUID with RequestID]): A =
     withEnrichedLoggingContext(
-      LoggingContextOf.label[JwtPayload],
+      LoggingContextOf.label[JwtPayloadTag],
       Map(
         "ledger_id" -> jwtPayload.ledgerId.toString,
         "act_as" -> jwtPayload.actAs.toString,
@@ -143,7 +143,9 @@ class Endpoints(
           Jwt,
           JwtWritePayload,
           JsValue,
-      ) => LoggingContextOf[JwtWritePayload with InstanceUUID with RequestID] => ET[T[ApiValue]]
+      ) => LoggingContextOf[JwtPayloadTag with InstanceUUID with RequestID] => ET[
+        T[ApiValue]
+      ]
   )(implicit
       lc: LoggingContextOf[InstanceUUID with RequestID],
       ev1: JsonWriter[T[JsValue]],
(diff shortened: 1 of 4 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -118,73 +119,102 @@ class Endpoints(
     }
   }
 
+  def withJwtPayloadLoggingContext[A, B](jwtPayload: JwtPayload)(
+      fn: LoggingContextOf[JwtPayload with InstanceUUID with RequestID] => A
+  )(implicit lc: LoggingContextOf[InstanceUUID with RequestID]): A =
```
S11001001, line 124: If you want to determine `B` inferrably, a label can sit alongside `jwtPayload`.

```suggestion
      fn: LoggingContextOf[JwtPayload with B] => A
  )(implicit lc: LoggingContextOf[B]): A =
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -118,73 +119,102 @@ class Endpoints(
...
+  // I think this is not the best name.
+  // Naming things is hard :/
+  def handleCommand[T[_]](req: HttpRequest)(
+      fn: (
+          Jwt,
+          JwtWritePayload,
+          JsValue,
+      ) => LoggingContextOf[JwtWritePayload with InstanceUUID with RequestID] => ET[T[ApiValue]]
+  )(implicit
+      lc: LoggingContextOf[InstanceUUID with RequestID],
```
S11001001, line 144: Likewise.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -118,73 +119,102 @@ class Endpoints(
...
+  )(implicit lc: LoggingContextOf[InstanceUUID with RequestID]): A =
+    withEnrichedLoggingContext(
+      LoggingContextOf.label[JwtPayload],
+      Map(
+        "ledger_id" -> jwtPayload.ledgerId.toString,
+        "act_as" -> jwtPayload.actAs.toString,
+        "application_id" -> jwtPayload.applicationId.toString,
+        "read_as" -> jwtPayload.readAs.toString,
+      ),
+    ).run(lc => fn(lc))
```
S11001001, line 133:
```suggestion
    ).run(fn)
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: True, this would be better. Thanks!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: Just noticed that this doesn't work because the other functions require the specific logging context within that function already. Going to revert the changes.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -118,73 +119,102 @@ class Endpoints(
...
+        "ledger_id" -> jwtPayload.ledgerId.toString,
+        "act_as" -> jwtPayload.actAs.toString,
+        "application_id" -> jwtPayload.applicationId.toString,
+        "read_as" -> jwtPayload.readAs.toString,
+      ),
+    ).run(fn)
+
+  // I think this is not the best name.
+  // Naming things is hard :/
+  def handleCommand[T[_]](req: HttpRequest)(
```
cocreature, line 137: Seems like there is a fair bit of duplication between the write & read endpoints. Can we share this? In the end both tokens can provide us with those fields so we should be able to make them extend a common trait (maybe we even have one already I forgot).

realvictorprm: I thought about union both jwtpayloads with a trait, however because it hasn't been done in the past already I hesitated to do it as both are similar but used in different scenarios.

However I would be much happier too with adding a trait and removing that duplicate code.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
S11001001: That's what the `label` is for, if you're referring to ambiguity.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: adressed with 0b0c824 & 24b0c33

ledger-service/http-json/src/main/scala/com/digitalasset/http/domain.scala
```diff
@@ -53,15 +53,26 @@ object domain {
...
 
+  trait JwtPayloadTag
+
+  trait JwtPayloadG[T[_]] {
+    val ledgerId: LedgerId
+    val applicationId: ApplicationId
+    val readAs: List[Party]
+    val actAs: T[Party]
+    val parties: OneAnd[Set, Party]
+  }
```
cocreature, line 64:
```suggestion
  trait JwtPayloadTag

  trait JwtPayloadG {
    val ledgerId: LedgerId
    val applicationId: ApplicationId
    val readAs: List[Party]
    val actAs: List[Party]
    val parties: OneAnd[Set, Party]
  }
```
and then rename the field in `JwtWritePayload` to `submitter` or something and define `actAs` to return the list. Seems like a much simpler and also more useful abstraction since you can actually use it generically with something more htan calling `toSrting` on it.

victormueller-da: True, that is more useful :thinking: Will do

realvictorprm: Adressed with 8ec7126

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -118,77 +119,97 @@ class Endpoints(
...
+      Map(
+        "ledger_id" -> jwtPayload.ledgerId.toString,
+        "act_as" -> jwtPayload.actAs.toString,
+        "application_id" -> jwtPayload.applicationId.toString,
+        "read_as" -> jwtPayload.readAs.toString,
+      ),
+    ).run(fn)
+
+  // I think this is not the best name.
+  // Naming things is hard :/
```
cocreature, line 136:
```suggestion
```

cocreature (approved):
Thanks!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: Many thanks, forgot that one :)

realvictorprm:
go mergify goooo :smile:

realvictorprm:
no you got to be kidding me

need to resolve that merge conflict first :fire:

realvictorprm:
now, finally resolved. Shouldn't take long till this is in :grin:

PR #9997, opened 2021-06-14
prepare for upgrade to Scalaz 7.3.3

S11001001:
This makes every Scalaz 7.2-compatible change from #9791; in other words, we remove 7.2-only constructs in favor of ones that are forward-compatible to 7.3.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks, lgtm!
I somewhat wonder if we should replace some/most uses of `\/` with `Either`. It doesn’t seem like we really use much of what scalaz provides here.

daml-lf/interface/src/main/scala/com/digitalasset/daml/lf/iface/reader/InterfaceReader.scala
```diff
@@ -29,15 +29,15 @@ object InterfaceReader {
   private def errorMessage(ctx: QualifiedName, reason: String): String =
     s"Invalid data definition: $ctx, reason: $reason"
 
-  private def invalidDataTypeDefinition(
+  private def invalidDataTypeDefinition[Bot](
```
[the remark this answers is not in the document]
S11001001: Bottom [type], or `\bot` in TeX.

extractor/src/main/scala/com/digitalasset/extractor/ledger/types/LedgerValue.scala
```diff
@@ -123,4 +127,8 @@ object LedgerValue {
       }
       .disjunction
   }
+
+  private[this] implicit final class `either covariant`[A](private val self: A) extends AnyVal {
+    def right[L, U >: A]: L \/ U = \/-(self)
```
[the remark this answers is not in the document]
S11001001: This is used in code above, replacing the Scalaz extension method of the same name.

language-support/scala/codegen/src/main/scala/com/digitalasset/codegen/CodeGen.scala
```diff
@@ -234,14 +236,16 @@ object CodeGen {
...
 
     // Finally we generate the "event decoder" and "package ID source"
     val specials =
       Seq(lf.EventDecoderGen.generate(util, templateIds.keySet), lf.PackageIDsGen.generate(util))
 
-    val specialPlans = specials map { case (fp, t) => \/-((None, fp, t)) }
+    val specialPlans = specials map { case (fp, t) => \/-((None, fp, t)): FilePlan }
 
     filePlans ++ specialPlans
   }
```
S11001001, line 251: > I somewhat wonder if we should replace some/most uses of `\/` with `Either`. It doesn’t seem like we really use much of what scalaz provides here.

@cocreature It depends on whether you agree with the underlying rationale of scalaz/scalaz#1469. The best illustration locally is this function, `produceTemplateAndTypeFilesLF`.

As a rule, the numerous variables in this function live in a tapestry of different subtypes of the goal type. Your `ntdRights` lives in a "righty" type and `tmplLefts` lives in a "lefty" type, for example. It is only later that we unify the types via covariance, and this proceeds through later in the file.

So if at any point you inspect the various variables, you will see that they exist not at the type we necessarily need them to be, but some subtype that happens to express how they were constructed. We know it works together because when we _unify_, it lazily works. However, unless you're familiar with how the goal relates to the components' types, if you change something upstream, the occurrence of the failure (at unification) bears less relationship to the part that actually needs to change. (I am speaking generally; I would hope at this point that Scala codegen's upstream will not change significantly. This is simply the most elaborate use of "natural" covariant unification I am aware of related to eithers in the daml codebase.)

By contrast, consider the situation with invariant `\/`. The type ascriptions added here are sort of declarations of how each component variable needs to arrange itself to fit into the goal type. So we ascribe `specialPlans` as a declaration that its type may not merely exist in the subtype lattice; its plans must declaratively be `FilePlan`s, so to speak. If I do not do this, the resulting error is [mostly] apparent in how it points to `specialPlans`'s problematic [sub]type, so should such an upstream change occur, it is [mostly] apparent what part of the unification tree needs to change.

So it really boils down to whether you consider the additional explicitness useful scaffolding, or a distraction. Or, whether the magic of *variant unification is good magic or wicked magic. (There are additional issues related to higher-kinded types, but that isn't strictly relevant to what I've just described.)

cocreature: In this PR it feels more like noise to me. Maybe I’ve written Scala for so long that I’m starting to like subtyping :scream_cat:

S11001001: It certainly felt like noise while working on the PR 😆 

However, I also think I'm operating from the privileged position of updating unification code _that I happen to know already works_, which is a little different from trying to assemble new lattices.

PR #10005, opened 2021-06-15
[JSON-API] Log templateId & choice name (if present) on command submissions in the json API

### LINKED ISSUE #9983: Log template id and (if present) choice name on command submissions in the JSON API
When submitting a command to the ledger from the JSON API we should also include the template id and (for exercise/create-and-exercise) the choice name. Probably wouldn’t throw it in the context but include it in the string just like we do for the type of command.

realvictorprm:
fixes #9983

changelog_begin

- [JSON API] The template id & choice name (if present) are now logged on command submissions in the Json API (at trace level)

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
@@ -47,7 +47,7 @@ class CommandService(
   )(implicit
       lc: LoggingContextOf[InstanceUUID with RequestID]
   ): Future[Error \/ ActiveContract[lav1.value.Value]] = {
-    logger.trace("sending create command to ledger")
+    logger.trace(s"sending create command to ledger, templateId: ${input.templateId}")
     val command = createCommand(input)
     val request = submitAndWaitRequest(jwtPayload, input.meta, command, "create")
     val et: ET[ActiveContract[lav1.value.Value]] = for {
@@ -64,7 +64,8 @@ class CommandService(
   )(implicit
       lc: LoggingContextOf[InstanceUUID with RequestID]
   ): Future[Error \/ ExerciseResponse[lav1.value.Value]] = {
-    logger.trace("sending exercise command to ledger")
+    logger.trace(s"sending exercise command to ledger, templateId: ${input.reference
+      .fold(_._1, _._1)}, choice: ${input.choice}")
     val command = exerciseCommand(input)
     val request = submitAndWaitRequest(jwtPayload, input.meta, command, "exercise")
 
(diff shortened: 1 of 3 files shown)


stefanobaghino-da (commented):
In general when logging some form of data it's probably better to put it in the logging context so that it's available as part of the structured part of the log entry.

cocreature:
> In general when logging some form of data it's probably better to put it in the logging context so that it's available as part of the structured part of the log entry.

I originally suggested not putting it in the logging context since it seemed like that might be getting a bit out of hand especially for things that are only logged once. But you have much more experience in that area so happy to follow your guidance and move it to the logging context if you think it makes sense.

stefanobaghino-da:
> > In general when logging some form of data it's probably better to put it in the logging context so that it's available as part of the structured part of the log entry.
> 
> I originally suggested not putting it in the logging context since it seemed like that might be getting a bit out of hand especially for things that are only logged once. But you have much more experience in that area so happy to follow your guidance and move it to the logging context if you think it makes sense.

The idea would be that one should be able to easily process logs in structured form without parsing the message, that's why I was suggesting that. However I don't have a strong opinion if for any reason we prefer to have this in the message.

realvictorprm:
OK folks, I'm also in favour of including it in the context, because I also had my thoughts about the usability here.

I think generally it is actually best to include everything in the context which later most likely is going to be matched for during debugging/investigating an issue.

ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
```diff
@@ -40,65 +42,88 @@ class CommandService(
...
   )(implicit
       lc: LoggingContextOf[InstanceUUID with RequestID]
-  ): Future[Error \/ ExerciseResponse[lav1.value.Value]] = {
-    logger.trace("sending exercise command to ledger")
-    val command = exerciseCommand(input)
-    val request = submitAndWaitRequest(jwtPayload, input.meta, command, "exercise")
+  ): Future[Error \/ ExerciseResponse[lav1.value.Value]] =
+    withTemplateChoiceLoggingContext(input.reference.fold(_._1, _._1), input.choice).run {
+      implicit lc =>
+        logger.trace(s"sending exercise command to ledger")
```
stefanobaghino-da, line 91:
```suggestion
        logger.trace("sending exercise command to ledger")
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
```diff
@@ -40,65 +42,88 @@ class CommandService(
...
-      response <- logResult(
-        Symbol("createAndExercise"),
-        submitAndWaitForTransactionTree(jwt, request),
-      )
-      exerciseResult <- either(exerciseResult(response))
-      contracts <- either(contracts(response))
-    } yield ExerciseResponse(exerciseResult, contracts)
+  ): Future[Error \/ ExerciseResponse[lav1.value.Value]] =
+    withTemplateChoiceLoggingContext(input.templateId, input.choice).run { implicit lc =>
+      logger.trace(s"sending create and exercise command to ledger")
```
stefanobaghino-da, line 113:
```suggestion
      logger.trace("sending create and exercise command to ledger")
```

realvictorprm:
Side note, I try to only force push if necessary, which currently is limited to rebasing.

stefanobaghino-da (approved):

ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
realvictorprm: great, thanks!

cocreature (approved):
tHANKS!

PR #10009, opened 2021-06-15
DPP-394 h2 support for append only schema

nmarton-da:
### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGSchema.scala
```diff
@@ -288,16 +287,20 @@ private[postgresql] object PGSchema {
...
-    override def booleanOptional[FROM, _](
-        extractor: FROM => Option[Boolean]
-    ): Field[FROM, Option[Boolean], _] =
-      PGBooleanOptional(extractor)
+    override def packageInsert(
+        tableName: String
+    )(fields: (String, Field[DbDto.Package, _, _])*): Table[DbDto.Package] =
+      PGTable.transposedInsertWithSuffix(tableName, "on conflict (package_id) do nothing")(
+        fields: _*
+      )
```
rautenrieth-da, line 45: Do you think it would make sense to make this more generic? Something like:
```scala
// Assumes the primary key uniquely identifies the content of the row.
// Conflicts on the primary key are safely and silently ignored.
def idempotentInsert(tableName: String, primaryKey: String)(...) =
  PGTable.transposedInsertWithSuffix(tableName, s"on conflict ($primaryKey) do nothing")
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
```diff
@@ -0,0 +1,69 @@
...
+    extends Field[FROM, Option[Int], java.lang.Integer] {
+  override def convert: Option[Int] => Integer = _.map(Int.box).orNull
+}
+
+case class Bigint[FROM](extract: FROM => Long) extends TrivialField[FROM, Long]
+
+case class SmallintOptional[FROM](extract: FROM => Option[Int])
+    extends Field[FROM, Option[Int], java.lang.Integer] {
+  override def selectFieldExpression(inputFieldName: String): String =
+    s"$inputFieldName::smallint"
```
rautenrieth-da, line 60: The double colon operator is not standard (not that "standard" would mean much when it comes to SQL dialects). I assume this is again a case where we refactor as necessary if we add more database backends?

nmarton-da: No!! This was completely not intentional, thanks for spotting that!
Unfortunately H2 supports the :: style cast (which I could not find evidence of in the docs), so even on my branch where H2 tests are green I could not have spotted that.
Changed it, and created/wired again the PGSmallintOptional for postgres

dasormeter: the sql standard syntax is ```cast ( $inputFieldName as smallint)``` and that should work for oracle and (presumably) h2 so would be the better cross-platform way.  i believe the double colon is just postgres shorthand for the same

nmarton-da: correct, the double colon is for postgres for sure but also for others (hence my surprise seeing that work with H2)
but the main thing here to keep it simple: in this case for H2 we do not even need the cast, we need it only for postgres, so a specialization in the postgres strategy is called for IMHO

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGSchema.scala
nmarton-da: good idea, will do

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGTable.scala
```diff
@@ -37,9 +37,15 @@ private[postgresql] case class PGTable[FROM](
   def prepareData(in: Vector[FROM]): Array[Array[_]] =
     fields.view.map(_._2.toArray(in)).toArray
 
-  def setupData(data: Array[Array[_]], preparedStatement: PreparedStatement): Unit =
-    fields.indices.foreach { i =>
-      preparedStatement.setObject(i + 1, data(i))
+  def executeInsert(data: Array[Array[_]], connection: Connection): Unit =
+    if (data(0).length > 0) { // data(0) accesses the array of data for the first column of the table. This is safe because tables without columns are not supported. Also because of the transposed data-structure here all columns will have data-arrays of the same length.
```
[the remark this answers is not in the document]
nmarton-da: sure thing

rautenrieth-da (approved):

PR #10010, opened 2021-06-15
Log context of all updates written to the database

mziolekda:
CHANGELOG_BEGIN
For every update in the index db log the full context at the INFO level.
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
+++ b/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
@@ -35,7 +35,7 @@ object Update {
       newConfiguration: Configuration,
   ) extends Update {
     override def description: String =
-      s"Configuration change '$submissionId' from participant '$participantId' accepted with configuration: $newConfiguration"
+      "configuration"
   }
 
   /** Signal that a configuration change submitted by this participant was rejected.
@@ -48,7 +48,7 @@ object Update {
       rejectionReason: String,
   ) extends Update {
     override def description: String = {
-      s"Configuration change '$submissionId' from participant '$participantId' was rejected: $rejectionReason"
+      "configuration rejection"
     }
   }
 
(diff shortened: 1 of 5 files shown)


ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
```diff
@@ -206,7 +206,7 @@ object Update {
       reason: RejectionReason,
   ) extends Update {
     override def description: String = {
-      s"Reject command ${submitterInfo.commandId}: $reason"
+      "transaction rejection"
```
gerolf-da, line 209: I think it wouldn't hurt to log the details. Otherwise the description makes an assumption on where it can be meaningfully used (i.e. only in an environment that uses logging context).

mziolekda: We don't use it anywhere atm.
The details really are just a repetition of the logging context in a less organized manner. I think if we have made that investment to log using context, let's just stick to it.

mziolekda: These descriptions used to show up in one trace log in the entire code-base and they looked awful

nmarton-da (approved):
LGTM

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v1/Update.scala
cocreature: fwiw we have an extension of the logging context which adds a type parameter exactly to solve this issue: if you don’t have the right context it simply won’t compile https://github.com/digital-asset/daml/blob/adeba206594b8ec3437efa99a44c4a5572d3ec6e/libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingContextOf.scala#L10.

mziolekda: Thanks for the pointer. The story doesn't end here for us. Tudor has started on another PR that makes the logging context usage in indexer more robust. That's where we will switch to LoggingContextOf

PR #10025, opened 2021-06-16
Force css-what resolution in compiler/daml-extension & fully regenera…

realvictorprm:
…te the lock file

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-extension/package.json
+++ b/compiler/daml-extension/package.json
@@ -152,5 +152,8 @@
         "@types/vscode": "1.39",
         "typescript": "3.6.3",
         "vsce": "1.88.0"
+    },
+    "resolutions": {
+        "**/css-what": "^5.0.1"
     }
 }
--- a/compiler/daml-extension/package.json
+++ b/compiler/daml-extension/package.json
@@ -152,8 +152,5 @@
         "@types/vscode": "1.39",
         "typescript": "3.6.3",
         "vsce": "1.88.0"
-    },
-    "resolutions": {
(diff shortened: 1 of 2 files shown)


realvictorprm:
ok testing this is somewhat difficulter than I expected. I build it via bazel and intended to install a version locally to see if everything works but I cannot find the vsix :|

compiler/daml-extension/package.json
```diff
@@ -152,5 +152,8 @@
         "@types/vscode": "1.39",
         "typescript": "3.6.3",
         "vsce": "1.88.0"
+    },
+    "resolutions": {
```
cocreature, line 156: Please only use resolutions when we have to force a version because a dependency has garbage bounds. in this case this is pulled in via a bound `    css-what "^5.0.0"` so we don’t have to mess with any bounds. Just deleting the yarn lock file and regenerating should do the trick.

cocreature:
> ok testing this is somewhat difficulter than I expected. I build it via bazel and intended to install a version locally to see if everything works but I cannot find the vsix :|

1. `daml-sdk-head --skip-jars`
2. `daml-head new foobar`
3. `cd foobar && daml studio --replace=always`

realvictorprm:
Somehow I still often enough forget about just using `daml-sdk-head` :D

cocreature (approved):
lgtm provided manual tests work

realvictorprm:
Works locally for me, everything seems normal with the extension. I consider it to work fine then.

PR #10027, opened 2021-06-16
I9996 value limits it fails oracle

danielporterda:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
@@ -67,7 +67,7 @@ private[dao] object SqlFunctions {
       val NumExtraChars = 20
       val OracleMaxStringLiteralLength = 4000
 
-      val groupedParties = parties.toList.sorted.foldLeft((List.empty[List[String]], 0))({
+      val groupedParties = parties.foldLeft((List.empty[List[String]], 0))({
         case ((prev, currentLength), party) =>
           if (
             currentLength + party.length + NumCharsBetweenParties > OracleMaxStringLiteralLength
--- a/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
@@ -62,8 +62,29 @@ private[dao] object SqlFunctions {
   object OracleSqlFunctions extends SqlFunctions {
     // TODO https://github.com/digital-asset/daml/issues/9493
     // This is likely extremely inefficient due to the multiple full tablescans on unindexed varray column
-    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String =
-      s"""JSON_EXISTS($arrayColumn, '$$[*]?(@ in ("${parties.mkString("""", """")}"))')"""
+    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String = {
+      val NUM_CHARS_BETWEEN_PARTIES = 3
+      val NUM_EXTRA_CHARS = 20
+      val ORACLE_MAX_STRING_LITERAL_LENGTH = 4000
+      parties.foreach { p => if(p.length > 3980) {
+        throw new Exception("Parties cannot be longer than 3980 characters.")
+      }}
+      val groupedParties = parties.toList.sorted.foldLeft((List.empty[List[String]], 0))({case ((prev, currentLength), party) =>
+        if(currentLength + party.length + NUM_CHARS_BETWEEN_PARTIES > ORACLE_MAX_STRING_LITERAL_LENGTH) {
+          (List(party) :: prev, party.length + NUM_EXTRA_CHARS)
+        }
+        else {
+          prev match {
+            case h :: tail => ((party :: h) :: tail, currentLength + party.length + NUM_CHARS_BETWEEN_PARTIES)
+            case Nil => (List(party) :: Nil, party.length + NUM_EXTRA_CHARS)
+          }
+        }
+      })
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
```diff
@@ -62,8 +62,31 @@ private[dao] object SqlFunctions {
   object OracleSqlFunctions extends SqlFunctions {
     // TODO https://github.com/digital-asset/daml/issues/9493
     // This is likely extremely inefficient due to the multiple full tablescans on unindexed varray column
-    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String =
-      s"""JSON_EXISTS($arrayColumn, '$$[*]?(@ in ("${parties.mkString("""", """")}"))')"""
+    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String = {
+      val NumCharsBetweenParties = 3
```
dasormeter, line 66: Is it worth actually parsing to get character lengths from the live expression on line 86 so these checks don't rot if the expression is changed in future?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
```diff
@@ -62,8 +62,31 @@ private[dao] object SqlFunctions {
...
+      val NumExtraChars = 20
+      val OracleMaxStringLiteralLength = 4000
+
+      val groupedParties = parties.toList.sorted.foldLeft((List.empty[List[String]], 0))({
+        case ((prev, currentLength), party) =>
```
dasormeter, line 71: why prev rather than next?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
```diff
@@ -62,8 +62,31 @@ private[dao] object SqlFunctions {
...
+          if (
+            currentLength + party.length + NumCharsBetweenParties > OracleMaxStringLiteralLength
+          ) {
+            (List(party) :: prev, party.length + NumExtraChars)
+          } else {
+            prev match {
```
dasormeter, line 77: worth adding a unit test for this?

ledger/participant-integration-api/src/main/scala/platform/store/dao/events/SqlFunctions.scala
```diff
@@ -62,8 +62,31 @@ private[dao] object SqlFunctions {
   object OracleSqlFunctions extends SqlFunctions {
     // TODO https://github.com/digital-asset/daml/issues/9493
     // This is likely extremely inefficient due to the multiple full tablescans on unindexed varray column
-    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String =
-      s"""JSON_EXISTS($arrayColumn, '$$[*]?(@ in ("${parties.mkString("""", """")}"))')"""
+    override def arrayIntersectionWhereClause(arrayColumn: String, parties: Set[Party]): String = {
```
nmarton-da, line 65: TLDR: do you really need this now? this might not needed later

the root problem here might be something else: this method for constructing the query is totally static, meaning: it is not like we have a prepare statement with something like ? in the middle which we pass to the prepared statement as a data -what we should be doing-, but rather we have a prepared statement where we statically string interpolate all the data (look for the # where this is used).
I can very well imagine if we would be able to pass the all parties as an array, we wouldn't have this problem.
Which would mean if we change this approach accordingly (this is on our horizon), this change will be lost.
So question is how bad you need this working on a short run? (if you do I am completely happy to go with it temporarily, although the complexity here scares me a bit)

dasormeter: @nmarton-da without this fix the ValueLimitsIt.scala fails on oracle for sandbox-on-x (that is what triggered the discovery) so this is defensiveness to deal with a big party list of submitters without requiring a hard limit.

@danielporterda also contemplated moving to a `?` prepared statement parameter replacement approach but that would have had a bigger knock-on impact to interfaces and across the stack which we deemed too impactful given all the moving parts happening now.

I would vote for merging this on the existing Oracle schema and then when we can revisit as we implement the equiv on the new append only.

danielporterda: yeah, I'm certainly not in love with this course of action, and I share your assessment of what the "root problem" is, but it felt like best of bad options for the moment. As Brian says, it's motivated by making the test pass. The other two possible implementations are

1. use a prepared statement as you suggest, or
2. render the JSON fields as a materialized view and do a subquery

It's not totally clear if a prepared statement will get around the issues with the string length, but that's worth investigating. The reason it felt like a non-starter was having to change a bunch of interfaces while the append only work is in motion.

As far as rendering the JSON array fields as a materialized view, this is appealing because it (should) mean that we can create an index on them, however @stephencompall-DA apparently ran into some issues with this that may or may not have implications about correctness - some issue with filtering out duplicates. This is second hand, but the upshot was that it was a larger change that wasn't guaranteed to work, and wasn't clear how it was going to play with your changes.

In general, I'm largely opposed to anything that depends on getting arithmetic correct, given how much evidence there is that programmers are terrible at it. Hard to get right, harder still to modify later. +1 to improving this later.

danielporterda: No worries, if you see me doing what looks like a footgun in the future, please feel free to speak up. Would prefer to explain myself over accidentally doing bad stuff. I'm not always so verbose, scout's honor :)

dasormeter (approved):
tested that this fixes failing ValueLimitsIT

PR #10032, opened 2021-06-16
Scala structural type application use case 

nmarton-da:
for minimizing the dependency surface of the client

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
@@ -6,8 +6,12 @@ package com.daml.platform.store.appendonlydao.events
 import java.sql.Connection
 import java.time.Instant
 
+import com.daml.ledger.api.domain.PartyDetails
 import com.daml.ledger.participant.state.v1.{CommittedTransaction, RejectionReasonV0}
-import com.daml.platform.store.backend.{ContractStorageBackend, PartyStorageBackend}
+import com.daml.lf.data.Ref
+
+import scala.language.reflectiveCalls
+import scala.util.Try
 
 /** Performs post-commit validation on transactions for Sandbox Classic.
   * This is intended exclusively as a temporary replacement for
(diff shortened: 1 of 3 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/PostCommitValidation.scala
```diff
@@ -50,8 +54,55 @@ private[appendonlydao] object PostCommitValidation {
       None
   }
 
+  type Dao = {
+    def maximumLedgerTime(ids: Set[ContractId])(connection: Connection): Try[Option[Instant]]
+    def parties(parties: Seq[Ref.Party])(connection: Connection): List[PartyDetails]
+    def contractKeyGlobally(key: Key)(connection: Connection): Option[ContractId]
+  }
```
S11001001, line 61: Presumably, the goal here is to avoid declaring this as `trait Dao` and then writing `with Dao` where you construct the single object that satisfies all 3 signatures?

nmarton-da: Maybe it was not entirely clear from the context but what we are having here is classic dependency injection.
- We have a very broad provider interface StorageBackend
- We have various implementations for this (factory: StorageBackend.of)
- We have a client depending on (part of) StorageBackend abstraction: PostCommitValidation   
Creating these service traits on the client and mixing it back either to the facade abstraction or implementations feels very wrong to me, it seems like the opposite of decoupling...or do I get this wrong somewhere? :thinking:

nmarton-da: would come handy if we could mix in traits to a value, then we could do it right at the caller, which would be fine...but that we cannot, can we?

nmarton-da: > `trait Dao` doesn't necessarily need to exist exactly here. And, in fact, when you define a method that matches multiple abstract methods in the declared nominal parents, that definition will be used for all of them, so `trait Dao` does not have to be related to `PartyStorageBackend` or `ContractStorageBackend` in any way.

I am not sure I get this...can you elaborate/show some examples?

nmarton-da: Regarding nominal vs representational: I hear you, and structural types are obviously representational. As function types are. And single abstract methods and anonymous functions dance to representational tune as well. I like them, because it makes programs leaner. Not to mention nominal types do not prevent you to inject the wrong dependency per-se ;)

But we do not necessarily have to agree on whether representational is good or bad, probably you would agree this is anyhow an opinionated discussion. 
As I said I would use this approach, but happy to refrain from usage if some sort of a consensus is forming against :ok_hand: (even better so if there is something which is better for this purpose!)

stefanobaghino-da: > My main concern is that the benefit here seems to be relatively small while we increase the complexity of the codebase and make it harder for people to understand it.

This is my main concern.

I'm not entirely sure about this point either:

> Creating these service traits on the client and mixing it back either to the facade abstraction or implementations feels very wrong to me, it seems like the opposite of decoupling

Let's talk about this, maybe next week, I would like to understand this better.

nmarton-da: it's settled then, thank you for your time

PR #10033, opened 2021-06-16
Speedy: Refactor contract Id/Key callback

remyhaemmerle-da:
Simplifies the logic for failling Speedy calbacks and their interaction with the scenario service.
 
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -20,6 +20,7 @@ import com.daml.lf.speedy.SValue.{SValue => SV}
 import com.daml.lf.transaction.{Transaction => Tx}
 import com.daml.lf.value.{Value => V}
 import com.daml.lf.transaction.{GlobalKey, GlobalKeyWithMaintainers, Node}
+import com.daml.scalautil.Statement.discard
 
 import scala.jdk.CollectionConverters._
 import scala.collection.immutable.TreeSet
@@ -1023,11 +1024,7 @@ private[lf] object SBuiltin {
               coid,
               templateId,
               onLedger.committers,
-              cbMissing = _ => false,
-              cbPresent = { case V.ContractInst(actualTmplId, V.VersionedValue(_, arg), _) =>
-                // Note that we cannot throw in this continuation -- instead
-                // set the control appropriately which will crash the machine
-                // correctly later.
+              { case V.ContractInst(actualTmplId, V.VersionedValue(_, arg), _) =>
                 if (actualTmplId != templateId) {
                   machine.ctrl =
                     SEDamlException(DamlEWronglyTypedContract(coid, templateId, actualTmplId))
@@ -1139,13 +1136,14 @@ private[lf] object SBuiltin {
     private[this] val typ = AstUtil.TContractId(Ast.TTyCon(templateId))
 
     final protected def importCid(cid: V.ContractId): SEImportValue =
+      // We have to check that the discriminator of cid does not conflict with a local ones
+      // however we cannot raise an exception in case of failure here.
+      // We delegate to CtrlImportValue the task to check cid.
       SEImportValue(typ, V.ValueContractId(cid))
     // Callback from the engine returned NotFound
-    def handleInputKeyNotFound(machine: Machine): Boolean
-    // CallBack from the engine returned a new Cid
     def handleInputKeyFound(machine: Machine, cid: V.ContractId): Unit
     // We already saw this key, but it was undefined or was archived
-    def handleInactiveKey(machine: Machine, gkey: GlobalKey): Unit
+    def handleKeyNotFound(machine: Machine, gkey: GlobalKey): Boolean
     // We already saw this key and it is still active
     def handleActiveKey(machine: Machine, cid: V.ContractId): Unit
 
@@ -1156,30 +1154,30 @@ private[lf] object SBuiltin {
     ): Unit =
       keyMapping match {
         case PartialTransaction.KeyActive(cid) => handleActiveKey(machine, cid)
-        case PartialTransaction.KeyInactive => handleInactiveKey(machine, gkey)
+        case PartialTransaction.KeyInactive => discard(handleKeyNotFound(machine, gkey))
       }
   }
 
   private[this] object KeyOperation {
     final class Fetch(override val templateId: TypeConName) extends KeyOperation {
-      override def handleInputKeyNotFound(machine: Machine): Boolean = false
       override def handleInputKeyFound(machine: Machine, cid: V.ContractId): Unit =
         machine.ctrl = importCid(cid)
-      override def handleInactiveKey(machine: Machine, gkey: GlobalKey): Unit =
+      override def handleKeyNotFound(machine: Machine, gkey: GlobalKey): Boolean = {
         machine.ctrl = SEDamlException(DamlEContractKeyNotFound(gkey))
+        false
+      }
+
       override def handleActiveKey(machine: Machine, cid: V.ContractId): Unit =
         machine.returnValue = SContractId(cid)
     }
 
     final class Lookup(override val templateId: TypeConName) extends KeyOperation {
-      override def handleInputKeyNotFound(machine: Machine): Boolean = {
+      override def handleInputKeyFound(machine: Machine, cid: V.ContractId): Unit =
+        machine.ctrl = SBSome(importCid(cid))
+      override def handleKeyNotFound(machine: Machine, key: GlobalKey): Boolean = {
         machine.returnValue = SValue.SValue.None
         true
       }
-      override def handleInputKeyFound(machine: Machine, cid: V.ContractId): Unit =
-        machine.ctrl = SBSome(importCid(cid))
-      override def handleInactiveKey(machine: Machine, key: GlobalKey): Unit =
-        machine.returnValue = SValue.SValue.None
       override def handleActiveKey(machine: Machine, cid: V.ContractId): Unit =
         machine.returnValue = SOptional(Some(SContractId(cid)))
     }
@@ -1200,7 +1198,11 @@ private[lf] object SBuiltin {
       )
     }
 
-    final def execute(args: util.ArrayList[SValue], machine: Machine, onLedger: OnLedger): Unit = {
+    final override def execute(
+        args: util.ArrayList[SValue],
+        machine: Machine,
+        onLedger: OnLedger,
+    ): Unit = {
       import PartialTransaction.{KeyActive, KeyInactive}
       val keyWithMaintainers = extractKeyWithMaintainers(args.get(0))
       if (keyWithMaintainers.maintainers.isEmpty)
@@ -1244,30 +1246,18 @@ private[lf] object SBuiltin {
                   onLedger.committers,
                   { result =>
                     cacheGlobalLookup(onLedger, gkey, result)
-                    // We need to check if the contract was consumed since we only
-                    // modify keys if the archive was for a key
-                    // already brought into scope.
-                    val activeResult = result match {
-                      case Some(cid) if onLedger.ptx.consumedBy.contains(cid) =>
-                        None
-                      case _ =>
-                        result
-                    }
-                    activeResult match {
-                      case Some(cid) =>
+                    result match {
+                      case Some(cid) if !onLedger.ptx.consumedBy.contains(cid) =>
                         onLedger.ptx = onLedger.ptx.copy(
                           keys = onLedger.ptx.keys.updated(gkey, KeyActive(cid))
                         )
-                        // We have to check that the discriminator of cid does not conflict with a local ones
-                        // however we cannot raise an exception in case of failure here.
-                        // We delegate to CtrlImportValue the task to check cid.
                         operation.handleInputKeyFound(machine, cid)
                         true
-                      case None =>
+                      case _ =>
                         onLedger.ptx = onLedger.ptx.copy(
                           keys = onLedger.ptx.keys.updated(gkey, KeyInactive)
                         )
-                        operation.handleInputKeyNotFound(machine)
+                        operation.handleKeyNotFound(machine, gkey)
                     }
                   },
                 )
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
@@ -33,11 +33,10 @@ object SResult {
       contractId: ContractId,
       templateId: TypeConName,
       committers: Set[Party],
-      // Callback to signal that the contract was not present
-      // or visible. Returns true if this was recoverable.
-      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
-      cbMissing: Unit => Boolean,
-      cbPresent: ContractInst[Value.VersionedValue[ContractId]] => Unit,
+      // Callback
+      // returns the next expression to evaluate.
+      // In case of failure the call back does not throw but returns a SErrorDamlException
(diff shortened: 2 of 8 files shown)


cocreature (commented):
I support the general direction here but I think it might be better to first merge https://github.com/digital-asset/daml/pull/10039. That should make it a lot easier to decouple the two.

remyhaemmerle-da (commented):
> I support the general direction here but I think it might be better to first merge #10039. That should make it a lot easier to decouple the two.

Let's do that.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1023,14 +1023,9 @@ private[lf] object SBuiltin {
               coid,
               templateId,
               onLedger.committers,
-              cbMissing = _ => machine.tryHandleSubmitMustFail(),
```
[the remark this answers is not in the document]
cocreature: we shouldn’t.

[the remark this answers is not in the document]
cocreature: if you look at my PR it kills the whole `tryHandleSubmitMustFail` stuff.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SError.scala
```diff
@@ -15,7 +15,10 @@ import com.daml.lf.scenario.ScenarioLedger
 object SError {
 
   /** Errors that can arise during interpretation */
-  sealed abstract class SError extends RuntimeException with Product with Serializable
+  sealed abstract class SError extends RuntimeException with Product with Serializable {
+    // override def toString: String =
+    //   s"$productPrefix(${productIterator.map(_.toString).mkString(",")})"
+  }
```
remyhaemmerle-da, line 21:
```suggestion
  sealed abstract class SError extends RuntimeException with Product with Serializable
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -33,11 +33,9 @@ object SResult {
       contractId: ContractId,
       templateId: TypeConName,
       committers: Set[Party],
-      // Callback to signal that the contract was not present
-      // or visible. Returns true if this was recoverable.
-      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
-      cbMissing: Unit => Boolean,
-      cbPresent: ContractInst[Value.VersionedValue[ContractId]] => Unit,
+      // returns the next expression to evaluate.
```
remyhaemmerle-da, line 36:
```suggestion
      // Callback
      // returns the next expression to evaluate.
```

daml-lf/scenario-interpreter/BUILD.bazel
```diff
@@ -19,6 +19,9 @@ da_scala_library(
     srcs = glob(["src/main/**/*.scala"]),
     main_class = "com.daml.lf.speedy.Main",
     tags = ["maven_coordinates=com.daml:daml-lf-scenario-interpreter:__VERSION__"],
+    versioned_scala_deps = {
+        "2.12": ["@maven//:org_scala_lang_modules_scala_collection_compat"],
+    },
```
remyhaemmerle-da, line 24:
```suggestion
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -33,11 +33,10 @@ object SResult {
...
       committers: Set[Party],
-      // Callback to signal that the contract was not present
-      // or visible. Returns true if this was recoverable.
-      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
-      cbMissing: Unit => Boolean,
-      cbPresent: ContractInst[Value.VersionedValue[ContractId]] => Unit,
+      // Callback
+      // returns the next expression to evaluate.
+      // In case of failure the call back does not throw but returns a SErrorDamlException
+      cbPresent: ContractInst[Value.VersionedValue[ContractId]] => SExpr,
```
remyhaemmerle-da, line 39:
```suggestion
      callback: ContractInst[Value.VersionedValue[ContractId]] => SExpr,
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -33,11 +33,10 @@ object SResult {
...
       committers: Set[Party],
-      // Callback to signal that the contract was not present
-      // or visible. Returns true if this was recoverable.
-      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
-      cbMissing: Unit => Boolean,
-      cbPresent: ContractInst[Value.VersionedValue[ContractId]] => Unit,
+      // Callback
+      // returns the next expression to evaluate.
+      // In case of failure the call back does not throw but returns a SErrorDamlException
+      callback: ContractInst[Value.VersionedValue[ContractId]] => SExpr,
```
remyhaemmerle-da, line 39: I see two alternative: 
```suggestion
      // Callback
      // returns the next expression to evaluate as Right
      // In case of failure the call back does not throw but returns a Left[SErrorDamlException]
      callback: ContractInst[Value.VersionedValue[ContractId]] => Either[SErrorDamlException, SExpr],
```
Not sure it is very convince since `SErrorDamlException` is a subtype of `SExpr`.

```suggestion
      // Callback
      // In case of failure the callback set the `ctrl` of the machine that have returned this `SResult`  to a SErrorDamlException.
      callback: ContractInst[Value.VersionedValue[ContractId]] => Unit,
```
Not very convince..

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1141,54 +1136,50 @@ private[lf] object SBuiltin {
...
-      override def handleInactiveKey(machine: Machine, gkey: GlobalKey): Unit =
-        machine.ctrl = SEDamlException(DamlEContractKeyNotFound(gkey))
-      override def handleActiveKey(machine: Machine, cid: V.ContractId): Unit =
-        machine.returnValue = SContractId(cid)
+      override def handleInputKeyFound(cid: V.ContractId): SExpr =
+        importCid(cid)
+      override def handleKeyNotFound(gkey: GlobalKey): SEDamlException =
+        SEDamlException(DamlEContractKeyNotFound(gkey))
+      override def handleActiveKey(cid: V.ContractId): SEValue =
+        SEValue(SContractId(cid))
```
remyhaemmerle-da, line 1162: We will paid a slight cost for using `SExpr` (and `machine.ctrl`)  instead of `SVaue` (and `machine.returnValue`). 
I do not think it is a big issue, as  underhanded the callback should be already very  costly.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -33,11 +33,10 @@ object SResult {
       contractId: ContractId,
       templateId: TypeConName,
       committers: Set[Party],
-      // Callback to signal that the contract was not present
-      // or visible. Returns true if this was recoverable.
-      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
-      cbMissing: Unit => Boolean,
```
remyhaemmerle-da, line 39: :fire:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
cocreature: I don’t really get what you gain by moving from `Unit` to `SExpr`. The fact that the caller now has to mess with the speedy machine ctrl seems to somewhat mix up responsibilities. I think we’re better off in a state where the caller sets the `ctrl` once at the beginning and then everything else is handled by speedy itself (potentially by providing callback opaque to the caller).

So my suggestion would be to stick with `Unit` for the success case. I don’t care all that much if we use `Either` or just set the ctrl  for the error case so if you have a strong preference go for that.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -237,19 +237,19 @@ object ScenarioRunner {
 
   // The interface we need from a ledger during submission. We allow abstracting over this so we can play
   // tricks like caching all responses in some benchmarks.
-  trait LedgerApi[R] {
+  abstract class LedgerApi[R] {
```
cocreature, line 240: I think I asked you this before but I forgot: Why do you prefer `abstract class` over `trait`?

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -258,22 +258,22 @@ object ScenarioRunner {
     ): Either[SError, R]
   }
 
-  case class ScenarioLedgerApi(ledger: ScenarioLedger)
-      extends LedgerApi[ScenarioLedger.CommitResult] {
+  class ScenarioLedgerApi(ledger: ScenarioLedger) extends LedgerApi[ScenarioLedger.CommitResult] {
```
cocreature, line 261: why? for something immutable like this one a case class seems reasonable to me.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -1388,7 +1388,9 @@ private[lf] object Speedy {
     */
   private[speedy] final case class SpeedyHungry(result: SResult)
       extends RuntimeException
-      with NoStackTrace
+      with NoStackTrace {
+    override def toString = s"SpeedyHungry($result)"
```
cocreature, line 1392: how is this different from what the autogenerated `toString` gives you?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
remyhaemmerle-da: :+1:

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
remyhaemmerle-da: @S11001001 told me that it is in general more efficient. So I started to use it blindly. 

After using this pattern blindly for some time, I like that because it is a simple pattern to avoid multi-intermittency : i.e. 
I tried to use only `abstract class` as super type,  and I  tried to use `trait` only for mixin (like for instance .

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
remyhaemmerle-da: I do not like it contains state full data. 
I will revert it anyway.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
S11001001: > it is a simple pattern to avoid multi-intermittency

The best reason.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
cocreature: `ScenarioLedger` is immutable. But perfectly fine to keep it, just wanted to understand.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
cocreature: seems reasonable, thanks for the explanation

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
remyhaemmerle-da: Here there is no autogenerated `toString`. The one from `Throwable` has precedence

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
nickchapman-da: > I think we’re better off in a state where the caller sets the `ctrl` once at the beginning and then everything else is handled by speedy itself

I agree strongly with this.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
remyhaemmerle-da: my mistake.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -296,26 +297,36 @@ object ScenarioRunner {
...
       val effectiveAt = ledger.currentTime
       val readers = actAs union readAs
 
-      def missingWith(err: SError) =
-        if (!canContinue(None))
-          throw SRunnerException(err)
+      def missingWith(err: SError) = {
+        callback(None)
+        machine.ctrl match {
+          case _: SExpr.SEDamlException =>
```
cocreature, line 324: I don’t understand this. Why can we not just continue interpretation and let it blow up?

cocreature (approved):
Nice! One small comment that I do not understand

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
remyhaemmerle-da: We want (do we really ?) rewrite the error in case of 
- `ScenarioLedger.LookupContractNotFound`
- `ScenarioLedger.LookupContractNotEffective`
- `ScenarioLedger.LookupContractNotActive`
To give a more precise error message. 

Should I kill `ScenarioLedger.LookupContractNotEffective` and `ScenarioLedger.LookupContractNotActive` ?

cocreature: oh I got confused, I thought we rethrow the `SEDamlException` but we rethrow what we pass to `missingWith`. I’m not entirely sure this is an improvement over what we had before. Now we depend on the implementation details of the callback here before instead of making it explicit by returning a boolean. But I don’t feel super strongly about this so if you prefer it fine to keep it.

cocreature: I do think we should keep it.

PR #10036, opened 2021-06-16
make json-api's scopt better-typed by passing JDBC drivers implicitly

S11001001:
This solves two warts in the code:

- the `validate`/`createUnsafe` double-parse, because scopt doesn't let you flatMap;
- the non-`JdbcConfig` sub-configs appeared to need to know the JDBC drivers when they really don't, because of a quirk in the inherited implementation (a "foolish consistency", literally)

That coherence of `scopt.Read` instances calls for all its dependencies to be coherent leads us to treat `supportedJdbcDriverNames` as a nullary typeclass instance.  This is a nullary typeclass by the same justification as `SupportedJdbcDriver`; see scaladoc there for more.

And we solve the latter problem by...adding a type parameter, how else.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
+++ b/ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
@@ -11,8 +11,9 @@ import akka.stream.ThrottleMode
 import com.daml.util.ExceptionOps._
 import com.daml.ledger.api.tls.TlsConfiguration
 import scalaz.std.option._
+import scalaz.syntax.tag._
 import scalaz.syntax.traverse._
-import scalaz.{Show, \/}
+import scalaz.{@@, Show, Tag, \/}
 
 import scala.concurrent.duration._
 import scala.util.Try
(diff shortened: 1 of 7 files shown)


realvictorprm (approved):
Looks good to me, although it took some time for me to get through how it is working now.

cocreature (approved):

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -207,9 +208,8 @@ private[http] object WebsocketConfig extends ConfigCompanion[WebsocketConfig]("W
   )
 
   override def create(
-      x: Map[String, String],
-      supportedJdbcDriverNames: Set[String],
-  ): Either[String, WebsocketConfig] =
+      x: Map[String, String]
+  )(implicit readCtx: DummyImplicit): Either[String, WebsocketConfig] =
```
stefanobaghino-da, line 212: Why is this necessary? Can we get away with renaming `create` to something else instead of introducing a dummy implicit?

stefanobaghino-da (approved):

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
S11001001: There has to be something there, to match the abstract signature.

My first instinct was initially to leave the argument where it was, but passing `()` explicitly was a little inelegant, but more importantly, it wouldn't mesh as well with supplying the `Read` instances.

PR #10039, opened 2021-06-16
Split scenario & ledger execution

cocreature:
This PR by no means aims to solve everything we can do here. It is
rather the minimal change that I could get to work that provides us
with one Speedy machine for scenario execution (which is still an
on-ledger machine to avoid having to change the callsites to much in
this PR) and one speedy machine per submission.

There is tons of cleanup we can do afterwards but this should
hopefully set the right foundations.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1023,7 +1023,7 @@ private[lf] object SBuiltin {
               coid,
               templateId,
               onLedger.committers,
-              cbMissing = _ => machine.tryHandleSubmitMustFail(),
+              cbMissing = _ => false,
```
cocreature, line 1026: we should be able to simplify this but I’ll leave that for a separate PR.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -35,6 +35,7 @@ object SResult {
       committers: Set[Party],
       // Callback to signal that the contract was not present
       // or visible. Returns true if this was recoverable.
+      // TODO (MK) Drop now that tryHandleSubmitMustFail is dead.
```
cocreature, line 38: same point as above

daml-lf/scenario-interpreter/BUILD.bazel
```diff
@@ -54,6 +54,8 @@ daml_compile(
     visibility = ["//visibility:public"],
 )
 
+# TODO (MK) Figure out what to do about the benchmark.
```
cocreature, line 57: This is the main thing that I still need to figure out. I think this should really just go via the scenario runner instead of reimplementing things but that needs some mechanism for caching to make it work.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -66,54 +73,165 @@ final case class ScenarioRunner(
...
-
         case SResultScenarioGetParty(partyText, callback) =>
           getParty(partyText, callback)
 
-        case SResultNeedKey(keyWithMaintainers, committers, cb) =>
-          // We never have readAs parties in scenarios.
-          lookupKeyUnsafe(keyWithMaintainers.globalKey, committers, Set.empty, cb)
+        case SResultScenarioSubmit(committers, commands, location, mustFail, callback) =>
+          val submitResult = submit(committers, commands, location)
+          // TODO (MK) We copy the ptx & commit location and the trace
```
cocreature, line 100: This is definitely not the desired endstate. I’d like to keep it for a separate PR though. Fixing this really only makes sense if we make the same change in Daml script (which currently does something similarly messy because it had to fit in what the scenario service did :facepalm:) and this PR is large enough.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -66,54 +73,165 @@ final case class ScenarioRunner(
...
+        case _: SResultNeedLocalKeyVisible =>
+          crash("SResultNeedLocalKeyVisible outside of submission")
       }
     }
     val endTime = System.nanoTime()
     val diff = (endTime - startTime) / 1000.0 / 1000.0
     (diff, steps, ledger, finalValue)
   }
 
+  private def submit(
```
cocreature, line 163: I think this part should be sharable with Daml script but again separate PR.

daml-lf/tests/scenario/stable/mustfails/EXPECTED.ledger
```diff
@@ -85,4 +85,4 @@ TX #21 1969-12-31T23:58:20Z [Test:345] version: 12
 mustFailAt actAs: {'Alice'} readAs: {} [Test:346]
 
 active contracts:
-   00067f889bdf49aa670596cab1a5907d75596d971df3477087c88ca86f281cee85, 005c2e250ff7988dbda72348b8172716801c6ad40cacecba387863315a8ed678db, 009097ab3d05b0538d961df2cb88558e4cd1947514d2f2c275363a3dc81e07b6aa, 00945c0656f04cdd9f6a85cb466acb0edcd75ea0f618362e83b92164c4cee523e9, 00a5627208a832b7d951562d793cb1c9e48b3459603d2543b48656ea3ac2debf41, 00d34fd9f2c2e66e6d68c344f1bec486f1c5624a4268bae020e2584a1ad411d65d, 00e52b332d66bbf5ac835144e35bf339a7ab319e505e005d1ddd9aed313764dd51
+   00049c3d61768cf931aa2b97a7ac28002d7673c3f8b996798d87227106e178286e, 00067f889bdf49aa670596cab1a5907d75596d971df3477087c88ca86f281cee85, 00154fe9827fdb40afae4ba484f3f51159d99a27c1463928ec761043c28353b76c, 007e5c9342585588fff0855b297c6e06bfb6a3d62578a28da860af02a98048bd68, 00945c0656f04cdd9f6a85cb466acb0edcd75ea0f618362e83b92164c4cee523e9, 00bf46b75494167be56ef270a63cbc740370b75c98fb517457d9859d55796f7fbf, 00e52b332d66bbf5ac835144e35bf339a7ab319e505e005d1ddd9aed313764dd51
```
cocreature, line 88: these were just broken. We don’t unwind the ptx properly in a few places.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -53,6 +60,15 @@ final case class ScenarioRunner(
     }
   }
 
+  private[this] def nextSeed(submissionSeed: crypto.Hash): crypto.Hash =
+    crypto.Hash.deriveTransactionSeed(
```
cocreature, line 64: I don’t particularly like any of the seed handling here and I’m almost certain the current behavior is not what anyone intended but for now this at least keeps the behavior.

remyhaemmerle-da (approved):
LGTM

nickchapman-da (commented):
Very nice! 
One question: What was the final status of the benchmark? Did the TODO get done?
(It looks like it did)

cocreature:
> Very nice!
> One question: What was the final status of the benchmark? Did the TODO get done?
> (It looks like it did)

Yes, it was a bit annoying to rewire it to make the caching work again but it seems to work now. Haven’t checked the numbers yet but given that scenarios are not really what we’re optimizing for at worst we’re paying a bit for the switch between the two machines which doesn’t matter all that much.

Issue #10020, opened 2021-06-16
Extend JSON API metrics


cocreature:
copying my [comment](https://github.com/digital-asset/daml/pull/9888#pullrequestreview-674921466) so it doesn’t get lost

- [x]    Concurrent requests: This should probably be split into 3 or maybe 4 metrics: concurrent command submissions, concurrent queries, concurrent websocket connections and maybe one for party & package allocation.
- [x]     Total duration of requests. This is what you already added. I think we should at least split that in command submissions and queries and lastly party/package management. Not quite sure what we actually measure for websocket connections in this PR. > The total duration there doesn’t seem all that interesting given that they run forever until the client closes the connection.
- More fine grained durations. A few examples:
  - [x] Validation & conversion of the incoming JSON values.
  - [ ] For command submissions the duration of the request to the ledger.
  - [x] Conversion on the outgoing side.
  - [x] Database operations at least split into reads & writes but probably more fine-grained is helpful.

This is not intended to be an exhaustive list and I expect we’ll expand it over time but I think with this we should have a good starting point.

I recommend to not tackle everything in one PR and instead tackle them separately (Some of them might be checked already, haven’t confirmed).
```
https://github.com/digital-asset/daml/pull/9888#pullrequestreview-674921466
<!--

This issue tracker is only for technical issues related to the Daml SDK.

If you suspect that your issue has a security impact, do not use this issue tracker. In that case, please contact security@digitalasset.com and follow our responsible disclosure policy: https://digitalasset.com/security/ .

For general questions about Daml and/or support requests you should use our forum: https://discuss.daml.com .

-->

realvictorprm:
I split this up into seperate issues to also refine specific things if necessary and also to track the seperate effort. Always good to have an issue to close with a PR :smile:

realvictorprm:
> Conversion on the outgoing side.
With this I assume the time spend on constructing the response payload?

PR #10045, opened 2021-06-17
[JSON-API] Add moar timing metrics

realvictorprm:
Solves part of #10020

I want to add the db metrics with a seperate PR as this seems to be a bit more complex (also because I have not worked in that part of the code yet).

changelog_begin

- [JSON-API] Timing metrics are now available for party management, package management, command submission and query endpoints.
- [JSON-API] Also added a timing metric for parsing and decoding of incoming json payloads

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -89,9 +89,11 @@ class Endpoints(
         case req @ HttpRequest(POST, Uri.Path("/v1/create-and-exercise"), _, _, _) =>
           (implicit lc => httpResponse(createAndExercise(req)))
       }
-    val queryDispatch = mkDispatchFunWithTimer(apiMetrics.queryTimer) {
+    val queryAllDispatch = mkDispatchFunWithTimer(apiMetrics.queryAllTimer) {
       case req @ HttpRequest(GET, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(retrieveAll(req)))
+    }
+    val queryMatchingDispatch = mkDispatchFunWithTimer(apiMetrics.queryMatchingTimer) {
       case req @ HttpRequest(POST, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(query(req)))
     }
@@ -138,7 +140,8 @@ class Endpoints(
     }
     import scalaz.std.partialFunction._, scalaz.syntax.arrow._
     ((commandDispatch orElse
-      queryDispatch orElse
+      queryAllDispatch orElse
+      queryMatchingDispatch orElse
       fetchDispatch orElse
       getPartyDispatch orElse
       allocatePartyDispatch orElse
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -67,54 +67,69 @@ class Endpoints(
   import util.ErrorOps._
   import Uri.Path._
 
+  // Parenthesis in the case matches below are required because otherwise scalafmt breaks.
+  //noinspection ScalaUnnecessaryParentheses
   def all(implicit
       lc: LoggingContextOf[InstanceUUID],
       metrics: Metrics,
   ): PartialFunction[HttpRequest, Http.IncomingConnection => Future[HttpResponse]] = {
+    val apiMetrics = metrics.daml.HttpJsonApi
     type DispatchFun =
-      PartialFunction[HttpRequest, LoggingContextOf[InstanceUUID with RequestID] => Future[HttpResponse]]
-    // Parenthesis are required because otherwise scalafmt breaks.
-    val commandDispatch: DispatchFun = ({
-      case req @ HttpRequest(POST, Uri.Path("/v1/create"), _, _, _) =>
-        (implicit lc => httpResponse(create(req)))
-      case req @ HttpRequest(POST, Uri.Path("/v1/exercise"), _, _, _) =>
-        (implicit lc => httpResponse(exercise(req)))
-      case req @ HttpRequest(POST, Uri.Path("/v1/create-and-exercise"), _, _, _) =>
-        (implicit lc => httpResponse(createAndExercise(req)))
-    }: DispatchFun) andThen (f =>
-      lc => Timed.future(metrics.daml.HttpJsonApi.commandSubmissionTimer, f(lc))
-    )
-    val queryDispatch: DispatchFun = ({
-      case req @ HttpRequest(POST, Uri.Path("/v1/fetch"), _, _, _) =>
-        (implicit lc => httpResponse(fetch(req)))
+      PartialFunction[HttpRequest, LoggingContextOf[InstanceUUID with RequestID] => Future[
+        HttpResponse
+      ]]
+    def mkDispatchFunWithTimer(timer: Timer)(fun: DispatchFun): DispatchFun =
+      fun andThen (f => lc => Timed.future(timer, f(lc)))
+    val commandDispatch =
+      mkDispatchFunWithTimer(apiMetrics.commandSubmissionTimer) {
+        case req @ HttpRequest(POST, Uri.Path("/v1/create"), _, _, _) =>
+          (implicit lc => httpResponse(create(req)))
+        case req @ HttpRequest(POST, Uri.Path("/v1/exercise"), _, _, _) =>
+          (implicit lc => httpResponse(exercise(req)))
+        case req @ HttpRequest(POST, Uri.Path("/v1/create-and-exercise"), _, _, _) =>
+          (implicit lc => httpResponse(createAndExercise(req)))
+      }
+    val queryDispatch = mkDispatchFunWithTimer(apiMetrics.queryTimer) {
       case req @ HttpRequest(GET, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(retrieveAll(req)))
       case req @ HttpRequest(POST, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(query(req)))
-    }: DispatchFun) andThen (f => lc => Timed.future(metrics.daml.HttpJsonApi.queryTimer, f(lc)))
-    val partyManagementDispatch: DispatchFun = ({
+    }
+    val fetchDispatch: DispatchFun = {
+      case req @ HttpRequest(POST, Uri.Path("/v1/fetch"), _, _, _) =>
+        (implicit lc => Timed.future(apiMetrics.fetchTimer, httpResponse(fetch(req))))
+    }
+    val getPartyDispatch = mkDispatchFunWithTimer(apiMetrics.getPartyTimer) {
       case req @ HttpRequest(GET, Uri.Path("/v1/parties"), _, _, _) =>
         (implicit lc => httpResponse(allParties(req)))
       case req @ HttpRequest(POST, Uri.Path("/v1/parties"), _, _, _) =>
         (implicit lc => httpResponse(parties(req)))
+    }
+    val allocatePartyDispatch: DispatchFun = {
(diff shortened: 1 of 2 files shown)


ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
```diff
@@ -708,6 +708,17 @@ final class Metrics(val registry: MetricRegistry) {
...
+      val commandSubmissionTimer: Timer = registry.timer(Prefix :+ "command_submission_timing")
+      // Meters how long processing of a query request takes
+      val queryTimer: Timer = registry.timer(Prefix :+ "query_timing")
+      // Meters how long processing of a party management request takes
+      val partyManagementTimer: Timer = registry.timer(Prefix :+ "party_management_timing")
+      // Meters how long processing of a package management request takes
+      val packageManagementTimer: Timer = registry.timer(Prefix :+ "package_management_timing")
+      // Meters how long parsing and decoding of incoming json payload takes
+      val incomingJsonParsingAndValidationTimer: Timer =
+        registry.timer(Prefix :+ "incoming_json_parsing_and_validation_timing")
```
stefanobaghino-da, line 721: It's probably fine to have all command endpoints tracked by a single timer, but shouldn't we rather have more fine-grained metrics to allow users to differentiate between the time taken to, say, make a query or fetch request?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -70,44 +71,62 @@ class Endpoints(
...
       case HttpRequest(GET, Uri.Path("/readyz"), _, _, _) =>
         _ => healthService.ready().map(_.toHttpResponse)
     }
     import scalaz.std.partialFunction._, scalaz.syntax.arrow._
-    (dispatch &&& { case r => r }) andThen { case (lcFhr, req) =>
+    ((commandDispatch orElse
+      queryDispatch orElse
+      partyManagementDispatch orElse
+      packageManagementDispatch orElse
+      liveOrHealthDispatch) &&& { case r => r }) andThen { case (lcFhr, req) =>
```
stefanobaghino-da, line 129: If `&&&` and `orElse` do the same thing we should probably choose one and stick to that.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -70,44 +71,62 @@ class Endpoints(
...
+      lc => Timed.future(metrics.daml.HttpJsonApi.commandSubmissionTimer, f(lc))
+    )
+    val queryDispatch: DispatchFun = ({
       case req @ HttpRequest(POST, Uri.Path("/v1/fetch"), _, _, _) =>
         (implicit lc => httpResponse(fetch(req)))
       case req @ HttpRequest(GET, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(retrieveAll(req)))
       case req @ HttpRequest(POST, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(query(req)))
+    }: DispatchFun) andThen (f => lc => Timed.future(metrics.daml.HttpJsonApi.queryTimer, f(lc)))
```
stefanobaghino-da, line 94: I don't remember off the top of my mind: what happens is a request is upgrade to `ws`? Is it handled here and would we start timing an arbitrary long WebSocket call, possibly polluting the stats?

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
realvictorprm: I can do that, I'm just unsure about what is most useful. Maybe I should just go the route "More metrics = better" :smile:

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: Oh didn't know that, many thanks. I personally would prefer `orElse` as this reads better than `&&&`. The latter is in F# btw bitwise-and :sweat_smile:

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: No need to worry here, web socket calls are handled seperate with different endpoints (see the `HttpServer.scala`).

realvictorprm:
Darn, CI failes due to formatting. How could I forgot to run a format before commiting :angry:

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
stefanobaghino-da: I do prefer `orElse` as well. @S11001001 what is the advantage of `&&&` over `orElse`? I would guess something related to type inference or variance but I really don't know.

stefanobaghino-da:
> Darn, CI failes due to formatting. How could I forgot to run a format before commiting

I have

```
#!/bin/sh

./ci/check-changelog.sh && ./fmt.sh --diff --test
```

as my `.git/hooks/pre-push`, this way I can't push changes that wouldn't pass platform-agnostic checks.

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
stefanobaghino-da: Not necessarily, I think that grouping commands together in sensible, but I would at least differentiate for things that clearly achieve different things, like between uploading and downloading a package.

tudor-da:
> > Darn, CI failes due to formatting. How could I forgot to run a format before commiting
> 
> I have
> 
> ```
> #!/bin/sh
> 
> ./ci/check-changelog.sh && ./fmt.sh --diff --test
> ```
> 
> as my `.git/hooks/pre-push`, this way I can't push changes that wouldn't pass platform-agnostic checks.

ha, nice one 👍

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: Just noticed that messed up, I do know the `&&&` operator and it doesn't mean `orElse` in this context. It means `combine`.

realvictorprm:
going to try that out too, thanks @stefanobaghino-da :smile:

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -66,64 +67,102 @@ class Endpoints(
...
+        case req @ HttpRequest(POST, Uri.Path("/v1/create-and-exercise"), _, _, _) =>
+          (implicit lc => httpResponse(createAndExercise(req)))
+      }
+    val queryDispatch = mkDispatchFunWithTimer(apiMetrics.queryTimer) {
       case req @ HttpRequest(GET, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(retrieveAll(req)))
       case req @ HttpRequest(POST, Uri.Path("/v1/query"), _, _, _) =>
         (implicit lc => httpResponse(query(req)))
+    }
+    val fetchDispatch: DispatchFun = {
```
realvictorprm, line 98: I think I'm going to merge the single case matches here into one because it's also weird to have an extra declaration for them then.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -66,64 +67,102 @@ class Endpoints(
...
+    val getPartyDispatch = mkDispatchFunWithTimer(apiMetrics.getPartyTimer) {
       case req @ HttpRequest(GET, Uri.Path("/v1/parties"), _, _, _) =>
         (implicit lc => httpResponse(allParties(req)))
       case req @ HttpRequest(POST, Uri.Path("/v1/parties"), _, _, _) =>
         (implicit lc => httpResponse(parties(req)))
+    }
+    val allocatePartyDispatch: DispatchFun = {
       case req @ HttpRequest(POST, Uri.Path("/v1/parties/allocate"), _, _, _) =>
-        (implicit lc => httpResponse(allocateParty(req)))
+        (
```
realvictorprm, line 110: All of this weird formatting just because scalafmt needs these braces, otherwise it fails with a parser error -.-

ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
realvictorprm: did split them up with 33842a5

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
stefanobaghino-da: What is `combine`?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
stefanobaghino-da: I'll take weird formatting over formatting discussions any time. 🙂

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
S11001001: This is a "the type is the documentation" situation for sure 🙂 https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Arrow.html#v:-38--38--38-

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -66,64 +67,102 @@ class Endpoints(
...
+    val commandDispatch =
+      mkDispatchFunWithTimer(apiMetrics.commandSubmissionTimer) {
+        case req @ HttpRequest(POST, Uri.Path("/v1/create"), _, _, _) =>
+          (implicit lc => httpResponse(create(req)))
+        case req @ HttpRequest(POST, Uri.Path("/v1/exercise"), _, _, _) =>
+          (implicit lc => httpResponse(exercise(req)))
+        case req @ HttpRequest(POST, Uri.Path("/v1/create-and-exercise"), _, _, _) =>
+          (implicit lc => httpResponse(createAndExercise(req)))
+      }
+    val queryDispatch = mkDispatchFunWithTimer(apiMetrics.queryTimer) {
```
cocreature, line 92: We probably should have separate timers for GET and POST here. The GET wildcard request which fetches all templates bypasses the database cache so we cannot really expect that it correlates with the POST requests which are cached.

cocreature (approved):
Thanks!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: will adjust that

realvictorprm: Addressed with f496c2c

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: on the other hand, mixing stuff together when it's split up that cleanly isn't looking that much better either.

realvictorprm:
I think mac is being flaky again
image

So much wow....

realvictorprm:
It's taking way longer than I want, excuse my impatience

cocreature:
`/azp run` reruns the whole pipeline. Usually you are better off waiting for the build to finish and rerunning only the jobs that failed via the GH ui.

realvictorprm:
I built on build caching for the successful run pipelines, however maybe it would have finished if I waited a bit longer. It just was kind of stuck in the notify_user stage.

realvictorprm:
as the machines were broken I unfortunately need to restart that whole stuff.

PR #10049, opened 2021-06-17
Allow constraints in any position in data-dependencies.

### LINKED ISSUE #8411: data-dependencies: Support constraints not in prenex position.
DAML/damlc is fine with constraints that show up in the middle of a type signature, e.g.

```
grantShowInt : (Show Int => t) -> t
```

or 

```
class Action1 m where
    buildAction : forall e t. (Action (m e) => m e t) -> m e t
```

But data-dependencies will not handle these type signatures correctly. It will ~~silently drop~~ misplace the inner constraint when re-exporting these functions as stubs. I think this is just an incorrect translation of the types. Either we should fix the translation, or warn about such functions. (I'd prefer the former if it's easy, since it's clear what this should translate to in DAML-LF.)

sofiafaro-da:
This fixes #8411. It turns out the issue was some missing parentheses. 😓 

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

PR #10051, opened 2021-06-17
10050 append only schema on oracle

dasormeter:
Get Append Only Schema working on Oracle

- [X] Instant -> java.sql.Timestamp conversion working, that looks to be the only problematic conversion need for packages related functionality
- [X] ran into reserved word problem with column type in packages table - rather than doing convoluted string escaping specific to the different dbs just renamed the field and included migration rename script for postgres (and oracle appendonly script)
- [X] broke out append only to separate folder a la postgres run on top of old schema in oracle folder, so the two will still both work during the migration dev
- [X] use `fetch next {pageSize} rows only` instead of `limit {pageSize` to avoid db specialisation
- [X] had to bring implicit conversion into scope for parser of party and party_entries for BigDecimal -> Boolean since Oracle does not have a native sql boolean
- [X] Oracle doesn't support boolean expressions or columns in select query - provide for specialized alternative to select queries checking for particular value of event_kind
- [X] When BLOB bytearray fields are accessed via a view, the column type of the view gets inferred from the first table used in creating the view -- matched order of tables for the view with postgres so that BLOB fields are defined as nullable
- [X] Include code to chunk party literal for json array lookup in max 4k expression to avoid hitting oracle string literal limit
- [X] Ensure conversion from serialised json array String to Iterable[Party] is used for flat events witnesses parsing
- [X] Wire in conformance tests on oracle for ledger-on-sql to ensure broader test coverage than from LedgerDao test suite alone
- [X] Avoid prepared statements that include multiple sql statements in one block (particularly for initialization and cleanup on indexer reset)


ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -42,7 +42,7 @@ CREATE TABLE packages (
     package_id VARCHAR PRIMARY KEY NOT NULL,
     upload_id VARCHAR NOT NULL,
     source_description VARCHAR,
-    size BIGINT NOT NULL,
+    siz BIGINT NOT NULL,
```
nmarton-da, line 45: nitpic: siz looks bit odd, can we go with something like package_size instead?

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V102_100__rename_packages_size_column.sql
```diff
@@ -0,0 +1,5 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+ALTER TABLE packages
```
nmarton-da, line 4: we have currently h2database and h2database-appendonly directories, this should be going to the appnedonly, righ? (because we deal hear with the appendonly changes I assume)

moreover: perheaps you noticed that the h2database-appendonly is starting from scratch, because we do not need data continuity there. therefore It would be much easier just to change the V1__Append_only_schema.sql directly, and regenerate the hash

nmarton-da: if you want to have clarity about how migrations play together with append-only: FlywayMigrations

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,484 @@
...
+       NULL          AS create_argument_compression,
+       participant_events_non_consuming_exercise.create_key_value_compression,
+       participant_events_non_consuming_exercise.exercise_argument_compression,
+       participant_events_non_consuming_exercise.exercise_result_compression
+FROM participant_events_non_consuming_exercise;
+
+
+-- Stores which events were touched by which migration.
+-- This metadata is not used for normal indexing operations and exists only to simplify fixing data migration issues.
+CREATE TABLE participant_migration_history_v100 (
```
nmarton-da, line 329: you do not need that, because no datacontinuity is needed for append only

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,484 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
nmarton-da, line 1: since no data migration is here, it would be much cleaner, more readable if we would start also the oracle append only from scratch, as in h2database. what do you think?

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,484 @@
...
+
+-- lookup by contract_key
+CREATE INDEX participant_events_create_create_key_hash_idx ON participant_events_create(create_key_hash, event_sequential_id);
+
+
+---------------------------------------------------------------------------------------------------
+-- Events table: consuming exercise
+---------------------------------------------------------------------------------------------------
+
+-- offset index: used to translate to sequential_id
```
nmarton-da, line 431: the index creation is separated originally from the tables to accommodate iterative migrations of heavy databases. since this is not needed, I would go ahead and put these next to the table definitions for clarity

ledger/participant-integration-api/src/main/resources/db/migration/postgres/V102_100__rename_packages_size_column.sql
```diff
@@ -0,0 +1,5 @@
+-- Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+ALTER TABLE packages
+    RENAME COLUMN size TO siz;
```
nmarton-da, line 5: same, with naming

ledger/participant-integration-api/src/main/scala/platform/store/backend/DbDto.scala
```diff
@@ -101,7 +101,7 @@ object DbDto {
       package_id: String,
       upload_id: String,
       source_description: Option[String],
-      size: Long,
+      siz: Long,
```
nmarton-da, line 104: same with naming

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -268,8 +268,9 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
       |    ledger_offset <= {endInclusive} and
       |    parameters.ledger_end >= ledger_offset
       |  order by ledger_offset asc
-      |  limit {pageSize}
-      |  offset {queryOffset}""".stripMargin
+      |  offset {queryOffset} rows
```
nmarton-da, line 271: according to postgres specification `next` and `rows` are noise words...are they noise words in oracle as well? (if not, that would be a reason enough to have limit for postgres and this for oracle separately)
if they are noise words, can we omit them?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -541,13 +545,11 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
   )(connection: Connection): Vector[(Offset, PackageLedgerEntry)] = {
     import com.daml.platform.store.Conversions.OffsetToStatement
-    SQL_GET_PACKAGE_ENTRIES
-      .on(
-        "startExclusive" -> startExclusive,
-        "endInclusive" -> endInclusive,
-        "pageSize" -> pageSize,
-        "queryOffset" -> queryOffset,
-      )
+
```
nmarton-da, line 548: @dasormeter would you be okay with not doing pure refactorings together with the meaty part? we can do this later, no functional change here, adds noise. I think a TODO comment to refactor these things is sufficient for now, and we can do refactorings later

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: this goes for all such constructs

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleField.scala
```diff
@@ -0,0 +1,134 @@
...
+      value: java.lang.Integer,
+  ): Unit = {
+    // TODO FIXME either cleanup and use Field from common, or change the line below
+    preparedStatement.setObject(index, value)
+  }
+}
+
+private[oracle] case class OracleBooleanField[FROM](extract: FROM => Boolean)
+    extends TrivialField[FROM, Boolean]
+//  {
```
nmarton-da, line 63: remove?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
```diff
@@ -0,0 +1,64 @@
...
+            Table.ifNonEmpty(data) {
+              val preparedStatement = connection.prepareStatement(insertStatement)
+              data(0).indices.foreach { dataIndex =>
+                fields(keyFieldIndex)._2.prepareData(
+                  preparedStatement,
+                  1,
+                  data(keyFieldIndex)(dataIndex),
+                )
+                fields.indices.foreach { fieldIndex =>
+                  preparedStatement.setObject(fieldIndex + 2, data(fieldIndex)(dataIndex))
```
nmarton-da, line 27: perhaps we need here the same treatment as in the other places:
fields(fieldIndex)._2.prepareData(
                    preparedStatement,
                    fieldIndex + 2,
                    data(fieldIndex)(dataIndex),
                  )

nmarton-da (commented):
promising

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
nmarton-da: or payload_size

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V102_100__rename_packages_size_column.sql
dasormeter: reverted

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
dasormeter: i do want this for oracle to avoid repetition of the initial definitions and have the old and new schema working for comparison -- so this is in the same style as postgres with postgres and postgres-appendonly

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
dasormeter: kept so we can test both old and new, and have just the changes and deltas in this -- once we remove old oracle completely maybe we move everything to this but that can be a follow-on.

now this follows same style as postgres

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
dasormeter: believe the noise words are needed for this to be portable across all the dbs

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
dasormeter: yeah this was a change when initially we thought the `SQL(...)` vs `SQL"..."` interpolation was causing an issue but it turned out that it was not needed -- i'll try to roll it back

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
dasormeter: split out now

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
dasormeter: reordered to put indices right under their relevant table defs

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
dasormeter: resolved

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleField.scala
dasormeter: yeah still need to clean up unused fields here

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
dasormeter: i think just this one place needed it

ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql
```diff
@@ -81,9 +81,8 @@ CREATE TABLE packages
 -- Table for storing a log of ledger configuration changes and rejections.
 CREATE TABLE configuration_entries
 (
-    ledger_offset    VARCHAR2(4000)  not null,
-    recorded_at      timestamp       not null, -- with time zone
-
+    ledger_offset    VARCHAR2(4000)  not null primary key,
+    recorded_at      timestamp       not null, -- TODO BH: in postgres this is with timezone
```
rautenrieth-da, line 85: The Postgres schema uses `timestamp`, which is [without](https://www.postgresql.org/docs/9.1/datatype-datetime.html) a timezone.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -264,12 +235,13 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
       |    configuration_entries,
       |    parameters
       |  where
-      |    ledger_offset > {startExclusive} and
+      |    ({startExclusive} is null or ledger_offset>{startExclusive}) and
```
rautenrieth-da, line 238: Why this change? Can the start offset ever be null?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -870,10 +887,11 @@ private[backend] object TemplatedStorageBackend {
       witnessesWhereClause: String,
       submittersInPartiesClause: String,
       filteredWitnessesClause: String,
+      columnEqualityBoolean: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.TreeEvent]] = {
     import com.daml.platform.store.Conversions.ledgerStringToStatement
     SQL"""select #$selectColumnsForTransactionTree, #$filteredWitnessesClause as event_witnesses,
-                 event_kind = 20 as exercise_consuming,
+                 #$columnEqualityBoolean as exercise_consuming,
```
rautenrieth-da, line 894: Did you check whether the query plan for
```
SELECT *
FROM participant_events
WHERE event_kind = 20
```
is equal to the one for
```
SELECT *
FROM participant_events_create
```
?

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -396,6 +402,8 @@ private[backend] object H2StorageBackend
       partyArrayContext = partyArrayContext,
       witnessesWhereClause =
         arrayIntersectionWhereClause("tree_event_witnesses", Set(requestingParty)),
+      submittersInPartiesClause = arrayIntersectionWhereClause("submitters", Set(requestingParty)),
+      columnEqualityBoolean = columnEqualityBoolean("event_kind", "20"),
```
rautenrieth-da, line 406: How about renaming this parameter to `createEventFilter = columnEqualityBoolean("event_kind", "20"),` the generic "columnEqualityBoolean" name makes it difficult to understand what value needs to be passed in.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
dasormeter: renamed

ledger/participant-integration-api/src/main/scala/platform/store/backend/DbDto.scala
dasormeter: renamed

ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql
dasormeter: every oracle timezone variant actually does store some time offset -- it looks like `TIMESTAMP WITH LOCAL TIME ZONE` is what the documentation recommends if you don't care about the time component, but i have experienced this get screwy in the past when you don't explicitly run your application servers and DB as UTC, as it shifts to local (particularly problematic for a global application across asia / emea / americas where a local shift actually changes the effective day)

we should probably add some tests at timezone shift boundaries (early morning hours asia where emea and americas are a day behind, late day americas where asia is a day ahead)

https://docs.oracle.com/en/database/oracle/oracle-database/19/nlspg/datetime-data-types-and-time-zone-support.html#GUID-2DFBD56C-5832-4AE8-831E-7EF1EFB7F0BB

```The TIMESTAMP WITH LOCAL TIME ZONE datatype stores the timestamp without time zone information. It normalizes the data to the database time zone every time the data is sent to and from a client. It requires 11 bytes of storage.```

```If you specify a date value without a time component, then the default time is midnight. If you specify a date value without a date, then the default date is the first day of the current month.

Oracle Database DATE columns always contain fields for both date and time. If your queries use a date format without a time portion, then you must ensure that the time fields in the DATE column are set to midnight. You can use the TRUNC (date) SQL function to ensure that the time fields are set to midnight, or you can make the query a test of greater than or less than (<, <=, >=, or >) instead of equality or inequality (= or !=). Otherwise, Oracle Database may not return the query results you expect.```

Comment (dasormeter) [reply ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql]: changing to `TIMESTAMP WITH LOCAL TIME ZONE` then confuses conversion with handling the java.sql.timestampz field

so i think in short, let's stick with `TIMESTAMP` and i'll remove the todos

```14:42:48.465 [daml.index.db.threadpool.connection.jdbcledgerdaooracleappendonlyspec-1] ERROR c.d.p.s.appendonlydao.DbDispatcher - Exception while executing SQL query. Rolled back. , context: {metric=get_completions}
anorm.AnormException: TypeDoesNotMatch(Cannot convert oracle.sql.TIMESTAMPLTZ@18fd069f: oracle.sql.TIMESTAMPLTZ to Date for column ColumnName(.RECORD_TIME,Some(RECORD_TIME)))
- should serialize a valid fetch *** FAILED *** (21 milliseconds)
  anorm.AnormException: TypeDoesNotMatch(Cannot convert oracle.sql.TIMESTAMPLTZ@18fd069f: oracle.sql.TIMESTAMPLTZ to Date for column ColumnName(.RECORD_TIME,Some(RECORD_TIME)))```

File: ledger/ledger-on-sql/src/test/lib/scala/com/daml/ledger/on/sql/MainWithEphemeralOracleUser.scala [hunk 16]
```diff
@@ -0,0 +1,32 @@
...
+        .getOrElse(sys.exit(1))
+
+    connectToOracle()
+    val user = createNewRandomUser()
+    sys.addShutdownHook(dropUser(user.name))
+    val oracleJdbcUrl = s"jdbc:oracle:thin:${user.name}/${user.pwd}@localhost:$oraclePort/ORCLPDB1"
+    val config = originalConfig.copy(
+      participants = originalConfig.participants.map(_.copy(serverJdbcUrl = oracleJdbcUrl)),
+      extra = ExtraConfig(jdbcUrl =
+        Some("jdbc:h2:mem:daml-on-sql-conformance-test")
```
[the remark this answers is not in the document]
dasormeter: i think it's just to say a generic h2 backed in memory conformance test ledger, and then the participant is oracle-backed for the test

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
dasormeter: i believe @nmarton-da ran into a problem where he was seeing null being passed (due to a bug) and put this in for defensiveness

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
dasormeter: this query is just returning a boolean for whether the column value is 20 though, since oracle does not have a native boolean type for a column return value -- it's not actually checking the validity of the view -- the view created for oracle is basically identical to what is done for postgres/h2

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
dasormeter: yeah it is a boolean derived in the select clause only (it is a good question why we need this boolean at all as we could just add a where clause to do the same in the queries that care about it, no?  here i just focused on getting functional equivalent to boolean used for postgres/h2 for oracle, since oracle does not have a native boolean type, so you have to do 0/1 numeric conversion sorcery

ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql
nmarton-da: just a note: running DB servers at UTC is pretty much a must for postgres as well

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: The ledger_offset is serialized from an Offset, this can be a null offset (empty array) in case we want to start the range from the very beginning, which will be an empty string, which works with h2 and postgres, but not with oracle. this special treatment takes care of it. @dasormeter I took your treatment for the mutable schema as is.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
```diff
@@ -0,0 +1,520 @@
...
+      witnessesWhereClause =
+        arrayIntersectionWhereClause("tree_event_witnesses", requestingParties),
+      filteredWitnessesClause = arrayIntersectionValues("tree_event_witnesses", requestingParties),
+      submittersInPartiesClause = arrayIntersectionWhereClause("submitters", requestingParties),
+      limitExpr = limitClause(limit),
+      fetchSizeHint = fetchSizeHint,
+      columnEqualityBoolean = columnEqualityBoolean("event_kind", "20"),
+    )(connection)
+
+  // TODO FIXME: confirm this works for oracle
```
nmarton-da, line 458: just checking: do you want something to do about this TODO, or shall we keep it and we take care about this TODO in cleanup/oracle perf tuning epic?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
```diff
@@ -0,0 +1,520 @@
...
+      "truncate table participant_events_non_consuming_exercise cascade",
+      "truncate table parties cascade",
+      "truncate table party_entries cascade",
+    ).map(SQL(_)).foreach(_.execute()(connection))
+
+  override def enforceSynchronousCommit(connection: Connection): Unit = () // Not supported
+
+  override def duplicateKeyError: String = "unique constraint"
+
+  val SQL_INSERT_COMMAND: String =
```
nmarton-da, line 47: is this the same SQL literal than what we have for H2?
if yes, can you add here a // TODO appen-only: reuse -- this is the same as the H2 counterpart

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: So comparing some string with a `null` value (`ledger_offset > null`) gives different results in Oracle and Postgres?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
rautenrieth-da: I understand that, I was more interested to see whether Oracle is smart enough to realize that it only needs to access the `participant_events_create` table and not the whole `participant_events` view (which is a union of all event partitions).

We could simplify a lot of queries using the above mentioned substitution, but so far there were not many arguments for doing it.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
nmarton-da: needed it for what? this approach bypasses the path what is layed down in the OracleFields.OracleTimestamp. This means even it is a timestamp it will be passed with setObject.
This is very disturbing...why is this working at all?
This is used for package insertion, it has a timestamp, and a mandatory one.
It seems to me the setTimestamp for the timestamp insertion is not needed....
But if that it not needed, then the whole .prepareData extension to the framework is not needed neither.
Do you know what is going on here?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
```diff
@@ -0,0 +1,64 @@
...
+      fields.view
+        .map(extractor)
+        .mkString(",")
+    val tableFields = commaSeparatedOf(_._1)
+    val selectFields = commaSeparatedOf { case (_, field) =>
+      field.selectFieldExpression("?")
+    }
+    val keyFieldName = fields(keyFieldIndex)._1
+    val keyFieldSelectExpression = fields(keyFieldIndex)._2.selectFieldExpression("?")
+    s"""MERGE INTO $tableName USING DUAL on ($keyFieldName = $keyFieldSelectExpression)
```
nmarton-da, line 51: is this the same as in H2? if yes we could just put this to common and make the default by viring to the common FieldStrategy

nmarton-da (commented):
looks good! great to see CI conformance tests!

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
nmarton-da: @rautenrieth-da I get your point, it might be worth double-checking it, but this is a wrong query for that here the event_kind = 20 populates the excercise_consuming field which is emitted to the resultset.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
nmarton-da: ahh I might have a theory: it is working because it is never null in this case, it might be that the null timestamps are not working, and they need to be put as .setTimestamp. If that is the case I would just change here to use the .prepareData (how I suggested above), and done

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
dasormeter: removed

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
dasormeter: Err I don't think we wrote this code, I think you did :) It's just copy pasta from the H2 implementation - I'm not sure I entirely understand the change you are requesting, but feel free to make it for both H2 and Oracle if you think it simplifies.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
dasormeter: ah this is because an empty string is considered `null` by oracle, so we have to add the extra check for null to handle that case for empty string https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Nulls.html#GUID-B0BA4751-9D88-426A-84AD-BCDBD5584071
```Note:Oracle Database currently treats a character value with a length of zero as null. However, this may not continue to be true in future releases, and Oracle recommends that you do not treat empty strings the same as nulls.```

File: ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala (hunk 14 above)
Comment (dasormeter) [reply ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala:894]: @rautenrieth-da when you query via the view oracle does a full table scan of all the tables associated with the view - so all the selects and union alls across all the tables.  when you select * from participant_events_create it does full table scan of just that table

Comment (dasormeter) [reply ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala:894]: when i query with an indexed field it does hit the indices of all the underlying tables behind the view though
image

File: ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala (hunk 15 above)
Comment (dasormeter) [reply ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala:406]: renamed as suggested

File: ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala (hunk 14 above)
Comment (dasormeter) [reply ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala:894]: i'm going to mark this as resolved -- if we want to do further optimization with view vs direct table queries let's do as a follow-on and try to approach it in a cross-db way

File: ledger/ledger-on-sql/BUILD.bazel [hunk 20]
```diff
@@ -359,6 +374,37 @@ conformance_test(
     ],
 )
 
+# TODO for Brian: please verify tags field, and concept, whether we can use this for CI conformance testing
```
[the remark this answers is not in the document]
dasormeter: removed

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,599 @@
...
+    application_id    NVARCHAR2(1000) not null,
+    submitters        CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
+    command_id        NVARCHAR2(1000) not null,
+
+    transaction_id    NVARCHAR2(1000), -- null if the command was rejected and checkpoints
+    status_code       INTEGER,         -- null for successful command and checkpoints
+    status_message    NVARCHAR2(1000)  -- null for successful command and checkpoints
+);
+
+create index participant_command_completions_idx on participant_command_completions(completion_offset, application_id);
```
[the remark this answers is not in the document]
dasormeter: corrected case

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,599 @@
...
+    tree_event_witnesses CLOB DEFAULT '[]' NOT NULL CONSTRAINT ensure_json_tree_event_witnesses CHECK (tree_event_witnesses IS JSON),       -- informees for create, exercise, and divulgance events
+
+    -- * divulgence and create events
+    create_argument BLOB,
+
+    -- * compression flags
+    create_argument_compression SMALLINT
+);
+
+---------------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
dasormeter: removed

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V100_0__Append_only_schema.sql
```diff
@@ -0,0 +1,599 @@
...
+    ledger_id                          NVARCHAR2(1000) not null,
+    -- stores the head offset, meant to change with every new ledger entry
+    ledger_end                         VARCHAR2(4000),
+    external_ledger_end                NVARCHAR2(1000),
+    participant_id                     NVARCHAR2(1000),
+    participant_pruned_up_to_inclusive VARCHAR2(4000),
+    ledger_end_sequential_id NUMBER
+);
+
+UPDATE parameters SET ledger_end_sequential_id = (
```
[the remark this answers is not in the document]
dasormeter: removed

ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql
```diff
@@ -154,8 +153,8 @@ CREATE UNIQUE INDEX idx_party_entries
 
 CREATE TABLE package_entries
 (
-    ledger_offset    VARCHAR2(4000)  not null,
-    recorded_at      timestamp       not null, --with timezone
+    ledger_offset    VARCHAR2(4000)  not null primary key,
+    recorded_at      timestamp       not null, --TODO BH in postgres this is with timezone
```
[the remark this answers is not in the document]
dasormeter: removed

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -480,7 +459,8 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
     Macro.parser[ParsedPackageData](
       "package_id",
       "source_description",
-      "size",
+      //confirm if change needed
```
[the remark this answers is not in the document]
dasormeter: removed

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -126,7 +127,7 @@ private[backend] object TemplatedStorageBackend {
                 COALESCE(divulgence_events.template_id, create_event_unrestricted.template_id),
                 COALESCE(divulgence_events.create_argument, create_event_unrestricted.create_argument),
                 COALESCE(divulgence_events.create_argument_compression, create_event_unrestricted.create_argument_compression)
-           FROM participant_events AS divulgence_events LEFT OUTER JOIN create_event_unrestricted USING (contract_id),
+           FROM participant_events divulgence_events LEFT OUTER JOIN create_event_unrestricted on (divulgence_events.contract_id = create_event_unrestricted.contract_id),
```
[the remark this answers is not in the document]
dasormeter: corrected case

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -289,6 +290,8 @@ private[backend] object TemplatedStorageBackend {
     Offset ~ String ~ Int ~ Long ~ String ~ String ~ Instant ~ Identifier ~ Option[String] ~
       Option[String] ~ Array[String]
 
+  import com.daml.platform.store.Conversions.ArrayColumnToStringArray.arrayColumnToStringArray
```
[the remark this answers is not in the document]
dasormeter: just to put it close to where it is used -- i moved it up to top of the object now

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -399,11 +404,12 @@ private[backend] object TemplatedStorageBackend {
...
       limitExpr: String,
       fetchSizeHint: Option[Int],
+      submittersInPartiesClause: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.FlatEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     SQL"""
-            select #$selectColumnsForTransactions, array[$party] as event_witnesses,
-                   case when submitters = #${partyArrayContext._1}$party#${partyArrayContext._2} then command_id else '' end as command_id
+            select #$selectColumnsForTransactions, #${partyArrayContext._1}$party#${partyArrayContext._2} as event_witnesses,
+                   case when #$submittersInPartiesClause then command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: it's not possible to do a straight equality comparison with a json array field, either needs to be a LIKE or the JSON_EXISTS specific query function we use here -- changed these to follow the same clause format as the other queries for simplicity.  is there any way to check for postgres perf regression off a PR?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -892,12 +910,14 @@ private[backend] object TemplatedStorageBackend {
...
+      columnEqualityBoolean: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.TreeEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     SQL"""
-        select #$selectColumnsForTransactionTree, array[$requestingParty] as event_witnesses,
-               event_kind = 20 as exercise_consuming,
-               case when submitters = #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} then command_id else '' end as command_id
+        select #$selectColumnsForTransactionTree, #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} as event_witnesses,
+               #$columnEqualityBoolean as exercise_consuming,
+               case when #$submittersInPartiesClause then command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: same as above

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -850,12 +865,14 @@ private[backend] object TemplatedStorageBackend {
...
+      columnEqualityBoolean: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.TreeEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     import com.daml.platform.store.Conversions.ledgerStringToStatement
-    SQL"""select #$selectColumnsForTransactionTree, array[$requestingParty] as event_witnesses,
-                 event_kind = 20 as exercise_consuming,
-                 case when submitters = #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} then command_id else '' end as command_id
+    SQL"""select #$selectColumnsForTransactionTree, #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} as event_witnesses,
+                 #$columnEqualityBoolean as exercise_consuming,
+                 case when #$submittersInPartiesClause then command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: same as above

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -726,11 +740,12 @@ private[backend] object TemplatedStorageBackend {
...
       partyArrayContext: (String, String),
       witnessesWhereClause: String,
+      submittersInPartiesClause: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.FlatEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     import com.daml.platform.store.Conversions.ledgerStringToStatement
-    SQL"""select #$selectColumnsForTransactions, array[$requestingParty] as event_witnesses,
-                 case when submitters = #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} then command_id else '' end as command_id
+    SQL"""select #$selectColumnsForTransactions, #${partyArrayContext._1}$requestingParty#${partyArrayContext._2} as event_witnesses,
+          case when #${submittersInPartiesClause} then command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: same as above

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -568,19 +576,21 @@ private[backend] object TemplatedStorageBackend {
...
   )(connection: Connection): Vector[EventsTable.Entry[Raw.FlatEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     import com.daml.platform.store.Conversions.OffsetToStatement
     import com.daml.platform.store.Conversions.IdentifierToStatement
-    SQL"""select #$selectColumnsForACS, array[$party] as event_witnesses,
-                   case when active_cs.submitters = #${partyArrayContext._1}$party#${partyArrayContext._2} then active_cs.command_id else '' end as command_id
-            from participant_events as active_cs
+    // TODO BH: figure out why interpolation is required for oracle but bombs h2 and postgres
+    SQL"""select #$selectColumnsForACS, #${partyArrayContext._1}$party#${partyArrayContext._2} as event_witnesses,
+                   case when #${submittersInPartiesClause} then active_cs.command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: same as above

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
```diff
@@ -535,18 +542,19 @@ private[backend] object TemplatedStorageBackend {
...
       fetchSizeHint: Option[Int],
+      submittersInPartiesClause: String,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.FlatEvent]] = {
     import com.daml.platform.store.Conversions.partyToStatement
     import com.daml.platform.store.Conversions.OffsetToStatement
-    SQL"""select #$selectColumnsForACS, array[$party] as event_witnesses,
-                   case when active_cs.submitters = #${partyArrayContext._1}$party#${partyArrayContext._2} then active_cs.command_id else '' end as command_id
-            from participant_events as active_cs
+    SQL"""select #$selectColumnsForACS, #${partyArrayContext._1}$party#${partyArrayContext._2} as event_witnesses,
+                   case when #${submittersInPartiesClause} then command_id else '' end as command_id
```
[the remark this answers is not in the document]
dasormeter: same as above

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleField.scala
```diff
@@ -0,0 +1,128 @@
...
+package com.daml.platform.store.backend.oracle
+
+import java.lang
+import java.sql.{PreparedStatement, Timestamp}
+import java.time.Instant
+import com.daml.platform.store.backend.common.{Field, TrivialField, TrivialOptionalField}
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+
+private[oracle] case class OracleStringField[FROM](extract: FROM => String)
```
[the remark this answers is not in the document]
dasormeter: cleaned up

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
dasormeter: No the Oracle error string is different so best to keep them separate

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
dasormeter: oracle needs parentheses around hte `($keyFieldName = $keyFieldSelectExpression)`, h2 does not : otherwise they should be equiv here- i'll try to extract out a common version that works for both

dasormeter: I'd rather leave this to refactoring as a follow-on PR as it involves some moving around of Table and Schema common code

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
dasormeter: removed

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -466,7 +477,8 @@ private[backend] object PostgresStorageBackend
 
   private def format(parties: Set[Party]): String = parties.view.map(p => s"'$p'").mkString(",")
 
-  private def limitClause(to: Option[Int]): String = to.map(to => s"limit $to").getOrElse("")
+  private def limitClause(to: Option[Int]): String =
+    to.map(to => s"fetch next $to rows only").getOrElse("")
```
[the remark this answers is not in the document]
dasormeter: yeah would rather leave as a TODO to cleanup in follow-on refactoring PR

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
rautenrieth-da: :+1:

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: Wow, not even string comparison is standartized in SQL, and Oracle evaluates `'' is null` to `true`?

The change makes sense now.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleTable.scala
nmarton-da: fixed

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: added todos

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/TemplatedStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: found a solution, check it out @dasormeter 
fixed

nmarton-da (approved):
LGTM, nice work!
@rautenrieth-da there will be couple of points captured in this PR which need to be dealt with/optimized later

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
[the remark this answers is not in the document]
rautenrieth-da: Are you tracking these TODOs somewhere? So far, we used to add `//TODO append-only:` to the code so that they can easily be found.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
```diff
@@ -0,0 +1,520 @@
...
+  )(connection: Connection): Vector[EventsTable.Entry[Raw.FlatEvent]] =
+    TemplatedStorageBackend.transactionsEventsSingleWildcardParty(
+      startExclusive = startExclusive,
+      endInclusive = endInclusive,
+      party = party,
+      partyArrayContext = partyArrayContext,
+      witnessesWhereClause = arrayIntersectionWhereClause("flat_event_witnesses", Set(party)),
+      limitExpr = limitClause(limit),
+      fetchSizeHint = fetchSizeHint,
+      submittersInPartiesClause = arrayIntersectionWhereClause("submitters", Set(party)),
```
rautenrieth-da: Previously, this used array equality, not array intersection. Would it be difficult to write a function that checks whether an array column contains exactly one given element?

In practice, the `transactionsEventsSingleWildcardParty` / `transactionsEventsSinglePartyWithTemplates` queries should be used much more often than `transactionsEventsOnlyWildcardParties` / `transactionsEventsSameTemplates`, as the latter are for the rarely used multi-party subscription feature.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
```diff
@@ -27,6 +28,20 @@ private[backend] abstract class Field[FROM, TO, CONVERTED](implicit
     input.view
       .map(extract andThen convert)
       .toArray(classTag)
+
+  final def prepareData(preparedStatement: PreparedStatement, index: Int, value: Any): Unit =
+    prepareDataTemplate(
+      preparedStatement,
+      index,
+      value.asInstanceOf[CONVERTED],
+    ) // this cast is safe by design
```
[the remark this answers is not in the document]
nmarton-da: well, the longer explanation is: if the Field-s prepareData is used for data which was created by the Field-s  toArray, which is the fundamental intention here, which with surrounding code must comply, then it is safe.

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -382,6 +386,7 @@ private[backend] object PostgresStorageBackend
       partyArrayContext = partyArrayContext,
       witnessesWhereClause =
         arrayIntersectionWhereClause("flat_event_witnesses", Set(requestingParty)),
+      submittersInPartiesClause = arrayIntersectionWhereClause("submitters", Set(requestingParty)),
```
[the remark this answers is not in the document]
nmarton-da: you were commenting on something outdated

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
nmarton-da: outdated again

rautenrieth-da (approved):

PR #8601, opened 2021-01-22
Removed logging JDBC url to prevent leaking passwords to log files. [DPP-140]

kamil-da:
Removed logging JDBC url to prevent leaking passwords to log files.

CHANGELOG_BEGIN
- not logging JDBC url
CHANGELOG_END

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [x] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10054, opened 2021-06-17
ledger: improve message for unsupported JDBC URL

meiersi-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
@@ -48,7 +48,7 @@ private[platform] object DbType {
     case pg if pg.startsWith("jdbc:postgresql:") => Postgres
     case oracle if oracle.startsWith("jdbc:oracle:") => Oracle
     case _ =>
-      sys.error(s"JDBC URL doesn't match any supported databases (h2, pg, oracle)")
+      sys.error(s"JDBC URL '$jdbcUrl' does not match any supported database (h2, postgresql, oracle)")
   }
 
   sealed trait AsyncCommitMode {


ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
```diff
@@ -48,7 +48,7 @@ private[platform] object DbType {
     case pg if pg.startsWith("jdbc:postgresql:") => Postgres
     case oracle if oracle.startsWith("jdbc:oracle:") => Oracle
     case _ =>
-      sys.error(s"JDBC URL doesn't match any supported databases (h2, pg, oracle)")
+      sys.error(s"JDBC URL '$jdbcUrl' does not match any supported database (h2, postgresql, oracle)")
```
cocreature, line 51: this looks like you leak passwords in the error. afaik we deliberately tried to avoid that everywhere at some point so going back to printing that out seems at least worth some consideration.

stefanobaghino-da: Cmp. https://github.com/digital-asset/daml/pull/8601

meiersi-da: o.O -- yes, that is indeed a problem. I've added it in the context of a command-line-tool call to `java -Xmx4g -jar bazel-bin/ledger/ledger-on-memory/app_deploy.jar dump-index-metadata` where the error message wasn't helping me understand what was wrong with my string.

Thanks for catching this. Let's close the PR then.

cocreature: fwiw in the JSON API we print jdbc urls in a few places but with passwords stripped which provides a middleground between debuggability & security.

PR #10060, opened 2021-06-21
[In-memory fan-out] Object liveness tracker 

tudor-da:
This PR introduces`ObjectLivenessTracker` which updates a counter based on the JVM reachability status of the objects it tracks. The feature can:
* Help discover memory leaks
* Track slow consumers and the buffer slices they are keeping alive (of the in-memory fan-out buffers for transactions).

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
+++ b/ledger/metrics/src/main/scala/com/daml/metrics/Metrics.scala
@@ -675,6 +675,9 @@ final class Metrics(val registry: MetricRegistry) {
             registry.timer(Prefix :+ "get_transaction_trees")
           val getFlatTransactions: Timer =
             registry.timer(Prefix :+ "get_flat_transactions")
+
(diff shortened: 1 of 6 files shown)


meiersi-da (commented):
@tudor-da : why did you choose this implementation over the one where the `LogEntryUpdate`s constructor and finalizer would increment and decrement the counter of how many instances there are?

I would expect the latter to have more predictable resource consumption, and still give us a good view on whether we are leaking these heavy objects.

Also: I believe we should add a comment to the code explaining why we are extra prudent wrt tracking the amount of live objects for `LogEntryUpdate`s. We might find that we cannot justify it or that there are also other like main-data-flow objects that we should be tracking.

tudor-da:
Observed experimentally that such an implementation leads to higher memory load. Discarded as not justified

PR #10066, opened 2021-06-21
kvutils: Refined transaction validation [KVL-1015]

hubert-da:
Implements transaction validation as described in the comment: https://github.com/digital-asset/daml/pull/10066#discussion_r664555881

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/keys/ContractKeysValidation.scala
```diff
@@ -42,18 +42,18 @@ private[transaction] object ContractKeysValidation {
         contractKeyDamlStateKeysToContractIds.map(m => m._1.getContractKey -> m._2)
 
       for {
-        stateAfterMonotonicityCheck <- checkContractKeysCausalMonotonicity(
+        stateAfterConsistencyChecks <- performTraversalContractKeysChecks(
```
cocreature, line 45: As far as I can tell with this change our checks are now ordered as follows:
1. Check key consistency for all nodes that have a key
2. Check key monotonicity
3. Reinterpret with the keys & cid lookups being checked via `contractIsActive` which not only checks activeness but also monotonicity. However, here we do not blow up with a monotonicity error. Instead we blow up with `Disputed`.

I _think_ this is fine if you have an honest submitter and will only produce bad error messages but there are a few points that worry me:

1. You do checks on inputs before checking that the transaction is okay, e.g., well authorized. This seems potentially leaky. E.g., I send you a negative lookup by key that is ill-authorized. You tell me that there is a mismatch and the key actually exists. Similarly you can leak activeness checks. If you trust your participant that may be not an issue but it’s at least sketchy.
2. You report monotonicity errors differently for keys & contract ids. For keys there is a dedicated error. For contract ids they are folded under general "transaction mismatch" so `Disputed` afaict. Not necessarily an issue I think but at least confusing.
3. You mix up activeness of contract ids with monotonicity afaict and both are handled implicitly during reinterpretation. That seems potentially problematic. Activeness can be a race. Monotonicity is a malicious participant.Both are reported the same way.

As mentioned above, I’m not sure you can exploit any of this without a malicious participant but I think the same was also true for the previous order of the checks so if we fix this it might be worth fixing it properly.

I’m by no means an expert on kvutils validation so take all of this with a grain of salt. I might just be missing something and this is all perfectly sensible but until someone explains it to me I’m at least not comfortable approving it.

gerolf-da: Thanks, @cocreature. I had a strong conviction that the model conformance was checked before contract key consistency that I read past it. 

For this PR: The order of the steps `validate_contract_keys` and `validate_model_conformance` should be changed.

For a later review: I think there should be a single monotonicity check for both contracts and keys after the model validation, and the contract lookup during interpretation shouldn't do a ledger effective time check.

cocreature: If you do model conformance before contract keys validation, then the example that produced `causal monotonicity violation` will fail with Disputed I believe which seems arguably worse.

cocreature: Ah no I think works for the case we encountered but breaks others. This is what we saw:
1. During submission, the key is inactive.
2. During model conformance during validation the key is active. The arguably incorrectly placed ledger effective time check however saves you and means interpretation will still treat the key as inactive.
3. So because of that you get past model conformance and than we fail with InconsistentKeys.

However now consider a case where in 2 the key is active and passes the monotonicity check. This will now blow up with `Disputed` even though it was just a race.

What seems questionable here to me (and I could just be missing something) is that you check model conformance with the current ledger state instead of the one during submission. If you look at the ledger model, model conformance does not depend on the ledger state. It is a property of a transaction.

So I think the checks maybe should look like the following:

1. Check conformance against the ledger state at submission. Reinterpret using that state, diff. Monotonicity checks can be done here but again they should be done against the legder state at submission not the one during validation.
2. Assuming the transaction is conformant, check consistency of key & contract ids.

hubert-da: JFYI, I’m currently looking into how the validation works in depth and will come up with a proper fix next week. On Mon and Tue I'm going to be OoO.

gerolf-da: Summary of a conversation with @cocreature, @hubert-da, @gerolf-da: 

We need to do two types of checks:
1. check the conformance of the transaction with the Daml model purely in terms of the transaction inputs itself.
A violation in this category generally falls into the `Disputed` class of errors.
2. check that the transaction is consistent with the current ledger state
A violation in this category generally falls into the `Inconsistent` class of errors.

### Transaction model conformance

1. Validate _model conformance_ of the transaction
    - use the contract keys of the transaction (`Transaction#contractKeyInput`) as input to `TransactionCommitter#lookupKey` instead of the contract keys from the commit context
    - because fetch nodes don't actually contain the contract, we still need to get the contracts from the commit context, but we must not check for activeness or causal monotonicity in `TransactionCommitter#lookupContract` during re-interpretation.
    - Any error in validation should result in `Disputed`.

2. check _causal monotonicity_ based on contracts used in the transaction (`GenTransaction#inputContracts`, contains contracts used in exercise, fetch, lookupByKey)
    - violations of causal monotonicity should result in `InvalidLedgerTime`.

### Check that transaction is still valid given the ledger state

1. check consistency of keys
     - check that keys in the transaction are the same as keys in state (already done in `ContractKeysValidation#performTraversalContractKeysChecks`)
2. check consistency of contracts
    - all contracts used in the transaction (`GenTransaction#inputContracts`) are still active. A violation results in the error `Inconsistent`.

### Cleanup
- Remove `KeyMonotonicityValidation`, it is superseded by the causal monotonicity check during model conformance validation.

### Ledger API Test Tool
The above changes will result in different errors than what the ledger api test tool currently checks.
We should also document in the release notes in which cases the error is now different.
----

@andreaslochbihler-da, @cocreature, @miklos-da: Could you please review this summary and point out any issues or give a 👍?

cocreature: sounds good to me, the only thing I’d add is that you might want to treat the problem around ledger time of divulged contracts as related since that somewhat destroys the assumption that model conformance can only be violated by a broken submitter.

andreaslochbihler-da: :+1: The summary sounds good.

I don't have a strong opinion on causal monotonicity violations ending in `InvalidLedgerTime` vs. `Disputed`. It could just as well be disputed because the submitting participant knows that the chosen ledger time is invalid, so it shouldn't have sent the transaction in the first place. I don't think that this difference can be tested with the ledger conformance tests right now because the participant server should never produce a transaction that violates causal monotonicity.

cocreature: I think they should be InvalidLedgerTime at least until we fixed the issue around divulged contracts I mentioned above. Otherwise you can get disputed from an honest submitter which is bad imho. Once that is fixed I agree with you that they can be treated the same.

gerolf-da: Thanks for the reminder, @cocreature.
There is a scenario where a participant could submit a transaction that violates causal monotonicity:
1. The participant's system clock lags behind in time of the committer's system clock.
2. The transaction only references divulged contracts, for which the participant doesn't (yet) keep a "substitute" ledger effective time (e.g. the LET from the divulging transaction), that were created and divulged within the window of the participants clock drift.
3. Therefore the participant will use its (lagging) system clock to determine LET and the contract's actual LET will be in the "future" compared to the submitted ledger effective time of the transaction.

I don't think we should fix this issue as part of KVL-990. And it already happens currently as well. So technically the situation doesn't get worse.

cocreature: Fine to leave it out as long as we don’t report is as Disputed which would be worse or at least different from the current state.

andreaslochbihler-da: > There is a scenario where a participant could submit a transaction that violates causal monotonicity:

Oh right. I forgot that the participant server doesn't keep track of ledger create time for divulged contracts (something we do in Canton).

gerolf-da: > Fine to leave it out as long as we don’t report is as Disputed which would be worse or at least different from the current state.

Currently, prior to the change described above, this case would also end up with a `Disputed` error.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
+      recordTime: Option[Timestamp],
+      rejections: Rejections,
+  )(
+      error: KeyInputError
+  )(implicit loggingContext: LoggingContext): StepResult[DamlTransactionEntrySummary] = {
+    val description = error match {
+      case DuplicateKeys(_) =>
+        "DuplicateKeys: the transaction contains a duplicate key"
+      case InconsistentKeys(_) =>
+        "InconsistentKeys: the transaction is internally inconsistent"
```
hubert-da, line 237: Do we want to return `RejectionReasonV0.Inconsistent` here even though it is the model conformance validation?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
+    }
+    rejections.buildRejectionStep(
+      transactionEntry,
+      RejectionReasonV0.Disputed(description),
+      recordTime,
+    )
+  }
+
   def rejectionReasonForValidationError(
```
hubert-da, line 246: Do we still want to return `Inconsistent` here (for the key lookups)?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -145,9 +343,70 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
   private val aKey = "key"
+  private val aKeyMaintainer = "maintainer"
   private val aDummyValue = TransactionBuilder.record("field" -> "value")
+  private val aTemplateId = "dummyPackage:DummyModule:DummyTemplate"
+
+  private val aSubmissionSeed = ByteString.copyFromUtf8("a" * 32)
+  private val ledgerEffectiveTime =
+    ZonedDateTime.of(2021, 1, 1, 12, 0, 0, 0, ZoneOffset.UTC).toInstant
+
+  private val aContractIdStateValue = {
```
hubert-da, line 361: Is it possible to make it more concise?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -145,9 +343,70 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
+          )
+          .build()
+      )
+      .build()
+  }
+
+  private def txBuilder = TransactionBuilder(TransactionVersion.VDev)
+
+  private def aGlobalKeyWithMaintainers(key: String, maintainer: String) = GlobalKeyWithMaintainers(
+    GlobalKey.assertBuild(
```
hubert-da, line 391: Is it possible to do it easier?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
+        val activeAt = Option(inputContractState.getActiveAt).map(parseTimestamp)
+        activeAt.exists(transactionEntry.ledgerEffectiveTime >= _)
     }
+
+    if (isCasualMonotonicityHeld)
+      StepContinue(transactionEntry)
+    else
+      rejections.buildRejectionStep(
+        transactionEntry,
+        RejectionReasonV0.InvalidLedgerTime("Causal monotonicity violated"),
```
hubert-da, line 218: @gerolf-da @moritzkiefer-da @miklos-da The existing tests haven't detected any changes in rejection reasons, as well as error codes. I would expect there is a change, at least here. Even though the test coverage is undoubtedly greater than before this PR, it would be good to write more tests. Also, I have no idea how it behaves in racy conditions like in the recent incident. Do you think it would be worth investing some time to check it (possibly as a separate story)?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -50,12 +95,164 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
+            ),
+          ),
+          transactionEntry1,
+        )
+      step shouldBe a[StepStop]
+      step
+        .asInstanceOf[StepStop]
+        .logEntry
+        .getTransactionRejectionEntry
+        .hasInconsistent shouldBe true
```
gerolf-da, line 175: As per my comment above, I think a `ReplayMismatch` error should resolve to a `Disputed` error.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
gerolf-da: I'm afraid not. That's the verbosity of protobuf :/

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
cocreature: yes this is model conformance -> malicious participant -> disputed.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
```diff
@@ -116,11 +108,38 @@ private[transaction] object ContractKeysValidator extends TransactionValidator {
...
+      !inputContracts(contractId).hasArchivedAt
+    )
+
+    if (areContractsConsistent)
+      StepContinue(transactionEntry)
+    else
+      rejections.buildRejectionStep(
+        transactionEntry,
+        RejectionReasonV0.Inconsistent(
+          "InconsistentContracts: at least one contract has been archived since the submission"
```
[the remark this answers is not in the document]
cocreature: I had a discussion with @andreaslochbihler-da around this a week ago. What should be fine is to show the following:
1. The list of contract ids that are no longer active. A handwavy argument for why this is not an issue is to consider a transaction with only a single contract where showing the contract id provides no additional information.
2. The list of contracts keys that have inconsistent assignments. Similar handwavy argument.

What you must not show is the following:
Contract key assignments in the case of inconsistent keys.
This already matters for an honest participant: A key lookup can be mapped to none if no stakeholder is on the participant.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
-    val stateKey = Conversions.globalKeyToStateKey(key.globalKey)
-    val contractId = for {
-      stateValue <- commitContext.read(stateKey)
-      if stateValue.getContractKeyState.getContractId.nonEmpty
-    } yield decodeContractId(stateValue.getContractKeyState.getContractId)
+    contractKeyInputs.get(key.globalKey) match {
+      case Some(KeyActive(cid)) => Some(cid)
+      case _ => None
+    }
+  }
```
fabiotudone-da, line 192: `{` and `}` shouldn't be needed here.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
-    // evidenced in the transaction. The activeness of the contract is checked when it is fetched.
-    contractId.orElse {
-      knownKeys.get(stateKey.getContractKey)
+    val inputContracts: Map[Value.ContractId, DamlContractState] = commitContext
+      .collectInputs {
+        case (key, Some(value)) if value.hasContractState =>
+          Conversions.stateKeyToContractId(key) -> value.getContractState
+      }
+
+    val isCasualMonotonicityHeld = transactionEntry.transaction.inputContracts.forall {
```
fabiotudone-da, line 206:
```suggestion
    val isCausallyMonotonic = transactionEntry.transaction.inputContracts.forall {
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -170,37 +182,67 @@ private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Me
...
-      stateValue <- commitContext.read(stateKey)
-      if stateValue.getContractKeyState.getContractId.nonEmpty
-    } yield decodeContractId(stateValue.getContractKeyState.getContractId)
+    contractKeyInputs.get(key.globalKey) match {
+      case Some(KeyActive(cid)) => Some(cid)
+      case _ => None
+    }
+  }
+
+  private[validation] def validateCausalMonotonicity(
```
fabiotudone-da, line 194:
```suggestion
  // Checks that input contracts are still active at ledger effective time.
  private[validation] def validateCausalMonotonicity(
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -40,99 +44,107 @@ import com.daml.logging.LoggingContext.withEnrichedLoggingContext
...
-      if contractIsActive(transactionEntry, contractState)
-      contract = Conversions.decodeContractInstance(contractState.getContractInstance)
-    } yield contract
+    // There is the possibility that the reinterpretation of the transaction yields a different
+    // result in a LookupByKey than the original transaction. This means that the contract state data for the
+    // contractId pointed to by that contractKey might not have been preloaded into the input state map.
+    // This is not a problem because after the transaction reinterpretation, we compare the original
+    // transaction with the reinterpreted one, and the LookupByKey node will not match.
+    commitContext
+      .read(stateKey)
```
hubert-da, line 144: A question from me and @fabiotudone-da: what if a contract is not in the state anymore?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
```diff
@@ -6,82 +6,74 @@ package com.daml.ledger.participant.state.kvutils.committer.transaction.validati
...
+    val contractKeysToContractIds: Map[DamlContractKey, RawContractId] = contractKeyState.collect {
+      case (k, v) if v.getContractId.nonEmpty =>
+        k.getContractKey -> v.getContractId
+    }
 
     val transaction = transactionEntry.transaction
 
+    import scalaz.std.either._
+    import scalaz.std.list._
+    import scalaz.syntax.foldable._
```
fabiotudone-da, line 76: Where is that used?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidatorSpec.scala
```diff
@@ -221,6 +230,28 @@ class ContractKeysValidatorSpec extends AnyWordSpec with Matchers {
...
+              )
+              .build()
+          )
+        ),
+      )
+      val builder = TransactionBuilder()
+      builder.add(archive(globalCreate, Set("Alice")))
+      val transaction = builder.buildSubmitted()
+      val result = validate(context, transaction)
+      result shouldBe a[StepStop]
```
fabiotudone-da, line 253: What is the expected rejection reason?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -50,12 +95,164 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
+            Map(
+              inputContractIdStateKey -> Some(makeContractIdStateValue()),
+              contractIdStateKey1 -> Some(aStateValueActiveAt(ledgerEffectiveTime.minusSeconds(1))),
+            ),
+          ),
+          rejections,
+        ) shouldBe StepContinue(transactionEntry1)
+    }
+
+    "reject transaction when causal monotonicity is not held" in {
```
fabiotudone-da, line 229:
```suggestion
    "reject transaction when causal monotonicity does not hold" in {
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -50,12 +95,164 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
+
+    "return None when mapping does not exist" in {
+      modelConformanceValidator.lookupKey(contractKeyInputs)(
+        aGlobalKeyWithMaintainers("nonexistentKey", "nonexistentMaintainer")
+      ) shouldBe None
+    }
+  }
+
+  "validateCausalMonotonicity" should {
+    "create StepContinue when causal monotonicity is held" in {
```
fabiotudone-da, line 214:
```suggestion
    "create StepContinue when causal monotonicity holds" in {
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -50,12 +95,164 @@ class ModelConformanceValidatorSpec extends AnyWordSpec with Matchers with Mocki
...
+      modelConformanceValidator.lookupContract(
+        createCommitContext(
+          None,
+          Map(
+            inputContractIdStateKey -> Some(
+              aContractIdStateValue
+            )
+          ),
+        )
+      )(Conversions.decodeContractId(inputContractId)) shouldBe a[Some[_]]
```
fabiotudone-da, line 190: Can this be more specific?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
```diff
@@ -6,82 +6,74 @@ package com.daml.ledger.participant.state.kvutils.committer.transaction.validati
...
+import com.daml.lf.value.Value
 import com.daml.logging.LoggingContext
 
-private[transaction] object ContractKeysValidator extends TransactionValidator {
+private[transaction] object TransactionConsistencyValidator extends TransactionValidator {
 
-  /** Creates a committer step that validates casual monotonicity and consistency of contract keys
-    * against the current ledger state.
+  /** Validates consistency of contracts and contract keys against the current ledger state.
+    * For contracts, checks whether all contracts used in the transaction are still active.
```
fabiotudone-da, line 33: And for keys?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
hubert-da: `traverse_` requires that

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
hubert-da: If they are consistent and there are no duplicates. Happy to add something if there are any suggestions.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidatorSpec.scala
hubert-da: Good catch, `Inconsistent`, adding this check.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
hubert-da: Good point, I was going to change this.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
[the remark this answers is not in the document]
hubert-da: As it is going to be an improvement to error messages, let me do it in a separate PR, please. This one is quite heavy already.

hubert-da:
@gerolf-da @moritzkiefer-da @fabiotudone-da JFYI I addressed the review comments

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
gerolf-da: It can only not be in the state, if the contract was archived and pruned on the committer.
The question is then, why was the participant able to produce such a transaction? For which the answer is: it used a divulged contract that appeared as active, because it didn't learn about the archival. On the other hand, using divulged contracts for interpretation has been [deprecated recently](https://docs.daml.com/support/component-statuses.html#ledger-api).

Technically I think it should result in an `Inconsistent` error though.

hubert-da: Thanks! I was also thinking about a racy situation.

cocreature: I don’t think you can get it through a race:

1. The participant will only use a contract if it’s been committed.
2. If it’s been committed it’s in the committer state and it stays there until committer pruning.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -40,99 +44,107 @@ import com.daml.logging.LoggingContext.withEnrichedLoggingContext
...
+  */
 private[transaction] class ModelConformanceValidator(engine: Engine, metrics: Metrics)
     extends TransactionValidator {
   import ModelConformanceValidator._
 
   private final val logger = ContextualizedLogger.get(getClass)
 
-  /** Creates a committer step that validates the submission's conformance to the Daml model. */
+  /** Validates model conformance based on the transaction itself (where it's possible).
+    * Because fetch nodes don't contain contracts, we still need to get them from the current state ([[CommitContext]]).
```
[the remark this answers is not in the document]
cocreature: See https://github.com/digital-asset/daml/pull/10066#discussion_r670611657, the only case where you don’t have it in the ledger state should be in some edge cases around comitter pruning.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
hubert-da: Thanks again, I get it now 👍🏻

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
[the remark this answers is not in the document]
hubert-da: Thanks. @fabiotudone-da I added a comment, are you satisfied with the conclusion?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -24,116 +19,121 @@ import com.daml.ledger.participant.state.kvutils.{Conversions, Err}
...
   }
 
-  // Helper to lookup contract instances. We verify the activeness of
-  // contract instances here. Since we look up every contract that was
+  // Helper to lookup contract instances. Since we look up every contract that was
   // an input to a transaction, we do not need to verify the inputs separately.
-  private def lookupContract(
-      transactionEntry: DamlTransactionEntrySummary,
-      commitContext: CommitContext,
+  // A contract may not be in the state only if it's an edge case around committer pruning.
```
fabiotudone-da, line 125: if the contract was archived and pruned on the committer.
```suggestion
  // Note that a contract may not be in the state only if it was archived and pruned on the committer.
  // Then, the participant able to produce such a transaction only by using a divulged contract that
  // appeared as active, because it didn't learn about the archival. On the other hand, using divulged
  // contracts for interpretation is deprecated.

File: ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala (hunk 18 above)
Comment (fabiotudone-da) [reply ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala:33]:
```suggestion
    * For contracts, checks whether all contracts used in the transaction are still active.
    * For keys, checks whether they are consistent and there are no duplicates.
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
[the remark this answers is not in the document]
fabiotudone-da: Yep, I'd be a bit more explicit in the comment though, pls. see my latest review.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
hubert-da: Done

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
hubert-da: Done

fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
```diff
@@ -6,82 +6,75 @@ package com.daml.ledger.participant.state.kvutils.committer.transaction.validati
...
+        stepResult <- validateConsistencyOfKeys(
+          commitContext,
           transactionEntry,
           rejections,
         )
-        finalState <- performTraversalContractKeysChecks(
-          commitContext.recordTime,
-          contractKeysToContractIds,
-          stateAfterMonotonicityCheck,
+        finalStepResult <- validateConsistencyOfContracts(
```
cocreature, line 50: did you deliberately check key consistency before contract consistency? I don’t think it matters apart from error messages and I haven’t really thought through which order gives us better error messages so just curious.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidatorSpec.scala
```diff
@@ -221,6 +230,33 @@ class ContractKeysValidatorSpec extends AnyWordSpec with Matchers {
...
+      result
```
cocreature, line 254: maybe consider an `inside` instead of a call to `asInstanceOf`.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
cocreature: don’t think so

cocreature (approved):
Nice work, thank you!

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -24,116 +19,125 @@ import com.daml.ledger.participant.state.kvutils.{Conversions, Err}
...
+  // Helper to lookup contract instances. Since we look up every contract that was
   // an input to a transaction, we do not need to verify the inputs separately.
-  private def lookupContract(
-      transactionEntry: DamlTransactionEntrySummary,
-      commitContext: CommitContext,
+  //
+  // Note that a contract may not be in the state only if it was archived and pruned on the committer.
+  // Then, the participant is able to produce such a transaction only by using a divulged contract that
+  // appeared as active, because it didn't learn about the archival. On the other hand, using divulged
+  // contracts for interpretation is deprecated.
```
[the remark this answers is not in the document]
hubert-da: According to this comment, it's now `Inconsistent`: https://github.com/digital-asset/daml/pull/10066#discussion_r672076310

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
hubert-da: I was also thinking about this. However, I just picked the order from the description: https://github.com/digital-asset/daml/pull/10066#discussion_r664555881 Shall we revisit this?

cocreature: I think this order makes sense or is at least fine. And in the case where a contract key changed value and the old assignment is no longer active, I think it’s more useful to report this as an InconsistentKeys than an inactive contract and that’s what your implementation does :+1:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -24,116 +19,125 @@ import com.daml.ledger.participant.state.kvutils.{Conversions, Err}
...
+      rejections: Rejections,
+  )(implicit loggingContext: LoggingContext): StepResult[DamlTransactionEntrySummary] = {
+    try {
+      val stepResult = for {
+        contractKeyInputs <- transactionEntry.transaction.contractKeyInputs.left
+          .map(rejectionForKeyInputError(transactionEntry, commitContext.recordTime, rejections))
+        _ <- validationResult
+          .consume(
+            lookupContract(commitContext),
+            lookupPackage(commitContext),
```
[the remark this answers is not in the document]
hubert-da: Not all of them. Errors that originate from the current ledger state i.e. `Err.MissingInputState` and `Err.DecodeError` end up with `Inconsistent` (see https://github.com/digital-asset/daml/pull/10066#discussion_r672076310). I'm actually not sure about `Err.DecodeError`. I wanted to preserve the current `lookupPackage` behavior and ignored it. Your comment made me realize that it was broken from the very beginning. There was no catch clause for `Err.DecodeError`. I changed it to a catch-all `Err` block and added more tests. Please check if it makes sense.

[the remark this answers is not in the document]
cocreature: I’m not quite sure Inconsistent makes sense.
1. For `DecodeError`, I think we’re really talking about an internal error. We’re not failing to decode what the participant send us, we’re failing to decode our own state. That should never happen and imho we shouldn’t report it as Inconsistent.
2. A missing package is similar to the missing contract case discussed in the other comment thread. However, there is no committer pruning of packages. So I think this can really only arise from a malicious/buggy participant. I think putting it under Inconsistent is fine since it’s not technically model conformance (because it isn’t a property of the transaction itself) but maybe at least add a comment that this can only arise from a malicous submitter.
3. For missing contracts, I think Inconsistent is probably the best choice.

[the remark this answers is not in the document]
hubert-da: 1. We can also not catch it at all, as it was previously. @SamirTalwar-DA told me that it's possible if we turn off the package validation in the Engine, and suggested a new rejection reason, which we probably don't want to introduce now.
2. Will add it tomorrow, thanks!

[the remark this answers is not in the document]
gerolf-da: 1) `DecodeError`: I agree that it should not be reported as inconsistent but rather as disputed. From the documentation of disputed:
> This means that the underlying ledger and its validation logic
    considered the transaction potentially invalid. This can be due to a bug
    in the submission or validation logic, or due to malicious behaviour.

2) I agree with @cocreature's assessment.
3) I agree with @cocreature's assessment.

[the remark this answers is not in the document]
gerolf-da: Another note regarding 1): In fact, we already have an appropriate exception that could be thrown, bot be more explicit about the specific case: `Err.ArchiveDecodingFailed`.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
```diff
@@ -24,121 +19,133 @@ import com.daml.ledger.participant.state.kvutils.{Conversions, Err}
...
-      transactionEntry: DamlTransactionEntrySummary,
-      contractState: DamlContractState,
-  ): Boolean = {
-    val activeAt = Option(contractState.getActiveAt).map(parseTimestamp)
-    !contractState.hasArchivedAt && activeAt.exists(transactionEntry.ledgerEffectiveTime >= _)
+          )
+      } yield ()
+      stepResult.fold(identity, _ => StepContinue(transactionEntry))
+    } catch {
+      case err: Err =>
```
gerolf-da, line 111: Here we need to differentiate between different "consumption" errors:
* `Err.MissingInputState` => `Inconsistent`
* `Err.DecodeError` => `Disputed`

hubert-da: Done

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidator.scala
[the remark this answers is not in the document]
hubert-da: Done, thanks

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/TransactionConsistencyValidator.scala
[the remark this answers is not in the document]
hubert-da: Added a new story for it: https://digitalasset.atlassian.net/browse/KVL-1026

PR #10067, opened 2021-06-21
fix 9_functional redirect

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END


realvictorprm (approved):

garyverhaegen-da:
@cocreature @bame-da Not sure what we should do for the existing deployment. It's easy enough to fix manually, though I generally don't like doing that.

cocreature:
Not that fuzzed about a broken redirect so I wouldn’t fix it manually but if @bame-da disagrees, I’m fine with that.

PR #10068, opened 2021-06-21
[JSON-API] Add db metrics & response construction metrics

realvictorprm:
Solves the last bits of #10020 except the `For command submissions the duration of the request to the ledger.` point.

changelog_begin

- [JSON-API] The database operations (regardless of in-memory or postgres) contain now metrics for fetching contracts by id or key (seperate metrics foreach)
- [JSON-API] The time for a repsonse payload construction of a request is now also tracked in a metric

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -471,22 +471,27 @@ class Endpoints(
 
   private def httpResponse[A: JsonWriter](
       result: ET[domain.SyncResponse[A]]
-  ): Future[HttpResponse] = {
-    val fa: Future[Error \/ (JsValue, StatusCode)] =
-      result.flatMap { x =>
-        either(SprayJson.encode1(x).map(y => (y, x.status)).liftErr(ServerError))
-      }.run
-    fa.map {
-      case -\/(e) =>
-        httpResponseError(e)
-      case \/-((jsVal, status)) =>
-        HttpResponse(
-          entity = HttpEntity.Strict(ContentTypes.`application/json`, format(jsVal)),
-          status = status,
-        )
-    }.recover { case NonFatal(e) =>
-      httpResponseError(ServerError(e.description))
-    }
+  )(implicit metrics: Metrics): Future[HttpResponse] = {
+    Timed.future(
+      metrics.daml.HttpJsonApi.responseCreationTimer,
+      result
+        .flatMap { x =>
+          either(SprayJson.encode1(x).map(y => (y, x.status)).liftErr(ServerError))
+        }
+        .run
+        .map {
+          case -\/(e) =>
+            httpResponseError(e)
+          case \/-((jsVal, status)) =>
+            HttpResponse(
+              entity = HttpEntity.Strict(ContentTypes.`application/json`, format(jsVal)),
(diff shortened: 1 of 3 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -471,22 +471,27 @@ class Endpoints(
...
+          case -\/(e) =>
+            httpResponseError(e)
+          case \/-((jsVal, status)) =>
+            HttpResponse(
+              entity = HttpEntity.Strict(ContentTypes.`application/json`, format(jsVal)),
+              status = status,
+            )
+        }
+        .recover { case NonFatal(e) =>
+          httpResponseError(ServerError(e.description))
```
stefanobaghino-da, line 492: I understand that this comes from existing code, but should we also take advantage of this chance to log this exception?

stefanobaghino-da (approved):

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
S11001001: I am generally against logging anything that is reported by proper channels already; it's a good way to leak sensitive data.

stefanobaghino-da: We are recovering from a server error here, I think it makes sense for the information to be logged on the server. I would argue it would probably make less sense to return the full description to the client (e.g. if the connection to the database fails I'd be likely be more interested to have it logged rather than reported to the client). Does that make sense?

S11001001: If you can honestly say "don't report the description to the client at all in this case", then it makes sense. Otherwise, we're pointing in one direction and jumping in the other.

stefanobaghino-da: If the errors caught here are meaningful to be reported in full only to the operator of the server, then yes. There are cases in which both a detailed report to the user _and_ logging to the advantage of the operator could make sense. Not sure of what could be thrown here. It looks to me like here most exceptions would be interesting from an operational point of view but less so from a user perspective, but I'll let you judge that.

realvictorprm: This escalated quickly :joy:

Thing is, how do we know whether the information is either best only for the user or the operator of the server?

realvictorprm: I want to postpone this to a seperate pr and better discuss this in slack or in an gh issue

realvictorprm:
If there are no further objections I would merge this

stefanobaghino-da:
> If there are no further objections I would merge this

If you have an approval, go for it. 🙂 

If there are open points on an approved PR it means they can be addressed in a separate PR, as long as it's done in a timely fashion.

Our approach is well described [here](http://neilmitchell.blogspot.com/2019/04/code-review-approve-with-suggestions.html).

realvictorprm:
Then LGTM :grin: First work PR to be merged today and second PR overall to be merged today :grin:

PR #10069, opened 2021-06-21
allow CI pools to use any zone

garyverhaegen-da:
This morning we started with very restricted CI pools (2/6 for Windows and 7/20 for Linux), apparently because the region we run in (us-east1) has three zones, two of them were unable to allocate new nodes, and the default policy is to distribute nodes evenly between zones.

I've manually changed the distribution policy. Unfortunately this option is not yet available in our version of the GCP Terraform plugin.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/modules/gcp_cdn_bucket/outputs.tf
+++ b/infra/modules/gcp_cdn_bucket/outputs.tf
@@ -3,10 +3,10 @@
 
 output external_ip {
   description = "The external IP assigned to the global fowarding rule."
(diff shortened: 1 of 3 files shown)


garyverhaegen-da:
Related [docs](https://cloud.google.com/compute/docs/instance-groups/regional-mig-distribution-shape).

cocreature (approved):
Thanks!

PR #10070, opened 2021-06-21
Run scenarios in off-ledger machine

cocreature:
This PR builds on the previous PR that split scenario execution in two
different speedy machines and now actually makes the machine that runs
scenarios run in off-ledger mode just like we handle Daml Script.

This required a bunch of refactoring to make it nice so apologies for
the slightly large PR. Hopefully it’s still relatively easy to follow
and luckily it deletes more code than it adds.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -29,8 +29,7 @@ final class Conversions(
 
   // The ledger data will not contain information from the partial transaction at this point.
   // We need the mapping for converting error message so we manually add it here.
-  private val ptxCoidToNodeId = incomplete
-    .fold[Map[NodeId, N.GenNode[NodeId, V.ContractId]]](Map.empty)(_.transaction.nodes)
+  private val ptxCoidToNodeId = incomplete.map(_.transaction.nodes).getOrElse(Map.empty)
     .collect { case (nodeId, node: N.NodeCreate[V.ContractId]) =>
       node.coid -> ledger.ptxEventId(nodeId)
     }
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -29,7 +29,9 @@ final class Conversions(
 
   // The ledger data will not contain information from the partial transaction at this point.
   // We need the mapping for converting error message so we manually add it here.
-  private val ptxCoidToNodeId = incomplete.map(_.transaction.nodes).getOrElse(Map.empty)
+  private val ptxCoidToNodeId = incomplete
+    .map(_.transaction.nodes)
+    .getOrElse(Map.empty)
     .collect { case (nodeId, node: N.NodeCreate[V.ContractId]) =>
       node.coid -> ledger.ptxEventId(nodeId)
     }
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -16,7 +16,7 @@ import scala.jdk.CollectionConverters._
 final class Conversions(
     homePackageId: Ref.PackageId,
     ledger: ScenarioLedger,
-    incomplete: IncompleteTransaction,
+    incomplete: Option[IncompleteTransaction],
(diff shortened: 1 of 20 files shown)


compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -29,7 +29,8 @@ final class Conversions(
 
   // The ledger data will not contain information from the partial transaction at this point.
   // We need the mapping for converting error message so we manually add it here.
-  private val ptxCoidToNodeId = incomplete.transaction.nodes
+  private val ptxCoidToNodeId = incomplete
+    .fold[Map[NodeId, N.GenNode[NodeId, V.ContractId]]](Map.empty)(_.transaction.nodes)
```
stefanobaghino-da, line 33:
```suggestion
  private val ptxCoidToNodeId = incomplete.map(_.transaction.nodes).getOrElse(Map.empty)
```
I prefer `fold` as well in general but the need to explicitly spell out the type could make `map`/`getOrElse` preferable here. Feel free to discard this suggestion.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
```diff
@@ -57,7 +57,7 @@ private[lf] object Pretty {
   def prettyParty(p: Party): Doc =
     char('\'') + text(p) + char('\'')
 
-  def prettyDamlException(ex: SErrorDamlException, ptx: PartialTransaction): Doc =
+  private def prettyDamlException(ex: SErrorDamlException, ptx: PartialTransaction): Doc =
```
stefanobaghino-da, line 60: 🚀

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -726,6 +726,8 @@ private[lf] object Speedy {
 
     private val damlTraceLog = LoggerFactory.getLogger("daml.tracelog")
 
+    def newTraceLog: TraceLog = RingBufferTraceLog(damlTraceLog, 100)
```
stefanobaghino-da, line 729: Where you interested in readability or were you unsure whether `RingBufferTraceLog` was actually constructed at every call? From some experimentation it looks like it should.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
```diff
@@ -451,4 +443,29 @@ object ScenarioRunner {
       // to avoid breaking all tests we keep it for now at least.
       Time.Timestamp.MinValue,
     )
+
+  sealed abstract class ScenarioResult {
+    def ledger: ScenarioLedger
+    def traceLog: TraceLog
+  }
```
stefanobaghino-da, line 450:
```suggestion
  } extends Product with Serializable
```
I'm fairly sure @S11001001 wrote about it but I couldn't find his post, [this](https://typelevel.org/blog/2018/05/09/product-with-serializable.html) is what I found. No need to accept this if you don't think it adds anything here.

stefanobaghino-da (commented):
I'll leave giving the green light to people more experienced than myself on Speedy and scenarios, but LGTM, I left a couple of suggestions.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
cocreature: I blame Scala for having shitty type inference.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: neither really. I use this in `ScenarioRunner` and didn’t want to duplicate it.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ScenarioRunner.scala
cocreature: I know about the pattern but I usually add it only when I actually use it in a way where type inference benefits from it. But I guess there is no harm in adding it prematurely.

aherrmann-da (approved):
Looks good to me, thank you!

remyhaemmerle-da (approved):
Nice.

PR #10075, opened 2021-06-22
Scenario: move Scenario Error Throwable to scenario package

remyhaemmerle-da:
com.daml.lf.speedy.SErrorScenario => com.daml.lf.scenario.Error

We move scenario errors out of speedy as the latter has no need to know about. 

part of #9974

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
@@ -15,8 +15,7 @@ import com.daml.lf.data.Ref.{DottedName, Identifier, ModuleName, PackageId, Qual
 import com.daml.lf.engine.script.ledgerinteraction.{IdeLedgerClient, ScriptTimeMode}
 import com.daml.lf.language.{Ast, LanguageVersion, Util => AstUtil}
 import com.daml.lf.scenario.api.v1.{ScenarioModule => ProtoScenarioModule}
-import com.daml.lf.speedy.{Compiler, ScenarioRunner, SDefinition, SExpr, Speedy}
-import com.daml.lf.speedy.SError._
+import com.daml.lf.speedy.{Compiler, SDefinition, SExpr, Speedy}
 import com.daml.lf.speedy.SExpr.{LfDefRef, SDefinitionRef}
 import com.daml.lf.validation.Validation
 import com.google.protobuf.ByteString
@@ -201,7 +200,7 @@ class Context(val contextId: Context.ContextId, languageVersion: LanguageVersion
     val participants = Participants(Some(ledgerClient), Map.empty, Map.empty)
     val (clientMachine, resultF) = runner.runWithClients(participants, traceLog)
 
-    def handleFailure(e: SError) = {
+    def handleFailure(e: Error) =
       // SError are the errors that should be handled and displayed as
       // failed partial transactions.
       Success(
@@ -216,7 +215,6 @@ class Context(val contextId: Context.ContextId, languageVersion: LanguageVersion
           )
         )
       )
-    }
 
     val dummyDuration: Double = 0
     val dummySteps: Int = 0
@@ -234,10 +232,11 @@ class Context(val contextId: Context.ContextId, languageVersion: LanguageVersion
             )
           )
         )
-      case Failure(e: SError) => handleFailure(e)
+      case Failure(e: Error) => handleFailure(e)
       case Failure(e: ScriptF.FailedCmd) =>
         e.cause match {
-          case e: SError => handleFailure(e)
+          case e: Error => handleFailure(e)
+          case e: speedy.SError.SError => handleFailure(Error.RunnerException(e))
           case _ => Failure(e)
         }
       case Failure(e) => Failure(e)
--- a/daml-lf/scenario-interpreter/BUILD.bazel
+++ b/daml-lf/scenario-interpreter/BUILD.bazel
@@ -13,12 +13,15 @@ load(
     "//rules_daml:daml.bzl",
     "daml_compile",
 )
+load("@scala_version//:index.bzl", "scala_major_version")
 
 da_scala_library(
     name = "scenario-interpreter",
     srcs = glob(["src/main/**/*.scala"]),
     main_class = "com.daml.lf.speedy.Main",
+    scala_deps = ["@maven//:org_typelevel_paiges_core"],
     tags = ["maven_coordinates=com.daml:daml-lf-scenario-interpreter:__VERSION__"],
+    versioned_scala_deps = {"2.12": ["@maven//:org_scalaz_scalaz_core"]},
     visibility = [
         "//visibility:public",
     ],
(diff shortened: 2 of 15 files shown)


compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
```diff
@@ -234,10 +232,11 @@ class Context(val contextId: Context.ContextId, languageVersion: LanguageVersion
             )
           )
         )
-      case Failure(e: SError) => handleFailure(e)
+      case Failure(e: Error) => handleFailure(e)
       case Failure(e: ScriptF.FailedCmd) =>
         e.cause match {
-          case e: SError => handleFailure(e)
+          case e: Error => handleFailure(e)
+          case e: speedy.SError.SError => handleFailure(Error.RunnerException(e))
```
stefanobaghino-da, line 239: Is this needed for backward compatibility?

daml-lf/scenario-interpreter/BUILD.bazel
```diff
@@ -13,11 +13,15 @@ load(
...
 )
+load("@scala_version//:index.bzl", "scala_major_version")
 
 da_scala_library(
     name = "scenario-interpreter",
     srcs = glob(["src/main/**/*.scala"]),
     main_class = "com.daml.lf.speedy.Main",
+    scala_deps =
+        ["@maven//:org_typelevel_paiges_core"] +  #
+        (["@maven//:org_scalaz_scalaz_core"] if scala_major_version == "2.12" else []),
```
stefanobaghino-da, line 24: Why a conditional dependency on Scalaz?

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/Error.scala
```diff
@@ -0,0 +1,78 @@
...
+    with NoStackTrace
+    with Product
+    with Serializable
+
+object Error {
+
+  final case class RunnerException(err: SError) extends Error
+
+  final case class Internal(reason: String) extends Error {
+    override def toString = "CRASH: " + reason
```
stefanobaghino-da, line 28: Is this necessary? Looks like `Pretty` already covers this.

stefanobaghino-da (commented):
LGTM but I'll leave an approval to more experienced people. 🙂 I left a couple of questions but nothing that looks off.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
remyhaemmerle-da: `FailCmd` may still contain some `SError`    -- I am not completley sure where they come from.
Refactoring `FailCmd` is not completly trivial as it may handle any `Throwable` but is definitively not part of this PR.

cocreature: the errors in `FailedCmd` should come from `IdeLedgerClient`.

daml-lf/scenario-interpreter/BUILD.bazel
remyhaemmerle-da: scala 2.12 complains when the dependency is missing
scala 2.13 complains when the dependency is there.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/Error.scala
remyhaemmerle-da: Pretty does not handle Internal Error.

daml-lf/scenario-interpreter/BUILD.bazel
remyhaemmerle-da: scala 2.12 complain about missing dependencies when missing.
scala 2.13 complain about unused dependencies when I add the dependencies

cocreature: can we use `versioned_scala_deps`?

stefanobaghino-da: 👆🏻

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/Error.scala
stefanobaghino-da: Ooops, sorry, I saw [this](https://github.com/digital-asset/daml/pull/10075/files#diff-da8d271bbcb2258273f76c8741a61b50e78d85bc4cb49f371c1d44cbf55fa354R32-R33) and thought it did, my bad.

daml-lf/scenario-interpreter/src/main/scala/com/digitalasset/daml/lf/Pretty.scala
```diff
@@ -0,0 +1,73 @@
...
+private[lf] object Pretty {
+
+  import speedy.Pretty._
+
+  def prettyError(err: Error, optPtx: Option[PartialTransaction] = None): Doc =
+    err match {
+      case Error.RunnerException(serror) =>
+        speedy.Pretty.prettyError(serror, optPtx)
+      case Error.Internal(msg) =>
+        text(s"CRASH: $msg ")
```
stefanobaghino-da, line 21: Looks to me like `Pretty` handles `Internal`. What am I missing?

remyhaemmerle-da: My mistake.
I am not sure dropping `Error.Internal#toString` is a good idea, as the error can be thrown directly and then potentially improperly catch.  I think having a proper to String  method can be useful for debugging, specially because Internal Error indicate a bug somewhere.

stefanobaghino-da: Then you probably also want to extend `getMessage`.

cocreature (commented):
Nice, thank you!

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
remyhaemmerle-da: @cocreature I will double check.

cocreature (approved):
and now I managed to click the right button

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Context.scala
remyhaemmerle-da: @cocreature `//compiler/damlc/tests:script-service` is failing when I remove the line 238.

cocreature: thx will take a look separately, let’s leave it out of this PR.

cocreature: nvm this is definitly expected. As discussed this morning the use of exceptions at least in some of those places is probably somewhat questionable but at least in the current state this makes sense.

PR #10083, opened 2021-06-22
DPP-428 Add missing indices for index initialization

rautenrieth-da:
This PR adds missing indices for the deletion of overspill entries during index initialization.


ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V102__add_initialization_indices.sql
```diff
@@ -0,0 +1,17 @@
...
+
+---------------------------------------------------------------------------------------------------
+-- V102: Add indices to speed up indexer initialization
+--
+-- At startup, the indexer deletes all entries with an offset beyond the stored ledger end.
+-- Such entries can be written when the indexer crashes right before updating the ledger end.
+-- This migration adds missing indices to speed up the deletion of such entries.
+---------------------------------------------------------------------------------------------------
+
+CREATE INDEX configuration_entries_ledger_offset_idx ON configuration_entries USING btree (ledger_offset);
```
rautenrieth-da, line 13: While adding this index, I realized that we don't prune configuration entries when we prune the participant. This seems to be a known shortcoming of the current pruning implementation.

nmarton-da: why is this needed? we have a primary key for the configuration_entries, which is the ledger_offset

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V102__add_initialization_indices.sql
```diff
@@ -0,0 +1,17 @@
...
+CREATE INDEX package_entries_ledger_offset_idx ON package_entries USING btree (ledger_offset);
```
nmarton-da, line 14: why is this needed? we have a primary key for the package_entries, which is the ledger_offset

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V102__add_initialization_indices.sql
```diff
@@ -0,0 +1,17 @@
...
+CREATE INDEX packages_ledger_offset_idx ON packages USING btree (ledger_offset);
+CREATE INDEX parties_ledger_offset_idx ON parties USING btree (ledger_offset);
+CREATE INDEX party_entries_ledger_offset_idx ON party_entries USING btree (ledger_offset);
```
nmarton-da, line 17: why is this needed? we have a primary key for the party_entries, which is the ledger_offset

nmarton-da (commented):
I do not understand why we need to add indexes for those tables which have already one on the ledger_offset.

Also: was it a conscious decision not adding these indexes to H2?

Also: did you create a task to replicate these new indexes for Oracle?

rautenrieth-da:
> I do not understand why we need to add indexes for those tables which have already one on the ledger_offset.

🤦

I saw that no index is created or used when deleting overspill entries from these tables, but I forgot to check the primary key. The fact that no index was used in my test is probably due to the very small number of parties/configurations/packages in the example data set. That only leaves `packages` and `parties` without an index, and of course I need to add it for H2 as well.

dasormeter:
> I do not understand why we need to add indexes for those tables which have already one on the ledger_offset.
> 
> Also: was it a conscious decision not adding these indexes to H2?
> 
> Also: did you create a task to replicate these new indexes for Oracle?

I believe these indexes already exist for oracle because an index is automagically created for any field marked as primary key

mziolekda:
Do we need to propagate this PR to 1.14 (provided the dispute with Marton is resolved)?

rautenrieth-da:
@mziolekda These indices only affect the performance of the indexer startup. They would be relevant for customers with large index databases (e.g., tens of thousands of parties) and strict HA requirements.

@dasormeter The `packages` and `parties` tables do not have the ledger offset as primary key. See [this commit](https://github.com/digital-asset/daml/pull/10083/commits/1c9af4a58ba0c243e2777bb52b1c1993c7e403f2) for what I did for H2, the same should be done for Oracle.

nmarton-da (approved):
LGTM

Also checked @dasormeter 's PR, it already contains the crucial indexes :+1:

PR #10088, opened 2021-06-23
dev-env: Use `nix-shell` for scripts instead of sourcing `dade-common`.

SamirTalwar-DA:
This introduces a _shell.nix_ file which exposes the various tooling, so that shell scripts can use it, as a more "standard" workflow than _dade-common_.

We then use this in _dev-env/bin/ghcide_ and _language-support/hs/bindings/export-package.sh_.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/dev-env/bin/ghcide
+++ b/dev-env/bin/ghcide
@@ -1,4 +1,9 @@
-#!/usr/bin/env bash
-DADE_CURRENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source "$DADE_CURRENT_SCRIPT_DIR/../lib/dade-common"
-bazel build @ghcide-exe//ghcide && bazel-bin/external/ghcide/ghcide-0.7.2.0/_install/bin/ghcide "$@"
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash ../../shell.nix
+
+set -e
+set -u
+set -x
+
+bazel build @ghcide-exe//ghcide
+exec ./bazel-bin/external/ghcide/ghcide-0.7.2.0/_install/bin/ghcide "$@"
--- a/dev-env/bin/ghcide
+++ b/dev-env/bin/ghcide
@@ -1,9 +1,7 @@
 #!/usr/bin/env nix-shell
 #!nix-shell -i bash ../../shell.nix
 
-set -e
-set -u
-set -x
+set -euo pipefail
 
 bazel build @ghcide-exe//ghcide
 exec ./bazel-bin/external/ghcide/ghcide-0.7.2.0/_install/bin/ghcide "$@"
--- a/language-support/hs/bindings/package.yaml
+++ b/language-support/hs/bindings/package.yaml
@@ -7,7 +7,7 @@ description:         Haskell bindings for a DAML Ledger
 license:             Apache-2.0
 author:              The DAML Authors
 maintainer:          nick.chapman@digitalasset.com
-github:              https://github.com/digital-asset/daml
+github:              digital-asset/daml
 
 dependencies:
 - async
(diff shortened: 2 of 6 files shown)


language-support/hs/bindings/package.yaml
```diff
@@ -7,7 +7,7 @@ description:         Haskell bindings for a DAML Ledger
 license:             Apache-2.0
 author:              The DAML Authors
 maintainer:          nick.chapman@digitalasset.com
-github:              https://github.com/digital-asset/daml
+github:              digital-asset/daml
```
SamirTalwar-DA, line 10: Without this change, `stack` kept complaining to me that it was invalid.

dev-env/bin/ghcide
```diff
@@ -1,4 +1,9 @@
-#!/usr/bin/env bash
-DADE_CURRENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source "$DADE_CURRENT_SCRIPT_DIR/../lib/dade-common"
-bazel build @ghcide-exe//ghcide && bazel-bin/external/ghcide/ghcide-0.7.2.0/_install/bin/ghcide "$@"
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash ../../shell.nix
+
+set -e
+set -u
+set -x
```
cocreature, line 6: do we want this here?

dev-env/bin/ghcide
```diff
@@ -1,4 +1,9 @@
-#!/usr/bin/env bash
-DADE_CURRENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source "$DADE_CURRENT_SCRIPT_DIR/../lib/dade-common"
-bazel build @ghcide-exe//ghcide && bazel-bin/external/ghcide/ghcide-0.7.2.0/_install/bin/ghcide "$@"
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash ../../shell.nix
+
+set -e
```
cocreature, line 4: any reason not to go for the standard `set -eou pipefail`?

language-support/hs/bindings/package.yaml
cocreature: i hate hpack

cocreature (approved):
:shipit:

dev-env/bin/ghcide
SamirTalwar-DA: It only prints two lines and I found it helpful. I can remove it if you like.

cocreature: fine to keep it. Just seemed a bit odd to include it in this one place which doesn’t actually do anything particularly interesting while we don’t have it in most other scripts.

dev-env/bin/ghcide
SamirTalwar-DA: I don't know how that became standard but I don't like it.

1. I find it unclear that the `o` is paired with `pipefail`.
2. We don't use pipes in this script.
3. Something something one thing per line.

I am happy to change this to `set -eu` or `set -e -u` if it makes you happier. :slightly_smiling_face:

cocreature: I get your point but in the end, I find it much easier to enforce that every script starts with the same prelude than make a case by case decision which can easily become outdated later if we start using pipes.
So I think this is one of the case where we’re better off using the same thing everywhere than using each person’s personal taste and having a bunch of different headers. I couldn’t care less about the exact formatting we use for the header as long as its consistent so if you do want to push to change it everywhere, I’m all for it.

aherrmann-da (approved):
LGTM

dev-env/bin/ghcide
nickchapman-da: I defer to the opinion of the master :)
https://cuddly-octo-palm-tree.com/posts/2021-01-31-bash-set-pipefail/

SamirTalwar-DA: That works for me. Done.

PR #10108, opened 2021-06-24
Fix Tx.Metadata for normalized transactions

nickchapman-da:
Fix Tx.Metadata for normalized transactions

The primary cause of the bug was the `Tx.Metadata` was referencing node-ids calculated prior to the normalization of the transaction. The bug did *not* manifest if normalization made no change to the node-ids, which was until recently was always the case if the transaction contained no rollback nodes.

With the recent change to `reinterpret` (adding a try-catch handler around the command being reinterpreted), the pre-normalized tx was constructed with transitory node-ids which did not end up matching the final node-ids after normalization... even in the case that the tx contains no rollback nodes!

The fix is to never use the transitory node-ids generated before normalization, specifically those which were in `actionNodeSeeds`, and instead calculate the correct NodeIds in `PartialTransaction.finish`. This is easy, so long as we always collect the actionNodeSeeds in pre-order. Which this PR also ensures.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -496,7 +496,7 @@ class EngineTest
     "reinterpret to the same result" in {
       val Right((tx, txMeta)) = interpretResult
 
-      val Right((rtx, _)) =
+      val Right((rtx, newMeta)) =
         reinterpret(
           engine,
           Set(party),
@@ -507,6 +507,7 @@ class EngineTest
           lookupPackage,
         )
       isReplayedBy(tx, rtx) shouldBe Right(())
+      txMeta.nodeSeeds shouldBe newMeta.nodeSeeds
     }
 
     "be validated" in {
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -2440,6 +2440,85 @@ class EngineTest
       }
     }
 
+    "action node seeds" should {
+      val (exceptionsPkgId, _, allExceptionsPkgs) = loadPackage("daml-lf/tests/Exceptions.dar")
+      val lookupPackage = allExceptionsPkgs.get(_)
+      val kId = Identifier(exceptionsPkgId, "Exceptions:K")
+      val seedId = Identifier(exceptionsPkgId, "Exceptions:NodeSeeds")
+      val let = Time.Timestamp.now()
+      val submissionSeed = hash("rollback")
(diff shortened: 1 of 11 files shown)


daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -507,6 +507,7 @@ class EngineTest
           lookupPackage,
         )
       isReplayedBy(tx, rtx) shouldBe Right(())
+      txMeta.nodeSeeds shouldBe newMeta.nodeSeeds
```
remyhaemmerle-da, line 510: :+1:

remyhaemmerle-da: Maybe we could have one more involved test that actually uses rollback node.

remyhaemmerle-da (approved):
Thanks.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
cocreature: A test for `submit` also would be nice. This really applies to everything as soon as a `try` is involved.

cocreature (approved):
thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -320,6 +322,10 @@ private[lf] case class PartialTransaction(
       sb.toString
     }
 
+  private def normalizedNodeSeeds(): NodeSeeds = {
+    ImmArray(actionNodeSeeds.toImmArray.toList.zipWithIndex.map { case (v, i) => (NodeId(i), v) })
```
remyhaemmerle-da, line 326: You should also avoid using `ImmArray#apply[X](xs: X*)` which is here more for convenience that for efficiency.

The optimal version going in the direction of @cocreature is as follow:
```suggestion
    actionNodeSeeds.toImmArray.toSeq.zipWithIndex.map { case (v, i) => (NodeId(i), v) }.toImmArray
```

remyhaemmerle-da:
I find a bit fragile to use `NodeId` for the node and not for the seed.
We could either: 
- mode the seed directly in the node
- handle everything in preorder traversal (without Node Id) until normalization.

Not for PR though.

cocreature:
> I find a bit fragile to use `NodeId` for the node and not for the seed.
> We could either:
> 
>     * mode the seed directly in the node
> 
>     * handle everything in preorder traversal (without Node Id) until normalization.
> 
> 
> Not for PR though.

yeah I agree it’s a bit messy.

oliverse-da:
Thanks all!

PR #10111, opened 2021-06-24
[Scala client bindings] Fix a "deadlock" in CommandRetryFlow

gerolf-da:
The "deadlock" occurs when a command is sent to the RETRY_PORT, but at
the same time new commands are coming in as well. If the
commandSubmissionFlow is limited (e.g. via the MaxInFlight flow), then
backpressure is exerted on the merge flow and ultimately back to the
retryDecider. This in turn prevents new command results from being
propagated to the PROPAGATE_PORT, and therefore the limited
commandSubmissionFlow ends up backpressuring itself into a "deadlock".

To circumvent this, two new measures are in place:
1. The Merge flow is now a MergePreferred flow, giving precedence to
submission retries.
2. An additional buffer is added to the retry feedback loop, so that the
limited commandSubmissionFlow can keep making progress.

Ideally this buffer would have an additional feature of rerouting new
elements to a completely different port if the buffer is full, but that
would likely require a custom graph stage.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
+++ b/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
@@ -8,12 +8,14 @@ import java.time.temporal.TemporalAmount
 import akka.NotUsed
 import akka.stream.{FlowShape, OverflowStrategy}
 import akka.stream.scaladsl.{Flow, GraphDSL, MergePreferred, Partition}
+import com.codahale.metrics.Counter
 import com.daml.api.util.TimeProvider
 import com.daml.ledger.api.refinements.ApiTypes.Party
 import com.daml.ledger.api.v1.command_submission_service.SubmitRequest
 import com.daml.ledger.api.v1.completion.Completion
 import com.daml.ledger.client.services.commands.CommandClient
 import com.daml.util.Ctx
+import com.daml.util.akkastreams.MaxInFlight
 import com.google.rpc.Code
 import com.google.rpc.status.Status
 import scalaz.syntax.tag._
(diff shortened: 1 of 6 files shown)


language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
```diff
@@ -98,12 +107,18 @@ object CommandRetryFlow {
             Ctx(retryInfo.newRetry, createRetry(retryInfo, failedCompletion), telemetryContext)
         })
 
+        // This buffer, together with the mergePreferred flow, is important to break the cycle around commandSubmission.
+        // Without this setup, mergeRetriesPreferred would backpressure all the way to retryDecider, effectively
+        // stopping the entire flow, i.e. command processing.
+        val buffer =
```
cocreature, line 113: So if I understand you correctly we still deadlock if this buffer is full?

gerolf-da: That's right, but together with the MergePreferred flow this should be dramatically reduced.
Apologies, I should have been explicit about it.

cocreature (approved):
Looks sensible, very nice work tracking this down!

I admit I’m slightly scared by the complexity of the retry flow especially since it only seems to be getting more and more complex but I don’t have a great suggestion for simplifying it either unfortunately.

I recommend waiting for a second review from someone who is a bit more of an akka streams expert than I am.

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
```diff
@@ -98,12 +107,18 @@ object CommandRetryFlow {
...
+          b.add(Flow[Out[RetryInfo[C]]].buffer(retryBufferSize, OverflowStrategy.backpressure))
+
         // format: off
-        merge.out            ~> commandSubmission ~> retryDecider.in
-        merge.in(RETRY_PORT) <~  convertToRetry   <~ retryDecider.out(RETRY_PORT)
+        mergeRetriesPreferred.out ~> commandSubmission ~> retryDecider.in
+        mergeRetriesPreferred.in(RETRY_PORT) <~ convertToRetry <~ buffer <~ retryDecider.out(RETRY_PORT)
```
S11001001, line 118: This issue existed before, but is exacerbated by the buffer:

1. What happens if there are commands in the buffer when a `CommandRetryFlow`'s upstream completes? (I believe this code implies "they are dropped", but could be mistaken.)
2. Is that behavior desirable? Is it acceptable?

S11001001 (approved):

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
gerolf-da: If the upstream completes, then the completion signal is propagated downstream and the retries are lost. Indeed, the buffer exacerbates that loss.

gerolf-da:
Thank you for your comments, @cocreature and @S11001001.
I think both are valid concerns.
Since the deadlock is technically still possible, do you want to track this separately?
Regarding the lost retries on upstream completion: I'd need to think a bit more about how to solve this, but I think this is a tricky one.

cocreature:
> Since the deadlock is technically still possible, do you want to track this separately?

opened #10113 and #10114

gerolf-da:
https://github.com/digital-asset/daml/pull/10111/commits/307b399dd8d48e0f689bff2796a11515b817f32c moves the `MaxInFlight` stage outside the `CommandRetryFlow`, which means there will be no backpressure internally to `CommandRetryFlow` and the deadlock will be prevented.

stefanobaghino-da (approved):
Nothing to add on top of what has already been mentioned.

I don't think this is a viable approach, but purely in the interest of putting the idea out there for consideration: what would be the impact of dropping retries entirely and ask the clients to use an explicit retry policy of their own?

gerolf-da:
@stefanobaghino-da: this is actually a very sensible suggestion. The place that currently uses `LedgerClientBinding#retryingConfirmedCommands` could also use `LedgerClientBinding#commands` and let the application handle the errors. I will check with if this works for the application in question.

gerolf-da:
Closing this stale PR.

PR #10116, opened 2021-06-25
DPP-457 Add UpdateToDbdto tests

rautenrieth-da:
This PR adds unit tests for `UpdateToDbDto`.

These tests can simulate any participant state update (including ones coming from privacy-aware or malicious participants) and check whether these updates are correctly translated into index database DTOs (each DTO corresponding to one row to be inserted).

### DIFF
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
@@ -54,6 +54,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
   import UpdateToDbDtoSpec._
 
   "UpdateToDbDto" should {
+
     "handle ConfigurationChanged" in {
       val update = Update.ConfigurationChanged(
         someRecordTime,
@@ -76,6 +77,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         )
       )
     }
+
     "handle ConfigurationChangeRejected" in {
       val rejectionReason = "Test rejection reason"
       val update = Update.ConfigurationChangeRejected(
@@ -100,6 +102,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         )
       )
     }
+
     "handle PartyAddedToParticipant (local party)" in {
       val displayName = "Test party"
       val update = Update.PartyAddedToParticipant(
@@ -133,6 +136,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle PartyAddedToParticipant (remote party)" in {
       val displayName = "Test party"
       val update = Update.PartyAddedToParticipant(
@@ -166,6 +170,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle PartyAllocationRejected" in {
       val rejectionReason = "Test party rejection reason"
       val update = Update.PartyAllocationRejected(
@@ -191,6 +196,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         )
       )
     }
+
     "handle PublicPackageUpload (two archives)" in {
       val sourceDescription = "Test source description"
       val update = Update.PublicPackageUpload(
@@ -231,6 +237,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle PublicPackageUploadRejected" in {
       val rejectionReason = "Test package rejection reason"
       val update = Update.PublicPackageUploadRejected(
@@ -252,6 +259,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         )
       )
     }
+
     "handle CommandRejected" in {
       val submitterInfo = SubmitterInfo(
         actAs = List(someParty),
@@ -287,6 +295,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (single create node)" in {
       val submitterInfo = someSubmitterInfo
       val transactionMeta = someTransactionMeta
@@ -332,7 +341,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
           create_argument = Some(emptyArray),
           create_signatories = Some(Set("signatory")),
           create_observers = Some(Set("observer")),
-          create_agreement_text = Some(createNode.coinst.agreementText).filter(_.nonEmpty),
+          create_agreement_text = Some(createNode.coinst.agreementText),
           create_key_value = None,
           create_key_hash = None,
           create_argument_compression = compressionAlgorithmId,
@@ -351,6 +360,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (single consuming exercise node)" in {
       val submitterInfo = someSubmitterInfo
       val transactionMeta = someTransactionMeta
@@ -429,6 +439,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (single non-consuming exercise node)" in {
       val submitterInfo = someSubmitterInfo
       val transactionMeta = someTransactionMeta
@@ -507,6 +518,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (nested exercise nodes)" in {
       // Previous transaction
       // └─ #1 Create
@@ -668,6 +680,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (single exercise node with divulgence)" in {
       // Previous transaction
       // └─ #1 Create
@@ -763,6 +776,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (transaction with local divulgence)" in {
       // Transaction
       // ├─ #1 Create
@@ -822,7 +836,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
           create_argument = Some(emptyArray),
           create_signatories = Some(Set("signatory")),
           create_observers = Some(Set("observer")),
-          create_agreement_text = Some(createNode.coinst.agreementText).filter(_.nonEmpty),
+          create_agreement_text = Some(createNode.coinst.agreementText),
           create_key_value = None,
           create_key_hash = None,
           create_argument_compression = compressionAlgorithmId,
@@ -882,6 +896,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (explicit blinding info)" in {
       val submitterInfo = someSubmitterInfo
       val transactionMeta = someTransactionMeta
@@ -977,6 +992,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (rollback node)" in {
       // Transaction
       // └─ #1 Rollback
@@ -1050,6 +1066,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         ),
       )
     }
+
     "handle TransactionAccepted (no submission info)" in {
       // Transaction that is missing a SubmitterInfo
       // This happens if a transaction was submitted through a different participant
@@ -1096,7 +1113,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
           create_argument = Some(emptyArray),
           create_signatories = Some(Set("signatory")),
           create_observers = Some(Set("observer")),
-          create_agreement_text = Some(createNode.coinst.agreementText).filter(_.nonEmpty),
+          create_agreement_text = Some(createNode.coinst.agreementText),
           create_key_value = None,
           create_key_hash = None,
           create_argument_compression = compressionAlgorithmId,
@@ -1105,6 +1122,7 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
         )
       )
     }
+
   }
 }
 
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
@@ -51,89 +51,7 @@ import org.scalatest.wordspec.AnyWordSpec
 import scala.concurrent.{ExecutionContext, Future}
 
 class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
-  // DbDto case classes are not comparable, because the contains Arrays.
-  // The DbDto trait mixes in SomeArrayEquals, but this does not help as some DTOs contain Arrays within Options.
-  // These tests work nevertheless because all array values use the same empty array instance.
-  private[this] val emptyArray = Array.emptyByteArray
-
-  // These tests do not check the correctness of the LF value serialization.
-  // All LF values are serialized into empty arrays in this suite.
-  private[this] val valueSerialization = new LfValueSerialization {
-    override def serialize(
-        contractId: ContractId,
-        contractArgument: Value.VersionedValue[ContractId],
-    ): Array[Byte] = emptyArray
-
-    /** Returns (contract argument, contract key) */
-    override def serialize(eventId: EventId, create: Create): (Array[Byte], Option[Array[Byte]]) =
-      (emptyArray, create.key.map(_ => emptyArray))
-
-    /** Returns (choice argument, exercise result, contract key) */
-    override def serialize(
-        eventId: EventId,
-        exercise: Exercise,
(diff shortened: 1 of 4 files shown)


ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1191 @@
...
+import com.google.protobuf.ByteString
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
+  // DbDto case classes are not comparable, because the contains Arrays.
+  // The DbDto trait mixes in SomeArrayEquals, but this does not help as some DTOs contain Arrays within Options.
+  // These tests work nevertheless because all array values use the same empty array instance.
```
stefanobaghino-da, line 56: Definitely outside the scope of this PR, but: general advice to use `immutable.ArraySeq` if you need to put arrays in `case class`es, using unsafe array wrapping methods the runtime cost is negligible at most and you don't mess with equality, which is one of the defining characteristics of `case class`es.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1191 @@
...
+  )
+  private[this] val someTransactionMeta = TransactionMeta(
+    ledgerEffectiveTime = Time.Timestamp.assertFromLong(2),
+    workflowId = Some(someWorkflowId),
+    submissionTime = Time.Timestamp.assertFromLong(3),
+    submissionSeed = someHash,
+    optUsedPackages = None,
+    optNodeSeeds = None,
+    optByKeyNodes = None,
+  )
```
stefanobaghino-da, line 136: Minor, but if you think it would improve readability, consider putting helpers in a companion object.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1191 @@
...
+        recordTime = someRecordTime,
+        divulgedContracts = List.empty,
+        blindingInfo = None,
+      )
+      val dtos = UpdateToDbDto(someParticipantId, valueSerialization, compressionStrategy)(
+        someOffset
+      )(update).toList
+
+      dtos should contain theSameElementsInOrderAs List(
+        // TODO append-only: Why is there a divulgence event? The divulged contract doesn't exist because it was rolled back.
```
stefanobaghino-da, line 1109: Should we open a ticket to track this?

stefanobaghino-da (commented):
LGTM, just a couple of minor notes. Leaving approval to someone closer to the implementation. 🙂

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1191 @@
...
+import com.daml.platform.store.dao.DeduplicationKeyMaker
+import com.google.protobuf.ByteString
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
+  // DbDto case classes are not comparable, because the contains Arrays.
+  // The DbDto trait mixes in SomeArrayEquals, but this does not help as some DTOs contain Arrays within Options.
```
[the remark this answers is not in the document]
nmarton-da: rant
actually this also means wartremover is a bit shortsighted
also seem to mean like what is ok for the scala-lib, is not ok for us to do :( 
also means like SomeArrayEquals has some limitations
all this make me wish just using plain case classes with whatever we want again :man_facepalming: 
/rant

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
nmarton-da: Thats a nice idea!
...if you want to have your case class comparable that regard, which we don't, seems like referential comparision works all nicely for tests, and no comparision is needed for prod

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1191 @@
...
+          node_index = Some(createNodeId.index),
+          event_id = Some(EventId(update.transactionId, createNodeId).toLedgerString),
+          contract_id = createNode.coid.coid,
+          template_id = Some(createNode.coinst.template.toString),
+          flat_event_witnesses = Set("signatory", "observer"),
+          tree_event_witnesses = Set("signatory", "observer"),
+          create_argument = Some(emptyArray),
+          create_signatories = Some(Set("signatory")),
+          create_observers = Some(Set("observer")),
+          create_agreement_text = Some(createNode.coinst.agreementText).filter(_.nonEmpty),
```
nmarton-da, line 1099: I think these kind of replication of the actual code under test is a bit on the non-practical side. would you mind changing this to the actual Option[String] instead?
(applies for similar constructs)

nmarton-da (commented):
Looks very thorough!
I am concerned about that TODO which might signal a bug somewhere
How about the TODO
```
        // TODO append-only:
        //   Covering this functionality with unit test is important, since at the time of writing kvutils ledgers purge RollBack nodes already on WriteService, so conformance testing is impossible
        //   Unit tests also need to cover the full semantic contract regarding fetch and lookup node removal as well
        //   Investigate possibility to encapsulate this logic in a common place
```

Did you consider encapsulating the logic there (...and injecting here, which would make these tests easier, focusing on node -> eent mapping, and we could have test for the injected part in an other place - hopefully LF?)

libs-scala/scala-utils/src/main/scala/com/daml/scalautil/SomeArrayEquals.scala
```diff
@@ -51,7 +51,7 @@ trait SomeArrayEquals extends Product with Serializable {
 }
 
 object SomeArrayEquals {
```
remyhaemmerle-da, line 53: I really not sure this should be move outside speedy.
This deals only with Array at the first level, if one of the argument of your `Product` is of type `List[Array[_]]` or `Option[Array[_]]`, this latter will be compare with standard equality.

remyhaemmerle-da: for instance the following test will succeed :

```
"equals" should {
    "distinguish nested Arrays" in {
      case class A(x: Option[Array[Int]])

      val a = A(Some(Array(1)))
      val b = A(Some(Array(1)))

      a shouldBe a
      a shouldNot be(b)
    }
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/DbDto.scala
```diff
@@ -5,10 +5,10 @@ package com.daml.platform.store.backend
 
 import java.time.Instant
 
-import com.daml.scalautil.NeverEqualsOverride
+import com.daml.scalautil.SomeArrayEquals
 
 sealed trait DbDto
-    extends NeverEqualsOverride
+    extends SomeArrayEquals
```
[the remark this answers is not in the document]
remyhaemmerle-da: Furthermore the field of type `Option(Array[Bytes])` will not be handle as you expect.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -0,0 +1,1195 @@
...
+import com.daml.platform.store.dao.DeduplicationKeyMaker
+import com.google.protobuf.ByteString
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
+  import UpdateToDbDtoSpec._
+
```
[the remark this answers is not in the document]
remyhaemmerle-da: to be cleaner the comparison should defined explicitly for all subtypes of DbDto.

remyhaemmerle-da (commented):
I do not think that `SomeArrayEqual` should be moved and used outside of speedy. 
It does not handle nested argument as expected.

remyhaemmerle-da (commented):
(deleted)

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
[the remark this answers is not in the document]
remyhaemmerle-da: I do not ask to add the following test, I just try to underline the problem of `SomeArrayEquals`

```suggestion

  "EventDivulgence equality" should {
    "be symmetric" in {

      val x = DbDto.EventDivulgence(
        event_offset = None,
        command_id = None,
        workflow_id = None,
        application_id = None,
        submitters = None,
        contract_id = "cid",
        template_id = None,
        tree_event_witnesses = Set.empty,
        create_argument = Some(Array(1)),
        create_argument_compression = None,
        event_sequential_id = 0,
      )

      (x: DbDto) shouldEqual x.copy(create_argument = Some(Array(1)))

    }

  }

```

remyhaemmerle-da:
I would like definitively the opinion of @S11001001  on that.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
rautenrieth-da: This particular issue doesn't come from the transaction traversal logic. It's the `BlindingInfo` class that produces a divulgence entry. There is even a [unit test for BlindingInfo](https://github.com/digital-asset/daml/blob/main/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/BlindingSpec.scala#L195) that handles divulgence under rollback nodes.

@moritzkiefer-da @remyhaemmerle-da Is it intended that `BlindingInfo` produces divulgence entries that reference create events rolled back in the same transaction? Is there any point in storing such divulgence events (maybe for a future where divulgence events are directly observable through the ledger API)?

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
remyhaemmerle-da: I would rather use `com.daml.lf.data.Bytes`.
If necessary I am fine to move it outside `com.daml.lf.data` and put it in `scala-utils`.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
cocreature: `BlindingInfo` included transient contracts already so contracts that are created and archived in the same transaction. Given that, it seemed natural to also include rolled-back creates. If you want to filter out transient and rolled-back creates you can use `inactiveContracts` https://github.com/digital-asset/daml/blob/729afa8c7b68580b6c4aa784e0f2428960eb0635/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala#L406

rautenrieth-da:
Thanks for the review and the suggestions, @remyhaemmerle-da! For reference for other reviewers:
- The `DbDto` objects contain various byte arrays (sometimes wrapped in `Option`s) that hold serialized Daml-Lf values
- The `DbDto` objects are not compared at run time
- The `DbDto` objects should have as little overhead as possible. In particular, there should be no overhead when sending their payload to JDBC.
- In the tests, we do need to compare `DbDto` objects. However, we know that all byte arrays will be empty, as the tests use a dummy value serialization.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
rautenrieth-da: A divulged transient contract should be visible over the transaction tree stream, so we can't filter out divulgence based on `inactiveContracts`. Divulging a rolled back contract on the other hand is something that "never happened" in my mental model.

If the semantics are not clear, I would opt for storing the divulgence event. Divulging rolled back contracts should be something rare, and we can always delete these events later if we don't need them.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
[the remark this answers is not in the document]
nmarton-da: such a nice idea defining the comparator for the test only in that fashion!

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
cocreature: > A divulged transient contract
I think you’re mixing up divulgence and disclosure here. Disclosure is what lets us see nodes on the ledger API, e.g., the create and exercise for a transient contract. At the moment we do not need this for rollback nodes but we will need it once they are exposed and ignoring the ledger API, you need it in the sense that if you look at privacy between participants those nodes will be shipped around (and this matters in Canton).

Divulgence (from an indexer pov) is what lets us fetch a contract in a later transaction (the fetch node is never exposed on the ledger API). You don’t need that for transient contracts and you don’t need it for rolled back contracts.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
S11001001: > general advice to use `immutable.ArraySeq` if you need to put arrays in `case class`es

In our codebase, use `ImmArray` or `ImmArraySeq` instead (depending on whether you want stdlib compatibility).

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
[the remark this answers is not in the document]
S11001001: >  actually this also means wartremover is a bit shortsighted

I can't blame wartremover here at all; we just aren't doing what really must be done to solve this, which is to avoid `Object#equals` and its proxy `==` altogether. (Note that we do not enable [the `Equals` wart](http://www.wartremover.org/doc/warts.html#equals).)

`==` is not type-safe and there is no way to make it type-safe by subtraction; if you want a safe equals, but not to define full typeclasses as Scalaz, Cats, and Spire do, the only way is [still by addition](https://github.com/japgolly/univeq#univeq).

(I come not to suppress your rant but to amplify it to its natural conclusion 😄 )

>  also means like SomeArrayEquals has some limitations

Absolutely. As a hack to make particular speedy case classes comparable without filling out a bunch of buggy boilerplate, it's not flexible beyond what was apparent in those classes.

S11001001:
> I do not think that `SomeArrayEqual` should be moved and used outside of speedy.

I don't think it _must_ be moved outside of speedy; I certainly never intended it for anything but speedy. However, if someone wants to use it, I have no problem with moving it outside.

> It does not handle nested argument as expected.

It is flawed in such-and-such ways, but `==` is flawed in these ways, and we use that, so I don't share this objection. If `==` is useful, then `SomeArrayEquals` is useful. SAE patches one hole in `==` without doing the whole repair job; sometimes, that's good enough. If we want to actually be principled about comparison, we must avoid both.

nmarton-da (approved):

rautenrieth-da:
@remyhaemmerle-da I have removed the usage of SomeArrayEquals and have moved it back to its original place. Instead, the code now uses a custom equality operator as you suggested.

remyhaemmerle-da (commented):
> @remyhaemmerle-da I have removed the usage of SomeArrayEquals and have moved it back to its original place. Instead, the code now uses a custom equality operator as you suggested.

Thanks. 
I do not approve because I am not confidence to asses the tests, however the rest look pretty good to me.
I believe the approval of @nmarton-da is sufficient to go ahead and merge.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
rautenrieth-da: You can still have a single transaction where a contract is created, then divulged through a fetch/exercise node, then archived, no? _Right now_ such a divulgence event has no use in the index database (and maybe it never will), but I would not change whether such events are stored as part of this PR. I have opened a separate issue to discuss this.

cocreature: As far as I know https://github.com/digital-asset/daml/blob/8747b3d7dcb172705353b59ad54648ca3875c1c0/ledger/participant-integration-api/src/main/scala/platform/store/dao/events/TransactionIndexing.scala#L221 filters out such divulence events for both transient & rolled back nodes but my understanding of the indexer is very superficial.

rautenrieth-da: That's the old indexer, the new one takes divulgences directly from the BlindingInfo: https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDbDto.scala#L241

PR #10118, opened 2021-06-25
Refactor error reporting in Daml Repl

### LINKED ISSUE #10098: Error reporting in Daml Repl
Currently our error reporting in Daml Repl is somewhat questionable and it silently swallows some errors.

The problem is that we report errors by printing them out in the Scala server and then fail the request.
On the client side we ignore failures because we pretend we already logged them on the server side.

This goes wrong as soon as the server side doesn’t log something or it isn’t even a server failure. E.g., a client-side timeout will clearly not be logged on the server side.

The reliable solution here is probably to stop doing what we do and instead send the errors to the client and print them there. Then there is never a case where we ignore errors because they might have been handled elsewhere.

cocreature:
fixes #10098

As mentioned in that issue, the current behavior is a mess which
silently drops errors in certain cases. This PR switches things around
so that errors are reported to the client and we print them there
making sure that no error should ever get lost.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
@@ -72,7 +72,9 @@ data Error
     | ExpectedExpression String
     | NotImportedModules [ModuleName]
     | TypeError -- ^ The actual error will be in the diagnostics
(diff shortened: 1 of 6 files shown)


aherrmann-da (approved):
Nice cleanup, thank you!

PR #10120, opened 2021-06-25
Stream (flat) transaction (tree) pages concurrently with a fixed page size

gerolf-da:
This avoids running expensive queries over a large range of events.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsRange.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsRange.scala
@@ -27,7 +27,7 @@ private[events] object EventsRange {
 
     /** Converts Offset range to Event Sequential ID range.
       *
-      * @param range offset range
+      * @param range      offset range
       * @param connection SQL connection
       * @return Event Sequential ID range
       */
@@ -45,7 +45,7 @@ private[events] object EventsRange {
     /** Converts ledger end offset to Event Sequential ID.
       *
       * @param endInclusive ledger end offset
-      * @param connection SQL connection
+      * @param connection   SQL connection
       * @return Event Sequential ID range.
       */
     def readEventSeqIdRange(endInclusive: Offset)(
@@ -69,18 +69,22 @@ private[events] object EventsRange {
       range: EventsRange[Long],
       pageSize: Int,
   ): Connection => Vector[A] = connection => {
-    val minPageSize = 10 min pageSize max (pageSize / 10)
-    val guessedPageEnd = range.endInclusive min (range.startExclusive + pageSize)
-    val arithPage =
-      read(range copy (endInclusive = guessedPageEnd), None, Some(pageSize))(connection)
-    val found = arithPage.size
-    if (guessedPageEnd == range.endInclusive || found >= minPageSize)
-      arithPage
-    else
-      arithPage ++ read(
-        range copy (startExclusive = guessedPageEnd),
-        Some(minPageSize - found),
-        Some(minPageSize - found),
-      )(connection)
+
+    def loop(newBegin: Long, result: Vector[A]): Vector[A] = {
+      val guessedPageEnd = range.endInclusive min (newBegin + pageSize)
+      val arithPage =
+        read(
+          EventsRange(startExclusive = newBegin, endInclusive = guessedPageEnd),
+          None,
+          Some(pageSize),
+        )(connection)
+      val newResult = result ++ arithPage
+      if (guessedPageEnd == range.endInclusive)
+        newResult
+      else
+        loop(newBegin = guessedPageEnd, newResult)
+    }
+
+    loop(range.startExclusive, Vector.empty)
   }
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
@@ -99,30 +99,31 @@ private[appendonlydao] final class TransactionsReader(
 
     val requestedRangeF = getEventSeqIdRange(startExclusive, endInclusive)
 
-    val query = (range: EventsRange[(Offset, Long)]) => {
-      implicit connection: Connection =>
-        logger.debug(s"getFlatTransactions query($range)")
-        queryNonPruned.executeSqlOrThrow(
-          getTransactions(
-            EventsRange(range.startExclusive._2, range.endInclusive._2),
-            filter,
-            pageSize,
-          )(connection),
-          range.startExclusive._1,
-          pruned =>
-            s"Transactions request from ${range.startExclusive._1.toHexString} to ${range.endInclusive._1.toHexString} precedes pruned offset ${pruned.toHexString}",
-        )
-    }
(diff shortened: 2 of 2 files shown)


gerolf-da:
This was the result of a few ideas and experiments on improving the [flat] transaction [tree] streams performance. This hasn't gone through the transaction service performance tool yet.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsRange.scala
```diff
@@ -69,18 +69,22 @@ private[events] object EventsRange {
...
-    if (guessedPageEnd == range.endInclusive || found >= minPageSize)
-      arithPage
-    else
-      arithPage ++ read(
-        range copy (startExclusive = guessedPageEnd),
-        Some(minPageSize - found),
-        Some(minPageSize - found),
-      )(connection)
+
+    def loop(newBegin: Long, result: Vector[A]): Vector[A] = {
```
tudor-da, line 73: `@tailrec`

tudor-da, line 73: Can you add the `@tailrec` annotation?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -99,30 +99,31 @@ private[appendonlydao] final class TransactionsReader(
...
-            EventsRange(range.startExclusive._2, range.endInclusive._2),
-            filter,
-            pageSize,
-          )(connection),
-          range.startExclusive._1,
-          pruned =>
-            s"Transactions request from ${range.startExclusive._1.toHexString} to ${range.endInclusive._1.toHexString} precedes pruned offset ${pruned.toHexString}",
-        )
-    }
+    val query = (fullRangeForPruningCheck: EventsRange[Offset]) =>
```
tudor-da, line 102: In the context where this `query` is passed, the fullRange is static, isn't it? Why do you need to have it as an input to this lambda?

tudor-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -501,19 +491,53 @@ private[appendonlydao] final class TransactionsReader(
...
+  )(implicit loggingContext: LoggingContext): Source[EventsTable.Entry[E], NotUsed] =
+    Source
+      .fromIterator(() => Iterator.iterate(fullRange.startExclusive._2)(_ + pageSize))
+      .sliding(2)
+      .map { case Seq(startExclusive, endInclusive) =>
+        EventsRange(startExclusive, endInclusive.min(fullRange.endInclusive._2))
       }
+      .takeWhile(_.endInclusive < fullRange.endInclusive._2, inclusive = true)
+      .mapAsync(5) { eventsRange =>
+        fetchEventsInRange(
```
nmarton-da, line 513: this flavor of paginated streams does something completely different as the semantics of these ranges go.
- this: a sequential partition of the complete range
- that: a sequential shrinking of the partition from the from part, based on the resultsets
(is that right? do I miss something?)
Yet the actual query implementation (fetchEventsInRange) seem to serve them both. Is that right? Is that intended?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -501,19 +491,53 @@ private[appendonlydao] final class TransactionsReader(
...
+      fullRange: EventsRange[(Offset, Long)],
+  )(implicit loggingContext: LoggingContext): Source[EventsTable.Entry[E], NotUsed] =
+    Source
+      .fromIterator(() => Iterator.iterate(fullRange.startExclusive._2)(_ + pageSize))
+      .sliding(2)
+      .map { case Seq(startExclusive, endInclusive) =>
+        EventsRange(startExclusive, endInclusive.min(fullRange.endInclusive._2))
       }
+      .takeWhile(_.endInclusive < fullRange.endInclusive._2, inclusive = true)
+      .mapAsync(5) { eventsRange =>
```
nmarton-da, line 512: this can lead to a multiplicative explosion of
A) Usage of  available DB connections
B) Usage of CPU intensive tasks (as we convert the result)
Is this intended? I am just wondering about the consequences...

This seems to be a serious number as far change in load characteristics go: maybe making this configurable? (if this is a PoC please disregard)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsRange.scala
```diff
@@ -69,18 +71,23 @@ private[events] object EventsRange {
...
-    val found = arithPage.size
-    if (guessedPageEnd == range.endInclusive || found >= minPageSize)
-      arithPage
-    else
-      arithPage ++ read(
-        range copy (startExclusive = guessedPageEnd),
-        Some(minPageSize - found),
-        Some(minPageSize - found),
-      )(connection)
+
```
nmarton-da, line 74: this changes completely how we interpret semantically the coming in range.
previously: try to grab a reasonable bite from the range, with first a limitless but fetchhinted range query from start to start+pageSize, to see whether we have enough results in that range, then if not do an additional query which is limited, to bite a bigger one with a limitted query, so that alltogether we have pageSize items, or the range is exhausted.
Note: this algorithm has no intention to exhaust the range.
now: 
do fetch hinted limitless queries, until we exhaust the range

Is this intentional? Isn't that dangerous to fetch here everything?

I did not understand the relevance of the previous approach regarding these various strategies regarding the specific combinations of limits/fetch hints, so it is probably a stupid question, but why is this approach with exclusively fetchinted limitless queries better?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsRange.scala
```diff
@@ -69,18 +71,23 @@ private[events] object EventsRange {
...
+    @tailrec
+    def loop(newBegin: Long, result: Vector[A]): Vector[A] = {
+      val guessedPageEnd = range.endInclusive min (newBegin + pageSize)
+      val arithPage =
+        read(
+          EventsRange(startExclusive = newBegin, endInclusive = guessedPageEnd),
+          None,
```
nmarton-da, line 81: We does not seem to use limit anymore, just fetch hints. Was this intentional?
If yes some code can be flushed this-regard.

nmarton-da (commented):
Not entirely sure about the purpose of this PR, but it changes a lot regards to streamed range query logic, so I tried to be more exhaustive in my comments (and risking to state the obvious some places...sorry if that was the case)

gerolf-da:
After some discussion with @nmarton-da, @tudor-da, and @mziolekda we came to the conclusion that we rather want to use the same parallelism mechanism for queries across the various streams (contract state, flat tx, tx trees, completions, ..) than introducing different mechanisms with different characteristics and potential for bugs.

PR #10123, opened 2021-06-25
ensure that signatories and observers are disjoint in ledger-api CreatedEvents

S11001001:
When a party occurs both in signatories and observers, we indirectly represent this in such a way that the party won't appear in observers when reading that contract over the ledger API. Now we document and test this fact in conformance tests.

This is entirely separate from the behavior of the `observer` method in `HasObserver`, which exactly reflects the `observer` declaration in the `template`. The `observer` method serves as input; by contrast, `CreatedEvent`'s fields serve as actual ledger state, to the extent they are disclosed.

### Pull Request Checklist

trigger the build.


stefanobaghino-da (approved):

ledger-api/grpc-definitions/com/daml/ledger/api/v1/event.proto
```diff
@@ -71,6 +71,7 @@ message CreatedEvent {
   repeated string signatories = 8;
 
   // The observers for this contract as specified explicitly by the template or implicitly as choice controllers.
+  // No member of signatories should be present in this list.
```
cocreature, line 74:
```suggestion
  // This field never contains parties that are signatories.
```
should seems a bit weak if this is a guarantee we provide to API users.

cocreature (approved):
Thanks!

PR #10124, opened 2021-06-25
avoid row_number/partition for single-party Oracle contract queries

S11001001:
#9895 reintroduced the `row_number` `over` `partition` to eliminate duplicates when querying the stakeholder-joined Oracle `contract` table.  However, as #10123 establishes, these duplicates cannot happen if we are querying for only one party.

Therefore, we special-case the single-party query case, for which we skip the `partition` + outer-query duplicate elimination steps. This skips these parts of the query plan:

```
|*  1 |  VIEW			       |		       |  1005 |  9976K|       |  2466	 (1)| 00:00:01 |
|*  2 |   WINDOW SORT PUSHED RANK      |		       |  1005 |     9M|  8048K|  2466	 (1)| 00:00:01 |

   1 - filter("ROWNUMBER"=1)
   2 - filter(ROW_NUMBER() OVER ( PARTITION BY "C"."CONTRACT_ID" ORDER BY  NULL )<=1)
```

### Pull Request Checklist

trigger the build.


cocreature (commented):
Are you able to get out any improvements out of this? In my tests, if anything it seems slightly slower (for the single-party case) measured via
```
bazel run //ledger-service/http-json:contractdao-bench -- -f 1 -i 5 -wi 1 -bm avgt "Query"
```
The slowness is probably just noise but adding complexity (by having two codepaths instead of one) even if just a small amount while not gaining anything still seems somewhat questionable.

S11001001:
@cocreature No. My hope is that the improved query plan will come into play with some other yet-to-be-discovered query improvement; it makes no sense that it would be free. Perhaps that is too optimistic.

cocreature:
> @cocreature No. My hope is that the improved query plan will come into play with some other yet-to-be-discovered query improvement; it makes no sense that it would be free. Perhaps that is too optimistic.

Doesn’t seem to do any harm either so I’ll approve it. Your choice if you want to merge it or not.

I think the reason why this might not be this expensive that this basically just boils down to "discard everything after the first row for each contract". That seems like something that you can implement quite efficiently but I have no idea how Oracle actually does it.

cocreature (approved):

PR #10134, opened 2021-06-28
contextualized-logging: Introduce different logging types for more structure. [KVL-996]

SamirTalwar-DA:
This introduces a `LoggingValue` enumeration that can serialize different data types in a more JSON-friendly way.

For example, a submission log line now looks like this:

```
15:16:38.856 [in-memory-ledger-akka.actor.default-dispatcher-8] INFO  c.d.p.a.s.ApiSubmissionService - Submitting transaction , context: {readAs: [], submittedAt: "2021-06-28T13:16:38.852Z", actAs: ["SemanticDoubleSpendShared-alpha-11e2a521f23e-party-0"], commandId: "SemanticDoubleSpendShared-alpha-11e2a521f23e-command-0", applicationId: "SemanticDoubleSpendShared", deduplicateUntil: "2021-06-29T13:16:38.852Z"}
```

Note that strings are correctly quoted, including inside JSON arrays. Numbers will remain unquoted. In addition, `None` will now serialize to `null`, not an empty string.

I have turned off quoting for the field names for now, to keep the output closer to the previous format, but it can easily be turned back on.

I think the implicit conversions make the code that uses the `withEnrichedLoggingContext` function much more pleasant, though of course you might disagree.

### Changelog

- The log output of Daml components has changed so that the structured part is closer to JSON. This allows us to distinguish and parse numbers and lists. If you are parsing this log output, you may need to change your parser.

  The log output has changed from:

      context: {a=b, x=1, foo=bar, parties=[alice, bob]}

  to:

      context: {a: "b", x: 1, foo: "bar", parties: ["alice", "bob"]}

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
@@ -16,7 +16,7 @@ import com.daml.lf.data.Ref.PackageId
 import com.daml.lf.data.Time.Timestamp
 import com.daml.lf.engine.{Engine, VisibleByKey}
 import com.daml.lf.language.Ast
-import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries, LoggingValue}
+import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries}
 import com.daml.metrics.Metrics
 import com.google.protobuf.ByteString
 
@@ -30,12 +30,6 @@ private[committer] object PackageCommitter {
   )
 
   type Step = CommitStep[Result]
-
-  // Orphan, but we can't do much about that.
-  // Adding this to Daml-LF would probably be overkill.
-  private implicit val `DamlLf.Archive to LoggingValue`
-      : LoggingValue.ToLoggingValue[DamlLf.Archive] =
-    value => LoggingValue.OfString(value.getHash)
 }
 
 final private[kvutils] class PackageCommitter(
@@ -53,7 +47,7 @@ final private[kvutils] class PackageCommitter(
 
   override protected def extraLoggingContext(result: Result): LoggingEntries =
     LoggingEntries(
-      "packages" -> result.uploadEntry.getArchivesList.asScala
+      "packages" -> result.uploadEntry.getArchivesList.asScala.view.map(_.getHash)
     )
 
   /** The initial internal state passed to first step. */
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
@@ -16,8 +16,7 @@ import com.daml.lf.data.Ref.PackageId
 import com.daml.lf.data.Time.Timestamp
 import com.daml.lf.engine.{Engine, VisibleByKey}
 import com.daml.lf.language.Ast
-import com.daml.logging.LoggingValue.ToLoggingValue
-import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries}
+import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries, LoggingValue}
 import com.daml.metrics.Metrics
 import com.google.protobuf.ByteString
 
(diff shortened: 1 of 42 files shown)


SamirTalwar-DA:
@gerolf-da, @mziolekda: Can you foresee any places where this would be seen as a breaking change?

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValue.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @inline
+  implicit def from[T](value: T)(implicit toLoggingValue: ToLoggingValue[T]): LoggingValue =
+    toLoggingValue(value)
+
+  private final class ToStringToLoggingValue[T] extends ToLoggingValue[T] {
+    override def apply(value: T): LoggingValue =
+      OfString(value.toString)
+  }
```
S11001001, line 34:
```suggestion
  private val ToStringToLoggingValue: ToLoggingValue[Any] =
    value => OfString(value.toString)
```

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValue.scala
```diff
@@ -0,0 +1,63 @@
...
+object LoggingValue {
+  final object Empty extends LoggingValue
+
+  final case class OfString(value: String) extends LoggingValue
+
+  final case class OfInt(value: Int) extends LoggingValue
+
+  final case class OfLong(value: Long) extends LoggingValue
+
+  final case class OfIterable(sequence: Iterable[LoggingValue]) extends LoggingValue
```
S11001001, line 21: You _could_ avoid the inner boxing layer on the elements by adding an existential here. It's always the same for every element, anyway. Then you'd add some GADT stuff to `LoggingValueSerializer`...totally up to you whether you want to try to get that working, it's a not-insignificant design change.

S11001001 (approved):

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValue.scala
SamirTalwar-DA: I _think_ I understand what you mean here. I was trying to avoid type parameters at all for `LoggingValue` (hence `ToLoggingValue`'s existence), mostly because I think it will pollute the API. Can you see a way to do this without making the API more confusing?

S11001001: Type parameters occur only within the data constructors; I do not consider `OfIterable` to be a part of the API, so to speak.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValue.scala
SamirTalwar-DA: Done.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValue.scala
SamirTalwar-DA: Oh, I see, that might work. I'll look into it but in a separate change; as you said, it's not insignificant and I'm not so knowledgeable about existentials.

If you feel like doing it at some point today, feel free. Otherwise I'll give it a crack once this is in.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
```diff
@@ -30,6 +30,12 @@ private[committer] object PackageCommitter {
   )
 
   type Step = CommitStep[Result]
+
+  // Orphan, but we can't do much about that.
+  // Adding this to Daml-LF would probably be overkill.
+  private implicit val `DamlLf.Archive to LoggingValue`
+      : LoggingValue.ToLoggingValue[DamlLf.Archive] =
+    value => LoggingValue.OfString(value.getHash)
```
stefanobaghino-da, line 38: Not a big problem but I would probably make it explicit on the call site that we're logging the hash.

stefanobaghino-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/PackageCommitter.scala
SamirTalwar-DA: Deleted this in favor of just `.map(_.getHash)` in the log instruction.

Issue #9947, opened 2021-06-10
Warn about use of divulged contracts in IDE (Scripts), Sandbox and Ledger API Server


bame-da:
As part of our rollout plan for #9283, we should start emitting warnings in IDE, Sandbox, and ideally all Ledger API Servers if a submission makes use of divulged contracts that could be pruned away.

The warning should be informative enough to identify where in the transaction or even in Daml the fetch or exercise occurred. Something like

"Warning: Use of divulged contract incompatible with Participant Pruning. 
The submission with CommandId = Foo with readAs Party/ies [Alice, Bob] makes use of contracts only known to those parties by divulgence. Contracts affected:

{
  template_id: {TEMPLATE_ID}
  contract_id: {CONTRACT_ID}
  stakeholders: [Carol, Doris]
  location: {PATH_TO_NODE}
},
{
  template_id: {TEMPLATE_ID}
  contract_id: {CONTRACT_ID}
  stakeholders: [Carol, Doris]
  location: {PATH_TO_NODE}
}
"

Since we are likely to land the change to pruning in 1.16, I would really like to get this into 1.15.

cocreature:
For the runtime warnings during interpretation, I think it makes sense to do those directly in the engine. Then we solve this for everything at once instead of having to worry about which ledgers warn and which don’t.

There is one issue to make this work: The engine does not currently have access to the `readAs` parties which we need here. This is less of a practical problem to fix (I already hacked up a prototype for another issue at some point) but it does somewhat break our current assumption that visibility checks are handled outside the engine.

But it isn’t obvious to me that we really gain anything by having that separation and this seems like a good argument for breaking it.

mziolekda:
I do not have any reservations for moving this check to the engine.

I am worried about something else. 
1. The format we want to use for this is very verbose. We should be able to turn it off, else it could flood the log-files of some users. 
2. Will we be able to distinguish the contracts that have been divulged vs those that have been disclosed and print this message only for the former? Again some users are relying on disclosed contracts.

cocreature:
I don’t think we want a different error for divulgence vs disclosure. Unless I misunderstood @bame-da both are deprecated.

cocreature:
For verbosity, I think we should at least have some way to turn off the path to the warning. The rest isn’t all that verbose.

bame-da:
[...]

PR #10136, opened 2021-06-28
Move visibility checks into speedy.

sofiafaro-da:
This is part of https://github.com/digital-asset/daml/issues/9947


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -76,34 +76,25 @@ object SResult {
...
-      committers: Set[Party],
-      // Callback.
-      // returns the next expression to evaluate.
-      // In case of failure the call back does not throw but set machine.ctrl to an SErrorDamlException
-      callback: SVisibleByKey => Unit,
-  ) extends SResult
-
-  sealed abstract class SVisibleByKey
-  object SVisibleByKey {
+  sealed abstract class SVisibility
```
cocreature, line 79: Maybe rename this to `SVisibleToStakeholders` or `SStakeholderAreReaders` (I’m sure you can come up with a less shitty name).

For contract ids it does not describe visibility properly and while we deprecate divulgence it’s still around for a while.

cocreature (approved):
Nice work, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
sofiafaro-da: Done, thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -87,6 +87,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
     */
   def submit(
       submitters: Set[Party],
+      readAs: Set[Party],
```
miklos-da, line 90: Can you extend the code documentation above, please?

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractDiscriminatorFreshnessCheckSpec.scala
```diff
@@ -135,6 +135,7 @@ class ContractDiscriminatorFreshnessCheckSpec
     engine
       .submit(
         submitters = Set(alice),
+        readAs = Set.empty,
```
miklos-da, line 138: What does an empty set mean for `readAs`?

sofiafaro-da: It means there are no additional readers (besides the submitters).

cocreature: same semantics as on the ledger API

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
sofiafaro-da: Good idea, thanks!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractDiscriminatorFreshnessCheckSpec.scala
miklos-da: Can you point me to the relevant documentation, please?

cocreature: https://github.com/digital-asset/daml/blob/ffc88d52b438dce5703b1b20d7d82f6f3d63337b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto#L97

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
miklos-da: Thanks extending the code documentation!
If these are really additional readers (i.e., besides the submitters), I'd suggest naming the parameter accordingly. Though, I might be missing some context regarding why we chose `readAs` originally.

miklos-da (approved):

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
cocreature: I’d prefer using the same name everywhere. Using different names for the same thing across our codebase is just confusing. As for why we have those names, I believe we tried to match the names we used for authorization when we added multitparty submissions.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
```diff
@@ -76,34 +76,25 @@ object SResult {
...
-      // Callback.
-      // returns the next expression to evaluate.
-      // In case of failure the call back does not throw but set machine.ctrl to an SErrorDamlException
-      callback: SVisibleByKey => Unit,
-  ) extends SResult
-
-  sealed abstract class SVisibleByKey
-  object SVisibleByKey {
+  sealed abstract class SVisibleToStakeholders
+  object SVisibleToStakeholders {
```
remyhaemmerle-da, line 80: A bit weird  to have the same name for the base type and the case object.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -120,7 +122,11 @@ private[lf] object Speedy {
       var globalDiscriminators: Set[crypto.Hash],
       var cachedContracts: Map[V.ContractId, CachedContract],
   ) extends LedgerMode {
-
+    private[lf] val visibleToStakeholders: Set[Party] => SVisibleToStakeholders =
+      if (validating) { _ => SVisibleToStakeholders.Visible }
```
remyhaemmerle-da, line 126: Naive question, why do we skip the check on validation ?

remyhaemmerle-da (approved):
LGTM

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: Visibility checks are a submission-only concept. They exist primarily to protect parties on the same participant from each other.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SResult.scala
sofiafaro-da: It's not a case object.

I didn't change anything here though, only renamed the existing type.

remyhaemmerle-da: My mistake.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
rautenrieth-da: The `readAs` and `actAs` terms were first used when we added ledger API authorization, for parties the bearer of the auth token is allowed to use for read-only (e.g., transaction service) and state modifying (e.g., command service) ledger API queries, respectively. At this time, the concept of multi-party submissions did not exist and `readAs` had nothing to do with contract visibility during command interpretation.

The terms then made it into the ledger API and the engine when we added multi-party submissions, where `readAs` means "parties that can be used to look up contracts/keys, in addition to the submitters". Using a `readAs` party for multi-party submissions requires a corresponding `readAs` claim in the auth token.

Note also the following details:
- The `readAs`/ `actAs` parties from auth tokens are a superset of the parties used in ledger API commands (you don't have to use all parties you are allowed to)
- The `readAs`/ `actAs` parties from ledger API commands are a superset of the minimal sets of parties required to successfully submit the command (you can submit a command as Alice+Bob, even though Alice alone would be enough to authorize the command in the Daml model)
- The ledger API server normalizes `readAs` parties by removing `actAs` parties (as `actAs` implies read access)

`submitters` and `actAs` are the probably the same thing.

PR #10141, opened 2021-06-29
participant-integration-api: Move transaction requests to structured logging. [KVL-996]

SamirTalwar-DA:
This also removes the "subscription ID". I am pretty sure it's not used except for a single log line, which makes it useless as a correlation ID.

If we want to correlate logs, let's add a correlation ID.

### Changelog

- **[Ledger API Server]** The amount of data logged in the API transaction service has been reduced at INFO level. Please enable TRACE logging to log the request data structures.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
@@ -3,14 +3,11 @@
 
 package com.daml.platform.apiserver.services.transaction
 
-import java.util.concurrent.atomic.AtomicLong
-
 import akka.NotUsed
 import akka.stream.Materializer
 import akka.stream.scaladsl.Source
-import com.daml.ledger.participant.state.index.v2.IndexTransactionsService
-import com.daml.lf.data.Ref.Party
 import com.daml.grpc.adapter.ExecutionSequencerFactory
+import com.daml.ledger
 import com.daml.ledger.api.domain._
 import com.daml.ledger.api.messages.transaction._
 import com.daml.ledger.api.v1.transaction_service.{
@@ -20,11 +17,12 @@ import com.daml.ledger.api.v1.transaction_service.{
   GetTransactionsResponse,
 }
 import com.daml.ledger.api.validation.PartyNameChecker
+import com.daml.ledger.participant.state.index.v2.IndexTransactionsService
+import com.daml.lf.data.Ref.Party
 import com.daml.logging.LoggingContext.withEnrichedLoggingContext
 import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries}
 import com.daml.metrics.Metrics
 import com.daml.platform.apiserver.services.{StreamMetrics, logging}
-import com.daml.ledger
 import com.daml.platform.server.api.services.domain.TransactionService
 import com.daml.platform.server.api.services.grpc.GrpcTransactionService
 import com.daml.platform.server.api.validation.ErrorFactories
@@ -61,8 +59,6 @@ private[apiserver] final class ApiTransactionService private (
 
   private val logger = ContextualizedLogger.get(this.getClass)
 
-  private val subscriptionIdCounter = new AtomicLong()
-
   override def getTransactions(
       request: GetTransactionsRequest
   ): Source[GetTransactionsResponse, NotUsed] =
@@ -71,8 +67,7 @@ private[apiserver] final class ApiTransactionService private (
       logging.endInclusive(request.endInclusive),
       logging.parties(request.filter.filtersByParty.keys),
     ) { implicit loggingContext =>
-      val subscriptionId = subscriptionIdCounter.incrementAndGet().toString
-      logger.info(s"Received request for transaction subscription $subscriptionId: $request")
+      logger.info(s"Received request for transaction: $request")
       transactionsService
         .transactions(request.startExclusive, request.endInclusive, request.filter, request.verbose)
         .via(logger.debugStream(transactionsLoggable))
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
@@ -221,6 +221,6 @@ private[apiserver] final class ApiTransactionService private (
       logging.commandId(commandId),
       logging.transactionId(transactionId),
       logging.workflowId(workflowId),
-      "offset" -> offset,
+      logging.offset(offset),
     )
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
@@ -22,6 +22,7 @@ import com.daml.lf.data.Ref.Party
 import com.daml.logging.LoggingContext.withEnrichedLoggingContext
 import com.daml.logging.{ContextualizedLogger, LoggingContext, LoggingEntries}
 import com.daml.metrics.Metrics
+import com.daml.platform.apiserver.services.transaction.ApiTransactionService._
 import com.daml.platform.apiserver.services.{StreamMetrics, logging}
 import com.daml.platform.server.api.services.domain.TransactionService
 import com.daml.platform.server.api.services.grpc.GrpcTransactionService
@@ -32,7 +33,6 @@ import scalaz.syntax.tag._
 import scala.concurrent.{ExecutionContext, Future}
 
 private[apiserver] object ApiTransactionService {
-
   def create(
       ledgerId: LedgerId,
       transactionsService: IndexTransactionsService,
@@ -48,6 +48,14 @@ private[apiserver] object ApiTransactionService {
       ledgerId,
       PartyNameChecker.AllowAllParties,
     )
+
+  @throws[StatusRuntimeException]
+  private def getOrElseThrowNotFound[A](a: Option[A]): A =
+    a.getOrElse(
+      throw Status.NOT_FOUND
+        .withDescription("Transaction not found, or not visible.")
+        .asRuntimeException()
+    )
 }
 
 private[apiserver] final class ApiTransactionService private (
@@ -56,8 +64,10 @@ private[apiserver] final class ApiTransactionService private (
 )(implicit executionContext: ExecutionContext, loggingContext: LoggingContext)
     extends TransactionService
     with ErrorFactories {
-
   private val logger = ContextualizedLogger.get(this.getClass)
+
+  override def getLedgerEnd(ledgerId: String): Future[LedgerOffset.Absolute] =
+    transactionsService.currentLedgerEnd().andThen(logger.logErrorsOnCall[LedgerOffset.Absolute])
 
   override def getTransactions(
       request: GetTransactionsRequest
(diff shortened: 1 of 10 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -71,45 +77,25 @@ private[apiserver] final class ApiTransactionService private (
       logging.endInclusive(request.endInclusive),
       logging.parties(request.filter.filtersByParty.keys),
     ) { implicit loggingContext =>
-      val subscriptionId = subscriptionIdCounter.incrementAndGet().toString
-      logger.info(s"Received request for transaction subscription $subscriptionId: $request")
+      logger.info("Received request for transactions.")
```
tudor-da, line 80: Any reason why we want to avoid logging the request?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -71,45 +77,25 @@ private[apiserver] final class ApiTransactionService private (
...
-  ): Source[GetTransactionTreesResponse, NotUsed] =
+  ): Source[GetTransactionTreesResponse, NotUsed] = {
     withEnrichedLoggingContext(
       logging.startExclusive(request.startExclusive),
       logging.endInclusive(request.endInclusive),
       logging.parties(request.parties),
     ) { implicit loggingContext =>
-      logger.info(s"Received request for transaction tree subscription: $request")
+      logger.info("Received request for transaction trees.")
+      logger.trace(s"Transaction tree request: $request")
```
miklos-da, line 98: Nit: for consistency, you might want to use the same pattern for all log messages, e.g., "\<what\> request: \<actual-request\>". None of the below log messages follow the pattern for this particular line.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -58,10 +64,10 @@ private[apiserver] final class ApiTransactionService private (
 )(implicit executionContext: ExecutionContext, loggingContext: LoggingContext)
     extends TransactionService
     with ErrorFactories {
-
   private val logger = ContextualizedLogger.get(this.getClass)
 
-  private val subscriptionIdCounter = new AtomicLong()
```
miklos-da, line 64: :+1:

miklos-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -71,45 +77,25 @@ private[apiserver] final class ApiTransactionService private (
       logging.endInclusive(request.endInclusive),
       logging.parties(request.filter.filtersByParty.keys),
     ) { implicit loggingContext =>
-      val subscriptionId = subscriptionIdCounter.incrementAndGet().toString
-      logger.info(s"Received request for transaction subscription $subscriptionId: $request")
+      logger.info("Received request for transactions.")
+      logger.trace(s"Transaction request: $request")
```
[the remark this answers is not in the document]
SamirTalwar-DA: I think we consider trace logging to be insecure and only to be used in development, but I'm not sure.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
SamirTalwar-DA: It's massive and the data is mostly in the logging context anyway.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
SamirTalwar-DA: Whoops, I seem to have deleted the word "request" half-way down. I'll add it back in.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
tudor-da: the `template_id` filter is not logged or is it?

mziolekda: I agree with the removal of subscription id, it came from before open-sourcing of daml.

I cannot agree with the removal of the request. It is a piece of info we look very often when trying to understand what is going in the production systems. In fact I was looking at this info this morning with Tudor. It is logged only when the subscription is made, so very rarely compared to let's say transaction submission

If we want to shorten the wording of the log for "Received request for" we should then change it across the board. ATM all client requests share this wording, which makes grepping easy.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -71,45 +77,25 @@ private[apiserver] final class ApiTransactionService private (
       logging.endInclusive(request.endInclusive),
       logging.parties(request.filter.filtersByParty.keys),
     ) { implicit loggingContext =>
-      val subscriptionId = subscriptionIdCounter.incrementAndGet().toString
-      logger.info(s"Received request for transaction subscription $subscriptionId: $request")
```
kamil-da, line 75: What was this used for?

SamirTalwar-DA: It pre-dates open-source history. I looked at the closed-source history and I can see that it was used in other log lines in this method, but these have been removed.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
```diff
@@ -215,12 +203,24 @@ private[apiserver] final class ApiTransactionService private (
...
+      commandId: String,
+      transactionId: String,
+      workflowId: String,
+      offset: String,
+  ): LoggingEntries =
+    LoggingEntries(
+      logging.commandId(commandId),
+      logging.transactionId(transactionId),
+      logging.workflowId(workflowId),
+      "offset" -> offset,
```
kamil-da, line 224: How about adding a specialised function `String => LoggingEntry` in `logging` package for the offset?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
kamil-da: Cool, thanks!

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
miklos-da: By changing the production logback settings we may enable logging the requests for this particular service for a specific ledger or disable such log lines for a specific ledger via the same mechanism.
Wouldn't we be able to grep for all requests by simply grepping for 'request' under the package `com.daml.apiserver.services`?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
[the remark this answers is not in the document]
miklos-da: Yes, we may log sensitive data under TRACE level so it is not meant to be enabled for production.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
SamirTalwar-DA: I think that's a good idea. I just moved this code but I'm happy to fix it up.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
SamirTalwar-DA: Can you explain which information you were looking at this morning? I am interested in whether this information was also in the logging context. I think that contains everything I would be looking at, but I can't speak for you.

tudor-da: @SamirTalwar-DA Yes, that one is missing, together with the `ledgerId` (which is arguably useless) and `verbose` flag. But thinking better about it: instead of making the loggingContext fatter from the beginning, isn't it better to log one-offs with the `request` inlined?

tudor-da: on the other hand, if you add the `template_id` filters to the `loggingContext`, the log line would have similar size as before with the extra penalty that you would log these extra fields all over the place downstream.

SamirTalwar-DA: Yeah, that's a good point. We definitely don't want this data in every log line.

We have two competing concerns:

1. @oliverse-da, among others, have asked for shorter log lines that don't contain unnecessary information. They have moved to multi-line logging which is incompatible with the Daml SDK without a lot of rewriting.
2. You, among others, require this information.
3. Our one-off logging information is unstructured, and hard to manipulate.

So here's what I suggest:

1. We move the request data to the structured logging context, but _only_ for the `info` line, after which it is discarded.
2. We look into correlation IDs for requests which allow us to correlate log lines.
3. In the future, we find a way to automatically log the entire request in a structured format (i.e. JSON), without explicit logging code.

What do you think?

tudor-da: Sure, makes sense, but that log wasn't generating multi-lines. So what was the issue with it?

SamirTalwar-DA: These logs are generating long lines. By moving the data to the structured context, I'm hoping @oliverse-da can then output the data over multiple lines without compromising on your debugging needs.

SamirTalwar-DA: How do you feel about these log lines?

```
16:30:08.949 [in-memory-ledger-akka.actor.default-dispatcher-75] INFO  c.d.p.a.s.t.ApiTransactionService - Received request for transactions. , context: {startExclusive: "00000000000000950000000000000000", endInclusive: "%end%", filters: {TXNoContractKey-alpha-359797a68be5-party-1: "all-templates", TXNoContractKey-alpha-359797a68be5-party-0: "all-templates"}}
16:30:08.950 [in-memory-ledger-akka.actor.default-dispatcher-73] INFO  c.d.p.a.s.t.ApiTransactionService - Received request for transactions. , context: {startExclusive: "000000000000008d0000000000000000", endInclusive: "%end%", filters: {TXAgreementTextExplicit-alpha-359797a68be5-party-0: ["e1f0a0f5fef4773fe78916593f7ff9c7a8ca65f9f538192e3a5335d778cae60f:Test:Dummy"]}}
```

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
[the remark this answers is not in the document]
matthiasS-da: I wouldn't consider the `request` as confidential data, as it seems not to contain any contract data.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/transaction/ApiTransactionService.scala
SamirTalwar-DA: Done, and I moved the logging context so it _only_ wraps the `info` line for each subscription request.

kamil-da (approved):

PR #10144, opened 2021-06-29
add ES cluster

garyverhaegen-da:
This PR adds a basic ES cluster to our infrastructure, completely open and unprotected but only accessible through VPN.

And, as of yet, through its IP address. I'm not sure whether it's worth adding a DNS for it.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -1,0 +1,352 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
This has already been deployed. @aherrmann-da we'll need to coordinate on what's missing here. I'd assume something along the lines of "add Kibana and this cron job to ingest data".

cocreature (approved):
looks sensible, thanks a lot!

aherrmann-da (approved):
Thank you!

> I'd assume something along the lines of "add Kibana and this cron job to ingest data".

Yes, that sounds right.

PR #10152, opened 2021-06-30
add kibana

garyverhaegen-da:
This PR adds a Kibana instance to each ES node, and duplicates the load
balancer mechanism to expose both raw ES and Kibana.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -18,6 +18,11 @@ locals {
       size           = 0,
     }
   ]
+
+  es_ports = [
+    { name = "es", port = "9200" },
+    { name = "kibana", port = "5601" },
+  ]
 }
 
 resource "google_compute_firewall" "es-ssh" {
@@ -53,7 +58,7 @@ resource "google_compute_firewall" "es-http" {
 
   allow {
     protocol = "tcp"
-    ports    = ["80"]
+    ports    = ["5601", "9200"]
   }
 }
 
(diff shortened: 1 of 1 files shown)


infra/es_cluster.tf
```diff
@@ -151,12 +156,26 @@ EOF
...
+docker run -d \
+           --name kibana \
+           -p 5601:5601 \
+           --link es:elasticsearch \
+           -e TELEMETRY_ENABLED=false \
+           docker.elastic.co/kibana/kibana:7.13.2
+
+## Getting container output directly to the GCP console
+
+( exec 1> >(while IFS= read -r line; do echo "elastic: $line"; done); docker logs -f es ) &
```
cocreature, line 173: maybe overkill but do they have an actual healthcheck that we can ping instead? Or even better, maybe they use the proper [docker healthchecks](https://docs.docker.com/engine/reference/builder/#healthcheck). Definitely not a blocker if it’s not easy to use either of those.

cocreature (approved):
Thank you!

infra/es_cluster.tf
garyverhaegen-da: Not quite sure why you comment here, this is just log redirection.

In terms of healthcheck, what we want to know at the load balancer/instance manager level is whether or not there is a process listening for HTTP connections. Checking that the port is open seems good enough for that; the actual health endpoints on ES give information about the state of the cluster as a whole, which is completely irrelevant for load balancing and potentially damaging for "kill node" decisions the instance manager would make.

On the Kibana side, the status endpoint seems to report on two things: whether it can connect to ES (given that this is an in-machine connection, the only way this doesn't work is if the ES container is down, in which case the machine will be killed by the ES healthcheck) and whether all plugins are correctly installed with their dependencies etc., which is unlikely to change.

Finally, regarding Docker health check, I don't think there's much value in having Docker try to restart the container vs. just starting a new machine.

It was worth looking into, so thanks for prompting me, but I think I'd rather leave the setup as it is.

cocreature: oh I somehow misread this as waiting until we see a specific line in the logs which it clearly doesn’t do. need more coffee.

garyverhaegen-da: Oh, no; I just want the logs of both containers to appear as logs emitted by the init script, so they show up in the "serial output" screen in the GCP console. And while I'm at it I prefix them with either `elastic:` or `kibana:` so I know which is which.

PR #10156, opened 2021-06-30
Retry bazel cache downloads

cocreature:
This is still a bit rough. It currently retries regardless of the
error (however, with an exponential backoff and a limit so not that
big of an issue, at worst we waste some time retrying). We also print
the exception to stderr which is a bit nasty but useful for debugging.

I did verify in logs that we hit the retry code path and that things
seem to be properly cached.

I’ll clean this up & upstream this to Bazel once we have verified for
a few days/weeks that it works well for us.

Reviewing patch files is nasty so I recommend to take a look at the
branch in my Bazel fork instead
https://github.com/cocreature/bazel/tree/cache-retry.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/nix/bazel-retry-cache.patch
+++ b/nix/bazel-retry-cache.patch
@@ -1,0 +1,257 @@
+diff --git a/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java b/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java
+index 57741a8f28..6673149a20 100644
+--- a/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java
++++ b/src/main/java/com/google/devtools/build/lib/remote/RemoteCacheClientFactory.java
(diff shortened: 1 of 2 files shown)


SamirTalwar-DA (approved):
This all seems quite sensible to me. I imagine the big risk is not catching all the places which can fail, which we can address if and when it comes up.

cocreature:
> This all seems quite sensible to me. I imagine the big risk is not catching all the places which can fail, which we can address if and when it comes up.

Definitely, I did try to do my due diligence and compared it to the download for gRPC caches which has retries in the corresponding places and only those so I have some confidence but :shrug:

garyverhaegen-da (approved):
I don't know much about the Bazel code base. Looking at [the diff](https://github.com/cocreature/bazel/commit/053fac128e6fcb2d3be9028af02bd8f07544da0d?w=1), this all seems fairly plausible, but I'm not sure you should give my opinion much more weight than the compiler's on this one.

aherrmann-da (commented):
Looks good, thank you! Yes, would be great to upstream this.

PR #10158, opened 2021-06-30
LF: avoid usage of GlobalKey.key in engine error

remyhaemmerle-da:
This will allow to drop the `key: Value[ContractId]` field from
GlobalKey in an upcomming PR.

addition, this is a nice addition for #9974 (no more transaction.Error)

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
+++ b/compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
@@ -8,7 +8,7 @@ import com.daml.lf.data.{ImmArray, Numeric, Ref}
 import com.daml.lf.ledger.EventId
 import com.daml.lf.scenario.api.{v1 => proto}
 import com.daml.lf.speedy.{SError, SValue, TraceLog}
-import com.daml.lf.transaction.{GlobalKey, IncompleteTransaction, Node => N, NodeId}
+import com.daml.lf.transaction.{IncompleteTransaction, Node => N, NodeId}
 import com.daml.lf.ledger._
 import com.daml.lf.value.{Value => V}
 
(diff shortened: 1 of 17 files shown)


cocreature (commented):
Can you explain how this will work if we actually drop the value from `GlobalKey`/make it optional?
Consider the following scenario:
1. I pass in a `FetchByKey` command to `reinterpret` specifying only the key.
2. Now the engine sends `SResultNeedKey`
3. The ledger gives us `None`
4. We throw `ContractKeyNotFound` but we have no value here.
If we just make the value in `GlobalKey` optional that’s easy. But this seems to go in the opposite direction, you mandate that we have a value when throwing an error and it’s not clear to me how that will work.

remyhaemmerle-da:
> Can you explain how this will work if we actually drop the value from `GlobalKey`/make it optional?
> Consider the following scenario:
> 
>     1. I pass in a `FetchByKey` command to `reinterpret` specifying only the key hash.
> 
>     2. Now the engine sends `SResultNeedKey`
> 
>     3. The ledger gives us `None`
> 
>     4. We throw `ContractKeyNotFound` but we have no value here.
>        If we just make the value in `GlobalKey` optional that’s easy. But this seems to go in the opposite direction, you mandate that we have a value when throwing an error and it’s not clear to me how that will work.

Good idea. 

My plan is to drop the value completely from GlobalKey. This PR shows this can be done without impact on the scenario service. 

Once the value field is drop from the GlobalKey, we could consider (no obligation):
 - adapt reinterpret of "by key" operations using only the hash of the key (but not the value).
 - drop completly the value from the node.  

In those cases I would make the key value "optional" in the Error.  If we  have the value in the error we display it properly (handy for the scenario service), otherwise we just print the hash.

One possibility would be to change the type of the Error as follows:

```
sealed abstract class ContractKey 
case class ContractKeyValue(templateId: Ref.Identifier) extends ContractKey
case class ContractKeyHash(globalKey: GlobalKey) extends ContractKey

case class ContractKeyNotFound(key: ContractKey)
```

cocreature:
I got confused by what our target endstate is and a bit concerned about potential regressions in error messages so I opened https://github.com/digital-asset/daml/issues/10159 to have a discussion. Please take a look if you have some time.

I don’t think that needs to block this PR though, will review tomorrow morning.

remyhaemmerle-da:
We decide to not remove value for Global key.

PR #10160, opened 2021-07-01
DEL-8657 update version info for Haskell LF libraries

jberthold-da:
We are past the 1.14 feature set so it is 1.15, even though more changes may come in before 1.15.0 is released.

### DIFF
--- a/build-scripts/build_hs_lf_tooling.sh
+++ b/build-scripts/build_hs_lf_tooling.sh
@@ -17,7 +17,7 @@ mkdir -p $TARGET_DIR
 
 # This version needs to be adapted in all cabal files, too
 # The script below will fail on the `cp` command otherwise.
-LIB_VERSION="0.1.13.0"
+LIB_VERSION="0.1.15.0"
 
 package_from_dir() {
     local dir=$1
--- a/compiler/daml-lf-ast/daml-lf-ast.cabal
+++ b/compiler/daml-lf-ast/daml-lf-ast.cabal
@@ -1,7 +1,7 @@
 cabal-version: 2.4
 name: daml-lf-ast
 build-type: Simple
-version: 0.1.13.0
+version: 0.1.15.0
 synopsis: DAML-LF AST
 license: Apache-2.0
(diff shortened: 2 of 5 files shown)


garyverhaegen-da (approved):
Not very familiar with this setup, but bumping seems sensible enough. Is there any reason why this should not be unified with, say, the max version from LATEST?

jberthold-da:
Thanks Gary! 
Using `LATEST` is definitely an option (I was unaware that this version file exists and is maintained). The only caveat would be that  modules might get added or removed every now and then, which could render the build invalid unless the `*.cabal` file reflects the change. In the past there was no desire from the language team to maintain these cabal files so I took it upon myself, and I can continue to do it manually.
### Options
1) keep the status-quo of manual version bumps and maintenance
2) use `LATEST` to determine the version number; accept temporary breakage and leave maintenance of `cabal` files with me
3) use `LATEST` to determine the version number; add a test for the artefact so that everybody working on these packages needs to adapt the `cabal` files when their cabal/stack build fails

Note that 3) still misses out on added modules if they are unused (but that is perfectly acceptable for my purpose). 3) also means to figure out how to reliably build the packages with stack plus a given stackage resolver, but that should not be hard.

garyverhaegen-da:
For context, the `LATEST` file is not just "maintained", it _is_ the trigger for releasing new versions. So you really can rely on it being up-to-date.

I don't know enough about the Haskell world to understand how much effort is involved in maintaining cabal files, and I don't know enough about these particular packages to have any idea how often they change and thus might break, so I'm afraid I won't be much help in making a decision here.

cocreature:
I’ll happily accept PRs that switch to `LATEST` (although it’s not entirely clear to me that this is correct, there is a reason why nothing else reads that file and we instead get the version from an env var) but given that these bindings are explicitly not supported externally, I’m not going to spend effort from my teams on those cabal files.

garyverhaegen-da:
Looks like option 1, then. Sorry for the noise.

PR #10163, opened 2021-07-01
daml-lf/data: Optionally truncate party names in structured logs. [KVL-996]

SamirTalwar-DA:
The party name can grow quite long, so we offer ledger implementors the opportunity to truncate it in structured log output.

Unfortunately, because we use Logback through the global `LoggerFactory`, there is no place to inject logging configuration. This means we also need to use global, mutable state to configure logging output. I have added a `LoggingConfiguration` class+object in Daml-LF Data, which may not be the best place, but I can't think of a better one right now. I suggest we leave it there until it has reason to grow, at which point we may want to move it.

I moved the logging implicits to a new package to reduce the number of explicit transitive dependencies.

### Pull Request Checklist

trigger the build.


oliverse-da (approved):
Looks good. Thank you, @SamirTalwar-DA

stefanobaghino-da (approved):

miklos-da (approved):

PR #10162, opened 2021-07-01
compress bazel build logs

garyverhaegen-da:
Anecdotally, I see a 25x reduction in size when compressing. Time to compress and decompress is negligible, whereas storage costs and transfer times may not be.

Since, as far as I'm aware, we don't currently have anything depending on the current format, I could run a script locally to transform all of the existing logs to match the new format.

CHANGELOG_BEGIN
CHANGELOG_END

PR #10176, opened 2021-07-05
restore print of job metadata

garyverhaegen-da:
I just noticed I erroneously removed that line in #10162.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #10179, opened 2021-07-05
Add divulgence warning and test in script service.

sofiafaro-da:
Part of #9947, building on the key visibility checks from #10136

This PR adds a divulgence warning inside the traceLog.
I wasn't sure whether:

1. these warnings should be kept in a separate structure from the
   traceLog, and therefore transmitted separately, or
2. these warnings should be kept together with traces, but the severity
   should be tracked and also transmitted over grpc, and warnings should
   be logged as warnings instead of as debug messages, or
3. these warnings should be kept together with traces, but
   logged as warnings instead of debug messages,
4. these warnings should be treated exactly like traces

I'm leaning toward option 2, but this PR implements option 3.

This PR tests the warning via the script service tests. There should be more tests in more places, but I think getting it working with script is a good first approximation.

The other thing this is missing is some context on where the fetch was performed.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -743,6 +745,16 @@ private[lf] object Speedy {
       returnValue = go(typ0, value0)
     }
 
+    def checkContractVisibility(onLedger: OnLedger, cid: V.ContractId, contract: CachedContract) = {
+      onLedger.visibleToStakeholders(contract.stakeholders) match {
+        case SVisibleToStakeholders.Visible => ()
+        case SVisibleToStakeholders.NotVisible(actAs, readAs) =>
+          this.traceLog.addWarning(
+            s"${contract.templateId} contract ${cid} not visible to actAs = ${actAs}, readAs = ${readAs}.",
```
cocreature, line 753:
```suggestion
            s"Tried to fetch or exercise ${contract.templateId} contract ${cid} but none of the reading parties to actAs = ${actAs}, readAs = ${readAs} are a stakeholder ${contract.stakeholders}. Use of divulged contracts is deprecated and incompatible with pruning",
```
Something like this maybe?

compiler/damlc/tests/src/DA/Test/ScriptService.hs
```diff
@@ -596,6 +596,53 @@ main =
...
+                    , "      do fetch cid"
+                    , ""
+                    , "testDivulge = do"
+                    , "  p1 <- allocateParty \"p1\""
+                    , "  p2 <- allocateParty \"p2\""
+                    , "  cid <- submit p1 (createCmd (T p1))"
+                    , "  divulgeCid <- submit p2 (createCmd (Divulge p1 p2 cid))"
+                    , "  submit p1 (exerciseCmd divulgeCid Accept)"
+                    , "  delegateCid <- submit p1 (createCmd (Delegate p1 p2 cid))"
+                    , "  submit p2 (exerciseCmd delegateCid Fetch)"
```
cocreature, line 639: maybe worth testing both fetch and exercise? I’d also like to see a test that a local contract created in the same transaction does not trigger the warning.

cocreature (approved):
Nice! A few comments. I think I agree with you that option 2 is the nicest but I’m happy to merge as is and then improve later.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
sofiafaro-da: Great suggestion, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/TraceLog.scala
```diff
@@ -27,6 +28,14 @@ private[lf] final case class RingBufferTraceLog(logger: Logger, capacity: Int) e
       size += 1
   }
 
+  def addWarning(message: String, optLocation: Option[Location]): Unit = {
+    logger.warn(s"Warning: $message")
```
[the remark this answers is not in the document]
sofiafaro-da: Removed it for now.

PR #10180, opened 2021-07-05
Improvements to the documentation with regards to offsets

stefanobaghino-da:
changelog_begin
[Docs] Improvements to the documentation with regards to offsets
changelog_end

- Simplify wording to explain the usage of `blockingForEach` in the Java
bindings quickstart page
- Describe offsets in prose in the page about the Ledger API services
- Suggest how to perform crash recovery with offsets on the application
architecture page
- Link to the relevant Protobuf reference docs where approriate

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/app-dev/services.rst
+++ b/docs/source/app-dev/services.rst
@@ -164,11 +164,11 @@ See :ref:`verbosity` above.
   The RPCs exposed as part of the transaction and active contracts services make use of offsets.
 
   An offset is an ever-growing number assigned by the participant to each transaction as they are received from the ledger.
-  This number is opaque to the client and must be used as-is (i.e. cliend-side transformations like incrementing it are not guaranteed to produce a valid offset).
+  This number is opaque to the client and must be used as-is (i.e. client-side transformations like incrementing it are not guaranteed to produce a valid offset).
   The state of a ledger (i.e. the set of active contracts) as exposed by the Ledger API is valid at a specific offset, which is why the last message your application receives when calling the ``ActiveContractsService`` is precisely that offset.
   In this way, the client can keep track of the relevant state without needing to invoke the ``ActiveContractsService`` again, by starting to read transactions from the given offset.
 
-  Offsets are also useful to perform crash recovery and fail over as documented more in depth in the :ref:`application architecture <dealing-with-failures>` page.
+  Offsets are also useful to perform crash recovery and failover as documented more in depth in the :ref:`application architecture <dealing-with-failures>` page.
 
   You can read more about offsets in the `protobuf documentation of the API <../app-dev/grpc/proto-docs.html#ledgeroffset>`__.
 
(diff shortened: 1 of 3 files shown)


docs/source/app-dev/services.rst
```diff
@@ -159,6 +159,19 @@ Verbosity
 
 See :ref:`verbosity` above.
 
+.. note::
+
+  The RPCs exposed as part of the transaction and active contracts services make use of offsets.
+
+  An offset is an ever-growing number assigned by the participant to each transaction as they are received from the ledger.
+  This number is opaque to the client and must be used as-is (i.e. cliend-side transformations like incrementing it are not guaranteed to produce a valid offset).
```
cocreature, line 167:
```suggestion
  This number is opaque to the client and must be used as-is (i.e. client-side transformations like incrementing it are not guaranteed to produce a valid offset).
```

cocreature: I believe we guarantee that they can be lexicographically ordered. Worth mentioning?

docs/source/app-dev/services.rst
```diff
@@ -159,6 +159,19 @@ Verbosity
...
+  The state of a ledger (i.e. the set of active contracts) as exposed by the Ledger API is valid at a specific offset, which is why the last message your application receives when calling the ``ActiveContractsService`` is precisely that offset.
+  In this way, the client can keep track of the relevant state without needing to invoke the ``ActiveContractsService`` again, by starting to read transactions from the given offset.
+
+  Offsets are also useful to perform crash recovery and fail over as documented more in depth in the :ref:`application architecture <dealing-with-failures>` page.
```
cocreature, line 171:
```suggestion
  Offsets are also useful to perform crash recovery and failover as documented more in depth in the :ref:`application architecture <dealing-with-failures>` page.
```
https://en.wikipedia.org/wiki/Failover

cocreature (approved):

docs/source/app-dev/services.rst
stefanobaghino-da: Mmmm, we do and we mention it in the (linked) referenced docs where they are just an array of byte. Here I call them just "ever-growing number"s for simplicity. Saying that they can be lexicographically ordered seems redundant. Does that make sense?

cocreature: I think ever-growing number is somewhat misleading. You basically give the impression of something with very strong guarantees (e.g., consecutive) and then step back in the next sentence. I personally find it more intuitive to go with "they’re completely opaque, you can only order them lexicographically” but both work so your choice.

stefanobaghino-da: Better? https://github.com/digital-asset/daml/pull/10180/commits/e857a7cf3f62c30816b79de67ea0dcb689a125a3

stefanobaghino-da: I'll assume yes, let me know if you prefer me to change the wording.

PR #10187, opened 2021-07-06
reduce noise in Scala bindings

garyverhaegen-da:
It's unlikely most clients will want all the details of the stack trace.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlow.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlow.scala
@@ -74,9 +74,9 @@ object CommandTrackerFlow {
                 1,
                 { case e =>
                   logger.warn(
-                    s"Completion Stream failed with an error. Trying to recover in ${backOffDuration} ..",
-                    e,
+                    s"Completion Stream failed with an error. Trying to recover in ${backOffDuration} .."
                   )
+                  logger.debug(e)
                   delayedEmptySource(backOffDuration)
                 },
               )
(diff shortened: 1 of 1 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlow.scala
```diff
@@ -74,9 +74,9 @@ object CommandTrackerFlow {
                 1,
                 { case e =>
                   logger.warn(
-                    s"Completion Stream failed with an error. Trying to recover in ${backOffDuration} ..",
-                    e,
+                    s"Completion Stream failed with an error. Trying to recover in ${backOffDuration} .."
                   )
+                  logger.debug(e)
```
cocreature, line 79: is it worth adding some text to make it clear that this belongs to the log statement above or is the logger + potential trace contexts enough?

cocreature (approved):
Thanks!

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlow.scala
garyverhaegen-da: I'd be enclined to say that logger context + stack trace + the fact that this will appear right after the WARN + the message in the exception itself should be enough, but I'm not going to be the one reading these logs, so I'm open to suggestions. A simple solution could be to just repeat the WARN message in the DEBUG logs:

```suggestion
                  logger.debug(
                    s"Completion Stream failed with an error. Trying to recover in ${backOffDuration} ..",
                    e,
                  )
```

cocreature: your choice, “right after” can be a bit of a tricky concept once you add concurrency but probably clear enough.

PR #10193, opened 2021-07-06
feed elasticsearch cluster

garyverhaegen-da:
This PR adds a machine that will, every 5 minutes, look at the GCS
bucket that stores Bazel metrics and push whatever it finds to
ElasticSearch.

A huge part of this commit is based on @aherrmann-da's work. You can
assume that all the good bits are his.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -78,16 +78,17 @@ resource "google_compute_firewall" "es-internal" {
   }
 }
 
-
-resource "google_service_account" "es" {
-  account_id   = "elastic"
-  display_name = "elastic"
+resource "google_service_account" "es-discovery" {
+  # account_id allows - but not _
+  account_id   = "es-discovery"
+  display_name = "es-discovery"
 }
 
-resource "google_project_iam_custom_role" "es" {
-  role_id     = "elastic"
-  title       = "elastic"
-  description = "elastic"
+resource "google_project_iam_custom_role" "es-discovery" {
+  # role_id allows _ but not -
+  role_id     = "es_discovery"
+  title       = "es-discovery"
+  description = "es-discovery"
   permissions = [
     # Cloud logging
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Because there is no great way to develop & test this kind of work without actually deploying all of it, it's all already deployed and currently running. We have about 3k jobs to ingest and it takes a few seconds per job, so hopefully by tomorrow morning we should have most of the data loaded. 🤞

infra/es_cluster.tf
```diff
@@ -208,10 +209,10 @@ resource "google_compute_instance" "es-init" {
   }
 
   service_account {
-    email = google_service_account.es.email
+    email = google_service_account.es-discovery.email
     scopes = [
       # Required for cloud logging
-      "cloud-platform",
+      "logging-write",
```
cocreature, line 215: is that less or more permissions or just different and the old one didn’t work?

cocreature (approved):
looks sensible, I admit I can’t claim to fully grasp the details and some of this seems like a "if it works it works type of thing” but don’t see anything off. Thank you!

infra/es_cluster.tf
garyverhaegen-da: It's less, but it's also unchanged. GCP has this great notion of having two different permission systems for instances: at one level, you specify "scopes", which is what we're doing here. This is the "old", deprecated-but-you-still-have-to-use-it way. The official recommendation (from GCP) here is to set it to "cloud-platform", which means "access to everything".

The "current" way of specifying permissions is via the service account associated with the instance.

In order for an action to succeed it needs to be permitted by both the scopes and the service account (which makes debugging permission issues a bit painful at times). So this is restricting the scope, but we were already (and still are) fully restricted to the absolute minimum set of rights by the service account, which is why "less, but actually the same".

PR #10200, opened 2021-07-07
ci/windows: disable spool

garyverhaegen-da:
We're not expecting to print anything, and @RPS5' security newsletter
says this is a vector of attack.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/vsts_agent_windows.tf
+++ b/infra/vsts_agent_windows.tf
@@ -13,7 +13,19 @@ locals {
       suffix     = "",
       size       = 6,
       assignment = "default",
+      extra      = "",
     },
+    {
+      suffix     = "-no-spool",
+      size       = 1,
+      assignment = "default",
+      extra      = <<EOF
+
+# Disable Print Sppoler service (security)
+Stop-Service -Name Spooler -Force
+Set-Service -Name Spooler -StartupType Disabled
+EOF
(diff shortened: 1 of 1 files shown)


infra/vsts_agent_windows.tf
```diff
@@ -83,7 +95,7 @@ Set-StrictMode -Version latest
 $ErrorActionPreference = 'Stop'
 
 # Disable Windows Defender to speed up disk access
-Set-MpPreference -DisableRealtimeMonitoring $true
+Set-MpPreference -DisableRealtimeMonitoring $true${local.w[count.index].extra}
```
cocreature, line 98: what does this change do?

garyverhaegen-da: This inserts lines 23-26 for the "no-spool" cohort, while doing no change at all to the default machines, so they don't get killed prematurely.

Arguably I could have put that on the blank line below.

The "mergeable" version of this PR will get back to just the default group, and this change inlined.

cocreature: Oh I see, I missed the fact that it has a newline at the beginning so it won’t actually affect this command.

cocreature (approved):
Thanks!

PR #10201, opened 2021-07-07
Warn on DA.Exception import

cocreature:
We had a few confused users that imported that module and got weird
errors. The warning should hopefully make things a bit clearer. We
could do this for other modules as well but I’ll leave that for
separate PRs.

changelog_begin

- [Daml Compiler] Importing DA.Exception on LF < 1.14 now produces a
warning that exceptions require Daml-LF >= 1.14.

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/Exception.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/Exception.daml
@@ -6,7 +6,9 @@
 #ifndef DAML_EXCEPTIONS
 
 -- | HIDE
-module DA.Exception where
+module DA.Exception
+  {-# WARNING "Your target Daml-LF version is too old for exceptions. You need Daml-LF 1.14 or newer" #-}
+  where
 
 #else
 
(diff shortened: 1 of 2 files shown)


sofiafaro-da (approved):
Good idea, thanks. Maybe it would be a good idea to a test in daml-test-files  (using @UNTIL-LF).

cocreature:
> Good idea, thanks. Maybe it would be a good idea to add a test in daml-test-files (using @UNTIL-LF).

done, slightly annoyed that UNTIL-LF is not exclusive. Something like UNTIL-LF-FEAUTURE exceptions (doesn’t exist yet but it shouldn’t be exclusive if the non-feature version is inclusive) would be nicer.

sofiafaro-da:
> > Good idea, thanks. Maybe it would be a good idea to add a test in daml-test-files (using @UNTIL-LF).
> 
> done, slightly annoyed that UNTIL-LF is not exclusive. Something like UNTIL-LF-FEAUTURE exceptions (doesn’t exist yet but it shouldn’t be exclusive if the non-feature version is inclusive) would be nicer.

I agree!

PR #10204, opened 2021-07-07
Release DAML SDK 1.14.1

tudor-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


cocreature (approved):
lgtm assuming vmbc tests passed which I’m sure you tested, thanks!

tudor-da:
> lgtm assuming vmbc tests passed which I’m sure you tested, thanks!

Not yet, but holding off until everything is fine there

tudor-da:
@moritzkiefer-da Updated this PR's `LATEST` file with the SHA for 1.14.1 to the latest commit published in `release/1.14.x` (after merging https://github.com/digital-asset/daml/pull/10212)

cocreature:
looks sensible, it’s slightly fishy to not release a separate snapshot first but given that the commit didn’t change anything in the code only build scripts, I’m fine with this but I would suggest to get @gerolf-da’s approval as well.

gerolf-da:
I'm inclined to rather use the exact same commit for the 1.14.1 RC-snapshot and the final 1.14.1 release. There's really no gain to be had from including this commit in the 1.14.1 release itself.
If I understand #10212 correctly, it's for branch hygiene only. Let's stick to good practices for releases.

LATEST
```diff
@@ -1,5 +1,5 @@
 62aabcc478d0357448c42d10b3c75d1ef8b13c10 1.15.0-snapshot.20210705.7286.0.62aabcc4
-8ec16e94fff58c55858c8c97378b62cab80e1180 1.14.1-snapshot.20210706.7133.0.8ec16e94
+a673f58850d9bd3b6ee32a8e7ac8f41f91ae21f7 1.14.1
```
gerolf-da, line 2:
```suggestion
8ec16e94fff58c55858c8c97378b62cab80e1180 1.14.1
```

cocreature:
> If I understand #10212 correctly, it's for branch hygiene only. Let's stick to good practices for releases.

It’s not, the build fails without this because we run the CI config from `main` with `build.sh` from the release branch which does not produce the log directory. The other alternative is to modify the CI config from `main` to skip the upload on releases or something like that.

tudor-da:
> > If I understand #10212 correctly, it's for branch hygiene only. Let's stick to good practices for releases.
> 
> It’s not, the build fails without this because we run the CI config from `main` with `build.sh` from the release branch which does not produce the log directory. The other alternative is to modify the CI config from `main` to skip the upload on releases or something like that.

I can redo both releases to stand by the clean practice. Fine by you, @gerolf-da ?

LATEST
gerolf-da: Withdrawing this suggestion, as without the commit we cannot build anymore.

gerolf-da:
Sorry, I missed it. In this case the new snapshot also wouldn't bring much value. Let's go for 1.14.1 on the latest commit on the branch then.

PR #10207, opened 2021-07-07
es-feed: ignore invalid files

garyverhaegen-da:
We currently have about 1% (28 out of 2756) of our build logs that have invalid JSON files. They are all about a `-profile` file being incomplete, and since those files represent a single JSON object we can't do smarter things like filtering invalid individual lines.

I haven't looked deeply into _why_ we create invalid files, but this should let our ingestion process make some progress in the meantime.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -797,24 +797,28 @@ bulk_upload() {
 }
 
 push() {
-  local job
+  local job f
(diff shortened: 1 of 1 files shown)


cocreature (approved):
looks very sensible, thank you!

PR #10211, opened 2021-07-08
Ensure lookupMaximumLedgerTime contention failures are propagated normally [DPP-481]

tudor-da:
This PR fixes a bug that bubbles up a contention-driven exception instead of passing it down via a failed Future. The issue manifests itself as a unhandled failure to the server instead of a handled submission result in `ApiSubmissionService`. 

The fix allows to return a `ContractNotFound` wrapped in a `Future` instead of throwing when a `lookupMaximumLedgerTime` hits a cached archived/not found contract.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
@@ -30,6 +30,7 @@ import com.daml.platform.store.interfaces.LedgerDaoContractsReader.{
 
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 private[platform] class MutableCacheBackedContractStore(
     metrics: Metrics,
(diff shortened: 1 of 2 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
```diff
@@ -200,6 +200,10 @@ private[apiserver] final class ApiSubmissionService private[services] (
       telemetryContext: TelemetryContext,
   ): Future[SubmissionResult] =
     for {
+      // Start the flatMap chain with a successful Future in order to ensure
```
cocreature, line 203: any reason not to fix the command executor instead? imho functions that return futures should never throw synchronously. Otherwise you get a confusing mess which is just asking for bugs.

tudor-da: The command executor is assumed fixed with this PR. But consider additional bugs that make the executor violate the contract and making it throw. What happens in that scenario is that this method will violate the contract as well which translates in bypassing the `stopDeduplicatingCommand` step.

tudor-da: > imho functions that return futures should never throw synchronously. Otherwise you get a confusing mess which is just asking for bugs.

True, but this contract is not enforced at typelevel but rather just by trust. As mentioned above, such a bug can leave the ledger in a bad state: it will not accept a re-submission, even though the current one did not go through.

cocreature: fair enough I guess my question is where you draw the line. If you never make that assumption you have to insert the dummy future everywhere which is not what we do atm afaict. Adding it in one place while keeping the assumption in other places seems inconsistent.

tudor-da: That future there is not best practice, but it does one thing in this particular case: it prevents a bug for causing more harm. I am happy to restructure this code a bit more to handle more clearly failures with always executing `stopDeduplicatingCommand`. Or you would rather not do that as well and just leave it like this?

cocreature: I’ll leave it up to you. I think what I was getting at is that we shouldn’t add inconsistent safe guards for bugs somewhere but instead fix the underlying bugs. But totally fine if you don't want to do this in this PR or not at all for now.

tudor-da: Ok, as this might be slightly controversial change, I'll drop it. I believe this part will be revisited anyhow as part of the command deduplication rearch

mziolekda (approved):
Nice catch! LGTM

PR #10219, opened 2021-07-08
tweak ES cluster

garyverhaegen-da:
This PR contains many small changes:

- A small refactoring whereby the "es-init" machine is now
  (syntactically) integrated with the two instance groups, to cut down a
  bit on repetition.
- The feeder machine is now preemptible, because I've seen it recover
  enough times that I'm confident this will not cause any issue.
- Indices are now sharded.
- Return values from ES are filtered, cutting down a bit on network
  usage and memory requirements to produce the responses.
- Bulk uploads for a single job are now done in parallel. This results
  in about a 2x speedup for ingestion.
- crontab was changed to very minute instead of every 5 minutes.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -6,17 +6,27 @@ resource "google_compute_network" "es" {
 }
 
 locals {
+  es_ssh  = 0
+  es_feed = 1
   es_clusters = [
     {
       suffix         = "-blue",
       ubuntu_version = "2004",
-      size           = 5,
+      size           = 10,
+      init           = "[]",
     },
     {
       suffix         = "-green",
       ubuntu_version = "2004",
       size           = 0,
-    }
+      init           = "[]",
+    },
+    {
+      suffix         = "-init",
+      ubuntu_version = "2004",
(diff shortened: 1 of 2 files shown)


infra/es_cluster.tf
```diff
@@ -6,17 +6,27 @@ resource "google_compute_network" "es" {
...
       suffix         = "-green",
       ubuntu_version = "2004",
       size           = 0,
-    }
+      init           = "[]",
+    },
+    {
+      suffix         = "-init",
+      ubuntu_version = "2004",
+      size           = 0,
```
cocreature, line 27: I’m confused, what does size 0 do?

cocreature (approved):
looks sensible, thanks!

infra/es_cluster.tf
garyverhaegen-da: That's the number of machines. In normal operation, you don't want any machine with an initial_master set. So this is there only as a knob to turn manually if/when reinitializing the cluster.

PR #10221, opened 2021-07-08
Multiplex multiple virtual query streams through a single web socket

### LINKED ISSUE #7796: Reduce the number of transaction streams resulting from JS stream queries
At the moment each call to `useStreamQueries` and `useStreamFetchByKeys` results in one websocket connection to the JSON API and this in turn results in one transaction stream per websocket connection. Even with less than < 100 of such connections we have seen issues on the ledger side (encountered by customers). Considering concurrent users, this is quite easy to reach with just 10 queries and 10 users so we should do something about this.

Specifically, we should reduce the number of transaction streams to 1 per browser session (~ user). Reducing to 1 websocket connection isn’t quite as crucial but I expect hundreds of websocket connections are also going to cause issues eventually and I think this is actually easier to solve together.

2 constraints that I’d like to stick to:

1. Keep the per-template queries in the JS library. Typing multi-template queries is awkward.

2. Related to 1: Don’t reimplement the query language provided by the JSON API on the client side.

There are various ways to go about this but here is one proposal:

1. Add a new endpoint to the JSON API. This endpoint accepts multiple queries (including key queries) each with a query id supplied by the user. The result is the same stream of queries that you get on the current streaming endpoints but each message has an extra field for the query id. If a transaction (or parts thereof) needs to be included in multiple queries, you will get multiple messages for each of them. No guarantees are made about the order of messages for different queries.

2. The JS library is changed to use a single websocket connection for all queries. Based on the query id, we associate messages with the relevant query. The user-facing APIs stay the same.

Open question:
The JS side is necessarily dynamic here wtr to the queries we make. We have two options:
1. Accept dynamic changes for the queries in the JSON API endpoint. This makes things easier for the JS libraries but does require a bit of juggling in the JSON API to stop and reopen transaction streams on changes.
2. No dynamic changes in the JSON API, instead close and reopen the websocket connection. Less complexity in the JSON API but more complexity in the JS libraries and does require streaming the whole ACS again.

I’m leaning towards 1 but I haven’t thought about this for very long.

stefanobaghino-da:
Fixes #7796

changelog_begin
[TS bindings] When using `daml-react` and `daml-ledger`, all streaming request to the query endpoint will be multiplexed through a single web socket
changelog_end

### Pull Request Checklist

trigger the build.


language-support/ts/daml-ledger/index.test.ts
```diff
@@ -153,7 +153,7 @@ describe("streamSubmit", () => {
     stream.on("change", mockChange);
     const restoreConsole = mockConsole();
     mockInstance.serverSend({ warnings: ["oh oh"] });
-    expect(console.warn).toHaveBeenCalledWith("Ledger.streamQueries warnings", {"warnings": ["oh oh"]});
+    expect(console.warn).toHaveBeenCalledWith("QueryStreamsManager warnings", {"warnings": ["oh oh"]});
```
garyverhaegen-da, line 156: The point here was to contain the function name the user actually called, so I'd have a slight preference for not changing the log messages if possible. QueryStreamsManager is an implementation detail the user should not care about, presumably.

language-support/ts/daml-ledger/index.ts
```diff
@@ -127,15 +131,15 @@ const decodeArchiveEventUnknown: jtv.Decoder<ArchiveEvent<object>> =
  * Decoder for an [[Event]].
  */
 const decodeEvent = <T extends object, K, I extends string>(template: Template<T, K, I>): jtv.Decoder<Event<T, K, I>> => jtv.oneOf<Event<T, K, I>>(
-  jtv.object({created: decodeCreateEvent(template)}),
+  jtv.object({created: decodeCreateEvent(template), matchedQueries: jtv.array(jtv.number())}),
   jtv.object({archived: decodeArchiveEvent(template)}),
```
garyverhaegen-da, line 135: Is it a JSON API limitation that you can't match archive events to specific streams?

language-support/ts/daml-ledger/index.ts
```diff
@@ -196,9 +200,7 @@ export type Query<T> = T extends object ? {[K in keyof T]?: Query<T[K]>} : T;
  * queries. As long as that restriction stays, there is no need for any kind of
  * encoding here.
  */
-function encodeQuery<T extends object, K, I extends string>(template: Template<T, K, I>, query?: Query<T>): unknown {
-  // I could not get the "unused" warning silenced, but this seems to count as "used"
-  [template];
+function encodeQuery<T extends object, K, I extends string>(_template: Template<T, K, I>, query?: Query<T>): unknown {
```
garyverhaegen-da, line 203: Fairly sure I'd tried that 🤔. Oh well. Thanks for fixing!

stefanobaghino-da: We may have changed the compiler version and/or configuration in the meantime, not sure.

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: Unfortunately yes. When reading the event from the Ledger API, creates are matched with the predicate based on the contract arguments. The archive event doesn't have the arguments, so this cannot be done. I guess it should be possible to achieve this by matching up the active contract set with the archive event and retrieving the argument from there, but this is not currently possible with the JSON API.

language-support/ts/daml-ledger/index.test.ts
stefanobaghino-da: Initially I was aiming to do this but decided to keep it simple as I didn't think it was too much of an issue, but this makes sense. I'll do it, thanks for the feedback.

language-support/ts/daml-ledger/index.ts
```diff
@@ -329,6 +331,277 @@ type LedgerOptions = {
...
+
+}
+
+type QueryResponseStream<T extends object, K, I extends string> = StreamEventEmitter<T, K, I, readonly CreateEvent<T, K, I>[]>;
+
+type StreamingQuery<T extends object, K, I extends string> = {
+    template: Template<T, K, I>,
+    queries: Query<T>[],
+    stream: QueryResponseStream<T, K, I>,
+    offset?: string | null, // undefined -> nothing has been received, null -> the JSON API returned a null offset
```
garyverhaegen-da, line 358: My TypeScript is a bit rusty; would it be possible to define some sort of enum or sentinel objects here? Googling yielded [this](https://2ality.com/2020/01/special-values-typescript.html), which suggests something like:

```typescript
const NoOffsetReceivedYet = Symbol('NoOffsetReceivedYet');
const NullOffsetReceived = Symbol('NullOffsetReceived');
...
    offset: string | typeof NoOffsetReceivedYet | typeof NullOffsetReceived,
```

language-support/ts/daml-ledger/index.ts
```diff
@@ -329,6 +331,277 @@ type LedgerOptions = {
...
+        map.set(key, [value]);
+    }
+}
+
+/**
+ * @deprecated All usages of this function should be replaced by just
+ *             iterating over the iterator. For this to happen, the TS
+ *             compiler requires the --downlevelIteration flag, which
+ *             however does not play nicely with Jest when running the
+ *             tests.
```
garyverhaegen-da, line 381: Another option would be to target es6, I think. A quick look at [can I use?](https://caniuse.com/?search=es6) seems to say the last major browser to offer compatibility was Firefox in June 2017. Do we still have a good reason to target es5?

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: I think this sounds very reasonable, thanks.

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: No idea. @cocreature?

language-support/ts/daml-ledger/index.ts
```diff
@@ -329,6 +331,277 @@ type LedgerOptions = {
...
+    //    - this causes the consumer to observe what is known as "phantom archives", which are known and documented
+    private matchIndexLookupTable: [StreamingQuery<object, unknown, string>, number][] = [];
+    private templateIdsLookupTable: {[templateId: string]: Set<StreamingQuery<object, unknown, string>>} = {};
+
+    // Accumulates each query in a flattened form to be sent as a single request to the JSON API
+    private request: StreamingQueryRequest[] = [];
+
+    // web socket handle and associated properties
+    private ws: WebSocket | null = null;
+    private wsLiveSince: number | undefined = undefined;
```
garyverhaegen-da, line 416: Why the distinction between null and undefined here?

language-support/ts/daml-ledger/index.ts
cocreature: not sure, it’s hard to verify if any of our users rely on es5

language-support/ts/daml-ledger/index.test.ts
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/10221/commits/0811eff4dc794a2f0ce12bd06ca85f40aef5cbbe

This also restores what would have been the previous behavior of having one log line for every outstanding query. Since (so far) I tried to make sure that the behavior of virtual streams is identical to the existing one backed by an actual web socket, this is a welcome improvement. 🙇🏻

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/10221/commits/e87bc5ca21b3d8f8b7b23ac6dff5fc40fe33e8f0

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: I'll leave this as is for the time being. No rush to take a decision. I would prefer to keep working against ES5 unless there are strong reasons not to. But honestly "we have been in the third decade of the third millennium for a while" is probably at least a good reason.

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: In my mind `undefined` is something which can be absent and will have an actual value at a later point, while `null` is something I explicitly want to define as such.

Would you have found

```suggestion
    private wsLiveSince?: number;
```

to be easier to read? And maybe use a symbol to denote when the web socket is shut down?

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: Embedded the idea as a `TODO` in the code in https://github.com/digital-asset/daml/pull/10221/commits/2cd0bc5ec99961cb088bd74516a1199a2a60e328

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: As mentioned offline, going with `?` for both fields seems the most sensible thing to do.

stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/10221/commits/8e1889d1c686611adc7fa02238e26327df302e17

PR #10224, opened 2021-07-08
Consolidating EventStorageBackend [DPP-461]

nmarton-da:
* Switching to composed anorm queries
* Switching to data-driven prepared statements where possible

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da (commented):
Looks good. 🙂 If you have data regarding the performance that would be useful.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -0,0 +1,464 @@
...
+        s"$selectColumnsForTransactionTree, ${eventStrategy.columnEqualityBoolean("event_kind", "20")} as exercise_consuming",
+      witnessesColumn = "tree_event_witnesses",
+    )(
+      limit = rangeParams.limit,
+      fetchSizeHint = rangeParams.fetchSizeHint,
+      filterParams,
+    )(connection)
+  }
+
+  // TODO the ordering of the results here is based on node_index in the previous implementation. current implementation uses event_sequential_id. Are these the same? (if not perhaps flatTransaction needs adapting too?)
```
rautenrieth-da, line 410: Did you check this TODO?

1. `event_sequential_id` is generated by the indexer and was introduced by the append-only schema, so we should be able to verify whether the ordering is the same
2. AFAICT, the ordering of events is only important when constructing the root nodes of the transaction, [see here](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/EventsTable.scala#L117).
3. The JdbcLedgerDao suite has transactions with multiple root nodes, [see here](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoSuite.scala#L610). Some of the tests would fail if the order of root nodes were shuffled, e.g., "filter correctly by multiple parties with the same template". The root nodes in the tests are always sorted by node id (because that's how TransactionBuilder adds them), not sure whether that is always the true and whether it affects the result.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -0,0 +1,464 @@
...
+    )(
+      limit = None,
+      fetchSizeHint = None,
+      filterParams,
+    )(connection)
+  }
+
+}
+
+trait EventStrategy {
```
rautenrieth-da, line 441: I like this style of customizing the queries :+1:

As the code base around constructing queries gets more and more complex, it would be great if you could document some of the interfaces. E.g., describe here the return type of the methods, in particular that the first element of the tuple is a plain string that contains a SQL fragment where named parameters can be included in curly braces.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -0,0 +1,464 @@
...
+  def filteredEventWitnessesClause(
+      witnessesColumnName: String,
+      parties: Set[Ref.Party],
+  ): (String, List[NamedParameter])
```
rautenrieth-da, line 445: I assume each method is responsible for finding globally unique names for the named parameters? Would anorm complain if you accidentally defined the value of a named parameter twice?

rautenrieth-da (commented):
Nice work consolidating the queries!

I am slightly worried about the fact that it gets more and more complicated to understand the SQL queries used by the index/indexer. You have to execute quite a lot of code in your head when trying to determine what SQL statement will be sent to the database, it's another reason to make sure our test coverage is really good.

As noted by @stefanobaghino-da, I would like to see a performance comparison before merging this.

nmarton-da:
Performance verification was concluded by @tudor-da (thank you!). Baseline is always the root of the branch

1: vanilla sample run for 1400 seconds@600TPS on single-node Sandbox-on-X
baseline: https://lr.da-ext.net/grafana/d/sandbox-on-x/sandbox-on-x?orgId=1&var-benchmark_id=1593&from=1625854284000&to=1625855961000
PR: https://lr.da-ext.net/grafana/d/sandbox-on-x/sandbox-on-x?orgId=1&var-benchmark_id=1594&from=1625893451000&to=1625895119000
Conclusion: No visible difference between baseline and PR

2: custom benchmark (with ledger-api-bench-tool) ran on Tudor's dev cluster, it is nearly identical to this (https://digitalasset.atlassian.net/wiki/spaces/DAMLX/pages/2438561809/Participant+stability+experiments#Repeated-GetTransactions%2FGetTransactionTrees-subscriptions)  run with two differences (1 hours duration, ofc the branch)
Visual data not availble, hence of private cluster.
Conclusion: No visible difference between baseline and PR

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
stefanobaghino-da: Wasn't the `event_sequential_id` introduced before append-only schema was a thing to ensure the correct usage of indexes for paginatation?

stefanobaghino-da:
> Performance verification was concluded by @tudor-da (thank you!). Baseline is always the root of the branch
> 
> 1: vanilla sample run for 1400 seconds@600TPS on single-node Sandbox-on-X
> baseline: https://lr.da-ext.net/grafana/d/sandbox-on-x/sandbox-on-x?orgId=1&var-benchmark_id=1593&from=1625854284000&to=1625855961000
> PR: https://lr.da-ext.net/grafana/d/sandbox-on-x/sandbox-on-x?orgId=1&var-benchmark_id=1594&from=1625893451000&to=1625895119000
> Conclusion: No visible difference between baseline and PR
> 
> 2: custom benchmark (with ledger-api-bench-tool) ran on Tudor's dev cluster, it is nearly identical to this (https://digitalasset.atlassian.net/wiki/spaces/DAMLX/pages/2438561809/Participant+stability+experiments#Repeated-GetTransactions%2FGetTransactionTrees-subscriptions) run with two differences (1 hours duration, ofc the branch)
> Visual data not availble, hence of private cluster.
> Conclusion: No visible difference between baseline and PR


Terrific work, thanks! 🙇🏻

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -949,8 +949,8 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
        FROM
            participant_events
        WHERE
-           event_sequential_id > #$startExclusive
-           and event_sequential_id <= #$endInclusive
+           event_sequential_id > $startExclusive
+           and event_sequential_id <= $endInclusive
```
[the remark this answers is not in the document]
nmarton-da: > Nice work consolidating the queries!
> 
> I am slightly worried about the fact that it gets more and more complicated to understand the SQL queries used by the index/indexer. You have to execute quite a lot of code in your head when trying to determine what SQL statement will be sent to the database, it's another reason to make sure our test coverage is really good.
> 
> As noted by @stefanobaghino-da, I would like to see a performance comparison before merging this.

Thanks. That is true, although we probably want to look at full SQL queries to understand what is going on: with the previous implementation we also had quite some stuff which was happening in different algorithms.

nmarton-da:
> Nice work consolidating the queries!
> 
> I am slightly worried about the fact that it gets more and more complicated to understand the SQL queries used by the index/indexer. You have to execute quite a lot of code in your head when trying to determine what SQL statement will be sent to the database, it's another reason to make sure our test coverage is really good.
> 
> As noted by @stefanobaghino-da, I would like to see a performance comparison before merging this.

Thanks. That is true, although we probably want to look at full SQL queries to understand what is going on: with the previous implementation we also had quite some stuff which was happening in different algorithms.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: Based on usage, but yes

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -0,0 +1,464 @@
...
+  ).mkString(", ")
+
+  private def events[T](
+      columnPrefix: String,
+      joinClause: String,
+      additionalAndClause: (String, List[NamedParameter]),
+      rowParser: RowParser[T],
+      selectColumns: String,
+      witnessesColumn: String,
+  )(
```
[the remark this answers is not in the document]
nmarton-da: Just to conceptually separate something which is data and injected functionality

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: @stefanobaghino-da you're right, it was introduced before as well. It slightly changed with parallel ingestion, because before it was populated by a db sequence, now scala-logic (to be able to assign these separately from actual parallel linsertion)

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: So these ordering should be the same, as confirmed by @remyhaemmerle-da 


Marton Nagy Today at 3:37 PM
Question about NodeId: if we traverse a transaction with .foldInExecutionOrder will that sequence result monotonically increasing nodeIds? (eg the ordering defined by execution-order and nodeId is the same?) 

remy
Yes, if the transaction you consider is generated from the engine.
If you omit end exerciseEnd and  rollbackEnd

Marton Nagy
we do that indeed (omit) (edited)

nmarton-da: Removed theTODO

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: Oh yes anorm will complain then! I added comments to clarify this behavior

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: Added documentation

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: Correction: anorm won't complain. I made a comment about that, will improve this in the next PR

stefanobaghino-da:
> > Nice work consolidating the queries!
> > I am slightly worried about the fact that it gets more and more complicated to understand the SQL queries used by the index/indexer. You have to execute quite a lot of code in your head when trying to determine what SQL statement will be sent to the database, it's another reason to make sure our test coverage is really good.
> > As noted by @stefanobaghino-da, I would like to see a performance comparison before merging this.
> 
> Thanks. That is true, although we probably want to look at full SQL queries to understand what is going on: with the previous implementation we also had quite some stuff which was happening in different algorithms.

Good point.

PR #10227, opened 2021-07-09
HA PoC behind a feature flag [DPP-426]

nmarton-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/BUILD.bazel
+++ b/ledger/participant-integration-api/BUILD.bazel
@@ -62,7 +62,6 @@ compile_deps = [
     "@maven//:io_opentelemetry_opentelemetry_api",
     "@maven//:io_opentelemetry_opentelemetry_context",
     "@maven//:org_slf4j_slf4j_api",
-    # this oracle import is problematic for daml assistant build
     "@maven//:com_h2database_h2",
     "@maven//:org_postgresql_postgresql",
     "@maven//:com_oracle_database_jdbc_ojdbc8",
--- a/ledger/participant-integration-api/BUILD.bazel
+++ b/ledger/participant-integration-api/BUILD.bazel
@@ -63,6 +63,9 @@ compile_deps = [
     "@maven//:io_opentelemetry_opentelemetry_context",
     "@maven//:org_slf4j_slf4j_api",
     # this oracle import is problematic for daml assistant build
+    "@maven//:com_h2database_h2",
+    "@maven//:org_postgresql_postgresql",
+    "@maven//:com_oracle_database_jdbc_ojdbc8",
 ]
 
 scala_compile_deps = [
@@ -79,9 +82,6 @@ scala_compile_deps = [
 
 runtime_deps = [
     "@maven//:ch_qos_logback_logback_classic",
-    "@maven//:com_h2database_h2",
-    "@maven//:org_postgresql_postgresql",
-    "@maven//:com_oracle_database_jdbc_ojdbc8",
 ]
 
 da_scala_library(
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerConfig.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerConfig.scala
@@ -6,6 +6,7 @@ package com.daml.platform.indexer
 import com.daml.lf.data.Ref
 import com.daml.platform.configuration.IndexConfiguration
 import com.daml.platform.indexer.IndexerConfig._
+import com.daml.platform.indexer.ha.HaConfig
 import com.daml.platform.store.DbType
 
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
@@ -23,6 +24,7 @@ case class IndexerConfig(
     allowExistingSchema: Boolean = false,
     // TODO append-only: remove after removing support for the current (mutating) schema
     enableAppendOnlySchema: Boolean = false,
+    // TODO append-only: this is now configuring only the append-only pool
     asyncCommitMode: DbType.AsyncCommitMode = DefaultAsyncCommitMode,
     maxInputBufferSize: Int = DefaultMaxInputBufferSize,
     inputMappingParallelism: Int = DefaultInputMappingParallelism,
@@ -32,6 +34,7 @@ case class IndexerConfig(
     tailingRateLimitPerSecond: Int = DefaultTailingRateLimitPerSecond,
     batchWithinMillis: Long = DefaultBatchWithinMillis,
     enableCompression: Boolean = DefaultEnableCompression,
+    haConfig: HaConfig = HaConfig(),
 )
 
 object IndexerConfig {
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerConfig.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerConfig.scala
@@ -24,7 +24,6 @@ case class IndexerConfig(
     allowExistingSchema: Boolean = false,
     // TODO append-only: remove after removing support for the current (mutating) schema
     enableAppendOnlySchema: Boolean = false,
-    // TODO append-only: this is now configuring only the append-only pool
     asyncCommitMode: DbType.AsyncCommitMode = DefaultAsyncCommitMode,
     maxInputBufferSize: Int = DefaultMaxInputBufferSize,
     inputMappingParallelism: Int = DefaultInputMappingParallelism,
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
@@ -27,11 +27,9 @@ import com.daml.platform.store.DbType.{
   SynchronousCommit,
 }
 import com.daml.platform.store.appendonlydao.events.{CompressionStrategy, LfValueTranslation}
-import com.daml.platform.store.backend.DataSourceStorageBackend.{
-  DataSourceConfig,
-  PgSynchronousCommitValue,
-}
+import com.daml.platform.store.backend.DataSourceStorageBackend.DataSourceConfig
 import com.daml.platform.store.backend.StorageBackend
+import com.daml.platform.store.backend.postgresql.PostgresDataSourceConfig
 import com.daml.platform.store.dao.LedgerDao
 import com.daml.platform.store.{DbType, FlywayMigrations, LfValueTranslationCache}
 
@@ -181,11 +179,13 @@ object JdbcIndexer {
         batchWithinMillis = config.batchWithinMillis,
         metrics = metrics,
         dataSourceConfig = DataSourceConfig(
-          pgSynchronousCommit = Some(config.asyncCommitMode match {
-            case SynchronousCommit => PgSynchronousCommitValue.On
-            case AsynchronousCommit => PgSynchronousCommitValue.Off
-            case LocalSynchronousCommit => PgSynchronousCommitValue.Local
-          })
+          postgresConfig = PostgresDataSourceConfig(
+            synchronousCommit = Some(config.asyncCommitMode match {
+              case SynchronousCommit => PostgresDataSourceConfig.SynchronousCommitValue.On
+              case AsynchronousCommit => PostgresDataSourceConfig.SynchronousCommitValue.Off
+              case LocalSynchronousCommit => PostgresDataSourceConfig.SynchronousCommitValue.Local
+            })
+          )
         ),
         haConfig = config.haConfig,
       )
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
@@ -135,7 +135,6 @@ object JdbcIndexer {
             servicesExecutionContext,
             metrics,
             lfValueTranslationCache,
-            jdbcAsyncCommitMode = config.asyncCommitMode,
             enricher = None,
             participantId = config.participantId,
           )
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
@@ -21,7 +21,16 @@ import com.daml.platform.common
 import com.daml.platform.common.MismatchException
 import com.daml.platform.configuration.ServerRole
 import com.daml.platform.indexer.parallel.ParallelIndexerFactory
+import com.daml.platform.store.DbType.{
+  AsynchronousCommit,
+  LocalSynchronousCommit,
+  SynchronousCommit,
+}
 import com.daml.platform.store.appendonlydao.events.{CompressionStrategy, LfValueTranslation}
+import com.daml.platform.store.backend.DataSourceStorageBackend.{
+  DataSourceConfig,
+  PgSynchronousCommitValue,
+}
 import com.daml.platform.store.backend.StorageBackend
 import com.daml.platform.store.dao.LedgerDao
 import com.daml.platform.store.{DbType, FlywayMigrations, LfValueTranslationCache}
@@ -171,6 +180,14 @@ object JdbcIndexer {
         tailingRateLimitPerSecond = config.tailingRateLimitPerSecond,
         batchWithinMillis = config.batchWithinMillis,
         metrics = metrics,
+        dataSourceConfig = DataSourceConfig(
+          pgSynchronousCommit = Some(config.asyncCommitMode match {
+            case SynchronousCommit => PgSynchronousCommitValue.On
+            case AsynchronousCommit => PgSynchronousCommitValue.Off
+            case LocalSynchronousCommit => PgSynchronousCommitValue.Local
+          })
+        ),
+        haConfig = config.haConfig,
       )
     }
 
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
@@ -16,38 +16,42 @@ import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, ExecutionContext, Future}
 
 /** A handle of a running program
-  * @param completed will complete as the program completes
+  * @param completed will complete right after the program completed
   *                  - if no KillSwitch used,
   *                    - it will complete successfully as program successfully ends
   *                    - it will complete with a failure as program failed
   *                  - if KillSwitch aborted, this completes with the same Throwable
   *                  - if KillSwitch shut down, this completes successfully
-  *                  As this completes, the program finished it's execution, also all resources released as well.
+  *                  After signalling completion, the program finished it's execution and has released all resources it acquired.
   * @param killSwitch to signal abortion and shutdown
   */
 case class Handle(completed: Future[Unit], killSwitch: KillSwitch)
 
-/** This functionality sign off a worker Connection, anc clears it for further usage.
+/** This functionality initializes a worker Connection, and clears it for further usage.
   * This only need to be done once at the beginning of the Connection life-cycle
   * On any error an exception will be thrown
   */
-trait SignConnection {
-  def sign(connection: Connection): Unit
+trait ConnectionInitializer {
+  def initialize(connection: Connection): Unit
 }
 
-/** To add High Availability related features to a program
+/** To add High Availability related features to a program, which intends to use database-connections to do it's work.
+  * Features include:
+  *   - Safety: mutual exclusion of these programs ensured by DB locking mechanisms
+  *   - Availability: release of the exclusion is detected by idle programs, which start competing for the lock to do
+  *     their work.
   */
 trait HaCoordinator {
 
-  /** Execute block in High Availability mode.
-    * Wraps around the Handle of the block.
+  /** Execute in High Availability mode.
+    * Wraps around the Handle of the execution.
     *
-    * @param block HaCoordinator provides a SignConnection which need to be used for all database connections to do work in the block
-    *              Future[Handle] embodies asynchronous initialisation of the block
-    *              (e.g. not the actual work. That asynchronous execution completes with the completed Future of the Handle)
+    * @param initializeExecution HaCoordinator provides a ConnectionInitializer which need to be used for all database connections during execution
+    *                            Future[Handle] embodies asynchronous initialisation of the execution
+    *                            (e.g. not the actual work. That asynchronous execution completes with the completed Future of the Handle)
     * @return the new Handle, which is available immediately to observe and interact with the complete program here
     */
-  def protectedBlock(block: SignConnection => Future[Handle]): Handle
+  def protectedExecution(initializeExecution: ConnectionInitializer => Future[Handle]): Handle
 }
 
 case class HaConfig(
@@ -64,10 +68,10 @@ object HaCoordinator {
   private val logger = ContextualizedLogger.get(this.getClass)
 
   /** This implementation of the HaCoordinator
-    * - provides a database lock based isolation of the protected blocks
-    * - will run the block at-most once during the entire lifecycle
-    * - will wait infinitely to acquire the lock needed to start the protected block
-    * - provides a SignConnection function which is mandatory to execute on all worker connections inside of the block
+    * - provides a database lock based isolation of the protected executions
+    * - will run the execution at-most once during the entire lifecycle
+    * - will wait infinitely to acquire the lock needed to start the protected execution
+    * - provides a ConnectionInitializer function which is mandatory to execute on all worker connections during execution
     * - will spawn a polling-daemon to observe continuous presence of the main lock
     *
     * @param dataSource to spawn the main connection which keeps the Indexer Lock
@@ -91,7 +95,7 @@ object HaCoordinator {
       def acquireLock(connection: Connection, lockId: LockId, lockMode: LockMode): Lock = {
         logger.debug(s"Acquiring lock $lockId $lockMode")
         storageBackend
-          .aquireImmediately(lockId, lockMode)(connection)
+          .tryAcquire(lockId, lockMode)(connection)
           .getOrElse(
             throw new Exception(s"Cannot acquire lock $lockId in lock-mode $lockMode: lock busy")
           )
@@ -127,7 +131,7 @@ object HaCoordinator {
           mainLockChecker <- go[PollingChecker](
             new PollingChecker(
               periodMillis = haConfig.mainLockCheckerPeriodMillis,
-              checkBlock = acquireMainLock(mainConnection),
+              check = acquireMainLock(mainConnection),
               killSwitch =
                 handle.killSwitch, // meaning: this PollingChecker will shut down the main preemptableSequence
             )
@@ -154,7 +158,7 @@ object HaCoordinator {
             mainLockChecker.check()
             logger.info(s"Preparing worker connection DONE.")
           }))
-          _ = logger.info("Step 6: initialize protected block - DONE")
+          _ = logger.info("Step 6: initialize protected execution - DONE")
           _ <- merge(protectedHandle)
         } yield ()
       }
@@ -162,6 +166,8 @@ object HaCoordinator {
 }
 
 object NoopHaCoordinator extends HaCoordinator {
-  override def protectedBlock(block: SignConnection => Future[Handle]): Handle =
-    Await.result(block(_ => ()), Duration.Inf)
+  override def protectedExecution(
+      initializeExecution: ConnectionInitializer => Future[Handle]
+  ): Handle =
+    Await.result(initializeExecution(_ => ()), Duration.Inf)
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
@@ -1,0 +1,167 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.indexer.ha
+
+import java.sql.Connection
+
+import akka.actor.Scheduler
+import akka.stream.KillSwitch
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.platform.store.backend.DBLockStorageBackend.{Lock, LockId, LockMode}
+import com.daml.platform.store.backend.DBLockStorageBackend
+import javax.sql.DataSource
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+/** A handle of a running program
+  * @param completed will complete as the program completes
+  *                  - if no KillSwitch used,
+  *                    - it will complete successfully as program successfully ends
+  *                    - it will complete with a failure as program failed
+  *                  - if KillSwitch aborted, this completes with the same Throwable
+  *                  - if KillSwitch shut down, this completes successfully
+  *                  As this completes, the program finished it's execution, also all resources released as well.
+  * @param killSwitch to signal abortion and shutdown
+  */
+case class Handle(completed: Future[Unit], killSwitch: KillSwitch)
+
+/** This functionality sign off a worker Connection, anc clears it for further usage.
+  * This only need to be done once at the beginning of the Connection life-cycle
+  * On any error an exception will be thrown
+  */
+trait SignConnection {
+  def sign(connection: Connection): Unit
+}
+
+/** To add High Availability related features to a program
+  */
+trait HaCoordinator {
+
+  /** Execute block in High Availability mode.
+    * Wraps around the Handle of the block.
+    *
+    * @param block HaCoordinator provides a SignConnection which need to be used for all database connections to do work in the block
+    *              Future[Handle] embodies asynchronous initialisation of the block
+    *              (e.g. not the actual work. That asynchronous execution completes with the completed Future of the Handle)
+    * @return the new Handle, which is available immediately to observe and interact with the complete program here
+    */
+  def protectedBlock(block: SignConnection => Future[Handle]): Handle
+}
+
+case class HaConfig(
+    mainLockAquireRetryMillis: Long = 500,
+    workerLockAquireRetryMillis: Long = 500,
+    workerLockAquireMaxRetry: Long = 1000,
+    mainLockCheckerPeriodMillis: Long = 1000,
+    indexerLockId: Int = 100,
+    indexerWorkerLockId: Int = 101,
+)
+
+object HaCoordinator {
+
+  private val logger = ContextualizedLogger.get(this.getClass)
+
+  /** This implementation of the HaCoordinator
+    * - provides a database lock based isolation of the protected blocks
+    * - will run the block at-most once during the entire lifecycle
+    * - will wait infinitely to acquire the lock needed to start the protected block
+    * - provides a SignConnection function which is mandatory to execute on all worker connections inside of the block
+    * - will spawn a polling-daemon to observe continuous presence of the main lock
+    *
+    * @param dataSource to spawn the main connection which keeps the Indexer Lock
+    * @param storageBackend is the database-independent abstraction of session/connection level database locking
+    * @param executionContext which is use to execute initialisation, will do blocking/IO work, so dedicated execution context is recommended
+    */
+  def databaseLockBasedHaCoordinator(
+      dataSource: DataSource,
+      storageBackend: DBLockStorageBackend,
+      executionContext: ExecutionContext,
+      scheduler: Scheduler,
+      haConfig: HaConfig,
+  )(implicit loggingContext: LoggingContext): HaCoordinator = {
+    implicit val ec: ExecutionContext = executionContext
+
+    val indexerLockId = storageBackend.lock(haConfig.indexerLockId)
+    val indexerWorkerLockId = storageBackend.lock(haConfig.indexerWorkerLockId)
+    val preemptableSequence = PreemptableSequence(scheduler)
+
+    asyncHandle =>
+      def acquireLock(connection: Connection, lockId: LockId, lockMode: LockMode): Lock = {
+        logger.debug(s"Acquiring lock $lockId $lockMode")
+        storageBackend
+          .aquireImmediately(lockId, lockMode)(connection)
+          .getOrElse(
+            throw new Exception(s"Cannot acquire lock $lockId in lock-mode $lockMode: lock busy")
+          )
+      }
+
+      def acquireMainLock(connection: Connection): Unit = {
+        acquireLock(connection, indexerLockId, LockMode.Exclusive)
+        ()
+      }
+
+      preemptableSequence.executeSequence { sequenceHelper =>
+        import sequenceHelper._
+        logger.info("Starting databaseLockBasedHaCoordinator")
+        for {
+          mainConnection <- go[Connection](dataSource.getConnection)
+          _ = logger.info("Step 1: creating main-connection - DONE")
+          _ = registerRelease {
+            logger.info("Releasing main connection...")
+            mainConnection.close()
+            logger.info("Released main connection")
+          }
+          _ <- retry(haConfig.mainLockAquireRetryMillis)(acquireMainLock(mainConnection))
+          _ = logger.info("Step 2: acquire exclusive Indexer Lock on main-connection - DONE")
+          exclusiveWorkerLock <- retry[Lock](
+            haConfig.workerLockAquireRetryMillis,
+            haConfig.workerLockAquireMaxRetry,
+          )(
+            acquireLock(mainConnection, indexerWorkerLockId, LockMode.Exclusive)
+          )
+          _ = logger.info("Step 3: acquire exclusive Indexer Worker Lock on main-connection - DONE")
+          _ <- go(storageBackend.release(exclusiveWorkerLock)(mainConnection))
+          _ = logger.info("Step 4: release exclusive Indexer Worker Lock on main-connection - DONE")
+          mainLockChecker <- go[PollingChecker](
+            new PollingChecker(
+              periodMillis = haConfig.mainLockCheckerPeriodMillis,
+              checkBlock = acquireMainLock(mainConnection),
+              killSwitch =
+                handle.killSwitch, // meaning: this PollingChecker will shut down the main preemptableSequence
+            )
+          )
+          _ = logger.info(
+            "Step 5: activate periodic checker of the exclusive Indexer Lock on the main connection - DONE"
+          )
+          _ = registerRelease {
+            logger.info(
+              "Releasing periodic checker of the exclusive Indexer Lock on the main connection..."
+            )
+            mainLockChecker.close()
+            logger.info(
+              "Released periodic checker of the exclusive Indexer Lock on the main connection"
+            )
+          }
+          protectedHandle <- goF(asyncHandle(workerConnection => {
+            // this is the checking routine on connection creation
+            // step 1: acquire shared worker-lock
+            logger.info(s"Preparing worker connection. Step 1: acquire lock.")
+            acquireLock(workerConnection, indexerWorkerLockId, LockMode.Shared)
+            // step 2: check if main connection still holds the lock
+            logger.info(s"Preparing worker connection. Step 2: checking main lock.")
+            mainLockChecker.check()
+            logger.info(s"Preparing worker connection DONE.")
+          }))
+          _ = logger.info("Step 6: initialize protected block - DONE")
+          _ <- merge(protectedHandle)
+        } yield ()
+      }
+  }
+}
+
+object NoopHaCoordinator extends HaCoordinator {
+  override def protectedBlock(block: SignConnection => Future[Handle]): Handle =
+    Await.result(block(_ => ()), Duration.Inf)
+}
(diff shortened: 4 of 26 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
```diff
@@ -0,0 +1,233 @@
...
+    * - KillSwitch events will be replayed to this handle
+    * - In case of abort/shutdown the PreemptableSequence's completion result will conform to the KillSwitch usage,
+    *   not to the completion of this handle (although it will wait for it naturally)
+    *
+    * @param handle The handle to delegate to
+    * @return the completion of the Handle
+    */
+  def merge(handle: Handle): Future[Unit]
+
+  /** The handle of the PreemprableSequence. This handle is available for sequence construction as well.
```
meiersi-da, line 80: typo

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
```diff
@@ -0,0 +1,233 @@
...
+trait UsedKillSwitch extends KillSwitch {
+  override def shutdown(): Unit = ()
+  override def abort(ex: Throwable): Unit = ()
+}
+case object ShutDownKillSwitch extends UsedKillSwitch
+case class AbortedKillSwitch(ex: Throwable, _myReference: AtomicReference[KillSwitch])
+    extends CaptureKillSwitch(_myReference)
+class CaptureKillSwitch(myReference: AtomicReference[KillSwitch]) extends KillSwitch {
+  override def shutdown(): Unit = myReference.set(ShutDownKillSwitch)
+  override def abort(ex: Throwable): Unit = myReference.set(AbortedKillSwitch(ex, myReference))
```
meiersi-da, line 100: I'm not sure this code implements the above requirements in a thread-safe way. The problem I see is that a thread can get interrupted by another thread between reading the reference and setting it.

Probably a better way would be to implement all the pure logic of what value overrides another one in case classes; and then have a single class that implements the management of the AtomicReference and its update.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
```diff
@@ -0,0 +1,213 @@
...
+    val connectionId =
+      List.fill(8)(scala.util.Random.nextPrintableChar()).mkString // TODO FIXME maybe not needed
+    logger.debug(s"$connectionId Creating new connection")
+    val connection = delegate.getConnection
+    try {
+      logger.debug(s"$connectionId Applying connection init hook")
+      initHook(connection)
+    } catch {
+      case t: Throwable =>
+        logger.info(s"$connectionId Init hook execution failed: ${t.getMessage}")
```
meiersi-da, line 57: Shouldn't this be a warning?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
```diff
@@ -521,4 +527,72 @@ private[backend] object OracleStorageBackend
...
+            lockmode => #$oracleLockMode,
+            timeout => 0
+          ) FROM DUAL"""
+      .as(get[Int](1).single)(connection) match {
+      case 0 => Some(DBLockStorageBackend.Lock(lockId, lockMode))
+      case 1 => None
+      case 2 => throw new Exception("Aquiring lock caused a deadlock!")
+      case 3 => throw new Exception("Parameter error")
+      case 4 => Some(DBLockStorageBackend.Lock(lockId, lockMode))
+      case 5 => throw new Exception("Illegal lock handle")
```
meiersi-da, line 561: Where are these exceptions logged? And will they carry the context information that this was a Oracle DBMS_LOCK request that failed?

meiersi-da (commented):
Good stuff @nmarton-da ! Some small comments, but nothing major.

ledger/participant-integration-api/src/main/scala/platform/indexer/JdbcIndexer.scala
```diff
@@ -20,7 +20,16 @@ import com.daml.platform.common
...
 import com.daml.platform.indexer.parallel.ParallelIndexerFactory
+import com.daml.platform.store.DbType.{
+  AsynchronousCommit,
+  LocalSynchronousCommit,
+  SynchronousCommit,
+}
 import com.daml.platform.store.appendonlydao.events.{CompressionStrategy, LfValueTranslation}
+import com.daml.platform.store.backend.DataSourceStorageBackend.{
+  DataSourceConfig,
+  PgSynchronousCommitValue,
```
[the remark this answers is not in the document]
nmarton-da: Yeah this is a bit weird but intentional: I'd like to leave StorageBackend agnostic of the logic what is the desirable configuration for which StorageBackend, and how we make that configurable otherwise.
So that config is part of the interface, it is deliberatly exposed. For example if we want to configure here oracle as well we will have to do that for both.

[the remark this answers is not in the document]
meiersi-da: Ack. I thought so once I've read the further code.

ledger/participant-integration-api/BUILD.bazel
```diff
@@ -61,6 +61,9 @@ compile_deps = [
     "@maven//:io_opentelemetry_opentelemetry_context",
     "@maven//:org_slf4j_slf4j_api",
     # this oracle import is problematic for daml assistant build
+    "@maven//:com_h2database_h2",
+    "@maven//:org_postgresql_postgresql",
+    "@maven//:com_oracle_database_jdbc_ojdbc8",
```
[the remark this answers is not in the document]
nmarton-da: That comment was there before, I did not touch that, I am not aware of problems that regard (CI passes)
@dasormeter can you pls comment on the intent of that comment here?

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
```diff
@@ -0,0 +1,167 @@
...
+    * @return the new Handle, which is available immediately to observe and interact with the complete program here
+    */
+  def protectedBlock(block: SignConnection => Future[Handle]): Handle
+}
+
+case class HaConfig(
+    mainLockAquireRetryMillis: Long = 500,
+    workerLockAquireRetryMillis: Long = 500,
+    workerLockAquireMaxRetry: Long = 1000,
+    mainLockCheckerPeriodMillis: Long = 1000,
```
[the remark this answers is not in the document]
nmarton-da: For one: this is not entirely true:
* Until fail-over is done, quite some other work is needed, so the total fail over time is probably much longer
* Tuning this guy only helps in a corner scenario, where on node1 just the main connection is lost, and node1 worker connections still grab on the shared worker lock-s. In this situation node2 can already grab the main lock, and is already in the loop for getting the exclusive worker lock, so this case also workerLockAquireRetryMillis would need adapting, to achive the desirable effect.
* This corner scenario above is quite unlikely, much more likely is that the connection is severed entirely, and all connections (therefore locks) released promptly. In this case it is indifferent what happens on node1, interesting would be mainly the mainLockAquireRetryMillis, which controls on node2 how fast it can acquire the main lock, and start the sequence from there.

Also for testing purposes I would prefer to keep this not too frequent, so we are not littering the logs aggressively.
Would you be okay to tune this (and preferably other durations here) for HA deployed scenarios with config overrides?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
nmarton-da: That is a good question: it depends on our logging policies. 
If potentially transient problems should materialize as warnings, then yes.
@mziolekda can you maybe comment on this

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
nmarton-da: The exception carries this information with the stack trace. 
In HaCoordinator it will be logged without stack trace (only on debug level while retying, and on info level if permanently failed). 
This code also runs in InitHookDataProxy: there it is just exception message on info level.
And bubbles up to HikariCP: there
- if hikari initializes it will be an ERROR log with the full stack trace
- if hikari already initialized it will be swallowed whole

On any rate: I was thinking that the exception messages only will point us to the right direction since it is unique to oracle.
what is your suggestion?

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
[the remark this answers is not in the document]
meiersi-da: Thanks for the detailed explanation. Let's tune this via config and leave the 1s default.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
meiersi-da: I would prefix the messages with `Oracle DB Error ${errorNo}: <your explanation>`. This way the message itself says clearly that this is Oracle specific.

I'd also change `Parameter error` to `Parameter error when trying to acquire lock` again with the motivation to make the error message itself more specific wrt the point that raised it.

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
```diff
@@ -0,0 +1,74 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+
+import akka.stream.KillSwitch
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+
+import scala.util.{Failure, Success, Try}
+
+class PollingChecker(
+    periodMillis: Long,
+    checkBlock: => Unit,
```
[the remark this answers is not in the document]
nmarton-da: no such block notion... I don't know where I got this, as a matter of fact I don't like it neither :smile: let's remove it

wrt Bool/Unit in the result: I would rather keep to the Unit, so no extra exception handling needed. Would like to follow the "whatever goes wrong is wrong, otherwise good" approach here. (this guy would have little room to maneuver if it gets an exception anyway)

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
nmarton-da: spot on! :dart: 
rewritten

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
```diff
@@ -40,7 +40,7 @@ trait OracleAround {
       val con = use(
         DriverManager.getConnection(
           s"jdbc:oracle:thin:@localhost:$port/ORCLPDB1",
-          systemUser,
+          "sys as sysdba", // TODO this is needed for being able to grant the execute access for the sys.dbms_lock below. Consider making this configurable
```
nmarton-da, line 43: @dasormeter can you please comment on this?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
nmarton-da: fixed accordingly

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
```diff
@@ -0,0 +1,212 @@
...
+}
```
meiersi-da, line 101: @nmarton-da : what is this code still used for?

nmarton-da: :man_facepalming: removed

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -235,6 +238,63 @@ object EventStorageBackend {
...
+      connectionInitHook: Option[Connection => Unit] = None,
+  )(implicit loggingContext: LoggingContext): DataSource
+}
+
+object DataSourceStorageBackend {
+  case class DataSourceConfig(
+      pgSynchronousCommit: Option[PgSynchronousCommitValue] = None
+  )
+
+  sealed abstract class PgSynchronousCommitValue(val pgSqlName: String)
```
mziolekda, line 255: Is the Pg prefix justified? Should this be generic instead?

nmarton-da: Same was raised by @meiersi-da as well: these are db-specific configurations, so I was opting for having ultimate level of configurability here. For example we do not have async commit  for H2, and we have it (currently not configurable) for Oracle, but with entirely different knobs, with entirely different meanings.
Abstracting over Async commit looses precision, and the idea here to have this loss of precision above the StorageBackend interface, to have free hands later on to implement mapping on-demand

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/KillSwitchCaptor.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.concurrent.atomic.AtomicReference
+
+import akka.stream.KillSwitch
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+
+/** A KillSwitch which captures it's usage
+  * - Shutdown always wins
+  * - From aborts, the last abort wins
+  * - With setDelegate() we can set a delegate KillSwitch, which to usage will be replayed
```
mziolekda, line 14: Can you rephrase this comment. It is not clear

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
```diff
@@ -0,0 +1,64 @@
...
+    periodMillis,
+    periodMillis,
+  )
+
+  // This is a cruel approach for ensuring single threaded usage of the mainConnection.
+  // In theory this could have been made much more efficient: not enqueueing for a check of it's own,
+  // but collecting requests, and replying in batches.
+  // Although experiments show approx 1s until a full connection pool is initialized at first
+  // (the peak scenario) which should be enough, and which can leave this code very simple.
+  def check(): Unit = synchronized {
```
mziolekda, line 50: Can you rename to avoid confusion with the parameter?

ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerConfig.scala
```diff
@@ -23,6 +24,7 @@ case class IndexerConfig(
     allowExistingSchema: Boolean = false,
     // TODO append-only: remove after removing support for the current (mutating) schema
     enableAppendOnlySchema: Boolean = false,
+    // TODO append-only: this is now configuring only the append-only pool
```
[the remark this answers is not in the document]
nmarton-da: added

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/KillSwitchCaptor.scala
nmarton-da: rephrased

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
nmarton-da: renamed

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
dasormeter: we need the extra permissions if we are making use of oracle locking - it's actually a good point that this may not be needed for index server as that locking functionality is only there for canton, and i don't think is applied for the index server

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
mziolekda: We are lacking a comprehensive strategy in that respect, but we have plenty of prior art. Transient problems pertaining to platform capabilities have usually been logged as warnings.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
```diff
@@ -0,0 +1,65 @@
...
+      case t: Throwable =>
+        logger.info(s"Init hook execution failed: ${t.getMessage}")
+        throw t
+    }
+    logger.info(s"Init hook execution finished successfully, connection ready")
+    connection
+  }
+
+  override def getConnection(s: String, s1: String): Connection = {
+    val connection = delegate.getConnection(s, s1)
```
mziolekda, line 47: Why the protective bolts and braces are omitted in this case?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
mziolekda: Is the idea to have other DB-specific flags as fields of the DataSourceConfig?
```
case class DataSourceConfig(
      pgSynchronousCommit: Option[PgSynchronousCommitValue] = None
      oracleSomeConfigKnob: Option[OracleSomeConfigKnobValue] = None
     etc...
  )
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DbDispatcher.scala
```diff
@@ -83,22 +82,24 @@ private[platform] object DbDispatcher {
...
       serverRole: ServerRole,
       jdbcUrl: String,
       connectionPoolSize: Int,
       connectionTimeout: FiniteDuration,
       metrics: Metrics,
-      connectionAsyncCommitMode: DbType.AsyncCommitMode,
   )(implicit loggingContext: LoggingContext): ResourceOwner[DbDispatcher] =
     for {
-      connectionProvider <- HikariJdbcConnectionProvider.owner(
+      hikariConnectionPool <- new HikariDataSourceOwner(
```
mziolekda, line 93: Technically speaking we are getting here HikariDataSource, consider renaming the variable

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
nmarton-da: warning it is, changing it

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
nmarton-da: exactly

libs-scala/oracle-testing/src/main/scala/testing/oracle/OracleAround.scala
nmarton-da: alright, I leave it like this, and extend `DPP-495 Unit testing: DBLockStorageBackend` with sorting this out with the option/flag.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
nmarton-da: well it is only about logging actually, but I totally missed that one here, thanks :raised_hands:

ledger/participant-integration-api/BUILD.bazel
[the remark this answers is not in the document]
nmarton-da: @moritzkiefer-da Originally you seem to added that comment when "@maven//:com_oracle_database_jdbc_ojdbc8", was still a compile dep. Then it was moved to runtime dep. Now I am moving back here to compile dep.
Shall we keep the comment?
Can this cause problems with daml-assistant?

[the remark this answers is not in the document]
cocreature: If CI passes you can remove this. In the past we did runtime detection in a few places where the presence of the oracle lib affected the available backends and due to the way jars are merged in the sdk that meant that sandbox pulling in the oracle jar resulted in the json API making the oracle backend available in the community edition. We switched a few months ago to hardcoding the available backends so this is no longer an issue.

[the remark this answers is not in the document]
nmarton-da: yeah something like that happening here as well, thanks @cocreature, removing TODO

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -522,6 +524,15 @@ object Config {
             "Experimental buffer for Ledger API streaming queries. Should not be used in production."
           )
           .action((_, config) => config.copy(enableInMemoryFanOutForLedgerApi = true))
+
+        // TODO ha: remove after stable
+        opt[Unit]("index-ha-unsafe")
```
meiersi-da, line 529:
```suggestion
        opt[Unit]("experimental-index-ha")
```
We don't want customers reading `unsafe` in our help messages ;-)

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -61,26 +67,37 @@ object ParallelIndexerFactory {
...
+              ExecutionContext.fromExecutorService(
+                Executors.newFixedThreadPool(
+                  1,
+                  new ThreadFactoryBuilder().setNameFormat(s"ha-coordinator-%d").build,
+                )
+              )
+            )
+            .map(
+              HaCoordinator.databaseLockBasedHaCoordinator(
+                dataSource = storageBackend.createDataSource(jdbcUrl),
```
mziolekda, line 88: Please comment why we need two data sources

meiersi-da (commented):
Only cosmetic changes. Feel free to postpone.

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
```diff
@@ -0,0 +1,64 @@
...
+    },
+    periodMillis,
+    periodMillis,
+  )
+
+  // This is a cruel approach for ensuring single threaded usage of the mainConnection.
+  // In theory this could have been made much more efficient: not enqueueing for a check of it's own,
+  // but collecting requests, and replying in batches.
+  // Although experiments show approx 1s until a full connection pool is initialized at first
+  // (the peak scenario) which should be enough, and which can leave this code very simple.
```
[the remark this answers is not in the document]
nmarton-da: I am little confused, all this is about the check() method being in a synchronized block, so about the very implementation which comes after this comment. Or do I miss something?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
nmarton-da: I was merely following the convention with the other feature flags we have currently for the indexer. Like the suggestion though, did it!

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -61,26 +67,37 @@ object ParallelIndexerFactory {
...
-          metrics = metrics,
-          connectionAsyncCommitMode = DbType.AsynchronousCommit,
-        )
       toDbDto = backend.UpdateToDbDto(
         participantId = participantId,
         translation = translation,
         compressionStrategy = compressionStrategy,
       )
+      haCoordinator <-
+        if (storageBackend.dbLockSupported && haConfig.enable)
```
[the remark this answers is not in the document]
nmarton-da: Well if AFAIK we want to enable HA everywhere we can especially so that it also ensures safety with async commit as well.
Also warn here if the DB is not supporting locking would leave all the H2 participant will start with a warning. I am not sure that is called for.
What do you think @rautenrieth-da @mziolekda @meiersi-da ?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/HikariJdbcConnectionProvider.scala
```diff
@@ -13,48 +13,40 @@ import com.daml.ledger.resources.{Resource, ResourceContext, ResourceOwner}
...
   override def acquire()(implicit context: ResourceContext): Resource[HikariDataSource] = {
     val config = new HikariConfig
-    val dbType = DbType.jdbcType(jdbcUrl)
-
+    config.setDataSource(dataSource)
     config.setJdbcUrl(jdbcUrl)
-    config.setDriverClassName(dbType.driver)
-    config.addDataSourceProperty("cachePrepStmts", "true")
-    config.addDataSourceProperty("prepStmtCacheSize", "128")
-    config.addDataSourceProperty("prepStmtCacheSqlLimit", "2048")
```
rautenrieth-da, line 48: These are the properties that have no effect in Postgres/H2/Oracle, right? Because you are removing them without replacement.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/HikariJdbcConnectionProvider.scala
```diff
@@ -13,48 +13,40 @@ import com.daml.ledger.resources.{Resource, ResourceContext, ResourceOwner}
...
+    extends ResourceOwner[DataSource] {
 
   private val logger = ContextualizedLogger.get(this.getClass)
 
   override def acquire()(implicit context: ResourceContext): Resource[HikariDataSource] = {
     val config = new HikariConfig
-    val dbType = DbType.jdbcType(jdbcUrl)
-
+    config.setDataSource(dataSource)
     config.setJdbcUrl(jdbcUrl)
```
rautenrieth-da, line 42: Do you need to set both the dataSource and the jdbcUrl?

rautenrieth-da (commented):
I have reviewed the first few commits in detail. I have not found any issue with the changed implementation of connection acquisition and async commit handling (but we should also rely on tests because this refactoring is hard to review).

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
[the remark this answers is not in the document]
rautenrieth-da: Do we want to enable HA everywhere by default? Or automatically choose HA mode depending on the backend?

If I understand correctly, then we can't provide HA if the backend doesn't support advisory locks. IMO we should crash if a participant was started in HA mode and the backend doesn't support HA. And if we choose the HA mode automatically, we should log the chosen HA mode at info level.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/HikariJdbcConnectionProvider.scala
mziolekda: As far as I understand these only make sense in MySQL, so they have been dead for some time now.

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
```diff
@@ -0,0 +1,183 @@
...
+      scheduler: Scheduler,
+      haConfig: HaConfig,
+  )(implicit loggingContext: LoggingContext): HaCoordinator = {
+    implicit val ec: ExecutionContext = executionContext
+
+    val indexerLockId = storageBackend.lock(haConfig.indexerLockId)
+    val indexerWorkerLockId = storageBackend.lock(haConfig.indexerWorkerLockId)
+    val preemptableSequence = PreemptableSequence(scheduler)
+
+    new HaCoordinator {
```
mziolekda, line 95: Thanks for having made it clearer!

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
```diff
@@ -0,0 +1,183 @@
...
+          acquireLock(connection, indexerLockId, LockMode.Exclusive)
+          ()
+        }
+
+        preemptableSequence.executeSequence { sequenceHelper =>
+          import sequenceHelper._
+          logger.info("Starting databaseLockBasedHaCoordinator")
+          for {
+            mainConnection <- go[Connection](dataSource.getConnection)
+            _ = logger.info("Step 1: creating main-connection - DONE")
```
mziolekda, line 118: If we hit the kill switch after this line main connection will never be closed

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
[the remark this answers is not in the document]
mziolekda: Yes, that is correct. Simon was alluding to the fact that comment explicitly mentions mainConnection, whereas in fact, polling checker could just as well be used for checking the temperature:
```ensuring single threaded usage of the mainConnection```

File: ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala [hunk 25]
```diff
@@ -121,16 +139,57 @@ object ParallelIndexerFactory {
...
-          .executeSql(metrics.daml.parallelIndexer.initialization)(storageBackend.initialize)
-          .map(initialized =>
-            ingest(initialized.lastEventSeqId.getOrElse(0L))(
-              readService.stateUpdates(beginAfter = initialized.lastOffset)
+      def subscribe(resourceContext: ResourceContext)(readService: ReadService): Handle = {
+        implicit val rc: ResourceContext = resourceContext
+        implicit val ec: ExecutionContext = resourceContext.executionContext
+        implicit val matImplicit: Materializer = mat
+        haCoordinator.protectedExecution { connectionInitializer =>
+          val killSwitchCaptor = Promise[KillSwitch]()
```
mziolekda, line 147:
```suggestion
          val killSwitchPromise = Promise[KillSwitch]()
```
As there s an actual class under that name I would use the word promise to aid readability.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -522,6 +524,15 @@ object Config {
             "Experimental buffer for Ledger API streaming queries. Should not be used in production."
           )
           .action((_, config) => config.copy(enableInMemoryFanOutForLedgerApi = true))
+
+        // TODO ha: remove after stable
+        opt[Unit]("experimental-index-ha")
```
mziolekda, line 529:
```suggestion
        opt[Unit]("index-ha-unsafe")
```
In accordance with the hitherto used naming convention

ledger/sandbox-classic/src/test/suite/scala/platform/sandbox/stores/ledger/sql/SqlLedgerSpecAppendOnly.scala
```diff
@@ -211,23 +210,6 @@ final class SqlLedgerSpecAppendOnly
         ledger.currentHealth() should be(Healthy)
       }
     }
-
-    /** Workaround test for asserting that PostgreSQL asynchronous commits are disabled in
```
mziolekda, line 215: Why are we getting rid of this test completely rather than replacing it with something equivalent? This would merit a TODO, wouldn't it?

mziolekda (approved):

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
nmarton-da: I would suggest to wait until we have async commit really coming for Oracle (if ever).

What I see in the code is that the DbType.AsyncCommitMode is already very much specialized for postgres async commit (so it is not abstracted away, only these configurations are not implemented for none of the backends except postgres). It is on, off, and local. (Local is expected to be needed by canton) Just for contrast in oracle we have two flags: immediate/batch, and wait/no wait. For me it is not trivial we can draw a one-size-fits-all abstraction over these. But this we will only know, if we will have the oracle usage. Until then I think it should be alright to take a postgre-style async commit config DbType.AsyncCommitMode and mapping this to a postgres specific configuration for the backend.

This concern was raised multiple times, so I would like to emphasize this here again: this is exposed on the Interface because this is exactly what we wish to configure, and only for one type of database-type. This is kind of an exception from the rule here. I was asking myself the question: what if we want to configure db specific datasource properties? Doing this with jdbc-rul properties? Doing this by passing untyped properties down to the driver? Or making them explicit and type safe and expose them on the interface? I was opting for the latter. Happy to continue this discussion :eyes:

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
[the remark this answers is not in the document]
nmarton-da: I altered the comment accordingly for more clarity

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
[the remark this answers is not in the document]
nmarton-da: @rautenrieth-da AFAIK we want to enable this by default. (because it is mainly to provide guards for mutual exclusion, thus not letting multiple indexer start using the same database).

`haConifig.enable` is the feature flag. That is planned to be removed as with the HA epic's cleanup

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -75,18 +75,7 @@ private[backend] object PostgresStorageBackend
...
-    try {
-      statement.execute()
-      ()
-    } finally {
-      statement.close()
-    }
-  }
-
-  override val duplicateKeyError: String = "duplicate key"
+  val duplicateKeyError: String = "duplicate key"
```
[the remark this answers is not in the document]
nmarton-da: no clue, readded :+1:

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -234,6 +236,38 @@ object EventStorageBackend {
...
+  object PgSynchronousCommitValue {
+    case object On extends PgSynchronousCommitValue("on")
+    case object Off extends PgSynchronousCommitValue("off")
+    case object RemoteWrite extends PgSynchronousCommitValue("remote_write")
+    case object RemoteApply extends PgSynchronousCommitValue("remote_apply")
+    case object Local extends PgSynchronousCommitValue("local")
+
+    def apply(s: String): PgSynchronousCommitValue = s.toLowerCase match {
```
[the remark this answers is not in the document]
nmarton-da: good point, I planned to wire the Async commit to ledger configuration, but that changed later, so this is not used anymore. removed it.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/InitHookDataSourceProxy.scala
```diff
@@ -0,0 +1,213 @@
...
+import com.daml.platform.store.backend.common.InitHookDataSourceProxy._
+
+private[backend] case class InitHookDataSourceProxy(
+    delegate: DataSource,
+    initHook: Connection => Unit,
+)(implicit loggingContext: LoggingContext)
+    extends DataSource {
+  override def getConnection: Connection = {
+    val connectionId =
+      List.fill(8)(scala.util.Random.nextPrintableChar()).mkString // TODO FIXME maybe not needed
```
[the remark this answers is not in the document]
nmarton-da: this whole thing is removed already in later steps

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/HikariJdbcConnectionProvider.scala
nmarton-da: yes

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/HikariJdbcConnectionProvider.scala
nmarton-da: of course not! so nice, more dependencies removed :tada:

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
nmarton-da: I think it might be still ok, because:
* The go wrapper checks the killswitch state before the wrapped future starts. (putting it differently: it does not let the continuation start in case shutdown is ongoing)
* the line 118 is part of the for comprehension, and in this case it will be hit after the go wrapped future already has it's result. At this point we don't care about the kill-switch
* at line 119 the release function is getting registered, but here we also do not care about kill-switches neither, so we just register the release function unaware of an ongoing shutdown process
* at line 124 we start the retry mechanism to grab the indexer main lock. Here we care about the killswitch. Here we will start the shutdown process, but at this point the release function is already registered, so we will release the connection.

With all that said this should be part of unit testing

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: good point. changed

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
nmarton-da: @meiersi-da @mziolekda 
I am good with both approaches. Can you agree on one? Then let me know, and I will change accordingly.

ledger/sandbox-classic/src/test/suite/scala/platform/sandbox/stores/ledger/sql/SqlLedgerSpecAppendOnly.scala
nmarton-da: I am not so sure. 
What happened here is a major change in how we configure asynchronous commit. 
Previously the general asynchronous commit configuration was overridden in a couple of cases (this included), so we still configure synchronous commit even if configuration was asynchronous commits.
Currently we do not care at all at all the places about asynchronous commit configuration (everything will be just default eg sync commit), except from one dedicated place: the parallel ingestion.
With this change it seemed to me unnecessary and very hard to test something like this.

nmarton-da: What do you think?

nmarton-da:
> I have reviewed the first few commits in detail. I have not found any issue with the changed implementation of connection acquisition and async commit handling (but we should also rely on tests because this refactoring is hard to review).

@mziolekda @meiersi-da We cannot really rely on tests regarding the async commit, because we do not have such tests which would capture the problematic theoretical scenarios. If you need more certainty on the async commit aspect, I would recommend to invite either @tudor-da or/and @oliverse-da to review. They are experts in how async commit was implemented before, and why. On any rate please give me a definitive go/no-go regarding the async commit angle. I won't merge until, since that aspect won't be behind a feature flag.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
meiersi-da: @nmarton-da : your rationale makes sense to me.

IMHO, the code would be totally fine with the Pg-specific option-only *provided* that option is annotated with comment containing the rationale you give here. I have a hunch that such a comment would even have sufficed to not have a discussion on this topic on the PR at all.

@mziolekda @rautenrieth-da what do you think?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
meiersi-da: @mziolekda is the owner of the code, so we go with his.

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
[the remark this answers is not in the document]
meiersi-da: The new comment is great. Thank you!

ledger/sandbox-classic/src/test/suite/scala/platform/sandbox/stores/ledger/sql/SqlLedgerSpecAppendOnly.scala
mziolekda: Sounds good

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
mziolekda: Yes, let's add a code comment and be done with the subject

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
mziolekda: Let's add a TODO then

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
nmarton-da: Good idea with gaining more clarity this-regard in code @meiersi-da , I added some comments, and also restructured the DataSourceConfig, so it reflects this particular nature: configurations regarding specific StorageBackend implementations.
Also moved the postgres specific part into the backend.postgresql package for more clarity.

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
nmarton-da: I just added a line to the respective Jira task to capture that testing aspect as well

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
nmarton-da: renamed

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
```diff
@@ -0,0 +1,192 @@
...
+
+      override def go[T](body: => T): Future[T] = goF[T](Future(body))
+
+      override def retry[T](waitMillisBetweenRetries: Long, maxAmountOfRetries: Long)(
+          body: => T
+      ): Future[T] =
+        go(body).transformWith {
+          // since we check countdown to 0, starting from negative means unlimited retries
+          case Failure(ex) if maxAmountOfRetries == 0 =>
+            logger.info(
```
tudor-da, line 129: Shouldn't the level here be `warn`?

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -108,12 +127,11 @@ object ParallelIndexerFactory {
                 .map(_ -> System.nanoTime())
             )
               .map(_ => ())
-              .keepAlive(
+              .keepAlive( // TODO ha: remove as stable. This keepAlive approach was introduced for safety with async commit. This is still needed until HA is mandatory for Postgres to ensure safety with async commit.
```
tudor-da, line 130: What removes the need of this keepAlive here? is it the `PollingChecker` from the `HaCoordinator`? 
If yes, it might be good to mention it here

tudor-da (approved):
LGTM. Left two minor comments regarding logging and comments

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
nmarton-da: The keepAlive's main purpose here is to let the indexing die if connectivity to DB is lost, to prevent the corner case scenario with async commit, and unnoticed DB restarts.
With postgres+async commit+ha DB restarts cannot go unnoticed, because if the main connection (which holds the main indexer lock) breaks, we loose the indexer lock, and shut down indexing as soon as detected.
(Also we cannot initialize an additional connection to database, if we cannot verify that the main connection still holds the lock.)
I'll make a note about this in the TODO

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PreemptableSequence.scala
nmarton-da: yes, since we switched another place logging to warn which was about possibly non-transient error, I think here it is called for as well. changed

PR #10236, opened 2021-07-09
es: switch to persistent nodes

garyverhaegen-da:
A few small tweaks, but the most important change is giving up on preemptible instances (for now at least), because I saw GCP kill 8 out of the 10 nodes at exactly the same time and I can't really expect the cluster to survive that.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #10238, opened 2021-07-11
Adds Unit tests for parallel-ingestion [DPP-455]

nmarton-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
@@ -1,0 +1,169 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.indexer.parallel
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+import com.daml.ledger.api.testing.utils.AkkaBeforeAndAfterAll
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.concurrent.duration.FiniteDuration
+
+class BatchingParallelIngestionPipeSpec
+    extends AsyncFlatSpec
+    with Matchers
+    with AkkaBeforeAndAfterAll {
+
+  // AsyncFlatSpec is with serial execution context
+  private implicit val ec: ExecutionContext = system.dispatcher
+
+  private val input = Iterator.continually(util.Random.nextInt()).take(100).toList
+
+  it should "end the stream successfully in a happy path case" in {
+    runPipe().map { case (ingested, ingestedTail, err) =>
+      err shouldBe empty
+      ingested.sortBy(_._1) shouldBe input.map(_.toString).zipWithIndex.map { case (s, i) =>
+        (i + 1, s)
+      }
+      ingestedTail.last shouldBe 100
+    }
+  }
+
+  it should "terminate the stream upon error in input mapper" in {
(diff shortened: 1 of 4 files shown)


ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
```diff
@@ -0,0 +1,169 @@
...
+  it should "hold the stream if a single ingestion takes too long" in {
+    runPipe(
+      ingesterHook = batch => {
+        if (batch.head == 21) Thread.sleep(1000)
+      },
+      timeout = FiniteDuration(100, "milliseconds"),
+    ).map { case (ingested, ingestedTail, err) =>
+      err should not be empty
+      err.get.getMessage shouldBe "timed out"
+      ingested.size shouldBe 25
```
rautenrieth-da, line 88: Why is this 25?

The test times out the 21st "batch", but the assertion is done on the number of "state updates" ingested. Does this test depend on the timing how updates are batched together?

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
```diff
@@ -0,0 +1,169 @@
...
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+import com.daml.ledger.api.testing.utils.AkkaBeforeAndAfterAll
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.concurrent.duration.FiniteDuration
+
+class BatchingParallelIngestionPipeSpec
```
rautenrieth-da, line 15: Is it worth adding tests that check whether a slow ingestion leads to full batches?

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/ParallelIndexerFactorySpec.scala
```diff
@@ -0,0 +1,311 @@
...
+        someParty,
+        someParty,
+        someParty,
+        someParty,
+      ),
+      batchSize = 3,
+      averageStartTime = 1000000001,
+      offsets = Vector("00", "01", "02").map(offset),
+    )
+    actual.copy(batch = Vector.empty) shouldBe expected.copy(batch = Vector.empty)
```
rautenrieth-da, line 158: Why do you define the value of `expected.batch` above if you discard it here?

rautenrieth-da (approved):

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
nmarton-da: no, as soon as the 21..25 batch is blocked, we still process the 26..30 batch, and only then the mapAsync stage is forced to block as well

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/BatchingParallelIngestionPipeSpec.scala
nmarton-da: I don't think so. Firstly it wont necessarily lead to full batches. The groupedWithin will take whatever it has if the within part breached. Secondly: I think that would be rather testing the framework.

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/parallel/ParallelIndexerFactorySpec.scala
nmarton-da: o damn, because I forgot to test whether the expected has the same entries as the result :man_facepalming: Since this is merged I will make note and fix this in an upcoming PR

PR #10243, opened 2021-07-12
LF: Make DarReader ZipEntries immulatble

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
+++ b/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
@@ -19,7 +19,8 @@ class GenDarReader[A](parseDalf: Bytes => Try[A]) {
 
   /** Reads an archive from a File. */
   def readArchiveFromFile(darFile: File): Try[Dar[A]] =
-    readArchive(darFile.getName, new ZipInputStream(new FileInputStream(darFile)))
+    bracket(Try(new ZipInputStream(new FileInputStream(darFile))))(zis => Try(zis.close))
+      .flatMap(readArchive(darFile.getName, _))
 
   /** Reads an archive from a ZipInputStream. The stream will be closed by this function! */
   def readArchive(
@@ -57,17 +58,13 @@ class GenDarReader[A](parseDalf: Bytes => Try[A]) {
       darStream: ZipInputStream,
       entrySizeThreshold: Int,
   ): Try[ZipEntries] =
-    bracket(Try(darStream))(is => Try(is.close()))
-      .flatMap(is =>
-        Try(
-          Iterator
-            .continually(darStream.getNextEntry)
-            .takeWhile(_ != null)
-            .map(entry => slurpWithCaution(entry.getName, is, entrySizeThreshold))
-            .toMap
-        )
-      )
-      .map(ZipEntries(name, _))
+    Try(
+      Iterator
+        .continually(darStream.getNextEntry)
+        .takeWhile(_ != null)
+        .map(entry => slurpWithCaution(entry.getName, darStream, entrySizeThreshold))
+        .toMap
+    ).map(ZipEntries(name, _))
 
   private[this] def parseAll(getPayload: String => Try[Bytes])(names: List[String]): Try[List[A]] =
     sequence(names.map(parseOne(getPayload)))
--- a/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
+++ b/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
@@ -92,31 +92,8 @@ object GenDarReader {
     private[GenDarReader] def readDalfNames: Try[Dar[String]] =
       bracket(get(ManifestName).map(_.toInputStream))(is => Try(is.close()))
         .flatMap(DarManifestReader.dalfNames)
-        .recoverWith { case NonFatal(e1) =>
-          findLegacyDalfNames.recoverWith { case NonFatal(_) =>
-            Failure(Error.InvalidDar(this, e1))
-          }
-        }
-
-    // There are three cases:
-    // 1. if it's only one .dalf, then that's the main one
-    // 2. if it's two .dalfs, where one of them has -prim in the name, the one without -prim is the main dalf.
-    // 3. parse error in all other cases
-    private[this] def findLegacyDalfNames: Try[Dar[String]] = {
-      val dalfs: List[String] = entries.keys.filter(isDalf).toList
-
-      dalfs.partition(isPrimDalf) match {
-        case (List(prim), List(main)) => Success(Dar(main, List(prim)))
-        case (List(prim), Nil) => Success(Dar(prim, List.empty))
-        case (Nil, List(main)) => Success(Dar(main, List.empty))
-        case _ => Failure(Error.InvalidLegacyDar(this))
-      }
-    }
+        .recoverWith { case NonFatal(e1) => Failure(Error.InvalidDar(this, e1)) }
   }
-
-  private[this] def isDalf(s: String): Boolean = s.toLowerCase.endsWith(".dalf")
-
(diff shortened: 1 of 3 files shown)


daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -98,53 +83,20 @@ object GenDarReader {
...
-        case (List(prim), List(main)) => Success(Dar(main, List(prim)))
-        case (List(prim), Nil) => Success(Dar(prim, List.empty))
-        case (Nil, List(main)) => Success(Dar(main, List.empty))
-        case _ => Failure(Error.InvalidLegacyDar(this))
-      }
-    }
-
-    private def isDalf(s: String): Boolean = s.toLowerCase.endsWith(".dalf")
-
-    private def isPrimDalf(s: String): Boolean = s.toLowerCase.contains("-prim") && isDalf(s)
```
remyhaemmerle-da, line 142: mostly dead code.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -30,66 +25,56 @@ class GenDarReader[A](
...
-        output.write(buffer, 0, n)
-        if (output.size >= entrySizeThreshold) throw Error.ZipBomb()
-      }
-      (output.size.toLong, new ByteArrayInputStream(output.toByteArray))
+  ): (String, Bytes) = {
+    val buffSize = 4 * 1024 // 4k
+    val buffer = new Array[Byte](buffSize)
+    var output = Bytes.Empty
+    Iterator.continually(zip.read(buffer)).takeWhile(_ >= 0).foreach { size =>
+      output = output ++ Bytes.fromByteArray(buffer, 0, size)
```
remyhaemmerle-da, line 50: concatenation is constant time.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -30,66 +25,51 @@ class GenDarReader[A](
...
-                }
-            }
-          )
-        case None => accT
-      }
-
-    go(Success(Map.empty)).map(ZipEntries(name, _))
-  }
+  ): Try[ZipEntries] =
+    bracket(Try(darStream))(is => Try(is.close()))
```
cocreature, line 56: I’m not sure this is a good idea. imho a function should not close a stream it didn’t open and we didn’t open the zip stream here.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -98,53 +78,20 @@ object GenDarReader {
   private val ManifestName = "META-INF/MANIFEST.MF"
   private[archive] val EntrySizeThreshold = 1024 * 1024 * 1024 // 1 GB
 
-  private[archive] case class ZipEntries(name: String, entries: Map[String, (Long, InputStream)]) {
+  private[archive] case class ZipEntry(size: Long, getStrean: () => InputStream)
```
cocreature, line 81:
```suggestion
  private[archive] case class ZipEntry(size: Long, getStream: () => InputStream)
```

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Bytes.scala
```diff
@@ -51,14 +50,43 @@ object Bytes {
...
+    new Bytes(ByteString.copyFrom(a, offset, size))
 
   def fromByteBuffer(a: ByteBuffer): Bytes =
     new Bytes(ByteString.copyFrom(a))
 
   def fromInputStream(a: InputStream): Bytes =
     new Bytes(ByteString.readFrom(a))
 
+  // reads at most `len` bytes, does not close `a`
+  def fromInputStream(a: InputStream, len: Int): Bytes = {
```
cocreature, line 65: this seems like a fair amount of complexity and potential source of bugs. do the performance gains matter?

cocreature (approved):
seems sensible, thanks!

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Bytes.scala
remyhaemmerle-da: I reverted the optimization. Do you prefer like that.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
remyhaemmerle-da: actually not.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
remyhaemmerle-da: 967146aec86b70968347210b47261bd3c5006779

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
remyhaemmerle-da: 3068bf08e9703c6a90a296a57a527045a76b687a

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -98,35 +79,30 @@ object GenDarReader {
...
-      parseDalfNamesFromManifest(readDalfNamesFromManifest).recoverWith { case NonFatal(e1) =>
-        findLegacyDalfNames().recoverWith { case NonFatal(_) =>
-          Failure(Error.InvalidDar(this, e1))
+    private[GenDarReader] def readDalfNames: Try[Dar[String]] =
+      bracket(get(ManifestName).map(_.toInputStream))(is => Try(is.close()))
+        .flatMap(DarManifestReader.dalfNames)
+        .recoverWith { case NonFatal(e1) =>
+          findLegacyDalfNames.recoverWith { case NonFatal(_) =>
+            Failure(Error.InvalidDar(this, e1))
+          }
```
remyhaemmerle-da, line 98: @cocreature Do we have to handle dar without Manifest ? 
If no, let me drop this code, otherwise should we have test for that (It seems we currently do not have any) ?

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Bytes.scala
cocreature: thx seems simpler! And just to be clear, I’m happy to make an optimization here if it provides a benefit just not without some measurements.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
cocreature: The compiler will always add a manifest and I’d generally consider anything not produced by the compiler to be unsupported so should be fine to drop it.

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Bytes.scala
remyhaemmerle-da: :+1:

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
remyhaemmerle-da: Great !

remyhaemmerle-da: :fire:

PR #10249, opened 2021-07-12
LF: Deprecate com.daml.lf.data.TryOps.Bracket

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/data/src/test/scala/com/digitalasset/daml/lf/data/TryOpsTest.scala
```diff
@@ -1,87 +0,0 @@
...
-package com.daml.lf.data
-
-import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AnyWordSpec
-import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
-import com.daml.lf.data.TryOps.Bracket.bracket
-
-import scala.util.{Failure, Success, Try}
-
-class TryOpsTest extends AnyWordSpec with Matchers with ScalaCheckDrivenPropertyChecks {
```
S11001001, line 13: Alternatively, you could deprecate this class.

S11001001 (approved):

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/TryOps.scala
```diff
@@ -17,6 +17,7 @@ private[daml] object TryOps {
       b <- tb
     } yield f(a, b)
 
+  @deprecated("use scala.util.Using", since = "1.16.0")
```
stefanobaghino-da, line 20: I think this is used only internally, any reason to leave it around?

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
```diff
@@ -90,7 +88,7 @@ object GenDarReader {
     }
 
     private[GenDarReader] def readDalfNames: Try[Dar[String]] =
-      bracket(get(ManifestName).map(_.toInputStream))(is => Try(is.close()))
+      get(ManifestName)
```
stefanobaghino-da, line 91: Does this change the guarantees given to the callers with regards to cleaning up?

daml-lf/data/src/test/scala/com/digitalasset/daml/lf/data/TryOpsTest.scala
```diff
@@ -1,87 +0,0 @@
-// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
cocreature, line 1: if we leave it around we should imho also leave the tests around.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
cocreature: There is no resource to cleanup. Previously, we opened the stream here and closed it here. Now `DarManifestReader.dalfNames` opens and closes the stream.

cocreature (approved):
Thanks!

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DarReader.scala
remyhaemmerle-da: `get(ManifestName)` returns a immutable value no need for clean up here anymore.
The creation of the `InputStream` (with its cleanup is done) in `DarManifestReader.scala` line 18.

daml-lf/data/src/test/scala/com/digitalasset/daml/lf/data/TryOpsTest.scala
remyhaemmerle-da: :+1:

daml-lf/data/src/test/scala/com/digitalasset/daml/lf/data/TryOpsTest.scala
remyhaemmerle-da: :+1:

ledger-service/jwt/BUILD.bazel
```diff
@@ -34,24 +38,13 @@ da_scala_library(
...
     ],
     deps = [
-        "//daml-lf/data",
-        "@maven//:com_auth0_java_jwt",
-        "@maven//:com_auth0_jwks_rsa",
-        "@maven//:com_google_guava_guava",
-        "@maven//:org_slf4j_slf4j_api",
+        ":jwt",
     ],
 )
```
remyhaemmerle-da, line 49: I am not sue why there is both the source 
```srcs = glob(["src/main/scala/**/*.scala"]),```
and the library
```
 ":jwt",
```

cocreature: looks broken, the library should exclude the `Main` class and then the binary should depend on the lib and the source for the Main class. But your change seems fine for now.

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/TryOps.scala
remyhaemmerle-da: I just prefer to deprecate anything which is public.

PR #10253, opened 2021-07-13
Separate traces from warnings in engine.

sofiafaro-da:
I decided to separate the engine warnings from the tracelog after all,
because I think it will make testing and maintenance easier in the
long run. For example, it makes it easier to display the warnings
separate from the traces, so they don't get buried in the traces.

Part of #9947, follow up from #10179

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -271,6 +275,11 @@ final class Conversions(
     builder.setMessage(msgAndLoc._1).build
   }
 
+  def convertSWarningMessage(msg: String): proto.WarningMessage = {
```
remyhaemmerle-da, line 278:
```suggestion
  private[this] def convertSWarningMessage(msg: String): proto.WarningMessage = {
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/WarningLog.scala
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf.speedy
+
+import org.slf4j.Logger
+import scala.collection.mutable.ArrayBuffer
+
+private[lf] final case class WarningLog (logger: Logger) {
```
remyhaemmerle-da, line 9: I dislike using `case class` for not mutable data. 
Since you instantiate the object only once, I think you can remove the `case` keyword here  and use a `new` there.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/WarningLog.scala
```diff
@@ -0,0 +1,20 @@
...
+  private val buffer = new ArrayBuffer[String](initialSize = 10)
```
remyhaemmerle-da, line 10:
```suggestion
  private[this] val buffer = new ArrayBuffer[String](initialSize = 10)
```

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
remyhaemmerle-da: Does this need to be public ?

remyhaemmerle-da (commented):
Scala looks fine for me.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/WarningLog.scala
sofiafaro-da: Done, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/WarningLog.scala
```diff
@@ -0,0 +1,19 @@
...
+
+package com.daml.lf.speedy
+
+import org.slf4j.Logger
+import scala.collection.mutable.ArrayBuffer
+
+private[lf] final class WarningLog(logger: Logger) {
+  private[this] val buffer = new ArrayBuffer[String](initialSize = 10)
+  def add(message: String): Unit = {
+    if (logger.isWarnEnabled) {
```
cocreature, line 12: I don’t think you need this. It makes sense for debug logs where you construct potentially expensive messages but here the string is already materialized and we expect warnings to always be on.

cocreature (approved):
Nice!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/WarningLog.scala
sofiafaro-da: I'll remove it, thanks!

PR #10256, opened 2021-07-13
Remove trace_context field from Ledger API [KVL-1021]

miklos-da:
CHANGELOG_BEGIN
* **[Integration Kit]** Removed `trace_context` field from Ledger API and its bindings as we now have trace context propagation support via gRPC metadata. If you are constructing or consuming Ledger API requests or responses directly, you may need to update your code.

CHANGELOG_END

In a follow-up PR we'll be adding documentation on how the current trace context propagation works.

### Pull Request Checklist

trigger the build.


ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
```diff
@@ -45,12 +43,6 @@ message GetActiveContractsRequest {
   // In particular, setting the verbose flag to true triggers the ledger to include labels for record fields.
   // Optional
   bool verbose = 3;
-
-  // Server side tracing will be registered as a child of the submitted context.
-  // This field is a future extension point and is currently not supported.
```
stefanobaghino-da, line 50: The future extension point that could not.

ledger/ledger-api-common/BUILD.bazel
```diff
@@ -47,7 +47,6 @@ da_scala_library(
         "@maven//:io_dropwizard_metrics_metrics_core",
         "@maven//:io_grpc_grpc_netty",
         "@maven//:io_netty_netty_handler",
-        "@maven//:io_zipkin_brave_brave",
```
stefanobaghino-da, line 50: If we don't use it anywhere else, we can probably remove it from `//bazel-java-deps.bzl` as well.

stefanobaghino-da (approved):
LGTM, thanks. Just one question: IIRC, since we never actually used this value ever, this is simply omitted in the serialized Protobuf, so even removing this field does not cause any backward-compatibility issue of sort. Can you confirm what I'm saying, please?

miklos-da:
> LGTM, thanks. Just one question: IIRC, since we never actually used this value ever, this is simply omitted in the serialized Protobuf, so even removing this field does not cause any backward-compatibility issue of sort. Can you confirm what I'm saying, please?

Yes, that's correct.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
```diff
@@ -45,12 +43,6 @@ message GetActiveContractsRequest {
...
-  // Optional
-  TraceContext trace_context = 1000;
```
S11001001, line 52: Perhaps 1000 should be reserved across this PR; maybe it doesn't matter.

SamirTalwar-DA (approved):

cocreature:
Slightly orthogonal to this PR but do we have any docs as to how you should pass trace context via metadata?

miklos-da:
> Slightly orthogonal to this PR but do we have any docs/example code in our repo as to how you should pass trace context via metadata?

Not yet, however, that'll be added in a separate PR (updated the description to state this).

ledger/ledger-api-common/BUILD.bazel
stefanobaghino-da: Addressed by https://github.com/digital-asset/daml/pull/10256/commits/751b11ec854d156409c0bb05de84f54c12103294 🙇🏻

ledger-api/grpc-definitions/com/daml/ledger/api/v1/active_contracts_service.proto
miklos-da: Indeed, we could make these fields reserved, however, as we aren't running different versions of the participant server against the same application this shouldn't matter (i.e., even if the application specified a field with ID 1000 it'll be ignored at all times).

PR #10261, opened 2021-07-13
participant-integration-api: Move `SeedService` here. [KVL-1002]

SamirTalwar-DA:
I don't see a reason why it's part of the participant state API, and it definitely doesn't need to change between v1 and v2.

### Changelog

- **[Integration Kit]** The class `SeedService` has been moved from the _participant-state_ Maven package to the _participant-integration-api_ Maven package, under the Java package name `com.daml.platform.apiserver` to reflect its usage by the API server, not the participant state API. If you use this class directly, you will need to change your imports.

### Pull Request Checklist

trigger the build.


miklos-da (approved):
Thanks for moving `SeedService` out from `participant-state`!

ledger-service/http-json-testing/src/main/scala/com/daml/http/HttpServiceTestFixture.scala
```diff
@@ -29,15 +30,15 @@ import com.daml.ledger.api.refinements.ApiTypes.ApplicationId
...
 import com.daml.ledger.api.tls.TlsConfiguration
 import com.daml.ledger.api.v1.{value => v}
-import com.daml.ledger.client.{LedgerClient => DamlLedgerClient}
 import com.daml.ledger.client.configuration.{
   CommandClientConfiguration,
   LedgerClientConfiguration,
   LedgerIdRequirement,
 }
-import com.daml.ledger.participant.state.v1.SeedService.Seeding
+import com.daml.ledger.client.{LedgerClient => DamlLedgerClient}
```
S11001001, line 38: Please don't use import cleaning in any way. We don't have an idempotent approach.

S11001001 (approved):

ledger-service/http-json-testing/src/main/scala/com/daml/http/HttpServiceTestFixture.scala
SamirTalwar-DA: I know it's not always idempotent, but I did verify that reordering these imports won't affect any behavior.

There are large parts of the codebase where import ordering is standard (e.g. pretty much everything under _ledger_), so I'd rather not turn it off. If you like, I can revert this file.

fabiotudone-da (approved):

SamirTalwar-DA:
Gonna merge in the interest of moving on, but please feel free to comment regardless and I'll address changes later.

ledger-service/http-json-testing/src/main/scala/com/daml/http/HttpServiceTestFixture.scala
S11001001: > I did verify that reordering these imports won't affect any behavior.

I don't think this is a good reason to add such hunks. I can do lots of import reordering that doesn't affect their behavior; even so, I'm still gumming up the review (and potentially forcing conflicts) with totally irrelevant diff lines.

> There are large parts of the codebase where import ordering is standard

I would say that it is more the case that they _look_ standard by the eye test, but do not actually match up to any specific standard. You will never reach an equilibrium with this approach, as things stand, but instead will add diff/conflict churn multiplied by however many people also do import reordering.

SamirTalwar-DA: That's fair. I agree that it's kind of a mess and not everyone follows the same conventions, uses the same version of the IDE, remembers to run the import reorganizer, or is in a situation where there is no equilibrium.

BRB gonna go write a tool to do this in a "standard" way.

S11001001: Alright, see you in a couple years 🤣

PR #10262, opened 2021-07-13
[Mutable cache] Resolve with full lookup on negative cache read-through lookups [DPP-501]

tudor-da:
When trying to resolve an input divulged contract for which the create is event is not present (i.e. the contract stakeholder is on a different participant), the mutable state cache will settle with a negative lookup and cache it. The current behavior prohibits recovery in this scenario, leading to continuous contract not found errors reported to the client.

This PR mitigates the problem by:
* Negative lookups (contracts not found) always result in a lookup with divulgence resolution `contractsReader.lookupActiveContractAndLoadArgument`. 
* Negative lookups are not cached and a warning is logged if encountered (after divulgence resolution). 
* Metrics added for counting divulgence and full lookups (for asserting the impact of more lookups due to more negative contract lookups).
* Metrics added for counting negative lookups after divulgence resolution.

**NOTE** _Due to always performing a second lookup after negative contract lookups, this change has a performance impact in applications that cause race conditions._ 

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
@@ -128,7 +128,9 @@ private[platform] class MutableCacheBackedContractStore(
         eventualValue = eventualValue.transformWith {
           case Success(NotFound) =>
             metrics.daml.execution.cache.readThroughNotFound.inc()
-            // Don't cache negative lookups
+            // We must not cache negative lookups by contract-id, as they can be invalidated by later divulgence events.
+            // This is OK from a performance perspective, as we do not expect uses-cases that require
+            // caching of contract absence or the results of looking up divulged contracts.
             Future.failed(ContractReadThroughNotFound(contractId))
           case result => Future.fromTry(result)
         },
@@ -161,11 +163,6 @@ private[platform] class MutableCacheBackedContractStore(
         // (the contract might have been divulged to the readers)
         // OR the contract was not found in the index
         //
-        // NOTE: The current implementation of mutable contract state cache provides an optimization aimed at performance-critical
-        //       applications that do not exercise this flow frequently (i.e. no negative contract lookups in command interpretation
-        //       or use of divulgence)
-        // TODO: Implement caching of divulged contracts for alleviating performance degradation in applications that
-        //       make use of divulgence OR applications that cause high numbers of negative contract lookups
         logger.debug(s"Checking divulgence for contractId=$contractId and readers=$readers")
         resolveDivulgenceLookup(contractStateValue, contractId, readers)
     }
@@ -390,7 +387,7 @@ private[platform] object MutableCacheBackedContractStore {
 
   final case class ContractReadThroughNotFound(contractId: ContractId) extends NoStackTrace {
     override def getMessage: String =
-      s"Contract not found for contract id $contractId. This is likely due to a race condition."
+      s"Contract not found for contract id $contractId. Hint: this could be due racing with a concurrent archival."
   }
 
   private[cache] class CacheIndex {
--- a/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
@@ -157,6 +157,15 @@ private[platform] class MutableCacheBackedContractStore(
       case Archived(stakeholders) if nonEmptyIntersection(stakeholders, readers) =>
         Future.successful(Option.empty)
       case contractStateValue =>
+        // This flow is exercised when the readers are not stakeholders of the contract
+        // (the contract might have been divulged to the readers)
+        // OR the contract was not found in the index
+        //
+        // NOTE: The current implementation of mutable contract state cache provides an optimization aimed at performance-critical
+        //       applications that do not exercise this flow frequently (i.e. no negative contract lookups in command interpretation
+        //       or use of divulgence)
+        // TODO: Implement caching of divulged contracts for alleviating performance degradation in applications that
+        //       make use of divulgence OR applications that cause high numbers of negative contract lookups
         logger.debug(s"Checking divulgence for contractId=$contractId and readers=$readers")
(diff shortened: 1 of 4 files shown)


meiersi-da:
@tudor-da : couldn't recovery happen via the processing of the divulgence event bringing the contract into scope?

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
```diff
@@ -107,13 +107,12 @@ private[platform] case class StateCache[K, V](cache: Cache[K, V], registerUpdate
               removeFromPending(key)
             }
             .getOrElse(logger.error(s"Pending updates tracker for $key not registered "))
-        },
-      (err: Throwable) => {
+        }
+      }
+      .recover { case err =>
```
mziolekda, line 112: Please note that now we recover all kinds of problems not just the new error type `ContractReadThroughNotFound`. I don't know if this is a problem though

ledger/participant-integration-api/src/test/suite/scala/platform/store/cache/MutableCacheBackedContractStoreSpec.scala
```diff
@@ -204,6 +208,37 @@ class MutableCacheBackedContractStoreSpec
...
+        succeed
+      }
+    }
+
+    "resort to resolveDivulgenceLookup on not found" in {
+      val spyContractsReader = spy(ContractsReaderFixture())
+      for {
+        store <- contractStore(cachesSize = 1L, spyContractsReader).asFuture
+        _ = store.cacheIndex.set(unusedOffset, 1L)
+        negativeLookup_cId6 <- store.lookupActiveContract(Set(bob), cId_7)
```
mziolekda, line 233:
```suggestion
        resolvedLookup_cId7 <- store.lookupActiveContract(Set(bob), cId_7)
```

ledger/participant-integration-api/src/main/scala/platform/store/cache/StateCache.scala
tudor-da: Indeed. First of all, the problems that might arise here that were previously "un-recovered" are internal to the cache. For these failures, it is actually better to recover and log since the failure of `putAsync` is not critical to the correctness guarantee of the mutable contract state cache. On the other hand, not recovering here would will stop command interpretation and report the failure to the client.

tudor-da:
> @tudor-da : couldn't recovery happen via the processing of the divulgence event bringing the contract into scope?

I'm not sure I'm following. Now we are not fetching divulgence events from neither of the update flows (read-throughs or the recency event stream). Can you please expand?

meiersi-da:
@tudor-da : I was essentially asking whether you've evaluated the alternative design option of also reading and processing the divulgence_events to keep the mutable state cache up to date.

As per our discussion: that option is technically more complex, and only required if we cannot afford reducing the scope of cacheable contract-id lookup calls as your PR does.

We've also discussed that you'd add some extra commentary to the code that only negative contract-id are not cached, and why they are not cached. Please ping me once you've done that for a review.

ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
```diff
@@ -124,7 +125,13 @@ private[platform] class MutableCacheBackedContractStore(
       _ <- contractsCache.putAsync(
         key = contractId,
         validAt = currentCacheSequentialId,
-        eventualValue = eventualValue,
+        eventualValue = eventualValue.transformWith {
+          case Success(NotFound) =>
+            metrics.daml.execution.cache.readThroughNotFound.inc()
+            // Don't cache negative lookups
```
meiersi-da, line 131:
```suggestion
            // We must not cache negative lookups by contract-id, as they can be invalidated by later divulgence events.
            // This is OK from a performance perspective, as we do not expect uses-cases that require
            // caching of contract absence or the results of looking up divulged contracts.
```

ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
```diff
@@ -149,31 +156,39 @@ private[platform] class MutableCacheBackedContractStore(
...
+      case contractStateValue =>
+        // This flow is exercised when the readers are not stakeholders of the contract
+        // (the contract might have been divulged to the readers)
+        // OR the contract was not found in the index
+        //
+        // NOTE: The current implementation of mutable contract state cache provides an optimization aimed at performance-critical
+        //       applications that do not exercise this flow frequently (i.e. no negative contract lookups in command interpretation
+        //       or use of divulgence)
+        // TODO: Implement caching of divulged contracts for alleviating performance degradation in applications that
+        //       make use of divulgence OR applications that cause high numbers of negative contract lookups
```
meiersi-da, line 168:
```suggestion
```
Move the note on the place where we don't cache negative lookups, as per my comment above.

I would not add the TODO, as it signals that we want to change that in the future, which we probably don't.

ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
```diff
@@ -373,6 +388,11 @@ private[platform] object MutableCacheBackedContractStore {
         "Cannot lookup the maximum ledger time for an empty set of contract identifiers"
       )
 
+  final case class ContractReadThroughNotFound(contractId: ContractId) extends NoStackTrace {
+    override def getMessage: String =
+      s"Contract not found for contract id $contractId. This is likely due to a race condition."
```
meiersi-da, line 393:
```suggestion
      s"Contract not found for contract id $contractId. Hint: this could be due racing with a concurrent archival."
```

meiersi-da: Weaken the text and clarify the kind of race condition to guide the user. In particular, avoid the bad connotation that the server itself has a race condition that leads to a bug.

meiersi-da (approved):
Good stuff @tudor-da ! Thank you!

PR #10263, opened 2021-07-13
[In-memory fan-out] Oracle support

tudor-da:
Enable Oracle support for in-memory fan-out
* Enable `--buffered-ledger-api-streams-unsafe` for Oracle conformance tests
* Add compatibility import for `rawTransactionEventParser` 
* Update Oracle SQL schema with null defaulting on nullable BLOB fields for `participant_events` view and regen SHA
* Add missing test suites to `JdbcLedgerDaoOracleAppendOnlySpec`

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10264, opened 2021-07-13
Correct so that in memory fan out enabled works on Oracle

dasormeter:
- Add implicit conversion and correct view defaulting for BLOB fields
- Add missing tests to oracle append only spec as it lagged behind postgres and h2
- Add additional ledger on sql oracle tests with different fan out config options, and for multi-participant IT
- Regen sql sha for oracle


oliverse-da (approved):
LGTM, thanks @dasormeter !

dasormeter:
superceded by https://github.com/digital-asset/daml/pull/10263

PR #10274, opened 2021-07-14
[JSON-API] Refactor Endpoints.scala to use path directives etc.

realvictorprm:
As the title states. This reduces the amount of code, makes it more readable and easier to edit and most importantly _no more code which needs formatter disabling comments_!

integration tests seem to pass locally, now let's see what the CI makes out of it (I still prefer my laptop to keep somewhat cool).

Will do some manual testing too, however I'm more or less confident that it should work if the it tests passed. 

changelog_begin
changelog_end

### Pull Request Checklist

- [x] Manually ensured that things work

trigger the build.

SIDE-NOTE:
Yeah much `apply` is used, however I wanted to reduce the huge amount of braces being created by the code.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -49,6 +49,8 @@ import com.daml.metrics.{Metrics, Timed}
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Directive
 
+import scala.annotation.nowarn
+
 class Endpoints(
     allowNonHttps: Boolean,
     decodeJwt: EndpointsCompanion.ValidateJwt,
@@ -72,6 +74,7 @@ class Endpoints(
   // Weird signature required, because it looks like Scala
   // does not see a difference between Future[HttpResponse] & other inner types.
   // Without it, it will complain that the function has been defined twice.
+  @nowarn("msg=parameter value ev in method toRoute is never used")
   def toRoute[A <: HttpResponse](res: Future[A])(implicit
       ev: A =:= HttpResponse
   ): Route = _ => res map Complete
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -75,7 +75,7 @@ class Endpoints(
   // does not see a difference between Future[HttpResponse] & other inner types.
   // Without it, it will complain that the function has been defined twice.
   @nowarn("msg=parameter value ev in method toRoute is never used")
-  def toRoute[A <: HttpResponse](res: Future[A])(implicit
+  private def toRoute[A <: HttpResponse](res: Future[A])(implicit
       ev: A =:= HttpResponse
   ): Route = _ => res map Complete
   // I think the arguments need to be lazy
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -4,7 +4,6 @@
 package com.daml.http
 
 import akka.NotUsed
-import akka.http.scaladsl.model.HttpMethods.{GET, POST}
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.{
   Authorization,
@@ -14,7 +13,7 @@ import akka.http.scaladsl.model.headers.{
   `X-Forwarded-Proto`,
 }
 import akka.http.scaladsl.server.Directives.extractClientIP
-import akka.http.scaladsl.server.{Rejection, RequestContext, Route, RouteResult}
+import akka.http.scaladsl.server.{Directive0, Route}
 import akka.http.scaladsl.server.RouteResult._
 import akka.stream.Materializer
 import akka.stream.scaladsl.{Flow, Source}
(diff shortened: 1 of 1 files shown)


stefanobaghino-da (approved):
Looks great, thanks.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -69,106 +70,72 @@ class Endpoints(
...
-  import Uri.Path._
+
+  // Weird signature required, because it looks like Scala
+  // does not see a difference between Future[HttpResponse] & other inner types.
+  // Without it, it will complain that the function has been defined twice.
+  @nowarn("msg=parameter value ev in method toRoute is never used")
+  def toRoute[A <: HttpResponse](res: Future[A])(implicit
+      ev: A =:= HttpResponse
+  ): Route = _ => res map Complete
+  // I think the arguments need to be lazy
```
cocreature, line 81: I thought the route dsl is already lazy? Have you tested this?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -69,106 +70,72 @@ class Endpoints(
...
   import util.ErrorOps._
-  import Uri.Path._
+
+  // Weird signature required, because it looks like Scala
+  // does not see a difference between Future[HttpResponse] & other inner types.
+  // Without it, it will complain that the function has been defined twice.
+  @nowarn("msg=parameter value ev in method toRoute is never used")
+  def toRoute[A <: HttpResponse](res: Future[A])(implicit
+      ev: A =:= HttpResponse
+  ): Route = _ => res map Complete
```
cocreature, line 80:
```suggestion
  def toRoute(res: Future[HttpResponse]): Route = _ => res map Complete
```
this compiles just fine for me. What am I missing?

cocreature: Alternatively, maybe consider just giving this function a different name instead of doing some weird overloading.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -69,106 +70,72 @@ class Endpoints(
   import encoder.implicits._
   import json.JsonProtocol._
   import util.ErrorOps._
-  import Uri.Path._
+
+  // Weird signature required, because it looks like Scala
+  // does not see a difference between Future[HttpResponse] & other inner types.
+  // Without it, it will complain that the function has been defined twice.
+  @nowarn("msg=parameter value ev in method toRoute is never used")
+  def toRoute[A <: HttpResponse](res: Future[A])(implicit
```
cocreature, line 78:
```suggestion
  private def toRoute[A <: HttpResponse](res: Future[A])(implicit
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -69,106 +70,72 @@ class Endpoints(
...
+      ev: A =:= HttpResponse
+  ): Route = _ => res map Complete
+  // I think the arguments need to be lazy
+  // because we extract the request early and pass it on to
+  // the method of each possible path.
+  // I assume that otherwise it would execute these
+  // in advance, although the path never would have matched.
+  def toRoute[A](
```
cocreature, line 86:
```suggestion
  private def toRoute[A](
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -69,106 +70,72 @@ class Endpoints(
...
+      res: => ET[domain.SyncResponse[A]]
+  )(implicit metrics: Metrics, jsonWriter: JsonWriter[A]): Route =
+    toRoute(httpResponse(res))
+  def toRoute(res: => Future[Error \/ SearchResult[Error \/ JsValue]]): Route =
```
cocreature, line 90:
```suggestion
  private def toRoute(res: => Future[Error \/ SearchResult[Error \/ JsValue]]): Route =
```

cocreature (approved):
Nice, thank you!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: Wrapping my head around how I can test it correctly :smile: Without lazyness the tests did pass too but are they covering scenario? I have my doubts right now (however I did not checked the tests yet).

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: ok let me try locally, maybe I did mess something up.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: guess that makes sense, thanks!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
cocreature: I was thinking of a simple manual test, add some `println`s in the futures and see if they get triggered for routes that aren’t hit.

realvictorprm: ok weird, you're right. Will change that stuff (as I just noticed that I cannot press the button here anymore after committing the other suggestions :sweat_smile:).

realvictorprm: many thanks, done with 2f55802

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: As discussed in slack. Turns out implicit calls ensure that every route after the path is being evaluated lazily (or to be more correct "call-by-name")
image
image

PR #10277, opened 2021-07-14
LF: change type from Try to Either in archive module

remyhaemmerle-da:
This is the first part of restructuring errors in archive module.
This is part of #9974.
    
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/Error.scala
+++ b/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/Error.scala
@@ -5,7 +5,7 @@ package com.daml.lf.archive
 
 import java.io.File
 
-class Error(val msg: String) extends RuntimeException(msg)
+sealed abstract class Error(val msg: String) extends RuntimeException(msg)
 
 object Error {
 
--- a/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/Error.scala
+++ b/daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/Error.scala
@@ -3,9 +3,15 @@
 
 package com.daml.lf.archive
 
+import java.io.File
+
 class Error(val msg: String) extends RuntimeException(msg)
 
 object Error {
+
+  final case class Internal(where: String, message: String) extends Error(s"IO error: $message")
+
+  final case class IO(where: String, cause: java.io.IOException) extends Error(s"IO error: $cause")
 
   import GenDarReader.ZipEntries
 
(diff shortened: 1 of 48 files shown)


daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/package.scala
```diff
@@ -8,8 +8,35 @@ import com.daml.lf.data.Ref.PackageId
 import com.daml.lf.language.{Ast, LanguageVersion}
 import com.google.protobuf.CodedInputStream
 
+import scala.util.Using
+import scala.util.Using.Releasable
+import scala.util.control.NonFatal
+
 package object archive {
 
+  type Result[X] = Either[Error, X]
```
S11001001, line 17:
```suggestion
  private[archive] type Result[X] = Either[Error, X]
```

Your choice, but aliases are expanded before testing for signature accessibility, so you may freely mark this as any privacy level even for public functions that use it, and getting `Result` out of the visible namespace might be a good idea.

cocreature: not a huge fan of the type synonyms for this type of stuff. You save very little typing and you have to keep chasing an indirection everywhere to make sense of the code.

cocreature (approved):
Thanks!

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/Error.scala
```diff
@@ -3,31 +3,39 @@
 
 package com.daml.lf.archive
 
+import java.io.File
+
 class Error(val msg: String) extends RuntimeException(msg)
```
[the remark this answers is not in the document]
remyhaemmerle-da: By diff, just a previous PR. ;)

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/package.scala
remyhaemmerle-da: I drop the type signature in 88dfffda244f564c3e6202385a538285da2f5d49

S11001001:
> not a huge fan of the type synonyms for this type of stuff.

I am a fan of this specific example, for two reasons:

1. It's of the "make me rewrite all these signatures once, shame on you, make me do it twice, shame on me" variety. Such a refactoring worth doing once is worth guarding against having to do again.
2. Specific to Scala, in situations where you can't infer the higher-kinded type `Either[Error, ?]`, it's nicer to have `Result` at hand to avoid the kind-projector, or worse, writing it out in the official way `({type l[a] = Either[Error, a]})#l`.

PR #10289, opened 2021-07-15
Register divulgence warning as a diagnostic

sofiafaro-da:
With this PR, the divulgence warnings for a scenario are registered as a diagnostic
when there isn't a scenario error. (Rationale: Scenario errors already
include the warnings in their text, after trace log entries.)

Part of #9947, follow up from #10253

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
+++ b/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
@@ -55,6 +55,7 @@ import qualified Data.Text.Extended as T
 import qualified Data.Text.Lazy as TL
 import Data.Tuple.Extra
 import Data.Typeable (Typeable())
+import qualified Data.Vector as V
 import Development.IDE.Core.Compile
 import Development.IDE.Core.OfInterest
 import Development.IDE.GHC.Error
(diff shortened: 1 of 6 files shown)


compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
```diff
@@ -904,19 +905,22 @@ toDiagnostic ::
...
-toDiagnostic file world range (Left err) =
-    Just $
-    (file, ShowDiag, ) $
-    Diagnostic
+toDiagnostic file world range = \case
+    Left err -> Just $ mkDiagnostic DsError $ formatScenarioError world err
+    Right SS.ScenarioResult{..}
+        | V.null scenarioResultWarnings -> Nothing
+        | otherwise -> Just $ mkDiagnostic DsWarning $
+            LF.prettyWarningMessages scenarioResultWarnings
```
cocreature, line 913: is it worth making this a separate diagnostic for each warning? Slightly orthogonal it could be very useful to embed to use the commit location (the callsite of `submit`) instead of the location of the scenario for the warnings which makes it much easier to pin down.

cocreature (approved):
Nice! Some minor ux suggestions

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
sofiafaro-da: I like the idea of using the submit location. Is that known to the engine, or would it be something we have to figure out how to propagate?

(Merging for now, since it should be a separate PR in any case.)

sofiafaro-da: As for separate warnings, I think it makes sense to have separate them if we separate the location. I don't think there's much value in having separate warnings otherwise.

cocreature: It is known to the engine and we already display it in Daml Studio. You can scrape it from https://github.com/digital-asset/daml/blob/cf2f79db4eaaaaba73bd3f1ccd503d9afcb0f07a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala#L121

sofiafaro-da: Thanks!

PR #10291, opened 2021-07-15
ledger-on-sql: Remove conformance tests for H2 and SQLite files.

SamirTalwar-DA:
We don't use these in any kind of production setting, and it reduces the CI overhead a lot.

### Pull Request Checklist

trigger the build.


cocreature (approved):
looks sensible, can we just stop supporting them completely and error out?

SamirTalwar-DA:
@cocreature: Not really, unless we want to parse JDBC URLs to a much greater degree.

fabiotudone-da (commented):
I think that, if we support something, we should test it; thus, I agree with @moritzkiefer-da that we should remove H2 and SQLite support altogether, and actually I think doing so should really be inseparable from the removal of the tests (else everything should stay as it is).

SamirTalwar-DA:
Well, we can't remove support altogether; we still need to support them for the in-memory version of Sandbox.

fabiotudone-da:
> @cocreature: Not really, unless we want to parse JDBC URLs to a much greater degree.

Can't we e.g. whitelist on the connection class?

fabiotudone-da:
> Well, we can't remove support altogether; we still need to support them for the in-memory version of Sandbox.

Both of them? Then I am actually not OK with removing the tests.

cocreature:
I think we need to clarify this a bit:
We use h2 an sqlite in in-memory mode (not sure why we need both :shrug:). We don’t use the file mode anywhere.

SamirTalwar-DA:
Yes, there are still tests for the in-memory modes.

I think we can probably drop support for H2 altogether, but I need to look into it more.

cocreature:
> I think we can probably drop support for H2 altogether, but I need to look into it more.

No, Canton uses this in their tests atm.

fabiotudone-da:
> (not sure why we need both)

> I think we can probably drop support for H2 altogether, but I need to look into it more.

Well that's perhaps another great PR.

> We don’t use the file mode anywhere

I see, thanks, then I am OK with this. I agree it would be better to error out on what we don't support.

fabiotudone-da (approved):

hubert-da (approved):

miklos-da (approved):

PR #10294, opened 2021-07-15
[JSON-API] Log json request & response bodies in debug

### LINKED ISSUE #10122: Log full request & response bodies in the JSON API under debug logging
Probably don’t want to enable this by default but at the moment it’s basically impossible to get them in a readable format (akka-http has something but ugh).

We can probably use https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/debugging-directives/logRequestResult.html for this.

realvictorprm:
This also readds logging of incoming requests and the responses which are being send out.

changelog_begin

- [JSON-API] You can now find for appropriate requests in the debug logs the request and response body within the logging context (The field names are "request-body" and "response-body")

changelog_end

Fixes #10122

### Pull Request Checklist

- [x] This was manually tested and worked great :star_struck: 

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/resources/logback.xml
+++ b/ledger-service/http-json/src/main/resources/logback.xml
@@ -9,7 +9,7 @@
                 <!-- encoders are assigned the type
                     ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
                 <encoder>
-                    <pattern>%date{"dd-MM-yyyy HH:mm:ss.SSS", UTC} [%thread] %-5level %logger{36} - %msg %replace(, context: %marker){', context: $', ''} %n</pattern>
+                    <pattern>%date{"dd-MM-yyyy HH:mm:ss.SSS", UTC} [%thread] %-5level %logger{36} - %msg%replace(, context: %marker){', context: $', ''} %n</pattern>
                 </encoder>
             </else>
         </if>
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -15,7 +15,7 @@ import akka.http.scaladsl.model.headers.{
 }
 import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directives.extractClientIP
-import akka.http.scaladsl.server.{Directive, Directive0, PathMatcher, Route}
+import akka.http.scaladsl.server.{Directive0, PathMatcher, Route}
 import akka.http.scaladsl.server.RouteResult._
 import akka.stream.Materializer
 import akka.stream.scaladsl.{Flow, Source}
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -10,10 +10,12 @@ import akka.http.scaladsl.model.headers.{
   ModeledCustomHeader,
   ModeledCustomHeaderCompanion,
   OAuth2BearerToken,
+  `Content-Type`,
   `X-Forwarded-Proto`,
 }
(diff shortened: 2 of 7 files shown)


ledger-service/http-json/src/main/resources/logback.xml
```diff
@@ -9,7 +9,7 @@
                 <!-- encoders are assigned the type
                     ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
                 <encoder>
-                    <pattern>%date{"dd-MM-yyyy HH:mm:ss.SSS", UTC} [%thread] %-5level %logger{36} - %msg %replace(, context: %marker){', context: $', ''} %n</pattern>
+                    <pattern>%date{"dd-MM-yyyy HH:mm:ss.SSS", UTC} [%thread] %-5level %logger{36} - %msg%replace(, context: %marker){', context: $', ''} %n</pattern>
```
realvictorprm, line 12: That thing here changes the formatting to instead of this "SomeSuperNiceLoggingText , context: ..." to "SomeSuperNiceLoggingText, context: ...". The space before the `,` annoyed me too much.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +79,84 @@ class Endpoints(
   private def toRoute(res: => Future[Error \/ SearchResult[Error \/ JsValue]]): Route =
     responseToRoute(httpResponse(res))
 
+  // Always put this directive after a path to ensure
+  // that you don't log request bodies multiple times (simply because a matching test was made multiple times).
+  // TL;DR JUST PUT THIS THING AFTER YOUR FINAL PATH MATCHING
```
realvictorprm, line 84: Just trying to clarify here that it is considered to be a bug if the directive is being executed without the path afterwards matching. This should prevent us from printing request bodies which are targeted to unmatched routes (And also as described in the comment can result in printing the body more than once).

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +79,84 @@ class Endpoints(
...
+            case _ =>
+              Future.failed _ apply
+                new RuntimeException(
+                  """Logging the request & response should never happen on routes which get rejected.
+                    |Make sure to place the directive only at places where a match is guaranteed (e.g. after the path directive).""".stripMargin
+                )
+          }
+        } yield response
+      }
+    }
```
realvictorprm, line 140: This whole body logging should be fully plugable, because in a production scenario you don't want to log the incoming request and outgoing response bodies (as these are potentially big and let the logs explode with time next to the performance implications).

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +79,84 @@ class Endpoints(
...
+          httpMessage
+            .header[`Content-Type`]
+            .map(_.contentType)
+            .contains(ContentTypes.`application/json`)
+        )
+          httpMessage
+            .entity()
+            .toStrict(maxTimeToCollectRequest)
+            .map(_.data.utf8String)
+            .map(_.parseJson)
```
realvictorprm, line 104: yes this can be quite the performance bottleneck as it's hella inefficient for the response body (which has been constructed by us which we are now parsing again :facepalm:), however I consider this to be a debugging helper.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: I guess this is the drawback of the directives, one has to be quite careful using them.

realvictorprm:
darn, noticed some flaws I need to address beforehand. Will re-open as soon as I did that.

PR #10298, opened 2021-07-15
[JSON-API] Put the request & response bodies into the log ctx if the log level is debug (for specific statements)

### LINKED ISSUE #10122: Log full request & response bodies in the JSON API under debug logging

realvictorprm:
This also readds logging of incoming requests and the responses which are being send out.

changelog_begin

- [JSON-API] Logging of the request and response bodies are now available for appropriate requests if the chosen log level is equal or lower than DEBUG. These can then be found in the logging context of the request begin & end log messages (The field names in the ctx are "request_body" and "response_body").

changelog_end

Fixes #10122

### Pull Request Checklist

- [x] This was manually tested and worked great :star_struck: 

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -178,17 +178,19 @@ class Endpoints(
     implicit val lc: LoggingContextOf[InstanceUUID with RequestID] =
       extendWithRequestIdLogCtx(identity)(lc0)
     import metrics.daml.HttpJsonApi._
-    def path[L](pm: PathMatcher[L]) = server.Directives.path(pm) & logRequestAndResult
+    val markThroughputAndLogProcessingTime: Directive0 = Directive { (fn: Unit => Route) =>
+      val t0 = System.nanoTime
+      metrics.daml.HttpJsonApi.httpRequestThroughput.mark()
+      fn(()).andThen(_.map { res =>
+        logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
+        res
+      })
+    }
+    def path[L](pm: PathMatcher[L]) =
+      server.Directives.path(pm) & markThroughputAndLogProcessingTime & logRequestAndResult
     def withTimer(timer: Timer) =
       Directive { (fn: Unit => Route) =>
-        val t0 = System.nanoTime
-        metrics.daml.HttpJsonApi.httpRequestThroughput.mark()
-        fn(()).andThen(res =>
-          for {
-            res <- Timed.future(timer, res)
-            _ = logger.trace(s"Processed request after ${System.nanoTime() - t0}ns")
-          } yield res
-        )
+        fn(()).andThen(res => Timed.future(timer, res))
       }
     val withCmdSubmitTimer: Directive0 = withTimer(commandSubmissionTimer)
     val withFetchTimer: Directive0 = withTimer(fetchTimer)
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
@@ -10,10 +10,12 @@ import akka.http.scaladsl.model.headers.{
   ModeledCustomHeader,
   ModeledCustomHeaderCompanion,
   OAuth2BearerToken,
+  `Content-Type`,
   `X-Forwarded-Proto`,
 }
+import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directives.extractClientIP
-import akka.http.scaladsl.server.{Directive0, Route}
+import akka.http.scaladsl.server.{Directive, Directive0, PathMatcher, Route}
 import akka.http.scaladsl.server.RouteResult._
 import akka.stream.Materializer
 import akka.stream.scaladsl.{Flow, Source}
@@ -47,7 +49,6 @@ import scala.util.control.NonFatal
 import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
 import com.daml.metrics.{Metrics, Timed}
 import akka.http.scaladsl.server.Directives._
-import akka.http.scaladsl.server.Directive
 
 class Endpoints(
     allowNonHttps: Boolean,
@@ -59,6 +60,7 @@ class Endpoints(
     healthService: HealthService,
     encoder: DomainJsonEncoder,
     decoder: DomainJsonDecoder,
+    shouldLogHttpBodies: Boolean,
     maxTimeToCollectRequest: FiniteDuration = FiniteDuration(5, "seconds"),
 )(implicit ec: ExecutionContext, mat: Materializer) {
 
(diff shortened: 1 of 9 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +80,117 @@ class Endpoints(
...
+      httpResponse => Future.successful(logger.info(mkResponseLogMsg(httpResponse))),
+    )
+
+  val logRequestAndResultFn: LoggingContextOf[InstanceUUID with RequestID] => Directive0 =
+    if (shouldLogHttpBodies) lc => logJsonRequestAndResult(lc)
+    else lc => logRequestAndResultSimple(lc)
+
+  def logRequestAndResult(implicit lc: LoggingContextOf[InstanceUUID with RequestID]): Directive0 =
+    logRequestAndResultFn(lc)
+
```
realvictorprm, line 175: I think I should move this somewhere else as this did quite explode in size.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValueSerializer.scala
```diff
@@ -25,6 +36,34 @@ private[logging] object LoggingValueSerializer {
...
         generator.writeEndArray()
+      case LoggingValue.OfJson(jsValue) =>
+        def write(jsValue: JsValue): Unit =
+          jsValue match {
+            case JsNull =>
+              generator.writeNull()
+            case JsTrue =>
+              generator.writeBoolean(true)
+            case JsFalse =>
+              generator.writeBoolean(false)
```
S11001001, line 47:
```suggestion
```

realvictorprm: Actually tried that already but the compiler complained ._.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +80,117 @@ class Endpoints(
...
+          .header[`Content-Type`]
+          .map(_.contentType)
+          .contains(ContentTypes.`application/json`)
+      )
+        httpMessage
+          .entity()
+          .toStrict(maxTimeToCollectRequest)
+          .map(_.data.utf8String)
+          .map(_.parseJson)
+          .map(body =>
```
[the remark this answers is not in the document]
realvictorprm: True 🤦 , thanks for this.

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
```diff
@@ -158,6 +158,12 @@ class OptionParser(getEnvVar: String => Option[String])(implicit
     .valueName(WebsocketConfig.usage)
     .text(s"Optional websocket configuration string. ${WebsocketConfig.help}")
 
+  opt[Unit]("log-request-and-response-bodies")
```
cocreature, line 161: Do we really want a separate option for this? I’d just fold this under the loglevel and turn it on for debug & trace (or just trace, I don’t feel strongly about this).

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValueSerializer.scala
cocreature: iirc it breaks exhaustiveness checks since `JsBoolean` is an `unapply` thingy.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValueSerializer.scala
```diff
@@ -25,6 +36,34 @@ private[logging] object LoggingValueSerializer {
         generator.writeStartArray()
         sequence.foreach(writeValue(_, generator))
         generator.writeEndArray()
+      case LoggingValue.OfJson(jsValue) =>
+        def write(jsValue: JsValue): Unit =
```
cocreature, line 40: maybe move it to a top-level definition?

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -77,17 +80,117 @@ class Endpoints(
...
+                new RuntimeException(
+                  """Logging the request & response should never happen on routes which get rejected.
+                    |Make sure to place the directive only at places where a match is guaranteed (e.g. after the path directive).""".stripMargin
+                )
+          }
+        } yield response
+      }
+    }
+
+  private def logJsonRequestAndResult(implicit
```
cocreature, line 114: I’m a bit confused, this seems like a fair amount of complexity when https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/debugging-directives/logRequestResult.html seems like it should do the job. What problem does this solve that `logRequestResult` does not solve?

ledger-service/http-json-cli/src/main/scala/com/daml/http/OptionParser.scala
realvictorprm: This is a very good idea, I would be fine with this too. Need to adjust the changelog then.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: I did not choose it, because it does not support using our contextualized logging (it seems the documentation is slightly outdated, the logging context parameter does not exist if I navigate the source in IntelliJ). However my implementation is pretty similar to the one from akka as I based on it :)

cocreature: Not sure contextualized logging matters all that much here given that this is purely for debugging but given that you already implemented it, we might as well keep it :+1:

cocreature (approved):
Thanks!

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: will do that later

realvictorprm:
Damn you Linux_scala_2_12 build (╯°□°)╯︵ ┻━┻

realvictorprm:
Well found some nits myself, after taking my rounds (nothing really serious, just me being nitpicky with myself). Hopefully this was the last commit I had to add for merging it.

libs-scala/contextualized-logging/src/main/scala/com/digitalasset/logging/LoggingValueSerializer.scala
S11001001: If that's true, then the `JsBoolean` case should be removed. Unless we are defensively programming for the addition of a third boolean value in the future 👻

PR #10301, opened 2021-07-15
Remove optLocation field from transaction nodes.

nickchapman-da:
Remove `optLocation` field from all transaction nodes.
- (This is a step towards building normalized transactions directly)
- Instead pass a extra `locationInfo` Map in `CompleteTransaction`; built by `PartialTransaction`
- fixup scenario-service `Conversions` to use the new map
- adapt rest of code base to remove `optLocation` field.

### Pull Request Checklist

trigger the build.


compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -553,10 +553,11 @@ final class Conversions(
       .build()
   }
 
-  def convertNode(
-      nodeWithId: (NodeId, N.GenNode[NodeId, V.ContractId])
-  ): proto.Node = {
+  def convertIncompleteNode(
```
nickchapman-da, line 556: renamed to avoid ad-hoc overloading with another function of the same name in this file.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -751,18 +760,23 @@ private[lf] case class PartialTransaction(
...
+      actionNodeLocations = actionNodeLocations :+ optLocation,
       nextNodeIdx = nextNodeIdx + 1,
       context = context.addActionChild(nid, version),
       nodes = nodes.updated(nid, node),
     )
   }
 
   /** Unwind the transaction aborting all incomplete nodes */
-  def unwind: PartialTransaction = {
+  def unwind(): PartialTransaction = {
```
nickchapman-da, line 779: Added parens for clarity. I find paren-less `def`s unnecessarily surprising!

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -224,22 +224,6 @@ class TransactionSpec
         na should not be nb
       }
     }
-
```
nickchapman-da, line 227: There is now no location to ignore! :smile:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -266,6 +270,7 @@ private[lf] case class PartialTransaction(
     keys: Map[GlobalKey, PartialTransaction.KeyMapping],
     globalKeyInputs: Map[GlobalKey, PartialTransaction.KeyMapping],
     localContracts: Set[Value.ContractId],
+    actionNodeLocations: BackStack[Option[Location]],
```
cocreature, line 273: probably a good idea to add a doc comment for this feld.

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
```diff
@@ -439,7 +439,7 @@ final class Conversions(
     val tx = incomplete.transaction
 
     val builder = proto.PartialTransaction.newBuilder
-      .addAllNodes(tx.nodes.map(convertNode).asJava)
+      .addAllNodes(tx.nodes.map(convertIncompleteNode(incomplete.locationInfo)).asJava)
```
cocreature, line 442: I think this is slightly misleading. The nodes are not necessarily incomplete only some might be. Maybe `convertPtxNode`?

cocreature (approved):
Nice, thank you!

miklos-da (approved):

compiler/scenario-service/server/src/main/scala/com/digitalasset/daml/lf/scenario/Conversions.scala
nickchapman-da: `convertIncompleteTransactionNode`

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: The scala convention, is "use parentheses if there is a side effect". I think it also fine to add it if the evaluation is computationally heavy. 
In this particular case, I am curious to know what did you find surprising.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -321,6 +329,12 @@ private[lf] case class PartialTransaction(
       sb.toString
     }
 
+  private def locationInfo(): Map[NodeId, Location] = {
+    this.actionNodeLocations.toImmArray.toSeq.zipWithIndex.collect { case (Some(loc), n) =>
```
remyhaemmerle-da, line 333: Is the `this.`  very useful here?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -632,7 +645,8 @@ private[lf] case class PartialTransaction(
           context =
             ec.parent.addActionChild(nodeId, exerciseNode.version min context.minChildVersion),
           nodes = nodes.updated(nodeId, exerciseNode),
-          actionNodeSeeds = actionNodeSeeds :+ actionNodeSeed,
+          actionNodeSeeds =
+            actionNodeSeeds :+ actionNodeSeed, //(NC) pushed by 'beginExercises'; why push again?
```
remyhaemmerle-da, line 649: >     pushed by 'beginExercises'; why push again?

probably a bug.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/transaction/TransactionSpec.scala
```diff
@@ -224,22 +224,6 @@ class TransactionSpec
...
-          case nf: Node.NodeFetch[V.ContractId] => nf copy (optLocation = None)
-          case ne: Node.NodeExercises[Nothing, V.ContractId] =>
-            ne copy (optLocation = None)
-          case nl: Node.NodeLookupByKey[V.ContractId] => nl copy (optLocation = None)
-        }
-        nodeWithoutLocation
-      }
-      isReplayedBy(withoutLocation, n) shouldBe Right(())
-      isReplayedBy(n, withoutLocation) shouldBe Right(())
-    }
```
remyhaemmerle-da, line 242: :fire:

remyhaemmerle-da (approved):
Nice.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: If was a line in `finish()` which I found really
```
val ptx = unwind
```
It looked like a variable renaming, but of course was a method call.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: probably not so much

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
remyhaemmerle-da: This is pretty standard scala. I think it is better you get use to it that working around. 
`val x = y` is definitely not a variable renaming in scala. Tale for example :

```
call f(x: => X) = {
    ...
    val y = x 
    ...
    g(x)
    ...
    h(x)
    ...
}
```

PR #10313, opened 2021-07-19
detect mount issue earlier

garyverhaegen-da:
When machine disks are full, we can't clean the Bazel cache if it happens to not be a mount point. I don't quite understand yet why it's not a mount point, but maybe I'll be able to investigate more if we catch the issue early, rather than waiting for the disk to be full and the clean-up to fail.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/clean-up.yml
+++ b/ci/clean-up.yml
@@ -16,9 +16,11 @@ steps:
     case "$UNAME" in
     Darwin)
       MPS="/var/tmp/_bazel_vsts /Users/vsts/.bazel-cache"
+      CMD="hdiutil info"
       ;;
     Linux)
       MPS="/home/vsts/.cache/bazel /home/vsts/.bazel-cache"
+      CMD="mount"
       ;;
     *)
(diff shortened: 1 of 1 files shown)


ci/clean-up.yml
```diff
@@ -10,6 +10,31 @@ steps:
...
+      MPS="/home/vsts/.cache/bazel /home/vsts/.bazel-cache"
+      ;;
+    *)
+      echo "Unexpected uname: $UNAME"
+      MPS=
+      ;;
+    esac
+
+    for path in $MPS; do
+      if ! (mount | grep $path); then
```
SamirTalwar-DA, line 30: Maybe use `grep -F` to escape the `.` characters?

SamirTalwar-DA (approved):

SamirTalwar-DA:
Looks like it's failing because:

> mount: command not found

cocreature (approved):
Thanks!

PR #10328, opened 2021-07-20
fix besu listing

garyverhaegen-da:
The current entry has been moved from "open-source integrations" to "commercial integrations" in 73b38f8add. However, the move was done verbatim, while the format for the two tables is different, making the Besu listing stick out a bit. Moreover, since the listing was added the URL of the press release has changed, landing people on the front page of the DA blog (from which it is possible to find the relevant press release, but quite frankly googling for it is faster).

For the record, I still think this page should be hosted elsewhere as the docs release cycle doesn't offer enough dynamism for it.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/docs/source/deploy/index.rst
+++ b/docs/source/deploy/index.rst
@@ -33,12 +33,12 @@ for production use today.
    * - `Sextant for Daml <https://blockchaintp.com/sextant/daml/>`__
      - `Amazon QLDB <https://aws.amazon.com/qldb/>`__
      - `Blockchain Technology Partners <https://blockchaintp.com/>`__
+   * - `Sextant for Daml <https://blockchaintp.com/sextant/daml/>`__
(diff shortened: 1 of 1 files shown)


cocreature (approved):

dasormeter:
@garyverhaegen-da @moritzkiefer-da there actually is an open source daml-on-besu offering that is using public kvutils https://github.com/blockchaintp/daml-on-besu. and a private offering that is closed source. To operationally run both can be via Sextant, but the open source one is available in the wild to play around with - this was open sourced a few months ago

garyverhaegen-da:
I don't personally know what the bar-to-pass is for inclusion in the open-source table here. @cocreature, do you know, or who to ask?

cocreature:
No idea, I’d wait for @bame-da to be back from vacation in 2 weeks. I don’t think this is urgent.

PR #10329, opened 2021-07-20
Report divulgence warning at commit location.

sofiafaro-da:
This PR records the commitLocation when the warning is generated, and uses it when creating the diagnostic.

Part of #9947, follow up from #10289

### DIFF
--- a/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
+++ b/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
@@ -821,9 +821,9 @@ runScenariosRule =
               let scenarioName = LF.qualObject scenario
               let mbLoc = NM.lookup scenarioName (LF.moduleValues m) >>= LF.dvalLocation
               let range = maybe noRange sourceLocToRange mbLoc
-              pure (toDiagnostic file world range res , (vr, res))
+              pure (toDiagnostics world file range res, (vr, res))
       let (diags, results) = unzip scenarioResults
-      pure (catMaybes diags, Just results)
+      pure (concat diags, Just results)
 
 runScriptsRule :: Options -> Rules ()
 runScriptsRule opts =
@@ -840,9 +840,9 @@ runScriptsRule opts =
               let scenarioName = LF.qualObject scenario
               let mbLoc = NM.lookup scenarioName (LF.moduleValues m) >>= LF.dvalLocation
               let range = maybe noRange sourceLocToRange mbLoc
-              pure (toDiagnostic file world range res, (vr, res))
+              pure (toDiagnostics world file range res, (vr, res))
       let (diags, results) = unzip scenarioResults
-      pure (catMaybes diags, Just results)
+      pure (concat diags, Just results)
 
 runScenariosScriptsPkg ::
        NormalizedFilePath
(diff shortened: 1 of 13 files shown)


compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
```diff
@@ -922,6 +925,26 @@ toDiagnostic file world range = \case
...
 
+    fileRangeFromMaybeLocation :: Maybe SS.Location -> (NormalizedFilePath, Range)
+    fileRangeFromMaybeLocation mbLocation =
+        fromMaybe (scenarioFile, scenarioRange) $ do
+            location <- mbLocation
+            lfModule <- LF.lookupLocationModule world location
+            filePath <- LF.moduleSource lfModule
+            Just (toNormalizedFilePath' filePath, rangeFromLocation location)
+
+    rangeFromLocation :: SS.Location -> LSP.Range
```
cocreature, line 936: completely orthogonal but we should also be able to use the commit location for errors during scenario interpretation that happen in a `submit`. I’ll open an issue.

cocreature (approved):
Nice, thank you! Can we change the tests you added before to match on the location?

compiler/damlc/tests/daml-test-files/Divulgence.daml
```diff
@@ -44,7 +44,8 @@ template DisclosureDelegation
         do
           fetch secretCid -- actor will divulge secretCid to p
 
--- @WARN range=48:1-48:5; Use of divulged contracts is deprecated
+-- @WARN range=49:1-49:5; Use of divulged contracts is deprecated
+-- @WARN range=49:1-49:5; Use of divulged contracts is deprecated
```
sofiafaro-da, line 48: This isn't the submit location :-(

compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs
sofiafaro-da: That's reasonable.

Also, I don't know exactly why, but this isn't working right now. I don't know if the commit location is off, or the source location, or something else.

cocreature: Let me know if you want me to take a closer look or need a rubber duck for debugging

sofiafaro-da: It looks like the commit location wasn't being set during a submit. I changed it, and now the locations show up correctly. (I think I still have an off by one error in the column, but that's easy to fix.)

cocreature: uh that’s not good, nice job catching this

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -798,7 +803,7 @@ private[lf] object Speedy {
             .initial(pkg2TxVersion, contractKeyUniqueness, submissionTime, initialSeeding),
           committers = committers,
           readAs = readAs,
-          commitLocation = None,
+          commitLocation = commitLocation,
```
cocreature, line 806: Nice :+1:

cocreature: I checked why this didn’t blow up before and it turns out we didn’t actually use commitLocation for anything. It was originally used before I split scenario interpretation into off-ledger and on-ledger interpretation but with the split we had access to it in `ScenarioRunner` already.

So at least nothing else was broken by the fact that we didn’t set this.

sofiafaro-da: Ok, good ... I was a bit worried.

PR #10336, opened 2021-07-20
fix broken triggers link

garyverhaegen-da:
CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/docs/source/tools/trigger-service/index.rst
+++ b/docs/source/tools/trigger-service/index.rst
@@ -10,7 +10,7 @@ Trigger Service
    ./authorization
    ./auth0_example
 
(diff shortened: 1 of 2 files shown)


garyverhaegen-da:
Current version links to [`/tools/triggers/index.html#running-a-daml-trigger`](https://docs.daml.com/tools/triggers/index.html#running-a-daml-trigger), which does not exist.

cocreature (approved):
I’ll never understand sphinx links

garyverhaegen-da:
> I’ll never understand sphinx links

I definitely can't claim to understand all of it, but in this case we had set it up as a "literal URL" link, not a Sphinx-knows-about-it link, so the build would not check if the link is valid. With the `:doc:` and `:ref:` directives at least the build fails when the link is invalid.

PR #10337, opened 2021-07-20
LF: Compile netsed ELets using constant stack

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
```diff
@@ -763,14 +763,17 @@ private[validation] object Typing {
...
-        introExprVar(vName, typ1).typeOf(body)
-      case Binding(_, _, bound @ _) =>
-        typeOf(body)
-    }
+    private def typeOfLet(bindings: List[Binding], body: Expr): Type =
+      bindings match {
+        case Binding(Some(vName), typ0, expr) :: tail =>
+          checkType(typ0, KStar)
+          val typ1 = checkExpr(expr, typ0)
+          introExprVar(vName, typ1).typeOfLet(tail, body)
```
cocreature, line 771: this doesn’t look stack safe, should we make it stack safe? I don’t expect the list of bindings to grow particularly large but it adds to an already large stack.

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
```diff
@@ -763,14 +763,17 @@ private[validation] object Typing {
...
+        case Binding(_, _, bound @ _) :: tail =>
```
cocreature, line 772:
```suggestion
        case Binding(None, _, bound @ _) :: tail =>
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -738,10 +738,18 @@ private[lf] final class Compiler(
...
+        case ELet(bindings, body) =>
+          val bounds = bindings.map { case Binding(optBinder, _, bound) =>
+            val se = withOptLabel(optBinder, compile(bound))
+            val p = nextPosition()
+            optBinder.foreach(addExprVar(_, p))
+            se
+          }
+          compile(body) match {
+            case SELet(bounds1, body1) =>
+              SELet(bounds ++ bounds1, body1)
```
cocreature, line 750: do we need to worry about the inner let having references to variables bound in outer let? I forgot if you can have references in multi-lets to previous bindings in daml-lf.

cocreature (approved):
Nice!

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
remyhaemmerle-da: We want, but another time

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
remyhaemmerle-da: reverted

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -734,17 +734,21 @@ private[lf] final class Compiler(
...
+  private[this] def compileELet(
+      eLet0: ELet,
+      bindings: mutable.Builder[SExpr, List[SExpr]] = List.newBuilder,
+  ): SELet = {
+    val binding = eLet0.binding
+    bindings += withOptLabel(binding.binder, compile(binding.bound))
+    val boundPos = nextPosition()
+    binding.binder.foreach(addExprVar(_, boundPos))
+    eLet0.body match {
+      case eLet1: ELet => compileELet(eLet1, bindings)
```
cocreature, line 748: you’re still merging nested lets here which is only valid if we can have references between bindings in a let so my previous comment still holds

remyhaemmerle-da: > do we need to worry about the inner let having references to variables bound in outer let? I forgot if you can have references in multi-lets to previous bindings in daml-lf.

We do not have to worry. In Speedy (like in the proto),  the  `Let` construction with multiple binding is a compressed form of nested let, each bound have access to the variable of previous bound. 

In any case you can see the function `let` using in the original implementation was already merging the `SELet`.

cocreature: great thanks for confirming

PR #10342, opened 2021-07-20
fix template dot-files

garyverhaegen-da:
The `cp $SRC/$d/*` does not copy dot files.

CHANGELOG_BEGIN
CHANGELOG_END


templates/BUILD.bazel
```diff
@@ -68,10 +68,13 @@ genrule(
         for d in skeleton empty-skeleton create-daml-app quickstart-scala quickstart-java; do
             mkdir -p $$OUT/$$d
             cp -rL $$SRC/$$d/* $$OUT/$$d/
-            # use default .gitignore and .dlint.yaml if they don't exist in the template
-            cp -n $$SRC/default-gitattributes $$OUT/$$d/.gitattributes
-            cp -n $$SRC/default-gitignore $$OUT/$$d/.gitignore
-            cp -n $$SRC/default-dlint-yaml $$OUT/$$d/.dlint.yaml
+            for f in gitattributes gitignore dlint.yaml; do
+                if [ -f "$$SRC/$$d/.$$f" ]; then
```
cocreature, line 72: I don’t get this part, when do we hit this?

cocreature: Are there templates that have custom files to override these?

garyverhaegen-da: `create-daml-app` has a custom [`.gitignore`](https://github.com/digital-asset/daml/blob/main/templates/create-daml-app/.gitignore) and [.dlint.yaml](https://github.com/digital-asset/daml/blob/main/templates/create-daml-app/.dlint.yaml).

I'm not aware of a template with a custom `.gitattributes` at this point, but I did not see a strong reason to make a special case out of it.

cocreature: makes sense, thanks

cocreature (approved):

PR #10344, opened 2021-07-20
use a single SQL query for any number of json-api query pairs

S11001001:
Since #8226 we have been doing multiple SQL queries and deduplicating them on the Scala side if there were multiple queries for a given template ID. Instead, we can rerun the predicate in the SELECT clause and append to a list of query numbers instead, making it so that we don't have to do any deduplication in Scala, as briefly imagined in #9278.

This is better for complex multiqueries, of course, and also makes it possible to consider streaming database results in _all_ cases in the future.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Nice, looks sensible albeit a bit intricate but hopefully tests catch the subtleties.

Thanks!

ledger-service/db-backend/src/test/scala/com/digitalasset/http/dbbackend/QueriesSpec.scala
```diff
@@ -3,33 +3,75 @@
...
+            sql"CASE WHEN ( tpid = ${66L}) THEN ${1}||''  WHEN ( tpid = ${55L}) THEN ${0}||''  ELSE NULL END ",
+        ),
+      ),
+      (
+        Seq(
+          (SurrogateTpId(55L), fr"foo") -> 0,
+          (SurrogateTpId(55L), fr"bar") -> 1,
+          (SurrogateTpId(66L), fr"baz") -> 2,
+        ),
+        Reordered(
```
cocreature, line 44: Should we just use an ordered map instead? This really shouldn’t be our performance bottleneck and that seems like a much nicer solution than hardcoding the hash order.

S11001001: Yeah, we can, I'm just being stubborn about my instances.

S11001001: fc7db44d36b

cocreature: nice, thank you

PR #10355, opened 2021-07-21
Cleanup perf

garyverhaegen-da:
We've removed some features, this can be simplified a bit now.

CHANGELOG_BEGIN
CHANGELOG_END


ci/cron/daily-compat.yml
```diff
@@ -90,17 +93,21 @@ jobs:
           eval "$(dev-env/bin/dade assist)"
           source $(bash_lib)
 
-          OUT="$(Build.StagingDirectory)/perf-results.json"
+          OUT="$(Build.StagingDirectory)/perf-results-speedy.json"
```
cocreature, line 96: inlining this seems to make it slightly harder to run this locally and test the script. Do we gain anything in return?

garyverhaegen-da: When running locally, why not just run the Bazel command?
```
bazel run daml-lf/scenario-interpreter:scenario-perf -- -rf json -rff $tmp
```
The underlying goal here is to make it easy to add other targets, like `//ledger-service/http-json:contractdao-bench`.

cocreature: but how do I test that I didn’t mess up the jq command? Not opposed to this change but I don’t really get why inlining this makes it easier to add new targets

garyverhaegen-da: I was planning to use the templating feature of Azure YAML, like we do in the compatibility job above. It does not _require_ inlining, I suppose, but the value of the external script seemed null to me and inlining makes this a little bit easier. I'm not worried about testing a jq expression that should not change until the format of the perf report changes and for which we'll have daily feedback on Slack.

cocreature: alright I’m convinced :)

cocreature (commented):
lgtm will approve once the temporary CI hackery is gone

garyverhaegen-da:
> lgtm will approve once the temporary CI hackery is gone

I think it's all good now.

cocreature (approved):
Thanks!

PR #10360, opened 2021-07-21
[JSON-API] Shutdown on startup if the db connection is invalid

### LINKED ISSUE #10241: JSON API does not fail immediately if connection to DB is broken
As discovered by @akshayshirahatti-da, if you run the JSON API with `createSchema=false` it will happily startup even if it cannot connect to the JSON API. We should try to fail as soon as possible.

realvictorprm:
Should fix #10241

changelog_begin

- [JSON-API] The json api now correctly shutdowns at startup if the provided db connection is invalid in case of `createSchema=false`

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
@@ -123,6 +123,18 @@ object Main {
             terminate()
             System.exit(ErrorCodes.StartupError)
         }
+      case (Some(dao), _) =>
+        Try(dao.isValid(120).unsafeRunSync()).toEither match {
+          case Right(false) =>
+            logger.error("Database connection is not valid.")
+            terminate()
+            System.exit(ErrorCodes.StartupError)
+          case Left(e) =>
+            logger.error("Failed to connect to the database", e)
+            terminate()
+            System.exit(ErrorCodes.StartupError)
+          case _ =>
+        }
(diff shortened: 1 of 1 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -123,6 +123,18 @@ object Main {
...
+        Try(dao.isValid(120).unsafeRunSync()).toEither match {
+          case Right(false) =>
+            logger.error("Database connection is not valid.")
+            terminate()
+            System.exit(ErrorCodes.StartupError)
+          case Left(e) =>
+            logger.error("Failed to connect to the database", e)
+            terminate()
+            System.exit(ErrorCodes.StartupError)
+          case _ =>
```
cocreature, line 136:
```suggestion
          case Right(true) =>
```
not a huge fan of wildcard matches when they’re easily avoidable. They tend to bite you when they suddenly end up catching more than you wanted to after refactoring.

cocreature (approved):
Looks reasonable, thanks! As mentioned on slack, you might be able to test this via the failure tests.

realvictorprm:
ok noticed that I would need to refactor this a bit to be able to test it.

realvictorprm:
And I think it is best to merge this for now, because the refactoring and adding more tests can be done later.

PR #10363, opened 2021-07-21
Fix resource cleanup in CommonStorageBackend.ledgerEnd

cocreature:
Got flagged by VeraCode (I think, it really doesn’t want to tell you
what exactly it doesn’t like).

If we throw in fetch we will never cleanup.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
@@ -40,7 +40,7 @@ import com.daml.platform.store.interfaces.LedgerDaoContractsReader.{
 import com.daml.scalautil.Statement.discard
 
 import scala.collection.mutable
-import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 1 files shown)


nmarton-da:
this part of code is under construction by @rautenrieth-da at DPP-460, where we most likely will switch to anorm as in other places (renders this change unnecessary)
if you do not need this urgently I would suggest to close this and wait.
if you need this to activate some additional CI validation, and this is in the way, please feel free to proceed.

cocreature:
>  if you need this to activate some additional CI validation, and this is in the way, please feel free to proceed.

We (or rather Ed) runs VeraCode once a month on our code and this was one thing it flagged this time. Given that we already have this PR and it shouldn’t make it harder to do the changes you want to do, my preference would be to merge it but your/your team’s choice. I can deal with the false positive for now as well.

rautenrieth-da:
Fixed by https://github.com/digital-asset/daml/pull/10472, which used anorm instead of the plain Java interfaces.

PR #10368, opened 2021-07-22
Limit length of package ids to 64 characters

cocreature:
In practice, you cannot use a package id that is not a valid sha256
outside of a test anyway since we validate them. However, we didn’t
enforce it at the daml-lf level which is a bit annoying.

This PR fixes this and adds a length restriction of 64
characters (i.e. 256 bytes for sha256).

We could go a bit further and also restrict the characters to valid
sha256 hashes (i.e. 64 character hex strings). I don’t feel all that
strongly about that either way. We use other characters in tests but
fixing that shouldn’t be particularly hard either. The extra
characters don’t seem to cause problem so far, so I think it’s fine to
keep that.

No changelog entry since I don’t see how a user can be affected by this.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/encoder/src/main/scala/com/digitalasset/daml/lf/archive/testing/DamlLfEncoder.scala
+++ b/daml-lf/encoder/src/main/scala/com/digitalasset/daml/lf/archive/testing/DamlLfEncoder.scala
@@ -29,7 +29,7 @@ private[daml] object DamlLfEncoder extends App {
     throw new Error("You should not get this error")
   }
 
-  private val pkgId = Ref.PackageId.assertFromString("-self-")
+  private val pkgId = Ref.PackageId.assertFromString("self")
 
   private def main() =
     try {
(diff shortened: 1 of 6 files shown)


daml-lf/encoder/src/main/scala/com/digitalasset/daml/lf/archive/testing/DamlLfEncoder.scala
```diff
@@ -29,7 +29,7 @@ private[daml] object DamlLfEncoder extends App {
     throw new Error("You should not get this error")
   }
 
-  private val pkgId = Ref.PackageId.assertFromString("-self-")
+  private val pkgId = Ref.PackageId.assertFromString("self")
```
remyhaemmerle-da, line 32: why do we need to change this one ?

remyhaemmerle-da (approved):

daml-lf/encoder/src/main/scala/com/digitalasset/daml/lf/archive/testing/DamlLfEncoder.scala
cocreature:
```suggestion
  private val pkgId = Ref.PackageId.assertFromString("-self-")
```

cocreature: leftover from when I tried to also restrict characters, dropped it

PR #10370, opened 2021-07-22
LF: Add check of nesting in SValue.toValue

remyhaemmerle-da:
The convertion of SValue to Value already ensure the resulting value
has a serializable type. Here we add a check to ensure it do not
overpass the maximun allow nesting.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/spec/daml-lf-1.rst
```diff
@@ -2874,7 +2874,6 @@ as described by the ledger model::
    —————————————————————————————————————————————————————————————————————— EvUpdCreateErr4
      'create' @Mod:T vₜ ‖ (st₀, keys₀)  ⇓ᵤ  (Err E, ε)
 
-
```
cocreature, line 2877: ?

cocreature: On that note, we probably should add it to the spec but doesn’t have too be in this PR.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -24,10 +24,49 @@ import scala.util.hashing.MurmurHash3
...
+          value.iterator.map { case (k, v) => k.mapContractId(f) -> v.mapContractId(f) },
+        )
+      case SAny(ty, value) =>
+        SAny(ty, value.mapContractId(f))
+    }
+}
+
+object SValue {
+
+  def toValue(v: SValue, maxNesting: Int = V.MAXIMUM_NESTING): V[V.ContractId] = {
```
cocreature, line 65: should we have some unit tests for this?

daml-lf/transaction/src/main/scala/com/daml/lf/Error.scala
```diff
@@ -104,4 +104,6 @@ object Error {
 
   final case class ContractIdInContractKey(key: Value[ContractId]) extends Error
 
+  final case object TooDeepValue extends Error
```
cocreature, line 107:
```suggestion
  final case object ValueExceedsMaxNesting extends Error
```
sounds a bit clearer but I don’t feel particularly strongly about this.

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -440,13 +440,5 @@ server_conformance_test(
...
-    test_tool_args = [
-        "--include=" + ",".join([
-            "DeeplyNestedValueIT:Accept",
-            "DeeplyNestedValueIT:RejectCreateCommand",
-            "DeeplyNestedValueIT:RejectExerciseCommand",
-            "DeeplyNestedValueIT:RejectCreateArgumentInCreateAndExerciseCommand",
-            "DeeplyNestedValueIT:RejectChoiceArgumentInCreateAndExerciseCommand",
-        ]),
-    ],
+    test_tool_args = ["--include=DeeplyNestedValueIT:Accept"],
```
cocreature, line 443: Why did these change?

ledger/sandbox/BUILD.bazel
```diff
@@ -325,10 +325,7 @@ server_conformance_test(
             "DeeplyNestedValueIT:AcceptFetchByKey46",
             "DeeplyNestedValueIT:AcceptFailingLookupByKey46",
             "DeeplyNestedValueIT:AcceptSuccessfulLookupByKey46",
-            "DeeplyNestedValueIT:RejectCreateCommand",
-            "DeeplyNestedValueIT:RejectExerciseCommand",
-            "DeeplyNestedValueIT:RejectCreateArgumentInCreateAndExerciseCommand",
-            "DeeplyNestedValueIT:RejectChoiceArgumentInCreateAndExerciseCommand",
+            "DeeplyNestedValueIT:Reject",
```
cocreature, line 328: Same question here

compiler/scenario-service/client/src/DA/Daml/LF/PrettyScenario.hs
```diff
@@ -410,6 +410,8 @@ prettyScenarioErrorError (Just err) =  do
           scenarioError_ContractIdInContractKeyKey
     ScenarioErrorErrorComparableValueError _ ->
       pure "Attend to compare incomparable values"
+    ScenarioErrorErrorTooDeepValueError _ ->
```
cocreature, line 413: Can we include the actual and the max nesting value in the errors

compiler/scenario-service/client/src/DA/Daml/LF/PrettyScenario.hs
```diff
@@ -410,6 +410,8 @@ prettyScenarioErrorError (Just err) =  do
...
+          pure "Exceed maximum nesting value"
```
cocreature, line 414:
```suggestion
          pure "Value exceeds maximum nesting value of 100"
```
you could include the 100 in the error to avoid duplicating it but I think we should show it in the error.

cocreature (approved):
lgtm with a few comments, I think it would be nice to have some simple unit tests for `toValue` before merging.

ledger/sandbox-classic/BUILD.bazel
remyhaemmerle-da: I should have remove also `:Accept`. All the test passed for sandbox classic.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
remyhaemmerle-da: added

remyhaemmerle-da: 9e21da583838c14d6581a950b15e9d10386543da

ledger/sandbox/BUILD.bazel
remyhaemmerle-da: More tests pass.

compiler/scenario-service/client/src/DA/Daml/LF/PrettyScenario.hs
remyhaemmerle-da: I added the nesting statically as it would be in the spec. 
For the actual nesting of the value : 
- I am not sure it is not very useful. 
- We have to compute it from the root of the value (and not from were we throw), This new method to compute the nesting of a value need to be stack safe, and we have to pass it through the proto. I think it is not worthwhile.

cocreature (approved):

PR #10374, opened 2021-07-22
[JSON-API] Validate schema version & add minimal options for schema creation

### LINKED ISSUE #10333: Validate JSON API schema
The JSON API currently makes no attempts at doing migrations and that’s fine for something that purely works as a cache.

However, we should at least do some form of validation of the schema and provide nice errors if it fails.

An easy solution seems an extra table with an integer that we increment on every schema change.
Alternatively, just embedding the sdk version would also work but that’s potentially more aggressive since it forces a reset even if nothing changed. However, from a user pov resetting on each upgrade is probably the most sensible option so maybe that’s sufficient.

realvictorprm:
should fix #10333

changelog_begin

- [JSON-API] Schema versioning was introduced to the db schema. Because of this the field `createSchema` in the jdbcConfig was deprecated. Via the field `start-mode` you can specify:
   1. `create-only`: This is equal to the behaviour of `createSchema=true` so the db schema is created and then the application terminates.
   2. `start-only`: With this the schema version is checked and if no version or an version was found which is not equal to the internal schema version then the application terminates. This is also the replacement of `createsSchema=false`.
   3. `create-if-needed-and-start`: With this the schema version is checked and if no version or an version was found which is not equal to the internal schema version then the schema will be created/updated and the application proceeds with the startup.
   4. `create-and-start`: Similar to the first option but instead of terminating the application proceeds with the startup.

changelog_end

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
```diff
@@ -52,8 +52,19 @@ object ContractDao {
     new ContractDao(Connection.connect(jdbcDriver, jdbcUrl, username, password)(cs))
   }
 
-  def initialize(implicit log: LogHandler, sjd: SupportedJdbcDriver): ConnectionIO[Unit] =
-    sjd.queries.dropAllTablesIfExist *> sjd.queries.initDatabase
+  def initialize(checkIfExists: Boolean)(implicit
```
cocreature, line 55: I don’t think this works properly. Afaict if I start up without schema creation which is what most users do atm simply because that’s the main thing you would do before, we don’t check anything. However, that’s the most important case to check. You want to blow up if someone starts on the wrong schema.

I think the logic should be something like the following:
1. If some specifies that they want to recreate the schema, always recreate. This was the previous behavior and can be useful if for whatever reason things broke.
2. When starting up the JSON API (currently only possible if you didn’t initialize, with your PR also possible after initialization), validate that you run against a schema that you understand. I don’t think we need a flag to turn this off. If the schema has changed, you need to recreate or bad things will happen.
3. With `createSchemaIfNeeded` or whatever that flag is called, create the schema if it has the wrong version or has no version.

realvictorprm: true, completely oversaw this flaw.

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
```diff
@@ -0,0 +1,64 @@
...
+          }
+      }
+    }
+    schemaHandling match {
+      case SchemaHandling.ForceCreateAndTerminate => reinit(shouldTerminate = true)
+      case SchemaHandling.ForceCreateAndContinue => reinit(shouldTerminate = false)
+      case SchemaHandling.CheckAndTerminateIfWrong =>
+        checkVersion(shouldTerminate = true)
+      case SchemaHandling.CreateOrUpdateAndContinue =>
+        checkVersion(shouldTerminate = false)
```
realvictorprm, line 61: I already think about abstracting this `true`, `false` stuff here :thinking:

realvictorprm:
In theory we could also allow to add another case `CreateOrUpdateAndTerminate` for the use case of just checking whether it needs to be updated and if so updating it directly with suitable credentials.

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
```diff
@@ -0,0 +1,30 @@
...
+package com.daml.http
+import scalaz.std.option._
+import scalaz.syntax.traverse._
+
+private[http] sealed trait SchemaHandling
+private[http] object SchemaHandling {
+  private[http] case object ForceCreateAndTerminate extends SchemaHandling
+  private[http] case object CheckAndTerminateIfWrong extends SchemaHandling
+  private[http] case object CreateOrUpdateAndContinue extends SchemaHandling
+  private[http] case object ForceCreateAndContinue extends SchemaHandling
```
cocreature, line 13:
```suggestion
  private[http] case object CreateSchema extends SchemaHandling
  private[http] case object Start extends SchemaHandling
  private[http] case object CreateSchemaIfNeededAndStart extends SchemaHandling
  private[http] case object CreateSchemaAndStart extends SchemaHandling
```
seems a bit clearer to me

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -129,21 +129,21 @@ private[http] object JdbcConfig
       s"${indent}url -- JDBC connection URL,\n" +
       s"${indent}user -- database user name,\n" +
       s"${indent}password -- database user password,\n" +
-      s"${indent}createSchema -- boolean flag, if set to true, the process will re-create database schema and terminate immediately.\n" +
+      s"${indent}schemaHandling -- option setting how the schema should be handled. Valid options are ForceCreateAndTerminate, CheckAndTerminateIfWrong, CreateOrUpdateAndContinue, ForceCreateAndContinue.\n" +
```
cocreature, line 132: this breaks backwards compatibility this is no good.

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  import scalaz.Validation.{success, failure}
+  import scalaz.Validation
+
+  private[http] def optionalSchemaHandlingField(
```
cocreature, line 18: this needs to be backwards compatible

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
```diff
@@ -110,30 +109,30 @@ object Main {
...
+        }
+        Try(SchemaHandlingResult.fromSchemaHandling(dao, c.schemaHandling).unsafeRunSync()) match {
+          case Success(SchemaHandlingResult.Terminate) =>
+            terminateProcess()
+          case Success(SchemaHandlingResult.Continue) => ()
           case Failure(e) =>
-            logger.error("Failed creating DB schema", e)
-            terminate()
-            System.exit(ErrorCodes.StartupError)
+            logger.error("Failed processing the schema handling", e)
```
cocreature, line 123: this sounds a bit weird, I think we should say something like "failed to create the schema", "schema version mismatch, expected X got Y", …..

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
```diff
@@ -0,0 +1,64 @@
...
+      dao.transact(ContractDao.initialize).map { _ =>
+        logger.info("DB schema created...")
+        SchemaHandlingResult.fromBool(shouldTerminate)
+      }
+    }
+    def checkVersion(shouldTerminate: Boolean): IO[SchemaHandlingResult] = {
+      logger.info("Checking for existing schema")
+      dao.transact(jdbcDriver.queries.version()).flatMap {
+        case None =>
+          logger.info("No schema version found")
```
cocreature, line 37:
```suggestion
          logger.info("No schema version found")
```
when terminating it is probably also nice to suggest to run with `createSchema=true` to create the schema.

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
```diff
@@ -0,0 +1,64 @@
...
+          if (shouldTerminate) terminate
+          else reinit(shouldTerminate = false)
+        case Some(version) =>
+          logger.info(s"DB schema version $version found")
+          if (version < jdbcDriver.queries.schemaVersion) {
```
cocreature, line 42:
```suggestion
          if (version != jdbcDriver.queries.schemaVersion) {
```
if you downgrade things are also broken. any mismatch needs to fail

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
```diff
@@ -0,0 +1,64 @@
...
+package com.daml.http
+
+import com.daml.http.dbbackend.ContractDao
+import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
+import cats.effect.IO
+import com.daml.http.util.Logging.InstanceUUID
+
+sealed trait SchemaHandlingResult
+object SchemaHandlingResult {
+  case object Terminate extends SchemaHandlingResult
```
cocreature, line 13: The logic here seems a bit awkward. I would suggest something like the following:
1. Have a flag for whether you want to start the server after schema initialization. This can either be a separate flag in `Config` or a `shouldStart` method on SchemaHandling. Handle this at the call site not inside `fromSchemaHandling`.
2. Rename `fromSchemaHandling` to something like `createAndValidateSchema` and depending on the flag only validate, always recreate or recreate if needed.
3. `createAndValidateSchema` either succeeds, in that case continue checking `shouldStart` and start if it’s true, otherwise exit with code 0. If createAndValidateSchema fails, always terminate with a non-zero exit code.

cocreature (changes requested):
Looks pretty close, I think the biggest issue is that you broke backwards compatibility which needs to be fixed before merging. The rest is mostly style nitpicks

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
realvictorprm: I'm not super happy with `Start`, it sounds a bit weird to me in the config. I think `JustStart` somehow sounds sensible, although it is still not the best naming. :thinking: :confused:

cocreature: I chose the names to make it clear that CreateSchemaAndStart is exactly doing first CreateSchema and then Start but I don’t feel all that strongly about the specific names.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
realvictorprm: I deprecated the old flag (so it's still there but advised to use the new one) and adjusted the info with cf1632f

ledger-service/http-json/src/main/scala/com/digitalasset/http/Main.scala
realvictorprm: I improved the logging on the processing side such that any additional logging at this location is not necessary anymore, see 34669f7

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
realvictorprm: I added suggestions to the log messages with 988e6a6

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
realvictorprm: Well let's go with it. If someone comes up with better names we can change it before the next release.

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
```diff
@@ -0,0 +1,30 @@
...
+  private[http] case object CreateSchemaIfNeededAndStart extends SchemaHandling
+  private[http] case object CreateSchemaAndStart extends SchemaHandling
+
+  import scalaz.Validation.{success, failure}
+  import scalaz.Validation
+
+  private[http] def optionalSchemaHandlingField(
+      m: Map[String, String]
+  )(k: String): Either[String, Option[SchemaHandling]] = {
+    val parse: String => Validation[String, SchemaHandling] = {
```
cocreature, line 21: Looking around a bit more, we have a fairly similar option for Daml on SQL already https://github.com/digital-asset/daml/blob/c3a3d600832bf1926a2421302e31abc1219c8f5a/ledger/sandbox-common/src/main/scala/platform/sandbox/config/PostgresStartupMode.scala#L9.

The names don’t quite work but I think we can at least follow the format of using lower-case hyphenated names.

I think we can probably also drop the `Schema` given that it’s clear you’re talking about the db from the option name and then we could go from `create` to `create-only` although I don’t feel particularly strongly about the last one.

1. `create-only`
2. `start

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -129,21 +131,22 @@ private[http] object JdbcConfig
       s"${indent}url -- JDBC connection URL,\n" +
       s"${indent}user -- database user name,\n" +
       s"${indent}password -- database user password,\n" +
-      s"${indent}createSchema -- boolean flag, if set to true, the process will re-create database schema and terminate immediately.\n" +
+      s"${indent}createSchema -- boolean flag, if set to true, the process will re-create database schema and terminate immediately. This is deprecated and replaced by schemaHandling, however if set it will always overrule schemaHandling.\n" +
+      s"${indent}schemaHandling -- option setting how the schema should be handled. Valid options are CreateSchema, Start, CreateSchemaIfNeededAndStart, CreateSchemaAndStart.\n" +
```
cocreature, line 135: Daml on SQL calls a very similar option `sql-start-mode` https://github.com/digital-asset/daml/blob/c3a3d600832bf1926a2421302e31abc1219c8f5a/ledger/sandbox-common/src/main/scala/platform/sandbox/config/PostgresStartupMode.scala#L9
Maybe consider using the same here?

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
cocreature: @realvictorprm you resolved this comment but you didn’t address it fully. You still conflate errors during initialization (e.g. version mismatch) with the user selecting to exit after creating the schema and both turn into exit code 0.

My suggestion above was to only handle "error vs no error" here and check if you continue and the call site rather than in here.

ledger-service/http-json-cli/src/main/scala/com/daml/http/SchemaHandling.scala
realvictorprm: "Ist gebongt", consistency is good :)

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
realvictorprm: Yeah, still in favour of consistency. Even if the schema name might get lost here.

ledger-service/http-json/src/main/scala/com/digitalasset/http/SchemaHandlingResult.scala
realvictorprm: True, I did not respect the error codes. They were not as important in my mind yet, but I see why it is important to keep these right. Will do a fancy fix.

realvictorprm: I think this should be OK now, the exit codes are now fixed.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
realvictorprm: Because we are already in the `jdbcConfig` I named it just `start-mode` (prior I even named it `startup-mode` but to preserve consistency I went for the other).

realvictorprm:
Last open point is to fix the scala 2.12 build, will take another look into that after lunch.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -160,23 +164,32 @@ private[http] object JdbcConfig
       url <- requiredField(x)("url")
       user <- requiredField(x)("user")
       password <- requiredField(x)("password")
-      createSchema <- optionalBooleanField(x)("createSchema")
+      createSchema <- optionalBooleanField(x)("createSchema").map(
+        _.map { createSchema =>
+          import DbStartupMode._
+          logger.warn(
+            s"The option 'createSchema' is deprecated. Please use for 'createSchema=true' => 'start-mode=${getConfigValue(CreateOnly)}' and for 'createSchema=false' => 'start-mode=${getConfigValue(StartOnly)}'"
```
cocreature, line 171:
```suggestion
            s"The option 'createSchema' is deprecated. Please use 'start-mode=${getConfigValue(CreateOnly)}' for 'createSchema=true' and 'start-mode=${getConfigValue(StartOnly)}'  for 'createSchema=false'"
```
sounds wrong otherwise but maybe my English is just borked

ledger-service/http-json/src/main/scala/com/digitalasset/http/DbStartupResult.scala
```diff
@@ -0,0 +1,78 @@
...
+package com.daml.http
+
+import com.daml.http.dbbackend.ContractDao
+import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
+import cats.effect.IO
+import com.daml.http.util.Logging.InstanceUUID
+
+sealed trait DbStartupResult
+object DbStartupResult {
+  case object GracefullyExit extends DbStartupResult
```
cocreature, line 13: you really don’t seem to like my suggestion to move the logic for this to the call site :grin: fair enough I don’t care too much

ledger-service/http-json/src/main/scala/com/digitalasset/http/DbStartupResult.scala
```diff
@@ -0,0 +1,78 @@
...
+  case object Continue extends DbStartupResult
+
+  private def fromBool(shouldGracefullyExit: Boolean): DbStartupResult =
+    if (shouldGracefullyExit) GracefullyExit else Continue
+
+  private[this] val logger = ContextualizedLogger.get(getClass)
+
+  def fromStartupMode(dao: ContractDao, dbStartupMode: DbStartupMode)(implicit
```
cocreature, line 21: This deserves a comment on the meaning of `None` and the difference between successfully returning `None` and throwing an exception. Looking at the code, it seems the difference is fairly arbitrary and the callsite handles both more or less the same so you could also consider if you really need to represent those two types of errors differently.

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
```diff
@@ -129,21 +131,23 @@ private[http] object JdbcConfig
       s"${indent}url -- JDBC connection URL,\n" +
       s"${indent}user -- database user name,\n" +
       s"${indent}password -- database user password,\n" +
-      s"${indent}createSchema -- boolean flag, if set to true, the process will re-create database schema and terminate immediately.\n" +
+      s"${indent}createSchema -- boolean flag, if set to true, the process will re-create database schema and terminate immediately. This is deprecated and replaced by start-mode, however if set it will always overrule start-mode.\n" +
+      s"${indent}start-mode -- option setting how the schema should be handled. Valid options are ${DbStartupMode.allConfigValues
```
cocreature, line 135: @S11001001 is there some logic behind where we use camelCase and where we use kebab-case?
It seems like currently the option names are all kebab-case but then in the option-value we use camelCase?

If it was just the start mode, I’d say switching to kebab-case would make sense since then we’re consistent everywhere but it’s not. The websocket config still uses camelCase.

cocreature (approved):
Nice, thank you!

ledger-service/http-json-cli/src/main/scala/com/daml/http/Config.scala
realvictorprm: I think you're right, thanks!

realvictorprm:
@cocreature OK, I decoupled the code a bit more, improved the tests further & made sure that all exceptions are properly handled with correct exit codes.

ledger-service/http-json/src/main/scala/com/digitalasset/http/DbStartupOps.scala
```diff
@@ -0,0 +1,106 @@
...
+      dao: ContractDao,
+      lc: LoggingContextOf[InstanceUUID],
+  ): IO[Option[DbVersionState]] = {
+    import dao.jdbcDriver
+    for {
+      _ <- IO(logger.info("Checking for existing schema in the DB"))
+      currentSchemaVersion = jdbcDriver.queries.schemaVersion
+      getVersionResult <- dao.transact(jdbcDriver.queries.version()).attempt
+    } yield getVersionResult
+      .fold(
```
cocreature, line 35: probably for a separate PR: maybe consider returning the error in an `Either` instead of returning `None`. That makes it very clear what the meaning of `Left` is where as `None` can be anything.

cocreature (approved):

ledger-service/http-json/src/main/scala/com/digitalasset/http/DbStartupOps.scala
realvictorprm: Deal :) I hesitated to go with an `Either[Unit, _]` :sweat_smile: But I agree that with `Option` is is more confusing right now.

cocreature: `Either[Throwable, _]` should also work I believe.

realvictorprm: Adjusted that small nit with my last build fix :) Hopefully the build finally passes :smile:

PR #10376, opened 2021-07-22
Introduce locally-defined `Rejection` reasons instead of the participant-state rejection reason type. [KVL-1002]

SamirTalwar-DA:
Instead of using the `RejectionReasonV0` enumeration defined in _participant-state_, we change the code to use new enumerations that are specific to the ledger driver. For each component, we define an enumeration named `Rejection`, encompassing all possible error states, and convert these cases to the _participant-state_ rejection when necessary.

The components in question are the Indexer (both the current indexer and the append-only one), Sandbox Classic, and kvutils.

Defining these enumerations makes the migration to the participant state v2 API a little easier, mostly because we have fewer dependencies, but also because the data is now structured, which will make it easier to construct valid v2 command rejections containing the [gRPC `ErrorInfo` type](https://github.com/googleapis/googleapis/blob/3859723d77717f5082e78b4b0565c6d08b3eedfe/google/rpc/error_details.proto#L88-L136).

Constructing those `ErrorInfo` objects will come in a future changeset.

### Pull Request Checklist

trigger the build.


miklos-da (approved):

andreaslochbihler-da (approved):
Looks good from the Canton perspective.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/Rejection.scala
```diff
@@ -0,0 +1,56 @@
...
+  val definiteAnswer: Boolean
+
+  def toDomainRejectionReason: domain.RejectionReason
+
+  def toStateV1RejectionReason: v1.RejectionReason =
+    toDomainRejectionReason.toParticipantStateRejectionReason
+}
+
+object Rejection {
+  val rejectionDomain = "com.daml.ledger"
```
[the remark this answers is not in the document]
SamirTalwar-DA: I shouldn't have committed this code; it's for v2 and I decided against including it in this PR. Deleting it.

fabiotudone-da (commented):
How about adding unit tests for the conversions?

SamirTalwar-DA:
@fabiotudone-da: Good call. I was being lazy because these conversions are temporary, but I think tests are still important because they'll force me to test the changes too.

ledger/participant-integration-api/src/test/suite/scala/platform/store/ConversionsSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package com.daml.platform.store
+
+import com.daml.ledger.api.domain
+import com.daml.ledger.participant.state.{v1 => state}
+import com.daml.platform.store.Conversions._
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AsyncWordSpec
+
+class ConversionsSpec extends AsyncWordSpec with Matchers {
```
SamirTalwar-DA, line 12: @fabiotudone-da: Tests are here.

fabiotudone-da: How about the `toStateV1RejectionReason` conversions?

Issue #10029, opened 2021-06-16
Removing dependencies on older LF versions


cocreature:
Currently a DAR built in LF 1.14 (or really any LF version) still pulls in dependencies on LF 1.6 (and versions in between) for types like `Archive`. You can see the full list of packages at https://github.com/digital-asset/daml/blob/7a2a349daf5e53e1f2643f63cdeda8ef34847049/compiler/damlc/stable-packages/lib/DA/Daml/StablePackages.hs#L21 (note that the exception packages are already in LF 1.14 so they are unaffected).

This has worked fine so far but now we have a concrete reason for a ledger to want to drop support for older LF versions: Canton would like to drop support for everything < LF 1.14 before going GA. This allows them to drop a number of backwards compatibility hacks around `byKey`, old contract ids and other stuff. This needs to happen before going GA since at that point dropping support is basically impossible (outside of major upgrades at least).

More generally, we should also expect that this is not the last time we want to drop support for an old LF version so the changes here should be repeatable for future deprecations.

## Changes for Daml Users
All of this only applies when targetting LF version 1.14. If you target an older LF version absolutely nothing changes.

* Packages will automatically depend on stable packages compiled in LF 1.14. These will all be mapped under `DA.Stable.V1_14.*`. E.g., `DA.Stable.V1_14.Archive` would contain the archive type. `DA.Stable.V1_14.Types` tuples and so on.
* `daml-stdlib` reexports the new types where we currently reexport the older types.
* The old types are still reexported in `daml-stdlib` under their old module names. Only if you reference them will you pull in the old stable packages as a dependency.
* For upgrades, you can simply import both of them. Qualified imports allow you to disambiguate.
* `data-dependencies` will automatically map the type to the correct module.
* Typeclass instances for the old types will have to be pulled in via an import of an old LF version. Current daml-stdlib will only have the instances for the newer types. This is necessary because we need to avoid having a dependency from daml-stdlib on the older LF packages.
* We will slightly relax the import restrictions to allow importing the modules where the old types currently live in directly even if they are under `DA.Internal`. The name makes no sense anyway since those modules are part of the public API via the ledger API.

## Breakage/Impact
1. This is a compiler change that only affects newer LF versions so unless you recompile in LF >= 1.14 there is no impact at all.
[...]

PR #10377, opened 2021-07-22
Let stable packages bypass LF version restrictions.

sofiafaro-da:
Part of #10029

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
@@ -69,6 +69,10 @@ object LanguageVersion {
   val EarlyAccessVersions: VersionRange[LanguageVersion] =
     StableVersions
 
+  // Only the current stable version, no older versions.
+  val StableOnlyVersions: VersionRange[LanguageVersion] =
+    StableVersions.copy(min = StableVersions.max)
+
   // All the versions
   val DevVersions: VersionRange[LanguageVersion] =
     StableVersions.copy(max = v1_dev)
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
@@ -69,10 +69,6 @@ object LanguageVersion {
   val EarlyAccessVersions: VersionRange[LanguageVersion] =
     StableVersions
 
-  // Only the current stable version, no older versions.
-  val StableOnlyVersions: VersionRange[LanguageVersion] =
-    StableVersions.copy(min = StableVersions.max)
-
   // All the versions
   val DevVersions: VersionRange[LanguageVersion] =
     StableVersions.copy(max = v1_dev)
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
@@ -10,28 +10,111 @@ object StablePackages {
   // TODO: Use this map to generate the mapping used in the test,
   // or generate both from a single source.
   val map = Map[PackageName, PackageId](
-    PackageName.assertFromString("daml-prim-DA-Exception-ArithmeticError-cb0552debf219cc909f51cbb5c3b41e9981d39f8f645b1f35e2ef5be2e0b858a") -> PackageId.assertFromString("cb0552debf219cc909f51cbb5c3b41e9981d39f8f645b1f35e2ef5be2e0b858a"),
-    PackageName.assertFromString("daml-prim-DA-Exception-AssertionFailed-3f4deaf145a15cdcfa762c058005e2edb9baa75bb7f95a4f8f6f937378e86415") -> PackageId.assertFromString("3f4deaf145a15cdcfa762c058005e2edb9baa75bb7f95a4f8f6f937378e86415"),
-    PackageName.assertFromString("daml-prim-DA-Exception-GeneralError-86828b9843465f419db1ef8a8ee741d1eef645df02375ebf509cdc8c3ddd16cb") -> PackageId.assertFromString("86828b9843465f419db1ef8a8ee741d1eef645df02375ebf509cdc8c3ddd16cb"),
-    PackageName.assertFromString("daml-prim-DA-Exception-PreconditionFailed-f20de1e4e37b92280264c08bf15eca0be0bc5babd7a7b5e574997f154c00cb78") -> PackageId.assertFromString("f20de1e4e37b92280264c08bf15eca0be0bc5babd7a7b5e574997f154c00cb78"),
-    PackageName.assertFromString("daml-prim-DA-Internal-Erased-76bf0fd12bd945762a01f8fc5bbcdfa4d0ff20f8762af490f8f41d6237c6524f") -> PackageId.assertFromString("76bf0fd12bd945762a01f8fc5bbcdfa4d0ff20f8762af490f8f41d6237c6524f"),
-    PackageName.assertFromString("daml-prim-DA-Internal-PromotedText-d58cf9939847921b2aab78eaa7b427dc4c649d25e6bee3c749ace4c3f52f5c97") -> PackageId.assertFromString("d58cf9939847921b2aab78eaa7b427dc4c649d25e6bee3c749ace4c3f52f5c97"),
-    PackageName.assertFromString("daml-prim-DA-Types-40f452260bef3f29dede136108fc08a88d5a5250310281067087da6f0baddff7") -> PackageId.assertFromString("40f452260bef3f29dede136108fc08a88d5a5250310281067087da6f0baddff7"),
-    PackageName.assertFromString("daml-prim-GHC-Prim-e491352788e56ca4603acc411ffe1a49fefd76ed8b163af86cf5ee5f4c38645b") -> PackageId.assertFromString("e491352788e56ca4603acc411ffe1a49fefd76ed8b163af86cf5ee5f4c38645b"),
-    PackageName.assertFromString("daml-prim-GHC-Tuple-6839a6d3d430c569b2425e9391717b44ca324b88ba621d597778811b2d05031d") -> PackageId.assertFromString("6839a6d3d430c569b2425e9391717b44ca324b88ba621d597778811b2d05031d"),
-    PackageName.assertFromString("daml-prim-GHC-Types-518032f41fd0175461b35ae0c9691e08b4aea55e62915f8360af2cc7a1f2ba6c") -> PackageId.assertFromString("518032f41fd0175461b35ae0c9691e08b4aea55e62915f8360af2cc7a1f2ba6c"),
-    PackageName.assertFromString("daml-stdlib-DA-Date-Types-bfcd37bd6b84768e86e432f5f6c33e25d9e7724a9d42e33875ff74f6348e733f") -> PackageId.assertFromString("bfcd37bd6b84768e86e432f5f6c33e25d9e7724a9d42e33875ff74f6348e733f"),
-    PackageName.assertFromString("daml-stdlib-DA-Internal-Any-cc348d369011362a5190fe96dd1f0dfbc697fdfd10e382b9e9666f0da05961b7") -> PackageId.assertFromString("cc348d369011362a5190fe96dd1f0dfbc697fdfd10e382b9e9666f0da05961b7"),
-    PackageName.assertFromString("daml-stdlib-DA-Internal-Down-057eed1fd48c238491b8ea06b9b5bf85a5d4c9275dd3f6183e0e6b01730cc2ba") -> PackageId.assertFromString("057eed1fd48c238491b8ea06b9b5bf85a5d4c9275dd3f6183e0e6b01730cc2ba"),
-    PackageName.assertFromString("daml-stdlib-DA-Internal-Template-d14e08374fc7197d6a0de468c968ae8ba3aadbf9315476fd39071831f5923662") -> PackageId.assertFromString("d14e08374fc7197d6a0de468c968ae8ba3aadbf9315476fd39071831f5923662"),
-    PackageName.assertFromString("daml-stdlib-DA-Logic-Types-c1f1f00558799eec139fb4f4c76f95fb52fa1837a5dd29600baa1c8ed1bdccfd") -> PackageId.assertFromString("c1f1f00558799eec139fb4f4c76f95fb52fa1837a5dd29600baa1c8ed1bdccfd"),
-    PackageName.assertFromString("daml-stdlib-DA-Monoid-Types-6c2c0667393c5f92f1885163068cd31800d2264eb088eb6fc740e11241b2bf06") -> PackageId.assertFromString("6c2c0667393c5f92f1885163068cd31800d2264eb088eb6fc740e11241b2bf06"),
-    PackageName.assertFromString("daml-stdlib-DA-NonEmpty-Types-e22bce619ae24ca3b8e6519281cb5a33b64b3190cc763248b4c3f9ad5087a92c") -> PackageId.assertFromString("e22bce619ae24ca3b8e6519281cb5a33b64b3190cc763248b4c3f9ad5087a92c"),
-    PackageName.assertFromString("daml-stdlib-DA-Semigroup-Types-8a7806365bbd98d88b4c13832ebfa305f6abaeaf32cfa2b7dd25c4fa489b79fb") -> PackageId.assertFromString("8a7806365bbd98d88b4c13832ebfa305f6abaeaf32cfa2b7dd25c4fa489b79fb"),
-    PackageName.assertFromString("daml-stdlib-DA-Set-Types-97b883cd8a2b7f49f90d5d39c981cf6e110cf1f1c64427a28a6d58ec88c43657") -> PackageId.assertFromString("97b883cd8a2b7f49f90d5d39c981cf6e110cf1f1c64427a28a6d58ec88c43657"),
-    PackageName.assertFromString("daml-stdlib-DA-Time-Types-733e38d36a2759688a4b2c4cec69d48e7b55ecc8dedc8067b815926c917a182a") -> PackageId.assertFromString("733e38d36a2759688a4b2c4cec69d48e7b55ecc8dedc8067b815926c917a182a"),
-    PackageName.assertFromString("daml-stdlib-DA-Validation-Types-99a2705ed38c1c26cbb8fe7acf36bbf626668e167a33335de932599219e0a235") -> PackageId.assertFromString("99a2705ed38c1c26cbb8fe7acf36bbf626668e167a33335de932599219e0a235")
+    PackageName.assertFromString(
+      "daml-prim-DA-Exception-ArithmeticError-cb0552debf219cc909f51cbb5c3b41e9981d39f8f645b1f35e2ef5be2e0b858a"
+    ) -> PackageId.assertFromString(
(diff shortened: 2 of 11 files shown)


remyhaemmerle-da (commented):
Look pretty good to me. 2 remarks thought:
+ Not completely sure it is worthwhile to parametrize the engine with stable package IDs.
+ we need tests.

cocreature (commented):
Looks very nice, thank you!
I tend to agree with @remyhaemmerle-da that the set of stable packages doesn’t have to be an option and arguably shouldn’t be since it is really an implementation detail

As for tests, one way would be to run the conformance tests with a ledger limited to LF 1.14 on LF >= 1.14. That’s arguably the most useful test in that it actually tests that a package produced by the compiler goes through the full stack. I think that might be slightly tricky to setup since we don’t have CLI flags to control those limits but adding those to ledger on memory for example seems perfectly reasonable.

Alternatively, a unit test somewhere is also fine with me.

ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -143,6 +143,18 @@ conformance_test(
...
+conformance_test(
+    name = "conformance-test-stable-only",
+    ports = [6865],
+    server = ":app",
+    server_args = [
+        "--contract-id-seeding=testing-weak",
+        "--participant=participant-id=example,port=6865",
+        "--stable-only",
+    ],
+    lf_versions = ["1.14"] # TODO: change to ["stable"] once 1.14 is considered stable in bazel
```
sofiafaro-da, line 155: Depends on https://github.com/digital-asset/daml/pull/10394

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
```diff
@@ -69,6 +69,10 @@ object LanguageVersion {
   val EarlyAccessVersions: VersionRange[LanguageVersion] =
     StableVersions
 
+  // Only the current stable version, no older versions.
+  val StableOnlyVersions: VersionRange[LanguageVersion] =
+    StableVersions.copy(min = StableVersions.max)
+
```
remyhaemmerle-da, line 75: What is the use case of this ?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -491,6 +491,14 @@ object Config {
...
           )
 
+        opt[Unit]("stable-only")
+          .optional()
+          .hidden()
+          .action((_, c) => c.copy(engineMode = EngineMode.StableOnly))
+          .text(
+            "Disable the loading of unstable LF packages outside of the current LF stable version. Should not be used in production."
+          )
+
```
remyhaemmerle-da, line 501: Why do we want to change the behavior of KV ?

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: I found that pretty weird this is potentially changing at each realize. 
Those constant are part of the Engine API, if there is not a solid reason to have a new one, the client should simply define is own range.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
sofiafaro-da: This is the parser that ledger-on-memory uses. The point is to be able to test without loading old (unstable) packages.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
sofiafaro-da: To be able to run things without enabling loading of old LF packages. As a preview for removing support for old versions.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
```diff
@@ -69,6 +69,10 @@ object LanguageVersion {
   val EarlyAccessVersions: VersionRange[LanguageVersion] =
     StableVersions
 
+  // Only the current stable version, no older versions.
+  val StableOnlyVersions: VersionRange[LanguageVersion] =
```
remyhaemmerle-da, line 73:
```suggestion
  val LastestOnly: VersionRange[LanguageVersion] =
```

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
```diff
@@ -69,6 +69,10 @@ object LanguageVersion {
   val EarlyAccessVersions: VersionRange[LanguageVersion] =
     StableVersions
 
+  // Only the current stable version, no older versions.
```
remyhaemmerle-da, line 72:
```suggestion
  // Only the latest stable version, no older versions.
```

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: In engine world Stable means non developments.

sofiafaro-da: Isn't that more confusing, with the distinction between stable and latest version?

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: > As a preview for removing support for old versions.
We do not really need to preview removing support for old versions. In general we will never do it. Th only use case I see in the foreseeable future (before daml 2) is canton that will not support version older that 14. 

In any case I do not really see the point to define a constant, the ledger can defined is own range.

sofiafaro-da: >  The only use case I see in the foreseeable future (before daml 2) is canton.

That's why I'm adding this now, and we can remove it once it's not useful. What's the reason to not define the range here?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
remyhaemmerle-da: OK the purpose is test. In this case I do not think  we need the constant `EngineMode.StableOnly`, we can defined it manually here.

I am a bit concern of changing the code of ledger-on-memory, to test a feature that will be used by canton only.
I think having a unit test of the Compiler and ask canton to add a test to be sure they refuse loading old LF version is enough.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: Canton will needs two ranges: 

```
  val CantonStableVersions: VersionRange[LanguageVersion] =
    StableRange.copy(min = v1_14)  
 
 val CantonDevVersions: VersionRange[LanguageVersion] =
    DevRange.copy(min = v1_14)
```

The min version should be fixed  and not changing every time we change `StableVersions.max` (which potentially changes at each release).

This is canton specific it should be defined in canton code, there is no reason to defined it here.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
sofiafaro-da: I don't understand what you mean by "define it manually here". Could you elaborate?

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: More confusing, for sure no ;). 
You maybe right, that it is maybe not less confusing. 
What about  `LatestStableVersion` ?

sofiafaro-da: I mean, ... what does a "latest stable version" mean when we have "latest version" and a "stable version" and these aren't always the same thing? 🤔

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
remyhaemmerle-da: In  `ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala` you could write : 
```
     case EngineMode.StableOnly => LanguageVersion.Stable.copy(max = LanguageVersion.Stable.max)
```

sofiafaro-da: Yes, but that still requires defining `EngineMode.StableOnly`?

cocreature: Sorry for only getting to this so late. How about adding a `min-lf-version` flag that allows you to set the minimum required LF version? That way we can set it to whatever we want in tests, we don’t need any constants in `Engine` and we don’t have to debate whether we call it latest, stable or latest stable?

cocreature: and you can replace `EngineMode` by `VersionRange[LanguageVersion]`.

remyhaemmerle-da: This is a separate issue. 

I think we should not have  `EngineMode.StableOnly` neither because I do not think we should change the code of KV for testing a feature that will be used by Canton only, but : 
- I care less 
- It should be approved by @miklos-da teams.

cocreature: It’s sort of related since a `min-lf-version` flag requires representing some form of range (or at least a lower bound) and at that point the redundancy between a new constructor to `EngineMode` and the existing constructors seems to do more harm than good. But yeah, I don’t care all that much either.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/LanguageVersion.scala
remyhaemmerle-da: The word "stable" in the code of the engine has a different meaning that in the world "stable" in the compiler code  
- "stable" in the engine means not development or nor preview version (there is no stable version, there are stable versions)
- "stable" in the compiler mean default output version. (there is only one stable version)
 
In this precise context there it does not make sens to add `OnlyStableVersions` because `StableVersions` already contains only stable versions. 

I am open to clean up this inconsistency but in a different PR (potentially merged before).

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
remyhaemmerle-da: @moritzkiefer-da > This is a separate issue.

remyhaemmerle-da: Despite the concern I express before, I will not veto changes in KV code. 
By the way I slightly prefer the `min-lf` flag.

sofiafaro-da: I added a `--daml-lf-min-version-1.14` flag in lieu of a full version parser.

cocreature (approved):
Nice, thank you!

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -486,19 +487,31 @@ object Config {
...
+          .optional()
+          .hidden()
+          .action((_, c) =>
+            c.copy(allowedLanguageVersions =
+              c.allowedLanguageVersions.copy(min = LanguageVersion.v1_14)
+            )
+          )
+          .text(
+            "Set minimum LF version for unstable packages to 1.14. Should not be used in production."
+          )
```
remyhaemmerle-da, line 513: @mziolekda @miklos-da 
This PR adds a feature needed by canton. However we need a way to test it through the full stack. 
Are you fine we add this flag to be able to use the `ledger-on-memory` to test the feature ?

remyhaemmerle-da (approved):
I would like some feedback from KV team.
However, If you do not get any by tonight could you please merge (we need the feature in tomorrow release) ? 
If some feedback arrive latter we will tackle that in another PR.

ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -143,6 +143,18 @@ conformance_test(
...
+    name = "conformance-test-min-1.14",
+    lf_versions = ["1.14"],
+    ports = [6865],
+    server = ":app",
+    server_args = [
+        "--contract-id-seeding=testing-weak",
+        "--participant=participant-id=example,port=6865",
+        "--daml-lf-min-version-1.14",
+    ],
+)
```
remyhaemmerle-da, line 156: a (non-conformance) test that actually checks that (non-stable) packages  older than 1.14 are rejected would be a nice to have.

sofiafaro-da: Hmm, there is already some tests for this at the engine level:

https://github.com/digital-asset/daml/blob/4b55f1ab6a7734886bb6ecc4bec5568119ccf17b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala#L2683

But there's room for a test of the new flag `--daml-lf-min-version-1.14`

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -486,19 +487,31 @@ object Config {
...
         opt[Unit]("daml-lf-dev-mode-unsafe")
           .optional()
           .hidden()
-          .action((_, c) => c.copy(engineMode = EngineMode.Dev))
+          .action((_, c) => c.copy(allowedLanguageVersions = LanguageVersion.DevVersions))
           .text(
             "Enable the development version of the Daml-LF language. Highly unstable. Should not be used in production."
           )
 
+        opt[Unit]("daml-lf-min-version-1.14")
```
miklos-da, line 503:
```suggestion
        opt[Unit]("daml-lf-min-version-1.14-unsafe")
```

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -486,19 +487,31 @@ object Config {
 
         opt[Unit]("early-access")
           .optional()
-          .action((_, c) => c.copy(engineMode = EngineMode.EarlyAccess))
+          .action((_, c) => c.copy(allowedLanguageVersions = LanguageVersion.EarlyAccessVersions))
```
miklos-da, line 490: This increases coupling between Daml-LF and kvutils-app config — we may have to add this back later.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf.language
+
+import com.daml.lf.data.Ref.{PackageId, PackageName}
+
+object StablePackages {
+  // Based on compiler/damlc/tests/src/stable-packages.sh
+  // TODO: Use this map to generate the mapping used in the test,
```
miklos-da, line 10: What are 'stable' packages?

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
miklos-da: Yes, with the above small change. Though, it would be preferred to parse the actual version so that we don't have to modify this command-line parameter for the next Daml-LF version.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
```diff
@@ -0,0 +1,120 @@
...
+  // or generate both from a single source.
+  val map = Map[PackageName, PackageId](
```
miklos-da, line 12: We don't need to expose this.
```suggestion
  private val nameToIdMapping = Map[PackageName, PackageId](
```

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
```diff
@@ -0,0 +1,120 @@
...
+    ) -> PackageId.assertFromString(
+      "733e38d36a2759688a4b2c4cec69d48e7b55ecc8dedc8067b815926c917a182a"
+    ),
+    PackageName.assertFromString(
+      "daml-stdlib-DA-Validation-Types-99a2705ed38c1c26cbb8fe7acf36bbf626668e167a33335de932599219e0a235"
+    ) -> PackageId.assertFromString(
+      "99a2705ed38c1c26cbb8fe7acf36bbf626668e167a33335de932599219e0a235"
+    ),
+  )
+  val ids = map.values.toSet
```
miklos-da, line 119:
```suggestion
  val Ids = map.values.toSet
```
(in the spirit of the [Scala style guide](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables))

miklos-da (approved):

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/StablePackages.scala
sofiafaro-da: Stable packages are LF packages that are part of the standard library, but don't change with SDK version. They expose Daml types that we want to remain stable across SDK upgrades.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
sofiafaro-da: > Though, it would preferred to parse the actual version so that we don't have to modify this command-line parameter for the next Daml-LF version.

I agree that it would be preferable to parse the actual min version, but note that there's no plan currently for the 1.14 minimum to change in the future, even when the next LF version is released.

PR #10380, opened 2021-07-22
Use pre-execution workflow for ledger-on-sql [KVL-812]

nicu-da:
This switches ledger-on-sql (and implicitly sandbox) to use the more efficient pre-execution workflow instead of the classic workflow

CHANGELOG_BEGIN
ledger-on-sql and sandbox are now using the pre-execution workflow
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
```diff
@@ -77,11 +75,6 @@ object SqlLedgerFactory extends LedgerFactory[ReadWriteService, ExtraConfig] {
         engine = engine,
         jdbcUrl = jdbcUrl,
         resetOnStartup = false,
-        logEntryIdAllocator = RandomLogEntryIdAllocator,
-        stateValueCache = caching.WeightedCache.from(
-          configuration = config.stateValueCache,
-          metrics = metrics.daml.kvutils.submission.validator.stateValueCache,
-        ),
```
SamirTalwar-DA, line 84: I think we need to put these lines back in.

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/SqlLedgerReaderWriter.scala
```diff
@@ -112,19 +132,30 @@ object SqlLedgerReaderWriter {
...
-          validator = validator,
-          postCommit = dispatcher.signalNewHead,
+        ledgerDataExporter <- LedgerDataExporter.Owner.acquire()
+        committer = new Committer(
+          transformStateReader = ledgerStateReader =>
+            CachingStateReader(
+              stateValueCache,
+              ImmutablesOnlyCacheUpdatePolicy,
+              DamlLedgerStateReader
+                .from(ledgerStateReader, StateKeySerializationStrategy.createDefault()),
```
SamirTalwar-DA, line 147: Please inject the `keySerializationStrategy` instead of constructing it here (and below).

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/SqlLedgerReaderWriter.scala
```diff
@@ -112,19 +132,30 @@ object SqlLedgerReaderWriter {
...
-          inStaticTimeMode = timeProvider != TimeProvider.UTC,
+        sqlLedgerStateAccess = new SqlLedgerStateAccess(database, metrics)
+        keyValueCommitting = new KeyValueCommitting(
+          engine,
+          metrics,
         )
-        committer = new ValidatingCommitter[Index](
-          () => timeProvider.getCurrentTime,
-          validator = validator,
-          postCommit = dispatcher.signalNewHead,
```
SamirTalwar-DA, line 127: Looks like we will never call `dispatcher.signalNewHead` any more. In the in-memory ledger this call is done after `commit`, but I'm not sure that's possible here.

You may need to adjust `PreExecutingValidatingCommitter` and `RawPostExecutionWriter` to return a tuple of `(Index, SubmissionResult)`. (The index can be retrieved from `LedgerStateOperations#appendToLog`.)

This should also simplify the in-memory code; you'll be able to delete the `InMemoryState#newHeadSinceLastWrite` method.

SamirTalwar-DA (commented):
Looks like good progress. Your tests are failing because you aren't signalling a new head on the `dispatcher`. I have a suggestion for how you can fix that in my comments.

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
nicu-da: Ah yes, forgot to readd the built cache. Good catch

ledger/ledger-on-sql/src/main/scala/com/daml/ledger/on/sql/SqlLedgerReaderWriter.scala
nicu-da: Got it, thanks for the explanation.

SamirTalwar-DA:
This issue is blocked on a problem with static time mode when running with pre-execution.

The problem: command deduplication doesn't work in static time mode. Once a command is submitted, any subsequent command with the same command ID will fail. This is because of the way we compute the time bounds for pre-execution.

In `LedgerTimeValidator`, the minimum time bound is computed as `max(deduplicateUntil, ledgerTime - maxSkew, recordTime - maxSkew)`. The maximum time bound is computed as `min(ledgerTime + minSkew, recordTime + minSkew)`.

This is an issue once we have a potential deduplication (i.e. a previous command), because `deduplicateUntil` uses wall clock time, whereas both `ledgerTime` and `recordTime` use the time mode specified. This means that the minimum record time can be `2021-07-27T07:00:00`, and the maximum ledger time will be `1970-01-01T00:01:00`, resulting in an out-of-bounds error.

In addition, the engine populates the submission time, which becomes the `ledgerTime` in the validator, as the same as the "ledger effective time", making these two fields identical.

I have spiked a quick "fix" that solves this problem, but not properly. To do so, I had to:
  1. Alter `CommandDeduplicationIT` to use `submitAndWait` instead of submit to correctly identify the problem.
  2. "Correct" the engine to accept the submission time as a parameter. I don't know if this is a good idea.
  3. Change `LedgerTimeValidator` to ignore the ledger time if it's in static time mode.
  4. Change `ConfigCommitter` to disable the TTL check in static time mode.
  5. Change the post execution write set selector in ledger-on-sql to use the current clock time, not the time from the time mode. (i.e. always wall clock time)

The changes are in https://github.com/digital-asset/daml/commit/c4d11ecd572447cbf5e2c1a01a93baa402d82d3e.

This "fix" is definitely not reasonable, because it disables the TTL check in the `ConfigCommitter`. (Other stuff may also not be reasonable.)

I believe the core problem here is that we are using two clocks at once without separating the concerns. We want to ensure that submissions have an expiry time, which is always going to be against a wall clock, not the time mode. However, we often use times provided by the time mode in these kinds of situations.

I think we need to isolate the command deduplication and TTL checks and make sure that they never reference the same values as the other time-based checks using the record time (e.g. causal monotonicity checks).

With the exception of the engine issue I mentioned above (no idea what the correct behavior should be), I think the issues here are constrained to KV.

Although static time mode is a somewhat niche feature, it's very valuable for testing and demonstrations, and therefore we need to get it right.

miklos-da:
Closing for now due to unclear timeline for resolving issue with static time mode & pre-execution.

PR #10381, opened 2021-07-22
treat exerciseResult structurally for isReplayedBy

nickchapman-da:
older versions of LF didn’t have the exerciseResult https://github.com/digital-asset/daml/commit/d725d50be98ce064c39d760d3658dced853b2427 so once the engine needed to produce this, we needed to allow for the recorded to not be present while the engine produces it. However, we don’t support that LF version anymore and on the other hand, with rollback nodes, the exerciseResult has become optional again so I think this is just broken and shouldn’t allow for None, Some(_).

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


cocreature (approved):
looks very reasonable thanks!
I’d like @remyhaemmerle-da to confirm that I’m not missing anything here

remyhaemmerle-da (approved):
LGTM. Thanks.

PR #10382, opened 2021-07-22
LF: drop old serializability check for Values.

remyhaemmerle-da:
We drop this check for the following reasons:

* Its only usage was inside the transaction builder, this is subsumed
  by the check we added in the SValue to Value translation (#10370)

* It seems to origianl check more that nesting (and so to be
  overcomplicate)

* It was used in non-consitent way. It was used to check create
  arguments and choice arguments, but niether choice result nor
  contract key.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -918,7 +918,7 @@ private[lf] object SBuiltin {
       mbKey.foreach { case Node.KeyWithMaintainers(key, maintainers) =>
         if (maintainers.isEmpty)
           throw SErrorDamlException(
-            IE.CreateEmptyContractKeyMaintainers(templateId, createArg.toValue, key)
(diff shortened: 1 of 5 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -918,7 +918,7 @@ private[lf] object SBuiltin {
       mbKey.foreach { case Node.KeyWithMaintainers(key, maintainers) =>
         if (maintainers.isEmpty)
           throw SErrorDamlException(
-            IE.CreateEmptyContractKeyMaintainers(templateId, createArg.toValue, key)
+            IE.CreateEmptyContractKeyMaintainers(templateId, createArgValue, key)
```
remyhaemmerle-da, line 921: Just a minor not directly related to the rest of the PR. 
Avoid recomputation already done at line 913

cocreature (approved):
Nice cleanup, thanks!

PR #10384, opened 2021-07-22
Issue: 10383 - Support deadline for command submission

dparmar:
### Pull Request Checklist

trigger the build.


cocreature (commented):
Hi @dparmar, thank you very much for the PR!

I think it would be nice to have a testcase before we merge this. You should be able to add something to https://github.com/digital-asset/daml/blob/ad13a86947df7fcb9746a3579cfcd92011f1e9e6/language-support/java/bindings-rxjava/src/test/scala/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImplTest.scala#L35 where you pass a future to https://github.com/digital-asset/daml/blob/ad13a86947df7fcb9746a3579cfcd92011f1e9e6/language-support/java/bindings-rxjava/src/test/scala/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImplTest.scala#L36 which never completes and you verify that the request still returns quickly whereas it blocks if you do not send a timeline.

There are also a few design considerations:
1. I think you probably want to be able to overwrite this per request.
2. As you mentioned, this should really apply to all (unidirectional) services.

I don’t think those are blockers if you do not want to address them in this PR.

dparmar:
Hi @cocreature , Thanks a lot for the review comments! I've pushed the changes to include test case. The two design considerations you mentioned (i.e. deadline for every unary calls and per call) would be very useful but I would prefer to address them outside this PR just to keep it simple considering my limited efficiency with coding in Scala :)

cocreature (approved):
Thanks! Looks good

cocreature:
Apologies, I accidentally closed this and I can’t reopen it for some reason. I recreated this at https://github.com/digital-asset/daml/pull/10522

dparmar:
Hi @cocreature,
While testing this feature yesterday, I realized my understanding was wrong and the idea of using a single deadline object does not work when using the same _serviceStub_ to send more than one requests. i.e. deadline starts ticking the moment it is created, which works fine for the first request but times out subsequent requests. 

After reading the comments on  [grpc issue 4305]( https://github.com/grpc/grpc-java/issues/4305), I believe ClientInterceptor is a good way to implement this feature. Hence I created [PR 10692](https://github.com/digital-asset/daml/pull/10692/).

PR #10391, opened 2021-07-23
Export import

### LINKED ISSUE #10379: daml ledger export: build sometimes fails due to "Import of internal module DA.Internal.Template is not allowed."
Note: this issue is a sibling to https://github.com/digital-asset/daml/issues/10378 as it can only be encountered after the path issue has been manually resolved.

Sometimes, the export execution produces an import which causes the build to fail.

This doesn't occur 100% of the time and I don't understand the cause, but when it does occur it requires the exporter code to be modified in order for it to build.

A sample build failure looks like this:

```$ daml build
Compiling export to a DAR.
File:     Export.daml
Hidden:   no
Range:    8:1-8:38
Source:   parser
Severity: DsError
Message:  Import of internal module DA.Internal.Template is not allowed.
ERROR: Creation of DAR file failed.
```

Removing the import line allows the build to succeed.

aherrmann-da:
Closes #10379 

Avoids imports of internal modules in the generated Daml ledger export script.

- Factor out `moduleRefs` which constructs the set of module imports based on references in ACS and transaction tree.
- Adds unit tests for `moduleRefs`.
- Filters out internal modules according to `isInternal` in `compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
@@ -446,7 +446,17 @@ object TreeUtils {
   ): Set[String] = {
     val fromAcs = acs.foldMap(ev => valueRefs(Sum.Record(ev.getCreateArguments)))
     val fromTrees = trees.foldMap(treeRefs(_))
-    (fromAcs ++ fromTrees).map(_.moduleName)
+    (fromAcs ++ fromTrees)
+      .map(_.moduleName)
+      .filterNot(isInternalModule)
+  }
+
+  private def isInternalModule(moduleName: String): Boolean = {
+    // Keep in sync with isInternal in compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
+    moduleName.startsWith("DA.Internal.") ||
+    moduleName.startsWith("GHC.") ||
+    Set("Control.Exception.Base", "Data.String", "LibraryModules", "DA.Types", "DA.Time.Types")
+      .contains(moduleName)
   }
 
   def treeRefs(t: TransactionTree): Set[Identifier] =
--- a/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
@@ -440,6 +440,15 @@ object TreeUtils {
     )
   }
 
+  def moduleRefs(
+      acs: Iterable[CreatedEvent],
+      trees: Iterable[TransactionTree],
+  ): Set[String] = {
+    val fromAcs = acs.foldMap(ev => valueRefs(Sum.Record(ev.getCreateArguments)))
+    val fromTrees = trees.foldMap(treeRefs(_))
+    (fromAcs ++ fromTrees).map(_.moduleName)
+  }
+
   def treeRefs(t: TransactionTree): Set[Identifier] =
     t.eventsById.values.foldMap(e => evRefs(e.kind))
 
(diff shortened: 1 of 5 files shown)


daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
```diff
@@ -440,6 +440,25 @@ object TreeUtils {
...
+    (fromAcs ++ fromTrees)
+      .map(_.moduleName)
+      .filterNot(isInternalModule)
+  }
+
+  private def isInternalModule(moduleName: String): Boolean = {
+    // Keep in sync with isInternal in compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
+    moduleName.startsWith("DA.Internal.") ||
+    moduleName.startsWith("GHC.") ||
+    Set("Control.Exception.Base", "Data.String", "LibraryModules", "DA.Types", "DA.Time.Types")
```
cocreature, line 458: Are you sure this works correctly? `DA.Time.Types` defines `RelTime` which is not exported by `Prelude`. As far as I can tell if I have a reference to `RelTime` this will now just result in an import error.

For `DA.Internal.Template` you are lucky because the `Archive` type is reexported from `Prelude`.

I haven’t checked if there are other types to which this applies in addition to `RelTime` but I suspect so.

The easiest fix would be to allow imports of these modules. While that sounds perhaps a bit odd, I think you can actually justify this:
1. The module names are already part of the public API if you’re working with the ledger API.
2. Any attempts at hiding constructors also don’t apply when working over the ledger API so I think providing an option to construct things without going through smart constructors directly in Daml is only reasonable.

wdyt? @sofiafaro-da I’d also appreciate your input here

sofiafaro-da: Another possibility is to rewrite these imports:

- DA.Time.Types -> DA.Time
- GHC.Stack.Types -> DA.Stack
- DA.Internal.Exception -> DA.Exception
- DA.Internal.Record -> DA.Record

Everything else seems to be exposed in Prelude, is hidden, or is special (e.g. Erased and PromotedText).

Making these four modules non-internal would be difficult without making all the other modules non-internal. But I'm also ok with us dropping the concept of "internal module" (that can't be imported) in the preprocessor, since that's not a thing in LF. The module name and lack of official docs should be enough to dissuade people to not import them directly unless they know what they're doing and are fine with being burned.

cocreature: Rewriting imports works only partially, For Daml script exports we will still fallover when we need to use the constructor for `RelTime` for example which is currently hidden from `DA.Time`.

sofiafaro-da: Got it. I'm happy for us to remove the internal module distinction altogether. (In any case, internal modules should not be considered a "stable" part of our standard library. Except for the ones that are in stable packages.)

cocreature: I think for the purposes of this discussion, stable packages are the only ones that matter. We will only get references to serializable types from the ledger and those are all in stable packages. Fully agree that having those in internal modules doesn’t make sense.

aherrmann-da: Yes, that's a good point. In case of `RelTime` we could also issue `DA.Time.convertMicrosecondsToRelTime` to work around the issue. But, trying to keep up with all the necessary rewrites sounds brittle.
All in favor of simply allowing these imports.
I'm happy to look into this.

PR #10395, opened 2021-07-23
add triggers to gsg

garyverhaegen-da:
Our team has recently had two new joiners, both of which were asked to develop a small Daml application as part of their onboarding, and both of which ended up with a great use-case for triggers while not knowing they existed. I also needed to look into how triggers work for unrelated reasons, so I thought it would make sense to add triggers to the GSG.

This is of course not meant to cover all of triggers (we already have doc pages for that), but just to make sure new Daml users know triggers exist.

CHANGELOG_BEGIN
CHANGELOG_END

--

See compiled version of the docs [here](https://garyverhaegen-da.github.io/docs-gsg-triggers/getting-started/triggers.html). I clearly need some guidance on how to do code blocks properly so they get the nice highlighting (and correct indentation for snippets), as well as possibly some guidance on how to make these part of some sort of test suite so at the very least we can check that the code keeps compiling across versions.

### DIFF
--- a/docs/source/getting-started/triggers.rst
+++ b/docs/source/getting-started/triggers.rst
@@ -16,7 +16,7 @@ Objective
 
 We're going to build a chatbot that answers evry message with
 
-  "Please tell me more about that."
+  "Please, tell me more about that."
 
 That should fool anyone and pass the Turing test, easily.
 
@@ -163,7 +163,7 @@ Open up the trigger code again (``daml/ChatBot.daml``), and change it to::
                 case replyTo of
                   None -> pure ()
                   Some (sender, _) -> do
-                    T.dedupExercise sender (User.SendMessage p "tell me more")
+                    T.dedupExercise sender (User.SendMessage p "Please, tell me more about that.")
               else pure ()
             None -> pure ()
       , registeredTemplates = T.AllInDar
--- a/docs/source/getting-started/triggers.rst
+++ b/docs/source/getting-started/triggers.rst
@@ -14,7 +14,7 @@ automation: :doc:`Daml Triggers </triggers/index>`.
 Objective
 *********
 
-We're going to build a chatbot that answers evry message with
+We're going to build a chatbot that answers every message with
 
   "Please, tell me more about that."
 
@@ -172,11 +172,11 @@ Open up the trigger code again (``daml/ChatBot.daml``), and change it to::
 
 Refresh ``daml start`` by pressing ``r`` (followed by ``Enter`` on Windows) in
 its terminal, then kill (CTRL-C) the trigger and start it again. Play a bit
-with ``alice`` and ``bob`` in your browser again, to get a feel for how the
+with ``alice`` and ``bob`` in your browser, to get a feel for how the
 trigger works. Watch both the messages in-browser and the debug statements
 printed by the trigger runner.
 
-Walking through the rule code line by line gives:
+Let's walk through the ``rule`` code lline-by-line:
 
 - We use the ``query`` function to get all of the ``Message`` templates visible
   to the current party (``p``; in our case this will be ``bob``). Per the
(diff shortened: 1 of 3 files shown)


nemanja-da:
While triggers are definitely useful they feel out-of-scope for a getting started guide. There's already plenty of things a noob has to go through 

- installing the SDK and the dependencies, which is not that easy
- running the app
- adding a new feature on the Daml and UI side

GSG was always meant to be just a way for someone to get a quick grasp of Daml (as you mention nothing exhaustive). I'm really not in favor of adding triggers to the GSG, think this should've been discussed with the DevRel team really as it impacts our OKR/KPI

nemanja-da:
The example would be a great addition to the `Triggers` section of the docs as it uses the same example as in the GSG though.

cocreature:
> GSG was always meant to be just a way for someone to get a quick grasp of Daml (as you mention nothing exhaustive). I'm really not in favor of adding triggers to the GSG, think this should've been discussed with the DevRel team really as it impacts our OKR/KPI

Let’s discuss after @garyverhaegen-da is back from vacation in a week. I think your concerns are valid but I do think that there is a real problem that we’re lacking "next steps" once you finish the current GSG and I think this section fits in there perfectly. I think if we structure it correctly it can just be an addition rather than an extra step someone has to go through.

cocreature:
Ah missed that you’re also on vacation, in that case I’ll discuss with Gary how we can integrate it without modifying the GSG and then we can revisit how we link to it from the gsg once you’re back.

nemanja-da:
We can discuss it, but I wouldn't say that pointing someone straight away to triggers is the way to go. Users still need to learn the basics of Daml really + how to deploy. Triggers are just a part if it. We're working on creating a better learning path on daml.com, should think of the docs as well

cocreature:
I don’t think this is limited to triggers but there is a problem that once you finished the GSG, you’re more or less stuck and it’s unclear to proceed. If you look at the recent onboarding daml apps that have been written all of them include a trigger somewhere so I think this is one possible next step, deployment & the daml intro are others. I think it stops making sense to treat it linearly after the current GSG.

nemanja-da:
Sure, let's discuss how to proceed after someone has completed the GSG outside of this PR

docs/source/getting-started/triggers.rst
```diff
@@ -0,0 +1,265 @@
...
+              if m.receiver == p
+              then do
+                users <- T.query @User.User
+                debug users
+                let isSender = (\user -> user.username == m.sender)
+                let replyTo = List.head $ filter (\(_, user) -> isSender user) users
+                case replyTo of
+                  None -> pure ()
+                  Some (sender, _) -> do
+                    T.dedupExercise sender (User.SendMessage p "Please, tell me more about that.")
```
cocreature, line 166: This dedup is fairly unreliable/slightly broken. Consider the following steps:

1. We get message M1 and respond here by sending a command asynchronously. Now we dedup on `(sender, SendMessage p "Please, tell me more about that.")`.
2. Now we get message M2 with a later time. We haven’t received a completion for M1 yet so dedup kicks in and the message gets droped.
3. Now we get a completion for M1 and dedup stops.

Arguably humans are slow at typing so the practical impacts of this are probably small but I don’t really like the idea of using a broken example in the GSG.

docs/source/getting-started/triggers.rst
```diff
@@ -0,0 +1,265 @@
...
+Objective
+*********
+
+We're going to build a chatbot that answers every message with
+
+  "Please, tell me more about that."
+
+That should fool anyone and pass the Turing test, easily.
+
+A No-Op Trigger
```
cocreature, line 23: At the risk of sidetracking this I wonder if an autofollow trigger that follows everyone that follows you might be a better example:
1. It doesn’t require any changes to the model.
2. It is easy to make it robust.
3. The code itself should be simpler as well.

garyverhaegen-da:
Closing per #10419.

PR #10397, opened 2021-07-23
Allow imports of internal modules

### LINKED ISSUE #10379: daml ledger export: build sometimes fails due to "Import of internal module DA.Internal.Template is not allowed."

aherrmann-da:
As discussed in https://github.com/digital-asset/daml/pull/10391#discussion_r675526700 changes the Daml compiler to no longer error on imports of internal modules.
Closes #10391
Closes #10379

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
+++ b/compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
@@ -8,9 +8,11 @@ module DA.Daml.Preprocessor
   , noPreprocessor
   ) where
 
+import           DA.Daml.LFConversion (convertModuleName)
 import           DA.Daml.Preprocessor.Records
 import           DA.Daml.Preprocessor.Generics
 import           DA.Daml.Preprocessor.EnumType
+import           DA.Daml.StablePackages (stablePackageByModuleName)
 
 import Development.IDE.Types.Options
 import qualified "ghc-lib" GHC
@@ -46,6 +48,11 @@ isInternal (GHC.moduleNameString -> x)
       , "DA.Types"
       , "DA.Time.Types"
       ]
+
+isUnstableInternal :: GHC.ModuleName -> Bool
+isUnstableInternal moduleName =
+  isInternal moduleName &&
+    convertModuleName moduleName `Map.notMember` stablePackageByModuleName
 
 preprocessorExceptions :: Set.Set GHC.ModuleName
 preprocessorExceptions = Set.fromList $ map GHC.mkModuleName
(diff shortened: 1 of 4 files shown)


cocreature (commented):
Have you checked if we run into any issues just allowing imports of the stable packages? Not necessarily suggesting that is a better idea but it’s at least sufficient for Daml Script exports so I think it would be useful to know if it works or if there are other issues with it.

aherrmann-da:
> Have you checked if we run into any issues just allowing imports of the stable packages? Not necessarily suggesting that is a better idea but it’s at least sufficient for Daml Script exports so I think it would be useful to know if it works or if there are other issues with it.

I've changed the PR accordingly and confirmed on the repro of #10379 that the generated Daml script, which contains an import of `DA.Internal.Template`, compiles successfully.

compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
```diff
@@ -163,8 +170,8 @@ checkModuleName (GHC.L _ m)
 -- | We ban people from importing modules such
 checkImports :: GHC.ParsedSource -> [(GHC.SrcSpan, String)]
 checkImports x =
-    [ (ss, "Import of internal module " ++ GHC.moduleNameString m ++ " is not allowed.")
-    | GHC.L ss GHC.ImportDecl{ideclName=GHC.L _ m} <- GHC.hsmodImports $ GHC.unLoc x, isInternal m]
+    [ (ss, "Import of unstable internal module " ++ GHC.moduleNameString m ++ " is not allowed.")
```
cocreature, line 173:
```suggestion
    [ (ss, "Import of internal module " ++ GHC.moduleNameString m ++ " is not allowed.")
```
I’d just keep it and say that the other ones are not internal even if they might have `Internal` in the name for historic reasons.

cocreature (approved):
Thanks!

compiler/damlc/daml-preprocessor/src/DA/Daml/Preprocessor.hs
aherrmann-da: Done

sofiafaro-da (approved):
Looks great, thank you!

PR #7618, opened 2020-10-08
upgrade doobie from 0.6.0 to 0.9.2

S11001001:
We've left the doobie version alone for a while; since pool exhaustion is on my mind, let's upgrade and then take extra care about how we set up our pools, more by being explicit than changing behavior.

The old `Fragment.const` behavior is now on `Fragment.const0`, but the new `const` seems like a safer default, so we leave it be.

Resources:

* [migration notes](https://tpolecat.github.io/doobie/migration.html)
* [latest release notes](https://github.com/tpolecat/doobie/releases)
* [earlier release notes](https://github.com/tpolecat/doobie/blob/master/CHANGELOG.md)
* ...the source code

First:

* [x] merge #7661

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10399, opened 2021-07-23
use doobie 0.9.0 Fragment-in-Fragment interpolation in json-api db-backend

S11001001:
Since tpolecat/doobie#1045 (and therefore #7618) doobie has supported interpolating fragments in fragments.  We've used this feature for several fragments written since #7618, but have left the ones written before alone to use `++`.  Here we change that where it meaningfully clarifies the SQL subexpression.

Note that this _does not_ entail a `Put` or `Write` instance for `Fragment`.  You cannot abstract over `Fragment` and arbitrary interpolated data in this way, because `Fragment`s are not treated as positional parameters; that would mean being able to put arbitrary SQL substrings in positional parameters.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -42,12 +42,12 @@ sealed abstract class Queries {
     sql"""
       CREATE TABLE
         contract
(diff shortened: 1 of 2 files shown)


realvictorprm (approved):
I shouldnt work anymore but couldn't resist.

Seems all fine to me as everything seems to be just a simple migration to interpolated sql strings 😃

cocreature (commented):
good stuff

PR #10402, opened 2021-07-25
ContractStorageBackend consolidation [DPP-462]

nmarton-da:
* Adding CompletionStorageBackendTemplate
* Adding ContractStorageBackendTemplate
* Splitting EventStrategy and creating QueryStrategy
* Adapting Backend implementations

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    import com.daml.platform.store.Conversions.HashToStatement
+    val sql =
+      s"""  WITH last_contract_key_create AS (
+         |         SELECT participant_events.*
+         |           FROM participant_events, parameters
+         |          WHERE event_kind = 10 -- create
+         |            AND create_key_hash = {keyHash}
+         |            AND event_sequential_id <= parameters.ledger_end_sequential_id
```
rautenrieth-da, line 405: Do you need this line if you have the `$validAtClause` below? I would have expected this line to be folded into the `validAtClause` if `validAt` is `None`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -0,0 +1,432 @@
...
+         |          FETCH NEXT 1 ROW ONLY
+         |       )
+         |  SELECT ${resultColumns.mkString(", ")}
+         |    FROM last_contract_key_create -- creation only, as divulged contracts cannot be fetched by key
+         |  WHERE $lastContractKeyFlatEventWitnessesClause -- check visibility only here
+         |    NOT EXISTS       -- check no archival visible
+         |         (SELECT 1
+         |            FROM participant_events, parameters
+         |           WHERE event_kind = 20 AND -- consuming exercise
+         |             event_sequential_id <= parameters.ledger_end_sequential_id AND
```
rautenrieth-da, line 418: Same here with `validAtClause`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -0,0 +1,432 @@
...
+        |         SELECT contract_id, ${resultColumns.mkString(", ")}
+        |           FROM participant_events, parameters
+        |          WHERE contract_id = {contractId}
+        |            AND event_kind = 10  -- create
+        |            AND event_sequential_id <= parameters.ledger_end_sequential_id
+        |          FETCH NEXT 1 ROW ONLY -- limit here to guide planner wrt expected number of results
+        |       ),
+        |       divulged_contract AS (
+        |         SELECT divulgence_events.contract_id,
+        |                -- Note: the divulgance_event.template_id can be NULL
```
rautenrieth-da, line 271:
```suggestion
        |                -- Note: the divulgence_event.template_id can be NULL
```

rautenrieth-da: Also, I think the comment can be rewritten more generally as:
```
[[ReadService]] integrations may omit communicating the full details of a divulged contract via [[TransactionAccepted.divulgedContracts]] if they know the participant has already seen the create event. Some columns (such as the template id and create argument) can therefore be NULL in the divulgence event.

If the divulgence event is missing some contract details, they can therefore be substituted from the corresponding create event.
```

rautenrieth-da (approved):
I have left some minor comments. Looks good in general, but as this PR touches queries and is hard to review, it would be good if we ran performance tests before merging.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -132,6 +97,24 @@ private[backend] object H2StorageBackend
...
+    override def arrayIntersectionNonEmptyClause(
+        columnName: String,
+        parties: Set[Party],
+        paramNamePostfix: String,
+    ): (String, List[NamedParameter]) =
+      (
+        if (parties.isEmpty)
+          "false"
+        else
+          parties.view.map(p => s"array_contains($columnName, '$p')").mkString("(", " or ", ")"),
```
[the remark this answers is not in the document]
nmarton-da: I think it could, H2 has an array_length function, so we can build here an sql query ensuring non-empty intersection of those arrays. Not sure it will be less or more efficient since the array_intersection is a scala code binding to H2.
It was not my goal to improve much on the H2 part of the query building. For example: in Oracle and Postgres I also put more emphasize to pass arguments as data instead of SQL code as well -- if that is the motivation behind the question.
If needed I would be ok to improve on the H2 part here in this PR on in a separate one as well.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/QueryStrategy.scala
```diff
@@ -0,0 +1,48 @@
...
+      .getOrElse("" -> Nil)
+
+  /** An expression resulting to a boolean to check whether:
+    *   - the party set defined by columnName and
+    *   - the party set defined by parties
+    * have at least one element in common (eg their intersection is non empty).
+    *
+    * @param columnName the SQL table definition which holds the set of parties
+    * @param parties set of parties
+    * @param paramNamePostfix a postfix to prevent collision with anorm NamedParameter-s
```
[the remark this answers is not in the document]
nmarton-da: I am not sure that is practical: to be able to do this on the caller side it would involve parsing the resturned clause string and replacing names in there, and also mapping the NamedParams similarily. Or maybe I am missing your point?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
nmarton-da: Probably not. But to do this we also need to remove the join as well. That's why I was thinking:
* it is simpler
* and also safer
To just let the two conditions be merged to one in the SQL engine.
Provided that it is not significantly less efficient which would motivate tradeoff on safety. (so benchmarks need to verify this cc @tudor-da )

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/QueryStrategy.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  /** This populates the following part of the query:
+    *   SELECT ... WHERE ... ORDER BY ... [THIS PART]
+    *
+    * @param limit optional limit
+    * @return a tuple:
+    *         - first part is a plain SQL which fits the query template
+    *         - second part is a list of NamedParameters
+    *         Important note: the NamedParameters should confirm to the generated SQL expression (all names there should match exactly)
+    *         Important note: implementor have to make sure that all used names are unique here in the scope of the whole query
```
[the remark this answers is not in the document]
nmarton-da: it is probably the other way around (like the implementor of the client of this needs to make sure not to use the names which occupied here in this function), but "globally" is scary and simple  enough :)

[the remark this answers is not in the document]
rautenrieth-da: - The caller of this function doesn't know what parameter names are being generated by this function.
- The implementor of this function doesn't know what other parameter names are used by the final query.

The only possibly way to choose a parameter name that is guaranteed to not collide with any other parameter name in the final query is to choose a globally unique string.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/QueryStrategy.scala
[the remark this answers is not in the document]
rautenrieth-da: True, mapping the NamedParams is easy, but parsing the query string is not. 

How about at least making sure all functions that generate clauses (i.e., ones that return `(String, List[NamedParameter])`) are given a unique suffix? I don't like the fact that helpers like `EventStorageBackendTemplate.witnessesWhereClause` or `QueryStrategy.limitClause` have to pick a suffix that is unique within the whole query, without knowing anything about the query.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
rautenrieth-da: Makes sense :+1:

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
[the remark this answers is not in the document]
rautenrieth-da: I was mainly asking because we use `array_intersection` in `H2EventStrategy.filteredEventWitnessesClause` a few lines below. No need to change anything in this PR, especially since it's not clear whether it's better.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -592,222 +584,6 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
 
   // Contracts
 
-  def contractKeyGlobally(key: Key)(connection: Connection): Option[ContractId] = {
```
mziolekda, line 595: //Contracts comment above is now orphaned

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -158,19 +142,30 @@ private[backend] object H2StorageBackend
...
           .map { case ((parties, templateIds), index) =>
+            val (clause, params) =
+              H2QueryStrategy.arrayIntersectionNonEmptyClause(
+                witnessesColumnName,
+                parties,
+                s"ptwwc$index",
+              )
             (
-              s"( (${arrayIntersectionWhereClause(witnessesColumnName, parties)}) AND (template_id = ANY({templateIdsArraywwc$index})) )",
+              s"( ($clause) AND (template_id = ANY({templateIdsArraywwc$index})) )",
```
mziolekda, line 165: nit: would it not make sense to maintain the same order for clauses and parameters?

mziolekda:
I have been able to review this today. Seems like Robert has done a very good job already weeding out all problems, because I have not spotted anything new despite going over every query and its usage meticulously.
Looks good

mziolekda (approved):

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
nmarton-da: with the new approach this comment is obsolete

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/QueryStrategy.scala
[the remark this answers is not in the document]
nmarton-da: Yeah this whole named parameters and composition is something I did not like much, so I took now a different approach to respect simplicity and related comments here.
* I was going back to the foundations with anorm, and observed with the SQL"" string interpolation we could eliminate all related issues at once
* Also noticed this is considered safer and faster approach compared to the SQL() style named parameter approach by anorm too (truly: the string interpolation is using macros + scala-string interpolation, and the named parameter is using parser combinators to rip apart strings)
* I draw a custom string interpolation layer on top of anorm, supporting SQL"" and cSQL"" interpolations where the latter results to CompositeSql-s which we can be used in the former, which behaves same as the anorm-counterpart. Simply put: this enables composition of string-interpolation style expressed anorm constructs.
* Quite happy with the results: client code got much easier, the whole naming/prefixing/collision problem is just gone, actually removed code...with the trade-off to have a non trivial string-fusing tree traversal algorithm to support nested CompositeSql-s. (if only anorm would support this by desing...)

check it out!

rautenrieth-da (approved):
The new query composition looks good, nice work! My only comment is that it would be good to have unit tests for `ComposableQuery`.

nmarton-da:
Added unit tests for the string interpolation flattening.
Switched to optional join on the parameters table in contract key queries due to degradation in performance benchmark results.

nmarton-da:
@tudor-da confirmed that LR run tests are now same as baseline. Merging

PR #10398, opened 2021-07-23
participant-integration-api: Switch to the v2 participant state API. [KVL-1002]

SamirTalwar-DA:
In which we [draw the rest of the owl](https://i.kym-cdn.com/photos/images/original/000/572/078/d6d.jpg).

Switching to the v2 participant state API means:

  - the API server and indexer expect v2 API traits
    - adapters are provided if you need to elevate your v1 API usage
  - the indexer internally uses v2 Updates
  - rejections are switched over to use the v2 format
  - Sandbox Classic uses v2 as the internal representation too (because it interacts directly with the underlying indexer representation, and is therefore tightly coupled)

_kvutils_ and other users of the `StandaloneApiServer` and `StandaloneIndexerServer` use the adapters.

### Changelog

- **[Integration Kit]** The API server and indexer have switched over to v2 of the participant-state API. You can continue to use the v1 API, but you will need to wrap your `ReadService` and `WriteService` objects in the `AdaptedV1ReadService` and `AdaptedV1WriteSerivce` classes.

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10405, opened 2021-07-26
Keep the participant-state API prefixed. [KVL-1002]

SamirTalwar-DA:
This makes it easier to switch from the v1 to the v2 API.

I also think it's clearer than prefixing with "v1", and the prefix may help make it clearer what we're referring to.

Extracted from #10398.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServices.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServices.scala
@@ -15,7 +15,7 @@ import com.daml.ledger.api.health.HealthChecks
 import com.daml.ledger.api.v1.command_completion_service.CompletionEndRequest
 import com.daml.ledger.client.services.commands.CommandSubmissionFlow
 import com.daml.ledger.participant.state.index.v2._
-import com.daml.ledger.participant.state.v1.WriteService
+import com.daml.ledger.participant.state.{v1 => state}
 import com.daml.ledger.resources.{Resource, ResourceContext, ResourceOwner}
 import com.daml.lf.data.Ref
 import com.daml.lf.engine._
@@ -71,7 +71,7 @@ private[daml] object ApiServices {
 
   final class Owner(
       participantId: Ref.ParticipantId,
-      optWriteService: Option[WriteService],
+      optWriteService: Option[state.WriteService],
       indexService: IndexService,
       authorizer: Authorizer,
       engine: Engine,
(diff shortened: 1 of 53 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServices.scala
```diff
@@ -15,7 +15,7 @@ import com.daml.ledger.api.health.HealthChecks
 import com.daml.ledger.api.v1.command_completion_service.CompletionEndRequest
 import com.daml.ledger.client.services.commands.CommandSubmissionFlow
 import com.daml.ledger.participant.state.index.v2._
-import com.daml.ledger.participant.state.v1.WriteService
+import com.daml.ledger.participant.state.{v1 => state}
```
fabiotudone-da, line 18: How about `v1state`?

fabiotudone-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServices.scala
fabiotudone-da: Or simply leaving `v1.state`

SamirTalwar-DA: Well, that defeats the point. A later commit will replace this line with:

```scala
import com.daml.ledger.participant.state.{v2 => state} 
```

The point is that the rest of the code then doesn't need to change.

fabiotudone-da: Isn't it useful to also read which version of the state the code is dealing with (without scrolling up to the imports, that is)?

fabiotudone-da: (Even though there are really also other ways to know in an IDE, so a very minor point)

SamirTalwar-DA: I'm not convinced it is. I don't particularly care about the versions of anything else I use; why is this different?

fabiotudone-da: Because it's our code so it's about orientation; still, I think it's minor.

miklos-da: The reference `state` should be always to the latest version of the participant state API, as we won't be supporting different versions apart from having adaptors till we deprecate the previous version.

miklos-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServices.scala
fabiotudone-da: > apart from having adaptors till we deprecate the previous version

Alright.

PR #10408, opened 2021-07-26
Windows dev-env Powershell 7 compatibility

aherrmann-da:
Testing a build of the daml repository on Windows server 2019 revealed that the Windows dev-env is not compatible with the later Powershell version 7 installed on that Windows version. This PR fixes the offending code.

- `Get-ChildItem` was used to list all manifest files assuming that it would return relative paths. However, on the later Windows version absolute paths are returned. Fixed by passing `-Name` to only return the filename (the listing is non-recursive).
- SSL certificate validation was disabled in a way that triggered the issue described in PowerShell/PowerShell#13597. An installation from scratch works without the validation check disabled, so it seems that this is no longer required.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/dev-env/windows/libexec/core.ps1
+++ b/dev-env/windows/libexec/core.ps1
@@ -49,7 +49,7 @@ function da_clear_buckets {
 
 function da_sync_buckets([String] $Directory) {
     da_clear_buckets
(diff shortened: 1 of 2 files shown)


SamirTalwar-DA (approved):

cocreature (approved):
Thanks!

PR #10411, opened 2021-07-26
Introduce buf checks [KVL-980]

miklos-da:
Summary of changes:
* Added `buf` to nix configuration.
* Introduced basic `buf` configuration for checking against breaking in Daml-LF and kvutils protos (see details [here](https://docs.buf.build/breaking-rules) on what is being checked).
* Hooked up running `buf check breaking` via `fmt.sh`.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA:
Perhaps instead of checking in _buf_image.bin_, this could be done inside the _compatibility_ checks, and we could generate the images from various versions on the fly?

nix/tools/buf/default.nix
```diff
@@ -0,0 +1,20 @@
+{ stdenv, fetchurl }:
```
cocreature, line 1: do you need this specific version? the version of nixpkgs already has `buf` in version 0.43.2 so unless there is a reason to stick to this old version, let’s just go with that. If we do need the old version, I would suggest to add a comment why we need it.

fmt.sh
```diff
@@ -72,11 +79,13 @@ USAGE
       javafmt_args=(--set-exit-if-changed --dry-run)
       dade_copyright_arg=check
       buildifier_target=//:buildifier
+      buf_args=(check breaking --against "${buf_image}")
       ;;
     --diff)
       shift
       merge_base="$(git merge-base origin/main HEAD)"
       scalafmt_args+=('--mode=diff' "--diff-branch=${merge_base}")
+      buf_args=(check breaking --against "${buf_image}")
```
cocreature, line 88: I’m not sure this makes sense. `--diff` shouldn’t change the semantics. It’s a fast version of running without `--diff` that only checks files that changed. With your PR `--diff` acts like `--test` for `buf`. Given that people tend to want to run `./fmt.sh` before committing, I don’t think this should update the image. Instead, I’d put that in a separate script or a separate flag in this script and then default to checking the diff for both no flag, `--diff` and `--test`.

nix/tools/buf/default.nix
miklos-da: No, I don't need this specific version. Switched to using the nixpgs provided one.

fmt.sh
miklos-da: Good point. Added an option `rebuild-buf-image` to `fmt.sh` and replaced the default behaviour to be always just running a `buf check breaking`.

cocreature (approved):
Thanks!

miklos-da:
> Perhaps instead of checking in _buf_image.bin_, this could be done inside the _compatibility_ checks, and we could generate the images from various versions on the fly?

Great idea — I'm happy to look into that as part of a follow-up story.

fmt.sh
```diff
@@ -11,6 +11,12 @@ cd "$(dirname "$0")"
 # load the dev-env
 eval "$(dev-env/bin/dade-assist)"
 
+# Location of reference image used for buf breaking check.
+# Note that in case breaking changes have been introduced, running ./fmt.sh will simply overwrite
+# this image. Hence, before committing the new image you should review the errors reported by buf
+# by restoring the reference image and running './fmt.sh --test'.
```
SamirTalwar-DA, line 17: I think this comment is out of date.

SamirTalwar-DA (approved):

fabiotudone-da (approved):

fmt.sh
miklos-da: Updated it.

PR #10412, opened 2021-07-26
[JSON-API] compile time check that we added new config fields to the logging

realvictorprm:
This ensures that whenever we add new config fields, they will never be forgotten to appear in the initial logging again. Also implicitly adds all the fields I forgot to add.

In the end I just want to avoid to have to think about even more places as soon as I introduce a new config field again.

PS: I'm pretty sure that the code could be even shorter, however my missing knowledge about shapeless brought me only this far in this short amount of time.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json/src/main/scala/com/digitalasset/http/ConfigOps.scala
+++ b/ledger-service/http-json/src/main/scala/com/digitalasset/http/ConfigOps.scala
@@ -1,0 +1,109 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.http
+
+import com.daml.cliopts.Logging.LogEncoder
+import com.daml.ledger.api.tls.TlsConfiguration
+import shapeless.{FieldPoly, LabelledGeneric}
+
+import java.nio.file.Path
+import scala.concurrent.duration.FiniteDuration
+import ch.qos.logback.classic.{Level => LogLevel}
+import com.daml.metrics.MetricsReporter
+import shapeless.labelled.FieldType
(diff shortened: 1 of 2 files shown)


ledger-service/http-json/src/main/scala/com/digitalasset/http/ConfigOps.scala
```diff
@@ -0,0 +1,109 @@
...
+// have a fitting implementation for displaying it on the command line.
+object ConfigOps {
+
+  // The signatures would make this code very unreadable
+  @SuppressWarnings(Array("org.wartremover.warts.ExplicitImplicitTypes"))
+  private object matcher extends FieldPoly {
+    type Field[Name, Typ] = Typ with FieldType[Symbol with shapeless.tag.Tagged[Name], Typ]
+    val W = shapeless.Witness
+
+    implicit def ledgerHost = at[Field[W.`"ledgerHost"`.T, String]](value => s"ledgerHost=$value")
```
realvictorprm, line 29: This needs to be that complicated because Scala 2.12 doesn't support the simple `"NiceFieldnameAsGeneric"` syntax.

cocreature (commented):
If I recall correctly the main reason to not use `toString` was that we have special `Show` instances. While there are reasons to prefer `Show` over `toString`, I don’t really see a big downside to overwriting `toString` for types that have a `Show` instance. Am I missing something?

It does seem like switching to `toString` would solve the issue in a much simpler way.

realvictorprm:
We probably should close this if we don't want it.

cocreature:
@realvictorprm @stefanobaghino-da do you want to close this?

stefanobaghino-da:
I wasn't even aware of its existence until now.

PR #10415, opened 2021-07-27
participant-integration-api: Fix completion debug log output.

SamirTalwar-DA:
So it doesn't embed Scala class names.

**Before:**

```
DEBUG c.d.p.a.s.ApiCommandCompletionService - Responding with completions: List(LoggingEntries(Map(commandId -> OfString(SemanticDoubleSpendBasic-alpha-2992c8731c2f-command-1), statusCode -> OfString(0)))), context: {parties: ["SemanticDoubleSpendBasic-alpha-2992c8731c2f-party-1"], offset: "000000000000000a0000000000000000"}
```

**After:**

```
DEBUG c.d.p.a.s.ApiCommandCompletionService - Responding with completions: [{commandId: SemanticDoubleSpendBasic-alpha-29a7c9d39f7b-command-1, statusCode: 0}], context: {parties: ["SemanticDoubleSpendBasic-alpha-29a7c9d39f7b-party-1"], offset: "000000000000000a0000000000000000"}
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
@@ -16,7 +16,6 @@ import com.daml.ledger.api.v1.command_completion_service._
 import com.daml.ledger.api.validation.PartyNameChecker
 import com.daml.ledger.participant.state.index.v2.IndexCompletionsService
 import com.daml.logging.LoggingContext.withEnrichedLoggingContext
-import com.daml.logging.entries.LoggingEntries
 import com.daml.logging.{ContextualizedLogger, LoggingContext}
 import com.daml.metrics.Metrics
 import com.daml.platform.api.grpc.GrpcApiService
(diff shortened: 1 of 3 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
```diff
@@ -52,27 +51,23 @@ private[apiserver] final class ApiCommandCompletionService private (
...
 
+  private def completionsLogLine(response: CompletionStreamResponse): String = {
+    val completionsString = response.completions.view
+      .map(c =>
+        s"{commandId: ${c.commandId}, statusCode: ${c.status.fold("none")(_.code.toString)}}"
+      )
+      .mkString("[", ", ", "]")
+    s"Responding with completions: $completionsString"
+  }
+
```
nicu-da, line 70: I've seen `Show` from scalaz being used in a few places, any reason not to use it here? Especially as I've see a few `.mkString("[", ", ", "]")` for collections

SamirTalwar-DA: Scalaz is used in a few places but I don't think we use it much in this module. I think in this case, we'd end up defining an orphan instance for `Show[CompletionStreamResponse]` which would be very specific to this particular log line; I'm not sure it's general enough to warrant a typeclass instance like that.

nicu-da: Fair enough.
I would normally still add an instance, even if it would be an orphan, can even be scoped locally, just so I don't have mentally process the call to print

nicu-da (approved):

miklos-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
SamirTalwar-DA: Well, if it's an orphan instance that's used once… that's the same as a function. Would it help if I pulled out a function here?

I think you make a good point about the `.mkString` call, too. Perhaps this stuff would be better in the logging context. Then we can use `ToLoggingContext`.

fabiotudone-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandCompletionService.scala
miklos-da: :+1: for putting these into the logging context.

SamirTalwar-DA: Done. I avoided implementing `ToLoggingValue[CompletionStreamResponse]` because it seemed a bit too niche, as I said above, but I think this code is a lot cleaner.

PR #10418, opened 2021-07-27
LF: Add context in missing package errors

remyhaemmerle-da:
explaining the origining of the missing package.

This is builton top of #10314.
This is part of #9974.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1398,26 +1398,14 @@ private[lf] object SBuiltin {
 
   /** $any-exception-message :: AnyException -> Text */
   final case object SBAnyExceptionMessage extends SBuiltin(1) {
-    private def exceptionMessage(tyCon: TypeConName, value: SValue, machine: Machine) =
-      machine.ctrl = SEApp(SEVal(ExceptionMessageDefRef(tyCon)), Array(SEValue(value)))
-
     override private[speedy] def execute(args: util.ArrayList[SValue], machine: Machine): Unit = {
       val exception = getSAnyException(args, 0)
-      val tyCon = exception.id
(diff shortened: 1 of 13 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1410,6 +1410,7 @@ private[lf] object SBuiltin {
         throw SpeedyHungry(
           SResultNeedPackage(
             tyCon.packageId,
+            language.Reference.Exception(tyCon),
```
cocreature, line 1413: Can we ever hit this? I thought loading the packages during preprocessing should be enough to make sure we never run into this so this is an internal error.

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: In off-ledger mode, there is no preprocessing and the validation can be skipped. 
So we can change that to internal error, assuming off-ledger engines have to preload everything.

cocreature: off-ledger engines should already preload everything I believe so that seems like a reasonable requirement. Doesn’t have to be in this PR though

remyhaemmerle-da: In this case it is probably fine to skip completely the check and wait the lookup of the definition fails. 
Since the ref of the definition is a `ExceptionMessageDefRef` the error message will get from `com.daml.lf.speedy.Speedy.Machine#lookupVal` should be clear enough.

remyhaemmerle-da:
cc @daravep

PR #10439, opened 2021-07-28
Move logging value definitions alongside their objects.

SamirTalwar-DA:
Rather than disassembling them elsewhere, let's do it in one place.

### Pull Request Checklist

trigger the build.


ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
```diff
@@ -259,4 +368,92 @@ object Update {
...
+      "recordTime" -> timestamp.toInstant
+
+    def submissionId(id: Ref.SubmissionId): LoggingEntry =
+      "submissionId" -> id
+
+    def submissionIdOpt(id: Option[Ref.SubmissionId]): LoggingEntry =
+      "submissionId" -> id
+
+    def participantId(id: Ref.ParticipantId): LoggingEntry =
+      "participantId" -> id
```
kamil-da, line 406: I understand that we would like to keep some of `T => LoggingEntry` methods private to specific classes/packages because of required domain context that is known only in such class/package. However, we could at least consider moving methods for `Ref` types to a common object. Currently we have multiple definitions for several of `Ref` types, e.g. for `Ref.SubmissionId`.
What do you think?

kamil-da (approved):
Looks good!
I've left a single comment for a possible follow-up PR.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
cocreature: I’d be fine with moving them to `Ref` but doesn’t have to be in this PR.

cocreature (approved):
Thanks!

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/domain.scala
```diff
@@ -288,9 +289,26 @@ object domain {
...
+        "applicationId" -> commands.applicationId,
+        "commandId" -> commands.commandId,
+        "actAs" -> commands.actAs,
+        "readAs" -> commands.readAs,
+        "submittedAt" -> commands.submittedAt,
+        "deduplicationDuration" -> commands.deduplicationDuration,
+      )
+  }
+
+  /** @param party      The stable unique identifier of a Daml party.
```
fabiotudone-da, line 309:
```suggestion
  /** @param party       The stable unique identifier of a Daml party.
```

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/CompletionInfo.scala
```diff
@@ -4,26 +4,27 @@
...
-  *                               it MUST adhere to the deduplication guarantee under a sensible interpretation
-  *                               of the corresponding [[SubmitterInfo.deduplicationPeriod]].
-  * @param submissionId An identifier for the submission that allows an application
-  *   to correlate completions to its submissions.
+  *                                Optional as some implementations may not be able to provide this deduplication information.
+  *                                If an implementation does not provide this deduplication information,
+  *                                it MUST adhere to the deduplication guarantee under a sensible interpretation
+  *                                of the corresponding [[CompletionInfo.deduplicationPeriod]].
+  * @param submissionId            An identifier for the submission that allows an application
+  *                               to correlate completions to its submissions.
```
fabiotudone-da, line 27: I am not sure I understand these ScalaDoc indentation changes

fabiotudone-da (approved):

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
SamirTalwar-DA: I'd be somewhat OK with this.

TBH I'm not sure these methods add too much but clearly others find value in them, so I guess there's something to it. I was tempted to inline them all.

That said, I'd prefer to find a way to derive these automatically from the type. Right now `SubmissionId` and some others are just type aliases to `LedgerString` though, so that wouldn't work.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/CompletionInfo.scala
SamirTalwar-DA: IntelliJ has made it happen. I actually hadn't realized it did it, it must have been on commit. Shame on me for not reading through my PR before opening it.

I will make sure they're all aligned.

SamirTalwar-DA: Re-aligned all logs in _domain.scala_, _CompletionInfo.scala_, and _Update.scala_.

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/domain.scala
```diff
@@ -338,4 +358,9 @@ object domain {
         reason: String,
     ) extends PackageEntry
   }
+
+  object Logging {
+    implicit def `tagged value to LoggingValue`[T: ToLoggingValue, Tag]: ToLoggingValue[T @@ Tag] =
+      value => value.unwrap
+  }
```
victormueller-da, line 365: I do not completely understand the code here yet but I have the feeling this could be quite useful for the json api too :thinking:

SamirTalwar-DA: It says that if we have a type `T`, a type `Tag`, and an instance of `ToLoggingValue[T]`, then we also have an instance of `ToLoggingValue[T @@ Tag]` (which just discards the tag).

For example, if we have a value of type `Ref.TransactionId @@ TransactionIdTag`, we can convert it to a logging value, because we know how to convert `Ref.TransactionId` to a logging value.

Happy to discuss further offline if you'd like to understand more.

PR #10440, opened 2021-07-28
PoC: Support Typeable constraints

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
+++ b/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
@@ -182,6 +182,10 @@ applySubstInExpr subst@Subst{..} = \case
         (applySubstInExpr subst e)
     ETypeRep t -> ETypeRep
         (applySubstInType subst t)
+    ETypeRepGeneric k t -> ETypeRepGeneric k
+        (applySubstInType subst t)
+    ETypeRepGenericApp k1 k2 ->
+        ETypeRepGenericApp k1 k2
     EToAnyException t e -> EToAnyException
         (applySubstInType subst t)
         (applySubstInExpr subst e)
--- a/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
+++ b/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
@@ -170,6 +170,7 @@ kindOf = \case
     introTypeVar v k $ checkType t1 KStar $> KStar
   TStruct recordType -> checkRecordType recordType $> KStar
   TNat _ -> pure KNat
+  TTypeRepGeneric k -> pure (KArrow k KStar)
 
 expandTypeSynonyms :: MonadGamma m => Type -> m Type
 expandTypeSynonyms = expand where
@@ -189,6 +190,7 @@ expandTypeSynonyms = expand where
       recordType' <- mapM (\(n,t) -> do t' <- expand t; return (n,t')) recordType
       return $ TStruct recordType'
     TNat typeLevelNat -> return $ TNat typeLevelNat
+    TTypeRepGeneric k -> return $ TTypeRepGeneric k
 
 expandSynApp :: MonadGamma m => Qualified TypeSynName -> [Type] -> m Type
 expandSynApp tsyn args = do
@@ -703,6 +705,12 @@ typeOf' = \case
   ETypeRep ty -> do
     checkGroundType ty
     pure $ TBuiltin BTTypeRep
+  ETypeRepGeneric k ty -> do
+    checkType ty k
+    checkGroundType' ty
+    pure $ TApp (TTypeRepGeneric k) ty
+  ETypeRepGenericApp k1 k2 -> do
+    pure $ TForall (alpha, KArrow k1 k2) $ TForall (beta, k1) (TApp (TTypeRepGeneric (KArrow k1 k2)) tAlpha :-> TApp (TTypeRepGeneric k1) tBeta :-> TApp (TTypeRepGeneric k2) (TApp tAlpha tBeta))
   EToAnyException ty val -> do
     checkExceptionType ty
     checkExpr val ty
--- a/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
+++ b/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
@@ -123,6 +123,20 @@ import qualified "ghc-lib-parser" BooleanFormula as BF
 import           Safe.Exact (zipExact, zipExactMay)
 import           SdkVersion
 
+-- NOTE [Typeable instances]
+-- We support Typeable instances in Daml, however
+-- not quite in the way GHC does.
+-- We have a polykinded TypeRep in Daml but
+-- no polykinds in general in Daml-LF.
+-- That means we cannot translate the polykinded
+-- Typeable definition. Instead, we inline it
+-- as needed and rely on the fact that all sensible uses
+-- are at a specific kind that we can fix without
+-- having to introduce kind variables.
+-- Since it is inlined everywhere, we can drop
+-- the actual Typeable definition so we never
+-- need to define a polykinded record or type synonym for it.
+
 ---------------------------------------------------------------------
 -- FAILURE REPORTING
 
@@ -485,6 +499,7 @@ convertTypeDef env o@(ATyCon t) = withRange (convNameLoc t) $ if
     | NameIn DA_Internal_Desugar n <- t
     , n `elementOfUniqSet` consumingTypes
     -> pure []
+    | NameIn Data_Typeable_Internal "TypeRep" <- t -> pure []
 
     -- Constraint tuples are represented by LF structs.
     | isConstraintTupleTyCon t
@@ -568,6 +583,8 @@ convertTypeSynonym env tycon
 
 convertClassDef :: Env -> TyCon -> ConvertM [Definition]
 convertClassDef env tycon
+--    See Note [Typeable instances]
+    | NameIn Data_Typeable_Internal "Typeable" <- tycon = pure []
     | Just cls <- tyConClass_maybe tycon
     = do
     let con = tyConSingleDataCon tycon
@@ -924,6 +941,7 @@ internalFunctions = listToUFM $ map (bimap mkModuleNameFS mkUniqSet)
     , ("GHC.Base",
         [ "getTag"
         ])
+    , ("Data.Typeable.Internal", ["mkTrCon", "mkTrApp", "mkTrFun", "$WTypeRep"])
     ]
 
 convertExpr :: Env -> GHC.Expr Var -> ConvertM LF.Expr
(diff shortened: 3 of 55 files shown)


cocreature:
@sofiafaro-da in case you’re interested, I hacked up a very messy PoC that uses GHC Typeable constraints to generate LF `TypeRep`s which seems to work surprisingly well.

I think to use this safely for generic templates we need to add a type parameter to `TypeRep` but apart from that this could be a relatively simple solution which would require little to know changes to the interpreter and have relatively limited performance implications (compared to dropping type erasure completely).

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -568,6 +568,20 @@ convertTypeSynonym env tycon
 
 convertClassDef :: Env -> TyCon -> ConvertM [Definition]
 convertClassDef env tycon
+    | getOccText tycon == "Typeable" = do
+      let tconName = mkTypeCon [getOccText tycon]
+          tsynName = mkTypeSyn [getOccText tycon]
+          tyVars = [(TypeVarName "fakeKind", KStar), (TypeVarName "a", KStar)]
```
sofiafaro-da, line 574: Without polykinds in LF, I'm a bit concerned we wont be able to construct parametrised typereps for e.g. `Numeric 10` because 10 has kind Nat. Or for `Optional Int` because `Optional` has kind `* -> *`.

sofiafaro-da: I see four outs: 

1. Stick to non-parametrised typereps, like implemented in this PoC, and do some runtime type checking on creates and exercises (i.e. given a typerep and a template value, check the value fits the typerep).

2. Introduce a non-type-erased forall in LF.

3. Remove type erasure completely.

4. Only allow template arguments that fit the very specific set of type reps we can fully support, like `T X1 ... Xn` where `X1` ... `Xn` are type constructor of kind star that themselves have no type arguments. 

I think 4 is a really bad restriction in general since it's non-compositional. E.g. you can't pass a generic template to a generic template.

sofiafaro-da: But maybe ... maybe we can translate the Typeable instance directly into the parametrised typerep? The parametrised typerep would be the only type that has a polykind-style type signature.

cocreature: I don’t think parametrized types are a problem in general. E.g., the tests already cover `[Int]` which is equivalent to `Optional Int` (and I confirmed the latter works)
What we cannot get is higher-kinded types as parameters but types with higher-kinded parameters are not serializable either so that seems like a consistent restriction with what we have already.
`Numeric 10` does indeed currently fail with the following error:
```
        error:
        Failed to load interface for ‘GHC.TypeNats’
        There are files missing in the ‘daml-prim-0.0.0’ package,
        try running 'ghc-pkg check'.
        Use -v (or `:set -v` in ghci) to see a list of the files searched for.
```
However, that’s not an LF error but a GHC error and I’m reasonably optimistic we can fix that and from there on it should work.

sofiafaro-da: The problem I'm thinking is how do you represent this in LF with parameters? I'm sure we can make this part work with the existing typereps in LF (see option 1 above), I'm just worried this isn't going to generalise to parametrised typereps.

cocreature: I don’t see the issue. We need a `TypeRep (Optional Int)`. We don’t need a `TypeRep Optional`.

sofiafaro-da: Maybe I just need to see it to believe it. My worry is how these typereps are getting constructed in a type-safe way. But maybe it's possible?

By the way, if we do support parametrised typereps at kind `*`, we can support any kind of template parameters since create/exercise can just ask for the typerep of the template.

cocreature: Let’s look at how I think this can work:
**LF**
In LF we currently have `ETypeRep_{type} : TypeRep` where `type` is a ground type. My proposal is to change (or rather add because backwards compat) this to `ETypeRep_{type} : TypeRep type`. We don’t add other primitives to construct primitives. Unless I’m missing something that is type safe.

**GHC & LF conversion**
I think we first have to clarify how GHC generates `Typeable` instances. There are no fixed instances for types and afaict instances don’t even propagate through modules. Instead they are made up adhoc. To make this concrete, let’s look at `Optional Int`:
There is no `Typeable a => Typeable (Optional a)` instance that GHC uses here. Instead, it generates an instance for `Optional Int` instance on the fly. If you use `Optional Bool`, it generates a typeable instance for that.
This PR exploits the fact that GHC is very eager to generate instances for ground types and generates an LF typerep based on that.
This doesn’t work completely, GHC does allow for code like the following:
```
f :: (Typeable f, Typeable x) => f x -> TypeRep (f x)
f _ = typeRep
```
This wouldn’t work with my PR and it wouldn’t work unless we add more ways to build typereps.
Basically the limitation is that you have to write your code such that the instance can get resolved directly to a ground type instead of being built up incrementally.
While that might sound bad, I don’t think this is a big limitation:
1. Even in Haskell, I’m not sure I’ve ever seen code like the one above. The wast majority of Typeable instances get resolved directly to a ground type.
2. The obvious case to get something like this would be if we had higher-kinded type parameters to templates which we don’t even support outside of templates atm for serializable types. I don’t think this is a route we should go down atm.
3. We can detect this reasonably easily in the LF conversion, I don’t expect most people to hit it and when you hit it you can probably adjust your code to work around it.

So yes, there is a real limitation with this approach wtr to how Typeable instances can be built up.
Looking through your options:
1. If we add new ways to build up typereps, I think we could make this work somehow but I don’t think it’s worth the complexity. But I really don’t like the idea of having to do runtime typechecking here.
2. Seems like a fair amount of complexity.
3. I don’t see how we can stop erasing types without compromising on performance.
4. Even more complexity than 2.
5. Seems more restrictive than the restriction on instance resolution.
So overall, I think the restriction on instance resolution for `Typeable` constraints is very reasonable compared to the alternatives.

sofiafaro-da: 👍 Thanks for explaining. I think I understand now better where you're coming from.

I still think this might be too limiting. E.g. it's not possible to generate the typereps for the create instance below:

```
template Template1 x 
    with ...
    where ...
       create @(Template2 (Optional x)) ...
```

Given that GHC generates typeable instances eagerly whenever it needs, what do you think about the sixth option -- making the new TypeRep primitive be polykind? Instead of translating `Typeable k t` constraint to an LF struct, we translate it directly to `TypeRep k t` where `TypeRep` is the new primitive. I think that would make it possible to construct typereps incrementally, and handle the case above as well as cases involving nats and * -> * and such. In actual LF code, the typerep's kind is always statically known.

cocreature: Making TypeRep polykinded sounds possible, I’m not quite convinced the complexity is justified (i.e., we’re back to I’d like to see usecases before we implement something) but I’ll play around with it a bit.

cocreature: I’ve pushed a very hacky poc that uses a polykinded `TypeRep`. It’s not too bad in terms of complexity and while it needs lots of cleanup it does seem to work.

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
```diff
@@ -703,6 +705,12 @@ typeOf' = \case
   ETypeRep ty -> do
     checkGroundType ty
     pure $ TBuiltin BTTypeRep
+  ETypeRepGeneric k ty -> do
+    checkType ty k
+    checkGroundType' ty
+    pure $ TApp (TTypeRepGeneric k) ty
+  ETypeRepGenericApp k1 k2 -> do
+    pure $ TForall (alpha, KArrow k1 k2) $ TForall (beta, k2) (TApp (TTypeRepGeneric (KArrow k1 k2)) tAlpha :-> TApp (TTypeRepGeneric k1) tBeta :-> TApp (TTypeRepGeneric k2) (TApp tAlpha tBeta))
```
sofiafaro-da, line 713:
```suggestion
    pure $ TForall (alpha, KArrow k1 k2) $ TForall (beta, k1) (TApp (TTypeRepGeneric (KArrow k1 k2)) tAlpha :-> TApp (TTypeRepGeneric k1) tBeta :-> TApp (TTypeRepGeneric k2) (TApp tAlpha tBeta))
```

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -961,6 +974,14 @@ convertExpr env0 e = do
...
     go env (VarIn GHC_Types "primitive") (LType (isStrLitTy -> Just y) : LType t : args)
         = fmap (, args) $ convertPrim (envLfVersion env) (unpackFS y) <$> convertType env t
+    go env (VarIn Data_Typeable_Internal "mkTrCon") (LType kind : LType typ : _tycon : _kindargs : args)
+        = do kind <- convertKind kind
+             typ <- convertType env typ
+             pure  (ETypeRepGeneric kind typ, args)
+    go env (VarIn Data_Typeable_Internal "mkTrApp") (LType k1 : LType k2 : args)
+        = do k1 <- convertKind k2
+             k2 <- convertKind k2
+             pure (ETypeRepGenericApp k1 k2, args)
```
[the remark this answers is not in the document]
cocreature: oh absolutely, I just tried to get the bare minimum working in a poc. This is nowhere near finished

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
sofiafaro-da: I'm glad it works :) I went through and reviewed most of it, and it looks good. The things I would consider changing is defining `TTypeRep<k>` as a builtin type if possible, since it has the right sort of equality. If I had implemented this, I would probably have tried to convert `(Typeable k t)` directly as a `(TypeRep<k> t)` instead of wrapping it in an LF struct `{ m_typeRep# : Unit -> TypeRep<k> t }`. But separating the two does make it possible to distinguish them later in data-dependencies, so I'm not sure my way would be better.

cocreature: Making families builtin in proto is a bit annoying (builtin types are an enum which can’t have parameters) and we deliberately avoided that at the value-level, e.g., `FromAny` is not a builtin. But maybe we could still make it a builtin type in the Haskell & Scala AST though. wdyt @remyhaemmerle-da?

cocreature: As for converting `Typeable` directly, I actually did something along those lines initially but now I can treat the body of the typeclass instance as a blackbox and only replace the primitives generated by GHC which seems a bit simpler. No strong feelings either way there though

sofiafaro-da: The reason fromAny isn't a builtin is because its type parameter has special rules (must be a ground type). We have builtin expressions that are not strictly enums (e.g. BEInt, BENumeric, etc). Looking at how builtin expressions are represented in the Haskell/Scala AST vs protobuf, it seems pretty ad hoc, unlike the types. (Btw, looking at this, it seems that `ETypeRepGenericApp` could be treated as a builtin expression, but `ETypeRepGeneric` cannot.)

I guess it's better to keep the types tidy as they are, and treat TTypeRep as a non-builtin.

sofiafaro-da: > As for converting `Typeable` directly, I actually did something along those lines initially but now I can treat the body of the typeclass instance as a blackbox and only replace the primitives generated by GHC which seems a bit simpler. No strong feelings either way there though

It's going to get a bit messier & inefficient when converting to/from the typeable instance to the typerep in template instance functions and choice bodies. I doubt it will be a big issue though.

cocreature: I had a chat with @remyhaemmerle-da on Thursday on the builtin stuff and he seemed to prefer both the types and the constructors to not be builtins/primitives so the current state in this PR.

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Subst.hs
```diff
@@ -182,6 +182,10 @@ applySubstInExpr subst@Subst{..} = \case
         (applySubstInExpr subst e)
     ETypeRep t -> ETypeRep
         (applySubstInType subst t)
+    ETypeRepGeneric k t -> ETypeRepGeneric k
+        (applySubstInType subst t)
```
sofiafaro-da, line 186: Hmm. In theory there should be no type variables in t...

sofiafaro-da: We do this sort of substitution in other places though that shouldn't have type variables, though. I just realised how technically unnecessary it is?

cocreature: yeah noticed that while implementing and decided to stick with what we do elsewhere rather than tryting to be clever.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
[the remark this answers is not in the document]
cocreature: implemented now.

compiler/damlc/daml-prim-src/Data/Typeable/Internal.daml
```diff
@@ -0,0 +1,40 @@
...
+mkTrApp :  forall k1 k2 (a : k1 -> k2) (b : k1).
+           TypeRep (a : k1 -> k2)
+        -> TypeRep (b : k1)
+        -> TypeRep (a b)
+mkTrApp = magic @"mkTrApp"
+
+mkTrFun : forall (r1 : RuntimeRep) (r2 : RuntimeRep)
+                  a b.
+           TypeRep a -> TypeRep b -> TypeRep ((a -> b))
+mkTrFun arg res = magic @"mkTrFun"
```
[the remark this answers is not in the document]
cocreature: it is now

cocreature:
closing for now since it looks like we probably don’t need this for daml interfaces at least.

PR #10444, opened 2021-07-29
[JSON-API] Add option for setting a table prefix

### LINKED ISSUE #10390: JSON API schema: table prefixes (or suffixes)
The JSON API is designed for running behind a load balancer (with a sticky bit on user credentials). It is also designed to have a per-instance db-backed cache.

This creates a situation where an operator might want to start N instances of the JSON API behind a load balancer. Currently, this would require said operator to supply a different JDBC URL to each JSON API instance to make them point to, at the very least, separate logical DBs (PostgreSQL schemas). This makes scaling unnecessarily difficult.

In order to work around that, I think we should have a CLI argument to the JSON API that sets a prefix (or suffix, or infix) on all table names. This would effectively namespace each JSON API instance, and make it possible to run multiple JSON API instances against the same logical DB, using the same set of credentials.

If the CLI flag is not set, its value should be generated randomly at startup (a slowdown at startup seems a better default than accidentally ending up with multiple JSON API instances writing on the same tables).

realvictorprm:
Should fix #10390

changelog_begin

- [JSON-API] A table prefix can now be specified in the jdbc config via `tablePrefix=<YourFancyTablePrefix>`. This was added to allow running multiple instances of the json api without having collisions (independent of the chosen database).

changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -40,6 +40,8 @@ sealed abstract class Queries(tablePrefix: String) {
   /** for use when generating predicates */
   protected[this] val contractColumnName: Fragment = sql"payload"
 
+  protected[this] val tablePrefixFr = Fragment.const0(tablePrefix)
+
   /** Table names with prefix * */
 
   val contractTableNameRaw = s"${tablePrefix}contract"
@@ -61,7 +63,7 @@ sealed abstract class Queries(tablePrefix: String) {
     sql"""
       CREATE TABLE
         $contractTableName
-        (contract_id $contractIdType NOT NULL CONSTRAINT contract_k PRIMARY KEY
+        (contract_id $contractIdType NOT NULL CONSTRAINT ${tablePrefixFr}contract_k PRIMARY KEY
         ,tpid $bigIntType NOT NULL REFERENCES $templateIdTableName (tpid)
         ,${jsonColumn(sql"key")}
         ,${jsonColumn(contractColumnName)}
@@ -106,7 +108,7 @@ sealed abstract class Queries(tablePrefix: String) {
     sql"""
       CREATE TABLE
         $templateIdTableName
-        (tpid $bigSerialType NOT NULL CONSTRAINT template_id_k PRIMARY KEY
+        (tpid $bigSerialType NOT NULL CONSTRAINT ${tablePrefixFr}template_id_k PRIMARY KEY
         ,package_id $packageIdType NOT NULL
         ,template_module_name $nameType NOT NULL
         ,template_entity_name $nameType NOT NULL
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -27,7 +27,7 @@ import cats.syntax.applicative._
 import cats.syntax.functor._
 import doobie.free.connection
 
-sealed abstract class Queries {
+sealed abstract class Queries(val tablePrefix: String) {
   import Queries.{Implicits => _, _}, InitDdl._
   import Queries.Implicits._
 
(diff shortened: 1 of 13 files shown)


realvictorprm:
Side note, from what I saw the ledger service does not support custom table prefixes yet :thinking:

realvictorprm:
merci @cocreature

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -709,33 +734,35 @@ private object OracleQueries extends Queries {
...
       doesTableExist <-
         sql"""SELECT EXISTS(
                 SELECT 1 FROM ALL_TABLES
-                WHERE table_name = 'json_api_schema_version'
+                WHERE table_name = '$jsonApiSchemaVersionTableName'
               )""".query[Boolean].unique
       version <-
         if (!doesTableExist) connection.pure(None)
-        else sql"SELECT version FROM json_api_schema_version".query[Int].option
+        else sql"SELECT version FROM $$jsonApiSchemaVersionTableName".query[Int].option
```
realvictorprm, line 765: ok this is wrong :facepalm:

realvictorprm: fixed

realvictorprm:
Finally the oracle build passes :smile_cat:

realvictorprm:
OK now everything passes :)

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -27,7 +27,7 @@ import cats.syntax.applicative._
 import cats.syntax.functor._
 import doobie.free.connection
 
-sealed abstract class Queries {
+sealed abstract class Queries(val tablePrefix: String) {
```
S11001001, line 30: Pretty sure you don't use this.

```suggestion
sealed abstract class Queries(tablePrefix: String) {
```

S11001001 (approved):

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: true, thanks for the heads up! :smiley:

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -40,13 +40,29 @@ sealed abstract class Queries {
   /** for use when generating predicates */
   protected[this] val contractColumnName: Fragment = sql"payload"
 
+  /** Table names with prefix * */
+
+  val contractTableNameRaw = s"${tablePrefix}contract"
```
cocreature, line 45: I think it would be great to test that this is set everywhere it needs to be.
I think there are two options for this:
1. Some form of unit test over `initDatabaseDdls` that checks that everything is correctly prefixed.
2. An integration test that spins up two instances of the JSON API with different prefixes in the same database and checks that both work.
I’d probably go for 2 but your choice.

cocreature (approved):
Thanks!

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -40,13 +40,29 @@ sealed abstract class Queries {
...
+  val templateIdTableNameRaw = s"${tablePrefix}template_id"
+  val templateIdTableName: Fragment = Fragment.const0(templateIdTableNameRaw)
+
+  val jsonApiSchemaVersionTableNameRaw = s"${tablePrefix}json_api_schema_version"
+  val jsonApiSchemaVersionTableName: Fragment =
+    Fragment.const0(jsonApiSchemaVersionTableNameRaw)
+
   private[this] val createContractsTable = CreateTable(
-    "contract",
+    contractTableNameRaw,
```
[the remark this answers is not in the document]
realvictorprm: I'm in favour of this, if I would knew how to extract the raw string from the fragment. So far I did not figure that out yet.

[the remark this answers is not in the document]
cocreature: why do you need to do this? you can interpolate a fragment in a fragment iirc

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: Just wrote a test locally which does `2` and caught a bug, thanks for the heads up :smile:

realvictorprm: @cocreature would you be fine with me pushing the last fixes I made to the code and adding the test in a seperate pr? It turned out to be a bit more code :smile:

cocreature: sounds reasonable

realvictorprm:
OK, going to merge this now. Further tests will be added in a follow up PR.

PR #10453, opened 2021-08-02
LF: restrict length of Name and DottedName

### LINKED ISSUE #10358: Limit length of template ids
Currently, Daml-LF imposes no limit on template ids, i.e., the 3 tuple (package id, modulename, entity name).

In practice, package ids are and have always been sha256sums and therefore limited to 64 characters so we could add that restriction in daml-lf without much trouble.

Module names and entity names are more tricky. When running on PostgreSQL, I believe both the JSON API and the participant ill actually support fairly arbitrary lengths.

However, on Oracle the situation is a bit different: Oracle doesn’t like arbitrary length strings so we add a bunch of random limits.

For the participant that limit is 4000 characters for the whole template id in the format `pkgid:module:entity` which is just the max string lenght on Oracle by default. For the JSON API, we have a limit of 1562 characters on the module name and on the entity name (separately) and a 64 char limit on the package id. The reason for that limit is that we have a `unique (pkgid, module, entity)` constraint and Oracle has a max size restriction on the things contributing to indices and constraints.

My suggestion would be to limit module name and entity name (including choice names and other identifiers) to length 1000 each (as a somewhat understandable length as opposed to 1562 and to have a bit of leeway) and enforce this in daml-lf.

The big question here is backwards compatibility. There are two options I see:

1. Restrict this for older lf versions as well. This will break a ledger that already has such templates. I think that’s highly unlikely but it is possible. There would also be no remedy in this case. You cannot prune packages so there is no solution like upgrading on the old ledger to something with shorter names.
2. Restrict this only for newer lf versions. In that case, a ledger running on Oracle should probably forbid old lf versions.

I have a slight preference for 1 since I think the risk is acceptable but ymmv.

cc @gerolf-da @vivd-da @bame-da @remyhaemmerle-da @sormeter

remyhaemmerle-da:
fixes #10358

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
@@ -126,6 +126,8 @@ object Ref {
   }
 
   object DottedName {
+    val maxLength = 1000
+
     def fromString(s: String): Either[String, DottedName] =
       if (s.isEmpty)
         Left(s"Expected a non-empty string")
@@ -154,8 +156,15 @@ object Ref {
     def assertFromSegments(s: Iterable[String]): DottedName =
       assertRight(fromSegments(s))
 
-    def fromNames(names: ImmArray[Name]): Either[String, DottedName] =
-      Either.cond(names.nonEmpty, new DottedName(names), "No segments provided")
+    def fromNames(names: ImmArray[Name]): Either[String, DottedName] = {
+      val length = names.foldLeft(-1)(_ + _.length + 1)
+      if (length < 0)
+        Left("No segments provided")
+      else if (length > maxLength)
+        Left("")
+      else
+        Right(new DottedName(names))
+    }
 
     @throws[IllegalArgumentException]
     def assertFromNames(names: ImmArray[Name]): DottedName =
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
@@ -156,15 +156,16 @@ object Ref {
     def assertFromSegments(s: Iterable[String]): DottedName =
       assertRight(fromSegments(s))
 
-    def fromNames(names: ImmArray[Name]): Either[String, DottedName] = {
-      val length = names.foldLeft(-1)(_ + _.length + 1)
-      if (length < 0)
+    def fromNames(names: ImmArray[Name]): Either[String, DottedName] =
+      if (names.isEmpty)
         Left("No segments provided")
-      else if (length > maxLength)
-        Left("")
-      else
-        Right(new DottedName(names))
-    }
+      else {
+        val length = names.foldLeft(-1)(_ + _.length + 1)
+        if (length > maxLength)
+          Left(s"""DottedName is too long (max: $maxLength)""")
+        else
+          Right(new DottedName(names))
+      }
 
     @throws[IllegalArgumentException]
     def assertFromNames(names: ImmArray[Name]): DottedName =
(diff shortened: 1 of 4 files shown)


daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
```diff
@@ -154,8 +156,15 @@ object Ref {
...
       assertRight(fromSegments(s))
 
-    def fromNames(names: ImmArray[Name]): Either[String, DottedName] =
-      Either.cond(names.nonEmpty, new DottedName(names), "No segments provided")
+    def fromNames(names: ImmArray[Name]): Either[String, DottedName] = {
+      val length = names.foldLeft(-1)(_ + _.length + 1)
+      if (length < 0)
+        Left("No segments provided")
+      else if (length > maxLength)
+        Left("")
```
cocreature, line 164: looks like you forgot to fill in the error message.

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
```diff
@@ -154,8 +156,15 @@ object Ref {
     def assertFromSegments(s: Iterable[String]): DottedName =
       assertRight(fromSegments(s))
 
-    def fromNames(names: ImmArray[Name]): Either[String, DottedName] =
-      Either.cond(names.nonEmpty, new DottedName(names), "No segments provided")
+    def fromNames(names: ImmArray[Name]): Either[String, DottedName] = {
+      val length = names.foldLeft(-1)(_ + _.length + 1)
+      if (length < 0)
+        Left("No segments provided")
```
cocreature, line 162:
```suggestion
      if (names. empty) {
        Left("No segments provided")
       } else {
      val length = names.view.map(_.length).sum + names.length - 1
```
seemed a bit easier to understand but feel free to ignore

daml-lf/spec/daml-lf-1.rst
```diff
@@ -390,16 +390,17 @@ and other similar pitfalls. ::
 We can now define a generic notion of *identifier* and *name*::
 
   identifiers:
-          Ident  ∈  [a-zA-Z_\$][a-zA-Z0-9_\$]       -- Ident
+          Ident  ∈  [a-zA-Z_\$][a-zA-Z0-9_\$]{999}  -- Ident
```
cocreature, line 393:
```suggestion
          Ident  ∈  [a-zA-Z_\$][a-zA-Z0-9_\$]{1000}  -- Ident
```

cocreature: or limit to 999 but currently your limit seems to be 1000 while you specify 999

cocreature (approved):
Thanksk

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: There is indeed an error here, but not the one you pointed out. 
The correct regexp is 
```
[a-zA-Z_\$][a-zA-Z0-9_\$]{0,999} 
```
or in full English: An `Identifier` is a non empty string starting with one character matching `[a-zA-Z_\$]`  followed by 0 to 999 character matching `[a-zA-Z0-9_\$]`.

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
remyhaemmerle-da: caa2888f26c632b9c89f5f3d10cbae9c19b31444

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Ref.scala
remyhaemmerle-da: caa2888f26c632b9c89f5f3d10cbae9c19b31444

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: `1 + 999 = 1000`

cocreature: duh ofc, I need more coffee

PR #10454, opened 2021-08-02
KV: Fix KV's TransactionCommitter's rejection trace logs [KVL-1023]

fabiotudone-da:
## Description

- A trace-level log line saying that the transaction has been rejected was being wrongly produced even when the rejection log entry was being created for pre-execution. This is fixed by logging only in functions that are being used to reject a transaction immediately.
- Renamed around a bit to clarify when we're rejecting immediately (and logging and updating metrics).
- Added a rejection trace log in the post-executor too.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
@@ -24,18 +24,18 @@ private[transaction] class Rejections(metrics: Metrics) {
 
   private final val logger = ContextualizedLogger.get(getClass)
 
-  def buildImmediateRejectionStep[A](
+  def reject[A](
       transactionEntry: DamlTransactionEntrySummary,
       rejection: Rejection,
       recordTime: Option[Timestamp],
   )(implicit loggingContext: LoggingContext): StepResult[A] =
-    buildImmediateRejectionStep(
+    reject(
       buildRejectionEntry(transactionEntry, rejection),
       rejection.description,
       recordTime,
     )
 
-  def buildImmediateRejectionStep[A](
+  def reject[A](
       rejectionEntry: DamlTransactionRejectionEntry.Builder,
       rejectionDescription: String,
       recordTime: Option[Timestamp],
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
@@ -53,7 +53,7 @@ private[transaction] class Rejections(metrics: Metrics) {
   def buildRejectionEntry(
       transactionEntry: DamlTransactionEntrySummary,
       rejection: Rejection,
-  )(implicit loggingContext: LoggingContext): DamlTransactionRejectionEntry.Builder = {
+  ): DamlTransactionRejectionEntry.Builder = {
     val builder = DamlTransactionRejectionEntry.newBuilder
     builder
       .setSubmitterInfo(transactionEntry.submitterInfo)
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
@@ -22,7 +22,7 @@ import com.daml.metrics.Metrics
 
 private[transaction] class Rejections(metrics: Metrics) {
 
-  private final val logger = ContextualizedLogger.get(getClass)
+  final private val logger = ContextualizedLogger.get(getClass)
 
   def reject[A](
       transactionEntry: DamlTransactionEntrySummary,
@@ -50,7 +50,13 @@ private[transaction] class Rejections(metrics: Metrics) {
     )
   }
 
-  def buildRejectionEntry(
+  def preExecutionOutOfTimeBoundsRejectionEntry(
+      transactionEntry: DamlTransactionEntrySummary,
+      rejection: Rejection,
+  ): DamlTransactionRejectionEntry =
+    buildRejectionEntry(transactionEntry, rejection).build
+
+  private def buildRejectionEntry(
       transactionEntry: DamlTransactionEntrySummary,
       rejection: Rejection,
   ): DamlTransactionRejectionEntry.Builder = {
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
@@ -3,6 +3,8 @@
 
 package com.daml.ledger.participant.state.kvutils.committer.transaction
 
+import java.time.Instant
+
 import com.codahale.metrics.Counter
 import com.daml.ledger.participant.state.kvutils.DamlKvutils.{
   DamlTransactionRejectionEntry,
@@ -52,9 +54,13 @@ private[transaction] class Rejections(metrics: Metrics) {
 
   def preExecutionOutOfTimeBoundsRejectionEntry(
       transactionEntry: DamlTransactionEntrySummary,
-      rejection: Rejection,
+      minimumRecordTime: Instant,
+      maximumRecordTime: Instant,
   ): DamlTransactionRejectionEntry =
-    buildRejectionEntry(transactionEntry, rejection).build
+    buildRejectionEntry(
+      transactionEntry,
+      Rejection.RecordTimeOutOfRange(minimumRecordTime, maximumRecordTime),
+    ).build
 
   private def buildRejectionEntry(
       transactionEntry: DamlTransactionEntrySummary,
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
@@ -24,22 +24,24 @@ private[transaction] class Rejections(metrics: Metrics) {
 
   private final val logger = ContextualizedLogger.get(getClass)
 
-  def buildRejectionStep[A](
+  def buildImmediateRejectionStep[A](
       transactionEntry: DamlTransactionEntrySummary,
       rejection: Rejection,
       recordTime: Option[Timestamp],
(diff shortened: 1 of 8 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -24,22 +24,24 @@ private[transaction] class Rejections(metrics: Metrics) {
...
       buildRejectionEntry(transactionEntry, rejection),
+      rejection.description,
       recordTime,
     )
-  }
 
-  def buildRejectionStep[A](
+  def buildImmediateRejectionStep[A](
       rejectionEntry: DamlTransactionRejectionEntry.Builder,
+      rejectionDescription: String,
```
nicu-da, line 40: wouldn't it make more sense for this description to be derived from the entry itself? With time there's a chance that we update only one of them and it creates very confusing logs.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -24,22 +24,24 @@ private[transaction] class Rejections(metrics: Metrics) {
...
       recordTime: Option[Timestamp],
-  ): StepResult[A] = {
+  )(implicit loggingContext: LoggingContext): StepResult[A] = {
     Metrics.rejections(rejectionEntry.getReasonCase.getNumber).inc()
+    logger.trace(s"Transaction rejected, $rejectionDescription.")
```
nicu-da, line 44: Would pre-execution still call this step and have the log produced? Or does it follow another code path?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: This variant is used only by the deduplication check that doesn't have a dedicated reason object. @SamirTalwar-DA is it better to add a reason for deduplication?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: No, pre-execution (i.e., when there is no record time, the code path is not super-obvious indeed) calls directly the function below as it pre-computes a rejection check for post-execution to potentially use (because the time check is done at post-execution).

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -52,8 +54,6 @@ private[transaction] class Rejections(metrics: Metrics) {
       transactionEntry: DamlTransactionEntrySummary,
       rejection: Rejection,
   )(implicit loggingContext: LoggingContext): DamlTransactionRejectionEntry.Builder = {
-    logger.trace(s"Transaction rejected, ${rejection.description}.")
```
SamirTalwar-DA, line 55: Do we log this somewhere else, e.g. in post-execution?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -24,22 +24,24 @@ private[transaction] class Rejections(metrics: Metrics) {
 
   private final val logger = ContextualizedLogger.get(getClass)
 
-  def buildRejectionStep[A](
+  def buildImmediateRejectionStep[A](
```
SamirTalwar-DA, line 27: I'm not sure what "immediate" means here. Could you please explain?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: In the function above.

fabiotudone-da: And also in post-execution I think, but I'll double-check.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: It means that the step has to be the result of the validation. Perhaps `buildConclusiveRejection`? Or even `reject`, so that it's clear that it has side effects (metrics and logging).

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: It looks like there are at least some post-executors that don't log which write set they've chosen => further PRs will come.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
SamirTalwar-DA: Yes, I think that if it records something, even metrics, we need to capture that in the name so it's clear it's not pure. `reject` works well IMO.

Are there any functions that build a rejection that could be thrown away later?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: I'll fix at least one in this PR though, as it seems fairly straightforward.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: Not so far but see [here](https://github.com/digital-asset/daml/pull/10454#pullrequestreview-720118302).

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
SamirTalwar-DA: Yes, but there's much larger changes to be made here to represent _all_ possible rejection reasons in a structured manner. I suggest we do it with a bit of up-front design as we need to support anything that gets written to the ledger.

fabiotudone-da: Alright, I suspected as much. I'll leave it as it is for now.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: 524e1a7115ac43a9f89b6e404c3271bd5f867242

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -51,9 +53,7 @@ private[transaction] class Rejections(metrics: Metrics) {
   def buildRejectionEntry(
```
SamirTalwar-DA, line 53: Maybe let's rename this to `buildPotentialRejectionEntry` or something to make it clear we may not use the result?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/validator/preexecution/WriteSetSelector.scala
```diff
@@ -3,9 +3,12 @@
...
 
-/** Selects a write set from a [[PreExecutionOutput]]. */
+import com.daml.logging.LoggingContext
+
 trait WriteSetSelector[ReadSet, WriteSet] {
 
-  def selectWriteSet(preExecutionOutput: PreExecutionOutput[ReadSet, WriteSet]): WriteSet
+  def selectWriteSet(preExecutionOutput: PreExecutionOutput[ReadSet, WriteSet])(implicit
+      loggingContext: LoggingContext
+  ): WriteSet
```
SamirTalwar-DA, line 12: When given the choice, I prefer to put the explicit arguments on newlines rather than the implicit ones, because those are the ones I care about most of the time.

```suggestion
  def selectWriteSet(
      preExecutionOutput: PreExecutionOutput[ReadSet, WriteSet]
  )(implicit loggingContext: LoggingContext): WriteSet
```

SamirTalwar-DA (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
fabiotudone-da: Problem is, then the fact that reject builds a "potential" rejection may become confusing. I'm tempted to make it `private` and wrap it as `buildPostExecutionRejectionEntry` to be used in the ledger time validator.

SamirTalwar-DA: That sounds like a good plan to me.

fabiotudone-da: 2f5068e

fabiotudone-da: And 94a7787; also, I think 1df697f is more correct for the post-executor.

fabiotudone-da:
@nicu-da I'd like to merge it if you don't have further concerns

fabiotudone-da:
@moritzkiefer-da @victormueller-da FYI, going to be merged soon

PR #10456, opened 2021-08-02
Build & releases 1.13 Ledger API test tool

cocreature:
Temporary hack to unblock Canton until they finish their exception support.

changelog_begin
changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10458, opened 2021-08-02
Reactive canton conformance test aginst LF 1.13

remyhaemmerle-da:
Now we compile the conformance test for 1.13, we can run conton
against 1.13.

follow up of #10456

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -4,6 +4,8 @@
 # The following dictionary alias LF versions to keywords:
 # - "legacy" is the keyword for last LF version that supports legacy
 #    contract ID scheme,
+# - "no-exceptions" is a keyword for the last version that does not
+#   support exceptions (can be dropped without notice),
 # - "default" is the keyword for the default compiler output,
 # - "latest" is the keyword for the latest stable LF version,
 # - "preview" is the keyword fort he next LF version, *not stable*,
--- a/daml-lf/language/daml-lf.bzl
+++ b/daml-lf/language/daml-lf.bzl
@@ -18,16 +18,14 @@
 
 lf_version_configuration = {
     "legacy": "1.8",
+    # TODO (MK) Drop no-exception once Canton fully supports exceptions.
+    "no-exceptions": "1.13",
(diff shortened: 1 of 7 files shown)


daml-lf/language/daml-lf.bzl
```diff
@@ -18,16 +18,14 @@
 
 lf_version_configuration = {
     "legacy": "1.8",
+    # TODO (MK) Drop no-exception once Canton fully supports exceptions.
```
cocreature, line 21: Have you checked that we don’t use this anywhere else? The reason why I kept it separate is that I wanted to limit the amount of stuff we have to compile & test against 1.13.

cocreature (approved):

daml-lf/language/daml-lf.bzl
remyhaemmerle-da: Yes. 

The only place where it was really used was for the rules:
- `//daml-lf/encoder:testing-dar-lookup-scala-%s` and 
- `//daml-lf/encoder:testing-dar-lookup-lib-%s` and 
The compilation of this two rules induced a minor overhead. 

However, I double check and realized that only `//daml-lf/encoder:testing-dar-lookup-lib-latest` is used, so  deactivate in b2c4e108393ce3b631f70dcef9afca0142ad2552 the compilation for the other keywords, meaning now compile even less.

remyhaemmerle-da:
The crux of the PR is actually in 29599b6c57e016b64fe8794c6548c3a76b459112

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -62,7 +62,7 @@ conformance_test(
         "@jdk11_nix//:bin/java",
     ],
     lf_versions = [
-        "legacy",
+        "no-exceptions",
         # FIXME: replace the previous line by the following ones once canton-test-runner supports LF 1.14
```
remyhaemmerle-da, line 66: @oliverse-da The two FIXME are for you.

PR #10461, opened 2021-08-02
participant-integration-api: Use `Scheduler`, and add more tests to configuration initialization. [KVL-1046]

SamirTalwar-DA:
This modifies the `IndexStreamingCurrentLedgerConfiguration` and `LedgerConfigProvisioner` classes to:

- switch out calls to `materializer.scheduleOnce` for `scheduler.scheduleOnce`,
- use the returned `Cancellation` to cancel scheduled tasks,
- use the services execution context instead of (ab)using the materializer,
- add a couple of clauses to handle exceptions (with `logger.error`),
- use the Akka TestKit scheduler in tests, and
- increase test coverage around shutting down.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
@@ -68,7 +68,7 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
   // stream of configuration changes.
   // If the source of configuration changes proves to be a performance bottleneck,
   // it could be replaced by regular polling.
-  private[this] def startLoading(): Future[Unit] = {
+  private[this] def startLoading(): Unit =
     index
       .lookupConfiguration()
       .map {
@@ -84,7 +84,11 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
           latestConfigurationState.set(None -> None)
       }
       .map(_ => startStreamingUpdates())
-  }
+      .failed
+      .foreach { exception =>
+        readyPromise.tryFailure(exception)
+        logger.error("Could not load the ledger configuration.", exception)
+      }
 
   private[this] def configFound(
       offset: LedgerOffset.Absolute,
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
@@ -5,6 +5,7 @@ package com.daml.platform.apiserver.configuration
 
 import java.util.concurrent.atomic.AtomicReference
 
+import akka.actor.Scheduler
 import akka.stream.scaladsl.{Keep, RestartSource, Sink}
 import akka.stream.{KillSwitches, Materializer, RestartSettings, UniqueKillSwitch}
 import akka.{Done, NotUsed}
@@ -26,11 +27,14 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
   * multiple services and validators require the latest ledger config.
   */
 private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private (
-    index: IndexConfigManagementService,
     ledgerConfiguration: LedgerConfiguration,
+    index: IndexConfigManagementService,
+    scheduler: Scheduler,
+)(implicit
     materializer: Materializer,
-)(implicit loggingContext: LoggingContext)
-    extends CurrentLedgerConfiguration
+    executionContext: ExecutionContext,
+    loggingContext: LoggingContext,
+) extends CurrentLedgerConfiguration
     with AutoCloseable {
 
   private[this] val logger = ContextualizedLogger.get(this.getClass)
@@ -48,15 +52,17 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
   // - Mark the provider as ready if no configuration was found after a timeout
   // - Submit the initial config if none is found after a delay
   startLoading()
-  materializer.scheduleOnce(
+  scheduler.scheduleOnce(
     ledgerConfiguration.configurationLoadTimeout.toNanos.nanos,
-    () => {
-      if (readyPromise.trySuccess(())) {
-        logger.warn(
-          s"No ledger configuration found after ${ledgerConfiguration.configurationLoadTimeout}. The ledger API server will now start but all services that depend on the ledger configuration will return UNAVAILABLE until at least one ledger configuration is found."
-        )
+    new Runnable {
+      override def run(): Unit = {
+        if (readyPromise.trySuccess(())) {
+          logger.warn(
+            s"No ledger configuration found after ${ledgerConfiguration.configurationLoadTimeout}. The ledger API server will now start but all services that depend on the ledger configuration will return UNAVAILABLE until at least one ledger configuration is found."
+          )
+        }
+        ()
       }
-      ()
     },
   )
 
@@ -65,7 +71,6 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
   // If the source of configuration changes proves to be a performance bottleneck,
   // it could be replaced by regular polling.
   private[this] def startLoading(): Future[Unit] = {
-    implicit val executionContext: ExecutionContext = materializer.executionContext
     index
       .lookupConfiguration()
       .map {
@@ -117,7 +122,7 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
           }
           .viaMat(KillSwitches.single)(Keep.right[NotUsed, UniqueKillSwitch])
           .toMat(Sink.ignore)(Keep.left[UniqueKillSwitch, Future[Done]])
-          .run()(materializer)
+          .run()
       )
     )
     ()
@@ -141,13 +146,18 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
 
 private[apiserver] object IndexStreamingCurrentLedgerConfiguration {
   def owner(
-      index: IndexConfigManagementService,
       ledgerConfiguration: LedgerConfiguration,
-  )(implicit
+      index: IndexConfigManagementService,
+      scheduler: Scheduler,
       materializer: Materializer,
-      loggingContext: LoggingContext,
+  )(implicit
+      loggingContext: LoggingContext
   ): ResourceOwner[IndexStreamingCurrentLedgerConfiguration] =
     ResourceOwner.forCloseable(() =>
-      new IndexStreamingCurrentLedgerConfiguration(index, ledgerConfiguration, materializer)
+      new IndexStreamingCurrentLedgerConfiguration(
+        ledgerConfiguration,
+        index,
+        scheduler,
+      )(materializer, materializer.executionContext, loggingContext)
     )
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
@@ -30,9 +30,9 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
     ledgerConfiguration: LedgerConfiguration,
     index: IndexConfigManagementService,
     scheduler: Scheduler,
+    servicesExecutionContext: ExecutionContext,
 )(implicit
     materializer: Materializer,
-    executionContext: ExecutionContext,
     loggingContext: LoggingContext,
 ) extends CurrentLedgerConfiguration
     with AutoCloseable {
@@ -70,13 +70,13 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
             s"Initial ledger configuration lookup did not find any configuration. Looking for new ledger configurations from the ledger beginning."
           )
           latestConfigurationState.set(None -> None)
-      }
-      .map(_ => startStreamingUpdates())
+      }(servicesExecutionContext)
+      .map(_ => startStreamingUpdates())(servicesExecutionContext)
       .failed
       .foreach { exception =>
         readyPromise.tryFailure(exception)
         logger.error("Could not load the ledger configuration.", exception)
-      }
+      }(servicesExecutionContext)
 
   private[this] def configFound(
       offset: LedgerOffset.Absolute,
@@ -132,7 +132,7 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private 
           ()
         }
       },
-    )
+    )(servicesExecutionContext)
   }
 
   /** This future will resolve successfully:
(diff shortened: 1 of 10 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
```diff
@@ -26,46 +27,36 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
+  private val latestConfigurationState = new AtomicReference[StateType](None -> None)
+  private val killSwitch = new AtomicReference[Option[UniqueKillSwitch]](None)
+  private val readyPromise = Promise[Unit]()
 
   // At startup, do the following:
   // - Start loading the ledger configuration
-  // - Mark the provider as ready if no configuration was found after a timeout
   // - Submit the initial config if none is found after a delay
+  // - Mark the provider as ready if no configuration was found after a timeout
+  private val scheduledTimeout = giveUpAfterTimeout()
```
nicu-da, line 52: This is probably out of scope but it seems that the current code is a bit too complex for all that it's done.

For the timeout, can't we use a simpler `Future.firstCompletedOf` where we have the initial config set using an akka `pattern.after` to complete the future after the delay? And create a source using `Source.future` from the previous result, concating the streaming update, then using a `via` with the killswitch. This would eliminate the need for the timeout and the state kept in the killSwitch and the readyPromise.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvider.scala
```diff
@@ -21,25 +21,33 @@ object LedgerConfigProvider {
...
         case Some(writeService) =>
           LedgerConfigProvisioner.owner(
+            ledgerConfiguration = ledgerConfiguration,
             currentLedgerConfiguration = currentLedgerConfiguration,
             writeService = writeService,
             timeProvider = timeProvider,
             submissionIdGenerator = SubmissionIdGenerator.Random,
-            ledgerConfiguration = ledgerConfiguration,
+            scheduler = scheduler,
+            executionContext = materializer.executionContext,
```
nicu-da, line 45: Is this 100% safe? With time you never know where the execution context gets passed, and I think this represents an internal akka executor, which we should try not to abuse.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
```diff
@@ -17,92 +16,87 @@ import com.daml.platform.configuration.LedgerConfiguration
...
+      executionContext: ExecutionContext,
+  )(implicit
+      loggingContext: LoggingContext
+  ): ResourceOwner[Unit] = {
+    implicit val ec: ExecutionContext = executionContext
+    ResourceOwner
+      .forCancellable(() =>
+        scheduler.scheduleOnce(
+          ledgerConfiguration.initialConfigurationSubmitDelay.toNanos.nanos,
+          new Runnable {
```
nicu-da, line 47: Can't we use a single abstract function here to omit the anonymous class?

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
```diff
@@ -81,12 +72,17 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private
           latestConfigurationState.set(None -> None)
       }
       .map(_ => startStreamingUpdates())
-  }
+      .failed
+      .foreach { exception =>
+        readyPromise.tryFailure(exception)
```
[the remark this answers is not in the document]
SamirTalwar-DA: Not sure what we'd log, but I'm open to suggestions.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
SamirTalwar-DA: Unfortunately not. `scheduler.scheduleOnce` is overloaded to accept either a `Runnable` or an `ActorRef`, so Scala doesn't know how to convert the lambda.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvider.scala
SamirTalwar-DA: Nope, I don't like this either. We know that the execution context is only used a little bit, but it could still interfere with Akka's own behavior. I'll pass the execution context designed for the API server through instead.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
SamirTalwar-DA: I think I see what you mean. Definitely worth a shot at cleaning up.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvider.scala
SamirTalwar-DA: Pushed.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
SamirTalwar-DA: I gave this a shot but everything I do makes it messier. Please feel free to take a look yourself once this is merged. :grinning:

nicu-da: Sounds good 😂

nicu-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
```diff
@@ -117,12 +113,28 @@ private[apiserver] final class IndexStreamingCurrentLedgerConfiguration private
...
     )
     ()
   }
 
+  private def giveUpAfterTimeout(): Cancellable = {
+    scheduler.scheduleOnce(
+      ledgerConfiguration.configurationLoadTimeout.toNanos.nanos,
+      new Runnable {
+        override def run(): Unit = {
+          if (readyPromise.trySuccess(())) {
```
fabiotudone-da, line 127: (I "love" the choice of `()` as the `Unit` value in Scala)

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
```diff
@@ -17,92 +16,86 @@ import com.daml.platform.configuration.LedgerConfiguration
...
-            writeService.submitConfiguration(
-              Timestamp.assertFromInstant(timeProvider.getCurrentTime.plusSeconds(60)),
-              submissionId,
-              config.initialConfiguration,
+  private def submitInitialConfig(
+      writeService: state.WriteConfigService,
+      timeProvider: TimeProvider,
+      submissionIdGenerator: SubmissionIdGenerator,
+  )(
+      initialConfiguration: Configuration
```
fabiotudone-da, line 69: It doesn't look like you're using this curry

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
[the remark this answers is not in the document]
fabiotudone-da: When can this happen?

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
```diff
@@ -17,92 +16,86 @@ import com.daml.platform.configuration.LedgerConfiguration
...
+  )(implicit executionContext: ExecutionContext, loggingContext: LoggingContext): Unit = {
+    Future.fromTry(Try(submissionIdGenerator.generate())).flatMap { submissionId =>
```
fabiotudone-da, line 71: When can this fail? Shouldn't we log something specific in that case?

ledger/participant-integration-api/src/test/suite/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfigurationSpec.scala
```diff
@@ -112,11 +132,80 @@ final class IndexStreamingCurrentLedgerConfigurationSpec
...
+        scheduler = scheduler,
+        materializer = materializer,
+        servicesExecutionContext = system.dispatcher,
+      )
+      val resource = owner.acquire()
+
+      resource.asFuture
+        .flatMap { currentLedgerConfiguration =>
+          currentLedgerConfiguration.ready.isCompleted should be(false)
+          resource.release().map(_ => currentLedgerConfiguration)
```
fabiotudone-da, line 179:
```suggestion
          resource.release() // Will cancel reading the configuration
            .map(_ => currentLedgerConfiguration)
```

libs-scala/resources-akka/src/main/scala/com/digitalasset/resources/akka/CancellableResourceOwner.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import akka.actor.Cancellable
+import com.daml.resources.{AbstractResourceOwner, HasExecutionContext, ReleasableResource, Resource}
+
+import scala.concurrent.Future
+
+class CancellableResourceOwner[Context: HasExecutionContext](acquireCancellable: () => Cancellable)
+    extends AbstractResourceOwner[Context, Cancellable] {
+  override def acquire()(implicit context: Context): Resource[Context, Cancellable] =
+    ReleasableResource(Future(acquireCancellable()))(cancellable =>
```
fabiotudone-da, line 14: I understand it's convenient here but I feel like exposing a cancellation as a resource release changes the connotation (and thus hinders readability), as a cancellation can (and perhaps usually will) also trigger some kind of failure. Is this change much worse off without this additional owner?

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/IndexStreamingCurrentLedgerConfiguration.scala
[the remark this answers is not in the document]
SamirTalwar-DA: If `lookupConfiguration` fails, or if the stream ends in a failure. Probably because the connection to the index DB breaks, or something.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
SamirTalwar-DA: No, but I like that the "data" parameters are separate from the "collaborator" arguments. Not sure of a better way to do this; I guess I could extract out a class.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
SamirTalwar-DA: It's injected, so theoretically, whenever. That said, this might be a bit too belt-and-suspenders, because there's only one implementation and it calls `UUID.randomUUID()`, which (I think) should never fail, so I think I'll undo it.

libs-scala/resources-akka/src/main/scala/com/digitalasset/resources/akka/CancellableResourceOwner.scala
SamirTalwar-DA: I wasn't aware that cancellations can trigger failures. Typically resources are only released when the application is shutting down; isn't this the kind of occasion where if there is something pending, it _should_ get some kind of cancellation error?

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
SamirTalwar-DA: I reverted this change.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
fabiotudone-da: I agree this intention is better conveyed by adding a bit more structure rather than through separate parameter lists.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
fabiotudone-da: I agree with the injected argument so I'd rather have logged the (theoretically possible) failure but it's a minor thing.

libs-scala/resources-akka/src/main/scala/com/digitalasset/resources/akka/CancellableResourceOwner.scala
fabiotudone-da: Akka doesn't seem to be explicit (for a change, although I must admit that I haven't looked around too long) about whether a `Cancellable` can throw; I was more thinking about (Java) thread interruption though. OK to leave it for me.

fabiotudone-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvisioner.scala
SamirTalwar-DA: OK, sounds good, I shall do that.

SamirTalwar-DA: Handled in https://github.com/digital-asset/daml/pull/10461/commits/45734de114b8c355355f1910ae896fb41d044a5f.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigProvider.scala
```diff
@@ -12,34 +12,46 @@ import com.daml.ledger.resources.ResourceOwner
 import com.daml.logging.LoggingContext
 import com.daml.platform.configuration.LedgerConfiguration
 
+import scala.concurrent.ExecutionContext
+
 object LedgerConfigProvider {
   def owner(
+      ledgerConfiguration: LedgerConfiguration,
```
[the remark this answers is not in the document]
SamirTalwar-DA: I plan on renaming `LedgerConfiguration` as part of subsequent work.

I'm not sure what the issue is with `CurrentLedgerConfiguration`. And I don't understand why you're suggesting to couple the interface to the implementation; my intent was to keep them separate.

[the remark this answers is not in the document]
miklos-da: Sounds good. My main issue with `CurrentLedgerConfiguration` is the confusing name, I agree with keeping the implementation separate from the interface definition.

PR #10466, opened 2021-08-03
[JSON-API] Change tests to run on the same db instance with different table prefixes

realvictorprm:
This catches more bugs than running e.g. the postgres integration tests
for each prefix while using a fresh db, because collisions cannot be
discovered with that approach.

I don't know yet whether the oracle tests will succeed but the CI will tell :smile: 

PS: This stuff here is full of lazy computations, simply because our jdbc config depends on the database url which isn't directly available at startup :face_with_head_bandage: 

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/http-json-oracle/src/it/scala/http/HttpServiceOracleInt.scala
+++ b/ledger-service/http-json-oracle/src/it/scala/http/HttpServiceOracleInt.scala
@@ -11,14 +11,13 @@ import org.scalatest.matchers.should.Matchers
 
 trait HttpServiceOracleInt extends AbstractHttpServiceIntegrationTestFuns with OracleAroundAll {
   this: AsyncTestSuite with Matchers with Inside =>
-
-  override final def jdbcConfig: Option[JdbcConfig] = Some(jdbcConfig_)
-
-  protected[this] def jdbcConfig_ = JdbcConfig(
+  protected[this] lazy val jdbcConfig_ = JdbcConfig(
     driver = "oracle.jdbc.OracleDriver",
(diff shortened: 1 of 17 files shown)


realvictorprm:
And the oracle build is already failing :joy:

realvictorprm:
image
And that is exactly why I changed the tests, just caught another bug :joy:

realvictorprm:
Good, oracle tests passed :smiley_cat:

cocreature (commented):
This looks broadly sensible but I’m wondering if it’s a bit overkill. Those tests are already slow, running the exact same test twice just to check for the different table prefixes seems excessive and also leads to a huge amount of changes as evidenced by the diff of this PR.

What do you think of the following?
1. Add a new tests which literally just create the schema twice with different schemas and then delete both. This ensures that they do not stomp on each other and they create disjoint tables, constraints, ….
2. Change (not duplicate) the existing tests to set a prefix (doesn’t matter which one).

1 ensures that our DDL operations set a prefix consistently. 2 ensures that our queries use that prefix. So in combination, I think we test the right things while keeping the existing changes more or less untouched and only adding a comparatively fast test for DDL stuff

realvictorprm:
@cocreature to be honest, I noticed that yesterday too and thought about how much sense it makes to do what you described vs what we currently do and also came to the conclusion you had yesterday :facepalm: :face_with_head_bandage:. However I did not decide on it yet because I wanted to have your feedback on the changes first (as I could have overseen something).

As you came to the same conclusion I'll change the tests accordingly :smiley:

PR #10467, opened 2021-08-03
Add submission id to the GRPC api [KVL-999]

nicu-da:
This allows to propagate a submission id through the GRPC api. 

If no id is submitted (the submission id is empty) then we generate a new submission id (previous behavior)

### Pull Request Checklist

trigger the build.


ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
```diff
@@ -180,6 +180,15 @@ final class CommandsValidator(ledgerId: LedgerId, submissionIdGenerator: Submiss
       case ProtoEmpty =>
         Left(missingField("command"))
     }
+
+  private def matchSubmissionId(commands: ProtoCommands) = {
```
miklos-da, line 184: I'd recommend `determineSubmissionId` or `takeOrGenerateSubmissionId` or similar name for this helper function ('match' implies we're using some kind of a mapping but we aren't).

nickchapman-da (commented):
Thanks. Haskell changes look good to me.

miklos-da (approved):

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
SamirTalwar-DA: Perhaps `extractOrGenerateSubmissionId`?

SamirTalwar-DA (approved):

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
nicu-da: Fully aggree, will update to `extractOrGenerateSubmissionId`

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
```diff
@@ -180,6 +180,15 @@ final class CommandsValidator(ledgerId: LedgerId, submissionIdGenerator: Submiss
...
         Left(missingField("command"))
     }
+
+  private def extractOrGenerateSubmissionId(commands: ProtoCommands) = {
+    if (commands.submissionId.isEmpty) {
+      domain.SubmissionId(submissionIdGenerator.generate())
+    } else {
+      domain.SubmissionId(Ref.SubmissionId.assertFromString(commands.submissionId))
+    }
+  }
```
fabiotudone-da, line 190: I think you don't need the curly braces here.

ledger-service/http-json/src/main/scala/com/digitalasset/http/CommandService.scala
```diff
@@ -42,15 +42,15 @@ class CommandService(
...
     withEnrichedLoggingContext(
       label[TemplateId.RequiredPkg],
       "template_id" -> templateId.toString,
     )
 
-  def withTemplateChoiceLoggingContext[T](
+  private def withTemplateChoiceLoggingContext[T](
       templateId: TemplateId.RequiredPkg,
       choice: domain.Choice,
   )(implicit lc: LoggingContextOf[T]): withEnrichedLoggingContext[Choice, RequiredPkg with T] =
```
fabiotudone-da, line 56: 👍🏼

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -95,6 +95,15 @@ message Commands {
   // to read contract data on behalf of each of the given parties.
   // Optional
   repeated string read_as = 13;
+
+  // A unique identifier to distinguish completions for different submissions with the same change ID.
+  // Typically a random UUID. Applications are expected to use a different UUID for each retry of a submission
+  // with the same change ID.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  //
+  // If omitted, the participant or Daml ledger may set a value of their choice.
```
fabiotudone-da, line 104: Is `Daml ledger` accepted docs terminology for the committer?

fabiotudone-da (approved):

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
nicu-da: Not really sure, I copied the description from the  [PR](https://github.com/digital-asset/daml/pull/9567/files#diff-ed4ed30e4f593aaf9d013b39ed5272dfc0df08c548cc78f110fd4471eaa6f59cR137)

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
nicu-da: I usually prefer to always have them in cases like this. Is the convention to remove them in the codebase?

SamirTalwar-DA: LOL, "convention".

I think we _tend_ to use the curly braces when there are multiple lines or side effects, and avoid them otherwise. In addition, if one side of the `if` gets braces, the other side gets them too.

This is just anecdotal, though. I don't think there's much consistency.

fabiotudone-da: > I usually prefer to always have them in cases like this

What do they add?

> I think we tend to use the curly braces when there are multiple lines or side effects, and avoid them otherwise. In addition, if one side of the if gets braces, the other side gets them too.

My personal guideline is "all else being equal, less is better".

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
fabiotudone-da: Asking because it sounds ambiguous to me, I'd rather have `the participant or the committer` or something very general like `daml runtime` or `daml system`

rautenrieth-da:
Allowing users to choose the submission ID means they might not be unique. Did you check whether this has any implications? We should add a test to the ledger conformance tool that submits two commands with the same submission id (and a test that simply uses a non-empty value for the submission id).

The party and config management services also allow the user to choose a submission ID. Because the submission ID needs to be unique for these services, the ledger API server appends a UUID to the user-chosen value. See also [this corresponding test](https://github.com/digital-asset/daml/blob/main/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala#L189). This approach might not be compatible with the command deduplication v2 design however.

miklos-da:
> Allowing users to choose the submission ID means they might not be unique. Did you check whether this has any implications? We should add a test to the ledger conformance tool that submits two commands with the same submission id (and a test that simply uses a non-empty value for the submission id).
> 
> The party and config management services also allow the user to choose a submission ID. Because the submission ID needs to be unique for these services, the ledger API server [appends a UUID to the user-chosen value](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiConfigManagementService.scala#L126). See also [this corresponding test](https://github.com/digital-asset/daml/blob/main/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala#L189). This approach might not be compatible with the command deduplication v2 design however.

Thanks for your input, @rautenrieth-da. We intend to extend conformance test coverage in a separate PR for populating submission IDs (we need to propagate these back to the client first). We don't rely on uniqueness of the submission ID itself for commands, these will only help the client to identify which completion belongs to which command submission.

Thanks for pointing out the assumptions and handling of submission IDs for party and config submissions.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
nicu-da: > What do they add?

The reduce the chance that someone adds new code that he thinks it's inside the if block but because of the lack of braces it's actually outside the if block. 

> My personal guideline is "everything else being equal, less is better".

My main reason is that I prefer consistency, I know exactly what to expect and having all the if blocks with braces makes it easier to compute. 


Will remove the braces though as it's used in multiple places without it.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
nicu-da: Good point, will update to `the participant or the committer`

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
fabiotudone-da: Note that I was referring to the outer braces, not the ones of the `if...else` expression

fabiotudone-da: I agree that consistency of braces for `if...else` is useful and I think in the committer team we settled on rather having them.

nicu-da: Ah, makes more sense now 👍 my brain fixated on the if. will change that

fabiotudone-da: (Although if you remove both the `if...else` braces and the outer braces, then there is no change one could inadvertently add code outside of the `if` nor `else` blocks)

PR #10469, opened 2021-08-03
feature: remember checkboxes in script view

### LINKED ISSUE #9177: Feature request: In IDE script table view remember checkbox choices
When making changes to a script, the "show archived" and "show detailed disclosure" choices get reset. Would be more convenient if these keep my last setting

robin-da:
Fixes #9177. This makes the script view remember the choices for
'show_archive', 'show detailed disclosure' and 'table/transaction' view
accross changes to the script.

Not sure if there is a test suite availabe for the extension where tests for
this could be added.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-extension/src/webview.js
+++ b/compiler/daml-extension/src/webview.js
@@ -3,17 +3,27 @@
 
 const vscode = acquireVsCodeApi();
 function show_archived_changed() {
-  document.body.classList.toggle('hide_archived', !document.getElementById('show_archived').checked);
+  const isChecked = document.getElementById('show_archived').checked;
+  document.body.classList.toggle('hide_archived', !isChecked);
+  vscode.postMessage({
+    'command': 'set_show_archived',
+    'value': isChecked
+  });
 }
 function toggle_detailed_disclosure() {
-  document.body.classList.toggle('hidden_disclosure', !document.getElementById('show_detailed_disclosure').checked);
+  const isChecked = document.getElementById('show_detailed_disclosure').checked;
+  document.body.classList.toggle('hidden_disclosure', !isChecked);
+  vscode.postMessage({
+    'command': 'set_show_detailed_disclosure',
+    'value': isChecked
+  });
 }
 function toggle_view() {
   document.body.classList.toggle('hide_transaction');
   document.body.classList.toggle('hide_table');
   vscode.postMessage({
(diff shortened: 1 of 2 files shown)


compiler/daml-extension/src/webview.js
```diff
@@ -34,9 +44,28 @@ window.addEventListener('message', event => {
...
+      if (message.value.showDetailedDisclosure) {
+        document.body.classList.remove('hidden_disclosure');
+        document.body.classList.add('show_disclosure');
+        document.getElementById('show_detailed_disclosure').checked = true;
+      } else {
+        document.body.classList.remove('show_disclosure');
+        document.body.classList.add('hidden_disclosure');
+        document.getElementById('show_detailed_disclosure').checked = false;
+      }
+
```
cocreature, line 68: I think you can factor this out into a function that accepts 4 arguments:
1. The boolean that controls the state.
2. The class that is active when the boolean is true.
3. The class that is active when the boolean is false
4. The id of the checkbox.

cocreature: You could also try fitting the transaction vs table view in there. Not sure if that’s helpful or not.

compiler/daml-extension/src/webview.js
```diff
@@ -3,17 +3,27 @@
...
+  });
 }
 function toggle_detailed_disclosure() {
-  document.body.classList.toggle('hidden_disclosure', !document.getElementById('show_detailed_disclosure').checked);
+  const isChecked = document.getElementById('show_detailed_disclosure').checked;
+  document.body.classList.toggle('hidden_disclosure', !isChecked);
+  vscode.postMessage({
+    'command': 'set_show_detailed_disclosure',
+    'value': isChecked
+  });
```
cocreature, line 19: These two can also be factored into one.

cocreature (approved):
Thanks! We don’t have tests for this atm so I’m fine with omitting them. A few suggestions for cleanup.

compiler/daml-extension/src/webview.js
robin-da: factored out in a function.

compiler/daml-extension/src/webview.js
robin-da: factored out.

PR #10471, opened 2021-08-03
LF: Log internal errors

remyhaemmerle-da:
part of #9974

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Utf8.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Utf8.scala
@@ -11,7 +11,6 @@ import com.google.protobuf.ByteString
 import scalaz.Order
 
 import scala.annotation.tailrec
-import scala.collection.compat._
 import scala.jdk.CollectionConverters._
 
 // The Daml-LF strings are supposed to be UTF-8 while standard java strings are UTF16
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Utf8.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Utf8.scala
@@ -73,10 +73,14 @@ object Utf8 {
   def unpack(s: String): ImmArray[Long] =
     ImmArray(s.codePoints().iterator().asScala.map(_.toLong).iterator.to(Iterable))
 
-  @throws[IllegalArgumentException]
-  def pack(codePoints: ImmArray[Long]): String = {
+  // Converts the List of Unicode code point into a String if all code point are valid.
+  // Returns the first invalid code point as Left otherwise.
+  def pack(codePoints: ImmArray[Long]): Either[Long, String] = {
     val builder = new StringBuilder()
-    for (cp <- codePoints) {
+    var illegalCodePoint = Option.empty[Long]
+    val iterator = codePoints.iterator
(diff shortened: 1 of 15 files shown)


daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/Utf8.scala
```diff
@@ -73,10 +73,14 @@ object Utf8 {
   def unpack(s: String): ImmArray[Long] =
     ImmArray(s.codePoints().iterator().asScala.map(_.toLong).iterator.to(Iterable))
 
-  @throws[IllegalArgumentException]
-  def pack(codePoints: ImmArray[Long]): String = {
+  // Converts the List of Unicode code point into a String if all code point are valid.
+  // Returns the first invalid code point as Left otherwise.
+  def pack(codePoints: ImmArray[Long]): Either[Long, String] = {
```
remyhaemmerle-da, line 78: Just better to return the error explicit

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -433,7 +433,12 @@ private[lf] object SBuiltin {
   final case object SBCodePointsToText extends SBuiltinPure(1) {
     override private[speedy] def executePure(args: util.ArrayList[SValue]): SText = {
       val codePoints = getSList(args, 0).map(_.asInstanceOf[SInt64].value)
-      SText(Utf8.pack(codePoints.toImmArray))
+      Utf8.pack(codePoints.toImmArray) match {
+        case Right(value) =>
+          SText(value)
+        case Left(cp) =>
+          crash(s"invalid code point 0x${cp.toHexString}.")
```
remyhaemmerle-da, line 440: Maybe  this one should be an `ArithmeticError` :cry:

remyhaemmerle-da:
@cocreature What do you think about the content ?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
cocreature: How hard would it be to make this dependent on the LF version? It seems like we could just have a flag in this primitive on whether it can throw or not and then the speedy compiler can set it to true f we’re in LF >= 1.15 while leaving it false for older versions.

cocreature (approved):
Looks very sensible, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: I think the simpler would be two have two different builtins in the scala AST. 
If we do not want to have two builtins in the compiler/archive, we can hack the archive decoder to map `CODE_POINTS_TO_TEXT` to one or the other depending of the LF version.

However I am not sure, the trouble is worthwhile. I think I will let the builtin crashing forever.

cocreature: yeah I think we can wait until we get a feature request for it. I’m not all that concerned about it.

PR #10472, opened 2021-08-03
DPP-460 Parameter storage consolidation

rautenrieth-da:
Changes (each in a separate commit):

- Change `DbDispatcher` such that it is possible to run SQL blocks in a custom transaction isolation mode
- Refactor `ParameterStorageBackend`:
  - Consolidate various methods for getting the ledger end into a single one that returns both the offset and event sequential id of the ledger end. **Note: this may have a performance impact**
  - Consolidate various ledger initialization methods into a single one that is guaranteed to be atomic
- Add a test for `ParameterStorageBackend.initializeParameters`. This test will likely have to be refactored as part of DPP-468, such that it runs on all backends.


rautenrieth-da:
@nmarton-da I have changed the `ParameterStorageBackend` interface to use a `Unit` return type instead of the previous result ADT. This means callers won't know anymore whether their call was responsible for initializing a *new* database. This has two implications:
- Logging has moved into `ParameterStorageBackend` (logging whether we initialized a new database or created a new one). The exact content of the log line will change as well.
- `SqlServer` does an extra query to find out whether it's a new database, because it needs to upload some packages only if the database is new. This makes the initialization not atomic, and we'd get problems if two `SqlServer`s start at the same time. We always had this problem, it just means this PR will not fix it.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -87,46 +87,66 @@ private class JdbcLedgerDao(
...
-    dbDispatcher.executeSql(metrics.daml.index.db.initializeLedgerParameters) {
-      implicit connection =>
-        storageBackend.updateLedgerId(ledgerId.unwrap)(connection)
-    }
-
-  override def initializeParticipantId(
-      participantId: ParticipantId
+    dbDispatcher
+      .executeSql(metrics.daml.index.db.getInitialLedgerEnd)(storageBackend.ledgerEnd)
+      .map { end =>
```
nmarton-da, line 133: how about .map(_.lastOffset)(DEC)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -87,46 +87,66 @@ private class JdbcLedgerDao(
...
+
+  override def initialize(
+      ledgerId: LedgerId,
+      participantId: ParticipantId,
   )(implicit loggingContext: LoggingContext): Future[Unit] =
-    dbDispatcher.executeSql(metrics.daml.index.db.initializeParticipantId) { implicit connection =>
-      storageBackend.updateParticipantId(participantId.unwrap)(connection)
-    }
+    dbDispatcher
+      .executeSql(metrics.daml.index.db.initializeLedgerParameters) { implicit connection =>
```
nmarton-da, line 142: yo can omit the `{ implicit => connection` ... `(connection) }` and switch to `( ... )` instead.
(curry instead of implicit resolution)

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -710,7 +730,7 @@ private class JdbcLedgerDao(
   private[this] def validateOffsetStep(offsetStep: OffsetStep, conn: Connection): Offset = {
     offsetStep match {
       case IncrementalOffsetStep(p, o) =>
-        val actualEnd = storageBackend.ledgerEndOffset(conn)
+        val actualEnd = storageBackend.ledgerEnd(conn).lastOffset.getOrElse(Offset.beforeBegin)
```
nmarton-da, line 733: as mentioned before, I would reach out here to some logic on ParameterStorageBackend for resolving defaults

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
```diff
@@ -168,7 +169,9 @@ object ParallelIndexerFactory {
             )
             .use { dbDispatcher =>
               dbDispatcher
-                .executeSql(metrics.daml.parallelIndexer.initialization)(storageBackend.initialize)
+                .executeSql(metrics.daml.parallelIndexer.initialization)(
```
nmarton-da, line 172: I am missing the part what we discussed before, and moves append only ledger initialization here. 
Will this be part of some other task or the same? Will this be part of a different PR?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -78,39 +78,43 @@ trait IngestionStorageBackend[DB_BATCH] {
...
-  def initialLedgerEnd(connection: Connection): Option[Offset]
+  def ledgerEnd(connection: Connection): StorageBackend.OptionalLedgerEnd
 
   /** Part of pruning process, this needs to be in the same transaction as the other pruning related database operations
     */
   def updatePrunedUptoInclusive(prunedUpToInclusive: Offset)(connection: Connection): Unit
   def prunedUptoInclusive(connection: Connection): Option[Offset]
+
+  /** Initializes the parameters table and verifies or updates ledger identity parameters:
+    *  - If no identity parameters are stored, then they are set to the given value.
```
nmarton-da, line 107: I think it would be sufficient to say that this method is idempotent. (if you prefer to still keep some/all of these details anyway, please add somewhere a line that it is idempotent)

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -111,17 +118,25 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
           |
           |""".stripMargin
       )
-    )(rs =>
-      StorageBackend.LedgerEnd(
+    )(rs => {
+      val rawLedgerEnd = rs.getString(1) // SQL NULL is mapped to JVM null
+      val rawEventSeqId = rs.getLong(2) // SQL NULL is mapped to JVM 0
+      val rawEventSeqIdIsNull = rs.wasNull()
```
nmarton-da, line 124: nice catch, previous code was not resilient for having null event-seq-ids (although accidentally it was still correct :) )

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -8,7 +8,7 @@ import java.time.Instant
 import java.util.Date
 
 import anorm.SqlParser.{array, binaryStream, bool, byteArray, date, flatten, get, int, long, str}
-import anorm.{Macro, Row, RowParser, SQL, SimpleSql, SqlParser, SqlStringInterpolation, ~}
+import anorm.{Macro, RowParser, SQL, SqlParser, SqlStringInterpolation, ~}
```
nmarton-da, line 11: just a note: I would have liked better not to review intermittent changes, and review the final version with anorm here instead (which now lives in another PR)

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -133,73 +148,74 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
+        discard(
+          SQL"update #$TableName set #$ParticipantIdColumnName=${participantId.unwrap: String}"
+            .execute()(connection)
+        )
+      case StorageBackend.OptionalIdentityParams(Some(`ledgerId`), Some(`participantId`)) =>
+        logger.info(
+          s"Found existing database for ledgerId '${params.ledgerId}' and participantId '${params.participantId}'"
+        )
+      case StorageBackend.OptionalIdentityParams(_, Some(existing))
+          if existing != params.participantId =>
```
nmarton-da, line 196: I think if you would move the line204 case up, then you would not need to check for equality here

ledger/participant-integration-api/src/main/scala/platform/store/dao/LedgerDao.scala
```diff
@@ -232,15 +232,22 @@ private[platform] trait LedgerReadDao extends ReportsHealth {
 
 private[platform] trait LedgerWriteDao extends ReportsHealth {
 
-  /** Initializes the ledger. Must be called only once.
+  /** Initializes the database with the given ledger identity.
+    * If the database was already intialized, instead compares the given identity parameters
+    * to the existing ones, and returns a Future failed with ??? if they don't match.
```
nmarton-da, line 237: ???

ledger/participant-integration-api/src/main/scala/platform/store/dao/ParametersTable.scala
```diff
@@ -56,9 +56,11 @@ private[store] object ParametersTable {
     )
 
   def getParticipantId(connection: Connection): Option[ParticipantId] =
-    SQL"select #$ParticipantIdColumnName from #$TableName".as(ParticipantIdParser.single)(
-      connection
-    )
+    SQL"select #$ParticipantIdColumnName from #$TableName"
```
nmarton-da, line 59: this changes the behavior. was it delibare? why is this needed?

ledger/participant-state-index/src/main/scala/com/daml/ledger/participant/state/index/v2/package.scala
```diff
@@ -114,4 +114,5 @@ package v2 {
   /** This command was submitted before. */
   final case class CommandDeduplicationDuplicate(deduplicateUntil: Instant)
       extends CommandDeduplicationResult
+
```
nmarton-da, line 117: why this change?

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendPostgresSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  // Each test gets its own database and its own connection pool
+  private var resource: Resource[DbDispatcher] = _
+  protected var dbDispatcher: DbDispatcher = _
+
+  protected def executeSql[T](sql: Connection => T): Future[T] = {
+    dbDispatcher.executeSql(metrics.test.db)(sql)
+  }
+
+  override protected def beforeEach(): Unit = {
```
nmarton-da, line 46: what ensures that each test is running in isolation? (If I understand correctly this is a prerequisite to be able to use this, right?)

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendPostgresSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    super.beforeEach()
+
+    // TODO: use a custom execution context, like JdbcLedgeDao.beforeAll()
+    implicit val resourceContext: ResourceContext = ResourceContext(system.dispatcher)
+
+    resource = for {
+      _ <- Resource.fromFuture(
+        new FlywayMigrations(jdbcUrl).migrate(enableAppendOnlySchema = true)
```
nmarton-da, line 54: I am slightly worried about creating a db, flywaymigrating it, and dropping it for each and every test case. Let's revisit this aspect if it becomes unfeasible. (for example JdbcLedgerDao tests use ...AroundAll instead)

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/StorageBackendPostgresInitializationSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import com.daml.ledger.api.domain.{LedgerId, ParticipantId}
+import com.daml.lf.data.Ref
+import com.daml.platform.common.MismatchException
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+final class StorageBackendPostgresInitializationSpec
+    extends AsyncFlatSpec
+    with StorageBackendPostgresSpec
```
nmarton-da, line 14: I think this needs some further improvements here: we need a framework to have tests declared only once, and executed for all backends (H2, Postgres, Oracle, similar to JdbcLedgerDao tests).

nmarton-da (changes requested):

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendPostgresSpec.scala
rautenrieth-da: This is given by ScalaTest. From [the docs](https://www.scalatest.org/user_guide/async_testing):

> By default (unless you mix in ParallelTestExecution), tests in an asynchronous style traits will be executed one after another, i.e., serially. This is true whether those tests return Assertion or Future[Assertion], no matter what threads are involved.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/StorageBackendPostgresInitializationSpec.scala
rautenrieth-da: This was just a quick implementation to have _some_ unit tests. PR https://github.com/digital-asset/daml/pull/10529 adds a whole framework for running the tests on each backend.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/common/MismatchException.scala
```diff
@@ -15,10 +15,14 @@ abstract class MismatchException[A](
 
 object MismatchException {
 
-  class LedgerId(existing: domain.LedgerId, provided: domain.LedgerId)
-      extends MismatchException[domain.LedgerId]("ledger id", existing, provided)
+  case class LedgerId(
```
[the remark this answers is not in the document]
rautenrieth-da: It was mostly so that we can pattern match on the exception.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -283,9 +287,13 @@ object DBLockStorageBackend {
 }
 
 object StorageBackend {
-  case class Params(ledgerEnd: Offset, eventSeqId: Long)
-
-  case class LedgerEnd(lastOffset: Option[Offset], lastEventSeqId: Option[Long])
+  case class LedgerEnd(lastOffset: Offset, lastEventSeqId: Long)
+  case class OptionalLedgerEnd(lastOffset: Option[Offset], lastEventSeqId: Option[Long])
```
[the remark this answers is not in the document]
rautenrieth-da: I thought about this, but I was worried about corner cases:
- The ledger_end offset moves for all kind of updates (including, e.g., party allocations), while the ledger_end_sequential_id only moves for `TransactionAccepted` updates.
- You can have an existing mutating schema database with a ledger end offset, but no events. If you migrate it, `ledger_end_sequential_id` will be `NULL`.

We could add a migration that fills the `ledger_end_sequential_id` column with 0 if the value is missing but `ledger_end` is defined. This would hardcode the constant for "no event yet" in a SQL migration, but that's ok. The the invariant you mentioned would hold.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -283,9 +287,13 @@ object DBLockStorageBackend {
...
+  case class IdentityParams(ledgerId: LedgerId, participantId: ParticipantId)
+  case class OptionalIdentityParams(
```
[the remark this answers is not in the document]
rautenrieth-da: > Simillarily here it seems not necessary since initialization always sets up both.

Same as above, it's not obvious that existing databases will always have both columns defined. We could make the participantId `NOT NULL` (same as the ledger id) and fix missing values in a migration to enforce this. This would clean up some code.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: Participant id mismatch has a higher priority than ledger id mismatch (in cases where both mismatch).

ledger/participant-integration-api/src/main/scala/platform/store/dao/ParametersTable.scala
rautenrieth-da: Before: crash if parameter table is empty
After: return `None` if parameter table is empty

This makes `getParticipantId` behave the same as `getLedgerId`. This was changed for sandbox initialization - SqlLedger reads the ledger and participant id before attempting to initializing the database.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendPostgresSpec.scala
rautenrieth-da: Good point. In https://github.com/digital-asset/daml/pull/10529, this is changed to only create one database for the entire suite, and resetting the content between tests.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -89,16 +94,18 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
       |""".stripMargin
   )
 
-  def updateParams(params: StorageBackend.Params)(connection: Connection): Unit = {
+  override def updateLedgerEnd(
```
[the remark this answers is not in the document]
rautenrieth-da: `CommonStorageBackend` currently implements all `StorageBackend` traits, why would you move only the `ParameterStorageBackend` implementation to a different place? 

Wouldn't we then need to create a `CommonPartyStorageBackend`, `CommonPackageStorageBackend`, `CommonEventStorageBackend`, ...?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: > The ledger_end offset moves for all kind of updates (including, e.g., party allocations), while the ledger_end_sequential_id only moves for TransactionAccepted updates.

That is true, but the ledger end never the less is set by setting them both. We do not set only one.

> You can have an existing mutating schema database with a ledger end offset, but no events. If you migrate it, ledger_end_sequential_id will be NULL.

That is true, but that can be circumvented by falling back to the default upon null, because this corner case allows it (offset some, event_seq_id null case). No additional migration is needed.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/common/MismatchException.scala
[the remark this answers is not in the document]
nmarton-da: makes sense. Still, would you be ok to remove those `new` -s?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: Kind of would make sense to me, yes. Not needs to be done in this iteration though.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: Simillarily as above, the idea would be to hide the special treatments of one set and the other not in implementation, and not letting this leak out on the interface level. 
But no need in this case, I am ok with how it is right now

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: are you sure? looking at the original indexer code I had another impression :thinking: 
also: I am not sure it matters too much, either way we are crashing, no?
(also: it seem to me both is possible as well, which would be quite hard to express with the current exceptions...I am not arguing to change that, just an observation)

ledger/participant-integration-api/src/main/scala/platform/store/dao/ParametersTable.scala
nmarton-da: makes sense

nmarton-da:
FYI: merged the ContractStorageBackend consolidation, conflict is coming from there probably, resolving should be trivial, but pls reach out if having trouble there

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/common/MismatchException.scala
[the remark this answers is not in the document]
rautenrieth-da: Of course, I'll remove those.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: Addressed by rewriting the code with anorm

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: I have merged the anorm change into this PR

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: Actually, you are right, I have moved the lines around. Thanks for catching this!

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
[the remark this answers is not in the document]
rautenrieth-da: I have refactored the code according to this discussion.

ledger/participant-integration-api/src/main/scala/platform/indexer/parallel/ParallelIndexerFactory.scala
rautenrieth-da: I wasn't planning on doing this as part of DPP-460

nmarton-da: sure no probs, created DPP-551 for this, I will do it as part of HA stabilization

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -78,39 +78,61 @@ trait IngestionStorageBackend[DB_BATCH] {
...
+
+  /** Query the current ledger end, returning a value that points to a point before the ledger begin
+    * if no ledger end exists.
+    * No significant CPU load, mostly blocking JDBC communication with the database backend.
+    *
+    * @param connection to be used to get the LedgerEnd
+    * @return the current LedgerEnd, or a LedgerEnd that points to before the ledger begin if no ledger end exists
+    */
+  final def ledgerEndOrBeforeBegin(connection: Connection): ParameterStorageBackend.LedgerEnd =
+    ledgerEnd(connection).getOrElse(ParameterStorageBackend.LedgerEnd(Offset.beforeBegin, 0L))
```
nmarton-da, line 109: this will be improved later, right?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -87,46 +87,55 @@ private class JdbcLedgerDao(
   override def currentHealth(): HealthStatus = dbDispatcher.currentHealth()
 
   override def lookupLedgerId()(implicit loggingContext: LoggingContext): Future[Option[LedgerId]] =
-    dbDispatcher.executeSql(metrics.daml.index.db.getLedgerId)(storageBackend.ledgerId)
+    dbDispatcher
+      .executeSql(metrics.daml.index.db.getLedgerId)(storageBackend.ledgerIdentity)
+      .map(_.map(_.ledgerId))(DEC)
```
nmarton-da, line 92: minor, not necessarily needs to change: alternatively if you do the .map(_.ledgerId) in the executeSql block, then we do not need the extra mapping, neither an EC.
upto you

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -24,43 +24,50 @@ import com.daml.platform.store.Conversions.{
...
-      "DELETE FROM package_entries WHERE ledger_offset > ?",
-      "DELETE FROM packages WHERE ledger_offset > ?",
-      "DELETE FROM participant_command_completions WHERE completion_offset > ?",
-      "DELETE FROM participant_events_divulgence WHERE event_offset > ?",
-      "DELETE FROM participant_events_create WHERE event_offset > ?",
-      "DELETE FROM participant_events_consuming_exercise WHERE event_offset > ?",
-      "DELETE FROM participant_events_non_consuming_exercise WHERE event_offset > ?",
-      "DELETE FROM parties WHERE ledger_offset > ?",
-      "DELETE FROM party_entries WHERE ledger_offset > ?",
+      SQL("DELETE FROM configuration_entries WHERE ledger_offset > {ledger_offset}"),
```
nmarton-da, line 46: just an extra info, for future reference: anorm claims string interpolation is superior (faster, safer), so we should probably aim to use that when possible.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -24,43 +24,50 @@ import com.daml.platform.store.Conversions.{
...
+import scalaz.syntax.tag._
 
 private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_BATCH] {
 
+  private val logger: ContextualizedLogger = ContextualizedLogger.get(this.getClass)
+
   // Ingestion
 
-  private val preparedDeleteIngestionOverspillEntries: List[String] =
+  private val SQL_DELETE_OVERSPILL_ENTRIES: List[SqlQuery] =
```
nmarton-da, line 44: constant naming is camelcase starting with capital in scala

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -70,62 +77,40 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
 
   // Parameters
 
-  private val preparedUpdateLedgerEnd: Connection => PreparedStatement = _.prepareStatement(
+  private val SQL_UPDATE_LEDGER_END = SQL(
```
nmarton-da, line 80: constant naming

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: Sorry, maybe there was a misunderstanding: I would definitely extract the parameter storage backend related stuff in this iteration, I would only not do the rest what you were referring to

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/common/MismatchException.scala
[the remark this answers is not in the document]
nmarton-da: I still see occurrences to new

nmarton-da (approved):
Awsome job navigating these troubled waters :+1: 
thanks for your patience with all my comments here
few minor things left, I reopened a couple of resolved comments as well. nothing big.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
rautenrieth-da: IMO, consistency within a project is more important than following naming standards. Almost everywhere in our code, SQL constants are all uppercase.

nmarton-da: well, strictly speaking this is not an SQL constant in that sense, but a list of SQL constants ;)
but you are right we really seem to have this pattern established.
after we cleaned dao/ most of it will disappear, and the rest should be rewritten to string interpolation anyway, so we could part from this approach if we want then. (I created a cleanup task DPP-553 to capture this aspect)

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
```diff
@@ -548,8 +548,7 @@ CREATE TABLE parameters
     ledger_id                          NVARCHAR2(1000) not null,
     -- stores the head offset, meant to change with every new ledger entry
     ledger_end                         VARCHAR2(4000),
-    external_ledger_end                NVARCHAR2(1000),
-    participant_id                     NVARCHAR2(1000),
+    participant_id                     NVARCHAR2(1000) not null,
```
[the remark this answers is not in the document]
rautenrieth-da: Resolved elsewhere: yes, we can still change it.

PR #10482, opened 2021-08-04
[JSON-API] Move database independent tests into a seperate abstract test

realvictorprm:
The DatabaseStartupOps tests are now also tested against Oracle.
Besides, an additional test will now cover that table creation
doesn't run into name collisions for different table prefixes within
the same database.

Besides, this also fixes a bug in the oracle query which was introduced by my schema version feature.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


realvictorprm:
The new tests already proof valueable, thanks to them I just discovered another bug for the oracle queries :facepalm: :smile:

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -120,7 +120,9 @@ sealed abstract class Queries(tablePrefix: String) {
 
   private[http] def dropAllTablesIfExist(implicit log: LogHandler): ConnectionIO[Unit] = {
     import cats.instances.vector._, cats.syntax.foldable.{toFoldableOps => ToFoldableOps}
-    initDatabaseDdls.reverse
+    initDatabaseDdls
```
cocreature, line 123: why is `createVersionTable` not part of `initDatabaseDdls`?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -727,7 +729,7 @@ private final class OracleQueries(tablePrefix: String) extends Queries(tablePref
   protected[this] override def agreementTextType = sql"NCLOB"
 
   protected[this] override def jsonColumn(name: Fragment) =
-    sql"$name CLOB NOT NULL CONSTRAINT ensure_json_$name CHECK ($name IS JSON)"
+    sql"$name CLOB NOT NULL CONSTRAINT ${tablePrefixFr}ensure_json_$name CHECK ($name IS JSON)"
```
cocreature, line 732: So this was one bug I assume? You only hit this if you actually set a prefix right?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -756,18 +758,20 @@ private final class OracleQueries(tablePrefix: String) extends Queries(tablePref
   protected[this] override def initDatabaseDdls =
     super.initDatabaseDdls ++ Seq(stakeholdersView, stakeholdersIndex)
 
-  protected[http] override def version()(implicit log: LogHandler): ConnectionIO[Option[Int]] = {
+  protected[http] override def version()(implicit log: LogHandler): ConnectionIO[Option[Int]] =
```
cocreature, line 761: Presumably this was the other bug. What did the code actually do? Return `None` in all cases making `createSchema=false` always fail since the schema doesn’t exist? How do any of our tests work in that case?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -756,18 +758,20 @@ private final class OracleQueries(tablePrefix: String) extends Queries(tablePref
...
-                WHERE table_name = '$jsonApiSchemaVersionTableName'
-              )""".query[Boolean].unique
+      // Note that Oracle table names seem to be somewhat case sensitive,
+      // but are inside the USER_TABLES table all uppercase.
+      res <-
+        sql"""SELECT 1 FROM USER_TABLES
+              WHERE TABLE_NAME = UPPER('$jsonApiSchemaVersionTableName')"""
+          .query[Int]
+          .option
+      _ = println(s"res: $res")
```
cocreature, line 770: leftover debug print

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -756,18 +758,20 @@ private final class OracleQueries(tablePrefix: String) extends Queries(tablePref
...
       version <-
-        if (!doesTableExist) connection.pure(None)
+        if (res.isEmpty) connection.pure(None)
         else sql"SELECT version FROM $jsonApiSchemaVersionTableName".query[Int].option
```
cocreature, line 773:
```suggestion
        res.flatMap { _ =>
         sql"SELECT version FROM $jsonApiSchemaVersionTableName".query[Int].option
       }
```

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
```diff
@@ -26,136 +20,6 @@ class HttpServiceWithPostgresIntTest
 
   override def wsConfig: Option[WebsocketConfig] = None
 
-  implicit lazy val logHandler: log.LogHandler = dao.logHandler
```
cocreature, line 29: so the issue was that we ran all startup tests only for postgres?

cocreature: Actually I still don’t get it. The regular tests should still create the schema and then startup. Do we skip schema validation in that case? That seems like something we might want to address. The startup shouldn’t work differently in the tests than it works outside of the tests.

cocreature (approved):
Thanks! I’d like to understand the effects of the buggy code a bit better:
1. It looks like a constraint wasn’t prefixed if you set one which doesn’t matter if you don’t set it so relatively small impact.
2. Afaict, the JSON API always failed schema validation on Oracle so it is impossible to start it. Is that correct?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: darn, thanks :face_with_head_bandage:

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: This is the bug which only appears if you run two instances of the json api with different prefixes.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: noice, thanks

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
realvictorprm: Yes, the startup specific tests were only run for postgres. However that they were not run for Oracle was more or less me not completely knowing & understanding the test suite yet.

Now that we are talking about it, you're right. This was never changed, however I will change this and hope nothing breaks.

realvictorprm: The reason why the tests were still passing is simply because the code did not contain an error but just an semantic bug. So it would always reinitialize the cache (if allowed by the startup mode).

realvictorprm:
> Afaict, the JSON API always failed schema validation on Oracle so it is impossible to start it. Is that correct?
Long story short, no but it would always reinitialize the cache, regardless of the state (if the startup mode allows).

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
realvictorprm: Ok addressed your request with a484f89. Let's see whether all tests still pass with that, locally I only ran HttpServiceWithPostgresIntTest which passed :)

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
realvictorprm: I wanted the version table to be always created last.  In the scenario of destroying all tables with something going wrong in between, it will always be seen as an invalid state, because the version table has been dropped first (see the oracle queries, there are more tables added which would mess with the order then which also brought me to this). Besides that, with this behavior it's also less likely to end up with version tables state which doesn't contain any version entry.

realvictorprm: Would be good to test what I'm telling here, but right now I don't know how I should simulate app exits in the middle of creating/dropping the tables.

cocreature: My worry is that you’re duplicating this assumption here and in `initDatabase`. If we just put it at the end of `initDatabaseDdls` you don’t have to duplicate this and afaict you get exactly the semantics you want. If you’re worried about someone changing the order in there, maybe factor it out into two variables. Onet he generic `initDatabaseDdls` and then `initDatabaseDdlsAndVersion` (feel free to come up with better names) which just appends the version to `initDatabaseDdls.

realvictorprm: OK I can add an extra val for that, better safe than sorry in the end :)

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
cocreature: > So it would always reinitialize the cache (if allowed by the startup mode).
meaning that with `createSchema=false` which all of our existing customers are using (as it was the only option so far), it would never start?

realvictorprm: Yes

realvictorprm: Side note, it would start if one uses `CreateAndStart`.

victormueller-da:
good stuff, everything is now green :partying_face: So if there aren't any more concerns I'll merge in a couple mins

PR #10483, opened 2021-08-04
Handle tuple values in Daml ledger export

aherrmann-da:
Encodes tuple values using tuple syntax.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/export/src/main/scala/com/daml/script/export/Encode.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/Encode.scala
@@ -142,9 +142,14 @@ private[export] object Encode {
       cidMap: Map[ContractId, String],
       v: Value.Sum,
   ): Doc = {
+    def isTupleRecord(recordId: Identifier): Boolean = {
+      recordId.moduleName == "DA.Types" && recordId.entityName.startsWith("Tuple")
+    }
     def go(v: Value.Sum): Doc =
       v match {
         case Sum.Empty => throw new IllegalArgumentException("Empty value")
+        case Sum.Record(value) if isTupleRecord(value.getRecordId) =>
+          tuple(value.fields.map(f => go(f.getValue.sum)))
         case Sum.Record(value) => encodeRecord(partyMap, cidMap, value)
         // TODO Handle sums of products properly
         case Sum.Variant(value) =>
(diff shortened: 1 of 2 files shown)


daml-script/export/src/main/scala/com/daml/script/export/Encode.scala
```diff
@@ -142,9 +142,14 @@ private[export] object Encode {
       cidMap: Map[ContractId, String],
       v: Value.Sum,
   ): Doc = {
+    def isTupleRecord(recordId: Identifier): Boolean = {
```
cocreature, line 145: given that these are all in a stable package, maybe worth checking the package id as well?

daml-script/export/src/test/scala/com/daml/script/export/EncodeValueSpec.scala
```diff
@@ -50,6 +50,31 @@ class EncodeValueSpec extends AnyFreeSpec with Matchers {
         |    c = 42
         |  c = M.R3""".stripMargin.replace("\r\n", "\n")
     }
+    "tuple" in {
+      def tupleId(n: Int): v.Identifier = v
+        .Identifier()
+        .withPackageId("40f452260bef3f29dede136108fc08a88d5a5250310281067087da6f0baddff7")
```
cocreature, line 56: can we get the package id from daml-lf/language?

cocreature (approved):
Thanks!

daml-script/export/src/main/scala/com/daml/script/export/Encode.scala
aherrmann-da: Done.

daml-script/export/src/test/scala/com/daml/script/export/EncodeValueSpec.scala
aherrmann-da: We could, but the lookup key is `daml-prim-DA-Types-40f452260bef3f29dede136108fc08a88d5a5250310281067087da6f0baddff7`. So, we'd still end up hard-coding the hash.

cocreature: ah fair enough that’s not very helpful

PR #10484, opened 2021-08-04
LF: revisit how contract IDs are handed in commands

remyhaemmerle-da:
Instead of the complicate (and probably buggy) logic in the command preprocessor that tries to detect contract ID freshness conflict statically, we  
- collect input contract IDs during the command preprocessing (without conflict detection) and store those in the Speedy command
- add a builtin to import dynamically those IDs and check freshness conflict at run time
- run the dynamic import before each command
- process a create follow by an exercise on the same cid as one `CreateAndExercise` command, when translation transaction.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -96,7 +96,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
     val submissionTime = cmds.ledgerEffectiveTime
     preprocessor
       .preprocessCommands(cmds.commands)
-      .flatMap { case (processedCmds, globalCids) =>
+      .flatMap { case processedCmds =>
         interpretCommands(
           validating = false,
           submitters = submitters,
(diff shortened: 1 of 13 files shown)


cocreature (changes requested):
This doesn’t work correctly unless you also change interpretation to go node by node. Consider this example https://gist.github.com/cocreature/724567664924cd36f58c86f3490fcd15 in LF 1.13.

The createAndExercise creates a `Keyed` contract which is now in the local cids.
The `exerciseByKey` then is turned into a regular exercise due to the lack of byKey flag and blows up.
It works correctly in LF 1.14.

andreaslochbihler-da (commented):
A few high-level comments as I'm not deeply familiar with the subtleties of Daml engine:
* The new built-in `SBUImportCids` makes a lot of sense and things do look cleaner. :+1: 
* Conceptually, I found it more intuitive to add all contract IDs in the commands before we (re)interpret the Speedy commands. This is now intertwined and seems to require a few contortions like a special case for dealing with createAndExercise, but that's probably a matter of taste.

remyhaemmerle-da:
As described in #10504 the contract ID discriminators in conceptually broken.

PR #10487, opened 2021-08-04
Replace LedgerConfiguration with InitialLedgerConfiguration or the load timeout. [KVL-1058]

SamirTalwar-DA:
Most things don't need all its constituent parts; this reduces the amount of unused properties.

In a future changeset, the initial ledger configuration will become optional. If it is not provided, the participant will not attempt to write a configuration.

### Changelog

- **[Integration Kit]** The `LedgerConfiguration` class has been removed in favor of `InitialLedgerConfiguration`. Its usage has been changed accordingly, with the `configurationLoadTimeout` property becoming part of `ApiServerConfig` instead.

  The default options provided by `LedgerConfiguration` have been removed; you are now encouraged to come up with sensible values for your own ledger. The `Configuration.reasonableInitialConfiguration` value may help.

### Pull Request Checklist

trigger the build.


ledger/ledger-on-memory/src/app/scala/com/daml/ledger/on/memory/InMemoryLedgerFactory.scala
```diff
@@ -50,8 +52,8 @@ private[memory] class InMemoryLedgerFactory(dispatcher: Dispatcher[Index], state
     )
   }
 
-  override def ledgerConfig(config: Config[Unit]): LedgerConfiguration =
-    LedgerConfiguration.defaultLocalLedger
+  override def initialLedgerConfig(config: Config[Unit]): InitialLedgerConfiguration =
+    super.initialLedgerConfig(config).copy(delayBeforeSubmitting = Duration.ZERO)
```
rautenrieth-da, line 56: A few ledgers now have different values for the delays. Was this intended?

The config submission delay used to be 500ms for kvutils ledgers (or any other single-node ledger where the index database is not the same as the ledger database). The previous comment says this is because there is a slight delay between the ledger and the index, although I don't think this would affect the initial submission for ledger-on-memory.

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
```diff
@@ -18,16 +19,19 @@ import com.daml.ledger.participant.state.kvutils.caching._
...
   override val defaultExtraConfig: ExtraConfig = ExtraConfig(
     jdbcUrl = None
   )
 
-  override def ledgerConfig(config: Config[ExtraConfig]): LedgerConfiguration =
-    super.ledgerConfig(config).copy(initialConfigurationSubmitDelay = Duration.ZERO)
+  override def initialLedgerConfig(config: Config[ExtraConfig]): InitialLedgerConfiguration =
+    InitialLedgerConfiguration(
+      configuration = Configuration.reasonableInitialConfiguration,
+      delayBeforeSubmitting = Duration.ZERO,
```
rautenrieth-da, line 33: Doesn't a delay of zero mean the participant will submit a new config change request every time it restarts? Even with an up-to-date index database, it will take some time until `CurrentLedgerConfiguration` sees a configuration (because of the round trip time to the database).

ledger/ledger-on-memory/src/app/scala/com/daml/ledger/on/memory/InMemoryLedgerFactory.scala
SamirTalwar-DA: Whoops, you're right, this should be 500ms.

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
SamirTalwar-DA: Yup, you are absolutely right. I screwed this up thinking this was Daml-on-SQL code, not ledger-on-sql.

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
```diff
@@ -21,13 +21,23 @@ final case class Configuration(
...
+  val NoGeneration: Long = 0
+  val StartingGeneration: Long = 1
 
   /** Version history:
     * V1: initial version
     * V2: added maxDeduplicationTime
     */
   val protobufVersion: Long = 2L
 
+  val reasonableMaxDeduplicationTime: Duration = Duration.ofDays(1)
```
nicu-da, line 33: Can we add a comment as to why is 1 day a reasonable deduplication time?

ledger/ledger-on-memory/src/app/scala/com/daml/ledger/on/memory/InMemoryLedgerFactory.scala
```diff
@@ -50,8 +52,8 @@ private[memory] class InMemoryLedgerFactory(dispatcher: Dispatcher[Index], state
     )
   }
 
-  override def ledgerConfig(config: Config[Unit]): LedgerConfiguration =
-    LedgerConfiguration.defaultLocalLedger
+  override def initialLedgerConfig(config: Config[Unit]): InitialLedgerConfiguration =
+    super.initialLedgerConfig(config).copy(delayBeforeSubmitting = Duration.ofMillis(500))
```
nicu-da, line 56: can we extract the duration into a variable and add a comment as to why we've chosen this duration

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
```diff
@@ -18,16 +18,16 @@ import com.daml.ledger.participant.state.kvutils.caching._
...
 
 object SqlLedgerFactory extends LedgerFactory[ReadWriteService, ExtraConfig] {
   override val defaultExtraConfig: ExtraConfig = ExtraConfig(
     jdbcUrl = None
   )
 
-  override def ledgerConfig(config: Config[ExtraConfig]): LedgerConfiguration =
-    super.ledgerConfig(config).copy(initialConfigurationSubmitDelay = Duration.ZERO)
+  override def initialLedgerConfig(config: Config[ExtraConfig]): InitialLedgerConfiguration =
+    super.initialLedgerConfig(config).copy(delayBeforeSubmitting = Duration.ofMillis(500))
```
nicu-da, line 30: Same here. Maybe we can extract to a constant and reuse if it has the same meaning

ledger/ledger-on-memory/src/app/scala/com/daml/ledger/on/memory/InMemoryLedgerFactory.scala
SamirTalwar-DA: Yeah, let's do it. I'd like to keep it KV-specific but I guess that works.

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
rautenrieth-da: The original value was chosen more or less randomly, because we thought one day is long enough for applications to recover and retry failed commands. AFAIK some ledgers will want to set this to one week though.

Participants may have to keep command deduplication data for up to this period of time. The only reason for having an upper bound on this duration is to avoid unbounded storage requirements.

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
rautenrieth-da: We had these constants defined in a common place ([here](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/configuration/LedgerConfiguration.scala#L46)). Thinking about it again, there should be 3 different common scenarios:

1. Ledger without persistence. The submission delay should be zero, as each application start will create a new ledger.
2. Participant where the index database is always up to date. The submission delay should be very short, as we only need to wait one roundtrip to the database to see whether there is a configuration.
3. Participant where the index database may have to be recreated from a distributed ledger. The submission delay should be a bit longer, to allow the indexer to ingest a few updates from the ReadService.

If the delay is too short, each participant start may produce a "Configuration change rejected" update on the ledger, which should not affect anything. Maybe we could review those constants and add them to `object InitialLedgerConfiguration`?

ledger/participant-integration-api/src/main/scala/platform/configuration/InitialLedgerConfiguration.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.daml.ledger.configuration.Configuration
+
+/** Instructions on how to initialize an empty ledger, without a configuration.
+  *
+  * A configuation is only submitted if one is not detected on the ledger.
+  *
+  * @param configuration         The initial ledger configuration to submit.
+  * @param delayBeforeSubmitting The delay until the participant tries to submit a configuration.
+  */
+final case class InitialLedgerConfiguration(
```
fabiotudone-da, line 17: I think I am confused by the multiple meanings of "configuration", as it is both "the result of configuring" and "the action of configuring", isn't it? Not sure what I can suggest though, to clarify that this is information about how to do it; `LedgerSetupInfo` or something like that?

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
```diff
@@ -21,13 +21,26 @@ final case class Configuration(
 )
 
 object Configuration {
+  val NoGeneration: Long = 0
+  val StartingGeneration: Long = 1
```
fabiotudone-da, line 25: Do we need them public?

ledger/ledger-on-memory/src/app/scala/com/daml/ledger/on/memory/InMemoryLedgerFactory.scala
SamirTalwar-DA: Done, but I kept it in the KV `Config` object, so other ledgers will have to make their own decisions. I think this is safer than just having one default for all ledgers.

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
SamirTalwar-DA: Done.

SamirTalwar-DA: GitHub is not refreshing comments for me until I hit Refresh. That's fun.

I'd be happy for driver authors to make this bigger; the important thing, as @rautenrieth-da says, is that it's bounded. However, I do think 1 day is totally reasonable for most set-ups.

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
SamirTalwar-DA: Yes, they're not used here, only elsewhere. Not much happens inside this package.

nicu-da (approved):

ledger/ledger-on-sql/src/app/scala/com/daml/ledger/on/sql/SqlLedgerFactory.scala
SamirTalwar-DA: I think the constants we had in `LedgerConfiguration` were quite misleading; ledgers can vary quite a bit, and we can't come up with decent values for all of them. I have extracted a couple of useful constants specifically for KV and added them to the KV `Config` object, but I don't think these should necessarily extend outwards to other driver implementations.

For example, I would expect an implementation based on Ethereum to be working in minutes, not seconds, and it makes very little sense for a Canton-based driver to allow the participant to set a configuration at all. (In the near future, this behavior will be completely optional, and Canton will not provide an `InitialLedgerConfiguration`.)

ledger/participant-integration-api/src/main/scala/platform/configuration/InitialLedgerConfiguration.scala
SamirTalwar-DA: I'm not sure `Info` helps.

I have tried to restrict the word "configuration" to only data objects representing the ledger configuration, whether the current configuration or a potential future one. The exception here is `CurrentLedgerConfiguration`, which I've renamed in some follow-up changes.

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
fabiotudone-da: I was thinking about package restriction; if they need to be fully public, maybe it's worth making them a bit more opaque?

fabiotudone-da (approved):

rautenrieth-da (approved):

ledger/ledger-configuration/src/main/scala/com/daml/ledger/configuration/Configuration.scala
SamirTalwar-DA: Other ledger operators may need to access `StartingGeneration` (for example, if they don't want to use `reasonableInitialConfiguration`). I don't see any particular harm in keeping that one public.

I just realized `NoGeneration` is not actually used except in a test case now, so I think I'll just delete that.

PR #10468, opened 2021-08-03
[JSON-API/Docs] Fix the date bounds specification in the docs to be 9999-12-31

realvictorprm:
I think this should fix #10449

changelog_begin

- [JSON-API] The date bounds specification of the json api has been corrected, the upper limit is now 9999-12-31 instead of the invalid date 9999-99-99

changelog_end

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10489, opened 2021-08-05
correct JSON API upper date bound

### LINKED ISSUE #10449: Data Bounds Incorrectly Designated (9999-99-99)
In the DA_SDK.pdf, page 259/577:

#### 3.3.1.7 Date

> The dates must be between the bounds specified by Daml-LF and ISO 8601, [0001-01-01, 9999-99-99].

Shouldn't this be:

> The dates must be between the bounds specified by Daml-LF and ISO 8601, [0001-01-01, 9999-12-31].

Or is this only a Numerical format, and that the implied use would correctly be 9999-12-31?

garyverhaegen-da:
As reported by @quid-agis. Fixes #10449.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature:
this is a duplicate of #10468

realvictorprm:
sorry that I did not take care of the additional test yet. Did not spend time on looking into the spec tests further.

garyverhaegen-da:
> this is a duplicate of #10468

But that's clearly dead: it's not moved in all of two days. Also this has tests now. 😇

(I blame GitHub for not refreshing #10449 so I could see the reference.)

ledger-service/lf-value-json/src/main/scala/com/digitalasset/daml/lf/value/json/ApiCodecCompressed.scala
```diff
@@ -136,7 +136,13 @@ class ApiCodecCompressed[Cid](val encodeDecimalAsString: Boolean, val encodeInt6
       case Model.DamlLfPrimType.Timestamp => { case JsString(v) =>
         V.ValueTimestamp(assertDE(Time.Timestamp fromString v))
       }
-      case Model.DamlLfPrimType.Date => { case JsString(v) => V.ValueDate.fromIso8601(v) }
+      case Model.DamlLfPrimType.Date => { case JsString(v) =>
+        try {
+          V.ValueDate.fromIso8601(v)
+        } catch {
+          case _: Exception => throw DeserializationException(s"Invalid date: $v")
```
garyverhaegen-da, line 143: The table of failures in `ApiCodeCompressedSpec` expects `DeserializationException` specifically. I wasn't sure which side should be changed.

realvictorprm:
> > this is a duplicate of #10468
> 
> But that's clearly dead: it's not moved in all of two days. Also this has tests now. 
> 
> (I blame GitHub for not refreshing #10449 so I could see the reference.)

Ignoring the fact that often enough PR's are left for one or two days :stuck_out_tongue: 

Jokes aside, thanks for doing this :cowboy_hat_face:

realvictorprm (approved):
Looks good to me :) Many thanks for taking over :+1:

ledger-service/lf-value-json/src/main/scala/com/digitalasset/daml/lf/value/json/ApiCodecCompressed.scala
cocreature: Changing here seems fine but I’d prefer to catch `java.time.format.DateTimeParseException` and `IllegalArgumentException` instead of catching `Exception`.

ledger-service/lf-value-json/src/main/scala/com/digitalasset/daml/lf/value/json/ApiCodecCompressed.scala
```diff
@@ -136,7 +136,13 @@ class ApiCodecCompressed[Cid](val encodeDecimalAsString: Boolean, val encodeInt6
       case Model.DamlLfPrimType.Timestamp => { case JsString(v) =>
         V.ValueTimestamp(assertDE(Time.Timestamp fromString v))
       }
-      case Model.DamlLfPrimType.Date => { case JsString(v) => V.ValueDate.fromIso8601(v) }
+      case Model.DamlLfPrimType.Date => { case JsString(v) =>
+        try {
+          V.ValueDate.fromIso8601(v)
```
cocreature, line 141: probably not in scope for this PR but we might want to consider having `fromIso8601`return an `Either` and having an `assertFromIso8601` that throws like we do in lots of other places.

cocreature (approved):
Thanks!

realvictorprm:
@garyverhaegen-da I can take over the requested changes if you want, just ping me

ledger-service/lf-value-json/src/main/scala/com/digitalasset/daml/lf/value/json/ApiCodecCompressed.scala
garyverhaegen-da: I slightly disagree but since you're not here to argue today I'll go with your approach.

ledger-service/lf-value-json/src/main/scala/com/digitalasset/daml/lf/value/json/ApiCodecCompressed.scala
garyverhaegen-da: @realvictorprm Since you offered, do you want to take care of this part?

realvictorprm: Will do :)

garyverhaegen-da: Thanks!

PR #10490, opened 2021-08-05
Handle exerciseByKey after exercise command

aherrmann-da:
The discovery of exerciseByKey commands currently doesn't handle the case where the contract or contract key is modified during the same transaction. Issuing a regular exercise command in this case is an error as it is not possible to reference a contract id that is generated during the same transaction.

This PR fixes this by tracking the impact of exercise events on contract keys.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
@@ -237,16 +237,31 @@ object TreeUtils {
   object Command {
     def fromTree(tree: TransactionTree): Seq[Command] = {
       val contractKeys = mutable.HashMap.empty[ContractId, Value]
+      def addContractKey(createdEvent: CreatedEvent): Unit = {
+        createdEvent.contractKey.foreach { contractKey =>
+          contractKeys += ContractId(createdEvent.contractId) -> contractKey
+        }
+      }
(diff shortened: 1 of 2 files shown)


daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
```diff
@@ -237,16 +237,31 @@ object TreeUtils {
   object Command {
     def fromTree(tree: TransactionTree): Seq[Command] = {
       val contractKeys = mutable.HashMap.empty[ContractId, Value]
+      def addContractKey(createdEvent: CreatedEvent): Unit = {
```
cocreature, line 240: I think I forgot how we handle ExerciseByKey. Can you confirm that my understanding upon rereading the code is correct:
We turn each exercise into an exerciseByKey provided that the contract has been created (and has a key) in the current transaction and has not been archived since.

cocreature (approved):
Thanks! One question on whether my understanding is correct but looks very sensible

daml-script/export/src/main/scala/com/daml/script/export/TreeUtils.scala
aherrmann-da: Correct, with one exception: If the exercise directly follows the create that created the contract, then we issue a createAndExercise instead of create followed by exerciseByKey.

PR #10498, opened 2021-08-06
participant-integration-api: Make the initial ledger config optional. [KVL-1046]

SamirTalwar-DA:
### Changelog

- **[Integration Kit]** The initial ledger configuration is now optional; if it is not specified, the participant will not attempt to submit a configuration, but instead wait for something else to provision the ledger with a configuration.

  This behavior is intended to be used by ledger drivers that have alternative means of setting up the initial ledger configuration. If no configuration is provisioned, the participant server will not be able to submit commands to the ledger.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServerConfig.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/ApiServerConfig.scala
@@ -25,7 +25,7 @@ case class ApiServerConfig(
     databaseConnectionTimeout: FiniteDuration,
     tlsConfig: Option[TlsConfiguration],
     maxInboundMessageSize: Int,
-    initialLedgerConfiguration: InitialLedgerConfiguration,
+    initialLedgerConfiguration: Option[InitialLedgerConfiguration],
     configurationLoadTimeout: Duration,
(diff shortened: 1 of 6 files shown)


nicu-da (approved):

fabiotudone-da (commented):
Does it makes sense to add a test for the new "don't inizialize a configuration if not asked to" behavior?

SamirTalwar-DA:
@fabiotudone-da: Yeah, you're right. I think `LedgerConfigInitializer` needs some tests, which is a shame because it's kind of heavyweight. I'll add some as part of this PR.

SamirTalwar-DA:
OK, looked into this and I think while it's possible to write tests for this, they'll have to be integrated tests and quite unpleasant. Perhaps some further refactoring could make it more testable but honestly, I'm not sure it's worth it.

fabiotudone-da (approved):

fabiotudone-da:
> OK, looked into this and I think while it's possible to write tests for this, they'll have to be integrated tests and quite unpleasant. Perhaps some further refactoring could make it more testable but honestly, I'm not sure it's worth it.

Then it's fine IMO, perhaps let's take a note for the future to refactor it further so that it's more testable.

PR #10500, opened 2021-08-06
participant-integration-api: Always wait for the first config lookup.

SamirTalwar-DA:
This means we don't have to guess how long a `lookupConfiguration()` round trip will take. Instead we can just submit the initial configuration if one is not found.

Some drivers may still need to wait a while for other reasons.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
@@ -24,7 +24,14 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
   * configuration available to consumers.
   *
   * This class helps avoiding code duplication and limiting the number of database lookups, as
-  * multiple services and validators require the latest ledger config.
+  * multiple services and validators require the latest ledger configuration.
+  *
+  * The [[subscription]] method returns a [[ResourceOwner]] which succeeds after the first lookup,
+  * regardless of whether the index contains a configuration or not. It then starts a background
+  * stream of configuration updates to keep its internal configuration state in sync with the index.
+  *
+  * The [[Subscription.ready]] method returns a [[Future]] that will not resolve until a
+  * configuration is found or the load timeout expires.
   */
 private[apiserver] final class LedgerConfigurationSubscriptionFromIndex(
     indexService: IndexConfigManagementService,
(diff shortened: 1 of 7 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
```diff
@@ -44,142 +51,97 @@ private[apiserver] final class LedgerConfigurationSubscriptionFromIndex(
...
+          .map {
+            case (offset, domain.ConfigurationEntry.Accepted(_, config)) =>
+              logger.info(s"New ledger configuration $config found at $offset")
+              scheduledTimeout.cancel()
+              latestConfigurationState.set(Some(offset) -> Some(config))
+              readyPromise.trySuccess(())
+              ()
+            case (offset, domain.ConfigurationEntry.Rejected(_, _, _)) =>
+              logger.info(s"New ledger configuration rejection found at $offset")
+              latestConfigurationState.updateAndGet(previous => Some(offset) -> previous._2)
```
nicu-da, line 123: any reason for the `updateAndGet` instead of `set` as used previosly?

SamirTalwar-DA: Yes; if we get a rejection but we previously had a configuration, we need to keep the previous configuration instead of setting it back to `None`.

(This code was just moved.)

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
```diff
@@ -24,7 +24,14 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
...
   * This class helps avoiding code duplication and limiting the number of database lookups, as
-  * multiple services and validators require the latest ledger config.
+  * multiple services and validators require the latest ledger configuration.
+  *
+  * The [[subscription]] method returns a [[ResourceOwner]] which succeeds after the first lookup,
+  * regardless of whether the index contains a configuration or not. It then starts a background
+  * stream of configuration updates to keep its internal configuration state in sync with the index.
+  *
+  * The [[Subscription.ready]] method returns a [[Future]] that will not resolve until a
+  * configuration is found or the load timeout expires.
```
fabiotudone-da, line 34: 👍🏼

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
fabiotudone-da: I think it's because `set` doesn't allow you to access the previous value. I'd propose to use a proper case class for the state though.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
```diff
@@ -44,142 +51,97 @@ private[apiserver] final class LedgerConfigurationSubscriptionFromIndex(
       override def acquire()(implicit
           context: ResourceContext
       ): Resource[LedgerConfigurationSubscription with IsReady] =
-        Resource(Future {
-          new Subscription(configurationLoadTimeout)
-        }(context.executionContext))(subscription =>
+        Resource(
+          indexService
+            .lookupConfiguration()
```
fabiotudone-da, line 56: What happens if this fails?

SamirTalwar-DA: The resource fails and the participant will crash.

This might sound scary, but this is already the case if the index fails to look up the ledger ID, so it doesn't add any new failure modes. Basically, participant without index is bad news.

fabiotudone-da: Worth logging something to clarify or will the cause of failure be clear enough?

rautenrieth-da: It's a good point. We generally don't retry database queries or otherwise handle failures on a low level.

For the indexer, we have `RecoveringIndexer` that restarts the whole indexer if anything fails (even a small thing like a transient database connection issue). This is not optimal, but at least the indexer component automatically and recovers from transient errors.

For the ledger API server, (AFAIK) we don't have such a thing. It would be the responsibility of the participant operator to monitor the health of the participant and restart if necessary.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
SamirTalwar-DA: I'll do it in a new changeset.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
SamirTalwar-DA: If only we could stop packaging the API server and indexer together, that would be trivial. :upside_down_face:

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
```diff
@@ -95,7 +96,7 @@ trait ConfigProvider[ExtraConfig] {
   def initialLedgerConfig(config: Config[ExtraConfig]): InitialLedgerConfiguration =
     InitialLedgerConfiguration(
       configuration = Configuration.reasonableInitialConfiguration,
-      delayBeforeSubmitting = Config.DefaultInitialConfigurationSubmissionDelay,
+      delayBeforeSubmitting = Duration.ZERO,
```
rautenrieth-da, line 99:
```suggestion
      // If a new index database is added to an already existing ledger,
      // a zero delay will likely produce a "configuration rejected" ledger entry,
      // because at startup the indexer hasn't ingested any configuration change yet.
      // Override this setting for distributed ledgers where you want to avoid these superfluous entries.
      delayBeforeSubmitting = Duration.ZERO,
```

rautenrieth-da (approved):
The change in the logic looks good :+1: I agree that a case class for the state would make the code nicer.

SamirTalwar-DA:
I have turned `(Option[LedgerOffset.Absolute], Option[Configuration])` into a `State` enumeration, but I'll push it separately.

ledger/participant-integration-api/src/main/scala/platform/apiserver/configuration/LedgerConfigurationSubscriptionFromIndex.scala
SamirTalwar-DA: #10508 introduces a `State` enumeration.

PR #10503, opened 2021-08-06
Track command - use types for error handling instead of grpc statuses [KVL-1005]

nicu-da:
Use an Either as a response from the command tracking client. We convert the `Completion` into a typed response as soon as possible and no longer count on grpc `Status` to determine if it was an error or not later on in the code.

No changes as to the way we actually process the response.

- We have the ability to add metadata to the `ErrorInfo` returned as part of the grpc exception  but this is not used just yet. (Will be covered in another PR)
- The failure from adding to the command queue is still directly propagated as an exception but will update that in a new PR so that we're consistent in the way we handle errors in this flow

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
+++ b/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
@@ -9,13 +9,13 @@ import akka.NotUsed
 import akka.stream.FlowShape
 import akka.stream.scaladsl.{Flow, GraphDSL, Merge, Partition}
 import com.daml.api.util.TimeProvider
+import com.daml.ledger.client.services.commands.tracker.CompletionResponse.CompletionResponse
 import com.daml.ledger.api.refinements.ApiTypes.Party
 import com.daml.ledger.api.v1.command_submission_service.SubmitRequest
-import com.daml.ledger.api.v1.completion.Completion
 import com.daml.ledger.client.services.commands.CommandClient
+import com.daml.ledger.client.services.commands.tracker.CompletionResponse
 import com.daml.util.Ctx
 import com.google.rpc.Code
-import com.google.rpc.status.Status
 import scalaz.syntax.tag._
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -23,9 +23,9 @@ import scala.concurrent.{ExecutionContext, Future}
 object CommandRetryFlow {
 
   type In[C] = Ctx[C, SubmitRequest]
-  type Out[C] = Ctx[C, Completion]
+  type Out[C] = Ctx[C, CompletionResponse]
   type SubmissionFlowType[C] = Flow[In[C], Out[C], NotUsed]
-  type CreateRetryFn[C] = (RetryInfo[C], Completion) => SubmitRequest
+  type CreateRetryFn[C] = (RetryInfo[C], CompletionResponse) => SubmitRequest
 
   private val RETRY_PORT = 0
   private val PROPAGATE_PORT = 1
@@ -70,25 +70,38 @@ object CommandRetryFlow {
             {
               case Ctx(
                     RetryInfo(request, nrOfRetries, firstSubmissionTime, _),
-                    Completion(_, Some(status: Status), _),
+                    result,
                     _,
                   ) =>
-                if (status.code == Code.OK_VALUE) {
-                  PROPAGATE_PORT
-                } else if (
-                  (firstSubmissionTime plus maxRetryTime) isBefore timeProvider.getCurrentTime
-                ) {
-                  RetryLogger.logStopRetrying(request, status, nrOfRetries, firstSubmissionTime)
-                  PROPAGATE_PORT
-                } else if (RETRYABLE_ERROR_CODES.contains(status.code)) {
-                  RetryLogger.logNonFatal(request, status, nrOfRetries)
-                  RETRY_PORT
-                } else {
-                  RetryLogger.logFatal(request, status, nrOfRetries)
-                  PROPAGATE_PORT
+                result match {
+                  case Left(value) =>
+                    value match {
+                      case CompletionResponse.NotOkResponse(_, status) =>
+                        if (
+                          (firstSubmissionTime plus maxRetryTime) isBefore timeProvider.getCurrentTime
+                        ) {
+                          RetryLogger.logStopRetrying(
+                            request,
+                            status,
+                            nrOfRetries,
+                            firstSubmissionTime,
+                          )
+                          PROPAGATE_PORT
+                        } else if (RETRYABLE_ERROR_CODES.contains(status.code)) {
+                          RetryLogger.logNonFatal(request, status, nrOfRetries)
+                          RETRY_PORT
+                        } else {
+                          RetryLogger.logFatal(request, status, nrOfRetries)
+                          PROPAGATE_PORT
+                        }
+                      case CompletionResponse.TimeoutResponse(_) =>
+                        PROPAGATE_PORT
+                      case CompletionResponse.NoStatusInResponse(commandId) =>
+                        statusNotFoundError(commandId)
+                    }
+                  case Right(_) =>
+                    PROPAGATE_PORT
                 }
-              case Ctx(_, Completion(commandId, _, _), _) =>
-                statusNotFoundError(commandId)
             },
           )
         )
--- a/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
+++ b/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
@@ -9,11 +9,14 @@ import akka.NotUsed
 import akka.stream.FlowShape
 import akka.stream.scaladsl.{Flow, GraphDSL, Merge, Partition}
 import com.daml.api.util.TimeProvider
-import com.daml.ledger.client.services.commands.tracker.CompletionResponse.CompletionResponse
 import com.daml.ledger.api.refinements.ApiTypes.Party
 import com.daml.ledger.api.v1.command_submission_service.SubmitRequest
 import com.daml.ledger.client.services.commands.CommandClient
 import com.daml.ledger.client.services.commands.tracker.CompletionResponse
+import com.daml.ledger.client.services.commands.tracker.CompletionResponse.{
+  CompletionFailure,
+  CompletionSuccess,
+}
 import com.daml.util.Ctx
 import com.google.rpc.Code
 import scalaz.syntax.tag._
@@ -23,9 +26,10 @@ import scala.concurrent.{ExecutionContext, Future}
 object CommandRetryFlow {
 
   type In[C] = Ctx[C, SubmitRequest]
-  type Out[C] = Ctx[C, CompletionResponse]
+  type Out[C] = Ctx[C, Either[CompletionFailure, CompletionSuccess]]
   type SubmissionFlowType[C] = Flow[In[C], Out[C], NotUsed]
-  type CreateRetryFn[C] = (RetryInfo[C], CompletionResponse) => SubmitRequest
+  type CreateRetryFn[C] =
+    (RetryInfo[C], Either[CompletionFailure, CompletionSuccess]) => SubmitRequest
 
   private val RETRY_PORT = 0
   private val PROPAGATE_PORT = 1
--- a/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
+++ b/language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
@@ -70,39 +70,34 @@ object CommandRetryFlow {
             {
               case Ctx(
                     RetryInfo(request, nrOfRetries, firstSubmissionTime, _),
-                    result,
+                    Left(CompletionResponse.NotOkResponse(_, status)),
+                    _,
+                  ) if RETRYABLE_ERROR_CODES.contains(status.code) =>
+                if ((firstSubmissionTime plus maxRetryTime) isBefore timeProvider.getCurrentTime) {
+                  RetryLogger.logStopRetrying(
+                    request,
+                    status,
+                    nrOfRetries,
+                    firstSubmissionTime,
+                  )
+                  PROPAGATE_PORT
+                } else {
+                  RetryLogger.logNonFatal(request, status, nrOfRetries)
+                  RETRY_PORT
+                }
+              case Ctx(
+                    RetryInfo(request, nrOfRetries, _, _),
+                    Left(CompletionResponse.NotOkResponse(_, status)),
                     _,
                   ) =>
-                result match {
-                  case Left(completionFailure) =>
-                    completionFailure match {
-                      case CompletionResponse.NotOkResponse(_, status)
-                          if RETRYABLE_ERROR_CODES.contains(status.code) =>
-                        if (
-                          (firstSubmissionTime plus maxRetryTime) isBefore timeProvider.getCurrentTime
-                        ) {
-                          RetryLogger.logStopRetrying(
-                            request,
-                            status,
-                            nrOfRetries,
-                            firstSubmissionTime,
-                          )
-                          PROPAGATE_PORT
-                        } else {
-                          RetryLogger.logNonFatal(request, status, nrOfRetries)
-                          RETRY_PORT
-                        }
-                      case CompletionResponse.NotOkResponse(_, status) =>
-                        RetryLogger.logFatal(request, status, nrOfRetries)
-                        PROPAGATE_PORT
-                      case CompletionResponse.TimeoutResponse(_) =>
-                        PROPAGATE_PORT
-                      case CompletionResponse.NoStatusInResponse(commandId) =>
-                        statusNotFoundError(commandId)
-                    }
-                  case Right(_) =>
-                    PROPAGATE_PORT
-                }
+                RetryLogger.logFatal(request, status, nrOfRetries)
+                PROPAGATE_PORT
+              case Ctx(_, Left(CompletionResponse.TimeoutResponse(_)), _) =>
+                PROPAGATE_PORT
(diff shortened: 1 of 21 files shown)


ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/CommandClientIT.scala
```diff
@@ -144,17 +148,19 @@ final class CommandClientIT
...
       expectedMessageSubString: String,
   ): Future[Assertion] =
-    submitCommand(submitRequest).map { completion =>
-      completion.getStatus.code should be(expectedErrorCode.value)
-      completion.getStatus.message should include(expectedMessageSubString)
+    submitCommand(submitRequest).map { result =>
+      inside(result) { case Left(NotOkResponse(_, grpcStatus)) =>
+        grpcStatus.code should be(expectedErrorCode.value)
+        grpcStatus.message should include(expectedMessageSubString)
+      }
```
fabiotudone-da, line 163: Probably just my ignorance about `Inside` but this match doesn't seem exhaustive and, if so, what (failure, hopefully) message do we get in other cases?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -0,0 +1,65 @@
...
+              transactionId = completion.transactionId,
+            )
+          )
+        } else {
+          Left(NotOkResponse(completion.commandId, grpcStatus))
+        }
+      case None =>
+        Left(NoStatusInResponse(completion.commandId))
+    }
+  }
```
fabiotudone-da, line 43: I think you don't need these curly braces

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerImpl.scala
```diff
@@ -95,24 +93,10 @@ private[services] object TrackerImpl {
...
-                GrpcStatus.INTERNAL
-                  .withDescription("Missing status in completion response.")
-              )
-
-            logger.trace(s"Completing promise with failure: $status")
-            promise.tryFailure(status.asException())
+        logger.trace("Completing promise")
+        val didCompletePromise = promise.trySuccess(result)
+        if (!didCompletePromise) {
+          logger.trace("Failed to complete promise")
```
fabiotudone-da, line 99: I'd perhaps change it to:

- Log only afterwards, both in the `true` and `false` cases
- Expand a bit the log message to give more context about what's happening

fabiotudone-da (approved):
Only minor comments, generally LGTM; I am not too familiar with this code so, although it looks to me like no user-facing changes are introduced, it's perhaps worth double-checking that this is indeed the case.

ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/CommandClientIT.scala
nicu-da: Yes. So in case of not matching we will get a failure.
I based the example on scalatest docs:

https://www.scalatest.org/user_guide/using_inside

Checking the source code we get a failure when the match is partial:

```
   if (pf.isDefinedAt(value)) {
     ......
    }
    else {
      Inside.level.set(Inside.level.get - 1)
      throw new TestFailedException((_: StackDepthException) => Some(Resources.insidePartialFunctionNotDefined(value.toString())), None, pos)
    }
```

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
```diff
@@ -70,25 +70,38 @@ object CommandRetryFlow {
...
-                ) {
-                  RetryLogger.logStopRetrying(request, status, nrOfRetries, firstSubmissionTime)
-                  PROPAGATE_PORT
-                } else if (RETRYABLE_ERROR_CODES.contains(status.code)) {
-                  RetryLogger.logNonFatal(request, status, nrOfRetries)
-                  RETRY_PORT
-                } else {
-                  RetryLogger.logFatal(request, status, nrOfRetries)
-                  PROPAGATE_PORT
+                result match {
```
SamirTalwar-DA, line 76: I think you can merge the `match` into the `case` above.

language-support/scala/bindings-akka/src/test/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlowUT.scala
```diff
@@ -32,13 +38,15 @@ class CommandRetryFlowUT extends AsyncWordSpec with Matchers with AkkaTest {
...
           if (commands.deduplicationTime.get.nanos == 0) {
-            Ctx(context, Completion(commands.commandId, Some(status)))
+            Ctx(context, CompletionResponse(Completion(commands.commandId, Some(status))))
           } else {
             Ctx(
               context,
-              Completion(
-                commands.commandId,
-                Some(status.copy(code = Code.OK_VALUE)),
+              CompletionResponse(
```
SamirTalwar-DA, line 45: Why not just use `CompletionResponse.CompletionSuccess` here?

language-support/scala/bindings-akka/src/test/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlowUT.scala
```diff
@@ -32,13 +38,15 @@ class CommandRetryFlowUT extends AsyncWordSpec with Matchers with AkkaTest {
       .map {
         case Ctx(context @ RetryInfo(_, _, _, status), SubmitRequest(Some(commands)), _) =>
           if (commands.deduplicationTime.get.nanos == 0) {
-            Ctx(context, Completion(commands.commandId, Some(status)))
+            Ctx(context, CompletionResponse(Completion(commands.commandId, Some(status))))
```
SamirTalwar-DA, line 41: How does this even work? I don't see a `CompletionResponse.unapply` method.

ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/CommandClientIT.scala
fabiotudone-da: Alright, it stringifies the whole value, which should be fine, thanks for checking and sorry for the laziness 👍🏼

language-support/scala/bindings-akka/src/test/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlowUT.scala
```diff
@@ -91,7 +99,7 @@ class CommandRetryFlowUT extends AsyncWordSpec with Matchers with AkkaTest {
       submitRequest(Code.OK_VALUE, Instant.ofEpochSecond(45)) map { result =>
         result.size shouldBe 1
         result.head.context.nrOfRetries shouldBe 0
-        result.head.value.status.get.code shouldBe Code.OK_VALUE
+        result.head.value shouldBe Symbol("right")
```
SamirTalwar-DA, line 102: Please use `matchPattern` rather than symbols. Much more type-safe.

```suggestion
        result.head.value should matchPattern { Right(_) => }
```

You could potentially use `inside(result)` in this test and subsequent ones to make it terser and get a better error message on failure (if, e.g. there are 2 results):

```scala
        inside(result) { case Seq(Ctx(context, Right(_), _) =>
          context.nrOfRetries shouldBe 0
        }
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
```diff
@@ -95,7 +95,10 @@ final class CommandClient(
   def trackCommands[Context](parties: Seq[String], token: Option[String] = None)(implicit
       ec: ExecutionContext
   ): Future[
-    Flow[Ctx[Context, SubmitRequest], Ctx[Context, Completion], Materialized[NotUsed, Context]]
+    Flow[Ctx[Context, SubmitRequest], Ctx[Context, CompletionResponse], Materialized[
+      NotUsed,
+      Context,
+    ]]
```
SamirTalwar-DA, line 101: Can you create a type alias for this and the below `Flow`s?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -0,0 +1,65 @@
...
+  final case class NotOkResponse(commandId: String, grpcStatus: StatusProto)
+      extends CompletionFailure
+  final case class TimeoutResponse(commandId: String) extends CompletionFailure
+  final case class NoStatusInResponse(commandId: String) extends CompletionFailure
+
+  final case class CompletionSuccess(commandId: String, transactionId: String)
+
+  def apply(completion: Completion): Either[CompletionFailure, CompletionSuccess] = {
+    completion.status match {
+      case Some(grpcStatus) =>
```
SamirTalwar-DA, line 29: If you like, you can use `case Some(GrpcStatus.OK())` here, and `case Some(_)` for the `else` branch.

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -236,7 +239,7 @@ class CommandTrackerFlowTest
             Some(Status(Code.ABORTED.value)),
           )
         completionStreamMock.send(CompletionStreamElement.CompletionElement(completion))
-        results.requestNext().value shouldEqual completion
+        results.requestNext().value shouldEqual CompletionResponse(completion)
```
SamirTalwar-DA, line 242: Not sure I like that we're using the smart constructor here; I'd prefer we write the expected value out explicitly (and below).

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -236,7 +239,7 @@ class CommandTrackerFlowTest
...
         succeed
```
SamirTalwar-DA, line 243: `succeed` is unnecessary here.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandService.scala
```diff
@@ -151,43 +153,65 @@ private[apiserver] final class ApiCommandService private (
...
     submitAndWaitInternal(request).flatMap { resp =>
-      val effectiveActAs = CommandsValidator.effectiveSubmitters(request.getCommands).actAs
-      val txRequest = GetTransactionByIdRequest(
-        request.getCommands.ledgerId,
-        resp.transactionId,
-        effectiveActAs.toList,
+      resp.fold(
+        failure => Future.failed(CompletionResponse.toException(failure)),
+        { resp =>
+          {
```
SamirTalwar-DA, line 202: This second set of curly braces is unnecessary.

ledger/sandbox-classic/src/test/suite/scala/platform/sandbox/services/command/CommandStaticTimeIT.scala
```diff
@@ -107,7 +105,7 @@ final class CommandStaticTimeIT
               )
             )
         } yield {
-          completion.status.value.code should be(Status.OK.getCode.value())
+          result shouldBe Symbol("right")
```
SamirTalwar-DA, line 108: Please use `matchPattern` here too.

SamirTalwar-DA (commented):
Looks like a really good change. I have some questions and suggestions.

language-support/scala/bindings-akka/src/test/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlowUT.scala
SamirTalwar-DA: Oh, my bad, it's not part of the `case`. I misunderstood.

ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/CommandClientIT.scala
nicu-da: 👍 I don't like the fact that this is not capture in the scaladocs themselves, had to dive into the source code to figure it out

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
nicu-da: Agree, was just being lazy while fixing the test. will update

ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/CommandClientIT.scala
fabiotudone-da: Sadly, it often happens with Scala libs.

language-support/scala/bindings-akka/src/test/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlowUT.scala
```diff
@@ -97,9 +92,9 @@ class CommandRetryFlowUT extends AsyncWordSpec with Matchers with AkkaTest with
 
     "propagete OK status" in {
```
SamirTalwar-DA, line 93:
```suggestion
    "propagate OK status" in {
```

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
```diff
@@ -70,25 +70,39 @@ object CommandRetryFlow {
...
+                  case Left(completionFailure) =>
+                    completionFailure match {
+                      case CompletionResponse.NotOkResponse(_, status)
```
SamirTalwar-DA, line 79: Why not:

```scala
             case Ctx(
                    RetryInfo(request, nrOfRetries, firstSubmissionTime, _),
                    Left(CompletionResponse.NoOkResponse(_, status)),
                    _,
                  ) =>
```

Instead of 3 levels of `match` unpacking?

SamirTalwar-DA (approved):

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/retrying/CommandRetryFlow.scala
nicu-da: Changed to a single level of matches but I'm in doubts as to if it makes it easier to read

SamirTalwar-DA: FWIW, I prefer it. I have to keep less state in my head when reading.

SamirTalwar-DA (approved):

cocreature (approved):
Thanks, seems reasonable. I suggest to notify our internal users of this API in advance of this change (ask me on Slack if you need details).

@bame-da suggested that it would be nice if we could get all changes around improved error reporting into a single release to avoid too much churn which I think is a nice goal for actual ledger API changes but for this one, I’m also fine if it ends up coming before the actual release.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -0,0 +1,62 @@
...
+
+object CompletionResponse {
+  type CompletionResponse = Either[CompletionFailure, CompletionSuccess]
+  sealed trait CompletionFailure {
+    def metadata: Map[String, String] = Map.empty
+  }
+  final case class NotOkResponse(commandId: String, grpcStatus: StatusProto)
+      extends CompletionFailure
+  final case class TimeoutResponse(commandId: String) extends CompletionFailure
+  final case class NoStatusInResponse(commandId: String) extends CompletionFailure
```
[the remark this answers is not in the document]
nicu-da: In theory it should never happen.
But we have to handle this case because of the type we receive, as the GRPC status is optional in the protobuf `Completion`.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.google.protobuf.Any
+import com.google.rpc
+import com.google.rpc.status.{Status => StatusProto}
+import io.grpc.Status.Code
+import io.grpc.{StatusRuntimeException, protobuf}
+
+import scala.jdk.CollectionConverters._
+
+object CompletionResponse {
+  type CompletionResponse = Either[CompletionFailure, CompletionSuccess]
```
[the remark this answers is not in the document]
nicu-da: Will inline it, I aggree that it will be easier to comprehend

nicu-da:
> Thanks, seems reasonable. I suggest to notify our internal users of this API in advance of this change (ask me on Slack if you need details).
> 
> @bame-da suggested that it would be nice if we could get all changes around improved error reporting into a single release to avoid too much churn which I think is a nice goal for actual ledger API changes but for this one, I’m also fine if it ends up coming before the actual release.

@cocreature Thank you for the review.
I aggree it would be nice to try to restrict the churn for changes like this.
 I assume the only public breaking change would be in `LedgerClientBinding`, I could easily keep the old return type there and convert back to a completion before returning so that we can do the breaking change later on.

nicu-da:
@SamirTalwar-DA Last review request, I promise 😂 If you can take a look at the last 2 commits, added a backwards compatibility helper

SamirTalwar-DA (approved):
:shipit:

PR #10506, opened 2021-08-06
Split the Daml ledger export integration tests

aherrmann-da:
Depends-On: #10492

This test suite performs the same test with different parameters three
times. These tests each setup some state on the ledger, then create a
Daml ledger export script, compile it, run it, and compare the new
ledger state to the previous state.

Running all this three times in the same target get exceed the default
timeout of 300s in some case. This splits these tests into three
separate targets to reduce the risk of timeouts.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

Issue #10419, opened 2021-07-27
Next steps in the GSG


cocreature:
Currently the GSG leaves people a bit on their own when it comes to what they should do next. Instead we should provide concrete suggestions, specifically the proposal is the following:

1. Move the "testing your app" somewhere to "building applications" and make it clear that this is only about doing e2e webui tests.
2. Provide 4 possible next steps (order and title/description to be determined)
    1. Deploying your app to Daml Hub linking to a section that still needs to be created
    2. Model automation/triggers linking to the current daml triggers section
    3. Model testing/daml script linking to the current daml script section
    4. In-depth Daml language intro linking to the existing section
3. The Daml script section will be changed to build upon the model in the getting started guide. Potential additional information that doesn’t quite fit but shouldn’t be lost can go in a subsection or at the end.
4. The Daml trigger section will be changed to build upon the model in the getting started guide. Potential additional information that doesn’t quite fit but shouldn’t be lost can go in a subsection or at the end.

The connect engineering teams will start with 1,3 and 4. We’ll do 2 together with @nemanja-da and team once he is back from vacation.

garyverhaegen-da:
Any news on this? 1 and 4 are done (resp. #10479 & #10509), and I don't think I'm the best person to address 3, but could give it a shot if we still think it needs to be done.

@nemanja-da any progress on 2? Do we still need/want to do any of this, or are we happy with the current state?

nemanja-da:
@garyverhaegen-da would wait with writing 2. as we're merging the docs and getting-started with Canton as part of the next steps might also make sense. I'm not sure to what extent things are clear when it comes to how that will play out (Canton<->Daml docs integration) and that's why I'm suggesting that we wait a bit. 

cc: @davidrichards-da @carrielaben-da

davidrichards-da:
I agree with Nemanja to wait.  There's a broader work stream going on to tackle the docs generally so I wouldn't want any work wasted.  @carrielaben-da, you should consider these next steps for the getting started guide in any docs changes that are considered.

garyverhaegen-da:
This doesn't seem to be planned anymore. @carrielaben-da please reopen if I'm wrong.

PR #10509, opened 2021-08-06
rewrite trigger docs to follow gsg

garyverhaegen-da:
Per #10419 point 4, I've rewritten the Triggers section to build upon the Getting Started Guide instead of inventing its own example.

Compared to #10395, this has a lot more explanations as this page must now serve the dual purpose of being a possible "next step" from the GSG and being the main reference page for triggers. It's also lost the "next steps" section, which I think is a bit of a shame, but it doesn't really make sense here.

There's also no easy way for people not interested in the GSG to follow along; should we expose the "completed GSG" as a tempate?

CHANGELOG_BEGIN
CHANGELOG_END


garyverhaegen-da:
[Deployed version](https://garyverhaegen-da.github.io/daml-docs/2021-08-06-5cf34e7504/triggers/index.html), in case that makes reviewing easier.

docs/source/triggers/index.rst
```diff
@@ -14,81 +14,63 @@ and the UI, Daml applications often need to automate certain
...
-that we need to archive ``Copy`` contracts if there is more than one for the same ``Original``, we need to archive
-``Copy`` contracts if the corresponding ``Original`` has been archived and we need to archive
-all ``Copy`` s for a given subscriber if the corresponding ``Subscriber`` contract has been archived.
-
-Implementing a Daml Trigger
----------------------------
+Daml Trigger Basics
+===================
 
 Having defined what our Daml trigger is supposed to do, we can now
```
cocreature, line 50: this sentence sounds very weird now given that 1. the explanation has shrunk down to a single sentence now and 2. it’s no longer "the trigger should" but "we will build a trigger". Not quite sure how to rephrase it.

docs/source/triggers/index.rst
```diff
@@ -14,81 +14,63 @@ and the UI, Daml applications often need to automate certain
...
 move on to its implementation. A Daml trigger is a regular Daml
 project that you can build using ``daml build``. To get access to the
 API used to build a trigger, you need to add the ``daml-triggers``
```
cocreature, line 53:
```suggestion
API used to build a trigger, you need to add the ``daml-trigger``
```
was there before but it’s still broken :)

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
 We can specify the templates that our trigger will operate
 on. In our case, we will simply specify ``AllInDar`` which means that
 the trigger will receive events for all template types defined in the
-DAR. It is also possible to specify an explicit list of templates,
-e.g., ``RegisteredTemplates [registeredTemplate @Original, registeredTemplate @Subscriber, registeredTemplate @Copy]``.
+DAR. It is also possible to specify an explicit list of templates. For example,
```
cocreature, line 114: Not for this PR but we might want to consider splitting this into two sections:
1. The next step section
2. all the gory details on registered templates, hearbeats, … which don’t matter for this.

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+
+    autoReply : T.Trigger ()
+    autoReply = T.Trigger
+      { initialize = pure ()
+      , updateState = \_ -> pure ()
+      , rule = \p -> do
+          message_contracts <- T.query @User.Message
+          let messages = map snd message_contracts
+          debug $ "Messages so far: " <> show (length messages)
+          let lastMessage = List.last $ List.sortOn (.receivedAt) messages
```
cocreature, line 247: definitely not for this PR but we should have a `maximumOn`.

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+                users <- T.query @User.User
+                debug users
+                let isSender = (\user -> user.username == m.sender)
+                let replyTo = List.head $ filter (\(_, user) -> isSender user) users
+                case replyTo of
+                  None -> pure ()
+                  Some (sender, _) -> do
+                    T.dedupExercise sender (User.SendMessage p "Please, tell me more about that.")
+              else pure ()
+            None -> pure ()
```
cocreature, line 262:
```suggestion
          whenSome lastMessage $ \m -> do
              when (m.receiver == p) $ do
                users <- T.query @User.User
                debug users
                let isSender = (\user -> user.username == m.sender)
                let replyTo = List.findOptional (\(_, user) -> isSender user) users
                whenSome replyTo $ \(sender, _) -> do
                    T.dedupExercise sender (User.SendMessage p "Please, tell me more about that.")
```
arguably primitive pattern matching makes for a nicer intro but I think `when` and `whenSome` are so common that we can use them here.

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+  the ``SendMessage`` choice, passing in the current party ``p`` as the sender.
+  See below for additional information on what that ``dedup`` in the name of the
+  command means.
+- If the receiver of the last message was not the current party, then the last
+  message was sent by it and we don't need to reply to ourselves, so we do
+  nothing.
+- Finally, if we had not found any message at all, and thus ``lastMessage`` was
+  ``None``, we have nothing to do.
+
+Deduplication
```
cocreature, line 335:
```suggestion
Command Deduplication
```

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+  `head <daml/stdlib/DA-List-Total.html#function-da-list-total-head-74336>`_
+  of that list, i.e. its first element.
+- Just like  ``last``, ``head`` will return an ``Optional a``, so the next step
+  is to check whether we have actually found the relevant ``User`` contract. In
+  most cases we should find it, but remember that users can send us a message if
+  *we* follow *them*, whereas we can only answer if *they* follow *us*.
+- If ``replyTo`` is ``None``, then the sender is not, or no longer, following
+  us, and we can't respond. In that case we just do nothing, which we indicate by
+  returning ``pure ()``.
+- If we did find some ``User`` contrac to reply to, we extract the
```
cocreature, line 322:
```suggestion
- If we did find some ``User`` contract to reply to, we extract the
```

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+
+If this simple deduplication is not suited to your use-case, you have two other
+tools at your disposal. The first one is the ``getCommandsInflight`` action
+(`doc <https://docs.daml.com/triggers/api/Daml-Trigger.html#function-daml-trigger-getcommandsinflight-32524>`__),
+which returns all of the commands this instance of the trigger runner has sent
+and that have not yet been resolved (i.e. either committed or failed). You can
+then build your own logic based on this list, the ACS, and possibly your own
+trigger state.
+
+The last tool you have at your disposal is the second argument to the ``emitCommands`` action
```
cocreature, line 385: I’d swap these. `emitCommands` is pretty common. Writing your own logic based on commands in flight is not.

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+certain contracts, this can be a simple, effective way to prevent triggering it
+multiple times.
+
+Finally, do keep in mind that all of these mechanisms rely on internal state
+from the trigger runner, which keeps track of which commands it has sent and
+for which it's not seen a completion. They will all fail to deduplicate if that
+internal state is lost, e.g. if the trigger runner is shut down and a new one
+is started. As such, these deduplication mechanisms should be seen as an
+optimization rather than a requirement for correctness. The Daml model should
+be deisnged such that duplicated commands are either rejected (e.g. using keys
```
cocreature, line 399:
```suggestion
be designed such that duplicated commands are either rejected (e.g. using keys
```

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+      , heartbeat = None
+      }
+
+Refresh ``daml start`` by pressing ``r`` (followed by ``Enter`` on Windows) in
+its terminal, then kill (CTRL-C) the trigger and start it again. Play a bit
+with ``alice`` and ``bob`` in your browser, to get a feel for how the
+trigger works. Watch both the messages in-browser and the debug statements
+printed by the trigger runner.
+
+Let's walk through the ``rule`` code line-by-line:
```
cocreature, line 273: The explanation of trigger behavior that makes the most sense to me is to view them as trying to go change the current ACS to a target ACS and then they become a noop. The event handling descriptions usually end up being a bit fuzzy and confusing imho and lead to wrong expectations around when the trigger gets triggered.

In your case, the trigger tries to change the ACS such that the last message was sent by it and then stops.

docs/BUILD.bazel
```diff
@@ -630,16 +630,6 @@ pkg_tar(
     visibility = ["//visibility:public"],
 )
 
-pkg_tar(
```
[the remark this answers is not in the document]
garyverhaegen-da: Removing it is hard, so, yes.

docs/source/triggers/index.rst
```diff
@@ -125,159 +111,316 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+
+Running a No-Op Trigger
+=======================
+
+To implement a no-op trigger, one could write the following in a separate
+``daml/ChatBot.daml`` file:
+
+.. code-block:: daml
+
+    module ChatBot where
```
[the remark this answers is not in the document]
garyverhaegen-da: Done.

docs/source/triggers/index.rst
garyverhaegen-da: I think just removing the sentence works.

docs/source/triggers/index.rst
garyverhaegen-da: Done in #10527; this has been rebased on now uses `maximumOn`.

docs/source/triggers/index.rst
garyverhaegen-da: Corrected. Thanks!

docs/source/triggers/index.rst
garyverhaegen-da: Done.

docs/source/triggers/index.rst
garyverhaegen-da: Done.

docs/source/triggers/index.rst
garyverhaegen-da: Corrected, thanks!

docs/source/triggers/index.rst
garyverhaegen-da: I've added a section at the top, and tried to tweak the step-by-step explanations to go in that direction.

garyverhaegen-da:
@cocreature I believe I have now addressed all of your "for-this-PR" concerns. Rendered result [here](https://garyverhaegen-da.github.io/daml-docs/2021-08-16-ddbfbd27d7/triggers/index.html), in case that helps reviewing.

cocreature (approved):
Thanks!

daml-assistant/exe/DA/Daml/Assistant.hs
```diff
@@ -354,7 +354,7 @@ argWhitelist = S.fromList
     , "install", "latest", "project"
     , "uninstall"
     , "studio", "never", "always", "published"
-    , "new", "skeleton", "empty-skeleton", "quickstart-java", "quickstart-scala", "copy-trigger"
```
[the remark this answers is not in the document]
garyverhaegen-da: Ah, that's a mistake. I'll revert that line. At some point in the process I had just removed the `copy-trigger` template, then decided to reinstate it. I missed this line when putting it back.

docs/scripts/live-preview.sh
```diff
@@ -44,6 +46,12 @@ bazel build //templates:create-daml-app-docs
 mkdir -p $BUILD_DIR/source/getting-started/code/templates-tarball
 tar -zxf ../../bazel-bin/templates/create-daml-app-docs.tar.gz -C $BUILD_DIR/source/getting-started/code/templates-tarball/
 
+# Templates
```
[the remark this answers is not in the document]
garyverhaegen-da: It seems more logical to have the templates-tarball produced by the templates/BUILD file based on files in templates, and then have the docs depend on that, than have the docs depend on inline files and then have the templates/BUILD file grab source code from the docs.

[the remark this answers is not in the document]
garyverhaegen-da: Also, yes, because it's a patchfile, something somewhere has to know how to produce the complete template, and it seems better for that to be templates than docs. Not having it as a patchfile seemed way worse, as we definitely want it in sync with the create-daml-app template.

templates/gsg-trigger.patch
```diff
@@ -0,0 +1,94 @@
+diff -Naur create-daml-app/daml/ChatBot.daml gsg-trigger/daml/ChatBot.daml
```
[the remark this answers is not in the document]
garyverhaegen-da: Isn't it there already? Line 793? Is there some other place I should add it to?

daml-assistant/exe/DA/Daml/Assistant.hs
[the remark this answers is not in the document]
garyverhaegen-da: Also  added `gsg-trigger` to that list; I have no idea what that list is for, but if `copy-trigger` is in there I think `gsg-trigger` should be too.

docs/source/triggers/index.rst
```diff
@@ -125,159 +137,269 @@ Like ``Scenario`` or ``Update``, you can use ``do`` notation and
...
+  follow *us*.
+- If we did find some ``User`` contract to reply to, we extract the
+  corresponding contract id (first element of the tuple, ``sender``) and
+  discard the payload (second element, ``_``), and we
+  `exercise <triggers/api/Daml-Trigger.html#function-daml-trigger-dedupexercise-37617>`_
+  the ``SendMessage`` choice, passing in the current party ``p`` as the sender.
+  See below for additional information on what that ``dedup`` in the name of the
+  command means.
+
+Command Deduplication
```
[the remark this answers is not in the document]
garyverhaegen-da: I'd named it just "Deduplication"; adding the `Command` was [your suggestion](https://github.com/digital-asset/daml/pull/10509#discussion_r685157395).

"Dealing with Repeated Actions"?

This also makes me wonder if we should perhaps mention a fourth technique here (probably in first position): designing your Daml model such that the commands sent by the trigger are idempotent.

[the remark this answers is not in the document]
cocreature: > This also makes me wonder if we should perhaps mention a fourth technique here (probably in first position): designing your Daml model such that the commands sent by the trigger are idempotent.
that’s pretty difficult without contract key uniqueness, i.e., multi-domain canton so I’m hesitant to encourage that. I think really we need to properly support ledger-side command dedup by allowing users to set command ids but that isn’t possible atm.

I guess we could title it "avoiding sending the same command multiple times" but that also sounds very awkward.  Maybe command dedup isn’t too bad :confused:

daml-assistant/exe/DA/Daml/Assistant.hs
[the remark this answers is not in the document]
cocreature: it’s an allow-list for things that we log via telemetry.

templates/gsg-trigger.patch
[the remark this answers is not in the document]
cocreature: oh missed that nvm

PR #10517, opened 2021-08-09
Upgrade rules_haskell and rules_nixpkgs

aherrmann-da:
Upgrade rules_haskell and rules_nixpkgs to their latest master.
Removes patches that have been upstreamed since and adapts the remaining patches.
Refactor `rules_haskell|nixpkgs` patches to automatically keep them in sync between `daml` and `compatibility`.

Removes the strict-source-names patch that forced source names to match module names and tracked individual .o and .hi output files.
This patch was mainly motivated by issues with directory outputs on Windows occasionally producing empty outputs that would poison the remote cache. However, we have continued seeing this type of issue despite this patch and have since resorted to a regular fixup job that cleans such empty items out of the cache.

With that the strongest motivation with this change is no longer valid and it doesn't seem worth it to maintain it any longer.

### Pull Request Checklist

trigger the build.


bazel_tools/haskell-windows-extra-libraries.patch
```diff
@@ -1,37 +1,8 @@
-diff --git a/haskell/assets/ghc_8_10_3_win_base.patch b/haskell/assets/ghc_8_10_3_win_base.patch
-new file mode 100644
-index 00000000..b205f35c
---- /dev/null
-+++ b/haskell/assets/ghc_8_10_3_win_base.patch
```
aherrmann-da, line 5: We no longer use 8.10.3. So, this patch on the bindist is no longer used.

cocreature (approved):
Thanks!

PR #10518, opened 2021-08-09
Add flag to enable/disable command deduplication [KVL-1006]

nicu-da:
Command deduplication is enabled/disabled based on the WriteService implementation

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -111,10 +111,10 @@ private[apiserver] final class ApiSubmissionService private[services] (
     withEnrichedLoggingContext(logging.commands(request.commands)) { implicit loggingContext =>
       logger.info("Submitting transaction")
       logger.trace(s"Commands: ${request.commands.commands.commands}")
-      ledgerConfigurationSubscription
-        .latestConfiguration()
-        .fold(Future.failed[Unit](ErrorFactories.missingLedgerConfig()))(ledgerConfiguration => {
-          if (writeService.isDeduplicationEnabled) {
+      val evaluatedCommand = ledgerConfigurationSubscription
+        .latestConfiguration() match {
+        case Some(ledgerConfiguration) =>
+          if (writeService.isApiDeduplicationEnabled) {
             deduplicateAndRecordOnLedger(
               seedService.nextSeed(),
               request.commands,
@@ -124,7 +124,9 @@ private[apiserver] final class ApiSubmissionService private[services] (
             evaluateAndSubmit(seedService.nextSeed(), request.commands, ledgerConfiguration)
               .transform(handleSubmissionResult)
           }
-        })
+        case None => Future.failed[Unit](ErrorFactories.missingLedgerConfig())
+      }
+      evaluatedCommand
         .andThen(logger.logErrorsOnCall[Unit])
     }
 
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -80,7 +80,8 @@ private[apiserver] object ApiSubmissionService {
     )
 
   final case class Configuration(
-      implicitPartyAllocation: Boolean
+      implicitPartyAllocation: Boolean,
+      commandDeduplicationEnabled: Boolean,
   )
 
 }
@@ -113,8 +114,18 @@ private[apiserver] final class ApiSubmissionService private[services] (
       logger.trace(s"Commands: ${request.commands.commands.commands}")
       ledgerConfigurationSubscription
         .latestConfiguration()
-        .map(deduplicateAndRecordOnLedger(seedService.nextSeed(), request.commands, _))
-        .getOrElse(Future.failed(ErrorFactories.missingLedgerConfig()))
+        .fold(Future.failed[Unit](ErrorFactories.missingLedgerConfig()))(ledgerConfiguration => {
+          if (configuration.commandDeduplicationEnabled) {
+            deduplicateAndRecordOnLedger(
+              seedService.nextSeed(),
+              request.commands,
+              ledgerConfiguration,
+            )
+          } else {
+            evaluateAndSubmit(seedService.nextSeed(), request.commands, ledgerConfiguration)
+              .transform(handleSubmissionResult)
+          }
+        })
         .andThen(logger.logErrorsOnCall[Unit])
     }
 
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -80,8 +80,7 @@ private[apiserver] object ApiSubmissionService {
     )
 
   final case class Configuration(
-      implicitPartyAllocation: Boolean,
-      commandDeduplicationEnabled: Boolean,
+      implicitPartyAllocation: Boolean
   )
 
 }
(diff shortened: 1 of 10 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
```diff
@@ -113,8 +114,18 @@ private[apiserver] final class ApiSubmissionService private[services] (
       logger.trace(s"Commands: ${request.commands.commands.commands}")
       ledgerConfigurationSubscription
         .latestConfiguration()
-        .map(deduplicateAndRecordOnLedger(seedService.nextSeed(), request.commands, _))
-        .getOrElse(Future.failed(ErrorFactories.missingLedgerConfig()))
+        .fold(Future.failed[Unit](ErrorFactories.missingLedgerConfig()))(ledgerConfiguration => {
```
SamirTalwar-DA, line 117: Honestly I think `match` would be easier to read here.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -369,6 +369,13 @@ object Config {
             "The maximum number of commands waiting to be submitted for each party. Overflowing this threshold will cause back-pressure, signaled by a RESOURCE_EXHAUSTED error code. Default is 512."
           )
 
+        opt[Boolean]("command-deduplication-enabled")
+          .optional()
+          .action((value, config) =>
+            config.copy(commandConfig = config.commandConfig.copy(deduplicationEnabled = value))
+          )
+          .text("Enable or disable command deduplication. Default is enabled.")
```
SamirTalwar-DA, line 377: I don't think it's a good idea to expose this as a CLI option.

We will switch this from enabled to disabled when we switch from v1 of the participant state API to v2.

Right now it's enough to verify this works in tests.

ledger/sandbox-common/src/main/scala/platform/sandbox/cli/CommonCliBase.scala
```diff
@@ -249,6 +249,13 @@ class CommonCliBase(name: LedgerName) {
           "The maximum number of commands waiting to be submitted for each party. Overflowing this threshold will cause back-pressure, signaled by a RESOURCE_EXHAUSTED error code. Default is 512."
         )
 
+      opt[Boolean]("command-deduplication-enabled")
+        .optional()
+        .action((value, config) =>
+          config.copy(commandConfig = config.commandConfig.copy(deduplicationEnabled = value))
+        )
+        .text("Enable or disable command deduplication. Default is enabled.")
```
SamirTalwar-DA, line 257: Sandbox should never turn this off.

SamirTalwar-DA:
You have a typo in "CHANGELOG_BEGIN".

SamirTalwar-DA:
I'm thinking that instead of exposing this as a separate flag on `StandaloneApiServer`, we may want to provide two constructors:

- one that takes the v1 `ReadService` and `WriteService`, and enables participant-side command deduplication, and
- one that takes the v2 `ReadService` and `WriteService`, and disables it.

These two features are coupled; it doesn't make sense to be able to configure them independently.

What do you think?

nicu-da:
> I'm thinking that instead of exposing this as a separate flag on `StandaloneApiServer`, we may want to provide two constructors:
> 
>     * one that takes the v1 `ReadService` and `WriteService`, and enables participant-side command deduplication, and
> 
>     * one that takes the v2 `ReadService` and `WriteService`, and disables it.
> 
> 
> These two features are coupled; it doesn't make sense to be able to configure them independently.
> 
> What do you think?

As discussed. I moved the `deduplication` flag into the `WriteService` definition itself as it makes no sense to have them configured separetly.
v1.WriteService enables deduplication while v2.WriteService disables it (the adapter used from v1 to v2 enables deduplication)

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
nicu-da: I'm a lot more used to `fold` and find it easier to read as it flows linearly, but aggree it's not always the case so I'll change to a match.

ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
```diff
@@ -81,4 +81,6 @@ final class TimedWriteService(delegate: WriteService, metrics: Metrics) extends
 
   override def currentHealth(): HealthStatus =
     delegate.currentHealth()
+
+  override def isDeduplicationEnabled: Boolean = false
```
SamirTalwar-DA, line 85: This needs to call `delegate.isDeduplicationEnabled`.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
```diff
@@ -103,4 +103,9 @@ trait WriteService
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
   )(implicit telemetryContext: TelemetryContext): CompletionStage[SubmissionResult]
+
+  /** Indicates whether command deduplication should be enabled when using this [[WriteService]]
+    * This is temporary until we fully transition from [[com.daml.ledger.participant.state.v1.WriteService]] to [[WriteService]]
+    */
+  def isDeduplicationEnabled: Boolean
```
SamirTalwar-DA, line 110: As discussed, let's default it to `false`.

We _always_ deduplicate; the question is whether the API should deduplicate, or the driver. So perhaps we can rename this to `isApiDeduplicationEnabled`?

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/LedgerBackedWriteService.scala
```diff
@@ -98,4 +98,6 @@ private[stores] final class LedgerBackedWriteService(ledger: Ledger, timeProvide
       submissionId: Ref.SubmissionId,
   ): CompletionStage[state.PruningResult] =
     CompletableFuture.completedFuture(state.PruningResult.NotPruned(Status.UNIMPLEMENTED))
+
+  override def isDeduplicationEnabled: Boolean = false
```
SamirTalwar-DA, line 102: I think we _do_ need to set this to `true` for Sandbox Classic, as it currently relies on that feature. We have upcoming work to fix this.

ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
nicu-da: 🤦

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
nicu-da: Added the false default.
I chose to not add the default so that any implementer would have to be explicit about it to avoid surprises. But if you think it makes sense to have the default I'll push it

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
nicu-da: Change to match, I find the code a lot less readable

SamirTalwar-DA (approved):

ledger/participant-integration-api/src/test/suite/scala/platform/apiserver/services/ApiSubmissionServiceSpec.scala
```diff
@@ -327,18 +310,127 @@ class ApiSubmissionServiceSpec
...
       }
   }
+
+  behavior of "command deduplication"
+
+  it should "use deduplication if enabled" in {
+    val partyManagementService = mock[IndexPartyManagementService]
+    val writeService = mock[state.WriteService]
+    val indexSubmissionService = mock[IndexSubmissionService]
+    val mockCommandExecutor = mock[CommandExecutor]
```
fabiotudone-da, line 320:
```suggestion
    val partyManagementService = mock[IndexPartyManagementService]
    val writeService = mock[state.WriteService]
    val indexSubmissionService = mock[IndexSubmissionService]
    val mockCommandExecutor = mock[CommandExecutor]
```

I suggest to have them named consistently, i.e. either all with a `mock` prefix (my preference, can make named parameters redundant) or all without.

fabiotudone-da (approved):

PR #10523, opened 2021-08-09
H2 Storage backend support for canton jdbc urls with user/password

oliverse-da:
The H2 database's JdbcDataSource does not like seeing user/password
properties in the jdbc url raising errors upon hikari connection pool
initialization:

`org.h2.jdbc.JdbcSQLNonTransientConnectionException: Duplicate property "USER"`

expecting to be able to set user and password separately from passing on
the jdbc url.

As h2 is not supported in production and to get canton integration tests past this
resorting to parsing the jdbc url looking for user/password properties, removing
them from the url and instead setting them explicitly on the data source object.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
@@ -197,7 +197,32 @@ private[backend] object H2StorageBackend
       connectionInitHook: Option[Connection => Unit],
   )(implicit loggingContext: LoggingContext): DataSource = {
     val h2DataSource = new org.h2.jdbcx.JdbcDataSource()
-    h2DataSource.setUrl(jdbcUrl)
+
+    // H2 (org.h2.jdbcx.JdbcDataSource) does not support setting the user/password within the jdbcUrl, so remove
+    // those properties from the url if present and set them separately. This helps run canton tests which set
+    // user/password. Note that Postgres and Oracle support user/password in the URLs, so we don't bother exposing
+    // user/password configs separately from the url just for h2 which is anyway not supported for production.
+    def setKeyValueAndRemoveFromUrl(url: String, key: String, setter: String => Unit): String = {
+      val separator = ";"
+      url.toLowerCase.indexOf(separator + key + "=") match {
+        case -1 => url
+        case i =>
+          val valueBegin = i + 1 + key.length + 1 // separator, key, "="
+          val (value, shortenedUrl) = url.indexOf(separator, i + 1) match {
+            case -1 => (url.substring(valueBegin), url.take(i))
+            case next =>
+              (url.substring(valueBegin, next), url.take(i) + url.takeRight(url.length - next))
+          }
+          setter(value)
+          shortenedUrl
+      }
+    }
+
+    val urlNoUser = setKeyValueAndRemoveFromUrl(jdbcUrl, "user", h2DataSource.setUser)
+    val urlNoUserNoPassword =
+      setKeyValueAndRemoveFromUrl(urlNoUser, "password", h2DataSource.setPassword)
+    h2DataSource.setUrl(urlNoUserNoPassword)
+
     InitHookDataSourceProxy(h2DataSource, connectionInitHook.toList)
   }
 
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
@@ -199,31 +199,35 @@ private[backend] object H2StorageBackend
     val h2DataSource = new org.h2.jdbcx.JdbcDataSource()
 
     // H2 (org.h2.jdbcx.JdbcDataSource) does not support setting the user/password within the jdbcUrl, so remove
-    // those properties from the url if present and set them separately. This helps run canton tests which set
-    // user/password. Note that Postgres and Oracle support user/password in the URLs, so we don't bother exposing
-    // user/password configs separately from the url just for h2 which is anyway not supported for production.
-    def setKeyValueAndRemoveFromUrl(url: String, key: String, setter: String => Unit): String = {
-      val separator = ";"
-      url.toLowerCase.indexOf(separator + key + "=") match {
-        case -1 => url
-        case i =>
-          val valueBegin = i + 1 + key.length + 1 // separator, key, "="
-          val (value, shortenedUrl) = url.indexOf(separator, i + 1) match {
-            case -1 => (url.substring(valueBegin), url.take(i))
-            case next =>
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -197,7 +197,32 @@ private[backend] object H2StorageBackend
       connectionInitHook: Option[Connection => Unit],
   )(implicit loggingContext: LoggingContext): DataSource = {
     val h2DataSource = new org.h2.jdbcx.JdbcDataSource()
-    h2DataSource.setUrl(jdbcUrl)
+
+    // H2 (org.h2.jdbcx.JdbcDataSource) does not support setting the user/password within the jdbcUrl, so remove
+    // those properties from the url if present and set them separately. This helps run canton tests which set
+    // user/password. Note that Postgres and Oracle support user/password in the URLs, so we don't bother exposing
```
rautenrieth-da, line 203: > This helps run canton tests which set user/password.

Do we need to mention Canton here? The change makes sense even without Canton.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -197,7 +197,32 @@ private[backend] object H2StorageBackend
...
+    // user/password configs separately from the url just for h2 which is anyway not supported for production.
+    def setKeyValueAndRemoveFromUrl(url: String, key: String, setter: String => Unit): String = {
```
rautenrieth-da, line 205: If you extract this to a public method, we can unit test this. I'm currently working on adding tests on `StorageBackend` level, where it would be easy to add a test for this.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -197,7 +197,32 @@ private[backend] object H2StorageBackend
...
+      val separator = ";"
+      url.toLowerCase.indexOf(separator + key + "=") match {
+        case -1 => url
+        case i =>
+          val valueBegin = i + 1 + key.length + 1 // separator, key, "="
+          val (value, shortenedUrl) = url.indexOf(separator, i + 1) match {
+            case -1 => (url.substring(valueBegin), url.take(i))
+            case next =>
+              (url.substring(valueBegin, next), url.take(i) + url.takeRight(url.length - next))
+          }
```
rautenrieth-da, line 215: Wouldn't it be easier to match for regex `s";${key}=([^;]*)"`?

rautenrieth-da (approved):
Looks good, left a few minor comments.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
oliverse-da: Indeed it does, so I am moving the reference to canton to a parenthetical note the end (so the canton dependency is clear in the code, as no canton code is yet in the daml repo).

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
oliverse-da: Makes sense. Done.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
oliverse-da: I actually looked into using regex as the first solution. I believe that would work as well although it seemed as I would have to make two separate regex calls - one to extract the value and another to `Regex.replaceFirst`.

rautenrieth-da: The regex I mentioned should give you two results: the first the substring that needs to be removed (can be passed to `String.replace`), the second the extracted value (can be passed to `setter`). I didn't actually try it though, so not sure if it would be actually better.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -197,10 +197,39 @@ private[backend] object H2StorageBackend
...
   )(implicit loggingContext: LoggingContext): DataSource = {
     val h2DataSource = new org.h2.jdbcx.JdbcDataSource()
-    h2DataSource.setUrl(jdbcUrl)
+
+    // H2 (org.h2.jdbcx.JdbcDataSource) does not support setting the user/password within the jdbcUrl, so remove
+    // those properties from the url if present and set them separately. Note that Postgres and Oracle support
+    // user/password in the URLs, so we don't bother exposing user/password configs separately from the url just for h2
+    // which is anyway not supported for production. (This also helps run canton h2 participants that set user and
+    // password.)
+    val urlNoUser = setKeyValueAndRemoveFromUrl(jdbcUrl, "user", h2DataSource.setUser)
```
nmarton-da, line 206: The parsing logic seems a non-trivial string logic, how about:
- having a more streamlined signature to the pure function like def extractUserPasswordAndRemoveFromUrl(jdbcUrl: String): (String, Option[String], Option[String]) // (clensedUrl, foundUser, foundPassword)
- having some simple unit tests to guard the pure function

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
nmarton-da: I think it is a great idea to use regex here as well, if we can!

nmarton-da (approved):
as discussed with @oliverse-da I am fine with this implementation as a temporary one, which will be improved later based on the review comments.

oliverse-da:
> as discussed with @oliverse-da I am fine with this implementation as a temporary one, which will be improved later based on the review comments.

Follow-up tracked in https://github.com/DACH-NY/canton/issues/6865

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
oliverse-da: Yes we can: Here is the regex version that seems to work when I tried manually:

```
  def setKeyValueAndRemoveFromUrl(url: String, key: String, setter: String => Unit): String = {
    val regex = new Regex(s".*(;${key}=([^;])).*")
    url match {
      case regex(keyAndValue, value) =>
        setter(value)
        url.replace(keyAndValue, "")
      case _ => url
    }
  }
```

PR #10524, opened 2021-08-09
Normalize transactions & values as a separate pass

nickchapman-da:
Normalize transactions and embedded values as a separate pass.
- This normalization pass is used by a new simpler implementation of `isReplayedBy`.
- The most significant aspect of the normalization is to drop the optional type info from Record and Variant values.

Eventually we would like to produce normalized values and transactions directly from the engine. When this is done, we will no longer need this separate normalization pass.

There are some failing ledger tests to be resolved before this PR is merged.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -222,7 +222,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       (rtx, _) = result
       validationResult <-
         transaction.Validation
-          .isReplayedBy(tx, rtx)
+          .isReplayedBy(transaction.Normalization.normalizeTx(tx), rtx)
           .fold(
             e => ResultError(Error.Validation.ReplayMismatch(e)),
             _ => ResultDone.Unit,
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -222,7 +222,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       (rtx, _) = result
       validationResult <-
         transaction.Validation
-          .isReplayedBy(transaction.Normalization.normalizeTx(tx), rtx)
+          .isReplayedBy(tx, rtx)
           .fold(
             e => ResultError(Error.Validation.ReplayMismatch(e)),
             _ => ResultDone.Unit,
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
@@ -137,6 +137,13 @@ class Normalization[Nid, Cid] {
         V.ValueGenMap(entries.map { case (k, v) => (normValue(k), normValue(v)) })
       case V.ValueTextMap(x) => V.ValueTextMap(x.mapValue(normValue))
 
+      // non-recursive cases (with normalization)
+      case V.ValueEnum(_, cons) =>
+        V.ValueEnum(
+          None, //norm
+          cons,
+        )
+
       // non-recursive cases
       case V.ValueContractId(_) => x
       case _: V.ValueCidlessLeaf => x
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
@@ -151,3 +151,9 @@ class Normalization[Nid, Cid] {
   }
 
 }
+
+object Normalization {
+  def normalizeTx[Nid, Cid](tx: VersionedTransaction[Nid, Cid]): VersionedTransaction[Nid, Cid] = {
+    new Normalization().normalizeTx(tx)
+  }
+}
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
@@ -72,8 +72,8 @@ class Normalization[Nid, Cid] {
 
       case old: NodeCreate[_] =>
         old
-          .copy(arg = normValue(old.arg))
-          .copy(key = old.key.map(normKWM))
+          .copy(arg = normValue(old.version)(old.arg))
+          .copy(key = old.key.map(normKWM(old.version)))
 
       case old: NodeFetch[_] =>
         (if (old.version >= TransactionVersion.minByKey) {
@@ -82,7 +82,7 @@ class Normalization[Nid, Cid] {
            old.copy(byKey = false)
          })
           .copy(
-            key = old.key.map(normKWM)
+            key = old.key.map(normKWM(old.version))
           )
 
       case old: NodeExercises[_, _] =>
@@ -92,14 +92,14 @@ class Normalization[Nid, Cid] {
            old.copy(byKey = false)
          })
           .copy(
-            chosenValue = normValue(old.chosenValue),
-            exerciseResult = old.exerciseResult.map(normValue),
-            key = old.key.map(normKWM),
+            chosenValue = normValue(old.version)(old.chosenValue),
+            exerciseResult = old.exerciseResult.map(normValue(old.version)),
+            key = old.key.map(normKWM(old.version)),
           )
 
       case old: NodeLookupByKey[_] =>
         old.copy(
-          key = normKWM(old.key)
+          key = normKWM(old.version)(old.key)
         )
 
       case old: NodeRollback[_] => old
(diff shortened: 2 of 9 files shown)


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -4,395 +4,53 @@
...
-    new Validation().valueIsReplayedBy(recorded, replayed)
-
 }
 
-sealed abstract class ReplayMismatch[Nid, Cid] extends Product with Serializable {
-  def recordedTransaction: VersionedTransaction[Nid, Cid]
-  def replayedTransaction: VersionedTransaction[Nid, Cid]
-
+final case class ReplayMismatch[Nid, Cid](
+    recordedTransaction: VersionedTransaction[Nid, Cid],
```
cocreature, line 50: this seems slightly sad, we produce somewhat worse errors now. I think this is fine since you never should see this without a malicious submitter but maybe worth considering if we should rely on structural equality at the node level but still traverse the transaction so we can pin it down to a node but let’s not do this in this transaction.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -4,395 +4,53 @@
...
-    * @note This function is asymmetric in order to provide backward compatibility.
-    *      For instance, some field may be undefined in the [[recorded]] transaction
-    *      while present in the [[replayed]] one.
+    * @note This function is symmetric
+    *
+    * 'isReplayedBy' normalizes both arguments, and then determines structural equality.
+    *
+    * The RIGHT `replayed` arg must be normalized because the engine currently does not.
+    *
+    * The LEFT `recorded` arg is normalized for the sake of test cases which do not ensure
```
cocreature, line 20: can we do this at the call site only in tests (potentially via a separate helper function) to avoid the performance impact on a real ledgers?

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
```diff
@@ -0,0 +1,153 @@
...
+    */
+
+  private type Val = V[Cid]
+  private type KWM = KeyWithMaintainers[Val]
+  private type Node = GenNode[Nid, Cid]
+  private type VTX = VersionedTransaction[Nid, Cid]
+
+  def normalizeTx(vtx: VTX): VTX = {
+    vtx match {
+      case VersionedTransaction(_, nodes, roots) =>
```
cocreature, line 48: similar point as below, I’d like to avoid mixing up things we do for convenience in tests with things we do on a production ledger so I’d prefer to not do the reversioning by default.

As for reusing `asVersionedTransaction`, I think you can already do this fairly easily, something like:
```
case VersionedTransaction(_, nodes, roots) =>
  val VersionedTransaction(version, _, _) = TransactionVersion.asVersionedTransaction(GenTransaction(nodes, roots))

File: daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala [hunk 4]
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  // It's ok to code this in stack unaware fashion, as values  have a depth limit of 100
+  private def normValue(x: Val): Val = {
+    x match {
+
+      // recursive cases (with normalization)
+      case V.ValueRecord(_, fields) =>
+        V.ValueRecord(
+          tycon = None, //norm
```
cocreature, line 117: That’s not right. We should only drop the type constructor for LF >= 1.12.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
```diff
@@ -0,0 +1,153 @@
...
+            (
+              None, //norm
+              normValue(v),
+            )
+          },
+        )
+
+      case V.ValueVariant(_, variant, v) =>
+        V.ValueVariant(
+          None, //norm
```
cocreature, line 128: Same here

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
```diff
@@ -0,0 +1,153 @@
...
+  private def normValue(x: Val): Val = {
+    x match {
+
+      // recursive cases (with normalization)
+      case V.ValueRecord(_, fields) =>
+        V.ValueRecord(
+          tycon = None, //norm
+          fields = fields.map { case (_, v) =>
+            (
+              None, //norm
```
cocreature, line 120: and here

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
```diff
@@ -0,0 +1,153 @@
...
+      // other recursive cases
+      case V.ValueList(list) => V.ValueList(list.map(normValue))
+      case V.ValueOptional(opt) => V.ValueOptional(opt.map(normValue))
+      case V.ValueGenMap(entries) =>
+        V.ValueGenMap(entries.map { case (k, v) => (normValue(k), normValue(v)) })
+      case V.ValueTextMap(x) => V.ValueTextMap(x.mapValue(normValue))
+
+      // non-recursive cases
+      case V.ValueContractId(_) => x
+      case _: V.ValueCidlessLeaf => x
```
cocreature, line 142: This isn’t correct. you need to normalise enums to drop the constructor in LF >= 1.12. I recommend to match on the individual types to avoid issues like this in the future. This is also another argument to not normalise the serialized version outside of testst which should have caught this bug.

cocreature (changes requested):
Thanks! Mostly looks good but there are a few issues in the normalization code.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
nickchapman-da: done
(by switching to existing code for doing value normalization)

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Normalization.scala
nickchapman-da: fixed
but then switched to existing code anyway!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -222,7 +222,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
       (rtx, _) = result
       validationResult <-
         transaction.Validation
-          .isReplayedBy(tx, rtx)
+          .isReplayedBy(transaction.Normalization.normalizeTx(tx), rtx)
```
cocreature, line 225: Why do we normalize here? Shouldn’t the submitted transaction already be serialized?

cocreature (approved):
Looks good except for the normalization in `validate`. Thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
nickchapman-da: I'm not sure we can rely on this. If we remove this call to `normalizeTx` then we get test failures.
For example:
```
bazel run //ledger/ledger-on-memory:ledger-on-memory-tests_test_suite_src_test_suite_scala_com_daml_ledger_on_memory_InMemoryLedgerReaderWriterIntegrationSpec.scala`
```

Tracing the stack shows the call to `Engine.validate` comes from `ModelConformanceValidator` from `TransactionCommitter`, and ultimately from `PreExecutingSubmissionValidator`.

I assume the TX has not been serialized at this point.

cocreature: Hm interesting, do you have a list of test failures? Wondering if it’s reasonable to insert normalization there in particular if it’s things like ledger-on-memory where we don’t care about performance rather than unnecessarily normalizing in places where performance might matter.

nickchapman-da: I just pushed a commit to remove the call to `normalizeTx` from `Engine.validate`, so CI can give us the full list...

All of the failures I have seen locally (so far) have looked like this:
```
The partial function passed as the second parameter to inside was not defined at the value passed as the first parameter to inside, which was: CommandRejected(2021-08-16T10:04:37.215Z,SubmitterInfo(List(alice),tests,X1,2021-08-16T10:04:47.158Z),Disputed(recreated and original transaction mismatch VersionedTransaction(V11,HashMap(),ImmArray()) expected, but VersionedTransaction(V10,Map(),ImmArray()) is recreated)) (ParticipantStateIntegrationSpecBase.scala:780)
```
So the mismatch is:
`VersionedTransaction(V11,HashMap(),ImmArray())`
vs
`VersionedTransaction(V10,Map(),ImmArray())`

This is quite interesting because the correct version for an empty TX is `V10`. But somewhere in the code is creating them at `V11`! Just need to find where this is happening...

nickchapman-da: The above badly version empty TX issue causes fails in the following 4 tests:
```
FAIL: //ledger/ledger-on-sql:ledger-on-sql-tests_test_suite_src_test_suite_scala_com_daml_ledger_on_sql_SqliteMemorySqlLedgerReaderWriterIntegrationSpec.scala
FAIL: //ledger/ledger-on-sql:ledger-on-sql-tests_test_suite_src_test_suite_scala_com_daml_ledger_on_sql_PostgresqlSqlLedgerReaderWriterIntegrationSpec.scala
FAIL: //ledger/ledger-on-sql:ledger-on-sql-tests_test_suite_src_test_suite_scala_com_daml_ledger_on_sql_H2MemorySqlLedgerReaderWriterIntegrationSpec.scala
FAIL: //ledger/ledger-on-memory:ledger-on-memory-tests_test_suite_src_test_suite_scala_com_daml_ledger_on_memory_InMemoryLedgerReaderWriterIntegrationSpec.scala
```

nickchapman-da: CI (windows/macos) confirms it is just these 4 failures.

cocreature: Interesting, wouldn’t have expected ledger on sql to fail. Fine to normalize for now and then drop it once we normalize within the engine.

nickchapman-da: :+1:

PR #10527, opened 2021-08-09
add minimumOn & maximumOn

garyverhaegen-da:
Requested in #10509. Should we also add `minimumBy` and `maximumBy` for consistency with `sortBy`?

```
CHANGELOG_BEGIN
- The DA.List and DA.List.Total modules now export minimumOn and
  maximumOn, respectively behaving similarly to (head . sortOn f) and
  (last . sortOn f).
CHANGELOG_END
```

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/List.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/List.daml
@@ -32,6 +32,8 @@ module DA.List
   , sort
   , sortBy
   , sortOn
+  , minimumOn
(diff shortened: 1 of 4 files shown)


cocreature (approved):
Thanks! I think `minimumBy` and `maximumBy` would be very nice additions as well.

garyverhaegen-da:
> Thanks! I think `minimumBy` and `maximumBy` would be very nice additions as well.

Added.

cocreature (approved):

PR #10528, opened 2021-08-09
factor the complicated bits of contract selection between PG, Oracle

S11001001:
#10344 brought the two implementations of `selectContractsMultiTemplate` close together enough that they can be usefully factored.  Here is that factoring.

Several of the arguments to `queryByCondition` take the form `(Read[T], T => Out)`, i.e. `Coyoneda`; we could invert the control by returning a data structure with coyonedas, but instead here we use a sort of continuation-passing style, so the coyonedas are embedded in the arguments to `queryByCondition`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -302,6 +302,57 @@ sealed abstract class Queries(tablePrefix: String) {
       ipol: SqlInterpol,
   ): Query0[DBContract[Mark, JsValue, JsValue, Vector[String]]]
 
+  // implementation aid for selectContractsMultiTemplate
+  @nowarn("msg=parameter value evidence.* is never used")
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

realvictorprm:
will restart the windows build as it seems to be only flaky.

realvictorprm (approved):
Well I do not know/understand yet what `Coyoneda`'s are but maybe you can shed some light on that in some sort of tech talk? Besides, although I don't understand the code completely here it still overall looks good to me, many thanks!

realvictorprm:
kicking off windows again

S11001001:
> Well I do not know/understand yet what `Coyoneda`'s are but maybe you can shed some light on that in some sort of tech talk?

@realvictorprm I don't know about that, but you may find [this](https://github.com/scalaz/scalaz/blob/v7.3.4/example/src/main/scala/scalaz/example/ContravariantCoyonedaUsage.scala) helpful. It's meant to be read through top to bottom.

PR #10529, opened 2021-08-09
DPP-468 StorageBackend tests

rautenrieth-da:
This PR adds tests for the `StorageBackend` interface:
- All tests run on all implementations (Postgres, H2, Oracle)
- The database is only created and migrated once, but the content of the database is `reset()` between each test
- For now, only functionality not already covered by JdbcLedgerDao is included


ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsIngestion.scala
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.backend
+
+import org.scalatest.flatspec.AsyncFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+private[backend] trait StorageBackendTestsIngestion extends Matchers with StorageBackendSpec {
```
nmarton-da, line 9: I think we need something more here. The initializeIngestion needs to be tested exhaustively:
* return value, corner case empty db
* for all types for ingested entries, the overspill entries should be wiped, but not the ones which are before the last ledger end
hint: for the overspill testing:
* either add capabilities of direct table observatin
* or move the ledger end sufficiently ahead, and test with backend retrival methods

nmarton-da (changes requested):
Looks very nice! Loved the safeguard around parallelism, and the reseting between run-s.
Added a commit which removes the need for tracking the type param for StorageBackend
Missing some tests for ingestion

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsIngestion.scala
```diff
@@ -0,0 +1,72 @@
...
+      configBeforeLedgerEndUpdate <- executeSql(backend.ledgerConfiguration)
+      _ <- executeSql(backend.updateLedgerEnd(ParameterStorageBackend.LedgerEnd(someOffset, 0)))
+      configAfterLedgerEndUpdate <- executeSql(backend.ledgerConfiguration)
+    } yield {
+      // The first query is executed before the ledger end is updated.
+      // It should not see the already ingested configuration change.
+      configBeforeLedgerEndUpdate shouldBe empty
+
+      // The second query should now see the configuration change.
+      configAfterLedgerEndUpdate should not be empty
```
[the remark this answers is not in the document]
rautenrieth-da: Good point. In general, the StorageBackend suite does not verify the serialization and deserialization of values like Daml-Lf values or Daml-Lf packages. They are all treated as opaque blobs.
`StorageBackend.ledgerConfiguration` is special because it returns a deserialized value, whereas all other read methods return streams or arrays of bytes. I have changed this line to assert on the configuration contents.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsIngestion.scala
rautenrieth-da: Added two tests:
- One that checks the return value of `initializeIngestion`
- One that checks `initializeIngestion` deleted overspill entries (by advancing the ledger end)

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -57,6 +57,25 @@ private[backend] object H2StorageBackend
     ()
   }
 
+  override def resetAll(connection: Connection): Unit = {
```
nmarton-da, line 60: maybe add reuse here?
how about unifying the truncation in common? the one-by-one style seems compatible with all. and the referential integrity treatment with H2 seems to be doable with some set to false;super.reset;set true sequence.

nmarton-da (approved):
Very nice addition to out test armada!

PR #10535, opened 2021-08-10
test: test for platform independent dars

robin-da:
We add a daily test to check that dars don't depend on the underlying
operating system where the dar is build.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build-unix.yml
+++ b/ci/build-unix.yml
@@ -38,6 +38,10 @@ steps:
       DAML_SDK_RELEASE_VERSION: ${{parameters.release_tag}}
       DAML_SCALA_VERSION: ${{parameters.scala_version}}
 
+  - template: publish-platform-independence-dar.yml
+    parameters:
+      platform: '${{parameters.name}}'
+
   - task: PublishBuildArtifacts@1
     condition: succeededOrFailed()
     continueOnError: true
--- a/ci/build-windows.yml
+++ b/ci/build-windows.yml
@@ -32,6 +32,10 @@ steps:
       #   number).
       SKIP_TESTS: ${{parameters.skip_tests}}
 
+  - template: publish-platform-independence-dar.yml
+    parameters:
+      platform: 'windows'
+
   - task: PublishBuildArtifacts@1
     condition: failed()
     displayName: 'Publish the bazel test logs'
--- a/ci/build.yml
+++ b/ci/build.yml
@@ -338,6 +338,48 @@ jobs:
         trigger_sha: '$(trigger_sha)'
     - template: report-end.yml
 
+- job: platform_independence_test
+  dependsOn:
+    - Windows
+    - Linux
+    - macOS
+  pool:
+    name: 'ubuntu_20_04'
+    demands: assignment -equals default
+  steps:
+    - checkout: self
+    - bash: ci/dev-env-install.sh
+      displayName: 'Build/Install the Developer Environment'
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifactName: platform-independence-dar-linux
+        targetPath: $(Build.StagingDirectory)/platform-independence/linux/
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifactName: platform-independence-dar-windows
+        targetPath: $(Build.StagingDirectory)/platform-independence/windows/
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        artifactName: platform-independence-dar-macos
+        targetPath: $(Build.StagingDirectory)/platform-independence/macos/
+    - bash: |
+        set -euo pipefail
+        eval "$(./dev-env/bin/dade-assist)"
+
+        DIR1=$(mktemp -d)
+        DIR2=$(mktemp -d)
+        DIR3=$(mktemp -d)
+        trap "rm -rf $DIR1; rm -rf $DIR2; rm -rf $DIR3" EXIT
+
+        unzip -d $DIR1 $(Build.StagingDirectory)/platform-independence/linux/platform-independence.dar
+        unzip -d $DIR2 $(Build.StagingDirectory)/platform-independence/windows/platform-independence.dar
+        unzip -d $DIR3 $(Build.StagingDirectory)/platform-independence/macos/platform-independence.dar
+
+        # hie/hi files may differ.
+        diff -r --strip-trailing-cr -x '*.hie' -x '*.hi' $DIR1 $DIR2
+        diff -r --strip-trailing-cr -x '*.hie' -x '*.hi' $DIR1 $DIR3
+      displayName: 'Compare platform-independence dars of different platforms.'
+
 - job: compatibility_ts_libs
   dependsOn:
     - da_ghc_lib
--- a/ci/publish-platform-independence-dar.yml
+++ b/ci/publish-platform-independence-dar.yml
@@ -1,0 +1,12 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+parameters:
+  platform: ''
+
+steps:
+  - task: PublishBuildArtifacts@1
+    inputs:
+        pathtoPublish: bazel-bin/compiler/damlc/tests/platform-independence.dar
+        artifactName: platform-independence-dar-${{parameters.platform}}
+
--- a/compiler/damlc/tests/BUILD.bazel
+++ b/compiler/damlc/tests/BUILD.bazel
@@ -4,7 +4,7 @@
 load("//bazel_tools:haskell.bzl", "da_haskell_binary", "da_haskell_library", "da_haskell_test")
 load("@os_info//:os_info.bzl", "is_windows")
 load(":util.bzl", "damlc_compile_test")
-load("//rules_daml:daml.bzl", "daml_compile")
+load("//rules_daml:daml.bzl", "daml_compile", "daml_compile_with_dalf")
 load("@build_environment//:configuration.bzl", "sdk_version")
 load("//daml-lf/language:daml-lf.bzl", "COMPILER_LF_VERSIONS", "lf_version_configuration")
 load("//compiler/damlc:util.bzl", "ghc_pkg")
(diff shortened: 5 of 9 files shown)


ci/cron/daily-compat.yml
```diff
@@ -67,6 +67,27 @@ jobs:
           pathtoPublish: '$(Build.StagingDirectory)'
           artifactName: 'Bazel Compatibility Logs'
 
+  - job: platform-independence-windows
```
garyverhaegen-da, line 70: Job names need to use underscores instead of dashes.

robin-da: fixed, thanks.

ci/cron/daily-compat.yml
```diff
@@ -67,6 +67,27 @@ jobs:
...
           artifactName: 'Bazel Compatibility Logs'
 
+  - job: platform_independence_windows
+    timeoutInMinutes: 60
+    pool:
+      name: windows-pool
+      demands: assignment -equals default
+    steps:
+      - checkout: self
+      - template: ../clean-up.yml
```
garyverhaegen-da, line 77: clean-up is unix-only
```suggestion
```

ci/platform-independence.yml
```diff
@@ -0,0 +1,22 @@
...
+    displayName: 'Configure Bazel for root workspace'
+    env:
+      IS_FORK: $(System.PullRequest.IsFork)
+      # to upload to the bazel cache
+      GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+  - task: DownloadPipelineArtifact@0
+    inputs:
+      artifactName: 'platform-independence-dar'
+      targetPath: $(Build.SourcesDirectory)/platform-independence.dar
+  - bash: |
```
garyverhaegen-da, line 17: You need to activate dev-env before you can use Bazel
```suggestion
  - bash: |
      set -euo pipefail
      eval "$(dev-env/bin/dade-assist)"
```

ci/platform-independence.yml
```diff
@@ -0,0 +1,22 @@
...
+      bazel build //compiler/damlc/tests:platform-independence.dar -o platform-independence-linux.dar
+    displayName: 'Rebuild platform-independence.dar'
+  - bash: |
```
garyverhaegen-da, line 20: Every `bash` task at the YAML level is a separate Bash process so you need to activate dev-env again here:
```suggestion
  - bash: |
      set -euo pipefail
      eval "$(dev-env/bin/dade-assist)"
```

compiler/damlc/tests/src/platform-independence-test.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+DAR1=$(sha1sum $1)
+DAR2=$(sha1sum $2)
+
+if [ $DAR1 -eq $DAR2 ]
```
garyverhaegen-da, line 10: This should be equivalent to just the condition, i.e.
```suggestion
[ $DAR1 -eq $DAR2 ]
```

Equivalently, the whole script here could be replaced by a call to `diff`.

compiler/damlc/tests/src/platform-independence-test.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+  exit 0
+else
+  exit 1
+fi
```
garyverhaegen-da, line 15: This should be equivalent to just the condition, i.e.
```suggestion
[ $DAR1 -eq $DAR2 ]
```

Equivalently, the whole script here could be replaced by a call to `diff`.

ci/cron/daily-compat.yml
robin-da: fixed, thanks!

ci/platform-independence-windows.yml
```diff
@@ -0,0 +1,16 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+steps:
+  - bash: ci/configure-bazel.sh
+    displayName: 'Configure Bazel for root workspace'
+    env:
+      IS_FORK: $(System.PullRequest.IsFork)
+      # to upload to the bazel cache
+      GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+  - powershell: 'bazel build //compiler/damlc/tests:platform-independence.dar -o platform-independence.dar'
```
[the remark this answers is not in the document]
robin-da: added the code from `build.ps1`.

ci/platform-independence.yml
```diff
@@ -0,0 +1,32 @@
...
+      GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+  - task: DownloadPipelineArtifact@0
+    inputs:
+      artifactName: 'platform-independence-dar'
+      targetPath: $(Build.StagingDirectory)/
+  - task: PublishPipelineArtifact@0
+      condition: succeededOrFailed()
+      inputs:
+        targetPath: bazel-bin/compiler/damlc/tests/platform-independence.dar
+        artifactName: platform-independence-dar-linux
```
garyverhaegen-da, line 21:
```suggestion
```

ci/platform-independence-windows.yml
```diff
@@ -0,0 +1,16 @@
...
+    env:
+      IS_FORK: $(System.PullRequest.IsFork)
+      # to upload to the bazel cache
+      GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+  - powershell: .\ci\platform-independence-dar.ps1
+    displayName: 'Build platform independence test dar'
+  - task: PublishPipelineArtifact@0
+    inputs:
+      targetPath: $(Build.StagingDirectory)/platform-independence.dalf
+      artifactName: "platform-independence-dar"
```
garyverhaegen-da, line 16:
```suggestion
      artifactName: "platform-independence-dalf"
```

robin-da: fixed.

ci/platform-independence.yml
```diff
@@ -0,0 +1,22 @@
...
+      # to upload to the bazel cache
+      GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+  - task: DownloadPipelineArtifact@0
+    inputs:
+      artifactName: 'platform-independence-dalf'
+      targetPath: $(Build.StagingDirectory)/
+  - bash: |
+      set -euo pipefail
+      eval "$(dev-env/bin/dade-assist)"
+      bazel build //compiler/damlc/tests:platform-independence.dalf
```
cocreature, line 20: can we check the full DAR rather than just this DALF? If those are identical the dalfs are clearly identical as well and our DARs should be built deterministically (if not we should probably fix that)so I think there shouldn’t be extra-nondeterminism introduced here.

ci/cron/daily-compat.yml
```diff
@@ -67,6 +67,26 @@ jobs:
...
+  - job: platform_independence_windows
+    timeoutInMinutes: 60
+    pool:
+      name: windows-pool
+      demands: assignment -equals default
+    steps:
+      - checkout: self
+      - template: ../platform-independence-windows.yml
+
+  - job: platform_independence
```
cocreature, line 79: probably a good idea to include macos as well. Not that I see a reason why that should be different but easy enough to add and better to check it than to be surprised later.

cocreature (approved):
Nice work!

ci/platform-independence.yml
garyverhaegen-da: The DARs are not identical: they include source code for a test file that differs on line endings, as well a .hie files that somehow differ too ("binary files differ", as `diff` puts it; I did not investigate further).

ci/cron/daily-compat.yml
garyverhaegen-da:
```suggestion
  - job: platform_independence
    strategy:
      matrix:
        linux:
          pool: ubuntu_20_04
        macos:
          pool: macOS-pool
```

garyverhaegen-da:
This may be a bit of a dumb/way-too-late question, but, is there any reason why this is a daily run rather than part of every PR? Adding a publish step to the existing build-everything Windows build shouldn't be that much of an overhead, and the checks on macOS and Linux will run in parallel and be fairly short.

cocreature:
> This may be a bit of a dumb/way-too-late question, but, is there any reason why this is a daily run rather than part of every PR? Adding a publish step to the existing build-everything Windows build shouldn't be that much of an overhead, and the checks on macOS and Linux will run in parallel and be fairly short.

I don’t really like the idea of adding something that is strictly sequential across to slow down builds even more. Yes assuming caching works out it’s probably only 2 minutes or so but  that’s still non-neglible and it can be a lot worse if caching fails.
Given how rarely we change the code that could affect this, the daily run seemed like a better option.

garyverhaegen-da:
> > This may be a bit of a dumb/way-too-late question, but, is there any reason why this is a daily run rather than part of every PR? Adding a publish step to the existing build-everything Windows build shouldn't be that much of an overhead, and the checks on macOS and Linux will run in parallel and be fairly short.
> 
> I don’t really like the idea of adding something that is strictly sequential across to slow down builds even more. Yes assuming caching works out it’s probably only 2 minutes or so but that’s still non-neglible and it can be a lot worse if caching fails.
> Given how rarely we change the code that could affect this, the daily run seemed like a better option.

If you're worried about caching you could instead have all three builds publish their result, and then have a separate job that literally just downloads the three DARs and calls diff. That should be super fast: DARs are only ~240kb each, `diff` is present by default with no dev-env activation needed.

Would make this a lot easier to test and avoid the big blob of duplicated PowerShell code nobody is going to keep in sync.

ci/cron/daily-compat.yml
```diff
@@ -324,6 +352,8 @@ jobs:
       compatibility_ts_libs: $[ dependencies.compatibility_ts_libs.result ]
       compatibility: $[ dependencies.compatibility.result ]
       compatibility_windows: $[ dependencies.compatibility_windows.result ]
+      platform_independence_windows: $[ dependencies.platform_independence-windows.result ]
```
garyverhaegen-da, line 355:
```suggestion
      platform_independence_windows: $[ dependencies.platform_independence_windows.result ]
```

PR #10536, opened 2021-08-10
Handle not quite empty directory digests in cache cleanup

cocreature:
Dropping the strict source name patch has resulted in empty
directories which Bazel is happy about but then fail at
runtime. They’re not quite empty, the digest points to 0x0a 0x00
so we match on that.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/src/BazelCache.hs
+++ b/ci/cron/src/BazelCache.hs
@@ -16,7 +16,7 @@ module BazelCache
     , run
     ) where
 
-import Build.Bazel.Remote.Execution.V2.RemoteExecution (ActionResult(..), Digest(..))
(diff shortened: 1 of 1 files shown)


ci/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         export BOTO_CONFIG=/dev/null
           # 90 minutes should provide enough overlap for an hourly
         # cronjob.
-        ./bazel-bin/ci/cron/cron bazel-cache --age 90 --delete
+        ./bazel-bin/ci/cron/cron bazel-cache --age 3000 --delete
```
cocreature, line 33: will revert before merging

aherrmann-da (approved):

PR #10537, opened 2021-08-10
kvutils: Improve the test coverage of OffsetBuilder, and support negative numbers.

SamirTalwar-DA:
I noticed the test coverage of `OffsetBuilder` was lacking so I added some.

In practice, there's no reason to support negative numbers (and it's probably a bad idea), but we should at least return a valid offset. Prior to these changes, a negative value passed to `OffsetBuilder.fromLong` would result in strange behavior.

I think `DataOutputStream` is probably also safer than clever bit arithmetic and `BigInt`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
@@ -3,6 +3,8 @@
 
 package com.daml.ledger.participant.state.kvutils
 
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
+
 import com.daml.ledger.offset.Offset
 
 /** Helper functions for generating 16 byte [[Offset]]s from integers.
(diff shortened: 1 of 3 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -24,44 +26,50 @@ object OffsetBuilder {
...
-    Offset.fromByteArray(bytes)
+    val bytes = new ByteArrayOutputStream
+    val stream = new DataOutputStream(bytes)
+    stream.writeLong(first)
+    stream.writeInt(second)
+    stream.writeInt(third)
+    Offset.fromByteArray(bytes.toByteArray)
+  }
+
+  // Doesn't call `split` because it's used a lot, so it's worth optimizing a little.
```
fabiotudone-da, line 58:
```suggestion
  // `highestIndex` is used a lot, so it's worth optimizing a little rather than reusing `split`.
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilderSpec.scala
```diff
@@ -4,46 +4,73 @@
 package com.daml.ledger.participant.state.kvutils
 
 import com.daml.ledger.offset.Offset
-import com.daml.lf.data
+import com.daml.lf.data.Bytes
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
```
fabiotudone-da, line 10: How much slower does it make the test?

SamirTalwar-DA: On my machine the test file runs in ~100ms.

fabiotudone-da: How long did it take before?

fabiotudone-da (approved):

hubert-da (commented):
lgtm

SamirTalwar-DA:
Hmmm, the PR has passed the build but GitHub doesn't seem to have picked that up.

Gonna rerun.

/azp run

PR #10538, opened 2021-08-10
Fix typo in windows extra libs patch

aherrmann-da:
This slipped in in https://github.com/digital-asset/daml/pull/10517/files

### Pull Request Checklist

trigger the build.


cocreature (approved):
looks good, why did this not break anything?

aherrmann-da:
> looks good, why did this not break anything?

I was surprised as well. I'm not 100% sure, but I think two things prevented this from causing any issues:
1. Many of our C library dependencies are through `haskell_cabal_library` targets which are unaffected by this.
2. rules_haskell generates package configs for linking which capture all those extra libraries as well (see [`create_link_config`](https://github.com/tweag/rules_haskell/pull/930/files#diff-b5bc2e06ac9cee9f7988c553089e6658494e67235e3be26d7b4bd16e2472258cR282) - the motivation is to avoid long command lines and back then mostly static linking).

PR #10552, opened 2021-08-11
Ledger API test tool: test that record time is monotonically increasing [KVL-1053]

fabiotudone-da:
CHANGELOG_BEGIN
- [Ledger API Specification] Introduced a new conformance test about record time in completions being monotonically increasing

CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
```diff
@@ -0,0 +1,48 @@
...
+          case (false, _) -> _ => (false, None)
+          case (true, None) -> firstTimestamp => (true, Some(firstTimestamp))
+          case (true, Some(previousTimestamp)) -> newTimestamp =>
+            TimestampConverters
+              .asJavaInstant(previousTimestamp)
+              .isBefore(TimestampConverters.asJavaInstant(newTimestamp)) ->
+              Some(newTimestamp)
+        }
+        ._1
+      assert(monotonicallyIncreasing, "record time is not monotonically increasing")
```
SamirTalwar-DA, line 45: Can you include the record times in this error message (preferably converted to something readable)?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
```diff
@@ -0,0 +1,48 @@
...
+      _ <- Future.traverse(1 to operations) { number =>
+        ledger.allocateParty(
+          partyIdHint = Some(
+            s"recordTimeMonotonicallyIncreasing_${number}_" + Random.alphanumeric.take(100).mkString
+          ),
+          displayName = Some(s"Clone $number"),
+        )
+      }
+      checkpoints <- ledger.checkpoints(operations, ledger.begin)()
+      recordTimes = checkpoints.flatMap(_.recordTime)
```
SamirTalwar-DA, line 32: This can be in the `yield` block.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
```diff
@@ -0,0 +1,48 @@
...
+        .foldLeft((true, None: Option[ProtoTimestamp])) {
+          case (false, _) -> _ => (false, None)
+          case (true, None) -> firstTimestamp => (true, Some(firstTimestamp))
+          case (true, Some(previousTimestamp)) -> newTimestamp =>
+            TimestampConverters
+              .asJavaInstant(previousTimestamp)
+              .isBefore(TimestampConverters.asJavaInstant(newTimestamp)) ->
+              Some(newTimestamp)
+        }
+        ._1
```
SamirTalwar-DA, line 44: You may be able to simplify this to:

```suggestion
      val recordTimes = checkpoints
        .flatMap(_.recordTime)
        .collect { case Some(time) => time } // `.flatten` might also work
        .map(TimestampConverters.asJavaInstant)
      val monotonicallyIncreasing = recordTimes.sorted == recordTimes
```
Given that we're dealing with 50 elements, sorting doesn't seem unreasonable.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
fabiotudone-da: Yes but with an additional and not very useful `val` prefix

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
fabiotudone-da: Not really sure how many we should try really

fabiotudone-da: f21292e

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
fabiotudone-da: f21292e

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/RecordTimeIT.scala
fabiotudone-da: f21292e but need to double-check

fabiotudone-da: b8b56df should be better

SamirTalwar-DA (approved):

fabiotudone-da:
@moritzkiefer-da This fails in `NonRepudiationProxyConformance` with e.g.:

```
some record times are not monotonically increasing: Vector(2021-08-11T13:29:08.905Z, 2021-08-11T13:29:08.896Z)
```

is that somewhat expected, or am I wrongly assuming that checkpoints will be streamed in offset order, or do you have any other clues why?

fabiotudone-da:
I am puzzled by the flakiness of this test. Basically, it ensures there are enough checkpoints to fetch later and checks that the record time in them increases monotonically, after sorting them by offset. @SamirTalwar-DA do you see any obvious flaws in it?

cocreature:
> I am puzzled by the flakiness of this test. Basically, it ensures there are enough checkpoints to fetch later and checks that the record time in them increases monotonically, after sorting them by offset. @SamirTalwar-DA do you see any obvious flaws in it?

Did you confirm that it’s generally flaky or only on the non-repudiation stuff? While I don’t understand it, I’m definitely not excluding the possibility that this is buggy

fabiotudone-da:
> > I am puzzled by the flakiness of this test. Basically, it ensures there are enough checkpoints to fetch later and checks that the record time in them increases monotonically, after sorting them by offset. @SamirTalwar-DA do you see any obvious flaws in it?
> 
> Did you confirm that it’s generally flaky or only on the non-repudiation stuff? While I don’t understand it, I’m definitely not excluding the possibility that this is buggy

It's flaky, in the sense that it doesn't always fail with the same ledger, although I don't know how generally. So far I have seen it failing with SQLs.

cocreature:
alright so at least nothing for me to worry about for now :)

fabiotudone-da:
I'd like to merge this test as optional (i.e., disabled by default), so that it can be run (only) on ledgers that actually support this property.

rautenrieth-da (commented):
Unlike pruning and multi-party submissions, monotonic record times are unlikely to ever be a required feature of the ledger API. I'm slightly concerned about adding ledger-specific tests to the ledger-agnostic test tool.

Can we at least make sure it's well documented what this test does? Maybe rename the test, and add some scaladoc comments explaining that it is only meant to be run on a certain type of ledgers.

fabiotudone-da:
> Unlike pruning and multi-party submissions, monotonic record times are unlikely to ever be a required feature of the ledger API. I'm slightly concerned about adding ledger-specific tests to the ledger-agnostic test tool.

I agree in principle; I'll transition it back to draft and think of other approaches that respect the ledger-agnostic / ledger-specific divide while still being practical.

miklos-da (approved):

hubert-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Tests.scala
```diff
@@ -53,6 +53,7 @@ object Tests {
     Vector(
       new ParticipantPruningIT,
       new MultiPartySubmissionIT,
+      new RecordTimeIT,
```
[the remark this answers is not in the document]
fabiotudone-da: Done

PR #10554, opened 2021-08-11
add cryptographic hash method to ChangeId and base equality on hash

andreaslochbihler-da:
Add a cryptographic hash method to `ChangeId`.

* Equality of `ChangeId` is based on the hash so that we don't have to worry about hash collisions when specifying the deduplication guarantee.
* Made the fields public so that other code can do something with a `ChangeId`.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
@@ -232,6 +232,7 @@ object Hash {
     val MaintainerContractKeyUUID = Purpose(4)
     val PrivateKey = Purpose(3)
     val ContractInstance = Purpose(5)
+    val ChangeId = Purpose(6)
   }
 
   // package private for testing purpose.
@@ -332,6 +333,17 @@ object Hash {
       arg: Value[Value.ContractId],
   ): Either[String, Hash] =
     handleError(assertHashContractInstance(templateId, arg))
+
+  def hashChangeId(
+      applicationId: Ref.ApplicationId,
+      commandId: Ref.CommandId,
+      actAs: Set[Ref.Party],
+  ): Hash =
+    builder(Purpose.ChangeId, noCid2String)
+      .add(applicationId)
+      .add(commandId)
+      .addStringSet(actAs)
+      .build
 
   def deriveSubmissionSeed(
       nonce: Hash,
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
@@ -223,7 +223,7 @@ object Hash {
   }
 
   // The purpose of a hash serves to avoid hash collisions due to equal encodings for different objects.
-  // Each purpose should be used at most one.
+  // Each purpose should be used at most once.
   private[crypto] case class Purpose(id: Byte)
 
   private[crypto] object Purpose {
--- a/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
+++ b/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
@@ -6,7 +6,7 @@ package com.daml.ledger.participant.state.v2
 import com.daml.lf.crypto.Hash
 import com.daml.lf.data.Ref
 
-/** Identifier for ledger changes used by command deduplication
+/** Identifier for ledger changes used by command deduplication.
   * Equality is defined in terms of the cryptographic hash.
   *
   * @see ReadService.stateUpdates for the command deduplication guarantee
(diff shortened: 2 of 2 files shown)


ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
```diff
@@ -3,14 +3,45 @@
...
 /** Identifier for ledger changes used by command deduplication
   *
   * @see ReadService.stateUpdates for the command deduplication guarantee
   */
-final case class ChangeId(
-    private val applicationId: Ref.ApplicationId,
-    private val commandId: Ref.CommandId,
-    private val actAs: Set[Ref.Party],
-)
+final class ChangeId(
```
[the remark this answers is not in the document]
andreaslochbihler-da: Because equality is based on the hash, not on the actual arguments to the constructor. It's similar to `GlobalKey`.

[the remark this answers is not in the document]
andreaslochbihler-da: This class is similar to `GlobalKey`, which isn't a case class either. When I see a `case class`, I expect equality to be defined in terms of the constructor arguments.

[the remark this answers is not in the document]
cocreature: I think the situation is a bit different. For GlobalKey the constructor is private and the hash is an argument to the constructor. Here you derive the hash in the constructor and the constructor is public.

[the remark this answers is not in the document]
andreaslochbihler-da: Can you elaborate why this difference matters here? `GlobalKey` has a public `apply` method which is as good as a constructor, isn't it?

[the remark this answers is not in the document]
andreaslochbihler-da: > provided you don’t assume your hash is broken.

That's precisely why I define equality using the hash, namely so that hash collisions are accounted for. But I won't fight over this. I'll turn it into a case class.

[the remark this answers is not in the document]
cocreature: Fine to keep it as well, I mostly wanted to understand the motivation.

cocreature (approved):
Thanks!

nicu-da (approved):

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
```diff
@@ -3,14 +3,31 @@
 
 package com.daml.ledger.participant.state.v2
 
+import com.daml.lf.crypto.Hash
 import com.daml.lf.data.Ref
 
 /** Identifier for ledger changes used by command deduplication
```
fabiotudone-da, line 9:
```suggestion
/** Identifier for ledger changes used by command deduplication.
```

fabiotudone-da: (Not really about this PR but since you're here)

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
```diff
@@ -232,6 +232,7 @@ object Hash {
     val MaintainerContractKeyUUID = Purpose(4)
```
fabiotudone-da, line 232: Also not really about this PR, but since you're here could you fix the typo above "one" -> "once"?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
```diff
@@ -3,14 +3,31 @@
...
-)
+    applicationId: Ref.ApplicationId,
+    commandId: Ref.CommandId,
+    actAs: Set[Ref.Party],
+) {
+
+  /** A stable hash of the change id.
+    * Suitable for storing in persistent storage.
+    */
+  val hash: Hash = Hash.hashChangeId(applicationId, commandId, actAs)
```
fabiotudone-da, line 23: I guess it's rather unimportant in practice but:

```suggestion
  lazy val hash: Hash = Hash.hashChangeId(applicationId, commandId, actAs)
```

fabiotudone-da (approved):

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
andreaslochbihler-da: Changed.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala
andreaslochbihler-da: Changed.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/crypto/Hash.scala
andreaslochbihler-da: Done.

PR #10555, opened 2021-08-11
fix hoogle

garyverhaegen-da:
Hoogle was down because the machines got stuck in the `apt-get update` stage.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

PR #10562, opened 2021-08-11
ledger-api-test-tool: Enforce test naming standards.

SamirTalwar-DA:
This changes a bunch of test names so that:

- test names are not prefixes of other test names, leading to accidental inclusion or exclusion, and
- test names are unique.

This is now enforced through unit tests.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
+++ b/ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
@@ -1,0 +1,31 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.testtool
+
+import com.daml.ledger.api.testtool.NamesSpec._
+import com.daml.ledger.api.testtool.tests.Tests
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class NamesSpec extends AnyWordSpec with Matchers {
+  "test suite names" should {
+    "only contain letters" in {
+      all(allTestSuiteNames) should fullyMatch regex """[A-Za-z]+""".r
+    }
+  }
+
+  "test identifiers" should {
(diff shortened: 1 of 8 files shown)


ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+        all(allTestIdentifiers.toSet - testIdentifier) should not startWith testIdentifier
+      }
+    }
+  }
+
+  "full test names" should {
+    "be unique" in {
+      allTestNames.foreach { name =>
+        allTestNames.filter(_ == name) should have size 1
+      }
```
nicu-da, line 40: Wouldn't it be simpler to convert `allTestNames` to a set and test for equal length?

nicu-da (approved):

ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
SamirTalwar-DA: I tried that but the error message was useless. "expected <some giant list> to have size 999 but was 972", or something. And an equality check resulted in "<some giant list> does not equal <some other giant list>".

This yields an error message that you can actually use to find the problem.

nicu-da: Ah yes, makes sense. Not sure if it's worth collecting all the duplicate tests though before failing, as I wouldn't expect to get a lot of duplicates.

cocreature: we already have a diffing lib in the legder API test tool, you could use that if you want but I’m also perfectly happy with this solution.

SamirTalwar-DA: Rule number one of writing tests: make sure it fails.

Rule number two: make sure the failure tells you something useful.

PR #10569, opened 2021-08-12
Bump bazel max jvm memory

cocreature:
I kept the starting memory the same. Happy to bump that as well if you
think it makes sense.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


aherrmann-da (approved):
LGTM

SamirTalwar-DA (approved):
Looks good. I suggest setting the starting heap size to the same as the max size because if it does cause problems, it'll probably cause them more consistently.

cocreature:
> Looks good. I suggest setting the starting heap size to the same as the max size because if it does cause problems, it'll probably cause them more consistently.

done

PR #10576, opened 2021-08-12
Upgrade Scalatest to v3.2.9.

SamirTalwar-DA:
Because of some coupling we also have to upgrade Scalaz to the latest v7.2 point release, v7.2.33.

The Scalatest changes are quite involved because the JAR has been broken up into several smaller JARs. Because Bazel expects us to specify all dependencies and doesn't allow transitive dependencies to be used directly, this means that we need to specify the explicit Scalatest components that we use.

As you can imagine, this results in quite a big set of changes. They are, however, constrained to dependency management; all the code remains the same.

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks! seems like a very annoying split but still better than staying on the old version

S11001001 (approved):

PR #10577, opened 2021-08-12
DPP-535 Verify postgres version

rautenrieth-da:
At startup, complain if the Postgres version is too old

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
@@ -112,12 +112,12 @@ private[backend] object PostgresStorageBackend
 
   def getPostgresVersion(
       connection: Connection
-  )(implicit loggingContext: LoggingContext): Option[(Int, Int, Int)] = {
+  )(implicit loggingContext: LoggingContext): Option[(Int, Int)] = {
     val version = SQL"SHOW server_version".as(get[String](1).single)(connection)
     logger.debug(s"Found Postgres version $version")
-    val versionPattern = """(\d)+[.](\d)+[.](\d)+""".r
+    val versionPattern = """(\d)+[.](\d)+.*""".r
     version match {
-      case versionPattern(major, minor, patch) => Some((major.toInt, minor.toInt, patch.toInt))
+      case versionPattern(major, minor) => Some((major.toInt, minor.toInt))
       case _ => None
     }
   }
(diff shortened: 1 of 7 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -86,6 +88,37 @@ private[backend] object PostgresStorageBackend
...
+  def getPostgresVersion(connection: Connection): Option[(Int, Int, Int)] = {
+    val version = SQL"SHOW server_version".as(get[String](1).single)(connection)
+    val versionPattern = """(\d)[.](\d)[.](\d)""".r
+    version match {
+      case versionPattern(major, minor, patch) => Some(major.toInt, minor.toInt, patch.toInt)
+      case _ => None
+    }
+  }
+
+  override def checkCompatibility(
```
nmarton-da, line 100: since this feature seems to be only about logging, why not just add this to the main entrypoint of connection creation: DataSourceStorageBackend.createDataSource, and only where we need it?
This way we could avoid increasing StorageBackend and LedgerDao surface.

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -86,6 +88,37 @@ private[backend] object PostgresStorageBackend
...
+            "Deprecated Postgres version. " +
+              s"Found Postgres version $major.$minor.$patch., minimum required Postgres version is 10. " +
+              "This application will continue running but is at risk of data loss, as Postgres < 10 does not support crash-fault tolerant hash indices. " +
+              "Please upgrade your Postgres database to version 10 or later to fix this issue. " +
+              "In the future, this deprecation warning may be upgraded to a fatal error."
+          )
+        }
+      case None =>
+        logger.warn(
+          s"Could not determine the version of the Postgres database. Please verify that this application is compatible with this Postgres version."
```
nmarton-da, line 116: if we fail with an error if we detect a wrong version, then perhaps we should fail with an error if version cannot be read?
BTW where are the requirements coming from wtr? (why error log in contrast to error log and fail?)

rautenrieth-da: This PR only "deprecates" old Postgres versions, it doesn't yet force users to upgrade. Later we can change both log outputs to throwing an error. which would force users to upgrade.

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
rautenrieth-da: The version is now checked within `createDataSource`.

PR #10579, opened 2021-08-14
Adding Ord instance for NonEmpty

hgrano:
As discussed in https://discuss.daml.com/t/no-ord-instance-for-nonempty/2964

CHANGELOG_BEGIN

- [Daml Standard Library] Instance of the Ord typeclass added for the NonEmpty data type.

CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
@@ -39,6 +39,7 @@ import DA.Action qualified as M
 
 deriving instance Eq a => Eq (NonEmpty a)
 deriving instance Show a => Show (NonEmpty a)
+deriving instance Ord a => Ord (NonEmpty a)
 
 -- NonEmpty is defined in a stable LF package so we need to handwrite the HasField instances here.
 
--- a/compiler/damlc/tests/daml-test-files/NonEmpty.daml
+++ b/compiler/damlc/tests/daml-test-files/NonEmpty.daml
@@ -13,3 +13,12 @@ testApplicative = scenario do
   (fs <*> l) === NonEmpty 2 [4, 6, 8, 3, 6, 9, 12]
   hd l === 1
   tl l === [2, 3, 4]
+
+testOrd = scenario do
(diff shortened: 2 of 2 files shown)


compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
```diff
@@ -40,6 +40,9 @@ import DA.Action qualified as M
 deriving instance Eq a => Eq (NonEmpty a)
 deriving instance Show a => Show (NonEmpty a)
 
+instance (Eq a, Ord a) => Ord (NonEmpty a) where
+  compare x y = compare (toList x) (toList y)
```
cocreature, line 44:
```suggestion
deriving instance Ord a => Ord (NonEmpty a)
```
We can just derive this instance similar to how we handle Eq and Show above.

Two small comments on your hand-written instance:
1. Eq is a superclass of Ord so the constraint is redundant.
2. `compare (toList x) (toList y)` is somewhat inefficient. There is no fusion or laziness in Daml so this has to build two new lists before comparing them. While that’s not super expensive it’s still unnecessary. The efficient way to write this is something like
```
compare x y = case compare (hd x) (hd y) of
  EQ -> compare (tl x) (tl y)
  LT -> LT
  GT -> GT
```
Luckily that’s exactly what the autoderived instance does so no need to write this ourselves.

hgrano: Thanks for the suggestion. My thinking was to keep the Ord instance behaviour always the same as the behaviour for List, but because of the way NonEmpty is implemented the derived instance will do the same and more efficiently.

cocreature (approved):
Thanks!

cocreature:
CI is failing due to the missing changelog entry. See https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md#pull-request-checklist for details.
You can push a new commit or amend the existing commit with the commit message containing something like this:
```
changelog_begin
- [Daml Stdlib] Add Ord instance for NonEmpty
changelog_end
```

PR #10585, opened 2021-08-16
ledger-api-test-tool: Split TransactionServiceIT into lots of suites.

SamirTalwar-DA:
I wanted to add another test, and there's no way I'm going to be responsible for making this class bigger.

### Changelog

- **[Ledger API Test Tool]** The `TransactionServiceIT` test suite has been split into many test suites. If you are including or excluding it, you will need to use the new test suite names, or you can use "TransactionService" as a prefix for all of them.

  If you are including or excluding individual tests, you will need to update your arguments with the new test suite. You can find the new test suite by running the test tool with the `--list-all` flag and looking for the test's short identifier. The short identifiers have not changed, with the exception of `TXNoContractKey`, which has been renamed to `CKNoContractKey` and is now in the `ContractKeysIT` test suite.

### Pull Request Checklist

trigger the build.


nicu-da (approved):
LGTM!
Two small things:
- do we support moving all this new suites into a package of it's own? Or do they have to be all part of the root 
- In the changelog, would it be useful to list the individual suites that we split the TransactionServiceIT?

stefanobaghino-da (commented):
I like the idea of splitting the suite in smaller files.

Have you considered splitting it in traits and bundling them together in a single `TransactionServiceIT` so that there is no impact for users?

stefanobaghino-da:
> The short identifiers have not changed, with the exception of `TXNoContractKey`, which has been renamed to `CKNoContractKey` and is now in the `ContractKeysIT` test suite.

I understand the reasoning behind this, but if we have some shuffling to do I think it should be a concerted effort, rather than a one-off thing. It's also probably better left out as a separate PR, especially if we find a way to not impact users.

> In the changelog, would it be useful to list the individual suites that we split the TransactionServiceIT?

If there's no way we still expose the `TransactionServiceIT` as a single unit, I think it would be a nice addition to ease the transition for users.

SamirTalwar-DA:
I'm happy to split and re-bundle these so there's no impact, but TBH I think as a consumer of the ledger API test tool, I also prefer having a bit more context in the suite name. This might just be a bias I have, though.

I'm very happy to extend the changelog to list all the new suite names.

Do you think I should rename `CKNoContractKey` back to `TXNoContractKey` (but still keep it in `ContractKeysIT`)?

SamirTalwar-DA:
I've verified with all known users of the Ledger API Test Tool that this has very little or no impact, so I think I'm going to just merge it.

I'll wait a couple more hours in case anyone wants to stop me.

miklos-da (approved):

PR #10588, opened 2021-08-16
daml ledger export: export all parties

### LINKED ISSUE #10436: daml ledger export: feature request - export ALL parties
Working with the ledger export tool, I would like to be able to export the entirety of the ledger contents, without having to enumerate the relevant parties myself.

It's possible to add this feature from the outside, using shellscripts:

```
daml ledger export script --host localhost --port 6865 --sdk-version 0.0.0 --output out/test --party $(daml ledger list-parties --host localhost --port 6865 --json | jq '.[] | .party' | awk '{printf "%s%s", NR==1?"":",", $1;'} | tr -d '"' | tr -d '\r')
```

I think it's enough of a common use case that it's fair to request it in the tool proper.

Could there be an `--allParties` flag added?

aherrmann-da:
Closes #10436

Adds an `--all-parties` flag to the Daml ledger export command. If set the tool will query all known parties from the ledger and generated an export of the ledger state as seen by these parties.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/export/src/main/scala/com/daml/script/export/Config.scala
+++ b/daml-script/export/src/main/scala/com/daml/script/export/Config.scala
@@ -7,6 +7,7 @@ import java.nio.file.{Path, Paths}
 import java.io.File
 
 import com.daml.auth.TokenHolder
+import com.daml.ledger.api.refinements.ApiTypes.Party
 import com.daml.ledger.api.tls.{TlsConfiguration, TlsConfigurationCli}
 import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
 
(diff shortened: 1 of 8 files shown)


daml-script/export/src/main/scala/com/daml/script/export/Config.scala
```diff
@@ -30,6 +31,13 @@ final case class ExportScript(
     damlScriptLib: String,
 ) extends ExportType
 
+// This is a product rather than a sum, so that we can raise
```
cocreature, line 34: we could consider using that only during config parsing and switching to a different type afterwards but probably not worth the complexity.

cocreature (approved):
Thanks!

daml-script/export/src/main/scala/com/daml/script/export/Config.scala
aherrmann-da: Agreed. I wish scopt had something like optparse-applicative's `Alternative` instance.

PR #10590, opened 2021-08-16
Move DeduplicationPeriod to ledger-api-domain [KVL-1047]

nicu-da:
- To enable use the reuse of `DeduplicationPeriod`, so it can be passed down from the API to the submitter info
- I'm not at all convinced about this move, please feel free to propose another option

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
+++ b/ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
@@ -3,7 +3,7 @@
 
 package com.daml.ledger.api
 
-import java.time.{Duration, Instant}
+import java.time.Duration
 
 import com.daml.ledger.offset.Offset
 import com.daml.logging.entries.{LoggingValue, ToLoggingValue}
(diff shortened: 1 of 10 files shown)


SamirTalwar-DA (approved):
Seems sensible to me.

fabiotudone-da (commented):
Agree with @SamirTalwar-DA, perhaps let's wait for a feedback from @digital-asset/kv-participant just to make sure there aren't caveats we are overlooking.

tudor-da (approved):
LGTM, although I'd appreciate @rautenrieth-da 's feedback on this one as well.

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package com.daml.ledger.participant.state.v2
+package com.daml.ledger.api
```
tudor-da, line 4: Where is this going to be used as part of the API. If it is only as part of command submission, I am wondering whether this shouldn't be in `com.daml.ledger.api.messages.command.completion`. l

nicu-da: Yes, so in the API it's gonna be part of `com.daml.ledger.api.Commands`. i'm up for moving it anywhere as I am not really familiar with the structure we have here

miklos-da (approved):
Looks good to me, i.e., deduplication period is not only relevant to participant state but is a more general concept used throughout the stack (both by the participant and the committer).

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
rautenrieth-da: `com.daml.ledger.api.messages.*` seems to only contain types corresponding to top-level messages used in the ledger API.

rautenrieth-da (approved):
The move looks sensible to me.

PR #10600, opened 2021-08-17
participant-integration-api: Store a status gRPC protobuf. [KVL-1005]

SamirTalwar-DA:
Instead of storing the status code and message, we store a serialized `google.rpc.Status` protocol buffers message. This allows us to pass through any additional information reported by the driver `ReadService`.

The migration is only done for the append-only index, and preserves old data in the existing columns. New data will only be written to the new column. We'll make sure this is OK by only supporting the append-only index with the v2 `ReadService`.

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
```diff
@@ -124,16 +124,17 @@ CREATE INDEX idx_party_entries ON party_entries(submission_id);
...
+    submitters                  CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
+    command_id                  NVARCHAR2(1000) NOT NULL,
 
-    transaction_id    NVARCHAR2(1000), -- null if the command was rejected and checkpoints
-    status_code       INTEGER,         -- null for successful command and checkpoints
-    status_message    CLOB  -- null for successful command and checkpoints
+    transaction_id              NVARCHAR2(1000), -- null if the command was rejected and checkpoints
+    rejection_status_code       INTEGER,         -- null for successful command and checkpoints
+    rejection_status_message    CLOB,            -- null for successful command and checkpoints
+    rejection_status            BLOB             -- null for successful command and checkpoints
```
SamirTalwar-DA, line 137: This assumes that we don't currently support migrations for Oracle. Let me know if I'm wrong, @dasormeter.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -10,60 +10,42 @@ import com.daml.ledger.api.v1.command_completion_service.{Checkpoint, Completion
...
+      checkpoint = Some(toApiCheckpoint(recordTime, offset)),
+      completions = Seq(Completion.of(commandId, Some(OkStatus), transactionId)),
     )
 
-  // Filter completions for transactions for which we have the full submitter information: appId, submitter, cmdId
-  // This doesn't make a difference for the sandbox (because it represents the ledger backend + api server in single package).
-  // But for an api server that is part of a distributed ledger network, we might see
-  // transactions that originated from some other api server. These transactions don't contain the submitter information,
-  // and therefore we don't emit CommandAccepted completions for those
-  def apply(
```
SamirTalwar-DA, line 37: This function was inlined into `InMemoryLedger` in Sandbox Classic, as it's only used there.

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V31__rejection_status_proto.sql
```diff
@@ -0,0 +1,16 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+-- Store the completion status as a serialized Protocol Buffers message of type `google.rpc.Status`.
+
+-- The `rejection_status` contains a Protocol-Buffers-serialized message of type
+-- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+-- driver). The `rejection_status_code` and `rejection_status_message` columns may contain old
+-- data and will always be `NULL` in a H2 ledger.
```
fabiotudone-da, line 9: > ...will always be `NULL` in a H2 ledger.

Why is that?

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
nicu-da, line 137: I assume this is not deployed anywhere yet

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,16 @@ CREATE TABLE participant_command_completions (
     application_id VARCHAR NOT NULL,
     submitters ARRAY NOT NULL,
     command_id VARCHAR NOT NULL,
+    -- The transaction ID is `NULL` for rejected transactions.
     transaction_id VARCHAR,
-    status_code INTEGER,
-    status_message VARCHAR
+    -- The rejection status is `NULL` if the completion is for an accepted transaction.
```
fabiotudone-da, line 126:
```suggestion
    -- The three following rejection status code columns are `NULL` if the completion is for an accepted transaction.
```

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,16 @@ CREATE TABLE participant_command_completions (
...
+    -- The `rejection_status` contains a Protocol-Buffers-serialized message of type
```
fabiotudone-da, line 127:
```suggestion
    -- The rejection status contains a Protocol-Buffers-serialized message of type
```

(Or citing the column name everywhere for uniformity)

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V106__add_rejection_status_proto_column.sql
```diff
@@ -0,0 +1,16 @@
...
+-- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+-- driver). We do not migrate the `rejection_status_code` and `rejection_status_message` columns,
+-- and so they may contain old data, which means we need to keep them.
+
+ALTER TABLE participant_command_completions
+    RENAME COLUMN status_code TO rejection_status_code;
+ALTER TABLE participant_command_completions
+    RENAME COLUMN status_message TO rejection_status_message;
+ALTER TABLE participant_command_completions
+    ADD COLUMN rejection_status BYTEA;
```
nicu-da, line 16: Do we have any HA deployments that might be affected by this?

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,16 @@ CREATE TABLE participant_command_completions (
...
+    -- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+    -- driver). The `rejection_status_code` and `rejection_status_message` columns will always be
+    -- `NULL` in an H2-backed index, but we keep them for parity with old data in other databases.
```
fabiotudone-da, line 130: Why is that?

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
```diff
@@ -124,16 +124,17 @@ CREATE INDEX idx_party_entries ON party_entries(submission_id);
...
-    submitters        CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
-    command_id        NVARCHAR2(1000) not null,
+    application_id              NVARCHAR2(1000) NOT NULL,
+    submitters                  CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
+    command_id                  NVARCHAR2(1000) NOT NULL,
 
-    transaction_id    NVARCHAR2(1000), -- null if the command was rejected and checkpoints
-    status_code       INTEGER,         -- null for successful command and checkpoints
-    status_message    CLOB  -- null for successful command and checkpoints
+    transaction_id              NVARCHAR2(1000), -- null if the command was rejected and checkpoints
```
fabiotudone-da, line 134:
```suggestion
    transaction_id              NVARCHAR2(1000), -- null if the command was rejected and for checkpoints
```

ledger/participant-integration-api/src/main/resources/db/migration/oracle/V1__Init.sql
```diff
@@ -176,16 +176,17 @@ CREATE UNIQUE INDEX idx_package_entries
...
-    submitters        CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
-    command_id        NVARCHAR2(1000) not null,
+    application_id              NVARCHAR2(1000) NOT NULL,
+    submitters                  CLOB NOT NULL CONSTRAINT ensure_json_submitters CHECK (submitters IS JSON),
+    command_id                  NVARCHAR2(1000) NOT NULL,
 
-    transaction_id    NVARCHAR2(1000), -- null if the command was rejected and checkpoints
-    status_code       INTEGER,         -- null for successful command and checkpoints
-    status_message    CLOB  -- null for successful command and checkpoints
+    transaction_id              NVARCHAR2(1000), -- null if the command was rejected and checkpoints
```
fabiotudone-da, line 186:
```suggestion
    transaction_id              NVARCHAR2(1000), -- null if the command was rejected and for checkpoints
```

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V106__add_rejection_status_proto_column.sql
```diff
@@ -0,0 +1,16 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+-- Store the completion status as a serialized Protocol Buffers message of type `google.rpc.Status`.
+
+-- The `rejection_status` contains a Protocol-Buffers-serialized message of type
+-- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+-- driver). We do not migrate the `rejection_status_code` and `rejection_status_message` columns,
+-- and so they may contain old data, which means we need to keep them.
```
fabiotudone-da, line 9: Do I understand correctly?

```suggestion
-- driver). We just rename the `rejection_status_code` and `rejection_status_message` columns,
-- so as to preserve existing data in them.
```

ledger/participant-integration-api/src/main/scala/platform/store/dao/CommandCompletionsTable.scala
```diff
@@ -25,19 +24,14 @@ private[platform] object CommandCompletionsTable {
...
 
   private val rejectedCommandParser: RowParser[CompletionStreamResponse] =
     sharedColumns ~ int("status_code") ~ str("status_message") map {
       case offset ~ recordTime ~ commandId ~ statusCode ~ statusMessage =>
-        CompletionStreamResponse(
-          checkpoint = toApiCheckpoint(recordTime, offset),
-          completions = Seq(Completion(commandId, Some(Status(statusCode, statusMessage)))),
-        )
+        val status = StatusProto.of(statusCode, statusMessage, Seq.empty)
+        CompletionFromTransaction.rejectedCompletion(recordTime, offset, commandId, status)
```
nicu-da, line 34: Shouldn't this parser check for rejection_status?

fabiotudone-da:
> The migration is only done for the append-only database

How about the "vintage" one?

nicu-da (commented):
Nice work!
Do we have any sort of integration tests that we can use to validate the behavior when reading completions that we stored both with the old `status` and `message` along with the new `rejection_status` format?

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V31__rejection_status_proto.sql
SamirTalwar-DA: Because it's in memory, so all migrations are always run on an empty ledger.

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
SamirTalwar-DA: I believe so but I'm waiting for @dasormeter to confirm. If he's happy then I will apply the suggestions above; otherwise I'll make a proper migration.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V106__add_rejection_status_proto_column.sql
SamirTalwar-DA: The participant is not highly available; HA is managed by running multiple participants, which do not share an index.

ledger/participant-integration-api/src/main/scala/platform/store/dao/CommandCompletionsTable.scala
SamirTalwar-DA: Not in this version, because this uses the "mutable" index, to which we're not making changes.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V106__add_rejection_status_proto_column.sql
nmarton-da: not sure about the question but that is not entirely true. with canton and HKEX we will have participant running in HA, sharing a replicated database, in a non-coordinated lock based fashion. PoC code with feature flag already landed. It is activa passive setup (so no load balancing), with ensuring only one indexer is active at the time.

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V31__rejection_status_proto.sql
fabiotudone-da: I can vaguely remember we also it use on-disk (but, if so, perhaps with sandbox) but I may be confused.

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V1__Append_only_schema.sql
danielporterda: Confirming that you're free to make breaking changes to this particular file for now - we do not have migration obligations for our August or September deliveries to HKEX.

ledger/participant-integration-api/src/main/resources/db/migration/h2database/V31__rejection_status_proto.sql
SamirTalwar-DA: We _can_, but it's not officially supported and you should not do that.

Using the append-only index, we do not support migrations at all on H2 (which is why I modified it directly).

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V106__add_rejection_status_proto_column.sql
SamirTalwar-DA: That's info I wasn't aware of. Thanks!

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala
```diff
@@ -6,42 +6,46 @@ package com.daml.platform.store.backend.common
...
-          completions = Seq(Completion(commandId, Some(Status(statusCode, statusMessage)))),
-        )
-    }
+  private val rejectedCommandParser: RowParser[CompletionStreamResponse] = {
+    val parserWithCodeAndMessage = sharedColumns ~
+      int("rejection_status_code") ~ str("rejection_status_message") map {
+        case offset ~ recordTime ~ commandId ~ rejectionStatusCode ~ rejectionStatusMessage =>
+          val status = StatusProto.of(rejectionStatusCode, rejectionStatusMessage, Seq.empty)
+          CompletionFromTransaction.rejectedCompletion(recordTime, offset, commandId, status)
+      }
```
nmarton-da, line 40: do you think we have tests somewhere to test this part?
normally it would not be tested because on the HEAD we would always populate the rejection_status

nmarton-da: @rautenrieth-da did you maybe encountered a test suit which would test this?

nmarton-da (approved):
nice'n'easy :+1:  thanks @SamirTalwar (commented only on minor stuff)

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -318,6 +327,9 @@ CREATE INDEX participant_events_consuming_exercise_tree_event_witnesses_idx ON p
 -- lookup by contract id
 CREATE INDEX participant_events_consuming_exercise_contract_id_idx ON participant_events_consuming_exercise (contract_id);
 
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
SamirTalwar-DA: IntelliJ keeps adding these for no reason. It really bugs me.

nicu-da:
@SamirTalwar-DA Thinking more about this, I'm not really convinced that we should store the raw `google.rpc.Status` directly. This creates a direct coupling with the external model which is not under our control at all, and in the long future if there's a breaking change this is gonna really complicate our life. 
It's also a really weird coupling between a completion rejection and gRPC response, I don't think we should have those two so tighly bound.
Wouldn't it make more sense to create our own protobuf model to represent completion failure reason?

SamirTalwar-DA:
@nicu-da: This is a reasonable concern. Fortunately, `google.rpc.Status` is a protocol buffers message, which means that (a) it should never change a backwards-incompatible way, and (b) if it does change in a way that we don't want to support, we can just copy it, rename it to `daml.Status`, and carry on using it, because the message type is not embedded in the message.

The reason we need to store a protobuf (as opposed to just having columns) is because with the new API, the ledger is capable of passing back arbitrary information as part of the `details`, which then needs to be sent to the ledger API client. The indexer and API server may not necessarily understand this information but they still need to send it on.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,18 @@ CREATE TABLE participant_command_completions (
...
+    -- The three columns below are `NULL` if the completion is for an accepted transaction.
+    -- The `rejection_status` contains a Protocol-Buffers-serialized message of type
+    -- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+    -- driver).
+    -- The `rejection_status_code` and `rejection_status_message` columns will always be
+    -- `NULL` in an H2-backed index, as data is not persisted across ledger restarts. However, we
+    -- keep them for parity with old data in other databases.
+    rejection_status_code INTEGER,
+    rejection_status_message VARCHAR,
+    rejection_status BLOB
```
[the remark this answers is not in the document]
SamirTalwar-DA: Done.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala
SamirTalwar-DA: We are very bad at testing migrations. I have no idea how we would go about testing this.

fabiotudone-da:
@SamirTalwar-DA could you elaborate on the reason why we only need to do this for the append-only schema?

SamirTalwar-DA:
@fabiotudone-da: We have decided that we will only support the v2 participant state API with the append-only index. This means that there will never be any "details" in the `rejection_status` when using the mutable index, and so we will always be able to fully reconstruct the status from its code and message.

We decided this because adding a new migration to the mutable index causes a whole load of problems right now. This is due to the way migrations are ordered within our codebase: all mutable index migrations must be applied before migrating to the append-only index. If we add a new one, it'll break any existing deployments using the append-only index schema.

We hope that the mutable index will be phased out fairly shortly, so this strange situation should ideally be temporary.

fabiotudone-da (approved):
LGTM!

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,18 @@ CREATE TABLE participant_command_completions (
...
+    -- The transaction ID is `NULL` for rejected transactions.
     transaction_id VARCHAR,
-    status_code INTEGER,
-    status_message VARCHAR
+    -- The three columns below are `NULL` if the completion is for an accepted transaction.
+    -- The `rejection_status` contains a Protocol-Buffers-serialized message of type
+    -- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
+    -- driver).
+    -- The `rejection_status_code` and `rejection_status_message` columns will always be
+    -- `NULL` in an H2-backed index, as data is not persisted across ledger restarts. However, we
```
miklos-da, line 131: H2 supports disk-based persistence as well, so this statement may be confusing.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -121,9 +121,18 @@ CREATE TABLE participant_command_completions (
     application_id VARCHAR NOT NULL,
     submitters ARRAY NOT NULL,
     command_id VARCHAR NOT NULL,
+    -- The transaction ID is `NULL` for rejected transactions.
     transaction_id VARCHAR,
-    status_code INTEGER,
-    status_message VARCHAR
+    -- The three columns below are `NULL` if the completion is for an accepted transaction.
+    -- The `rejection_status` contains a Protocol-Buffers-serialized message of type
+    -- `google.rpc.Status`, containing the code, message, and further details (decided by the ledger
```
miklos-da, line 128: Instead of storing the full serialized `google.rpc.Status` message, another option worth considering is storing the contents of `Status.details` only in a serialized form (i.e., in addition to the existing `code` and `message` fields). This may be a good compromise between storing the full serialized status message vs. completely decoupling from it, as we don't have a type for the contents of the `details` field anyway.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala
rautenrieth-da: The [compatibility package](https://github.com/digital-asset/daml/tree/main/compatibility) tests migrations in a way. It does:

1. install an old SDK version
2. push some data into a sandbox ledger, using the ledger API
3. upgrade to the next SDK version (up to and including the head version)
4. use the ledger API to check that the ledger content still looks the same
5. go to step 2

The application that uploads and checks data is a custom client application written in Haskell. There are checks for comparing the ACS and contract visibility due to divulgence. It could probably be extended to compare completions as well.

nmarton-da: Yeah but do we also test rejections this way? Never mind me, it will probably work...although we would not necessarily would have here dead code if we follow @miklos-da useful suggestion

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
nmarton-da: I like this suggestion.
Especially so that it would still make use of the old fields, so no worries with migration, maintaining dead compatibility code etc.
Also this covers @nicu-da 's concern with the dedicated, in-house protobuf schema
Also it would allow more easily to pass additional detail content for accpeted completion as well (not sure that is needed at all)

meiersi-da: That is a great suggestion. It will also simplify ad-hoc debugging in the IndexDB, as status_code and status_message are still accessible to queries.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
meiersi-da:
```suggestion
    -- `NULL` in an H2-backed index, as we do not support disk-based H2 persistence. However, we
```

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
fabiotudone-da: The price is probably a bit more logic but I'd also lean towards "blob == only things that _need_ to be opaque", if possible.

SamirTalwar-DA: I think this is reasonable. However, we'd need to create a new protobuf for the details, or a custom serialization mechanism, as it's a `repeated Any`, which cannot, AFAIK, be easily serialized and deserialized.

Another way of dealing with this might be to use `BYTEA ARRAY`, or just `ARRAY` in H2.

What do you think?

meiersi-da: I would advise to use a well-defined proto message with buf checks then.

Oracle does not seem to support `BYTEA ARRAY`: https://stackoverflow.com/questions/16885526/return-list-type-array-of-blobs-in-oracle-11g

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala
rautenrieth-da: I don't think we test rejections, but we could extend the compatibility tests to cover them.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
fabiotudone-da: > Another way of dealing with this might be to use BYTEA ARRAY, or just ARRAY in H2.

From the modelling perspective, in this case an array looks more convenient to me and it doesn't really make the details less opaque. I am not sure that there aren't DB-related impacts though. @digital-asset/kv-participant?

SamirTalwar-DA: The downside of using a protobuf message is that it can't be easily embedded into a `Status` without deserialization and reserialization, so we're still paying that cost.

That said, I think it's still worth doing to see how it goes.

SamirTalwar-DA: Made a PR on top of this one to get a separate review. Please take a look at https://github.com/digital-asset/daml/pull/10613 and let me know if you approve.

nmarton-da: > The downside of using a protobuf message is that it can't be easily embedded into a Status without deserialization and reserialization, so we're still paying that cost.

I am sceptical about this aspect:
* To my best knowledge this is tiny data, so additional serialization effort will be also tiny
* Since it is null for all successful completion, we would optimize only the rejection case, which is also (hopefully) tiny fraction compared to successful completions

SamirTalwar-DA: OK, merged these changes into this PR branch. Let me know if you have any further concerns.

miklos-da (approved):
Elegant solution!

PR #10601, opened 2021-08-17
Add multiple ways of specifying deduplication [KVL-1047]

nicu-da:
### Pull Request Checklist
The previous way of specifying deduplication is kept as `deduplication_time` and represents a duration (no changes there)

~~Introduce alternative of specyfing `deduplication_start` which represents a Timestamp in the past for the start of deduplication.~~

~~For backwards compatiblity we transform `deduplication_start` into a duration using the formula `deduplication_start + minSkew - submission_time`.~~

Add `deduplication_offset` as a representation for deduplication period. This represents an offset from which deduplication starts. It is supported only for v2 WriteServices.

Add an optional conformance test `CommandDeduplicationOffsetIT`, which should be enabled only for ledgers that have support for offset based deduplication.

trigger the build.


language-support/hs/bindings/src/DA/Ledger/Types.hs
```diff
@@ -103,6 +103,11 @@ data Command
       }
     deriving (Eq,Ord,Show)
 
+data Deduplication
```
SamirTalwar-DA, line 106: Can we go with `DeduplicationPeriod`?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -225,10 +226,19 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
...
-              )
-              Instant.now().plusSeconds(dedup.seconds).plusNanos(dedup.nanos.toLong)
+              lazy val protoMaxDedup = ProtoDuration.of(maxDedup.getSeconds, maxDedup.getNano)
+              val timeoutDuration = commands.deduplication match {
+                case Deduplication.Empty =>
+                  protoMaxDedup
+                case Deduplication.DeduplicationTime(value) =>
+                  value
+                case Deduplication.DeduplicationStart(_) =>
+                  protoMaxDedup
```
SamirTalwar-DA, line 236: If we're going to ignore this, can we make sure the work is captured somewhere? And add a big fat comment explaining this?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -225,10 +226,19 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
...
-              val dedup = commands.deduplicationTime.getOrElse(
-                ProtoDuration.of(maxDedup.getSeconds, maxDedup.getNano)
-              )
-              Instant.now().plusSeconds(dedup.seconds).plusNanos(dedup.nanos.toLong)
+              lazy val protoMaxDedup = ProtoDuration.of(maxDedup.getSeconds, maxDedup.getNano)
+              val timeoutDuration = commands.deduplication match {
+                case Deduplication.Empty =>
+                  protoMaxDedup
+                case Deduplication.DeduplicationTime(value) =>
+                  value
```
SamirTalwar-DA, line 234: If you converted this to a `Duration` here, it would simplify the rest of the function.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -225,10 +226,19 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
             }
             val commandTimeout = {
```
SamirTalwar-DA, line 227: Maybe pull this out into its own function.

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -80,8 +80,10 @@ class CommandTrackerFlowTest
       Some(
         Commands(
           commandId = commandId,
-          deduplicationTime =
-            dedupTime.map(t => com.google.protobuf.duration.Duration(t.getSeconds)),
+          deduplication = dedupTime
+            .map(t => com.google.protobuf.duration.Duration(t.getSeconds))
+            .map(Commands.Deduplication.DeduplicationTime)
```
SamirTalwar-DA, line 85: Please fuse these maps.

I think it's reasonable to alias `com.google.protobuf.duration.Duration` as `DurationProto` in the imports for terseness.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala
```diff
@@ -59,10 +60,12 @@ final class CommandsValidator(ledgerId: LedgerId, submissionIdGenerator: Submiss
             s"Can not represent command ledger time $ledgerEffectiveTime as a Daml timestamp"
           )
         )
-      deduplicationDuration <- validateDeduplicationDuration(
-        commands.deduplicationTime,
+      deduplication <- validateDeduplicationDuration(
```
SamirTalwar-DA, line 63: Please use `deduplicationPeriod`.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/services/grpc/GrpcCommandSubmissionService.scala
```diff
@@ -61,7 +62,13 @@ class GrpcCommandSubmissionService(
         .value(
           metrics.daml.commands.validation,
           validator
-            .validate(request, currentLedgerTime(), currentUtcTime(), maxDeduplicationTime()),
+            .validate(
```
SamirTalwar-DA, line 65: You can join these two lines.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
```diff
@@ -99,32 +102,64 @@ trait FieldValidations {
   def requirePresence[T](option: Option[T], fieldName: String): Either[StatusRuntimeException, T] =
     option.fold[Either[StatusRuntimeException, T]](Left(missingField(fieldName)))(Right(_))
 
+  /** We validate only using current time because we set the currentTime as submitTime so no need to check both
+    */
   def validateDeduplicationDuration(
-      durationO: Option[com.google.protobuf.duration.Duration],
+      deduplication: Deduplication,
```
SamirTalwar-DA, line 108: please use `deduplicationPeriod` here too.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +57,22 @@ message Commands {
   // Required
   repeated Command commands = 8;
 
-  // The length of the time window during which all commands with the same ``act_as`` parties and
-  // the same command ID will be deduplicated.
-  // Duplicate commands submitted before the end of this window return an ALREADY_EXISTS error.
-  // Optional
-  google.protobuf.Duration deduplication_time = 9;
+  oneof deduplication {
```
SamirTalwar-DA, line 60: I thought we were going with `deduplication_period`?

Might be wrong, but I definitely find it more understandable.

SamirTalwar-DA (approved):
The logic looks good to me. I just have questions about naming.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
```diff
@@ -99,32 +102,64 @@ trait FieldValidations {
...
+            s"The given deduplication time of $result exceeds the maximum deduplication time of $maxDeduplicationTime",
+          ).map(DeduplicationPeriod.DeduplicationDuration)
+        case Deduplication.DeduplicationStart(startFrom) =>
+          val start = TimestampConversion.toInstant(startFrom)
+          for {
+            maxSkew <- maxSkew.toRight(missingLedgerConfig())
+            _ <- Either.cond(
+              start.isBefore(currentTime),
+              start,
+              invalidField(fieldName, "Deduplication start time is after current time"),
```
fabiotudone-da, line 146:
```suggestion
              invalidField(fieldName, "Deduplication start time is equal to or after current time"),
```

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
```diff
@@ -99,32 +102,64 @@ trait FieldValidations {
...
+            )
+            currentTimeDuration = DeduplicationPeriod.deduplicationDurationFromTime(
+              currentTime,
+              start,
+              maxSkew,
+            )
+            _ <- validateDuration(
+              currentTimeDuration,
+              s"The given deduplication start has a current time based duration of $currentTimeDuration which exceeds the maximum deduplication time of $maxDeduplicationTime",
```
fabiotudone-da, line 155:
```suggestion
              s"The requested deduplication start yields a duration of $currentTimeDuration which exceeds the maximum deduplication period of $maxDeduplicationTime",
```

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
```diff
@@ -487,7 +487,7 @@ reset args = do
                     , cid = L.CommandId $ TL.fromStrict $ UUID.toText cmdId
                     , actAs = parties
                     , readAs = []
-                    , dedupTime = Nothing
+                    , dedup = Nothing
```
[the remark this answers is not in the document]
SamirTalwar-DA: It may not be a duration. "Period" is what we're using elsewhere.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
```diff
@@ -43,6 +48,70 @@ final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterva
...
+    "CDSimpleDeduplicationUsingStartTimestamp",
+    "Deduplicate commands within the deduplication time window which is specified by deduplication start",
+    allocate(SingleParty),
+    runConcurrently = false, //we modify the time model
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    ledger
+      .getTimeModel()
+      .flatMap(timeModelResponse => {
+        val timeModel = timeModelResponse.getTimeModel
+        val time = Instant.now()
```
miklos-da, line 64:
```suggestion
        val deduplicationTimepoint = Instant.now()
```
or
```suggestion
        val deduplicateFrom = Instant.now()
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
```diff
@@ -43,6 +48,70 @@ final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterva
...
+        val maxSkew = deduplicationTime
+        val deduplicatedResult = for {
+          ledgerTime <- ledger.time()
+          mrt = ledgerTime.plusSeconds(30)
```
miklos-da, line 68:
```suggestion
          maxRecordTime = ledgerTime.plusSeconds(30)
```

language-support/hs/bindings/src/DA/Ledger/Types.hs
nicu-da: Good point, should've cleaned and used the consistent naming everywhere

daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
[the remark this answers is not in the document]
nicu-da: Updating to period everywhere

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
nicu-da: Agree, deduplication_period it is

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +57,22 @@ message Commands {
...
+
+      // Specifies the length of the deduplication period.
```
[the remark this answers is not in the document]
nicu-da: Good point regarding this change. If you check the future lines it says:

```
    // The participant or Daml ledger converts the duration into a timestamp ``deduplication_start``
      // by subtracting the duration from its local clock at some point during processing the submission.
```


So the deduplication is calculated in the past for the new version, but in the future for the old version. How do we want to capture this in the documentation? @SamirTalwar-DA @fabiotudone-da

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +57,22 @@ message Commands {
...
+      // The participant or Daml ledger converts the duration into a timestamp ``deduplication_start``
+      // by subtracting the duration from its local clock at some point during processing the submission.
```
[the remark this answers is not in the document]
nicu-da: I'm all good to remove it. I copied the descriptions from the existing draft PR https://github.com/digital-asset/daml/pull/9567/files#diff-ed4ed30e4f593aaf9d013b39ed5272dfc0df08c548cc78f110fd4471eaa6f59c and I do not posses the knowledge to determine how detailed/abstract we should be in this descriptions. All good from my part to drop it if you think so

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: There is a story to remove this code all together and should be done soon. Do you think I should comment with the issue id here?
I decided to not calculate the duration because I would not want to have to handle `maxSkew` as well here, and it's just safer to use max duration.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -225,10 +226,19 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
...
+                  protoMaxDedup
+                case Deduplication.DeduplicationTime(value) =>
+                  value
+                case Deduplication.DeduplicationStart(_) =>
+                  protoMaxDedup
+              }
+              Instant
+                .now()
+                .plusSeconds(timeoutDuration.seconds)
+                .plusNanos(timeoutDuration.nanos.toLong)
```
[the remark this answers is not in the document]
nicu-da: Yes, should be removed quite soon

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -33,10 +72,14 @@ object DeduplicationPeriod {
   /** The `offset` defines the start of the deduplication period. */
   final case class DeduplicationOffset(offset: Offset) extends DeduplicationPeriod
 
+  final case class DeduplicationFromTime(start: Instant) extends DeduplicationPeriod
```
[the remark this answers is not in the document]
nicu-da: Renamed to be consistent with the proto namings.
Regarding the comments should we keep the simple ones (just the first line) or the full comments from the draft?

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,45 @@ sealed trait DeduplicationPeriod extends Product with Serializable
...
+    case DeduplicationDuration(duration) =>
+      time.plus(duration)
+    case DeduplicationOffset(_) =>
+      throw new NotImplementedError("Offset deduplication is not supported")
+    case DeduplicationFromTime(start) =>
+      val duration = deduplicationDurationFromTime(time, start, maxSkew)
+      time.plus(duration)
+  }
+
+  /** deduplication_start: compute the duration as submissionTime + config.minSkew - deduplication_start
```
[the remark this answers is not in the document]
nicu-da: Really good catch, the comment is actually correct 🤦

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,45 @@ sealed trait DeduplicationPeriod extends Product with Serializable
...
+    * We measure `deduplication_start` on the ledger’s clock, and thus need to add the minSkew to compensate for the maximal skew that the participant might be behind the ledger’s clock.
```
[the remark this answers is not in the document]
nicu-da: Not sure I follow, which conventions exactly?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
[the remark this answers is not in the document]
fabiotudone-da: @SamirTalwar-DA do you know if that was intentionally so detailed in the proto doc?

andreaslochbihler-da:
> For backwards compatiblity we transform `deduplication_start` into a duration using the formula `deduplication_start + maxSkew - submission_time`.

Ultimately, `deduplication_start` should be passed through to the `WriteService` as part of the `SubmitterInfo`. Canton, e.g., converts a `deduplication_time` back into a `deduplication_start`. If the ledger API server first converts a the `deduplication_start` into a `deduplication_duration` and then the WriteService back, we'll have some precision loss as the two conversions most likely do not use the same clock.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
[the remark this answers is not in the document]
andreaslochbihler-da: In the proposal PR https://github.com/digital-asset/daml/pull/9567/files#diff-ed4ed30e4f593aaf9d013b39ed5272dfc0df08c548cc78f110fd4471eaa6f59cR72-R95, we decided to not reuse the existing field for the deduplication_duration and instead introduce a new field, so that we can then later properly deprecate the field and issue warnings if someone still uses it.

I'm against documenting the future aspect of this field. For backwards-compatibility, we want to emulate the old behaviour by treating the future-looking period as if it was backward-looking. If an application specifies the same deduplication time for each repetition of its commands, then it won't notice a difference. If it uses different times for different resubmissions, it may notice a difference, but @bame-da agreed that command dedup is currently sufficiently broken to warrant such a breaking change. Otherwise, if we document "future-looking" here, then all applications will break as soon as they switch to a Read/WriteService-v2 implementation because that would then have to reject deduplication_time specifications.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -313,6 +307,21 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
...
 
+  private def maxTimeoutFromDeduplicationPeriod(deduplicationPeriod: DeduplicationPeriod) = {
+    lazy val maxDeduplicationDuration = maxDeduplicationTime()
+    val timeoutDuration = deduplicationPeriod match {
+      case DeduplicationPeriod.Empty =>
+        maxDeduplicationDuration
+      case DeduplicationPeriod.DeduplicationTime(duration) =>
+        JDuration.ofSeconds(duration.seconds, duration.nanos.toLong)
+      case DeduplicationPeriod.DeduplicationStart(_) =>
+        maxDeduplicationDuration
```
andreaslochbihler-da, line 318: Shouldn't this raise an error if the deduplication start is earlier than what the `maxDeduplciationDuration` allows?

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
```diff
@@ -99,32 +102,64 @@ trait FieldValidations {
   def requirePresence[T](option: Option[T], fieldName: String): Either[StatusRuntimeException, T] =
     option.fold[Either[StatusRuntimeException, T]](Left(missingField(fieldName)))(Right(_))
 
-  def validateDeduplicationDuration(
-      durationO: Option[com.google.protobuf.duration.Duration],
+  /** We validate only using current time because we set the currentTime as submitTime so no need to check both
+    */
+  def validateDeduplicationPeriod(
```
andreaslochbihler-da, line 107: Do these checks run on all submissions or only for those where the WriteService has https://github.com/digital-asset/daml/blob/main/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala#L110 set to `true`? I'd strongly prefer it this kind of validation is done by the WriteService that's responsible for command deduplication. Otherwise, we're again using different clocks at different points on the submission path, which is a recipe for sporadic failures..

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: That should already be the case, as by this point the deduplication should've already been validated. To be true to specifics though I will calculate the deduplication duration here and not just use the max duration.

nicu-da:
> > For backwards compatiblity we transform `deduplication_start` into a duration using the formula `deduplication_start + maxSkew - submission_time`.
> 
> Ultimately, `deduplication_start` should be passed through to the `WriteService` as part of the `SubmitterInfo`. Canton, e.g., converts a `deduplication_time` back into a `deduplication_start`. If the ledger API server first converts a the `deduplication_start` into a `deduplication_duration` and then the WriteService back, we'll have some precision loss as the two conversions most likely do not use the same clock.

@andreaslochbihler-da That is done only for the v2 to v1 WriteService adapter, as the v1 WriteService requires an `Instant` which specifies the `deduplicate_until`. For the v2 we just pass through the deduplication period and let the WriteService handle it as it wishes.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
nicu-da: @andreaslochbihler-da 
Currently it runs for all submissions. I agree that it would a lot better to just validate for api deduplication when the api deduplication is enabled. i could for sure change that. Do you think it would make more sense to remove the validation all together?
@SamirTalwar-DA Any thoughts on this?

andreaslochbihler-da: I don't know. The validation may be useful for other implementations of the read/write service. But then this should be a utility function that lives somewhere where the `DeduplicationPeriod`s are defined, shouldn't it?

andreaslochbihler-da:
> @andreaslochbihler-da That is done only for the v2 to v1 WriteService adapter, as the v1 WriteService requires an `Instant` which specifies the `deduplicate_until`. For the v2 we just pass through the deduplication period and let the WriteService handle it as it wishes.

Great. But then, you'll also have to add the `DeduplicationTimestamp` case from https://github.com/digital-asset/daml/pull/9638/files#diff-11f765f35b6ddb52aac00cb67263587c80b5b61b45796da65e60ee46567c717aR23 to the `DeduplicationPeriod`, don't you?

nicu-da:
> > @andreaslochbihler-da That is done only for the v2 to v1 WriteService adapter, as the v1 WriteService requires an `Instant` which specifies the `deduplicate_until`. For the v2 we just pass through the deduplication period and let the WriteService handle it as it wishes.
> 
> Great. But then, you'll also have to add the `DeduplicationTimestamp` case from https://github.com/digital-asset/daml/pull/9638/files#diff-11f765f35b6ddb52aac00cb67263587c80b5b61b45796da65e60ee46567c717aR23 to the `DeduplicationPeriod`, don't you?

It's already there but uses the name `DeduplicationStart`, will rename to be consistent with the protobuf naming.

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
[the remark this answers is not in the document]
miklos-da: The ones described in `DeduplicationPeriod`, specifically for the option that takes a point in time for the start of the deduplication period (however we end up calling it). I.e., I just would like to ask to align the scaladoc for this helper method vs. the classes implementing the `DeduplicationPeriod` trait.

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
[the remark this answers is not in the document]
miklos-da: I would suggest keeping as much of the full comments from the draft as possible, however, trying to keep the contract as generic as possible (hence removing/minimizing implementation specific details).

buf.yaml
```diff
@@ -30,3 +30,6 @@ lint:
 breaking:
   use:
     - WIRE_JSON
+  ignore_only:
+    FIELD_SAME_ONEOF:
+      - com/daml/ledger/api/v1/commands.proto
```
SamirTalwar-DA, line 35: Please add a new line.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
```diff
@@ -180,16 +181,18 @@ final class CommandClient(
...
-        )
-        r.copy(commands = Some(commands.copy(deduplicationTime = updateDedupTime)))
+        val updatedDeduplicationPeriod = commands.deduplicationPeriod match {
+          case DeduplicationPeriod.Empty =>
+            DeduplicationPeriod.DeduplicationTime(
+              Duration
+                .of(
+                  config.defaultDeduplicationTime.getSeconds,
+                  config.defaultDeduplicationTime.getNano,
+                )
```
SamirTalwar-DA, line 191: We've really gotta extract out some conversion functions somewhere.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -318,6 +318,25 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => Opti
...
+      case DeduplicationPeriod.DeduplicationTime(duration) =>
+        JDuration.ofSeconds(duration.seconds, duration.nanos.toLong)
+      case DeduplicationPeriod.DeduplicationOffset(
+            _
+          ) => //no way of extracting the duration from here, will be removed soon
+        maxDeduplicationDuration
+    }
+    Instant
+      .now()
+      .plus(timeoutDuration)
```
SamirTalwar-DA, line 337:
```suggestion
    Instant.now().plus(timeoutDuration)
```

SamirTalwar-DA: You may want to inject `now` here, too, as it's likely to cause issues if you use slightly different definitions throughout a request.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
SamirTalwar-DA: Sorry, I missed this question.

I agree that this should be done by the driver (on the participant or the committer, as necessary), not the ledger API server, with the same clock and definition of "now" as is going to be used for the submission itself.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
```diff
@@ -99,32 +102,47 @@ trait FieldValidations {
...
+      def validateDuration(duration: Duration, exceedsMaxDurationMessage: String) = {
+        if (duration.isNegative)
+          Left(invalidField(fieldName, "Duration must be positive"))
+        else if (duration.compareTo(maxDeduplicationDuration) > 0)
+          Left(
+            invalidField(
+              fieldName,
+              exceedsMaxDurationMessage,
+            )
+          )
```
SamirTalwar-DA, line 125:
```suggestion
          Left(invalidField(fieldName, exceedsMaxDurationMessage))
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto
```diff
@@ -57,11 +57,20 @@ message Commands {
...
-  // Duplicate commands submitted before the end of this window return an ALREADY_EXISTS error.
-  // Optional
-  google.protobuf.Duration deduplication_time = 9;
+  oneof deduplication_period {
+
+      // Specifies the length of the deduplication period.
+      // It is interpreted relative to the local clock at some point during the submission's processing.
+      //
+      // Must be non-negative.
+      google.protobuf.Duration deduplication_time = 9;
```
SamirTalwar-DA, line 66: Did we agree or not on deprecating this field, and introducing a new one, to force people to understand the new "looking back" semantics?

Happy to do it in a separate PR, but I don't want to drop it.

nicu-da: Yes, adding a new one in a different PR because this one is too big already so I'm trying to limit the changes I introduce.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+        )
+      _ <- ledger.submit(requestA1)
+      completionsA1First <- ledger.firstCompletions(
+        ledger.completionStreamRequest(ledgerEnd1)(party)
+      )
+      ledgerEndAfterFirstSubmit <- ledger.currentEnd()
+      _ <- ledger.submit(requestA1)
+      completionsA1Second <- ledger.firstCompletions(
+        ledger.completionStreamRequest(ledgerEndAfterFirstSubmit)(party)
+      ) // this will deduplicated as the deduplication offset includes the first submitted command
```
miklos-da, line 37:
```suggestion
      ) // This will be deduplicated as the deduplication offset includes the first submitted command.
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+      )
+      val failureCommandId1 =
+        assertSingleton("Expected only one first failure", completionsA1Second)
+      assert(
+        failureCommandId1.commandId == requestA1.commands.get.commandId,
+        "The command ID of the second completion does not match the command ID of the submission",
+      )
+      assert(
+        failureCommandId1.status.get.code == Code.ALREADY_EXISTS.value,
+        s"Second completion for the first submit was not deduplicated ${failureCommandId1.status}",
```
miklos-da, line 73:
```suggestion
        s"Second completion for the first submission was not deduplicated ${failureCommandId1.status}",
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+
+      val completionCommandId2 =
+        assertSingleton(
+          "Expected only one second successful completion",
+          completionsA2First,
+        )
+
+      assert(
+        completionCommandId2.commandId == requestA2.commands.get.commandId,
+        "The command ID of the second completion does not match the command ID of the submission",
```
miklos-da, line 84:
```suggestion
        "The command ID of the second completion does not match the command ID of the second submission",
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+        ledger.completionStreamRequest(ledgerEnd2)(party)
+      )
+      activeContracts <- ledger.activeContracts(party)
+    } yield {
+      assert(ledgerEnd1 != ledgerEnd2)
+      val completionCommandId1 =
+        assertSingleton("Expected only one first completion", completionsA1First)
+      assert(
+        completionCommandId1.commandId == requestA1.commands.get.commandId,
+        "The command ID of the first completion does not match the command ID of the submission",
```
miklos-da, line 59: Minor recommended clarification:
```suggestion
        "The command ID of the first completion does not match the command ID of the first submission",
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+      )
+      assert(
+        completionCommandId1.status.get.code == Code.OK.value,
+        s"First command did not complete OK. Had status ${completionCommandId1.status}",
+      )
+      val failureCommandId1 =
+        assertSingleton("Expected only one first failure", completionsA1Second)
+      assert(
+        failureCommandId1.commandId == requestA1.commands.get.commandId,
+        "The command ID of the second completion does not match the command ID of the submission",
```
miklos-da, line 69:
```suggestion
        "The command ID of the second completion does not match the command ID of the first submission",
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationOffsetIT.scala
```diff
@@ -0,0 +1,97 @@
...
+      )
+      assert(
+        completionCommandId2.status.get.code == Code.OK.value,
+        s"Second command did not complete OK. Had status ${completionCommandId1.status}",
+      )
+      assert(
+        activeContracts.size == 2,
+        s"There should be 2 active contracts, but received ${activeContracts.size} contracts, with events: $activeContracts",
+      )
+
```
miklos-da, line 94:
```suggestion
```

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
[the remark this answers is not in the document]
miklos-da: The comment is confusing as it's referencing names out of context here.
```suggestion
  /** Computes deduplication duration as the duration `time + minSkew - deduplicationStart`.
```

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,41 @@ sealed trait DeduplicationPeriod extends Product with Serializable
...
+    case DeduplicationDuration(duration) =>
+      time.plus(duration)
+    case DeduplicationOffset(_) =>
+      throw new NotImplementedError("Offset deduplication is not supported")
+  }
+
+  /** deduplication_start: compute the duration as submissionTime + config.minSkew - deduplication_start
+    * We measure `deduplication_start` on the ledger’s clock, and thus need to add the minSkew to compensate for the maximal skew that the participant might be behind the ledger’s clock.
+    * @param time submission time or current time
+    * @param deduplicationStart the [[Instant]] from where we should start deduplication. it must be < than time
```
miklos-da, line 41:
```suggestion
    * @param deduplicationStart the [[Instant]] from where we should start deduplication, must be < than time
```

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,41 @@ sealed trait DeduplicationPeriod extends Product with Serializable
...
+    * @param minSkew The ledger min skew duration
```
miklos-da, line 42:
```suggestion
    * @param minSkew the minimum skew as specified by the current ledger time model
```

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,41 @@ sealed trait DeduplicationPeriod extends Product with Serializable
 
 object DeduplicationPeriod {
 
+  /** Backwards compatibility
+    * Transforms the [[period]] into an [[Instant]] to be used for deduplication into the future.
+    * Offset deduplication is not supported
```
miklos-da, line 23:
```suggestion
  /** Transforms the [[period]] into an [[Instant]] to be used for deduplication into the future.
    * Only used for backwards compatibility. Offset deduplication is not supported.
```

ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala
```diff
@@ -18,6 +18,41 @@ sealed trait DeduplicationPeriod extends Product with Serializable
...
+      period: DeduplicationPeriod,
+  ): Instant = period match {
+    case DeduplicationDuration(duration) =>
+      time.plus(duration)
+    case DeduplicationOffset(_) =>
+      throw new NotImplementedError("Offset deduplication is not supported")
+  }
+
+  /** deduplication_start: compute the duration as submissionTime + config.minSkew - deduplication_start
+    * We measure `deduplication_start` on the ledger’s clock, and thus need to add the minSkew to compensate for the maximal skew that the participant might be behind the ledger’s clock.
```
miklos-da, line 39: Can you break this long line, please?

miklos-da (commented):
Added a few recommendations for improving code documentation.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandClient.scala
nicu-da: I do aggree, so far I've seen it done either like this inline, using implicit extension methods, using a `Conversion` object to do the conversion, I would love to decide on one way of doing things so that we know where to find the conversions when we need them.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: I do aggree, but for now I would not change this because we also use Instant.now() to verify the timeout. I will capture it in another PR

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/FieldValidations.scala
nicu-da: As we removed for now the deduplication period which specifies a timestampt, and for offset we currently have no way of validating here, we only validate when receiving a duration so this is no longer relevant.

SamirTalwar-DA (approved):
Happy to merge this with the expectation that the changes we talked about will arrive in future changes.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: @SamirTalwar-DA I'm working on another PR to fix this. But cannot identify a position where we can inject the now from without calling Instant.now(). Throughout the flow I could not find an easy way to inject the now. 

And in retrospect it makes sense:
We want to measure the timeout inside the `CommandTracker` so we're only interested in the time it passed since we've acknowledged the command until it times out, and that has use the same definition (which it does). It has no direct correlation with time used in another parts of the request flow.

cocreature (approved):
Haskell changes look good, I’ll leave the rest of the review to someone else.

PR #10602, opened 2021-08-17
[JSON-API] Use the token from incoming requests to update the package list

### LINKED ISSUE #10192: Drop package token in the JSON API
Currently the JSON API requires a path to a separate token file which it uses to fetch and refresh packages. Those packages are required for type-aware json encoding/decoding and to infer package ids where possible.

The current approach suffers from two main issues:
1. Managing the token including token refresh is annoying. Several users have build their own scripts/utilities that do this but there’s nothing generic.
2. We are planning to add some form of visibility restrictions to packages later in the year, i.e., per-party packages. At that point the current approach simply doesn’t work.

Instead, we should fetch packages on-demand using the user-supplied token of the given request. We can apply the following caching strategy:
1. If a package id is included, refresh iff we don’t have a package with that id. At least initially, I wouldn’t worry about being fast for package not found. If we care, we could skip the refresh within a given time interval.
2. If a package id is not included, check if we have exactly one package with the given module name. If we have more than one error. If there is none, refresh (similar point to the above, we could skip the refresh within a time interval). If there is one, refresh if there have been more than X seconds since the last refresh, otherwise return that directly.

This completely removes the need for the token file. It has no performance impact beyond the first request if you use package ids. If you do not use package ids, you get a slight performance hit every so often configurable by the refresh interval which seems perfectly acceptable.

This doesn’t yet provide us with per-party packages but should make adding it a lot easier later.

realvictorprm:
should fix #10192

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/util/ErrorOps.scala
```diff
@@ -22,4 +22,13 @@ object ErrorOps {
     def liftErrS[M](msg: String)(f: String => M)(implicit L: Show[L]): M \/ R =
       liftErr(x => f(msg + " " + x))
   }
+
+  implicit final class `EitherT WSS extras`[F[_]: Functor, L, R](private val self: EitherT[F, L, R])
```
realvictorprm, line 26: IDK what WSS means but I just took that name from above :joy:

stefanobaghino-da: I believe it's "web socket service".

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+            }
+            .map { res =>
+              lastUpdated = LocalDateTime.now()
+              res
+            }
+            .run
+        try {
+          ongoingPackageUpdateTask = f.map(_ => ().right)
+          // The task future has been updated, now all waiting threads can continue and pick it up.
+          isValid = true
```
stefanobaghino-da, line 102: What is the subject of `isValid`? What is valid is `isValid` is `true`?

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Concurrent.scala
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.http.util
+
+object Concurrent {
+
+  /** Equal to a Semaphore with max count equal to one. */
+  final case class Mutex() extends java.util.concurrent.Semaphore(1, true)
+}
```
stefanobaghino-da, line 10: I don't think we need this, especially as a publicly available class.

A simple
```
private[this] def newMutex() = new java.util.concurrent.Semaphore(1, true)
```
in `PackageService` will do.

I'm also trying to have a better look at the code that uses this, as I'd like to understand why `synchronized` won't do.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+        updateTaskUpdateSync.release()
+        logger.debug(s"Picking up ongoing update task, is valid? $isValid")
+        ongoingPackageUpdateTask
+      }
     }
 
+    (jwt: Jwt, ec: ExecutionContext, lc: LoggingContextOf[InstanceUUID]) => reload(jwt)(ec, lc)
+  }
+
+  // synchronized via a mutex, so two threads cannot reload it concurrently
```
stefanobaghino-da, line 140: This comment covers an implementation detail that exists somewhere else. I would rather comment on the mutexes internally as to why they are needed and what shared state they are protecting from concurrent access.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+    // was moved into a separate block, so it's easier to track down all interactions.
+    val packageUpdateSync = util.Concurrent.Mutex()
+    val updateTaskUpdateSync = util.Concurrent.Mutex()
+    @volatile var ongoingPackageUpdateTask = Future.successful(().right)
+    @volatile var isValid = false
+
+    def reload(jwt: Jwt)(implicit
+        ec: ExecutionContext,
+        lc: LoggingContextOf[InstanceUUID],
+    ): Future[Error \/ Unit] = {
```
stefanobaghino-da, line 76: I've tried to go through this method a couple of times but it's still unclear to me why:
1. we need two mutexes
2. we need explicit mutexes

Can you guide through these changes, please?

stefanobaghino-da (commented):
No hard blockers on my side, but:
1. the new package reloading method is quite complicated and uses explicit mutexes, which are rarely needed, I would like to better understand why we need this
2. I noticed that overall there's an increase of indentation level that can make the code hard to follow. Some of these seem to follow a pattern (e.g. the changes in `WebSocketService`) that could probably used being factored out in its own method and re-used.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+          f.transform { res =>
+            // regardless of whether this was a success or failure
+            // we always want to release the mutex here.
+            isValid = false
+            packageUpdateSync.release()
+            logger.debug("Released mutex after task finished")
+            res
+          }
+        } catch {
+          case t: Throwable =>
```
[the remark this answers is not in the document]
realvictorprm: Good catch, completely forgot about that :man_facepalming:

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+  @volatile private var lastUpdated = LocalDateTime.MIN
+
+  // Regular updates should happen regardless of the current state every minute.
+  private def packagesShouldBeFetchedAgain =
+    lastUpdated.until(LocalDateTime.now(), temporal.ChronoUnit.SECONDS) >= 60L
+
+  private val _reload = {
+    // Because things can go so easily wrong, the declaration of the mutex
+    // was moved into a separate block, so it's easier to track down all interactions.
+    val packageUpdateSync = util.Concurrent.Mutex()
```
cocreature, line 68: so if I read this correctly the way the concurrency logic works is as follows:
`updateTaskUpateSync` synchronizes updates to the `ongoingPackageUpdateTask` variable.
`updateTaskUpdateSync` is locked while the future executes. I don’t actually understand why this needs to be a mutex.

I agree with @stefanobaghino-da that we can simplify that a bit.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
realvictorprm: My comment unfortunately was lost ._.

So in short.

Under the effort of preventing unnecessary concurrent package reloads, I synchronized the access to the cache with the first mutex `packageUpdateSync` with threads that cannot acquire it picking up the ongoing update task/future. The second future is there to synchronize access to that future, because it could already be outdated (I had to introduce this after the  integration tests failed, they made me aware of that problem).

I'm not super convinced by my solution for the `ongoingPackageUpdateTask` as it's forcing sequential processing. Moreover I think it would be best to have used a fibre with support for semaphores, because os thread level semaphores have quite the overhead.

PS: We can throw the whole synchronization out if we think that the reloading the package cache is super cheap. However be aware of the scenario that with over a minute since the last update, every incoming request at this time could trigger a package reload potentially causing e.g. 1000 reloads. That could result in bad performance.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+    val updateTaskUpdateSync = util.Concurrent.Mutex()
+    @volatile var ongoingPackageUpdateTask = Future.successful(().right)
+    @volatile var isValid = false
+
+    def reload(jwt: Jwt)(implicit
+        ec: ExecutionContext,
+        lc: LoggingContextOf[InstanceUUID],
+    ): Future[Error \/ Unit] = {
+      def update() = {
+        val f =
```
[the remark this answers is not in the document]
realvictorprm: Very true, will change the name.

PS: Remnants of me having refactored part of this code out into a method only to revert that later on :joy:

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
realvictorprm: I will remove that as it doesn't serve a purpose except for debugging

ledger-service/http-json/src/main/scala/com/digitalasset/http/util/Concurrent.scala
realvictorprm: I agree, in the past there was a bit more stuff in here but the code changed again. Thanks for the suggestion :)

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -83,84 +108,132 @@ object HttpService {
...
+          )
+          builder.maxInboundMetadataSize(clientConfig.maxInboundMetadataSize)
+          builder.maxInboundMessageSize(clientConfig.maxInboundMessageSize)
+          builder.build()
+        }
+      )
 
-      ledgerId = apiLedgerId(client.ledgerId): lar.LedgerId
+      lazyLedgerClient =
+        LazyF[Future, Jwt, DamlLedgerClient] { case Jwt(token) =>
```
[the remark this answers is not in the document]
realvictorprm: With this approach everything seems to work correctly.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
-            logger.debug(s"new package IDs not found")
-            ()
+  // volatile, because we want to reduce the probability of unnecessary package updates
+  // due to multiple threads reading this.
+  // Writes to this are synchronized with a mutex, so in theory this should work out fine.
+  @volatile private var lastUpdated = LocalDateTime.MIN
+
+  // Regular updates should happen regardless of the current state every minute.
+  private def packagesShouldBeFetchedAgain =
+    lastUpdated.until(LocalDateTime.now(), temporal.ChronoUnit.SECONDS) >= 60L
```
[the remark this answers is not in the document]
realvictorprm: Will do, also planned for a seperate PR.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+
+  private val _reload = {
+    // Because things can go so easily wrong, the declaration of the mutex
+    // was moved into a separate block, so it's easier to track down all interactions.
+    val packageUpdateSync = util.Concurrent.Mutex()
+    val updateTaskUpdateSync = util.Concurrent.Mutex()
+    @volatile var ongoingPackageUpdateTask = Future.successful(().right)
+    @volatile var isValid = false
```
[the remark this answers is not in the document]
realvictorprm: Yeah, however will remove that stuff

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
realvictorprm: Another avenue would be a optimistic approach where we are fine with some unnecessary updates at the same time.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+      lc: LoggingContextOf[InstanceUUID],
+  ): ET[B] = for {
+    packagesHadToBeFetchedAgain <-
+      if (packagesShouldBeFetchedAgain) EitherT(reload(jwt)).map(_ => true)
+      else EitherT.pure(false): ET[Boolean]
+    _ = logger.trace(s"Did we have to reload the packages: $packagesHadToBeFetchedAgain")
+    result <- EitherT.pure(fn()): ET[B]
+    _ = logger.trace(s"Result: $result")
+    finalResult <-
+      if (!checkIfSuccess(result) && !packagesHadToBeFetchedAgain) {
```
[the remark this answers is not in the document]
realvictorprm: I can adjust it to be that fine grained. In the end my approach was just the easiest initial solution to me.

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -62,8 +59,36 @@ object HttpService {
     def fromLedgerClientError(e: LedgerClientBase.Error): Error = Error(e.message)
   }
 
+  final case class LazyF[F[_], A, B](private val fn: A => F[B]) {
```
akshayshirahatti-da, line 62: given this is fairly generic , would make more sense for this to live in some`util` package

realvictorprm: I agree

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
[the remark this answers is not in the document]
akshayshirahatti-da: +1 I would also be wary of this change maybe @stefanobaghino-da can comment this

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
```diff
@@ -403,25 +413,35 @@ class Endpoints(
...
+                      .search(jwt, jwtPayload, cmd)
+                      .map(
+                        domain.SyncResponse.covariant.map(_)(
+                          _.via(handleSourceFailure)
+                            .map(_.flatMap(toJsValue[domain.ActiveContract[JsValue]](_)))
+                        )
+                      )
+                  }
+                }
+                .sequence
```
akshayshirahatti-da, line 440: generally I'd start looking to break things into more isolated/well contained methods when the nesting runs this deep as it becomes harder to understand the encapsulation of functions.

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -371,7 +359,7 @@ object HttpService {
 
   private def uploadDarAndReloadPackages(
       f: LedgerClientJwt.UploadDarFile,
-      g: LoggingContextOf[InstanceUUID] => Future[PackageService.Error \/ Unit],
+      g: Jwt => LoggingContextOf[InstanceUUID] => Future[PackageService.Error \/ Unit],
   )(implicit ec: ExecutionContext): LedgerClientJwt.UploadDarFile =
-    (x, y) => implicit lc => f(x, y)(lc).flatMap(_ => g(lc).flatMap(toFuture(_): Future[Unit]))
+    (x, y) => implicit lc => f(x, y)(lc).flatMap(_ => g(x)(lc).flatMap(toFuture(_): Future[Unit]))
```
akshayshirahatti-da, line 364: another place IMO where we can probably use slightly more verbose variable and function names :)

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -48,32 +53,130 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+    def reload(jwt: Jwt)(implicit
+        ec: ExecutionContext,
+        lc: LoggingContextOf[InstanceUUID],
+    ): Future[Error \/ Unit] = {
+      def update() = {
+        val f =
+          EitherT
+            .eitherT(
+              Future(
+                logger.debug("Trying to execute a package update")
```
akshayshirahatti-da, line 82: can we get rid of this `Future` to do debug logging ? would the below not work ?
```scala
EitherT.eitherT({
                logger.debug("Trying to execute a package update")
                reloadPackageStoreIfChanged(jwt)(state.packageIds)
            })
```

realvictorprm: To my knowledge `EitherT.eitherT` is not lazily evaluating it's parameter.

ledger-service/http-json/src/main/scala/com/digitalasset/http/Endpoints.scala
realvictorprm: yeah that function truly turned into a mess :smile:

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
realvictorprm: however I think the result of using your code would in this case be the same

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
realvictorprm: I changed that stuff to use an atomic boolean and a countdownlatch

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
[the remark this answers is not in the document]
akshayshirahatti-da: @cocreature whats the reason behind the refresh on point 4 ? i.e if we have the package (but no package id) what do we expect to get from the refresh ?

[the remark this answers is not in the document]
cocreature: the json API errors out if the module name is not unique across modules. If we never refresh on 4 we will keep selecting a package even if it is no longer unique.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
akshayshirahatti-da: ahh ok cheers, my suggestion was motivated by an effort to reduce object creation where possible as long as we don't loose functionality or readability.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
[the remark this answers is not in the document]
realvictorprm: > 1. package id and we have the package

Do you actually mean here:

`package id provided which is in our cache and a template id could be found`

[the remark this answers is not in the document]
cocreature: no if you know the package and you don’t find it, you also don’t need to refresh. packages don’t mutate so a refresh isn’t going to result in a successful lookup.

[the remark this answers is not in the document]
realvictorprm: ok for point 4 however we need to assume that the search was successful, otherwise how should I know that we have the package if no package id is provided :D?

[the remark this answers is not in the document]
cocreature: For 3 you do, for 4 you don’t. If the refresh fails you just resolve to the package you already know. The refresh in 4 is only to turn a successful resolve into an ambiguity error. We want to make sure that this happens at some point but I don’t think it should fail any individual request.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -44,36 +50,116 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
+                if (state.packageIds.contains(packageId)) {
+                  logger.trace("package id and we have the package, never refresh")
+                  keep(result)
+                }
+                // package id and we don’t have the package, always refresh
+                else {
+                  logger.trace("package id and we don’t have the package, always refresh")
+                  doReloadAndSearchAgain()
+                }
+              }
```
realvictorprm, line 157: @cocreature can you take a look whether this is what you wanted and if not what should be changed?

cocreature: looks sensible :+1:

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -527,18 +607,24 @@ class WebSocketService(
       jwt: Jwt,
       jwtPayload: JwtPayload,
   )(implicit
-      lc: LoggingContextOf[InstanceUUID with RequestID]
+      ec: ExecutionContext,
+      lc: LoggingContextOf[InstanceUUID],
+      _jwt: Jwt = jwt,
```
stefanobaghino-da, line 612: Why are there two `Jwt`s here, one of which is `implicit`?

stefanobaghino-da (commented):
The concurrency part is much better now. I think there are a few parts where readability can be improved. More importantly, I think it's cleaner to take a step back and improve the `LedgerClient`  to make it work without requiring a token right away, rather than pursuing the `LazyF` road.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -44,36 +50,116 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
...
           case Some(diff) =>
-            this.state = this.state.append(diff)
+            updateState(diff)
             logger.info(s"new package IDs loaded: ${diff.keySet.mkString(", ")}")
             logger.debug(s"loaded diff: $diff")
-            ()
-          case None =>
-            logger.debug(s"new package IDs not found")
-            ()
+          case None => logger.debug(s"new package IDs not found")
```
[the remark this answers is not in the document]
stefanobaghino-da: This is also an issue that pre-dates this PR, but let's clean it up. 🙂

realvictorprm:
@stefanobaghino-da will take care of that stuff tomorrow. Happy to do some refactoring

stefanobaghino-da:
> @stefanobaghino-da will take care of that stuff tomorrow. Happy to do some refactoring

Definitely no need to work on this outside of working hours. 😄

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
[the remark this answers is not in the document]
stefanobaghino-da: Also quite important: this PR is already very large. If you work on the `LedgerClient` to support accepting a token lazily please do so in another PR and rebase this on top of that.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -1122,38 +1131,43 @@ abstract class AbstractHttpServiceIntegrationTest
...
     instanceUUIDLogCtx(implicit lc =>
-      jsonCodecs(client).map { case (encoder, decoder) =>
-        testCreateCommandEncodingDecoding(encoder, decoder)
-        testExerciseCommandEncodingDecoding(encoder, decoder)
+      jsonCodecs(client).flatMap { case (encoder, decoder) =>
+        testCreateCommandEncodingDecoding(encoder, decoder) *>
+          testExerciseCommandEncodingDecoding(
+            encoder,
+            decoder,
+          )
```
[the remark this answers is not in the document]
realvictorprm: It's a short variant for `m1.flatMap(_ => m2)`. More readable in my opinion because we don't introduce more braces

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
```diff
@@ -44,36 +50,116 @@ private class PackageService(reloadPackageStoreIfChanged: PackageService.ReloadP
     }
   }
 
-  // volatile, reading threads don't need synchronization
-  @volatile private var state: State =
-    State(Set.empty, TemplateIdMap.Empty, Map.empty, Map.empty, Map.empty)
+  private case class StateCache() {
```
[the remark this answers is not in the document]
realvictorprm: Well a normal class doesn't have a apply method out of the box (which I prefer over writing`new XYZ()`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -442,13 +520,15 @@ object WebSocketService {
   ): doobie.Fragment =
     sjd.queries.keyEquality(LfValueDatabaseCodec.apiValueToJsValue(k))
 
-  private[this] object InitialEnrichedContractKeyWithStreamQuery
-      extends EnrichedContractKeyWithStreamQuery[Unit] {
+  private[WebSocketService] final case class InitialEnrichedContractKeyWithStreamQuery()(implicit
+      ec: ExecutionContext
+  ) extends EnrichedContractKeyWithStreamQuery[Unit] {
```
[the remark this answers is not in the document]
realvictorprm: Same here

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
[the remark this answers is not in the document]
stefanobaghino-da: I see, thanks for clarifying.

ledger-service/http-json/src/main/scala/com/digitalasset/http/PackageService.scala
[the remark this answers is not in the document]
stefanobaghino-da: Making it a `case class` also makes it extend `Product` and `Serializable` and synthesizes `equals`, `hashCode`, `toString` and an `unapply` extractor in the companion object (none of which mean anything sensible for this class). If you really can't stand writing `new` you can write the `apply` method yourself, but I would prefer to not abuse `case class`es just for this. What's the problem in writing `new` explicitly?

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
realvictorprm: got rid of the implict jwt. Better pass it explicit.

ledger-service/utils/src/main/scala/com/digitalasset/ledger/service/LedgerReader.scala
```diff
@@ -3,12 +3,15 @@
...
 
+import com.daml.ledger.api.domain.LedgerId
 import com.daml.lf.archive
 import com.daml.lf.data.Ref.{Identifier, PackageId}
 import com.daml.lf.iface.reader.InterfaceReader
 import com.daml.lf.iface.{DefDataType, Interface}
 import com.daml.ledger.api.v1.package_service.GetPackageResponse
 import com.daml.ledger.client.services.pkg.PackageClient
+import com.daml.ledger.client.services.pkg.withoutledgerid.{PackageClient => LoosePackageClient}
+//import com.daml.ledger.rxjava.PackageClient
```
stefanobaghino-da, line 14: No need to comment out code, just drop it.

stefanobaghino-da (commented):
LGTM. In the future you may want to leave rearrangement of argument lists separate to reduce noise on the PR main objective.

ledger-service/utils/src/main/scala/com/digitalasset/ledger/service/LedgerReader.scala
realvictorprm: good catch, thanks

realvictorprm:
> In the future you may want to leave rearrangement of argument lists separate to reduce noise on the PR main objective.

Are you referring to me adding the ledger id inbetween? I guess after all you are right, however I was very hestitant to not adjust it, because the old order otherwise would have forced me to write more unnecessary code (replicating the default values of the parameters).

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -92,7 +95,7 @@ trait AbstractHttpServiceIntegrationTestFuns extends StrictLogging {
   protected val metadata2: MetadataReader.LfMetadata =
     MetadataReader.readFromDar(dar2).valueOr(e => fail(s"Cannot read dar2 metadata: $e"))
 
-  protected val jwt: Jwt = jwtForParties(List("Alice"), List(), testId)
+  protected implicit val jwt: Jwt = jwtForParties(List("Alice"), List(), testId)
```
S11001001, line 98:
```suggestion
  protected val jwt: Jwt = jwtForParties(List("Alice"), List(), testId)
```

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -1120,44 +1137,53 @@ abstract class AbstractHttpServiceIntegrationTest
...
 
     val command0: domain.CreateCommand[v.Record, OptionalPkg] = iouCreateCommand()
 
-    val x = for {
-      jsVal <- encoder.encodeCreateCommand(command0).liftErr(JsonError)
-      command1 <- decoder.decodeCreateCommand(jsVal)
+    val x: EitherT[Future, JsonError, Assertion] = for {
+      jsVal <- EitherT.either[Future, JsonError, JsValue](
+        encoder.encodeCreateCommand(command0).liftErr(JsonError)
+      )
```
S11001001, line 1167: Ascription is more declarative than tparam-passing.

```suggestion
    type F[A] = EitherT[Future, JsonError, A]
    val x: F[Assertion] = for {
      jsVal <- EitherT.either(
        encoder.encodeCreateCommand(command0).liftErr(JsonError)
      ): F[JsValue]
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -138,17 +161,16 @@ class ContractsService(
...
+        lc: LoggingContextOf[InstanceUUID with RequestID]
     ): Future[Option[domain.ActiveContract[LfValue]]] = {
-      import ctx.{jwt, parties, templateIds => templateId}
+      import ctx.{jwt, parties, templateIds => templateId, ledgerId}
       for {
-        resolvedTemplateId <- toFuture(resolveTemplateId(templateId)): Future[
-          TemplateId.RequiredPkg
-        ]
+        resolvedTemplateId <- resolveTemplateId(lc)(jwt, ledgerId)(templateId)
+          .map(_.toOption.flatten.get)
```
S11001001, line 169:
```suggestion
          .flatMap(_.toOption.flatten.cata(Future.successful, Future.failed(...))
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -160,18 +182,19 @@ class ContractsService(
...
-        lc: LoggingContextOf[InstanceUUID]
+        lc: LoggingContextOf[InstanceUUID with RequestID]
     ): Future[Option[domain.ActiveContract[LfValue]]] = {
-      import ctx.{jwt, parties, templateIds => templateId}
+      import ctx.{jwt, parties, templateIds => templateId, ledgerId}
       for {
 
         resolvedTemplateIds <- templateId.cata(
-          x => toFuture(resolveTemplateId(x).map(Set(_))),
+          x => resolveTemplateId(lc)(jwt, ledgerId)(x).map(_.toOption.flatten.get).map(Set(_)),
```
S11001001, line 191: Same with flatMap, Future.failed.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -206,54 +229,63 @@ class ContractsService(
...
-  ): SearchResult[Error \/ domain.ActiveContract[JsValue]] = {
-
-    val (resolvedTemplateIds, unresolvedTemplateIds) = resolveTemplateIds(templateIds)
+      lc: LoggingContextOf[InstanceUUID with RequestID]
+  ): Future[SearchResult[Error \/ domain.ActiveContract[JsValue]]] = for {
+    res <- resolveTemplateIds(jwt, ledgerId)(templateIds)
+    (resolvedTemplateIds, unresolvedTemplateIds) = res
 
-    val warnings: Option[domain.UnknownTemplateIds] =
+    warnings: Option[domain.UnknownTemplateIds] =
```
S11001001, line 274: Congrats, you taught me a Scala thing. I didn't know ascription was allowed here.

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
```diff
@@ -484,14 +530,24 @@ class ContractsService(
...
       xs: OneAnd[Set, Tid]
-  ): (Set[domain.TemplateId.RequiredPkg], Set[Tid]) = {
+  )(implicit
+      lc: LoggingContextOf[InstanceUUID with RequestID]
+  ): Future[(Set[domain.TemplateId.RequiredPkg], Set[Tid])] = {
+    import scalaz.syntax.traverse._
     import scalaz.std.iterable._
-    import scalaz.syntax.foldable._
+    import scalaz.Scalaz.listInstance
+    import scalaz.Scalaz.futureInstance
```
S11001001, line 543: Don't refer to Scalaz (or scala-library, for that matter) implicits by name, use wildcards instead.

```suggestion
    import scalaz.std.list._, scalaz.std.scalaFuture._
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -116,48 +112,50 @@ object HttpService {
...
-      // load all packages right away
-      _ <- eitherT(packageService.reload).leftMap(e => Error(e.shows)): ET[Unit]
-
-      _ = schedulePackageReload(packageService, packageReloadInterval)
-
       commandService = new CommandService(
-        LedgerClientJwt.submitAndWaitForTransaction(client),
-        LedgerClientJwt.submitAndWaitForTransactionTree(client),
+        (jwt, req) => LedgerClientJwt.submitAndWaitForTransaction(client)(jwt, req),
+        (jwt, req) => LedgerClientJwt.submitAndWaitForTransactionTree(client)(jwt, req),
```
S11001001, line 123:
```suggestion
        LedgerClientJwt.submitAndWaitForTransaction(client) _,
        LedgerClientJwt.submitAndWaitForTransactionTree(client) _,
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -116,48 +112,50 @@ object HttpService {
...
         packageService.resolveTemplateId,
         packageService.allTemplateIds,
-        LedgerClientJwt.getActiveContracts(client),
-        LedgerClientJwt.getCreatesAndArchivesSince(client),
-        LedgerClientJwt.getTermination(client),
+        (jwt, ledgerId, filter, verbose) =>
+          LedgerClientJwt.getActiveContracts(client)(jwt, ledgerId, filter, verbose),
+        (jwt, ledgerId, filter, offset, terminates) =>
+          LedgerClientJwt
+            .getCreatesAndArchivesSince(client)(jwt, ledgerId, filter, offset, terminates),
```
S11001001, line 133: Likewise with `_`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -116,48 +112,50 @@ object HttpService {
...
       )
 
       partiesService = new PartiesService(
-        LedgerClientJwt.listKnownParties(client),
-        LedgerClientJwt.getParties(client),
-        LedgerClientJwt.allocateParty(client),
+        LedgerClientJwt.listKnownParties(client).apply,
+        (jwt, partyIds) => LedgerClientJwt.getParties(client)(jwt, partyIds),
+        (jwt, identifierHint, displayName) =>
+          LedgerClientJwt.allocateParty(client)(jwt, identifierHint, displayName),
```
S11001001, line 143: Likewise with `_`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
```diff
@@ -116,48 +112,50 @@ object HttpService {
...
-        LedgerClientJwt.getPackage(pkgManagementClient),
-        uploadDarAndReloadPackages(
-          LedgerClientJwt.uploadDar(pkgManagementClient),
-          implicit lc => packageService.reload,
-        ),
+        (jwt, ledgerId) =>
+          lc => LedgerClientJwt.listPackages(pkgManagementClient)(jwt, ledgerId)(lc),
+        { case (jwt, ledgerId, packageId) =>
+          lc => LedgerClientJwt.getPackage(pkgManagementClient)(jwt, ledgerId, packageId)(lc)
+        },
```
S11001001, line 151: Likewise with `_`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -23,17 +23,20 @@ import doobie.syntax.string._
 import scalaz.syntax.bifunctor._
 import scalaz.syntax.std.boolean._
 import scalaz.syntax.std.option._
-import scalaz.syntax.traverse._
+import scalaz.Scalaz.futureInstance
```
S11001001, line 26: Likewise regarding referring to implicits.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -23,17 +23,20 @@ import doobie.syntax.string._
...
 import scalaz.std.map._
 import scalaz.std.option._
 import scalaz.std.set._
 import scalaz.std.tuple._
 import scalaz.std.vector._
+import scalaz.syntax.traverse._
+import scalaz.Scalaz.listInstance
```
S11001001, line 33: Likewise regarding referring to implicits.

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -332,41 +380,61 @@ object WebSocketService {
...
+          .map(
+            _.valueOr(_ => a.map(_ => com.daml.lf.value.Value.ValueUnit))
           ) // unit will not match any key
 
     }
 
-  private[this] sealed abstract class EnrichedContractKeyWithStreamQuery[Cid]
-      extends StreamQuery[NonEmptyList[domain.ContractKeyStreamRequest[Cid, LfV]]] {
+  private[this] sealed abstract class EnrichedContractKeyWithStreamQuery[Cid](implicit
+      val ec: ExecutionContext
```
S11001001, line 436:
```suggestion
      ec: ExecutionContext
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/WebSocketService.scala
```diff
@@ -375,30 +443,40 @@ object WebSocketService {
...
+        lc: LoggingContextOf[InstanceUUID]
+    ): Future[StreamPredicate[Positive]] = {
 
       // invariant: every set is non-empty
-      val q: Map[domain.TemplateId.RequiredPkg, HashSet[LfV]] =
+      def getQ(
+          resolvedWithKey: Set[
+            (
+                com.daml.http.domain.TemplateId.RequiredPkg,
+                com.daml.http.query.ValuePredicate.LfV,
```
S11001001, line 457:
```suggestion
                domain.TemplateId.RequiredPkg,
                query.ValuePredicate.LfV,
```

ledger-service/http-json/src/main/scala/com/digitalasset/http/json/DomainJsonDecoder.scala
```diff
@@ -6,15 +6,23 @@ package com.daml.http.json
...
 import scalaz.syntax.show._
 import scalaz.syntax.std.option._
 import scalaz.syntax.traverse._
-import scalaz.{Traverse, \/, \/-}
+import scalaz.{EitherT, Traverse, \/, \/-}
+import scalaz.EitherT.{either, eitherT}
 import spray.json.{JsValue, JsonReader}
 
+import scala.concurrent.{ExecutionContext, Future}
+import scalaz.Scalaz.futureInstance
```
S11001001, line 23: Likewise regarding referring to implicits.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/pkg/PackageClient.scala
```diff
@@ -10,16 +10,16 @@ import com.daml.ledger.api.v1.package_service.PackageServiceGrpc.PackageServiceS
 import scala.concurrent.Future
 
 class PackageClient(val ledgerId: LedgerId, service: PackageServiceStub) {
-  private val it = new withoutledgerid.PackageClient(service)
+  private[daml] val it = new withoutledgerid.PackageClient(service)
```
S11001001, line 13: scared of that 👻

S11001001 (approved):
Looks reasonable to me, just some 🚲 🏚️

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
stefanobaghino-da: You taught me a Scala thing

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/pkg/PackageClient.scala
realvictorprm: That stuff needs adjustment, I'm not super happy right now :sweat:

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsService.scala
realvictorprm: ok, thanks :)!

ledger-service/http-json/src/main/scala/com/digitalasset/http/HttpService.scala
realvictorprm: good catch, thanks!

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/pkg/PackageClient.scala
realvictorprm: but that is for another day

PR #10604, opened 2021-08-17
Use the port file and dynamic port generation in client/server tests.

SamirTalwar-DA:
This creates a runner named `runner_with_port_file` which knows how to interpolate two variables, `%PORT_FILE%` and `%PORT%`. This allows us to use the `port-file` argument to the kvutils runner rather than hard-coding a port for conformance tests.

For now, we only use this for generating the kvutils reference ledger export.

This pull request brought to you by me getting annoyed that the assigned port was already taken, breaking my build on CI.

### Pull Request Checklist

trigger the build.


bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+    maybePort <- forM portFile $ \file -> do
+      putStrLn $ "Waiting for the port in the file \"" <> file <> "\" to open..."
+      waitForConnectionInPortFile (threadDelay 500000) file
+    let clientArgs' = case maybePort of
+                        Nothing -> splitArgs clientArgs
+                        Just port -> map (interpolateVariable "PORT" (show port)) (splitArgs clientArgs)
+    callProcess clientExe clientArgs'
+  forM_ portFile removeFile
+
+interpolateVariable :: String -> String -> String -> String
```
cocreature, line 36: we already depend on `extra` which has `replace` on String.

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+interpolateVariable variable replacement =
+  Text.unpack . Text.replace ("%" <> Text.pack variable <> "%") (Text.pack replacement) . Text.pack
+
+generatePortFile :: [String] -> IO (Maybe String, [String])
+generatePortFile args = do
+  if any (Text.isInfixOf "%PORT_FILE%" . Text.pack) args
```
cocreature, line 42: is there a good reason to not do this in all cases?

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+  let serverProc = proc serverExe serverArgs'
+  withCreateProcess serverProc $ \_stdin _stdout _stderr _ph -> do
+    maybePort <- forM portFile $ \file -> do
+      putStrLn $ "Waiting for the port in the file \"" <> file <> "\" to open..."
+      waitForConnectionInPortFile (threadDelay 500000) file
+    let clientArgs' = case maybePort of
+                        Nothing -> splitArgs clientArgs
+                        Just port -> map (interpolateVariable "PORT" (show port)) (splitArgs clientArgs)
+    callProcess clientExe clientArgs'
+  forM_ portFile removeFile
```
cocreature, line 34: this is not exception safe. If your code fails before this the file might persist. I recommend to use `withTempDir` to create a directory and then chose a file path in there. Then you also don’t have to worry about deleting the temp file you just created. We use that a fair bit throughout the codebase, e.g., https://github.com/digital-asset/daml/blob/0c1878530b5341e52f1eb5f107ce109ce5953f19/daml-assistant/daml-helper/src/DA/Daml/Helper/Start.hs#L103

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+waitForConnectionInPortFile sleep portFile = do
+  let hints = defaultHints { addrFlags = [AI_NUMERICHOST, AI_NUMERICSERV], addrSocketType = Stream }
+  untilJust $ do
+    eitherPortFileContents <- tryIO (Text.unpack . Text.strip . Text.pack <$> readFile portFile)
+    case eitherPortFileContents of
+      Left _ -> sleep *> pure Nothing
+      Right portFileContents ->
+        case reads @Int portFileContents of
+          [(port, "")] -> do
+            addr : _ <- getAddrInfo (Just hints) (Just "127.0.0.1") (Just $ show port)
```
cocreature, line 61: do we need this? the rest of our codebase assumes that the portfile is only written when the connection is available and afaik that has never caused issues so this seems like a sensible pattern to rely on.

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+
+waitForConnectionInPortFile :: IO () -> FilePath -> IO Int
+waitForConnectionInPortFile sleep portFile = do
+  let hints = defaultHints { addrFlags = [AI_NUMERICHOST, AI_NUMERICSERV], addrSocketType = Stream }
+  untilJust $ do
+    eitherPortFileContents <- tryIO (Text.unpack . Text.strip . Text.pack <$> readFile portFile)
+    case eitherPortFileContents of
+      Left _ -> sleep *> pure Nothing
+      Right portFileContents ->
+        case reads @Int portFileContents of
```
cocreature, line 59: `readMay` might be a better choice and you don’t have to strip for that to work so you can simplifyf the code a bit. See how `readPortFile` works.

cocreature: or maybe try actually using that here if possible.

ledger/participant-state/kvutils/BUILD.bazel
```diff
@@ -238,8 +238,6 @@ proto_jars(
 
 REFERENCE_LEDGER_EXPORT_NAME = "reference-ledger-export"
 
-REFERENCE_LEDGER_EXPORT_PORT = 65102
```
arobert-da, line 241: Why not just move this out of emphemeral range and be done? Want to "solve it for once and for all" to avoid chance of conflict with other fixed ports?

cocreature: I’m not sure what you’re suggesting? In this PR, @SamirTalwar-DA switched to using port 0 which choses a free open port so there is no chance of it colliding with anything.

arobert-da: Correct me if I'm wrong, but the collision observed in practice was likely due to the 65102 being taken by an outgoing connection, whereas if it were out of the ephemeral range (i.e., in 1025-32767) this, at least, would not happen. I just wanted to understand the motivation for going the extra mile.

cocreature: I don’t know how Samir ended up with the collision but generally just making sure that our own tests don’t conflict with each other is already painful and we have a lot of tests that cannot run in parallel with other tests exactly because they hardcode the same port number. I’m absolutely in favor of using port 0 everywhere and I think the complexity for this is justified.

bazel_tools/client_server/runner_with_port_file/Main.hs
SamirTalwar-DA: No, but we use this information to decide whether to wait for the port to open.

That said, we can probably just delete that behavior, as the Ledger API Test Tool also waits.

bazel_tools/client_server/runner_with_port_file/Main.hs
SamirTalwar-DA: That makes sense; I'll do that. Thanks.

ledger/participant-state/kvutils/BUILD.bazel
SamirTalwar-DA: Yeah, I started by changing the port and then decided to go overboard. I am happy to go with the simpler change but my real intent with this is to use it across the conformance tests so they don't accidentally collide.

It really bugs me when I have a ledger running locally for testing and then try running some conformance tests, only to watch them fail because they picked the same default port.

arobert-da: Right.. local running, parallel tests, oh my – thanks for the explanations. Now I understand a bit about the differences between testing here and in LR. :-)

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+import Control.Monad.Loops (untilJust)
+import Data.List.Split (splitOn)
+import Network.Socket
+import System.Directory (removeFile)
+import System.Environment
+import System.IO.Temp (emptySystemTempFile)
+import System.Process
+import qualified Data.Text as Text
+
+main :: IO ()
```
aherrmann-da, line 20: Taking the other feedback into account, about only waiting on the port-file, but not also the connection, how significantly does this runner still differ from [`//bazel_tools/client_server/runner`](https://github.com/digital-asset/daml/blob/64abf8aaa9f63386a33182fc6a9e77da60ff0000/bazel_tools/client_server/runner/Main.hs)?
That other runner also supports port 0 and a port-file, it also waits for the creation of that port-file.
One difference that stands out is that the other runner hard-codes the shape of the `--port-file` and `--port` arguments. If that's the only remaining difference, perhaps the other runner could be generalized to make these args configurable instead of adding an additional runner?

SamirTalwar-DA: You're right, they're similar. My goal is to eventually merge all three runners together. However, I wanted to start small and just change a single target.

aherrmann-da: Thanks for clarifying. :+1: Makes sense.

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,38 @@
...
+  withSystemTempDirectory "runner" $ \tempDir -> do
+    (portFile, interpolatedServerArgs) <-
+      if any (isJust . stripInfix "%PORT_FILE%") splitServerArgs
+        then do
+          portFile <- emptyTempFile tempDir "port"
+          removeFile portFile
+          let interpolatedArgs = map (replace "%PORT_FILE%" portFile) splitServerArgs
+          return (Just portFile, interpolatedArgs)
+        else
+          return (Nothing, splitServerArgs)
```
cocreature, line 29:
```suggestion
  withTempDir "runner" $ \tempDir -> do
    (portFile, interpolatedServerArgs) <-
      if any (isJust . stripInfix "%PORT_FILE%") splitServerArgs
        then do
          let portFile = tempDir </> "portfile"
          let interpolatedArgs = map (replace "%PORT_FILE%" portFile) splitServerArgs
          return (Just portFile, interpolatedArgs)
        else
          return (Nothing, splitServerArgs)
```

bazel_tools/client_server/runner_with_port_file/Main.hs
```diff
@@ -0,0 +1,38 @@
...
+    let serverProc = proc serverExe interpolatedServerArgs
+    withCreateProcess serverProc $ \_stdin _stdout _stderr _ph -> do
+      maybePort <- mapM (readPortFile maxRetries) portFile
+      let splitClientArgs = splitArgs clientArgs
+      let interpolatedClientArgs =
+            case maybePort of
+              Nothing -> splitClientArgs
+              Just port -> map (replace "%PORT%" (show port)) splitClientArgs
```
cocreature, line 37: just always doing that now that you dropped the wait seems like it might simplify things a bit

cocreature (approved):
Thanks! two small comments

fabiotudone-da (approved):

bazel_tools/client_server/runner_with_port_file/Main.hs
SamirTalwar-DA: Dropped the wait, sure, but we still have no guarantee that the server actually wrote a port file.

I kept this as a `Maybe` because it allows us to use one runner everywhere. Do you think we're better off only using it where we definitely have a port file?

cocreature: Before your PR we have two types of client_server runners:

1. Based on port files which waits for the server to write a port file.
2. Based on hardcoded ports which tests the connection until the server is up.

The client needs to know the port the server is listening on so if you don’t get it from the port file you don’t have a lot of other options than hardcoding it in some form.

I understood this PR as introducing a variation of 1 where the port file is specified differently meaning it doesn’t make sense to use it if the server does not write a port file. It sounds like you have something different in mind but I think I don’t quite understand what.

SamirTalwar-DA: My end goal is that we have one runner. I can see an argument for one (everything is done through Bazel configuration), and an argument for two (two different ways of starting the server).

Which would you prefer?

cocreature: Definitely in favor of one runner but that doesn’t really answer the question I clearly failed to ask properly :)

We currently have waiting for a port file and waiting for a connection. You’re adding a third option in here where we wait for nothing. What do we need this for? In my ideal world everything runs on ephemeral ports for which we need the port file (otherwise we can’t tell the client tell what to connect to). Where that’s not possible, we hardcode and wait for the connection. I don’t see the need for the third option.

SamirTalwar-DA: I think I was getting ahead of myself. For now I'll simplify, though I think this logic will need to come back if we merge this runner with `runner_with_port_check`.

PR #10608, opened 2021-08-18
Remove GenMissingString class because it is not used

pbatko-da:


cocreature (approved):
deleted code is the best code

PR #10611, opened 2021-08-18
fix gsg-trigger template

garyverhaegen-da:
It currently has all it needs for the docs to build, but not quite for people to use it as a useful result of the triggers docs page: without the UI components, it's hard to see anything the triggers are supposed to do.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/templates/gsg-trigger.patch
+++ b/templates/gsg-trigger.patch
@@ -1,6 +1,6 @@
 diff -Naur create-daml-app/daml/ChatBot.daml gsg-trigger/daml/ChatBot.daml
 --- create-daml-app/daml/ChatBot.daml	1970-01-01 01:00:00.000000000 +0100
-+++ gsg-trigger/daml/ChatBot.daml	2021-08-16 13:05:51.000000000 +0200
++++ gsg-trigger/daml/ChatBot.daml	2021-08-18 13:15:25.000000000 +0200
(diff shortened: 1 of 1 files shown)


templates/gsg-trigger.patch
```diff
@@ -1,6 +1,6 @@
 diff -Naur create-daml-app/daml/ChatBot.daml gsg-trigger/daml/ChatBot.daml
 --- create-daml-app/daml/ChatBot.daml	1970-01-01 01:00:00.000000000 +0100
-+++ gsg-trigger/daml/ChatBot.daml	2021-08-16 13:05:51.000000000 +0200
++++ gsg-trigger/daml/ChatBot.daml	2021-08-18 13:15:25.000000000 +0200
```
cocreature, line 3: instead of inlining this into this patch could we first apply `messaging.patch` and then this one to avoid the two getting out of sync?

garyverhaegen-da: I wish we could, but `messaging.patch` applies on a realized version of the template, not on the template itself. Fixing that seemed like a bit more effort than I thought reasonable to put into this at this point.

cocreature: makes sense, thanks for the explanation

cocreature (approved):
Thanks!

PR #10613, opened 2021-08-18
participant-integration-api: Store the rejection status as 3 columns.

SamirTalwar-DA:
Serializing the details but keeping the code and message columns populated.


ledger/participant-integration-api/BUILD.bazel
```diff
@@ -16,7 +17,20 @@ load(
...
+    name = "participant-integration-api-proto",
+    srcs = glob(["src/main/protobuf/**/*.proto"]),
+    maven_artifact_prefix = "participant-integration-api",
+    maven_group = "com.daml",
+    strip_import_prefix = "src/main/protobuf",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@com_google_protobuf//:any_proto",
+    ],
+)
```
meiersi-da, line 30: Thank you @SamirTalwar for adding this. Is there an easy way to add `buf` checks for these protos?

SamirTalwar-DA: What kind of checks would you like? Backwards-compatibility checks against each released version?

I have suggested in the past that we add these as part of the _compatibility_ checks we run on every PR. However, that is a big piece of work and needs to be scoped and scheduled accordingly.

We do run some `buf` checks against a checked-in image, which I am not a fan of, because it doesn't provide any guarantees across 3 or more versions.

meiersi-da: I was thinking of backwards compatibility checks against released versions. It's OK though if the enabler for this work is not yet in place.

In that case I'd just add a BIG FAT disclaimer to this proto that all changes must be backwards compatible.

SamirTalwar-DA: Makes sense; I will make it clear that the protobuf is used for storage.

nmarton-da (approved):
LGTM

PR #10615, opened 2021-08-18
bazel-tools: Replace `runner` with either `runner_with_port_check` or `runner_with_port_file`.

SamirTalwar-DA:
Then there were 2.

This replaces `runner` with the `runner_with_port_check` for Canton tests (because it uses multiple ports, which we don't support in a port file for now), and `runner_with_port_file` everywhere else.

Conformance tests still use `runner_with_port_check`, which I'd like to stop doing, but we need to handle multiple ports first.

### Pull Request Checklist

trigger the build.


aherrmann-da (approved):
Looks good

PR #10617, opened 2021-08-18
detect unsynchronized contract table and retry

### LINKED ISSUE #10334: handle concurrent-update race condition for DB-backed websocket queries
@S11001001 suspects/knows that we currently do not work correctly at read-committed level which is the default in both Oracle & PostgreSQL. I don’t quite remember the problematic case so @S11001001 please leave a comment :)

We should either change the serialization level (probably acceptable at least short term) or fix our code to handle read-committed.

S11001001:
After the websocket DB query, we check the offset table to ensure that it's consistent with what we just wrote to it in the fetch-and-persist phase. If it isn't, we try again. (We can always use the offset that we just wrote in our final response body, because if it doesn't match, we're going to use the new offset and re-fetch anyway.)

We try 10 times, and if we still don't get consistent results, we fail the WS request. I haven't explored more complex backoff strategies here; the focus of this PR is getting a minimal update that is least likely to livelock (and thus exhaust all retries).

We _could_ also use the `fetchAndPersistBracket` function for the non-WS queries. I'm not sure that's necessary, though.

Fixes #10334. See that issue for a longer explanation of the possible race conditions this PR addresses.

```rst
CHANGELOG_BEGIN
- [JSON API] Under rare conditions, a multi-template query backed by database
  could have an ACS portion that doesn't match its transaction stream, if
  updated concurrently.  This conditions is now checked and accounted for.
  See `issue #10617 <https://github.com/digital-asset/daml/pull/10617>`__.
CHANGELOG_END
```

* [x] get rid of XTag/XX unless I need it for a different `minimumViableOffsets` approach

### Pull Request Checklist

trigger the build.


stefanobaghino-da (commented):
How will this impact performance? Is there some visible effect on existing performance tests?

ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
```diff
@@ -143,6 +148,126 @@ object ContractDao {
...
+  ): Option[(Off, Set[OTpId])] = {
+    import scalaz.syntax.foldable1.{ToFoldableOps => _, _}
+    import scalaz.std.iterable._
+    val lagging: Option[Lagginess[ITpId, Off]] =
+      (unsynced: Iterable[(ITpId, Map[Party, Off])]) foldMap1Opt {
+        case (surrogateTpId, partyOffs) =>
+          val maxLocalOff = partyOffs
+            .collect {
+              case (unsyncedParty, unsyncedOff)
+                  if queriedParty(unsyncedParty) || unsyncedOff > expectedOffset =>
```
akshayshirahatti-da, line 212: what should be the behaviour for an `unsyncedParty` which is not a `queriedParty` and has `unsyncedOff < expectedOffset` ?

ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
```diff
@@ -143,6 +148,126 @@ object ContractDao {
...
+                unsyncedOff
+            }
+            .maximum
+            .getOrElse(expectedOffset)
+          val singleton = Set(surrogateTpId)
+          // if a queried party is not in the map, we can safely assume that
+          // it is exactly at expectedOffset
+          val caughtUp = maxLocalOff <= expectedOffset ||
```
akshayshirahatti-da, line 220: again slightly confused on how `maxLocalOff` being < `expectedOffset` is considered as caughtUp

ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
S11001001: We can safely ignore it. This is the test `"not require update for lagging, but unqueried, party"`.

ledger-service/http-json/src/main/scala/com/digitalasset/http/dbbackend/ContractDao.scala
S11001001: We have the invariant that every `queriedParty` is already at least `expectedOffset`, but it may not be in the map. Therefore, if the condition you mention holds, then `maxLocalOff` _must_ have come from an unqueried party, which is fine as stated in the prior comment.

S11001001: Put another way, we care only that the `queriedParty` is caught up.

akshayshirahatti-da (approved):
functionally looks fine to me, thanks for explaining 👍

ledger-service/http-json/src/test/scala/com/digitalasset/http/dbbackend/ContractDaoTest.scala
```diff
@@ -0,0 +1,90 @@
...
+    "not require update for lagging, but unqueried, party" in {
+      mvo(3, Map(0 -> Map((-4: Byte) -> 1))) should ===(None)
+    }
+
+    "require update for ahead, albeit unqueried, party" in {
+      mvo(3, Map(0 -> Map((-4: Byte) -> 4))) should ===(Some((4, Set(0))))
+    }
+
+    "report desync, but no updates, if consistent" in {
+      mvo(3, Map(0 -> everyParty(5), 1 -> everyParty(5))) should ===(
```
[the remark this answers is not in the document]
S11001001: `expectedOffset` is only "what we just wrote"; it never comes from reading. So this can happen if all these template-IDs/parties got updated from 3 to 5 after our own fetch.

S11001001:
> How will this impact performance? Is there some visible effect on existing performance tests?

@stefanobaghino-da There's no visible effect on `AsyncQueryConstantAcs` between 9641fd5f83 and 5da2571c14. The `ContractDao` benchmarks don't use `ContractsFetch` so have no code changes.

A couple things on concurrent updates:

1. If we rerun, we _don't_ rollback; we read forward as if we were just starting the request. So successive retries won't redo any fetching that was already done in previous tries.
2. If you want to follow the common (no concurrent update) case while thinking about the code, it's as follows:
    1. `Queries#unsyncedOffsets` returns a map whose keys are all the template IDs and values are empty maps; the query it does has no rows.
    2. In `minimumViableOffsets`,
        1. `partyOffs` always has only unqueried parties;
        2. `maxLocalOff` is always `expectedOffset`;
        3. `caughtUp` is always true by the first branch;
        4. `lagging` is always `Some`, but `minimumViableOffsets` will return `None`.
    3. In the `Lagginess` append,
        1. `offA === offB`;
        2. `nc`, `ncA`, and `ncB` are empty;
        3. `c` accumulates all the surrogate template IDs we're querying.

PR #10619, opened 2021-08-19
Augment `completion.proto` with deduplication-related info [KVL-1057]

fabiotudone-da:
## Description

This is the first change in a series to let the Ledger API provide more command de-duplication information in completions (only the submission rank and deduplication offsets are deferred to later stages). It updates the proto definition of a completion (courtesy of @SamirTalwar-DA) and leaves the new fields blank.

Based on the original deduplication Ledger API draft #9567.

- [x] Ideally merged after #10625.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


fabiotudone-da:
Opening for review because it builds from `main`; I currently target #10625 because I'd like to make sure that the changes are backward-compatible.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -29,7 +29,13 @@ private[platform] object CompletionFromTransaction {
   ): CompletionStreamResponse =
     CompletionStreamResponse.of(
       checkpoint = Some(toApiCheckpoint(recordTime, offset)),
-      completions = Seq(Completion.of(commandId, Some(OkStatus), transactionId)),
+      completions = Seq(
+        Completion().update(
```
SamirTalwar-DA, line 33: I'd prefer `Completion.of` with the relevant fields set to empty values, so it's explicit.

If you're very much against this, you can just do `Completion(commandId, OkStatus, transactionId)`, as that constructor takes default parameters.

fabiotudone-da: I agree about being explicit, considering that adding TODOs would be bad practice.

nicu-da: I think it might be worth also considering disabling adding defaults to the constructor from `scalapb`. That way if you add a new field you have to explictly add it everywhere. And in tests you can always use `Type.defaultInstance`

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,40 @@ message Completion {
...
+  // The actual deduplication window used for the submission, which is derived from
+  // ``Commands.deduplication_period``. The ledger may convert the deduplication period into other
+  // descriptions and extend the period in implementation-specified ways.
+  //
+  // Used to audit the deduplication guarantee described in ``commands.proto``.
+  //
+  // Optional; the deduplication guarantee applies even if the completion omits this field.
+  oneof deduplication_period {
+    // Specifies the length of the deduplication period.
+    // It is interpreted relative to the local clock at some point during the submission's processing.
```
andreaslochbihler-da, line 61: For submission, we don't specify the clock, but for completions we do want to so that one can audit the completion stream.
```suggestion
    // It is interpreted on the record time of completions
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,40 @@ message Completion {
...
+    //
+    // Must be non-negative.
+    google.protobuf.Duration deduplication_time = 11;
+
+    // Specifies a point in time in the past after which this submission should be de-duplicated w.r.t. previous submissions of the same commands.
+    // It is interpreted relative to the local clock at some point during the submission's processing.
+    google.protobuf.Timestamp deduplication_start = 12;
```
andreaslochbihler-da, line 68: What has happened to the `deduplication_offset`?

Currently, `DeduplicationPeriod` in `state.v2` offers only `DeduplicationDuration` and `DeduplicationOffset`. Why do we have in completions now duration and timestamp, but not start?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,40 @@ message Completion {
...
+  // Optional; the deduplication guarantee applies even if the completion omits this field.
+  oneof deduplication_period {
+    // Specifies the length of the deduplication period.
+    // It is interpreted relative to the local clock at some point during the submission's processing.
+    //
+    // Must be non-negative.
+    google.protobuf.Duration deduplication_time = 11;
+
+    // Specifies a point in time in the past after which this submission should be de-duplicated w.r.t. previous submissions of the same commands.
+    // It is interpreted relative to the local clock at some point during the submission's processing.
```
andreaslochbihler-da, line 67: Same as for deduplication time.
```suggestion
    // It is measured on the record time of completions.
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
andreaslochbihler-da: See https://github.com/digital-asset/daml/blob/main/ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/DeduplicationPeriod.scala#L25

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,40 @@ message Completion {
...
+  //
+  // Used to audit the deduplication guarantee described in ``commands.proto``.
+  //
+  // Optional; the deduplication guarantee applies even if the completion omits this field.
+  oneof deduplication_period {
+    // Specifies the length of the deduplication period.
+    // It is interpreted relative to the local clock at some point during the submission's processing.
+    //
+    // Must be non-negative.
+    google.protobuf.Duration deduplication_time = 11;
```
andreaslochbihler-da, line 64: Just out of curiosity: Why do we skip field numbers 7 to 10?

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -29,7 +29,13 @@ private[platform] object CompletionFromTransaction {
   ): CompletionStreamResponse =
```
andreaslochbihler-da, line 29: Why does this method not simply take a `CompletionInfo` as an argument? Then we'd have all the current and future extensions automatically available (submission ID, submission rank, ...) so that we can set the fields in the `Completion` message.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
SamirTalwar-DA: Didn't we agree that we'd not worry about deduplication offsets for now?

andreaslochbihler-da: No. We agreed on not worrying about deduplication starting points for now.

SamirTalwar-DA: Hmmm, there was clearly a miscommunication here. Let's take this to a different channel.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
fabiotudone-da: > I think it might be worth also considering disabling adding defaults to the constructor from scalapb

I agree it's a good idea but for a separate PR.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
fabiotudone-da: This is easier to do in a subsequent PR, after the DB schemas have been updated.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -40,7 +50,17 @@ private[platform] object CompletionFromTransaction {
...
-      completions = Seq(Completion.of(commandId, Some(status), RejectionTransactionId)),
+      completions = Seq(
+        Completion.of(
+          commandId = commandId,
+          status = Some(status),
+          transactionId = RejectionTransactionId,
+          applicationId = null,
+          actAs = null,
+          submissionId = null,
+          deduplicationPeriod = null,
```
nicu-da, line 61: In this case I wouldn't use the `.of` constructor as we set a subset of the fields. Instead I would use the default constructor `Completion(
commandId = commandId,
...
)`.

fabiotudone-da: I thought we agreed that being explicit also about the fields not being set is actually better, why not using `.of` then?

nicu-da: I agree with being explicit if we set all or most fields, but in this case it seems we set a minority of the fields so I find the explicit `of` to actually make things muddier.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
fabiotudone-da: Added 89336ab

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
fabiotudone-da: d566b97

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
fabiotudone-da: d566b97

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
nicu-da: If not then I'm not sure if it works out with null. We would need to pass the default for each field type, eg: "" for String, Seq.empty for repeated elems, DeduplicationPeriod.Empty for the `oneOf`

nicu-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
fabiotudone-da: 1ba2045

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
SamirTalwar-DA: I would recommend at least leaving a gap for `submission_rank`.

fabiotudone-da: I'd like to justify it, maybe a `reserved`?

fabiotudone-da: efa01ff

SamirTalwar-DA (approved):

fabiotudone-da:
@digital-asset/kv-participant I am going to merge as soon as the build is green, let me know if you have remarks.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,50 @@ message Completion {
   // Must be a valid LedgerString (as described in ``value.proto``).
   // Optional
   string transaction_id = 3;
+
+  // The application ID that was used for the submission, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional for historic completions where this data is not available.
```
fabiotudone-da, line 37: @SamirTalwar-DA Isn't it optional for an application to set it in the submission?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
```diff
@@ -29,4 +31,50 @@ message Completion {
...
+  // The set of parties on whose behalf the commands were executed.
+  // Contains the union of ``party`` and ``act_as`` from ``commands.proto``.
+  // The order of the parties need not be the same as in the submission.
+  // Each element must be a valid PartyIdString (as described in ``value.proto``).
+  // Optional for historic completions where this data is not available.
+  repeated string act_as = 5;
+
+  // The submission ID this completion refers to, as described in ``commands.proto``.
+  // Must be a valid LedgerString (as described in ``value.proto``).
+  // Optional for historic completions where this data is not available.
```
fabiotudone-da, line 49: @SamirTalwar-DA Isn't it optional for an application to set it in the submission?

fabiotudone-da: I think it is set by the participant in that case, though. Is that right?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
fabiotudone-da: Ignore, I've checked and it's required.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/completion.proto
fabiotudone-da: Checked, the participant or the committer, so the participant will have it (starting with a certain SDK). The comment is correct.

PR #10621, opened 2021-08-19
Pruning offset comparisons should not log errors [DPP-557]

tudor-da:
Debatable, but the SQL query execution can lead to soft errors, hence the log level change. 
Far from ideal to log everything as WARNs, since query execution errors can be due to programming errors as well, which should definitely be ERROR.

Eager to learn about others' opinions

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DbDispatcher.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DbDispatcher.scala
@@ -56,7 +56,7 @@ private[platform] final class DbDispatcher private (
           result
         } catch {
           case NonFatal(e) =>
-            logger.error("Exception while executing SQL query. Rolled back.", e)
+            logger.warn("Exception while executing SQL query. Rolled back.", e)
             throw e
           // fatal errors don't make it for some reason to the setUncaughtExceptionHandler
           case t: Throwable =>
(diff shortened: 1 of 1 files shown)


SamirTalwar-DA:
I'm not sure this is the "right" change. I would prefer we catch the error before executing the SQL at all.

I know this is a concern with participant pruning; can't we validate first? Are there cases where race conditions can cause SQL errors?

tudor-da:
> I'm not sure this is the "right" change. I would prefer we catch the error before executing the SQL at all.
> 
> I know this is a concern with participant pruning; can't we validate first? Are there cases where race conditions can cause SQL errors?

I agree, thought better about it and actually most of the errors coming here should be programming errors. Moving it to draft

nmarton-da:
> > I'm not sure this is the "right" change. I would prefer we catch the error before executing the SQL at all.
> > I know this is a concern with participant pruning; can't we validate first? Are there cases where race conditions can cause SQL errors?
> 
> I agree, thought better about it and actually most of the errors coming here should be programming errors. Moving it to draft

I second the concern. How about keeping the log levels here, and doing something special for the new code like catching exceptions there, packing result in an either for client evaluation, re-throwing unknown ones, but swallowing known ones...

SamirTalwar-DA:
What's the status of this?

tudor-da:
Not needed anymore. This logs as INFO now.

PR #10623, opened 2021-08-19
Make Index DB enable multiple party additions [DPP-546]

nmarton-da:
**Preparation, small fixes**

* Remove unnecessary TODO
* Fix FieldStrategy.idempotentInsert type parameters

**Changing the DB-Schema**

This change adapts the DB schema for all supported backends. After this change we only populate the party_entries table, and on the query side we reconstruct the state from this.

* Drop the party table
* Add indexing to party_entries

**Adapting StorageBackend: ingestion**

Since we only ingest party_entries, the party population needs to be removed.

* Drop the party table in ingestion code
* Fixes test

**Adapting StorageBackend: queries**

Queries needs to be adapted to construct the state from the read side.

* Rewrite queries.
* Fixes reset implementations.

**Adapting JdbcLedgerDao**

Since underlying storage changed, JdbcLedgerDao can be simplified: no special treatment needed with duplicate errors, since these errors are impossible to happen.
Removing JdbcLedgerDao tests, and adding a new test, testing the behavior of the new event-source party model. Please note: this database refactoring only applies to the append-only schema, so for the mutating schema the test is disabled.
During implementation a bug surfaced: it was not possible anymore to store the is_local information via JdbcLedgerDao.storePartyEntry. Although this bug is a minor issue, since that method is only used from single participant environment, still a fix was also implemented for this, by passing a magic participantId upon non-local party storage.

* Simplify storePartyEntry.
* Fixes bug introduced by append-only.
* adds/adapts tests

**Refactoring: remove not used duplicateKeyError from StorageBackend**

Changes to JdbcLedgerDao rendered this duplicateKeyError unused.

* Removes unused duplicateKeyError

**Adapting sandbox-classic**

In sandbox-classic it is not allowed to have updates for parties. Essentially the updates concerning already existent parties were dropped silently with logging without effect.
Here I started by pinning down this behaviour in the SqlLedgerSpec and SqlLedgerSpecAppendOnly. These tests were implemented with the original code in mind.
Then adapted the SqlLedger method: making sure of uniqueness by first trying to lookup the to-be-persisted party.

* Added tests grabbing a hold on original behavior
* Adapted implementation to ensure same behavior

**Switching to correct is_local derivation for party queries as per review**

* Adapting implementation: switching to aggregated table and a query on that
* Introducing QueryStrategy.booleanOrAggregationFunction to support Oracle
* Moving party related queries to PartyStorageBackendTemplate
* Fixes JdbcLedgerDaoPartiesSpec tests, and add another test case

**Also:**

* Align Update interface documentation
* Switching to explicit optionality in party query implementation asa per review

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -383,16 +382,31 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
+        |    row_number() over (partition by party order by ledger_offset desc) as rn
+        |  from party_entries, parameters
+        |  where
+        |    party_entries.party in ({parties})
+        |    and party_entries.typ = 'accept'
+        |    and party_entries.ledger_offset <= parameters.ledger_end
+        |) as parties
+        |where
+        |  rn = 1
+        |""".stripMargin)
```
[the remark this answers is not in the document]
meiersi-da: Consider specializing to https://www.postgresql.org/docs/9.6/functions-window.html, LAST_VALUE

consider why we need the offset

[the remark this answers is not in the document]
meiersi-da: drop offset

meiersi-da (commented):
Comments from our f2f discussion

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -419,10 +432,24 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
+        |    party_entries.is_local as is_local,
+        |    row_number() over (partition by party order by ledger_offset desc) as rn
+        |  from party_entries, parameters
+        |  where
+        |    party_entries.typ = 'accept'
+        |    and party_entries.ledger_offset <= parameters.ledger_end
+        |) as parties
+        |where
+        |  parties.rn = 1
+        |""".stripMargin)
```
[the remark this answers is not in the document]
nmarton-da: solved by DRY

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -223,40 +221,31 @@ private class JdbcLedgerDao(
       }
     }
 
+  private val nonLocalParticipantId = Ref.ParticipantId.assertFromString("non-local-participant-id")
```
meiersi-da, line 224:
```suggestion
  private val nonLocalParticipantId = Ref.ParticipantId.assertFromString("__HACK__non-local-participant-id")
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -223,40 +221,31 @@ private class JdbcLedgerDao(
...
-              logger.warn(
-                s"Ignoring duplicate party submission with ID ${partyDetails.party} for submissionId $submissionIdOpt"
+          sequentialIndexer.store(
+            conn,
+            offset,
+            Some(
+              state.Update.PartyAddedToParticipant(
+                party = partyDetails.party,
+                displayName = partyDetails.displayName.orNull,
+                participantId = if (partyDetails.isLocal) participantId else nonLocalParticipantId,
```
meiersi-da, line 242:
```suggestion
                // HACK: the `PartyAddedToParticipant` transmits `participantId`s, while here we only have the information
                // whether the party is locally hosted or not. We use the `nonLocalParticipantId` to get the desired effect of
                // the `isLocal = False` information to be transmitted via a `PartyAddedToParticpant` `Update`.
                //
                // This will be properly resolved once we move away from the `sandbox-classic` codebase.
                participantId = if (partyDetails.isLocal) participantId else nonLocalParticipantId,
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -383,51 +383,50 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
-
-  def parties(parties: Seq[Ref.Party])(connection: Connection): List[PartyDetails] = {
-    import com.daml.platform.store.Conversions.partyToStatement
-    SQL_SELECT_MULTIPLE_PARTIES
-      .on("parties" -> parties)
-      .as(PartyDataParser.*)(connection)
-      .map(constructPartyDetails)
+  private def queryParties(parties: Set[String], connection: Connection): Vector[PartyDetails] = {
+    val partyFilter =
+      if (parties.isEmpty) cSQL""
```
meiersi-da, line 401: What about using `Option[Set[String]]` instead of redefining the meaning of filtering for the empty set?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -383,51 +383,50 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
...
+    SQL"""
+        SELECT
+          parties.party,
+          parties.display_name,
+          parties.is_local
+        FROM (
+          SELECT
+            party_entries.party party,
+            party_entries.display_name display_name,
+            party_entries.is_local is_local,
```
meiersi-da, line 412: Important: As we are retrieving `is_local` here and that one has a set-semantics, we must use `bool_or(party_entries.is_local)` to return the correct information wrt locality.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
```diff
@@ -172,33 +158,60 @@ private[dao] trait JdbcLedgerDaoPartiesSpec {
...
+        partyEntries <- ledgerDao
+          .getPartyEntries(beforeStartOffset, nextOffset())
+          .runWith(Sink.collection)
+      } yield {
+        parties shouldBe List(dan4)
+        val partyEntriesMap = partyEntries.toMap
+        partyEntriesMap(firstOffset).asInstanceOf[AllocationAccepted].partyDetails shouldBe dan
+        partyEntriesMap(secondOffset).asInstanceOf[AllocationAccepted].partyDetails shouldBe dan2
+        partyEntriesMap(thirdOffset).asInstanceOf[AllocationAccepted].partyDetails shouldBe dan3
+        partyEntriesMap(lastOffset).asInstanceOf[AllocationAccepted].partyDetails shouldBe dan4
```
meiersi-da, line 203: Here we would expect that `dan` is still considered to be locally hosted.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/Update.scala
```diff
@@ -93,8 +93,11 @@ object Update {
   }
 
   /** Signal that a party is hosted at a participant.
+    * There might be multiple such events for the same party, in this case always the last one is the current one.
+    * Only exception from this rule is in case of participantId: once a party is added to a participant, it cannot be removed from there.
+    * (Which makes the paricipantId an ever extending set)
```
meiersi-da, line 98:
```suggestion
    * 
    * Repeated `PartyAddedToParticipant` updates are interpreted in the order of their offsets as follows:
    * - last-write-wins semantics for `displayName`
    * - set-union semantics for `participantId`; i.e., parties can only be added to, but not removed from a participant
    * The `recordTime` and `submissionId` are always metadata for their specific `PartyAddedToParticipant` update.
```

meiersi-da (commented):
LGTM, up to one important query change required: `is_local` needs to be retrieved with a set-semantics using `bool_or`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: I was thinking about that, but it seemed to me the empty set is anyhow an invalid case here. Since you mentioned, I am changing it.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
```diff
@@ -143,21 +144,6 @@ private[dao] trait JdbcLedgerDaoPartiesSpec {
     }
   }
 
-  it should "inform the caller if they try to write a duplicate party" in {
```
mziolekda, line 146: Shouldn't this test be retained for old schema?

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: ouch, complete misunderstanding from my part: I was thinking we should retrieve the latest.
That complicates query big time in complexity and performance.
Not sure there is a bool_or for Oracle, probably we need something more portable.

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
nmarton-da: oh, yeah! Re-added those with conditions: they will be only running for mutable schema!

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -223,40 +221,37 @@ private class JdbcLedgerDao(
       }
     }
 
+  private val nonLocalParticipantId =
+    Ref.ParticipantId.assertFromString("__HACK__non-local-participant-id")
```
mziolekda, line 225: HACK is a scary word, and probably to harsh, I would suggest saying RESTRICTED_NON_LOCAL_PARTICIPANT_ID. Effectively this is now a restricted string.

Should we maybe move it to the package object?

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlLedger.scala
```diff
@@ -484,19 +484,33 @@ private final class SqlLedger(
...
-            timeProvider.getCurrentTime,
-            PartyDetails(party, displayName, isLocal = true),
-          ),
-        )
-        .map(_ => ())(DEC)
-        .recover { case t =>
-          //recovering from the failure so the persistence stream doesn't die
-          logger.error(s"Failed to persist party $party with offset: ${offset.toApiString}", t)
-          ()
+        .getParties(Seq(party))
```
mziolekda, line 487: Do you have an idea what's the perf hit of resolving the conflict by hand in this manner. I am asking because daml-hub (and others) rely on several k of parties being created in a span of first few minutes of a ledger run

mziolekda: Actually running conformance test `PartyManagementServiceIT:PMAllocateOneHundred` should give us an idea

mziolekda (approved):
Few minor comments, otherwise ok

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: as discussed offline with @meiersi-da , we switched to group-by aggregated table selects, and proper implementation for is_local

ledger/participant-integration-api/src/test/lib/scala/platform/store/dao/JdbcLedgerDaoPartiesSpec.scala
nmarton-da: fixed

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
nmarton-da: I am fine with the placement and with the name, but also fine with restricted. What do you think @meiersi-da , @mziolekda ?

meiersi-da: Let's use @mziolekda 's proposal.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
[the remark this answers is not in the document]
nmarton-da: this is outdated with the lates queries with bool_or

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/sql/SqlLedger.scala
nmarton-da: that is a good point, thanks for raising it @mziolekda 
I did some quick measurements locally, these are the total runtimes of that particular test you mentioned int the interesting 4 combinations:
```
postgres, mutable, with query: 1113ms, 1124ms, 965ms
postgres, mutable, without query: 952ms, 1025ms, 939ms
postgres, append-only, with query: 1055ms, 1050ms, 964ms
postgres, append-only, without query: 1017ms, 951ms, 987ms
```

For my part I am comfortable with this level of barely noticeable degradation. @mziolekda @meiersi-da  do you think we should take further precaution wrt this?

(Please note after the PR is in the final state I plan to run a benchtool LR performance verification)

mziolekda: This looks ok!

meiersi-da (approved):
:beers: :hamburger: and a :heavy_check_mark: for @nmarton-da :smile: 

enjoy :raised_hands:

PR #10625, opened 2021-08-19
Ledger API: add `buf` checks [KVL-1045]

fabiotudone-da:
Depends on #10624.

### Pull Request Checklist

trigger the build.


fmt.sh
```diff
@@ -156,6 +156,7 @@ run scalafmt "${scalafmt_args[@]:-}"
 run bazel run "$buildifier_target"
 
 # Check for breaking proto changes.
+run buf mod update
```
SamirTalwar-DA, line 159: What does this command do? I can't find the relevant docs.

fabiotudone-da: Found it [here](https://docs.buf.build/tour/add-a-dependency/#depend-on-googleapis).

fabiotudone-da: This is needed because `completion.proto` depends on gRPC's `Status` and that's in `googleapis`.

miklos-da (approved):

Issue #10504, opened 2021-08-06
Contract IDs freshness is not possible. 


remyhaemmerle-da:
In general the contract ID discriminator freshness based on input/output as described in the [contract ID specification](https://github.com/digital-asset/daml/blob/main/daml-lf/spec/contract-id.rst#discriminator-freshness), is not possible. 

The reason is that in projected transaction, there is actually no proper way to known what is input and what is output. 
From a abstract point of view, you cannot have capture-avoiding substitution for contract ID of local contract. 

To work around this problem:

* In fully distributed framework (Canton): 
  1. we will use suffix to distinguish local from global CID (local without suffix, global with suffix) 
  2. we will do the minimal change to ensures that the evaluation result is never changed by suffixing of local CID.

* In Central Committer framework (Sandbox, KV) 
    We will continue to use non-suffixed CIDs and rely on the fact the ledger ensure no two contracts can be created with the same ID. 

Implementation wise this translates into: 

- [x] Disable completely the contract ID freshness check (done during preprocessing and dynamically) #10620
- [x] Enforcing that LF Comparisons (both Ordering **and** Equality) fail at runtime if comparing local vs global CID. (enforcing point i above)  #10630
  * Ordering is already failing, but we need to expose the error (it is currently an internal error),
  * Equality must be changed to fail when comparing a local and global CID with same discriminator,
- [x] #10675
  * Checking during preprocessing, that all the CIDs in commands are suffixed #10642
  * Adding a flag in the Engine Config to (dis/en)able this check (Sandbox = off, KV = off, Canton = on) #10642
  * Conformance tests #10654
- [x] Fix the CID specification accordingly  #10703

This is a breaking change, however it should not impact any production system that do not have malicious participant.

PR #10630, opened 2021-08-19
LF: Comparisons fail at runtime if comparing local vs global CIDs

remyhaemmerle-da:
This PR enforces that LF Comparisons (both Ordering and Equality)
fails at runtime if comparing local vs global CID.

This corresponds to the second task of #10504.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/svalue/Ordering.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/svalue/Ordering.scala
@@ -106,27 +106,38 @@ object Ordering extends scala.math.Ordering[SValue] {
     diff
   }
   @inline
-  private[this] def compareCid(cid1: ContractId, cid2: ContractId): Int =
-    (cid1, cid2) match {
-      case (ContractId.V0(s1), ContractId.V0(s2)) =>
-        s1 compareTo s2
-      case (ContractId.V0(_), ContractId.V1(_, _)) =>
-        -1
-      case (ContractId.V1(_, _), ContractId.V0(_)) =>
-        +1
-      case (ContractId.V1(hash1, suffix1), ContractId.V1(hash2, suffix2)) =>
-        val c1 = crypto.Hash.ordering.compare(hash1, hash2)
-        if (c1 != 0)
-          c1
-        else if (suffix1.isEmpty == suffix2.isEmpty)
-          Bytes.ordering.compare(suffix1, suffix2)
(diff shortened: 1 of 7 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/svalue/Ordering.scala
```diff
@@ -106,27 +106,38 @@ object Ordering extends scala.math.Ordering[SValue] {
...
-          Bytes.ordering.compare(suffix1, suffix2)
-        else
-          // We crash here because we should have catch this beforehand
-          // when preprocessing or importing values.
-          throw SError.SErrorCrash(
-            getClass.getCanonicalName + ".compare",
-            "Unexcpected contract ID freshness Error",
-          )
+  private[this] def compareCid(cid1: ContractId, cid2: ContractId): Int = {
+    cid1 match {
```
cocreature, line 110: Personally I found the pattern match on the pair we had before easier to follow but that’s clearly subjective so your choice

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/svalue/Ordering.scala
remyhaemmerle-da: 45ad0f16f278fad24faddd34a8445fb1bba3e009

remyhaemmerle-da:
@moritzkiefer-da following the discussion we had this morning I revert the special case of create an exercise 45ad0f16f278fad24faddd34a8445fb1bba3e009

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -2751,6 +2764,13 @@ object EngineTest {
...
   }
 
+  private val dummySuffix = Bytes.assertFromString("00")
+
+  private def suffix(tx: Tx.Transaction) =
+    data.assertRight(tx.suffixCid(_ => dummySuffix))
+
+  // Mimics Canton reinterpreation
+  // requires a suffixed transaction.
   private def reinterpret(
```
remyhaemmerle-da, line 2774: This function mimic Canton reinterpretion, for this reason I change the test that use it, to require CID suffixing.

PR #10637, opened 2021-08-20
participant-integration-api: Fix a flaky test.

SamirTalwar-DA:
The `LedgerConfigurationSubscriptionFromIndexSpec` was flaky due to stubs not specifying all behavior for `IndexConfigManagementService`.

This fixes the underlying issue by avoiding stubs in favor of fakes, which means that _all_ behavior must be modelled.

Note: Martin Fowler has an excellent, terse description of [the various forms of test doubles][TestDouble].

[TestDouble]: https://www.martinfowler.com/bliki/TestDouble.html

### Pull Request Checklist

trigger the build.


nicu-da (approved):
Looks good. Not sure if this might be some `FakeServices` that we might reuse in other tests, it might be useful but it's always a problem of discovering these fakes when you need them.

SamirTalwar-DA:
I think we might want to reuse some of these but I'm also a little wary of doing so. It'd reduce duplication but also increase coupling between tests, which might result in strange failures in the future if improvements to Test A requires a change to a shared test-only class.

That said, I'm open to the idea if you can see a use case.

PR #10641, opened 2021-08-20
clean-up gsg README

garyverhaegen-da:
The current README dates back to when `create-daml-app` was its own
repo, and seems mostly written from the perspective of being the default
home page for the project. These days, `create-daml-app` serves two
purposes:

1. As the core starting point for the Getting Started Guide, we can
   assume people who are totally new to Daml will be following the
   getting started guide. How else would they even know about this
   template? In that case, I think it's fair to assume the user will *not*
   be looking to the README for guidance. Therefore, repeating the
   documentation here seems a bit pointless.
2. As a template to quickstart the creation of a "custom" app that is
   not the GSG. For that use-case, the README is the place where people
   would look for information, but at that point the information they're
   looking for are more along the lines of quick refreshers of what
   commands to run than explanations on how to use the app as Alice and
   Bob.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

Issue #6665, opened 2020-07-09
Test transaction validation.


remyhaemmerle-da:
Transaction validation should be extensively tested to avoid bug described in #6664

In particular the following point should be tests: 
* translation of transaction nodes to speedy command
* computation of global contract IDs during transaction validation (following the scheme described in  the [contract ID specification](https://github.com/digital-asset/daml/blob/master/daml-lf/spec/contract-id.rst) 
* comparison of submitted transaction with the replay transaction.

Consider also specifying validation. 

Also using this issue for refactoring work:
- Normalizing transactions and values constructed by the engine (making `isReplayedBy` simply structural equality).
- See google doc: https://docs.google.com/document/d/1Hhn8DZKwrSyIm3HuomtAUGaBgoTuBMqh8CZH7LvoHaI

remyhaemmerle-da:
@nickchapman-da @cocreature I think we can close this one.

PR #10648, opened 2021-08-23
Normalize transaction values within the engine

nickchapman-da:
Normalize transaction values in the engine as they are converted from speedy-values.
- Except in some limited cases! (scenario-runner, sandbox, some tests)
- `isReplayedBy` now finally becomes structural equality!
- Normalization controlled by a new `EngineConfig` flag (`transactionNormalization`)
- which is passed when a speedy `Machine` is constructed

This PR advances #6665

### DIFF
--- a/compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/speedy/ReplServiceMain.scala
+++ b/compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/speedy/ReplServiceMain.scala
@@ -278,7 +278,7 @@ class ReplService(
               }
             case RunScriptRequest.Format.JSON =>
               try {
-                LfValueCodec.apiValueToJsValue(v.toValue).compactPrint
+                LfValueCodec.apiValueToJsValue(v.toUnnormalizedValue).compactPrint
               } catch {
                 case SError.SErrorCrash(_, _) => throw new ReplServiceMain.NonSerializableValue()
 
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
@@ -310,6 +310,7 @@ class Engine(val config: EngineConfig = Engine.StableConfig) {
         readAs = readAs,
         validating = validating,
         contractKeyUniqueness = config.contractKeyUniqueness,
+        transactionNormalization = config.transactionNormalization,
       )
       interpretLoop(machine, ledgerTime)
     }
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
@@ -17,6 +17,7 @@ import com.daml.lf.transaction.ContractKeyUniquenessMode
   * @param allowedLanguageVersions The range of language versions the
   *     engine is allowed to load.  The engine will crash if it asked
   *     to load a language version that is not included in this range.
+  * @param transactionNormalization Normalizes transaction nodes according to the LF version
   * @param stackTraceMode The flag enables the runtime support for
   *     stack trace.
   * @param profileDir The optional specifies the directory where to
@@ -29,6 +30,7 @@ import com.daml.lf.transaction.ContractKeyUniquenessMode
   */
 final case class EngineConfig(
     allowedLanguageVersions: VersionRange[language.LanguageVersion],
+    transactionNormalization: Boolean = true,
     packageValidation: Boolean = true,
     stackTraceMode: Boolean = false,
     profileDir: Option[Path] = None,
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/LargeTransactionTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/LargeTransactionTest.scala
@@ -10,7 +10,7 @@ import com.daml.bazeltools.BazelRunfiles
 import com.daml.lf.archive.UniversalArchiveDecoder
 import com.daml.lf.data.Ref._
 import com.daml.lf.data.{FrontStack, ImmArray, Ref, Time}
-import com.daml.lf.language.Ast
+import com.daml.lf.language.{Ast, LanguageVersion}
 import com.daml.lf.scenario.ScenarioLedger
 import com.daml.lf.transaction.SubmittedTransaction
 import com.daml.lf.transaction.Transaction.Transaction
@@ -101,7 +101,8 @@ class LargeTransactionTest extends AnyWordSpec with Matchers with BazelRunfiles 
   private def report(name: String, quantity: Quantity[Double]): Unit =
     println(s"$name: $quantity")
 
-  private val engine: Engine = Engine.DevEngine()
+  private val engine: Engine =
+    new Engine(new EngineConfig(LanguageVersion.DevVersions, transactionNormalization = false))
 
   List(5000, 50000, 500000)
     .foreach { txSize =>
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
@@ -9,6 +9,7 @@ import java.util
 import com.daml.lf.data._
 import com.daml.lf.data.Ref._
 import com.daml.lf.language.Ast._
+import com.daml.lf.transaction.{TransactionVersion, Util}
 import com.daml.lf.value.Value.ValueArithmeticError
 import com.daml.lf.value.{Value => V}
 import com.daml.nameof.NameOf
@@ -26,7 +27,17 @@ sealed trait SValue {
 
   import SValue.{SValue => _, _}
 
-  def toValue: V[V.ContractId] = SValue.toValue(this)
+  /** Convert a speedy-value to a value which may not be correctly normalized.
+    * And so the resulting value should not be serialized.
+    */
+  def toUnnormalizedValue: V[V.ContractId] = SValue.toValue(this)
+
+  /** Convert a speedy-value to a value normalized according to the LF version.
+    */
+  def toNormalizedValue(version: TransactionVersion): V[V.ContractId] = {
+    val v = SValue.toValue(this)
+    Util.assertNormalizeValue(v, version) //TODO: avoid separate pass; inline norm into toValue
+  }
 
   def mapContractId(f: V.ContractId => V.ContractId): SValue =
     this match {
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
@@ -185,6 +185,7 @@ private[lf] object PartialTransaction {
       contractKeyUniqueness: ContractKeyUniquenessMode,
       submissionTime: Time.Timestamp,
       initialSeeds: InitialSeeding,
+      transactionNormalization: Boolean,
   ) = PartialTransaction(
     pkg2TxVersion,
     contractKeyUniqueness = contractKeyUniqueness,
@@ -199,6 +200,7 @@ private[lf] object PartialTransaction {
     globalKeyInputs = Map.empty,
     localContracts = Set.empty,
     actionNodeLocations = BackStack.empty,
+    transactionNormalization = transactionNormalization,
   )
 
   type NodeSeeds = ImmArray[(NodeId, crypto.Hash)]
@@ -281,6 +283,7 @@ private[lf] case class PartialTransaction(
     globalKeyInputs: Map[GlobalKey, PartialTransaction.KeyMapping],
     localContracts: Set[Value.ContractId],
     actionNodeLocations: BackStack[Option[Location]],
+    transactionNormalization: Boolean,
 ) {
 
   import PartialTransaction._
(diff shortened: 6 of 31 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
```diff
@@ -317,6 +317,7 @@ class Engine(val config: EngineConfig = new EngineConfig(LanguageVersion.StableV
         readAs = readAs,
         validating = validating,
         contractKeyUniqueness = config.contractKeyUniqueness,
+        valueNormalization = config.valueNormalization,
```
remyhaemmerle-da, line 320: Do we really need to make that parametric ?

remyhaemmerle-da (approved):
This makes the jobs. 
However I not very fan of the end result: 
-  The API of `PartialTransaction` now mixes `Value` and `SValue`
-  The version is calculated several time and in different place (In `SBuiltin` and in `PartialTransaction`). I would prefer is done in one or the other.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
```diff
@@ -396,7 +412,7 @@ private[lf] case class PartialTransaction(
   def insertCreate(
```
remyhaemmerle-da, line 412: The API is now mixing `Value` and `SValue`. I do not find that very neat. 
I would argue, either you do not change the type of `arg` or you change both `arg` and `key`.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
```diff
@@ -13,20 +13,20 @@ private final class Validation[Nid, Cid]() {
     *   the root nodes of [[recorded]].
     * @note This function is symmetric
     *
-    * 'isReplayedBy' normalizes its 2nd argument, and then determines structural equality.
+    * 'isReplayedBy' is simply structural equality.
```
cocreature, line 16: I’m confused, how can that be true? Where do you normalize for example `byKey`?

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/validation/ValidationSpec.scala
```diff
@@ -229,10 +226,10 @@ class ValidationSpec extends AnyFreeSpec with Matchers with TableDrivenPropertyC
   //--[predicates]--
   // Some tweaks have version dependant significance.
 
-  private def versionBeforeMinByKey(v: TransactionVersion): Boolean = {
+  /*  private def versionBeforeMinByKey(v: TransactionVersion): Boolean = {
```
cocreature, line 229: can we remove this?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
cocreature: as far as I can tell there are two cases where we turn this off:
1. `LargeTransactionTest`
2. `Sandbox`
I don’t see a reason why we should need it in either of them and in combination with the fact that `byKey` flags don’t seem to be normalized I’m a bit worried that this is just resulting in us never running tests that validate on LF < 1.14.

cocreature (changes requested):
Mostly looks good but I’d like to understand how we handle `byKey` and if we really need to make normalization optional before approving.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
cocreature: fair enough, especially if we end up killing `toValue` I’m fine with doing it outside of ptx.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
nickchapman-da: Yes. This has  been lost.
I mean to directly build the correct normalized value for byKey.

daml-lf/transaction/src/test/scala/com/digitalasset/daml/lf/validation/ValidationSpec.scala
nickchapman-da: yup

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
nickchapman-da: We also turn of for the scenario-runner.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: I agree it's a big ugly that some stuff is in ptx and other stuff outside. I will see if I can improve this.
The reason for being inside ptx is that this is where the `valueNormalization` flag is accessed.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
cocreature: the scenario runner seems like a reasonable usecase. I think ideally that wouldn’t affect speedy and we would instead call the valueenricher but that can definitely be in a separate PR. Also interested in whether @remyhaemmerle-da agrees that this is preferable.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/LargeTransactionTest.scala
```diff
@@ -101,7 +101,8 @@ class LargeTransactionTest extends AnyWordSpec with Matchers with BazelRunfiles
   private def report(name: String, quantity: Quantity[Double]): Unit =
     println(s"$name: $quantity")
 
-  private val engine: Engine = Engine.DevEngine()
+  private val engine: Engine =
+    new Engine(new EngineConfig(LanguageVersion.DevVersions, valueNormalization = false))
```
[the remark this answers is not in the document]
nickchapman-da: Because the test inspects record values embedded in a transaction, and it looks up fields by name
I expect it would be possible to change the test to avoid use of the field name, and so work with normalized values.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -28,6 +29,11 @@ sealed trait SValue {
 
   def toValue: V[V.ContractId] = SValue.toValue(this)
 
+  def toValueNorm(version: TransactionVersion): V[V.ContractId] = {
```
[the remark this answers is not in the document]
nickchapman-da: ok

I have wanted to not introduce a new function, and instead repurpose `toValue` (taking a `version`) to enforce normalization always. The problem is that in various places we don't have a `version` to hand. For values which are not destined to end up in a transaction.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
```diff
@@ -17,6 +17,7 @@ import com.daml.lf.transaction.ContractKeyUniquenessMode
   * @param allowedLanguageVersions The range of language versions the
   *     engine is allowed to load.  The engine will crash if it asked
   *     to load a language version that is not included in this range.
+  * @param valueNormalization Strips type-annotations from values.
```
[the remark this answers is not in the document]
nickchapman-da: yup, that makes sense

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: But since I now defined `ptx.normValue`  (didn't have this originally)
```
def normValue(templateId: TypeConName, svalue: SValue): Value[Value.ContractId]
```
then I should be able to call this from outside.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/LargeTransactionTest.scala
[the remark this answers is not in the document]
nickchapman-da: But I think I prefer to leave the test working with type-annotated values.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
nickchapman-da: AHA. Explains this failure on CI (only Linux) ?
```
//ledger/ledger-api-test-tool:conformance-test-1.8
```

cocreature: yep seems very likely

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Engine.scala
remyhaemmerle-da: Scenario runner is indeed a reasonable use case. 
I incline toward using the value enricher:
- this would limit the number of parameter in the production code
- this would make the scenario service working similarly as the lerger API in verbose mode.

nickchapman-da: I'll take a stab at using the value enricher for the scenario runner in a subsequent PR.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: @remyhaemmerle-da 
> The API is now mixing Value and SValue. I do not find that very neat.
I hope things are now more to your satisfaction :)

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Validation.scala
nickchapman-da: I fixed the `byKey` normalization. This test passes again.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/transaction/PartialTransaction.scala
nickchapman-da: I rediscovered the other reason why some of the normalization (speedy-value --> value) was done inside ptx...

In the case of the call to `endExercises` from `Speedy.scala`, the `templateId` was not readily available at the call site, but it was available inside ptx in the `ExercisesContextInfo`.

But I fixed it now to thread the `templateId` from `Compiler.scala` through `ScopeExercises` and into `KCloseExercise` where it is needed.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
```diff
@@ -17,6 +17,7 @@ import com.daml.lf.transaction.ContractKeyUniquenessMode
   * @param allowedLanguageVersions The range of language versions the
   *     engine is allowed to load.  The engine will crash if it asked
   *     to load a language version that is not included in this range.
+  * @param transactionNormalization Normalizes values according to the LF version
```
cocreature, line 20:
```suggestion
  * @param transactionNormalization Normalizes transaction nodes according to the LF version
```
byKey is not part of the value

compiler/repl-service/server/src/main/scala/com/digitalasset/daml/lf/speedy/ReplServiceMain.scala
```diff
@@ -278,7 +278,7 @@ class ReplService(
               }
             case RunScriptRequest.Format.JSON =>
               try {
-                LfValueCodec.apiValueToJsValue(v.toValue).compactPrint
+                LfValueCodec.apiValueToJsValue(v.toUnNormalizedValue).compactPrint
```
cocreature, line 281:
```suggestion
                LfValueCodec.apiValueToJsValue(v.toUnnormalizedValue).compactPrint
```
nit: just looks weird to me otherwise

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -26,7 +27,17 @@ sealed trait SValue {
 
   import SValue.{SValue => _, _}
 
-  def toValue: V[V.ContractId] = SValue.toValue(this)
+  /** Convert a speedy-value to a value which may not be correctly normalized.
+    * And so the resulting value should not be embedded into a transaction.
```
cocreature, line 31:
```suggestion
    * And so the resulting value should not be serialized.
```

cocreature (approved):
Nice, thanks!
Can you please post the list of sandbox test cases that fail if we enable normalization there before merging?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
nickchapman-da: done

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
nickchapman-da: done

nickchapman-da:
> Nice, thanks!
> Can you please post the list of sandbox test cases that fail if we enable normalization there before merging?

will do

remyhaemmerle-da (commented):
I like it more like that. Thanks a lot.

nickchapman-da:
If we remove `transactionNormalization = false,`
from `ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxServer.scala`
then the following tests fail:

```
//compiler/damlc/tests:repl
//language-support/ts/codegen/tests:build-and-lint-test
//ledger-service/http-json:integration-tests-ce
//ledger-service/http-json:integration-tests-ee
//ledger/sandbox-classic:conformance-test-exceptions-memory-1.14
//ledger/sandbox-classic:conformance-test-static-time-memory-1.14
//ledger/sandbox-classic:conformance-test-wall-clock-memory-1.14
```

cocreature:
That’s not good, it’s breaking `verbose` mode on the ledger API not just internal tests. I’m fine with disabling this in this PR but this really should go through the value enricher.

cocreature:
Looking at the code, the enricher is missing from https://github.com/digital-asset/daml/blob/ef239fdd1a8e5ddcaa5154ce4826422a76820dca/ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala#L115. Do you want to take a look at fixing this after this PR @nickchapman-da?

nickchapman-da:
> Looking at the code, the enricher is missing from
> 
> https://github.com/digital-asset/daml/blob/ef239fdd1a8e5ddcaa5154ce4826422a76820dca/ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala#L115
> 
> . Do you want to take a look at fixing this after this PR @nickchapman-da?

Sure

PR #10649, opened 2021-08-23
Dpp-558 Fix startexclusive queries on oracle

rautenrieth-da:
# Issue

Oracle treats empty byte arrays as null. We handled this in some places, but not everywhere.

# Why wasn't this caught

JdbcLedgerDao tests are not isolated. They look up the ledger end at the start of the test, and query from there. The tests never query from ledger begin.

# Fix

The following columns store offsets and are potentially affected. I have searched for all comparisons on these columns and fixed two places where the comparison was not guarded against null.

- parameters.ledger_end
- parameters.participant_pruned_up_to_inclusive
- configuration_entries.ledger_offset
- packages.ledger_offset
- package_entries.ledger_offset
- parties.ledger_offset
- party_entries.ledger_offset
- participant_command_completions.completion_offset
- participant_events_*.event_offset

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
@@ -312,7 +312,7 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
 
   private val SQL_GET_PARTY_ENTRIES = SQL(
     """select * from party_entries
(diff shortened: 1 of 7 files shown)


oliverse-da (approved):
Looks good.

Just curious, is it necessary in the unit tests to write a configuration update as the first update?

rautenrieth-da:
> Just curious, is it necessary in the unit tests to write a configuration update as the first update?

I often try to write "realistic" sequences of updates, and you can't have a transaction update without a preceding configuration update. But it's not required for this test, we could remove it.

nmarton-da (approved):

PR #10650, opened 2021-08-23
participant-integration-api: Accommodate changes to max dedup time.

SamirTalwar-DA:
Previously, if the max deduplication time was extended, the participant _might_ retain the old time for certain submissions.

@fabiotudone-da explains it better than I could:

- Previously, the max deduplication time was read from the ledger configuration only upon a tracker's creation.
- This meant that alive trackers would not take into account updates to it (via ledger configuration updates) during their lifetime.
- Now, the max deduplication time is checked every time it is needed.
- This also means that the failure when the ledger configuration is missing happens later.

### Changelog

- **[Ledger API Server]** The API server manages a single command tracker per (application ID × submitters) pair. This tracker would read the current ledger configuration's maximum deduplication time on creation, but never updated it, leading to trackers that might inadvertently reject a submission when it should have been accepted. The tracker now reads the latest ledger configuration.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -14,19 +14,19 @@ import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
 import com.daml.ledger.client.services.commands.tracker.CompletionResponse.{
   CompletionFailure,
   CompletionSuccess,
+  NotOkResponse,
 }
 import com.daml.ledger.client.services.commands.{CompletionStreamElement, tracker}
+import com.daml.platform.server.api.validation.ErrorFactories
 import com.daml.util.Ctx
-import com.google.protobuf.duration.{Duration => ProtoDuration}
 import com.google.protobuf.empty.Empty
-import com.google.rpc.code._
-import com.google.rpc.status.Status
-import io.grpc.{Status => RpcStatus}
+import com.google.rpc.status.{Status => StatusProto}
+import io.grpc.Status
 import org.slf4j.LoggerFactory
 
 import scala.collection.compat._
 import scala.collection.{immutable, mutable}
-import scala.concurrent.duration._
+import scala.concurrent.duration.DurationInt
 import scala.concurrent.{Future, Promise}
 import scala.util.control.NoStackTrace
 import scala.util.{Failure, Success, Try}
@@ -49,7 +49,7 @@ import scala.util.{Failure, Success, Try}
   * We also have an output for offsets, so the most recent offsets can be reused for recovery.
   */
 // TODO(mthvedt): This should have unit tests.
-private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDuration)
+private[commands] class CommandTracker[Context](maxDeduplicationTime: () => Option[JDuration])
     extends GraphStageWithMaterializedValue[CommandTrackerShape[Context], Future[
       immutable.Map[String, Context]
     ]] {
@@ -208,13 +208,13 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
         }
       }
 
-      private def registerSubmission(submitRequest: Ctx[Context, SubmitRequest]): Unit = {
-        submitRequest.value.commands
-          .fold(
+      private def registerSubmission(submitRequest: Ctx[Context, SubmitRequest]): Unit =
+        submitRequest.value.commands match {
+          case None =>
             throw new IllegalArgumentException(
               "Commands field is missing from received SubmitRequest in CommandTracker"
             ) with NoStackTrace
-          ) { commands =>
+          case Some(commands) =>
             val commandId = commands.commandId
             logger.trace("Begin tracking of command {}", commandId)
             if (pendingCommands.contains(commandId)) {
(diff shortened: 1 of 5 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -223,23 +223,28 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
...
-              Instant.now().plusSeconds(dedup.seconds).plusNanos(dedup.nanos.toLong)
+            maxDeduplicationTime() match {
+              case None =>
+                emit(
+                  resultOut,
+                  submitRequest.map { _ =>
+                    val status = GrpcStatus.toProto(ErrorFactories.missingLedgerConfig().getStatus)
+                    Left(NotOkResponse(commandId, status))
+                  },
+                )
```
nicu-da, line 234: Why would we delay the check until here? Wouldn't it make more sense to fail fast during startup as we know if the config is missing or not.

SamirTalwar-DA: Well, the config can be missing for the first 5 seconds, and then show up.

`LedgerConfigSubscription` is mutable; the state can change.

fabiotudone-da (commented):
If I understand correctly:

- Previously, the dedup time was read from the ledger configuration only upon a tracker's creation.
- This led to alive trackers not to take into account updates to it (via ledger configuration updates) during their lifetime.
- Now, the dedup time is checked every time it is needed.
- This also means that the failure when the ledger configuration is missing happens later.

If so, can you expand the PR's description? Also, isn't this an API behavior change that needs a changelog and a heads-up?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: Can't we validate it though before we get to this point? We're creating the tracker in an invalid state then from the beggining and this config validation seem to leak way too deep into the tracker.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -223,23 +223,28 @@ private[commands] class CommandTracker[Context](maxDeduplicationTime: () => JDur
                 s"A command with id $commandId is already being tracked. CommandIds submitted to the CommandTracker must be unique."
               ) with NoStackTrace
             }
-            val commandTimeout = {
-              lazy val maxDedup = maxDeduplicationTime()
-              val dedup = commands.deduplicationTime.getOrElse(
-                ProtoDuration.of(maxDedup.getSeconds, maxDedup.getNano)
-              )
-              Instant.now().plusSeconds(dedup.seconds).plusNanos(dedup.nanos.toLong)
+            maxDeduplicationTime() match {
```
nicu-da, line 226: can we rename this to `commandTimeout`?

SamirTalwar-DA:
Yes, you are correct. Thank you for explaining it better than I could.

As for the changelog, it's the kind of API behavior that should never affect a properly functioning ledger, because the configuration will show up within a couple of seconds at most. I'm not sure it's helpful to include changelog entries for edge cases. Do you think it would help either a participant operator or client diagnose a failure?

nicu-da:
> If I understand correctly:
> 
>     * Previously, the dedup time was read from the ledger configuration only upon a tracker's creation.
> 
>     * This led alive trackers not to take into account updates to it (via ledger configuration updates) during their lifetime.
> 
>     * Now, the dedup time is checked every time it is needed.
> 
>     * This also means that the failure when the ledger configuration is missing happens later.
> 
> 
> If so, can you expand the PR's description? Also, isn't this an API behavior change that needs a changelog and a heads-up?

Is this the case?
This dedup time was previosly part of the request, using the configure default if no dedup was set in the request. 
In the tracker we are passing a function which afaik it's based on the latest config so not sure where it would be stale. 
Also this is the command timeout, we should not reference dedup anymore.

SamirTalwar-DA:
> This dedup time was previosly part of the request, using the configure default if no dedup was set in the request.

Correct.

> In the tracker we are passing a function which afaik it's based on the latest config so not sure where it would be stale.

It's based on the latest config at the time it was captured, which was during tracker creation.

> Also this is the command timeout, we should not reference dedup anymore.

That work is upcoming; haven't implemented it yet.

fabiotudone-da:
> I'm not sure it's helpful to include changelog entries for edge cases. Do you think it would help either a participant operator or client diagnose a failure?

Potentially, if the changelog conveying process is structured in such a way to allow it (and to make it easy). I see no harm in including it and a potential harm in not including it.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
SamirTalwar-DA: I suggest we rename this as part of upcoming work to separate command deadlines from the deduplication time.

SamirTalwar-DA:
Changelog added. :man_dancing:

nicu-da (approved):

fabiotudone-da (approved):

fabiotudone-da:
@digital-asset/kv-participant Any remarks?

PR #10651, opened 2021-08-24
participant-integration-api: Simplify convertOffset.

SamirTalwar-DA:
This function used to be memoized but I accidentally removed the behavior a year ago. As a reasonable client will never request the transactions from `LEDGER_END` to `LEDGER_END`, the behavior did not change.

This simplifies the function further.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
@@ -116,13 +116,19 @@ private[platform] final class LedgerBackedIndexService(
         .map(_._2)
     }
 
-  private def convertOffset(offset: LedgerOffset)(implicit
-      loggingContext: LoggingContext
-  ): Try[Offset] = offset match {
-    case LedgerOffset.LedgerBegin => Success(Offset.beforeBegin)
-    case LedgerOffset.LedgerEnd => Success(ledger.ledgerEnd())
-    case LedgerOffset.Absolute(offset) => ApiOffset.fromString(offset)
-  }
+  private def convertOffset(
+      offset: LedgerOffset
+  )(implicit loggingContext: LoggingContext): Try[Offset] =
+    offset match {
+      case LedgerOffset.LedgerBegin => Success(Offset.beforeBegin)
+      case LedgerOffset.LedgerEnd => Success(ledger.ledgerEnd())
+      case LedgerOffset.Absolute(offset) => ApiOffset.fromString(offset)
+    }
+
+  private def convertOptionalOffset(
+      offset: Option[LedgerOffset]
+  )(implicit loggingContext: LoggingContext): Try[Option[Offset]] =
+    offset.map(convertOffset(_).map(Some(_))).getOrElse(Success(None))
 
   private def between[A](
       startExclusive: domain.LedgerOffset,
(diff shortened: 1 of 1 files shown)


hubert-da (approved):

ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
```diff
@@ -114,33 +114,26 @@ private[platform] final class LedgerBackedIndexService(
           verbose = verbose,
         )
         .map(_._2)
-    })
+    }
 
-  // Returns a function that memoizes the current end
-  // Can be used directly or shared throughout a request processing
-  private def convertOffset(implicit
+  private def convertOffset(offset: LedgerOffset)(implicit
```
fabiotudone-da, line 119:
```suggestion
  private def convertOffset(offset: domain.LedgerOffset)(implicit
```

ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
```diff
@@ -114,33 +114,26 @@ private[platform] final class LedgerBackedIndexService(
...
-      endInclusive
-        .map(convert(_).map(Some(_)))
-        .getOrElse(Source.single(None))
-        .flatMapConcat {
-          case Some(`begin`) =>
-            Source.empty
+  ): Source[A, NotUsed] =
+    convertOffset(startExclusive)
+      .flatMap { begin =>
+        endInclusive.map(convertOffset(_).map(Some(_))).getOrElse(Success(None)).map {
```
fabiotudone-da, line 135: I think binding `convertOffset(_).map(Some(_))` could make it more readable.

fabiotudone-da (commented):
Is there test coverage for this?

SamirTalwar-DA:
There is no direct test coverage for `LedgerBackedIndexService`. Unfortunately we rely far too much on conformance tests and other integrated tests here.

I'm not sure how to begin to test it with the current architecture; it would be almost meaningless.

ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
SamirTalwar-DA: Why do you want to prefix it?

fabiotudone-da: It seems to be prefixed elsewhere and imo it clarifies what is being converted.

fabiotudone-da:
> There is no direct test coverage for `LedgerBackedIndexService`. Unfortunately we rely far too much on conformance tests and other integrated tests here.
> 
> I'm not sure how to begin to test it with the current architecture; it would be almost meaningless.

I understand, on the other hand refactorings without checks concern me. I start wondering if we shouldn't consider feature branches in some cases.

ledger/participant-integration-api/src/main/scala/platform/index/LedgerBackedIndexService.scala
SamirTalwar-DA: I think this is to differentiate it from the `LedgerOffset` protocol buffers message. IMO this is the wrong way around; we should prefix or alias the protobuf, not the domain type.

I'll add the prefix but I'll revisit this at some point to standardize it.

SamirTalwar-DA:
To clarify: yes, there are tests, but they exist in `TransactionStreamIT` and other conformance tests.

fabiotudone-da:
> To clarify: yes, there are tests, but they exist in `TransactionStreamIT` and other conformance tests.

Sure, I am talking about unit tests.

SamirTalwar-DA:
Eh, you're right, we need to get this stuff under test first.

I'll close the PR and we can worry about this another time, after some architectural changes to make it testable.

PR #10652, opened 2021-08-24
Command dedup: add columns to completions in append-only schemas [KVL-1057]

fabiotudone-da:
## Description

This is a further change in a series (after #10619) to let the Ledger API provide more command de-duplication information in completions as well (only the submission rank is deferred to later stages), so that the application can use it in correlation with the one provided at command submission time.

It updates the index DB append-only schema to allow storing the additional command de-duplication information and provides the read/write logic for it.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


fabiotudone-da:
@digital-asset/kv-participant Could you please have a first stab at the schema changes?

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
```diff
@@ -0,0 +1,19 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+-- Add columns to store command deduplication information in completions.
+
+ALTER TABLE participant_command_completions
+    -- The application ID has to be provided by the application.
+    ADD COLUMN application_id text NOT NULL,
+    -- The submission ID will be provided by the participant or driver if the application didn't provide one.
+    ADD COLUMN submission_id text NOT NULL,
```
fabiotudone-da, line 10: I guess these two will need to be nullable to accomodate historical data.

SamirTalwar-DA: Yes, absolutely. And I recommend making them nullable on all databases so they're not out of sync.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -122,6 +122,19 @@ CREATE TABLE participant_command_completions (
     command_id VARCHAR NOT NULL,
     -- The transaction ID is `NULL` for rejected transactions.
     transaction_id VARCHAR,
+    -- The application ID has to be provided by the application.
+    application_id VARCHAR NOT NULL,
```
SamirTalwar-DA, line 126: There is already an `application_id` above.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
```diff
@@ -0,0 +1,19 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+-- Add columns to store command deduplication information in completions.
+
+ALTER TABLE participant_command_completions
+    -- The application ID has to be provided by the application.
+    ADD COLUMN application_id text NOT NULL,
```
SamirTalwar-DA, line 8: The H2 and Oracle append-only index databases already have an `application_id`. Are you sure that the PostgreSQL index DB doesn't have this column already?

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
nmarton-da: Will this work? How do we populate for old data?

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
fabiotudone-da: 8079673

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
fabiotudone-da: 8079673

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
fabiotudone-da: Made both nullable in all the schemas: 3e5fa10

fabiotudone-da (commented):
A couple of additional things I am not sure about:

- Do we have a policy for the position of newly added columns?
- I couldn't find another instance of duration, is the double i64+i32 column (as in protobuf) OK?

nmarton-da:
> A couple of additional things I am not sure about:
> 
>     * Do we have a policy for the position of newly added columns?
AFAIK no policy regarding position ATM
> 
>     * I couldn't find another instance of the duration, is the double i64+i32 column (as in protobuf) OK?
Do you need such precision? Wouldn't be seconds or millis enough (resulting in one bigint column)?
Besides no idea what datatype best to store duration, I would go with millis

nmarton-da:
General comment to this PR
Not sure about you intentions, but it would be great (most probably necessary) to make the DB changes together with the DB facing code changes:
* Reviewers would gain more clarity where the schema is heading semantically
* It would be clear how backwards compatibility is ensured
* It would probably also include db facing tests, which would ensure smooth operation between schema and code
* It would successfully prevent possible multiple iterations on the schema due to integration with code (indexes, data-type changes, constraint changes, migration changes etc)

fabiotudone-da:
> General comment to this PR

@nmarton-da Thanks for the feedback; the intention was to break down the work in small, separate reviews as much as possible, starting with the ones requiring the most feedback (like the ledger API proto and the DB schema, the latter purely on a technical level initially).
I think creating additional PRs on top of this one with the code changes (and waiting for a review on them too before merging anything) could have the advantages of both approaches. WDYT?

fabiotudone-da:
> Do you need such precision? Wouldn't be seconds or millis enough (resulting in one bigint column)?
Besides no idea what datatype best to store duration, I would go with millis/bigint

nmarton-da:
> > General comment to this PR
> 
> @nmarton-da Thanks for the feedback; the intention was to break down the work in small, separate reviews as much as possible, starting with the ones requiring the most feedback (like the ledger API proto and the DB schema, the latter purely on a technical level initially).
> I think creating additional PRs on top of this one with the code changes (and waiting for a review on them too before merging anything) could have the advantages of both approaches. WDYT?

I think the schema change and the code which works with that should be same unit in terms of work / mergable / PR.
How about keeping this a draft, after agreement reached, closing it, and then open another one which is about the change of the persistence with this content, and with the code changes?

Don't get me wrong please, small PRs is a very good thing, but there must be a PR which is too small :) I claim making db changes without the supporting code falls into the too small category. Also don't get me wrong please on the other end neither: I am not saying that because that should belong together with something else, then everything needs to belong together: feel free to maintain granular changes later as you see fit!

TLDR:
I would applaud if we could have a selfcontained PR:
* having db schema changes
* having StorageBackend code changes
* having JdbcLedgerDao code changes if needed
* and db facing tests proving successful SQL-scala integration
...together in one PR

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
```diff
@@ -251,47 +253,81 @@ class UpdateToDbDtoSpec extends AnyWordSpec with Matchers {
...
       val status = StatusProto.of(Status.Code.ABORTED.value(), "test reason", Seq.empty)
-      val update = state.Update.CommandRejected(
-        someRecordTime,
-        completionInfo,
-        new state.Update.CommandRejected.FinalReason(status),
-      )
-      val dtos = UpdateToDbDto(someParticipantId, valueSerialization, compressionStrategy)(
-        someOffset
-      )(update).toList
+      forAll(deduplicationPeriods) {
```
fabiotudone-da, line 280: I feel like I am hijacking a couple of test cases here but I need to check all the current dedup cases on a rejection and a successful completion (I don't really care which transaction). Let me know if you can think of a better way.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
```diff
@@ -233,6 +237,20 @@ private[backend] object StorageBackendTestValues {
...
+        case _ => None
+      },
+      deduplication_time_seconds = deduplication.flatMap {
+        case Middle3((seconds, _)) => Some(seconds)
+        case _ => None
+      },
+      deduplication_time_nanos = deduplication.flatMap {
+        case Middle3((_, nanos)) => Some(nanos)
+        case _ => None
+      },
```
fabiotudone-da, line 252: Could extract these together and bind-deconstruct them earlier, maybe slightly better.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala
```diff
@@ -76,4 +58,77 @@ trait CompletionStorageBackendTemplate extends CompletionStorageBackend {
...
+  ).?
+  private val deduplicationStartColumn: RowParser[Instant] = instant("deduplication_start")
+
+  private val acceptedCommandParser: RowParser[CompletionStreamResponse] =
+    acceptedCommandSharedColumns ~
+      deduplicationOffsetColumn ~
+      deduplicationTimeSecondsColumn ~ deduplicationTimeNanosColumn ~
+      deduplicationStartColumn map {
+        case offset ~ recordTime ~ commandId ~ _ ~ _ ~ transactionId ~
+            _ ~ _ ~ _ ~ _ =>
```
fabiotudone-da, line 88: These are the newly added (+ pre-existing application ID) command dedup columns, they will be used in the next PR.
```

File: ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CompletionStorageBackendTemplate.scala [hunk 7]
```diff
@@ -76,4 +58,77 @@ trait CompletionStorageBackendTemplate extends CompletionStorageBackend {
...
+  private val rejectedCommandParser: RowParser[CompletionStreamResponse] =
+    sharedColumns ~
+      deduplicationOffsetColumn ~
+      deduplicationTimeSecondsColumn ~ deduplicationTimeNanosColumn ~
+      deduplicationStartColumn ~
+      rejectionStatusCodeColumn ~
+      rejectionStatusMessageColumn ~
+      rejectionStatusDetailsColumn map {
+        case offset ~ recordTime ~ commandId ~ _ ~ _ ~
+            _ ~ _ ~ _ ~ _ ~
```
fabiotudone-da, line 106: These are the newly added (+ pre-existing application ID) command dedup columns, they will be used in the next PR.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -117,11 +117,25 @@ CREATE TABLE participant_command_submissions (
...
+    submission_id VARCHAR,
+    -- The three alternatives below are mutually exclusive, i.e. the deduplication
+    -- interval could have specified by the application as one of:
+    -- 1. an initial offset
+    -- 2. a duration (split into two columns, seconds and nanos, mapping protobuf's 1:1)
+    -- 3. an initial timestamp
+    deduplication_offset VARCHAR,
+    deduplication_time_seconds BIGINT,
+    deduplication_time_nanos INT,
+    deduplication_start TIMESTAMP,
```
fabiotudone-da, line 138: This way of specifying the deduplication period is currently unsupported but it belongs here for imminent use, unsure if removing it or leaving it for posterity.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
SamirTalwar-DA: You could maybe do something like:

```scala
      deduplication_offset = deduplication.leftOr(None)(Some),
      deduplication_time_seconds = deduplication.middleOr(None)(Some),
      deduplication_time_nanos = deduplication.rightOr(None)(Some),
```

That said, I think I'd prefer a custom set of classes, with appropriate getters.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
SamirTalwar-DA: I agree with leaving it and keeping the migration count down.

SamirTalwar-DA (commented):
Looks pretty good to me, but I'll leave approval to the index maintainers.

SamirTalwar-DA:
> Good point, I couldn't find any mention of resolution restrictions w.r.t. the API type (i.e., protobuf's). @SamirTalwar-DA Do you know anything more here?

I highly suggest not throwing away information at this point. If you want to store the value in one column, with a different precision, I'd recommend doing that before it makes it to the ledger.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
fabiotudone-da: It would become even more cryptic:

```
      deduplication_offset = deduplication.flatMap(_.leftOr(None[_])(Some(_))),
      deduplication_time_seconds = deduplication.flatMap(_.middleOr(None[_])(Some(_))).map(_._1),
      deduplication_time_nanos = deduplication.flatMap(_.middleOr(None[_])(Some(_))).map(_._2),
```

I think I'll go with an ADT.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
```diff
@@ -205,25 +205,61 @@ private[backend] object StorageBackendTestValues {
     )
   }
 
+  sealed trait Deduplication extends Product with Serializable {
```
fabiotudone-da, line 208:
```suggestion
  sealed trait Deduplication extends Product {
```

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -106,11 +106,25 @@ CREATE TABLE participant_command_submissions (
 CREATE TABLE participant_command_completions (
     completion_offset VARCHAR NOT NULL,
     record_time TIMESTAMP NOT NULL,
-    application_id VARCHAR NOT NULL,
+    -- The application ID has to be provided by the application.
```
nmarton-da, line 109: discussions on the precision made me think more about this:
* Foreword: sorry to raise this so late, but this is exactly what is my main concern regarding too granular PRs: not knowing the whole picture makes it quite hard to reason about this
* It seems to me that all the old fields are pass-through info for the indexer (except from the offset). How about the new ones? Does the indexer need to be aware of all this info? Can't we just squeeze all the non-relevant fields into a proto which could be possibly passed through the system? (this pattern can be captured in all the contract related serialized data and also in here since @SamirTalwar-DA 's commit, the rejection_status_details)

nmarton-da (commented):
* The test for the low level database layer is located in StorageBackendSuite
* If you would want to test with that test suite, you would require to change the CompletionStreamResponse, as I suggested, otherwise you won't be able to observe outcome. (whether CompletionStreamResponse is both a persistent and API data-type is good or not, it debatable. BTW this is also a sign which made me think about making all this info a pass-through for the indexer/participant-api)
* So if you would have wanted to organize you're changes in PRs, and keep the persistence code tested the same time you would need to follow a different grouping: first doing db-schema changes, Dto, Schema, CompletionStreamResponse, and test then can be added to SotrageBackendSuite, and only then doing the UpdateToDbDto changes, with unit tests (which actually would unlock JdbcLedgerDao related testing, and integration testing as well)

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
fabiotudone-da: My general idea is that everything that needs to be serialized (e.g. b/c it's driver-dependent) should be, but nothing more because that makes using the info more complicated.

In this case, it's the new command dedup model and it's API behavior, so I think we shouldn't make it opaque at the index level and it's supposed not to change in the (near) future; for this reason, maintenance should also be minimal after this initial work.

Being core info, it's also possible that the indexer may benefit from being aware of these new fields in the future (although the design trend for command deduplication implementations is currently driver-based). In order to make it easy to use (and to migrate) them in future designs/architectures, I think it's better if they are known to the DB.
Another use case is debugging/troubleshooting; it is true that there are other means (e.g. logging) but being able to look at the actual data in the DB at a given point in time can be pretty useful in some circumstances.

I am aware that even just adding pass-through fields is a lot of work and code but maybe that's the actual issue and we should rather deal with that.

Does it make sense?

nmarton-da: After discussion with @meiersi-da and @rautenrieth-da the concolusion is that it is okay in this case to keep this pass-thru data flattened :ok_hand:

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
```diff
@@ -205,25 +205,61 @@ private[backend] object StorageBackendTestValues {
...
+    final case class Span(span: SecondsAndNanos) extends Deduplication {
+      require(
+        span.allComponentsAreNonNegative,
+        s"All the deduplication window components must not be negative: $span",
+      )
+    }
+    final case class Start(instant: Instant) extends Deduplication
+  }
+
   def dtoCompletion(
```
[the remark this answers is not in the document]
fabiotudone-da: > wouldn't it will be simpler just to extend dtoCompletion with the default null-s, and override the needed ones in the test

Yes and necessary to test for unexpected data. Done so.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
fabiotudone-da: Actually, application ID was not nullable in the PSQL schema, I was thrown off by a comment there. Reverted to non-nullable in all schemas.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
```diff
@@ -41,6 +41,7 @@ private[backend] object StorageBackendTestValues {
     ParameterStorageBackend.IdentityParams(someLedgerId, someParticipantId)
   val someParty: Ref.Party = Ref.Party.assertFromString("party")
   val someApplicationId: Ref.ApplicationId = Ref.ApplicationId.assertFromString("application_id")
+  val someSubmissionId: String = "submission_id"
```
rautenrieth-da, line 44: Can you make this a `Ref.SubmissionId`? All other `some*` values are properly typed as well.

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
rautenrieth-da: Maybe you could have left `"handle TransactionAccepted (single create node)"` alone and added a new test `"handle TransactionAccepted (all deduplication data)"`. This would make it more clear what the purpose of each test is, at the cost of duplicated code.

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -60,4 +81,64 @@ private[platform] object CompletionFromTransaction {
...
+        )
+    }
+  }
+
+  private def toApiDeduplicationPeriod(
+      maybeDeduplicationOffset: Option[String],
+      maybeDeduplicationTimeNanos: Option[Int],
+      maybeDeduplicationTimeSeconds: Option[Long],
+  ): Completion.DeduplicationPeriod =
+    // The only invariant tha should hold, considering legacy data, is that either
```
rautenrieth-da, line 126:
```suggestion
    // The only invariant that should hold, considering legacy data, is that either
```

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestValues.scala
fabiotudone-da: 87da1dd

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
fabiotudone-da: f32346e

ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDbDto.scala
```diff
@@ -257,22 +253,54 @@ object UpdateToDbDto {
...
 
+  private def commandCompletion(
+      offset: Offset,
+      recordTime: Time.Timestamp,
+      transactionId: Option[Ref.TransactionId],
+      completionInfo: CompletionInfo,
+  ): DbDto.CommandCompletion = {
+    val (deduplicationTimeSeconds, deduplicationTimeNanos) =
+      completionInfo.optDeduplicationPeriod
+        .flatMap {
```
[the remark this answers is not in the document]
fabiotudone-da: 482891e and  804c99e

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V108_add_deduplication_info_to_completions.sql
fabiotudone-da: (More correctly, it is still nullable because it also contained checkpoints but they were migrated out at some point)

ledger/participant-integration-api/src/test/suite/scala/platform/store/backend/UpdateToDbDtoSpec.scala
fabiotudone-da: Yes, I think that test purpose clarity is more important so I will do as you suggest before merging.

fabiotudone-da: bde3d1e, there's a comprehensive code deduplication work that can be done but I'd rather do it in a separate future PR.

nmarton-da (approved):
Looks solid.
Thanks for your patience regarding the back and forth-s

ledger/participant-integration-api/src/main/scala/platform/store/backend/UpdateToDbDto.scala
```diff
@@ -257,22 +253,48 @@ object UpdateToDbDto {
...
+  ): DbDto.CommandCompletion = {
+    val (deduplicationOffset, deduplicationTimeSeconds, deduplicationTimeNanos) =
+      completionInfo.optDeduplicationPeriod
+        .map {
+          case DeduplicationOffset(offset) =>
+            (Some(offset.toHexString), None, None)
+          case DeduplicationDuration(duration) =>
+            (None, Some(duration.getSeconds), Some(duration.getNano))
+        } match {
+        case Some(value) => value
```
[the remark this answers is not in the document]
fabiotudone-da: 4a94562

ledger/participant-integration-api/src/main/scala/platform/store/CompletionFromTransaction.scala
```diff
@@ -60,4 +81,82 @@ private[platform] object CompletionFromTransaction {
       recordTime = Some(fromInstant(recordTime)),
       offset = Some(LedgerOffset.of(LedgerOffset.Value.Absolute(offset.toApiString))),
     )
+
+  private def toApiCompletion(
```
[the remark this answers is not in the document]
fabiotudone-da: It's mostly because of scalapb plus proto's "non-optional optionals"; I simplified it further in 2749c75 and I think a unit test at this level doesn't add much value because it's very thin logic and the test would basically duplicate it in order to check it. I think testing the `CompletionFromTransaction` unit at its interface (i.e., `acceptedCompletion` and `rejectedCompletion`) is in a similar situation as it just delegates to the private `toXXX` utilities.

Also, this logic should  already have full coverage by the DB-level integration test cases in `StorageBackendTestsCompletions`. WDYT?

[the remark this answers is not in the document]
fabiotudone-da: I'll merge this as soon as the build passes and we'll deal with further improvements in a separate PR, if need be.

PR #10662, opened 2021-08-25
participant-integration-api: Ensure that all waiting, failed, and closed trackers are cleaned up.

SamirTalwar-DA:
We didn't handle these edge cases as well as we probably should have. Not a big deal, but potentially something that could grow.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
@@ -97,7 +97,7 @@ private[services] final class TrackerMap[Key](
             trackerBySubmitter -= submitter
             tracker.close()
           }
-        case Waiting => // do nothing
+        case Waiting(_) => // do nothing
         case Failed(_) | Closed =>
           trackerBySubmitter -= submitter
       }
@@ -141,7 +141,8 @@ private[services] object TrackerMap {
   }
 
   private sealed trait AsyncResourceState[+T <: AutoCloseable]
-  private final case object Waiting extends AsyncResourceState[Nothing]
+  private final case class Waiting[T <: AutoCloseable](future: Future[T])
+      extends AsyncResourceState[T]
   private final case class Ready[T <: AutoCloseable](resource: T) extends AsyncResourceState[T]
   private final case object Closed extends AsyncResourceState[Nothing]
   private final case class Failed(exception: Throwable) extends AsyncResourceState[Nothing]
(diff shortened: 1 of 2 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
```diff
@@ -125,53 +140,56 @@ private[services] object TrackerMap {
...
 
-    def ifPresent[U](f: T => U): Option[U] = state.get() match {
-      case Ready(t) => Some(f(t))
-      case _ => None
-    }
+    private[TrackerMap] def currentState: AsyncResourceState[T] = state.get()
+
+    def withResource[U](f: T => Future[U])(implicit ex: ExecutionContext): Future[U] =
+      currentState match {
+        case Waiting(future) => future.flatMap(_ => withResource(f)) // try again
```
nicu-da, line 172: Isn't this going to go in a loop like crazy? What's the benefit vs the previous behavior?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
```diff
@@ -59,32 +62,44 @@ private[services] final class TrackerMap[Key](
...
+      trackerResource.currentState match {
+        case Ready(tracker) =>
+          if (currentTime - tracker.getLastSubmission > retentionNanos) {
+            logger.info(
+              s"Shutting down tracker for $submitter after inactivity of $retentionPeriod"
+            )(trackerResource.loggingContext)
+            trackerBySubmitter -= submitter
+            tracker.close()
+          }
+        case Waiting(_) => // do nothing
```
nicu-da, line 100: Shouldn't we call close here?

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
SamirTalwar-DA: Nope, it will wait on the future and then run again, which guarantees that the state will never be `Waiting` again.

The previous behavior would just sometimes do nothing, which was… not OK.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
SamirTalwar-DA: This is the cleanup function. It should only close stuff that's expired. If it's still waiting, it's not expired.

nicu-da: Alright got it. So the retention is only for ready trackers. Can we update the comment 
`  * @param retentionPeriod The minimum finite duration for which to retain idle trackers.` to have the same terminology

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
nicu-da: I still fail to see the difference. I assume this is a replacement for the previous `flatMap` implementation, where we would flatMap the resulting future directly, now we flatMap it through an intermediary call to `withResource`. If the `future` never completes it will still do nothing.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
SamirTalwar-DA: Can you be clear about what you'd like to change? I don't understand how this comment and the behavior are misaligned.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
SamirTalwar-DA: If the future never completes, it will do nothing. The previous behavior would just discard `f` entirely if it was in a waiting state. The new behavior waits on `future`, which guarantees that `state` will become `Ready` or `Failed`, and _then_ tries again.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
nicu-da: We're implicitly saying `ready` = `idle` so I would propose updating the comment to `...for which to retain ready trackers`

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
nicu-da: Ok, I need to take a closer look at the state transitions and how does the `future` changes between them as I had other conceptions about it. 

No reason to block this PR because of this though.

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/TrackerMap.scala
SamirTalwar-DA: Updated a bunch of comments to make this clearer.

nicu-da (approved):

PR #10665, opened 2021-08-25
Include committers in PartialTransaction root context

cocreature:
This gets us into a state where PartialTransaction always has the
authorization context which is much more sensible than having the
committers in speedy while the rest is in ptx.

The reason for having the split in the first place was the scenario
service but now that we create a new speedy machine per submission
that’s no longer a valid reason.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


remyhaemmerle-da (commented):
Much nicer!

remyhaemmerle-da (approved):

nickchapman-da (approved):
Nice!

PR #10668, opened 2021-08-25
Introduce a new `Offset` format [KVL-1063]

hubert-da:
CHANGELOG_BEGIN

- [Integration Kit] Changes the Offset format to contain a version and therefore reduces the highest index size by one byte

CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
@@ -3,13 +3,11 @@
 
 package com.daml.ledger.participant.state.kvutils
 
-import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream, DataOutputStream}
-
 import com.daml.ledger.offset.Offset
 
 /** Helper functions for generating 16 byte [[Offset]]s from integers.
   * The created offset will look as follows:
-  * | highest index (64 bits) | middle index (32 bits) | lowest index (32 bits) |
+  * | zeros (8 bits) | highest index (56 bits) | middle index (32 bits) | lowest index (32 bits) |
   * Leading zeros will be retained when generating the resulting offset bytes.
   *
   * Example usage:
@@ -18,6 +16,7 @@ import com.daml.ledger.offset.Offset
   *   - If you may have multiple records per block then use [[OffsetBuilder.fromLong]] with the index within the block as the second argument.
   *
   * @see com.daml.ledger.offset.Offset
+  * @see com.daml.ledger.offset.VersionedOffsetBuilder
   * @see com.daml.ledger.participant.state.kvutils.api.KeyValueParticipantStateReader
   */
 object OffsetBuilder {
@@ -26,50 +25,27 @@ object OffsetBuilder {
   private[kvutils] val lowestStart = 12
   private[kvutils] val end = 16
 
-  def onlyKeepHighestIndex(offset: Offset): Offset = {
-    val highest = highestIndex(offset)
-    fromLong(highest)
-  }
+  private val delegate = VersionedOffsetBuilder(version = 0)
 
-  def dropLowestIndex(offset: Offset): Offset = {
-    val (highest, middle, _) = split(offset)
-    fromLong(highest, middle)
-  }
+  def onlyKeepHighestIndex(offset: Offset): Offset = delegate.onlyKeepHighestIndex(offset)
 
-  def setMiddleIndex(offset: Offset, middle: Int): Offset = {
-    val (highest, _, lowest) = split(offset)
-    fromLong(highest, middle, lowest)
-  }
+  def dropLowestIndex(offset: Offset): Offset = delegate.dropLowestIndex(offset)
 
-  def setLowestIndex(offset: Offset, lowest: Int): Offset = {
-    val (highest, middle, _) = split(offset)
-    fromLong(highest, middle, lowest)
-  }
+  def setMiddleIndex(offset: Offset, middle: Int): Offset = delegate.setMiddleIndex(offset, middle)
 
-  def fromLong(first: Long, second: Int = 0, third: Int = 0): Offset = {
-    val bytes = new ByteArrayOutputStream
-    val stream = new DataOutputStream(bytes)
-    stream.writeLong(first)
-    stream.writeInt(second)
-    stream.writeInt(third)
-    Offset.fromByteArray(bytes.toByteArray)
-  }
+  def setLowestIndex(offset: Offset, lowest: Int): Offset = delegate.setLowestIndex(offset, lowest)
 
-  // `highestIndex` is used a lot, so it's worth optimizing a little rather than reusing `split`.
-  def highestIndex(offset: Offset): Long = {
-    val stream = new DataInputStream(new ByteArrayInputStream(offset.toByteArray))
-    stream.readLong()
-  }
+  def fromLong(first: Long, second: Int = 0, third: Int = 0): Offset =
+    delegate.of(first, second, third)
 
-  def middleIndex(offset: Offset): Int = split(offset)._2
+  def highestIndex(offset: Offset): Long = VersionedOffsetBuilder.highestIndex(offset)
 
-  def lowestIndex(offset: Offset): Int = split(offset)._3
+  def middleIndex(offset: Offset): Int = VersionedOffsetBuilder.middleIndex(offset)
 
-  def split(offset: Offset): (Long, Int, Int) = {
-    val stream = new DataInputStream(new ByteArrayInputStream(offset.toByteArray))
-    val highest = stream.readLong()
-    val middle = stream.readInt()
-    val lowest = stream.readInt()
+  def lowestIndex(offset: Offset): Int = VersionedOffsetBuilder.lowestIndex(offset)
+
+  private[kvutils] def split(offset: Offset): (Long, Int, Int) = {
+    val (highest, middle, lowest) = VersionedOffsetBuilder.split(offset)
     (highest, middle, lowest)
   }
 }
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
@@ -38,14 +38,14 @@ object OffsetBuilder {
   def fromLong(first: Long, second: Int = 0, third: Int = 0): Offset =
     delegate.of(first, second, third)
 
-  def highestIndex(offset: Offset): Long = VersionedOffsetBuilder.highestIndex(offset)
+  def highestIndex(offset: Offset): Long = delegate.highestIndex(offset)
 
-  def middleIndex(offset: Offset): Int = VersionedOffsetBuilder.middleIndex(offset)
+  def middleIndex(offset: Offset): Int = delegate.middleIndex(offset)
 
-  def lowestIndex(offset: Offset): Int = VersionedOffsetBuilder.lowestIndex(offset)
+  def lowestIndex(offset: Offset): Int = delegate.lowestIndex(offset)
 
   private[kvutils] def split(offset: Offset): (Long, Int, Int) = {
-    val (highest, middle, lowest) = VersionedOffsetBuilder.split(offset)
+    val (highest, middle, lowest) = delegate.split(offset)
     (highest, middle, lowest)
   }
 }
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/VersionedOffsetBuilder.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/VersionedOffsetBuilder.scala
@@ -59,27 +59,20 @@ class VersionedOffsetBuilder(version: Byte) {
     stream.writeInt(lowest)
     Offset.fromByteArray(bytes.toByteArray)
   }
-}
-
-object VersionedOffsetBuilder {
-
-  val MaxHighest: Long = (1L << 56) - 1
-
-  private val highestStartByte = 1
-  private val highestSizeBytes = 7
-  private val highestMask = 0x00ffffffffffffffL
-
-  def apply(version: Byte): VersionedOffsetBuilder = new VersionedOffsetBuilder(version)
 
   def version(offset: Offset): Byte = {
     val stream = toDataInputStream(offset)
-    stream.readByte()
+    val extractedVersion = stream.readByte()
+    validateVersion(extractedVersion)
+    extractedVersion
   }
 
   // `highestIndex` is used a lot, so it's worth optimizing a little rather than reusing `split`.
   def highestIndex(offset: Offset): Long = {
     val stream = toDataInputStream(offset)
-    readHighest(stream)
+    val (extractedVersion, highest) = readVersionAndHighest(stream)
+    validateVersion(extractedVersion)
+    highest
   }
 
   def middleIndex(offset: Offset): Int = split(offset)._2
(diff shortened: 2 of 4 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,100 @@
...
+    of(version, highest, middle, lowest)
   }
 
-  def fromLong(first: Long, second: Int = 0, third: Int = 0): Offset = {
+  def of(version: Byte, first: Long, second: Int = 0, third: Int = 0): Offset = {
     val bytes = new ByteArrayOutputStream
     val stream = new DataOutputStream(bytes)
-    stream.writeLong(first)
+    stream.writeByte(version.toInt)
+    writeSevenByteLong(first, stream)
```
hubert-da, line 57: Do we want to throw an exception if `first: Long` is too long?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilderSpec.scala
```diff
@@ -14,67 +14,68 @@ class OffsetBuilderSpec extends AnyWordSpec with Matchers with ScalaCheckDrivenP
...
-      val offset = OffsetBuilder.fromLong(0)
+      val offset = OffsetBuilder.of(OffsetBuilder.DefaultVersion, 0)
 
       offset should be(zeroOffset)
     }
 
     "always return an offset of the same length" in {
-      forAll { (highest: Long, middle: Int, lowest: Int) =>
-        val offset = OffsetBuilder.fromLong(highest, middle, lowest)
+      forAll { (version: Byte, highest: Int, middle: Int, lowest: Int) =>
```
hubert-da, line 23: I changed the `highest` to `Int` to reduce the boilerplate. If you're not fine with that, I'll have to use `whenever` or something similar.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
SamirTalwar-DA: Yes, good idea. I think you just want to check if `first < 0`.

With tests, please.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
+  *   - If you have one record per block then use [[OffsetBuilder.of]] with a single argument, the block ID.
+  *   - If you may have multiple records per block then use [[OffsetBuilder.of]] with the index within the block as the second argument.
   *
   * @see com.daml.ledger.offset.Offset
   * @see com.daml.ledger.participant.state.kvutils.api.KeyValueParticipantStateReader
   */
 object OffsetBuilder {
-  private[kvutils] val highestStart = 0
+  val DefaultVersion: Byte = 0
+  val HighestIndexSize = 7
```
SamirTalwar-DA, line 26: Does this need to be public?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilderSpec.scala
SamirTalwar-DA: Please use `whenever`, or preferably, `Gen.chooseNum`. We want to make sure that everything in the range works.

Something like this should work.

```suggestion
      val genHighest = Gen.chooseNum(0L, (1L << 56) - 1)
      forAll(arbitrary[Byte], genHighest, arbitrary[Int], arbitrary[Int]) {
        (version, highest, middle, lowest) =>
```

You can reuse that generator in other tests.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
+
+  private[kvutils] val highestStart = 1
   private[kvutils] val middleStart = 8
   private[kvutils] val lowestStart = 12
   private[kvutils] val end = 16
 
   def onlyKeepHighestIndex(offset: Offset): Offset = {
```
SamirTalwar-DA, line 33: I think you're going to need a method to extract the version, too.

hubert-da: Isn't this one enough: https://github.com/digital-asset/daml/pull/10668/files#diff-95c4c69783488dd7b39c4269e0d257f53a8f49dfdeb4994e6bfc8cf600ecc5d9R64 (`def version(offset: Offset): Byte = split(offset)._1`)?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
hubert-da: No, don't think so, thanks

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
SamirTalwar-DA: Oh, missed that! Thanks.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
+      ((readBuffer(3) & 255).toLong << 24) +
+      ((readBuffer(4) & 255) << 16) +
+      ((readBuffer(5) & 255) << 8) +
+      ((readBuffer(6) & 255) << 0)
+  }
+
+  private def writeSevenByteLong(first: Long, stream: DataOutputStream): Unit = {
+    val buffer = ByteBuffer.allocate(8)
+    buffer.putLong(first << 8) // shift to preserve the sign bit
+    stream.write(buffer.array(), 0, HighestIndexSize)
```
SamirTalwar-DA, line 102: I'm not understanding the shift, and why you want to preserve the sign bit. Why not something like:

```suggestion
    buffer.putLong(first)
    stream.write(buffer.array(), 1, HighestIndexSize + 1)
```

If you preserve the sign bit, won't shifting overwrite bit 2? Surely if the sign bit is set, we should throw or something?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
hubert-da: What do you think about using an unsigned long? Haven't ever used that but maybe it makes sense.

hubert-da: Also, I think I would check the maximum value as well. WDYT?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
+    val version = stream.readByte()
+    val highest = readSevenByteLong(stream)
     val middle = stream.readInt()
     val lowest = stream.readInt()
-    (highest, middle, lowest)
+    (version, highest, middle, lowest)
+  }
+
+  private def readSevenByteLong(stream: DataInputStream): Long = {
+    // mostly copied from DataInputStream.readLong
```
fabiotudone-da, line 87: I am against replicating this; how instead about letting `DataInputStream.readLong` read it from 8-bytes where the first one is an addition and is empty (given that the expected byte order in the buffer is big endian)?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilderSpec.scala
```diff
@@ -14,67 +14,68 @@ class OffsetBuilderSpec extends AnyWordSpec with Matchers with ScalaCheckDrivenP
 
   "OffsetBuilder" should {
```
fabiotudone-da, line 15: I think it's important to make sure we test corner cases around the 7-bytes `Long`.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
SamirTalwar-DA: Scala/Java has unsigned longs?

SamirTalwar-DA: And yes, you're right, you need to check the maximum value too; I was getting bits and bytes confused.

SamirTalwar-DA: This didn't show up for me:

> I think returning an Option or Try would be better.

I'd prefer not to; it would complicate the API for an edge case that should, in theory, never ever happen.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
-    val highest = stream.readLong()
+    val version = stream.readByte()
+    val highest = readSevenByteLong(stream)
     val middle = stream.readInt()
     val lowest = stream.readInt()
-    (highest, middle, lowest)
+    (version, highest, middle, lowest)
+  }
+
+  private def readSevenByteLong(stream: DataInputStream): Long = {
```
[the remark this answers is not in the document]
SamirTalwar-DA: "Seven-byte long" is, I believe, correct English. "Seven bytes long" doesn't read as well to me.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
fabiotudone-da: Exception at least, then, but not sure what are the consequences.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
hubert-da: Initially, I wasn't sure if we allow negative numbers. The tests were checking them. I had an impression that there can be a situation where the byte array looks like this: [-1, 2, 3, 4, 5, 6, 7, 8] and we have to shift to make it look like [-2, 3, 4, 5, 6, 7, 8, 0]. It may be total BS, so if we don't support negative numbers, let's not bother with that. `+ 1` is not needed, as it is a length.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
-  *   - If you may have multiple records per block then use [[OffsetBuilder.fromLong]] with the index within the block as the second argument.
+  *   - If you have one record per block then use [[OffsetBuilder.of]] with a single argument, the block ID.
+  *   - If you may have multiple records per block then use [[OffsetBuilder.of]] with the index within the block as the second argument.
   *
   * @see com.daml.ledger.offset.Offset
   * @see com.daml.ledger.participant.state.kvutils.api.KeyValueParticipantStateReader
   */
 object OffsetBuilder {
-  private[kvutils] val highestStart = 0
+  val DefaultVersion: Byte = 0
```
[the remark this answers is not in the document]
hubert-da: What do you propose?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
hubert-da: > Scala/Java has unsigned longs?

Since Java 8, I think. Let me try it out and see how it looks.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
hubert-da: How would you like to provide this additional byte?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
```diff
@@ -4,72 +4,101 @@
...
+      ((readBuffer(2) & 255).toLong << 32) +
+      ((readBuffer(3) & 255).toLong << 24) +
+      ((readBuffer(4) & 255) << 16) +
+      ((readBuffer(5) & 255) << 8) +
+      ((readBuffer(6) & 255) << 0)
+  }
+
+  private def writeSevenByteLong(first: Long, stream: DataOutputStream): Unit = {
+    val buffer = ByteBuffer.allocate(8)
+    buffer.putLong(first << 8) // shift to preserve the sign bit
```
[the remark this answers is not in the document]
hubert-da: Ignore this, I'm just going to go with:
```
buffer.putLong(first)
stream.write(buffer.array(), 1, highestIndexSize)
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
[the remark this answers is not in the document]
fabiotudone-da: I think we went with a `vN' package and a `Version` constant somewhere else and I think it makes sense. It would also conveniently (as we're going to have another one) make this builder the first version.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/OffsetBuilder.scala
SamirTalwar-DA: > Exception at least, then, but not sure what are the consequences.

Probably more pleasant than the previous consequences of exceeding the limit, which were to roll around to `Long.MinValue` (`-9223372036854775808`).

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/VersionedOffsetBuilder.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  def setLowestIndex(offset: Offset, lowest: Int): Offset = {
+    val (highest, middle, _) = split(offset)
+    of(highest, middle, lowest)
+  }
+
+  def of(first: Long, second: Int = 0, third: Int = 0): Offset = {
+    if (first < 0 || first > MaxHighest)
+      throw new IllegalArgumentException(s"Highest: $first is out of range [0, $MaxHighest]")
```
fabiotudone-da, line 48: Shouldn't we also validate that `second` and `third` are positive?

SamirTalwar-DA: I think that's a good idea.

hubert-da: Done

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/VersionedOffsetBuilderSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+          the[IllegalArgumentException] thrownBy VersionedOffsetBuilder(version).of(
+            highest,
+            middle,
+            lowest,
+          ) should have message s"requirement failed: lowest ($lowest) is lower than 0"
+      }
+    }
+
+    "fail on a wrong version" in {
+      forAll(genHighest, Gen.posNum[Int], Gen.posNum[Int], genDifferentVersions) {
```
miklos-da, line 68: Can you add test that we do check the offset's version for `VersionedOffsetBuilder.split` as well, please? That should cover the other setters/getters as well like `setLowestIndex`.

SamirTalwar-DA (approved):

miklos-da (approved):

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/VersionedOffsetBuilderSpec.scala
hubert-da: Done

PR #10669, opened 2021-08-25
[LEDGER-API-CLIENT] Allow to skip ledger id verification

realvictorprm:
if it only fails due to a missing token (given a user specified ledger id).

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
@@ -3,6 +3,8 @@
 
 package com.daml.ledger.client
 
+import com.daml.grpc.GrpcException
+
 import java.io.Closeable
 import java.util.concurrent.TimeUnit
 
@@ -32,6 +34,7 @@ import io.grpc.stub.AbstractStub
 import io.grpc.{Channel, ManagedChannel}
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
 final class LedgerClient private (
     val channel: Channel,
@@ -100,10 +103,28 @@ object LedgerClient {
     for {
       ledgerId <- new LedgerIdentityClient(
         stub(LedgerIdentityServiceGrpc.stub(channel), config.token)
-      )
-        .satisfies(config.ledgerIdRequirement)
+      ).satisfies(config.ledgerIdRequirement)
+        .transformWith {
+          case Success(Right(res)) =>
+            Future.successful(res)
+          case Success(Left(err)) =>
+            Future.failed(new IllegalArgumentException(err))
+          // This means we don't have a token for authorization
+          // such that ledger id verification is not possible.
+          // In this case we try to use the ledgerId we want to verify against which is hopefully provided
+          // via the config.
+          case Failure(e @ GrpcException.UNAUTHENTICATED()) =>
+            config.ledgerIdRequirement.optionalLedgerId
+              .map(LedgerId(_))
+              .fold(Future.failed[LedgerId](e))(Future.successful(_))
+          case Failure(t) => Future.failed(t)
+        }
     } yield {
-      new LedgerClient(channel, config, ledgerId)
+      new LedgerClient(
+        channel,
+        config,
+        ledgerId,
+      )
     }
   }
 
--- a/ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/LedgerClientAuthIT.scala
+++ b/ledger/ledger-api-client/src/it/scala/com/digitalasset/ledger/client/LedgerClientAuthIT.scala
@@ -18,6 +18,7 @@ import com.daml.platform.sandboxnext.SandboxNextFixture
 import org.scalatest.Inside
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AsyncWordSpec
+import scalaz.syntax.tag._
 
 final class LedgerClientAuthIT
     extends AsyncWordSpec
@@ -86,7 +87,7 @@ final class LedgerClientAuthIT
       }
     }
 
-    "it does not have a token" should {
+    "it does not have a token & no ledgerIdRequirement specified" should {
       "fail to construct" in {
         for {
           exception <- LedgerClient(channel, ClientConfigurationWithoutToken).failed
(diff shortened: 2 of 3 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
```diff
@@ -100,10 +103,28 @@ object LedgerClient {
     for {
       ledgerId <- new LedgerIdentityClient(
         stub(LedgerIdentityServiceGrpc.stub(channel), config.token)
-      )
-        .satisfies(config.ledgerIdRequirement)
+      ).satisfies(config.ledgerIdRequirement)
```
cocreature, line 106: how do you set the ledger id in the JSON API on startup for this to work out? Currently users do not have to specify one and forcing this is both clunky and backwards compat breaking.

cocreature: actually I think I’m even more confused. You are not actually skipping the call to the ledger identity service so how does this work without a ledger id?

realvictorprm: We try to verify the ledger id first (Which so far my understanding is should also work without token) and if that fails due to an auth error we just pick up the ledger id the user specified.

I still think about better ways to solve this without invasive changes.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
```diff
@@ -100,10 +103,28 @@ object LedgerClient {
...
+        .transformWith {
+          case Success(Right(res)) =>
+            Future.successful(res)
+          case Success(Left(err)) =>
+            Future.failed(new IllegalArgumentException(err))
+          // This means we don't have a token for authorization
+          // such that ledger id verification is not possible.
+          // In this case we try to use the ledgerId we want to verify against which is hopefully provided
+          // via the config.
+          case Failure(e @ GrpcException.UNAUTHENTICATED()) =>
```
cocreature, line 116: how do you set the ledger id in the JSON API on startup for this to work out? Currently users do not have to specify one and forcing this is both clunky and backwards compat breaking.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
cocreature: sorry reshuffled my comment to the right place I misread the code first. I think it still applies though.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
cocreature: @realvictorprm currently the JSON API does not set `optionalLedgerId` on startup so this is not going to work without changing that.

realvictorprm: Shouldn't this be backward compatible, because prior in the case of unauthenticated the application would have failed? My changes just add an avenue to handle this without crashing.

realvictorprm: > @realvictorprm currently the JSON API does not set optionalLedgerId on startup so this is not going to work without changing that.

I initially planned to add that change in my PR which removes the token file, however I can already add it in this PR.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
stefanobaghino-da: > Which so far my understanding is should also work without token

No, `satisfies` does eagerly call the `LedgerIdentityService`. What changes is whether or not the user-provided ledger identifier will be compared with the actual one. If the Ledger API requires calls to be authorized, the `LedgerIdentityService` requires it as well.

> we just pick up the ledger id the user specified

This would require to know the ledger identifier beforehand, which we cannot rely upon.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
cocreature: It works if you keep the package token where we will never hit this. What I’m trying to say is that this does not help you without the package token:
1. In that case you cannot query for the ledger id.
2. You also don’t know it on startup since users do not specify it and I don’t think they should have to (clunky & backwards compat breaking unless you still want to support the case where someone specifies a package token).

realvictorprm: OK I see the point for the remove-token-file PR. That change would force users to instead of specifying the token file, to specify the ledger id.

realvictorprm: Geez, I already have locally a non working version where I made the ledger id optional, however that is causing a shit-ton of changes across the whole code base.

stefanobaghino-da (commented):
This approach won't work unfortunately. I believe a possible approach is to make the services `lazy`, add a `lazy val ledgerId` and have the (private) constructor accept a function to retrieve the ledger identifier that will be invoked when the `ledgerId` is first accessed.

Does this make sense?

stefanobaghino-da:
> Allow to skip ledger id verification

Rather than skipping ledger id verification, I think the point is to rather delay calling the `LedgerIdentityService` until the ledger id is actually needed (that is, when a service which requires it -- like the `PackageService` -- is used to perform a call).

realvictorprm:
shooo Mergify :P

cocreature:
> This approach won't work unfortunately. I believe a possible approach is to make the services `lazy`, add a `lazy val ledgerId` and have the (private) constructor accept a function to retrieve the ledger identifier that will be invoked when the `ledgerId` is first accessed.
> 
> Does this make sense?

alternatively you can keep the services strict but make the ledger id passed to them optional. In that scenario you never need to call the ledger identity service. It is really only used for two purposes:
1. Validate that the one passed on a specific request matches the one we got on startup. We can simply skip that check and let the ledger blow up server side.
2. On some services we have convenience methods that don’t require you to pass in the ledger id and instead use the one we already know. We could add the ledger id as an extra optional parameter to these and then use the global ledger id, the one passed to the request or error out if neither is specified (and if both are specified error out if they mismatch).

This is backwards compatible and it works well for the JSON API since we know the ledger id on each request by virtue of having it in the token.

realvictorprm:
Will continue working then on my initial approach which I only have offline

PR #10677, opened 2021-08-25
ledger-on-memory: If the index provided already exists, fail.

SamirTalwar-DA:
Pointing ledger-on-memory at an existing index is a recipe for disaster, as the ledger will have no entries but the index might have some.

This enforces an empty index by failing if any migrations have been applied.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerStartupMode.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerStartupMode.scala
@@ -17,4 +17,6 @@ object IndexerStartupMode {
 
   case object MigrateOnEmptySchemaAndStart extends IndexerStartupMode
 
+  case object EnforceEmptySchemaAndMigrate extends IndexerStartupMode
+
 }
--- a/ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
@@ -99,7 +99,7 @@ private[platform] class FlywayMigrations(
           throw StoppedProgressing(pendingMigrations)
         } else {
           logger.debug(
-            s"Concurrent migration has reduced the pending migrations set to ${pendingMigrations}, waiting until pending set is empty.."
+            s"Concurrent migration has reduced the pending migrations set to $pendingMigrations, waiting until pending set is empty.."
           )
           Thread.sleep(1000)
           flywayMigrationDone(retries - 1, Some(pendingMigrations))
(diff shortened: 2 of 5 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerStartupMode.scala
```diff
@@ -17,4 +17,6 @@ object IndexerStartupMode {
 
   case object MigrateOnEmptySchemaAndStart extends IndexerStartupMode
 
+  case object EnforceEmptySchemaAndMigrate extends IndexerStartupMode
```
nicu-da, line 20: Would be nice to add some documentation on the exact difference between these startup modes. I think the names are pretty self-explanatory already but it's difficult to grasp the small difference between `MigrateOnEmptySchemaAndStart` and `EnforceEmptySchemaAndMigrate` without mentally desiccating the names for a while.

nicu-da (approved):
LGTM, just a small comment with regards to docs

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -153,6 +153,36 @@ private[platform] class FlywayMigrations(
...
+      val flyway = configurationBase(dbType, enableAppendOnlySchema, additionalMigrationPaths)
+        .dataSource(ds)
+        .ignoreFutureMigrations(false)
+        .load()
+      logger.info(
+        "Ensuring Flyway migration has either not started or there are no pending migrations..."
+      )
+      val flywayInfo = flyway.info()
+
+      flywayInfo.applied().length match {
```
tudor-da, line 170: Actually this limits something that is not a problem. We have a problem if there are any offsets in the index. Should we make it so restrictive?

tudor-da: I don't know if there are any other use-cases that we would be impeding other than testing migrations, though..

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -201,15 +231,19 @@ private[platform] object FlywayMigrations {
...
+      extends RuntimeException(s"Ran out of retries with $pendingMigrations migrations remaining")
   case class StoppedProgressing(pendingMigrations: Int)
-      extends RuntimeException(s"Stopped progressing with ${pendingMigrations} migrations")
+      extends RuntimeException(s"Stopped progressing with $pendingMigrations migrations")
   case class MigrateOnEmptySchema(appliedMigrations: Int, pendingMigrations: Int)
       extends RuntimeException(
-        s"Asked to migrate-on-empty-schema, but encountered neither an empty database with ${appliedMigrations} " +
-          s"migrations already applied nor a fully-migrated databases with ${pendingMigrations} migrations pending."
+        s"Asked to migrate-on-empty-schema, but encountered neither an empty database with $appliedMigrations " +
+          s"migrations already applied nor a fully-migrated databases with $pendingMigrations migrations pending."
```
miklos-da, line 243:
```suggestion
          s"migrations already applied nor a fully-migrated database with $pendingMigrations migrations pending."
```

ledger/participant-integration-api/src/main/scala/platform/indexer/IndexerStartupMode.scala
miklos-da: Do we still use all of these startup modes?

hubert-da (commented):
LGTM but I don't know too much about Flyway

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
SamirTalwar-DA: I agree, but this is only used by a test ledger, so I don't really mind being a little more restrictive. The goal is to stop us doing silly things like using it to test migrations.

PR #10679, opened 2021-08-25
Command dedup: migrate kvutils to use v2 services [KVL-1049]

nicu-da:
- Add extra fields to `kvutils.proto` which are required to update to v2 (submission id and deduplication period)
- Migrate everything to use v2 services only and drop the usage of adapters
- Enable api deduplication for existing kv ledgers

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/api/KeyValueParticipantState.scala
```diff
@@ -40,15 +40,22 @@ class KeyValueParticipantState(
     reader: LedgerReader,
     writer: LedgerWriter,
     metrics: Metrics,
+    apiDeduplicationEnabled: Boolean = false,
```
SamirTalwar-DA, line 43: Can't we just hard-code this to `true` for now?

ledger/ledger-on-memory/src/test/suite/scala/com/daml/ledger/on/memory/InMemoryLedgerWriterSpec.scala
```diff
@@ -45,7 +46,7 @@ class InMemoryLedgerWriterSpec
         )(any[ExecutionContext])
       )
         .thenReturn(
-          Future.successful(SubmissionResult.InternalError("Validation failed with an exception"))
+          Future.successful(SubmissionResult.SynchronousError(Status()))
```
SamirTalwar-DA, line 49: Can you pull out this value and then assert with `should be` below?

ledger/participant-state/kvutils/BUILD.bazel
```diff
@@ -51,13 +51,15 @@ da_scala_library(
...
         "//ledger/caching",
+        "//ledger/ledger-api-domain",
         "//ledger/ledger-api-health",
         "//ledger/ledger-configuration",
         "//ledger/ledger-configuration/protobuf:ledger_configuration_proto_java",
         "//ledger/ledger-offset",
         "//ledger/ledger-resources",
         "//ledger/metrics",
         "//ledger/participant-state",
+        "//ledger-api/grpc-definitions:ledger_api_proto_scala",
```
SamirTalwar-DA, line 62: Please move this above all the `//ledger` entries, as below. I'm surprised the autoformatter didn't do this for you.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -183,20 +189,59 @@ message DamlTransactionBlindingInfo {
...
+    InconsistentKeys internal_inconsistent_keys = 12;
+    InconsistentContracts external_inconsistent_contracts = 13;
+    DuplicateKeys external_duplicate_keys = 14;
+    InconsistentKeys external_inconsistent_keys = 15;
+    MissingInputState missing_input_state = 16;
+    RecordTimeOutOfRange record_time_out_of_range = 17;
+    CausalMonotonicityViolated causal_monotonicity_violated = 18;
+    SubmittingPartyNotKnownOnLedger submitting_party_not_known_on_ledger = 19;
+    PartiesNotKnownOnLedger parties_not_known_on_ledger = 20;
+    InvalidParticipantState  invalid_participant_state = 21;
```
SamirTalwar-DA, line 215:
```suggestion
    InvalidParticipantState invalid_participant_state = 21;
```

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -183,20 +189,59 @@ message DamlTransactionBlindingInfo {
 // A transaction rejection entry.
 message DamlTransactionRejectionEntry {
   DamlSubmitterInfo submitter_info = 1;
-
   reserved 5; // was reason.maximum_record_time_exceeded
 
   oneof reason {
+    InvalidLedgerTime invalid_ledger_time = 9;
+    SubmitterCannotActViaParticipant submitter_cannot_act_via_participant = 8;
```
SamirTalwar-DA, line 196: Why are these two not deprecated? Are they used by both?

If they are, can you add a comment explaining this?

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -183,20 +189,59 @@ message DamlTransactionBlindingInfo {
...
+    //v1 reasons DEPRECATED
```
SamirTalwar-DA, line 197:
```suggestion
    // v1 reasons (these are DEPRECATED)
```

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -183,20 +189,59 @@ message DamlTransactionBlindingInfo {
...
     Inconsistent inconsistent = 2;
     Disputed disputed = 3;
     ResourcesExhausted resources_exhausted = 4;
     Duplicate duplicate_command = 6;
     PartyNotKnownOnLedger party_not_known_on_ledger = 7;
-    SubmitterCannotActViaParticipant submitter_cannot_act_via_participant = 8;
-    InvalidLedgerTime invalid_ledger_time = 9;
+    //v2 reasons
```
SamirTalwar-DA, line 203:
```suggestion
    // v2 reasons
```

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -183,20 +189,59 @@ message DamlTransactionBlindingInfo {
 // A transaction rejection entry.
 message DamlTransactionRejectionEntry {
   DamlSubmitterInfo submitter_info = 1;
-
   reserved 5; // was reason.maximum_record_time_exceeded
 
   oneof reason {
```
SamirTalwar-DA, line 194: These changes make me very happy.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -132,24 +135,47 @@ private[state] object Conversions {
...
+    }
+    CompletionInfo(
       actAs = subInfo.getSubmittersList.asScala.toList.map(Ref.Party.assertFromString),
       applicationId = Ref.LedgerString.assertFromString(subInfo.getApplicationId),
       commandId = Ref.LedgerString.assertFromString(subInfo.getCommandId),
-      deduplicateUntil = parseTimestamp(subInfo.getDeduplicateUntil).toInstant,
+      optDeduplicationPeriod = deduplicationPeriod,
+      submissionId = Ref.SubmissionId.assertFromString(
+        subInfo.getSubmissionId
+      ), // FIXME this can be missing, what to do what to do
```
SamirTalwar-DA, line 175: If there's no submission ID, we need to generate one somewhere.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejection.scala
```diff
@@ -8,40 +8,29 @@ import java.time.Instant
...
       override val description: String =
         "DuplicateKeys: the transaction contains a duplicate key"
-
-      override def toStateV1RejectionReason: v1.RejectionReason =
-        v1.RejectionReasonV0.Disputed(description)
     }
 
     object InconsistentKeys extends Rejection {
       override val description: String =
         "InconsistentKeys: the transaction is internally inconsistent"
```
SamirTalwar-DA, line 33: Please knock out these extra empty lines.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -70,21 +72,77 @@ private[transaction] class Rejections(metrics: Metrics) {
...
+        builder.setMissingInputState(
+          MissingInputState.newBuilder().setKey(key)
+        )
+      case Rejection.InvalidParticipantState(error) =>
+        builder.setInvalidParticipantState(
+          InvalidParticipantState
+            .newBuilder()
+            .setDetails(
+              error.getMessage
+            )
```
SamirTalwar-DA, line 102:
```suggestion
            .setDetails(error.getMessage)
```

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala
```diff
@@ -327,15 +320,16 @@ abstract class ParticipantStateIntegrationSpecBase(implementationName: String)(i
         }
       }
 
-      "reject duplicate commands" in participantState.use { ps =>
-        val commandIds = ("X1", "X2")
-
+      //ignore as we don't support command deduplication currently at this level
```
SamirTalwar-DA, line 323: Please make sure we don't lose track of this.

buf.yaml
```diff
@@ -33,3 +33,4 @@ breaking:
   ignore_only:
     FIELD_SAME_ONEOF: # ignore field_same_oneof to not trigger an error from moving an existing field into a new oneof
       - com/daml/ledger/api/v1/commands.proto
+      - com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```
SamirTalwar-DA, line 36: Please add a newline at the end here.

ledger/ledger-on-memory/src/test/suite/scala/com/daml/ledger/on/memory/InMemoryLedgerReaderWriterIntegrationSpec.scala
```diff
@@ -44,6 +44,10 @@ class InMemoryLedgerReaderWriterIntegrationSpec
         engine = Engine.DevEngine(),
         committerExecutionContext = committerExecutionContext,
       )
-    } yield new KeyValueParticipantState(readerWriter, readerWriter, metrics)
+    } yield new KeyValueParticipantState(
+      readerWriter,
+      readerWriter,
+      metrics,
+    )
```
SamirTalwar-DA, line 51: I think this change was probably unnecessary.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -268,6 +294,159 @@ private[state] object Conversions {
...
+        code.value,
+        message,
+        Seq.empty,
+      )
+    }
+
+    val status = entry.getReasonCase match {
+      case DamlTransactionRejectionEntry.ReasonCase.INVALID_LEDGER_TIME =>
+        val rejection = entry.getInvalidLedgerTime
+        Some(
```
SamirTalwar-DA, line 311: Maybe pull out `Some(…)` here.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -268,6 +294,159 @@ private[state] object Conversions {
...
+  def decodeTransactionRejectionEntry(
+      entry: DamlTransactionRejectionEntry
+  ): Option[FinalReason] = {
+    def buildStatus(code: Code, message: String) = {
+      Status.of(
+        code.value,
+        message,
+        Seq.empty,
+      )
+    }
```
SamirTalwar-DA, line 306: You can shorten this, and perhaps pull it out of the enclosing function, as it doesn't require the method parameter.

```suggestion
    def buildStatus(code: Code, message: String) =
      Status.of(code.value, message, Seq.empty)
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
```diff
@@ -70,21 +73,77 @@ private[transaction] class Rejections(metrics: Metrics) {
...
-      case RejectionReasonV0.Inconsistent(reason) =>
-        builder.setInconsistent(Inconsistent.newBuilder.setDetails(reason))
-      case RejectionReasonV0.Disputed(reason) =>
-        builder.setDisputed(Disputed.newBuilder.setDetails(reason))
-      case RejectionReasonV0.ResourcesExhausted(reason) =>
-        builder.setResourcesExhausted(ResourcesExhausted.newBuilder.setDetails(reason))
-      case RejectionReasonV0.PartyNotKnownOnLedger(reason) =>
-        builder.setPartyNotKnownOnLedger(PartyNotKnownOnLedger.newBuilder.setDetails(reason))
-      case RejectionReasonV0.SubmitterCannotActViaParticipant(details) =>
+    rejection match {
```
SamirTalwar-DA, line 76: Can we add some tests that check this in combination with _Conversions.scala_? I am thinking some table-driven checks that verify that `Rejection.Xyz` converts to `Status.of(code1, message1)`, etc.

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala
```diff
@@ -472,7 +467,7 @@ abstract class ParticipantStateIntegrationSpecBase(implementationName: String)(i
 
           offset2 should be(toOffset(2))
           inside(update2) { case CommandRejected(_, _, reason) =>
-            reason should be(a[RejectionReasonV0.PartyNotKnownOnLedger])
+            reason should be(a[FinalReason])
```
SamirTalwar-DA, line 470: Can you check the exact reason?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumptionSpec.scala
```diff
@@ -114,10 +114,9 @@ class KeyValueConsumptionSpec extends AnyWordSpec with Matchers {
 
     "generate a rejection entry for a transaction if record time is out of time bounds" in {
       def verifyCommandRejection(actual: Option[Update]): Unit = actual match {
-        case Some(Update.CommandRejected(recordTime, submitterInfo, reason)) =>
+        case Some(Update.CommandRejected(recordTime, completionInfo, _)) =>
           recordTime shouldBe aRecordTime
-          submitterInfo shouldBe Conversions.parseSubmitterInfo(someSubmitterInfo)
-          reason shouldBe a[RejectionReasonV0.InvalidLedgerTime]
```
SamirTalwar-DA, line 120: Can we assert the exact rejection reason?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/api/KeyValueParticipantState.scala
nicu-da: Done

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
nicu-da: Exactly, these two errors kept their usage. Will add a comment.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
nicu-da: There's one None so not sure how to make the compiler happy without this.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/Rejections.scala
nicu-da: Yes, this is the next step.
Actually plan on moving this code to conversions as well so that it's in the same place, and then add tests for 2-way conversions. 

On another note I was also thinking of using these conversions as part of the `RejectionReasonUpdateNormalizer`, this would at least add some form of validation on the rejection as well.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
SamirTalwar-DA: Sorry, must have missed that.

SamirTalwar-DA:
Looking very solid, IMO.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -93,6 +107,195 @@ class ConversionsSpec extends AnyWordSpec with Matchers {
...
+            rejection,
+          )
+          .build()
+        Conversions
+          .decodeTransactionRejectionEntry(encodedEntry)
+          .value
+          .code shouldBe expectedCode.value()
+      }
+
+      "convert validation failure" in {
```
SamirTalwar-DA, line 127: Consider using `Table` for these tests, to make it terser and potentially clearer?

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/ParticipantStateIntegrationSpecBase.scala
```diff
@@ -327,15 +322,16 @@ abstract class ParticipantStateIntegrationSpecBase(implementationName: String)(i
         }
       }
 
-      "reject duplicate commands" in participantState.use { ps =>
-        val commandIds = ("X1", "X2")
-
+      //ignore as we don't support command deduplication currently at this level
```
miklos-da, line 325: This is only testing committer-side command deduplication, doesn't it?

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -184,20 +190,35 @@ message DamlTransactionBlindingInfo {
...
     Disputed disputed = 3;
     ResourcesExhausted resources_exhausted = 4;
     Duplicate duplicate_command = 6;
     PartyNotKnownOnLedger party_not_known_on_ledger = 7;
-    SubmitterCannotActViaParticipant submitter_cannot_act_via_participant = 8;
-    InvalidLedgerTime invalid_ledger_time = 9;
+
+    // v2 reasons
+    ValidationFailure validation_failure = 10;
+    DuplicateKeys internal_duplicate_keys = 11;
```
miklos-da, line 209: Was there a reason to diverge from the respective Scala rejection names such as `InternallyInconsistentTransaction.DuplicateKeys`? In order to be consistent with the Scala class names I'd suggest using `internally_duplicate_keys`, `internally_inconsistent_keys`, etc. as field names here, or even better, create a separate message mirroring the Scala class and declare two separate fields for rejections because of internally vs. externally inconsistent transactions so that we keep this message easier to read.
E.g., this message could look something like:
```protobuf
message DamlTransactionRejectionEntry {
  ...
  oneof reason {
    <common rejection reasons>
    //
    // Rejections used by participant.state.v1 API.
    // Note that these are deprecated.
    //
    Inconsistent inconsistent = 2;
    Disupted disupted = 3;
    ...

    //
    // Rejections used by participant.state.v2 API.
    //
    ValidationFailure validation_failure = 10;
    InconsistentTransaction internally_inconsistent_transaction = ...;
    InconsistentTransaction externally_inconsistent_transaction = ...;
    ...
  }
}
```
Another option to consider is to introduce a separate message encompassing all rejection reasons specific to `participant.state.v2`.
@nicu-da What do you think?

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -374,7 +395,7 @@ message DamlCommandDedupKey {
   string application_id = 2;
   string command_id = 3;
 }
-
+// should this be removed? how do we handle transitions?
```
miklos-da, line 398: This may not be removed yet. For command deduplication related changes we'll define a process for the SDK upgrade that'll ensure that no old deduplication state interferes with the new code.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -469,6 +490,8 @@ message DamlContractKeyState {
 
 // Errors
 
+// V1
```
miklos-da, line 493: If you want to mark a block of messages belonging to a version, please either create a multi-line header such as the below:
```protobuf
//
// Used by participant.state.v1 API.
//
```
or add a comment e.g. `// Used by participant.state.v1 API.` to each message.

miklos-da (commented):
Looks good! We should consider some alternatives for changing `daml_kvutils.proto` so that the schema is easier to maintain.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
nicu-da: Makes sense, I like a lot more the header field for versioning. Will update that and naming.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -540,3 +568,33 @@ message TimedOut {
 message AlreadyExists {
   string details = 1;
 }
+
+// V2
+
+message ValidationFailure {
+    string details = 1;
```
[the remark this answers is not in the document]
nicu-da: > Looks good! We should consider some alternatives for changing `daml_kvutils.proto` so that the schema is easier to maintain.

I would split it a bit into multiple files tbh. All the rejections code, for example, it would make a lot more sense to have it in its own file.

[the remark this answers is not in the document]
nicu-da: Digging through IJ to fix the formatter that I use as it's not the first time it causes issues for me

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
nicu-da: I think mirroring the scala namespacing for "external" vs "internal" doesn't really make sense, because internally we have only "inconsistent" and "duplicate" keys while externally we also have "inconsistent contracts" so the models don't really map 1:1.

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityChecker.scala
```diff
@@ -125,13 +125,12 @@ class IntegrityChecker[LogResult](
   private def indexStateUpdates(
       config: Config,
       metrics: Metrics,
-      replayingReadService: ReplayingReadService,
+      readService: ReplayingReadService,
```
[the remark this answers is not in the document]
nicu-da: 👍 agree, the original name makes a lot more sense

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
nicu-da: Ok, done this by using the entry id as that's the only constant which is unique in the entry that we have

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
nicu-da: Done, I'm usually not the biggest fan as it tends to make tests a lot more unreliable and I would take the extra boilerplate to have the test easier to read, but in this case I think it makes sense as the inputs are really simple

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/ModelConformanceValidatorSpec.scala
```diff
@@ -188,9 +173,9 @@ class ModelConformanceValidatorSpec
           aTransactionEntry,
         )
       inside(step) { case StepStop(logEntry) =>
-        logEntry.getTransactionRejectionEntry.hasInconsistent shouldBe true
-        logEntry.getTransactionRejectionEntry.getInconsistent.getDetails should startWith(
-          "Missing input state for key contract_id: \"#inputContractId\""
+        logEntry.getTransactionRejectionEntry.getReasonCase shouldBe DamlTransactionRejectionEntry.ReasonCase.MISSING_INPUT_STATE
+        logEntry.getTransactionRejectionEntry.getMissingInputState.getKey.toString should startWith(
```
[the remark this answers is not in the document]
nicu-da: I agree that it would make things easier and nicer to read, just tried to only "update" the tests by keeping them as they were as much as possible. Will try to fix this as well

fabiotudone-da (commented):
I believe this PR (also) wants to add a (proto-wire-)backward-compatible increment to the protos that is also a stable one. If so it should also update `buf_image.bin`.

nicu-da:
> I believe this PR (also) wants to add a (proto-wire-)backward-compatible increment to the protos that is also a stable one. If so it should also update `buf_image.bin`.

I was thinking of doing that in another PR and removing the "exception" that we added in the buf config. Wdyt @fabiotudone-da ?

fabiotudone-da:
> > I believe this PR (also) wants to add a (proto-wire-)backward-compatible increment to the protos that is also a stable one. If so it should also update `buf_image.bin`.
> 
> I was thinking of doing that in another PR and removing the "exception" that we added in the buf config. Wdyt @fabiotudone-da ?

So are more changes planned to the protos that are not backward compatible with the ones here? If so, sure; in general consider that your changes won't be stable until you do, including in the next SDK snapshot (if you don't make it before that).

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ClosedWorldIT.scala
```diff
@@ -28,7 +28,11 @@ class ClosedWorldIT extends LedgerTestSuite {
         .create(payer, Iou(payer, binding.Primitive.Party("unallocated"), onePound))
         .mustFail("referencing an unallocated party")
     } yield {
-      assertGrpcError(failure, Status.Code.INVALID_ARGUMENT, "Party not known on ledger")
+      assertGrpcError(
+        failure,
+        Status.Code.INVALID_ARGUMENT,
+        "Parties not known on ledger [unallocated]",
+      )
```
SamirTalwar-DA, line 35:
```suggestion
      assertGrpcError(
        failure,
        Status.Code.INVALID_ARGUMENT,
        "Part(y|ies) not known on ledger",
      )
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ContractKeysIT.scala
```diff
@@ -195,7 +195,7 @@ final class ContractKeysIT extends LedgerTestSuite {
       assertGrpcError(
         duplicateKeyFailure,
         Status.Code.ABORTED,
-        Some(Pattern.compile("Inconsistent")),
+        Some(Pattern.compile("DuplicateKeys: at least one contract key is not unique")),
```
SamirTalwar-DA, line 198: Sounds like we need to keep this particular assertion. You can fix the error by changing the message in `ActiveLedgerStateManager` instead.

SamirTalwar-DA: No, wait, I misunderstood. I think this is a legit change. My solution was correct though: you can fix the failure in Sandbox Classic by changing the message in `ActiveLedgerStateManager`.

nicu-da: This message, in particular, comes from `ExternallyInconsistentTransaction.DuplicateKeys`, I assume that one we need to change to `Inconsistent: DuplicateKey: contract key is not unique`

SamirTalwar-DA: I guess whatever it was before?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -126,24 +137,61 @@ private[state] object Conversions {
...
+      case DeduplicationPeriodCase.DEDUPLICATION_OFFSET =>
+        Some(
+          DeduplicationPeriod.DeduplicationOffset(
+            Offset.fromHexString(Ref.HexString.assertFromString(subInfo.getDeduplicationOffset))
+          )
+        )
+      case DeduplicationPeriodCase.DEDUPLICATE_UNTIL =>
+        //backwards compatibility
+        // we no longer support deduplicate until, and in v1 it was always relative to record time
+        // to keep backwards compatibility and simulate the new deduplication period we calculate the duration from the record time
```
miklos-da, line 173: Feel free to ignore/edit, just a suggestion for improving the wording here:
```suggestion
        // For backwards compatibility with rejections generated by participant.state.v1 API.
        // As the deduplicate until timestamp is always relative to record time, we take the duration
        // between record time and the previous timestamp as the deduplication period (duration).
```

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/UpdateNormalizer.scala
```diff
@@ -67,45 +69,17 @@ object RecordTimeNormalizer extends UpdateNormalizer {
...
-      val newReason = reason match {
-        case RejectionReasonV0.Disputed(description) =>
-          RejectionReasonV0.Inconsistent(description)
-        case RejectionReasonV0.InvalidLedgerTime(description) =>
-          RejectionReasonV0.Inconsistent(description)
-        case _ => reason
-      }
-      commandRejected.copy(reason = newReason)
+    case commandRejected @ Update.CommandRejected(_, _, _) =>
+      val newReason = FinalReason(
```
miklos-da, line 77: This is a radical normalization. Can we do something more refined here, please? I.e., are there cases that we could possibly recognize and map between before vs. after?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/CompletionInfo.scala
```diff
@@ -31,15 +31,19 @@ import com.daml.logging.entries.{LoggingValue, ToLoggingValue}
   *                               this deduplication information, it MUST adhere to the deduplication
   *                               guarantee under a sensible interpretation of the corresponding
   *                               [[CompletionInfo.optDeduplicationPeriod]].
+  *
   * @param submissionId           An identifier for the submission that allows an application to
   *                               correlate completions to its submissions.
+  *
+  *                                Optional as entries created by the participant.state.v1 API do not have this filled.
```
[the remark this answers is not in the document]
nicu-da: Yes, I actually liked more like this because it's visually indicative that this text still is part of the submissionId documentation but I agree that being consistent is more important so I'll remove the extra indentation.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -93,6 +94,190 @@ class ConversionsSpec extends AnyWordSpec with Matchers {
...
+      "calculate duration for deduplication for backwards compatibility with deduplicate until" in {
+        val completionInfo = parseCompletionInfo(
+          recordTime,
+          submitterInfo.setDeduplicateUntil(buildTimestamp(recordTime.plusSeconds(30))).build(),
+        )
+        completionInfo.optDeduplicationPeriod.value shouldBe DeduplicationPeriod
+          .DeduplicationDuration(Duration.ofSeconds(30))
+      }
+
+      "handle deduplication which is the past relative to record time" in {
```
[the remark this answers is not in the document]
nicu-da: Exactly, added this detail to the test description

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/UpdateNormalizer.scala
nicu-da: I agree that this is basically a nuke. 
I'm not that knowledgeable as to how exactly does the integrity checker work (my assumption is that we have exports of the ledger and we replay those using the current code and compare. If that's the case then we could in theory just drop this normalizer as it should be compatible without it. Will test this before doing so.

SamirTalwar-DA: There really aren't many; it would be a huge endeavor to create a mapping.

If we are to do this, I suggest we do it later.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ClosedWorldIT.scala
```diff
@@ -28,7 +30,11 @@ class ClosedWorldIT extends LedgerTestSuite {
         .create(payer, Iou(payer, binding.Primitive.Party("unallocated"), onePound))
         .mustFail("referencing an unallocated party")
     } yield {
-      assertGrpcError(failure, Status.Code.INVALID_ARGUMENT, "Party not known on ledger")
+      assertGrpcError(
+        failure,
+        Status.Code.INVALID_ARGUMENT,
+        Some(Pattern.compile("Part(y|ies) not known on ledger")),
```
SamirTalwar-DA, line 36: I think you can use the `.r` method here to create a `Pattern`.

```suggestion
        Some("Part(y|ies) not known on ledger".r),
```

SamirTalwar-DA (approved):

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
nicu-da: Just double-checking this.
For the new API, to construct this message we still call the deprecated `.getDeduplicateUntil`. This means that the field is gonna be empty when `deduplication_period` is set but not in the deprecated format. 
Should this field be updated to the new deduplication period format as well? @miklos-da

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/KVTest.scala
```diff
@@ -430,13 +431,15 @@ object KVTest {
...
+      submissionId: Ref.SubmissionId,
   ): SubmitterInfo =
     SubmitterInfo(
       actAs = List(submitter),
       applicationId = Ref.LedgerString.assertFromString("test"),
       commandId = commandId,
-      deduplicateUntil = recordTime.addMicros(deduplicationTime.toNanos / 1000).toInstant,
+      deduplicationPeriod = DeduplicationPeriod.DeduplicationDuration(deduplicationTime),
+      submissionId = submissionId,
+      ledgerConfiguration = null,
```
fabiotudone-da, line 442: Elsewere we just don't pass unset fields to scalapg's constructors; is there a reason why it's not done here? If not, can you remove it here and elsewhere?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/KVUtilsTransactionSpec.scala
```diff
@@ -182,10 +174,7 @@ class KVUtilsTransactionSpec extends AnyWordSpec with Matchers with Inside {
           resultA.successfulLogEntry.getPayloadCase shouldEqual DamlLogEntry.PayloadCase.TRANSACTION_ENTRY
 
           resultB.successfulLogEntry.getPayloadCase shouldEqual DamlLogEntry.PayloadCase.TRANSACTION_REJECTION_ENTRY
-          resultB.successfulLogEntry.getTransactionRejectionEntry.getReasonCase shouldEqual DamlTransactionRejectionEntry.ReasonCase.INCONSISTENT
-          resultB.successfulLogEntry.getTransactionRejectionEntry.getInconsistent.getDetails should startWith(
-            "InconsistentKeys"
-          )
```
fabiotudone-da, line 188: Are we OK with removing these string matches then?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -93,6 +96,190 @@ class ConversionsSpec extends AnyWordSpec with Matchers {
       val key2 = deduplicationKeyBytesFor(List("alice", "bob"))
       key1 should not be key2
     }
+
+    "encode/decode rejections" should {
+
+      val submitterInfo = DamlSubmitterInfo.newBuilder().build()
+      val now = Instant.now
+
+      "convert rejection to proto models and back to expected grpc code" in {
```
fabiotudone-da, line 105: What is our strategy for checking (driver-specific afair) details?

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/CompletionInfo.scala
```diff
@@ -33,13 +33,16 @@ import com.daml.logging.entries.{LoggingValue, ToLoggingValue}
   *                               [[CompletionInfo.optDeduplicationPeriod]].
   * @param submissionId           An identifier for the submission that allows an application to
   *                               correlate completions to its submissions.
+  *
+  *                               Optional as entries created by the participant.state.v1 API do not have this filled.
+  *                               Only set for participant.state.v2 created entries
```
fabiotudone-da, line 38:
```suggestion
  *                               Only set for entries created by participant.state.v2.
```

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/UpdateNormalizer.scala
miklos-da: Agreed regarding fixing this normalizer in a follow-up PR to limit scope.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
miklos-da: Because we're depending on the deduplication period to generate the proper min/max record times during pre-execution, we need to add the new deduplication period here as well and convert it into a maximum record time in [LedgerTimeValidator](https://github.com/digital-asset/daml/blob/main/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/validation/LedgerTimeValidator.scala#L52).

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/KVTest.scala
nicu-da: Set it to null because this is not scalapb code, but our own model

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/KVUtilsTransactionSpec.scala
nicu-da: Yes because we can assert on the actual type, v2 rejections are more refined type-wise.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
nicu-da: Nothing really being done for now. We only verify status codes and some messages in a few conformance tests, but nothing really extended

ledger/participant-state/kvutils/src/test/lib/scala/com/daml/ledger/participant/state/kvutils/KVTest.scala
fabiotudone-da: Why are we setting nulls in our own models?

nicu-da: Update to use some defaults

fabiotudone-da (approved):

PR #10681, opened 2021-08-25
Add LedgerClientWithoutLedgerId next to the LedgerClient

realvictorprm:
For changes to the json api I'm in need of a ledger client which does not need the ledgerid for initialization. In short the LedgerClientWithoutLedgerId is doing the same as the LedgerClient but one needs to pass for every call the ledger id.

However I'm not sure if it's even necessary to do it like this, because the token passed over should already contain the ledger id.

### Pull Request Checklist

trigger the build.


language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/LedgerClientBinding.scala
```diff
@@ -57,6 +57,7 @@ class LedgerClientBinding(
       templateSelector: TemplateSelector,
       startOffset: LedgerOffset,
       endOffset: Option[LedgerOffset],
+      token: Option[String],
```
stefanobaghino-da, line 60:
```suggestion
      token: Option[String] = None,
```
Should this have a default value to avoid breaking downstream?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
```diff
@@ -138,12 +134,6 @@ object LedgerClient {
       builder: NettyChannelBuilder,
       configuration: LedgerClientConfiguration,
   )(implicit ec: ExecutionContext, esf: ExecutionSequencerFactory): Future[LedgerClient] = {
-    val channel = GrpcChannel(builder, configuration)
-    sys.addShutdownHook {
-      channel.shutdownNow()
-      ()
-    }
-    apply(channel, configuration)
+    apply(GrpcChannel.withShutdownHook(builder, configuration), configuration)
```
stefanobaghino-da, line 137:
```suggestion
    LedgerClient(GrpcChannel.withShutdownHook(builder, configuration), configuration)
```
Not necessary to change this, the argument I would make is that this eliminates any ambiguity for the reader as to which `apply` method is this.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/GrpcChannel.scala
```diff
@@ -44,4 +44,15 @@ object GrpcChannel {
       )
   }
 
+  def withShutdownHook(
```
stefanobaghino-da, line 47: Good call factoring this out of the main logic, makes it more readable. 👍🏻

stefanobaghino-da (commented):
Looks good, thanks. 👍🏻 

This PR has client-facing consequences (the Scala bindings are deprecated but still part of our externally supported components), so please add a couple of lines to the change log detailing what has been added.

language-support/scala/bindings-akka/src/main/scala/com/digitalasset/ledger/client/binding/LedgerClientBinding.scala
realvictorprm: This is a good idea, thanks!

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClient.scala
realvictorprm: I agree, thanks!

realvictorprm:
@stefanobaghino-da addressed the changelog with e937cea. I hope it is sufficient.

realvictorprm:
Because no mails get sent out: Enabled auto-squash-merge

realvictorprm:
Looks like I need to rebase again :smile:

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/LedgerClientWithoutLedgerId.scala
```diff
@@ -0,0 +1,112 @@
...
+import com.daml.ledger.client.services.transactions.TransactionClientWithoutLedgerId
+import com.daml.ledger.client.services.version.VersionClientWithoutLedgerId
+import io.grpc.{Channel, ManagedChannel}
+import io.grpc.netty.NettyChannelBuilder
+
+import java.io.Closeable
+import java.util.concurrent.TimeUnit
+import scala.concurrent.ExecutionContext
+
+class LedgerClientWithoutLedgerId private (
```
S11001001, line 33: Use a package instead of a different name to distinguish from the with-ledger-ID variants. e.g. `com.daml.ledger.client.withoutledgerid`.

S11001001 (commented):
I am not really a fan of these structurally-equal-but-nominally-incompatible newly introduced types.

But if you don't want to undergo the type-unification of `LedgerClient` with `withoutledgerid...LedgerClient` _et al_ in this PR, may I suggest making every newly-introduced type `private[daml]`, which would make it clear that this is not an extension of the public API, and making the question of later refactoring simpler.

realvictorprm:
> I am not really a fan of these structurally-equal-but-nominally-incompatible newly introduced types.
> 
> But if you don't want to undergo the type-unification of `LedgerClient` with `withoutledgerid...LedgerClient` _et al_ in this PR, may I suggest making every newly-introduced type `private[daml]`, which would make it clear that this is not an extension of the public API, and making the question of later refactoring simpler.

I'm not completely happy with this either and aware of the fact that using generics would have made this code less-duplicated, however I did not want to change the api surface further. However, maybe you have other specific ideas how I could have solved this, I'm very happy to hear them :)

It's a good idea to make it `private[daml]`, many thanks for the suggestion. That should give us the freedom to do what we like with it.

realvictorprm:
@S11001001 I made the types private, however that unfortunately caused that there is no inheritance anymore for these. But now they are really private and we can do what we want with them without the need of taking care & changelogs.

akshayshirahatti-da (approved):
overall no blockers , looks good to me thanks 👍

victormueller-da:
side note, the changelog vanished by accident but that is fine because everything I wrote in there didn't apply anymore. So in the end it is intentionally, that the changelog for this PR is empty.

victormueller-da:
> overall no blockers , looks good to me thanks +1

Many thanks, now I can proceed with my other PR.

PR #10687, opened 2021-08-26
LF: V1 Contract ID check in Preprocessor

remyhaemmerle-da:
This PR makes possible to reject V0 contract IDs during preprocessing.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
@@ -35,9 +35,8 @@ final case class EngineConfig(
     stackTraceMode: Boolean = false,
     profileDir: Option[Path] = None,
     contractKeyUniqueness: ContractKeyUniquenessMode = ContractKeyUniquenessMode.On,
-    // TODO: https://github.com/digital-asset/daml/issues/10504
-    //  switch the default to true
-    requireSuffixedGlobalCids: Boolean = false,
+    requireV1ContractId: Boolean = false,
+    requireSuffixedGlobalContractId: Boolean = false,
 ) {
 
   private[lf] def getCompilerConfig: speedy.Compiler.Config =
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
@@ -105,9 +105,24 @@ object Error {
         s"Provided value exceeds maximum nesting level of ${Value.MAXIMUM_NESTING}"
     }
 
-    final case class IllegalContractId(cid: Value.ContractId) extends Error {
+    final case class IllegalContractId(cid: Value.ContractId, reason: IllegalContractId.Reason)
+        extends Error {
       override def message: String =
-        s"""Illegal Contract ID "${cid.coid}""""
+        s"""Illegal Contract ID "${cid.coid}, """" + reason.details
+    }
+
+    object IllegalContractId {
+      sealed abstract class Reason extends Serializable with Product {
+        def details: String
+      }
+      case object V0ContractId extends Reason {
+        def details = "V0 Contract IDs are forbidden"
+        def apply(cid: Value.ContractId.V0): IllegalContractId = IllegalContractId(cid, this)
+      }
+      case object NonSuffixV1ContractId extends Reason {
+        def details = "non-suffixed V1 Contract IDs are forbidden"
+        def apply(cid: Value.ContractId.V1): IllegalContractId = IllegalContractId(cid, this)
+      }
     }
 
     final case class RootNode(nodeId: NodeId, override val message: String) extends Error
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
@@ -100,14 +100,14 @@ object Error {
         override val message: String,
     ) extends Error
 
-    final case class ValueNesting(value: Value[Value.ContractId]) extends Error {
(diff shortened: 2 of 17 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
```diff
@@ -35,9 +35,8 @@ final case class EngineConfig(
     stackTraceMode: Boolean = false,
     profileDir: Option[Path] = None,
     contractKeyUniqueness: ContractKeyUniquenessMode = ContractKeyUniquenessMode.On,
-    // TODO: https://github.com/digital-asset/daml/issues/10504
-    //  switch the default to true
-    requireSuffixedGlobalCids: Boolean = false,
+    requireV1ContractId: Boolean = false,
```
cocreature, line 38: maybe forbidV0 would be more future proof if we ever add v2?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
```diff
@@ -100,14 +100,14 @@ object Error {
...
     ) extends Error
 
-    final case class ValueNesting(value: Value[Value.ContractId]) extends Error {
+    final case class ValueNesting(culprit: Value[Value.ContractId]) extends Error {
       override def message: String =
         s"Provided value exceeds maximum nesting level of ${Value.MAXIMUM_NESTING}"
     }
 
-    final case class NonSuffixedCid(cid: Value.ContractId.V1) extends Error {
+    final case class IllegalContractId(cid: Value.ContractId) extends Error {
```
cocreature, line 108: should we include the reason here and tell users that it failed because of lack of suffix or because they used v0?

cocreature (approved):
Nice, thank you!

daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ledgerinteraction/IdeLedgerClient.scala
```diff
@@ -47,7 +47,11 @@ class IdeLedgerClient(
   def currentSubmission: Option[ScenarioRunner.CurrentSubmission] = _currentSubmission
 
   private[this] val preprocessor =
-    new preprocessing.CommandPreprocessor(compiledPackages.interface, requiredCidSuffix = false)
+    new preprocessing.CommandPreprocessor(
+      compiledPackages.interface,
+      requireV1ContractId = false,
```
cocreature, line 52: Can we not enforce v1 cids in Daml Studio?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
```diff
@@ -35,9 +35,8 @@ final case class EngineConfig(
...
+    requireSuffixedGlobalContractId: Boolean = false,
```
cocreature, line 39: did you drop the todo intentionally?

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
remyhaemmerle-da: :+1:

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
remyhaemmerle-da: Yes. After our discussion last week where you told me you are not concern about misusage of the flags (because we have only 3 actively maintain ledgers) I just thought the default behavior does not really matter.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
remyhaemmerle-da: I did not feel it was necessary, in the sens that if it's happen it means the user try to feed to the ledger ID that it has cooked himself. If the user want to shoot himself in the foot, I do not see why I shoot tell him which foot is it. 
If you really think it is useful, I will add the reason.

daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ledgerinteraction/IdeLedgerClient.scala
remyhaemmerle-da: will try.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
cocreature: fair enough, I tend to lean on the side of including too much information rather than too little in case we need it later but I don’t feel all that strongly about this one.

remyhaemmerle-da: 88c5c2dabe5be062dab899680144636703e9f079

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/EngineConfig.scala
oliverse-da: Ok so in canton, I would rename use use of `requireV1ContractId` to `forbidV0ContractId`.

And no need to use `requireV1ContractIdSuffix`, right, @remyhaemmerle-da ?

cocreature: @oliverse-da you should use both: `forbidV0ContractId` to make sure you never get a global v0 cid. `requireV1ContractIdSuffix` to make sure that all global v1 cids have a suffix.

Issue #10155, opened 2021-06-30
JSON API multi-user performance testing


cocreature:
Currently our JSON API performance tests focus on the performance of individual queries.

We should also test how performance is impacted by having ~300 concurrent users and which parameters in akka-http or in our own code, e.g., database connection pool size can help.

akshayshirahatti-da:
As part of the Multi user performance tests the following changes was carried out

- Addition of a Connection pool, results captured in 
https://docs.google.com/spreadsheets/d/1HNqzOlMJX0X963DSGDZAYmk3JbWJHwvCbC4TtkUP9iQ/edit?usp=sharing
- Introduction of a key hash field, the introduction of which lead to notable performance benefit.
baseline

```

> Number of requests                                  1000 (OK=1000   KO=-     )
> Min. response time                                   487 (OK=487    KO=-     )
> Max. response time                                  4631 (OK=4631   KO=-     )
> Mean response time                                  2462 (OK=2462   KO=-     )
> Std. deviation                                       787 (OK=787    KO=-     )
> response time 90th percentile                       2507 (OK=2507   KO=-     )
> response time 95th percentile                       3002 (OK=3002   KO=-     )
> response time 99th percentile                       3637 (OK=3637   KO=-     )
> response time 99.9th percentile                     4126 (OK=4126   KO=-     )
> Mean requests/second                               3.906 (OK=3.906  KO=-     )
---- Response time distribution ------------------------------------------------
> t < 800 ms                                             7 (  1%)
> 800 ms < t < 1200 ms                                  67 (  7%)
> t > 1200 ms                                          926 ( 93%)
> failed                                                 0 (  -%)
```
query using hashed key field

```
> Number of requests                                  1000 (OK=1000   KO=-     )
> Min. response time                                   206 (OK=206    KO=-     )
> Max. response time                                  2893 (OK=2893   KO=-     )
> Mean response time                                  1074 (OK=1074   KO=-     )
> Std. deviation                                       348 (OK=348    KO=-     )
> response time 90th percentile                       1100 (OK=1100   KO=-     )
> response time 95th percentile                       1330 (OK=1330   KO=-     )
> response time 99th percentile                       1577 (OK=1577   KO=-     )
> response time 99.9th percentile                     1832 (OK=1832   KO=-     )
> Mean requests/second                                8.85 (OK=8.85   KO=-     )
---- Response time distribution ------------------------------------------------
> t < 800 ms                                           217 ( 22%)
[...]

PR #10688, opened 2021-08-26
[JSON-API] vanilla oracle_perf ci job

akshayshirahatti-da:
Addition of a perf oracle job , which runs the multi user scenarios to write data and run fetch by key and payload search idx based queries.  Running higher number of queries post search idx changes by Stephen

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+              docker rm -f oracle
+          }
+          trap cleanup EXIT
+          testConnection() {
+              docker exec oracle bash -c 'sqlplus -L '"$ORACLE_USERNAME"'/'"$ORACLE_PWD"'@//localhost:'"$ORACLE_PORT"'/ORCLPDB1 <<< "select * from dba_users;"; exit $?' >/dev/null
+          }
+
+          # dont want to wait forever to test connection , 15m is more than sufficient here.
+          declare -fxr testConnection
+          timeout 15m bash -c 'until testConnection; do echo "Could not connect to Oracle, trying again..." ; sleep 1 ; done'
```
garyverhaegen-da, line 220: AFAICT `timeout` is not a Bash builtin, so may be worth adding to dev-env to get reproducible behaviour across platforms.

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+
+          # Actually run some tests
+
+          bazel build //ledger-service/http-json-perf/...
+          DAR="${PWD}/bazel-bin/ledger-service/http-json-perf/LargeAcs.dar"
+          JWT="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJodHRwczovL2RhbWwuY29tL2xlZGdlci1hcGkiOnsibGVkZ2VySWQiOiJNeUxlZGdlciIsImFwcGxpY2F0aW9uSWQiOiJmb29iYXIiLCJhY3RBcyI6WyJBbGljZSJdfX0.VdDI96mw5hrfM5ZNxLyetSVwcD7XtLT4dIdHIOa9lcU"
```
garyverhaegen-da, line 226: Would be nice to put (in a comment, perhaps) what that JWT decodes to, to save people a round trip to jwt.io.

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+
+          TEST_CASES="\
+           populateCache \
+           fetchByKey \
+           fetchByQuery \
+          "
+
+          START=$(git log -n1 --format=%cd --date=format:%Y%m%d).$(git rev-list --count HEAD).$(Build.BuildId).$(git log -n1 --format=%h --abbrev=8)
```
garyverhaegen-da, line 234: `START` feels like a bit of a weird name for a variable that does not include the current time.

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+          OUT="$(Build.StagingDirectory)/${REPORT_ID}"
+
+          # Overall we want to populate the test cache , run some fetch by key queries and then some index queries
+
+          export NUM_RECORDS=100000 # 100k ACS for a single template
+          export NUM_QUERIES=10000 # 10k queries in total
+          export NUM_READERS=100 # 100 users in parallel.
+
+          for CASE in $TEST_CASES; do
+            RUN_MODE=${CASE} USE_DEFAULT_USER=true RETAIN_DATA=true bazel run //ledger-service/http-json-perf:http-json-perf-binary-ee -- \
```
garyverhaegen-da, line 245:
```suggestion
            RUN_MODE=${CASE} \
            USE_DEFAULT_USER=true \
            RETAIN_DATA=true \
            bazel run //ledger-service/http-json-perf:http-json-perf-binary-ee -- \
```

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+              --scenario=com.daml.http.perf.scenario.OracleMultiUserQueryScenario \
+              --dars=${DAR} \
+              --reports-dir=${OUT}_${CASE} \
+              --jwt=${JWT} \
+              --query-store-index=oracle
+          done
+
+          GZIP=-9 tar -zcvf ${OUT}.tgz ${OUT}_*
```
garyverhaegen-da, line 253: Do we need to be verbose here?

garyverhaegen-da: Also `${OUT}_*` looks like a list of files. I personally like my tar files to be contained in a folder; could you maybe do `--reports=${OUT}/${CASE}` instead of `${OUT}_${CASE}`?

garyverhaegen-da (approved):
TIL two new Bash features. Thanks!

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,94 @@ jobs:
...
+          set -euo pipefail
+          eval "$(dev-env/bin/dade assist)"
+          source $(bash_lib)
+
+          docker login --username "$DOCKER_LOGIN" --password "$DOCKER_PASSWORD"
+          IMAGE=$(cat ci/oracle_image)
+          docker pull $IMAGE
+          # Cleanup stray containers that might still be running from
+          # another build that didn’t get shut down cleanly.
+          docker rm -f oracle || true
```
stefanobaghino-da, line 204: Is there a chance the `oracle` container is removed while another pipeline is using it?

ledger-service/http-json-perf/src/main/scala/com/daml/http/perf/scenario/OracleMultiUserQueryScenario.scala
```diff
@@ -12,7 +12,7 @@ import scala.concurrent.duration._
 
 private[scenario] trait HasRandomCurrency {
   protected val rng = new scala.util.Random(123456789)
-  final val currencies = List("USD", "GBP", "EUR", "CHF", "AUD")
+  final val currencies = Range.apply(0, 1000).map(i => s"CUR_$i").toList
```
stefanobaghino-da, line 15:
```suggestion
  final val currencies = Range(0, 1000).map(i => s"CUR_$i")
```
The explicit `apply` is not needed and I believe the `toList` is probably not necessary:
```
scala> :t Range(0, 1000).map(i => s"CUR_$i")
IndexedSeq[String]
```

Otherwise, if you want there's ~an app for that~ `List.tabulate` that does this in the standard library:
```suggestion
  final val currencies = List.tabulate(1000)(i => s"CUR_$i")
```

stefanobaghino-da (commented):
Looks good, thanks. I left one minor suggestion on the Scala code and I have one doubt regarding the `oracle` container (maybe @garyverhaegen-da knows about it).

ci/cron/daily-compat.yml
akshayshirahatti-da: renamed to `METADATA`

ledger-service/http-json-perf/src/main/scala/com/daml/http/perf/scenario/OracleMultiUserQueryScenario.scala
akshayshirahatti-da: `List.tabulate` seems like the better choice thanks

ci/cron/daily-compat.yml
akshayshirahatti-da: This logic is similar to what we use for our `Linux_oracle` jobs , as long as we don't run multiple jobs in parallel on the same azure boxes this should be fine IMO

cocreature: We only ever have one job running per node so that’s not an issue.

ci/cron/daily-compat.yml
```diff
@@ -173,6 +173,107 @@ jobs:
...
+          GOOGLE_APPLICATION_CREDENTIALS_CONTENT: $(GOOGLE_APPLICATION_CREDENTIALS_CONTENT)
+      - template: ../bash-lib.yml
+        parameters:
+          var_name: bash_lib
+      - bash: |
+          set -euo pipefail
+          eval "$(dev-env/bin/dade assist)"
+          source $(bash_lib)
+
+          docker login --username "$DOCKER_LOGIN" --password "$DOCKER_PASSWORD"
```
cocreature, line 199: doesn’t have to be in this PR but I think we should factor out the logic here into a bash script so we can share it between this job and the other one.

cocreature (approved):
Nice work!

ci/cron/daily-compat.yml
stefanobaghino-da: Cool, thanks! 🙇🏻

akshayshirahatti-da:
work done as part of [#10155](https://github.com/digital-asset/daml/issues/10155)

Issue #10383, opened 2021-07-22
Support deadline for command submission


dparmar:

dparmar:
This is a feature request to support [deadline](https://grpc.io/blog/deadlines/) parameter for CommandSubmissionClient or any unary RPC for that matter.

dparmar:
Created a PR for the same - https://github.com/digital-asset/daml/pull/10384

realvictorprm:
did #10522 fix this?

dparmar:
> did #10522 fix this?

Yes, it did.

PR #10692, opened 2021-08-27
Fix for #10383

dparmar:
CHANGELOG_BEGIN
Revert "Issue: 10383 - Support deadline for command submission (#10522)"
Support request specific deadline using interceptor.
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
@@ -12,8 +12,10 @@ module DA.Daml.Compiler.Repl
     , ReplLogger(..)
     ) where
 
-import TcRnTypes (tcg_rdr_env)
-import BasicTypes (Boxity(..))
+import FastString
+import TysWiredIn (unitDataCon, unitTyCon)
+import BasicTypes (Boxity(..), PromotionFlag(..), Origin(..))
+import DynFlags
 import Bag (bagToList, unitBag)
 import Control.Applicative
 import Control.Concurrent.Extra
@@ -25,8 +27,13 @@ import qualified Control.Monad.State.Strict as State
 import Control.Monad.Trans.Maybe
 import DA.Daml.Compiler.Output (printDiagnostics)
 import qualified DA.Daml.LF.Ast as LF
+import qualified DA.Daml.LF.InferSerializability as Serializability
+import qualified DA.Daml.LF.Simplifier as LF
+import qualified DA.Daml.LF.TypeChecker as LF
 import DA.Daml.LF.Ast.Optics (packageRefs)
 import qualified DA.Daml.LF.ReplClient as ReplClient
+import DA.Daml.LFConversion (convertModule)
+import DA.Daml.LFConversion.UtilLF (buildPackage)
 import DA.Daml.LFConversion.UtilGHC
 import DA.Daml.Options.Types
 import qualified DA.Daml.Preprocessor.Records as Preprocessor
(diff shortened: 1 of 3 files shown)


language-support/java/bindings-rxjava/src/test/scala/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImplTest.scala
```diff
@@ -36,7 +35,7 @@ class CommandSubmissionClientImplTest
   it should "timeout after deadline exceeded" in {
     ledgerServices.withCommandSubmissionClient(
       Future.never,
-      deadline = Optional.of(Deadline.after(5, TimeUnit.SECONDS)),
+      interceptors = Array(new TimeoutInterceptor(5, TimeUnit.SECONDS)),
     ) { (client, serviceImpl) =>
       val commands = genCommands(List.empty)
       expectDeadlineExceeded(
```
stefanobaghino-da, line 41: There is no test validating that the bug has been fixed.

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/DamlLedgerClient.java
```diff
@@ -68,14 +68,14 @@ public Builder withAccessToken(@NonNull String accessToken) {
       return this;
     }
 
-    public Builder withDeadline(@NonNull Deadline deadline) {
-      this.deadline = Optional.of(deadline);
+    public Builder withInterceptors(ClientInterceptor... interceptors) {
+      this.interceptors = interceptors;
```
stefanobaghino-da, line 72: Two notes:
1. `withDeadline` has been shipped as part of a stable release (1.16) and we can't just drop it
2. In general I think client interceptors are a good idea but I would prefer to have a bit of design work happen before committing to something in that regard

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImpl.java
```diff
@@ -30,11 +30,11 @@ public CommandSubmissionClientImpl(
       @NonNull String ledgerId,
       @NonNull Channel channel,
       Optional<String> accessToken,
-      Optional<Deadline> deadline) {
+      ClientInterceptor[] interceptors) {
     this.ledgerId = ledgerId;
     this.serviceStub =
         StubHelper.authenticating(
-            CommandSubmissionServiceGrpc.newFutureStub(channel).withDeadline(deadline.orElse(null)),
+            CommandSubmissionServiceGrpc.newFutureStub(channel).withInterceptors(interceptors),
```
stefanobaghino-da, line 37: I think the best solution is to:
1. revert to accept an optional deadline
2. store the deadline in the object
3. move the `withDeadline` call to when the call stub is created as part of `submit` (should be around line 65)

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/grpc/TimeoutInterceptor.java
```diff
@@ -0,0 +1,23 @@
...
+    this.duration = duration;
+    this.unit = unit;
+  }
+
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    return next.newCall(method, callOptions.withDeadlineAfter(duration, unit));
+  }
+}
```
stefanobaghino-da, line 23: Under the solution proposed in the previous comment, this class can be removed.

stefanobaghino-da (changes requested):
Thanks for spotting the bug and contributing back, it's very much appreciated!

I left a few comments that I'd like to be addressed. In summary:
1. even though the feature is not working as intended, the bindings should not potentially break existing client code
2. there is a solution that I suggested (keep the deadline and apply it on a per-call basis)
3. allowing interceptors is a good idea but I would prefer to have some time to design it properly
4. most importantly: there should be a test validating that the bug has been fixed (lowering the timeouts from 5 to 1 seconds should allow you to add the relevant test while lowering the amount of time required to run the tests overall)

stefanobaghino-da:
> CHANGELOG_BEGIN
> Revert "Issue: 10383 - Support deadline for command submission (#10522)"
> Support request specific deadline using interceptor.
> CHANGELOG_END

The change log tags should be in the `git` commit history, not in the GitHub PR description. I would recommend to add another tag describing the bug that was fixed explicitly as part of addressing the review comments.

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/DamlLedgerClient.java
dparmar: 1. Looks like `withDeadline` hasn't been shipped in 1.16.
2. Sounds good.

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImpl.java
dparmar: I agree. However since a `Deadline` object represents an absolute deadline (i.e. it starts ticking as soon as it is created), it is not suitable to reuse the same instance in multiple requests. So I used `Duration` and created deadline per request.

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/grpc/TimeoutInterceptor.java
dparmar: Done

language-support/java/bindings-rxjava/src/main/java/com/daml/ledger/rxjava/grpc/CommandSubmissionClientImpl.java
dparmar: I closed the PR and created a new one https://github.com/digital-asset/daml/pull/10713

PR #10697, opened 2021-08-30
DPP-368 unhide append-only CLI flags

rautenrieth-da:
This PR unhides the CLI flags that enable the append-only schema. This effectively publicly releases the append-only schema, and should go together with a release note.

### Changelog entry:

```
- [Sandbox, participant] Added a flag to enable a new append-only database schema.
  This schema was designed to support significantly higher performance.
  In a future release, all applications will automatically migrate to the new schema.
```


mziolekda (approved):
Thank you, looks good!

PR #10701, opened 2021-08-30
Construct ParsedModule directly in Daml Repl

### LINKED ISSUE #10073: REPL needs manual import of DA.TextMap
I've got a template that uses `DA.TextMap`:

```
template Bar
  with
    p : Party
    matrix : [TextMap Foo]
  where
    signatory p
```

When I start the repl with my usual command `daml repl .daml/dist/large-contract-0.0.1.dar -i large-contract --ledger-host localhost --ledger-port 6865`, and try to do anything with `Bar`, I get this error:

```
File:     Line2.daml
Hidden:   no
Range:    6:21-6:43
Source:   typecheck
Severity: DsError
Message:
  Line2.daml:6:21: error:
  Not in scope: type constructor or class ‘DA.Internal.LF.TextMap’
  No module named ‘DA.Internal.LF’ is imported.
```

Adding `DA.TextMap` manually via `:module + DA.TextMap` fixes this, but it's surprising to me that that is necessary, and I don't think the error message would lead many users to that conclusion.

### LINKED ISSUE #10664: Improve error message in Daml Repl if types are not in scope
The way Daml Repl currently works is that if you bind an identifier
```
x <- …
```
following lines are typechecked as a function with `x` being an argument. This requires us to refer to the type of `x` in the type signature.
We try to be somewhat clever here and figure out how we can refer to the type even if the defining module is not in scope but in some cases that’s just impossbible and things fail.
If the defining module is from a loaded package you get a remotely sensible error that the module is not imported. However if the package is not in scope you get a parse error since we end up printing something like `pkg-version:Module.Entity` which is not valid Daml.

Ideally we would get rid of this issue completely but I don’t really know how. Baring that, we should at least produce a better error.

cocreature:
This is a relatively large change unfortunately which unfortunately
requires reimplementing parts of the logic of the typechecker & core
compilation. I don’t think it is too bad but we might want to think
over time if we can factor this better.

This fixes #10073 and fixes #10664 by referencing the exact types
instead of going via the renamer.

There are some minor changes around error messages for "module not
found" errors. This is because these are now caught in the
typechecker instead of in our own code. We could keep the errors but
it requires duplicating even more logic and I don’t really see what it
buys us so I think I prefer the approach here.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/Repl.hs
@@ -12,8 +12,10 @@ module DA.Daml.Compiler.Repl
     , ReplLogger(..)
     ) where
 
-import TcRnTypes (tcg_rdr_env)
-import BasicTypes (Boxity(..))
+import FastString
+import TysWiredIn (unitDataCon, unitTyCon)
+import BasicTypes (Boxity(..), PromotionFlag(..), Origin(..))
+import DynFlags
 import Bag (bagToList, unitBag)
 import Control.Applicative
 import Control.Concurrent.Extra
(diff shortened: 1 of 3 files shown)


sofiafaro-da (approved):
LGTM, thank you! 👍

PR #10703, opened 2021-08-30
LF: Update specification with Contract ID Comparability check

remyhaemmerle-da:
This is part of #10504.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
@@ -96,7 +96,7 @@ private[lf] object Pretty {
         text("Update failed due to a unknown contract") & prettyContractId(cid)
       case NonComparableValues =>
         text("functions are not comparable")
-      case ContractIdFreshness(globalCid) =>
+      case ContractIdComparability(globalCid) =>
         text(s"The global contract ID $globalCid conflicts with a local contract ID")
       case ContractIdInContractKey(key) =>
         text(
--- a/daml-lf/spec/contract-id.rst
+++ b/daml-lf/spec/contract-id.rst
@@ -26,17 +26,18 @@ projection. That is, the contract IDs for contracts created in the
 projection of a transaction to a set of parties can be computed
 solely from the projection and input seeds.
 
-**Unlinkability**: It is computationally infeasible to link the contract
-contents to the contract ID unless the create node is witnessed or the input seeds are known. The
-contract contents include the contract instance, the template ID, the
-stakeholders / signatories / maintainers, and the contract key.
+**Unlinkability**: It is computationally infeasible to link the
+contract contents to the contract ID unless the create node is
+witnessed or the input seeds are known. The contract contents include
+the contract instance, the template ID, the stakeholders / signatories
+/ maintainers, and the contract key.
 
-**Freshness**: It is computationally infeasible to find contracts C1 and
-C2 with the same contract ID such that one of the following holds:
+**Freshness**: It is computationally infeasible to find contracts C1
+and C2 with the same contract ID such that one of the following holds:
 
 * The contracts have different stakeholders.
 * The contracts are created by different transactions.
-* The contracts are created by different subactions of the same
+* The contracts are created by different sub-actions of the same
   transaction forest where the seeds are pairwise different.
 
 **Distinctness**: The ledger can enforce that the seeds are pairwise
@@ -64,8 +65,8 @@ where
   uniqueness of the contract ID in a distributed ledger.
 
 
-Discriminator freshness
------------------------
+Local/Global Contract IDs
+-------------------------
 
 In a transaction we distinguish two kinds of contract IDs:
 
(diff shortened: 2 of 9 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Pretty.scala
```diff
@@ -96,7 +96,7 @@ private[lf] object Pretty {
         text("Update failed due to a unknown contract") & prettyContractId(cid)
       case NonComparableValues =>
         text("functions are not comparable")
-      case ContractIdFreshness(globalCid) =>
+      case ContractIdComparability(globalCid) =>
```
cocreature, line 99: Did this sneak in from #10687?

remyhaemmerle-da: need more coffee.

remyhaemmerle-da: In fact, it is the contrary I sneak in  #10687 this. 

The correct place for this change is this PR, because this is related to V1 contract ID.

daml-lf/spec/contract-id.rst
```diff
@@ -64,37 +65,29 @@ where
...
 * The *global* contract IDs are the contract IDs that:
-   
+
    * appear in the commands that produced the transaction. This
      includes the IDs of the exercised contract, together with all the
      IDs referenced in the arguments of the create and exercise
      commands;
    * that are fetched or looked up by key unless they are local;
-   * are referenced in the input contracts.
+   * are referenced in the payloads of the fetched contracts.
```
cocreature, line 83: nit: arguably input is more correct since an exercise does not have a direct fetch

daml-lf/spec/contract-id.rst
```diff
@@ -116,6 +109,23 @@ discriminator allocation scheme ensures in this case the uniqueness of
...
+-------------------------
+
+The so-called *contract ID comparability restriction*, states that the
+comparison of a local contract ID with a global contract IDs
+that have the same discriminator is forbidden. Engine compliant with
+this specification, should reject with a fatal error any attempt to
+compare during interpretation such IDs, either explicitly (using
+Generic equality and order builtins) or implicitly (though ordering of
+key in generic maps). Ledger implementations that suffix contract IDs
+should furthermore enforce that all global Contract IDs are suffixed.
```
cocreature, line 122:
```suggestion
that have the same discriminator is forbidden. Any attempt to
compare such IDs during interpretation, either explicitly (using
Generic equality and order builtins) or implicitly (though ordering of
key in generic maps), must be rejected. Ledger implementations that suffix contract IDs
should furthermore enforce that all global Contract IDs are suffixed.
```

daml-lf/spec/contract-id.rst
```diff
@@ -116,6 +109,23 @@ discriminator allocation scheme ensures in this case the uniqueness of
 allocated discriminators.
 
 
+Contract ID Comparability
+-------------------------
+
+The so-called *contract ID comparability restriction*, states that the
+comparison of a local contract ID with a global contract IDs
+that have the same discriminator is forbidden. Engine compliant with
```
cocreature, line 117:
```suggestion
The so-called *contract ID comparability restriction*, states that the
comparison of a local contract ID with a global contract ID
with the same discriminator is forbidden. Engine compliant with
```

daml-lf/spec/contract-id.rst
```diff
@@ -232,9 +242,11 @@ The submission performs the following steps:
 
 Depending on the ledger implementation, the local contract IDs are
 suffixed with a suffix in a later step. This yields the *committed
-transaction*. For ledgers that do not require suffixing, committed and submitted
-transactions coincide. Committed transactions are the source of truth to
-derive the state of the ledger.
+transaction*. In the case the ledger require suffixing
```
cocreature, line 245: incomplete sentence?

daml-lf/spec/daml-lf-1.rst
```diff
@@ -3554,6 +3554,15 @@ The following builtin functions defines an order on the so-called
 abstractions, functions, partially applied builtin functions, and
 updates.
 
+Note that as described in the `V1 Contract ID allocation scheme
+specification <./contract-id.rst>`_ the comparison of two V1 contract
+identifiers may fail at run time. For the purpose of this
+specification, we will say that two contract identifiers are *not
+comparable* if (i) both of them are V1 contract identifiers, (ii) one
```
cocreature, line 3561: I’m a bit confused by the condition here, you have i && ii && iii but I would have expected !i || (ii && iii) meaning either one or both is a v0 cid or both are v1 cids and 3 apply.

cocreature (approved):
lgtm with a few comments, thanks!

daml-lf/spec/contract-id.rst
remyhaemmerle-da:
```suggestion
   * are referenced in the input contracts.
```

daml-lf/spec/contract-id.rst
remyhaemmerle-da:
```suggestion
transaction*. 
```

daml-lf/spec/daml-lf-1.rst
remyhaemmerle-da: I am problably way more confused that you ;) .

Note that here I am specifying *not comparable* (maybe it is a bad idea).


Let me redefine the opposite condition:
 
Two cids are comparable if one of the following condition is true: 
- at least one of them is V0: `!(i)`
- both are suffixed : `!(ii)` 
- none of them is a prefix of the other : `!(iii)`

So we have cids are comparable if `!(i)` || `!(ii)` || `!(iii)` or cids are not comparable if  `(i) && (ii) && (iii)`.

remyhaemmerle-da: Does this make sens.

cocreature: Doesn’t make sense to me. Why are v0 cids comparable? I thought comparisons require v1?
or is comparable only equality but not ordering?

cocreature: Recording our conversation: It does make sense, we do allow for comparisons of v0 contract ids since they are always global (or you run in an engine that does not provide comparison primitives).

PR #10704, opened 2021-08-30
participant-integration-api: Use deadlines, not deduplication times, for expiring trackers. [KVL-1009]

SamirTalwar-DA:
I wanted to add a high-level test for this but unfortunately I can't, because gRPC deadlines are handled on both the client and the server. This means that if I set a gRPC deadline on the client, even if the server handles it correctly by aborting the request, I can't tell because the client will *also* handle it and cancel the request before it can be aborted.

I tested it semi-manually by reducing the deadline on the server by 1 second to ensure it does the right thing, but I don't know how to automate this. I think the unit test coverage is good enough though.

### Changelog

- **[Ledger API Server]** The command deduplication time is no longer used for determining the period of time to track the command before giving up. Instead, the gRPC deadline is used. If no deadline is provided (or if the deadline exceeds the command tracker retention period), the  tracker retention period is used instead.

### Pull Request Checklist

trigger the build.


hubert-da (approved):
LGTM

nicu-da (approved):

gerolf-da:
I think there is limited utility for using the gRPC deadline mechanism, if the various bindings (at least java and typescript?) don't support setting this header (cc @stefanobaghino-da)

SamirTalwar-DA:
@gerolf-da: This is a good point. I reinstated the previous behavior for now. The tracker will first check for a deadline, and if this is not set, check for `deduplication_time`.

SamirTalwar-DA:
We would like to remove this behavior because deduplication can now be specified against a specific offset, which cannot necessarily be converted into a timestamp on the participant.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -222,11 +223,28 @@ private[commands] class CommandTracker[Context](
...
+              // We keep supporting the `deduplication_time` field as the command timeout,
+              // for historical reasons.
+              case Commands.DeduplicationPeriod.DeduplicationTime(deduplicationTimeProto) =>
+                val deduplicationTime = Duration.ofSeconds(
+                  deduplicationTimeProto.seconds,
+                  deduplicationTimeProto.nanos.toLong,
+                )
+                durationOrdering.min(deduplicationTime, maximumCommandTimeout)
+              // All other deduplication periods were added recently and should not influence the
+              // command timeout.
```
miklos-da, line 240: Can you remove the time reference from this comment, please? I.e., 'recently' won't mean the same in a few months.
```suggestion
              // All other deduplication period types should not influence the command timeout.
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -222,11 +223,28 @@ private[commands] class CommandTracker[Context](
             s"A command with id $commandId is already being tracked. CommandIds submitted to the CommandTracker must be unique."
           ) with NoStackTrace
         }
-        val commandTimeout =
-          timeoutFromDeduplicationPeriod(commands.deduplicationPeriod, maximumCommandTimeout)
+        val commandTimeout = submission.value.timeout match {
```
miklos-da, line 226: This is really a duration so should be named accordingly (e.g., to avoid confusion with the corresponding parameter in `TrackingData`).

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -342,7 +358,8 @@ class CommandTrackerFlowTest
         val Handle(submissions, results, _, completionStreamMock) =
           runCommandTrackingFlow(allSubmissionsSuccessful)
         val timedOutCommandId = "timedOutCommandId"
-        val submitRequestShortDedupTime = newSubmission(timedOutCommandId, Some(shortDuration))
+        val submitRequestShortDedupTime =
```
miklos-da, line 361: We should not be referencing 'deduplication time' here anymore, right?
```suggestion
        val submitRequestShortTimeout =
```

miklos-da (approved):

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
SamirTalwar-DA: Well, we use "timeout" everywhere else. I'm happy to change it but let's do it in a separate PR, after proper discussion.

I don't think `commandDuration` really captures the meaning. I went for `commandExpiry` before but that's also confusing.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
SamirTalwar-DA: Done. Thanks for catching this.

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
SamirTalwar-DA: I inlined the variable.

PR #10708, opened 2021-08-30
Enable pruning in the sandbox-classic when the append-only schema is used [DPP-567]

kamil-da:
### Main change
- enabled participant pruning in the `sandbox-classic` when the append-only schema is used

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/participant/ParticipantTestContext.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/participant/ParticipantTestContext.scala
@@ -724,7 +724,7 @@ private[testtool] final class ParticipantTestContext private[participant] (
   def prune(
       pruneUpTo: LedgerOffset,
       attempts: Int = 10,
-      pruneAllDivulgedContracts: Boolean = true,
+      pruneAllDivulgedContracts: Boolean = false,
   ): Future[PruneResponse] =
     prune(pruneUpTo.getAbsolute, attempts, pruneAllDivulgedContracts)
 
--- a/ledger/sandbox-classic/BUILD.bazel
+++ b/ledger/sandbox-classic/BUILD.bazel
@@ -372,7 +372,7 @@ server_conformance_test(
         "default",
     ],
     server_args = [
-        "--static-time",
+        "--wall-clock-time",
         "--contract-id-seeding=testing-weak",
         "--enable-append-only-schema",
     ],
--- a/ledger/sandbox-classic/BUILD.bazel
+++ b/ledger/sandbox-classic/BUILD.bazel
@@ -380,6 +380,8 @@ server_conformance_test(
     test_tool_args = [
         "--open-world",
         "--include=ParticipantPruningIT",
+        # Excluding tests that require using pruneAllDivulgedContracts option that is not supported by sandbox-classic
+        "--exclude=ParticipantPruningIT:PRLocalAndNonLocalRetroactiveDivulgences,ParticipantPruningIT:PRRetroactiveDivulgences,ParticipantPruningIT:PRDisclosureAndRetroactiveDivulgence",
     ],
 )
 
--- a/ledger/sandbox-classic/BUILD.bazel
+++ b/ledger/sandbox-classic/BUILD.bazel
@@ -367,6 +367,23 @@ server_conformance_test(
 )
 
 server_conformance_test(
+    name = "conformance-test-pruning-append-only",
+    lf_versions = [
+        "default",
+    ],
+    server_args = [
+        "--static-time",
+        "--contract-id-seeding=testing-weak",
+        "--enable-append-only-schema",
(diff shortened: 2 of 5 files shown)


ledger/sandbox-classic/BUILD.bazel
```diff
@@ -366,6 +366,23 @@ server_conformance_test(
     ],
 )
 
+server_conformance_test(
+    name = "conformance-test-pruning-append-only",
+    lf_versions = [
+        "default",
+    ],
+    server_args = [
+        "--static-time",
```
mziolekda, line 375: static-time and weak seeding are corner cases that we should try and avoid in a more general test.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/LedgerBackedWriteService.scala
```diff
@@ -92,13 +96,15 @@ private[stores] final class LedgerBackedWriteService(ledger: Ledger, timeProvide
...
 
-  // WriteParticipantPruningService - not supported by sandbox-classic
   override def prune(
       pruneUpToInclusive: Offset,
       submissionId: Ref.SubmissionId,
       pruneAllDivulgedContracts: Boolean,
   ): CompletionStage[state.PruningResult] =
-    CompletableFuture.completedFuture(state.PruningResult.NotPruned(Status.UNIMPLEMENTED))
+    CompletableFuture.completedFuture {
+      if (enablePruning) state.PruningResult.ParticipantPruned
```
nmarton-da, line 105: this is not enabling pruning, just pretending to prune, no?
how is the conformance tests pass then?
am I missing something? :thinking:

kamil-da: If I understand correctly the `ApiParticipantPruningService` performs pruning in two steps:
1. pruning the write service
2. pruning the index db
For the `sandbox-classic` they are the same thing, so returning `ParticipantPruned` for the step 1. enables the step two.

nmarton-da: yupp, makes sense, a comment would be nice to clarify this behavior in the write service

tudor-da: `pruneAllDivulgedContracts` is not meant to be supported for `sandbox-classic` because the backfiling lookup queries can start failing if used. Please return `UNIMPLEMENTED` if that flag is set

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/SandboxIndexAndWriteService.scala
```diff
@@ -117,19 +124,25 @@ private[sandbox] object SandboxIndexAndWriteService {
         templateStore,
         ledgerEntries,
       )
-    owner(MeteredLedger(ledger, metrics), participantId, timeProvider)
+    owner(
+      ledger = MeteredLedger(ledger, metrics),
+      participantId = participantId,
+      timeProvider = timeProvider,
+      enablePruning = false,
```
tudor-da, line 131: Why can't we enable pruning for in-memory ledger?

tudor-da: Maybe you can document it in the PR header why this one is left out

mziolekda: It is a corer case that we should not spend any time debating at this point. It would have to go through another story if we wanted that support.

tudor-da: fine by me. Just observed that there is no mention in the ticket or the PR regarding it.

ledger/sandbox-classic/BUILD.bazel
kamil-da: Done.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/participant/ParticipantTestContext.scala
```diff
@@ -724,7 +724,7 @@ private[testtool] final class ParticipantTestContext private[participant] (
   def prune(
       pruneUpTo: LedgerOffset,
       attempts: Int = 10,
-      pruneAllDivulgedContracts: Boolean = true,
+      pruneAllDivulgedContracts: Boolean = false,
```
kamil-da, line 727: There are only three test cases in `ParticipantPruningIT` that specifically test pruning divulged contracts. For these cases we enable `pruneAllDivulgedContracts` explicitly, leaving the flag disabled for all other cases.

tudor-da (approved):

PR #10709, opened 2021-08-30
Support TLS in Daml Helper --json-api requests

cocreature:
This PR accepts the --tls flag but rejects the flags to set certs
because those don’t really make sense here (or are at least not easily
settable).

Unfortunately no great way to test this without a huge amount of test
infrastructure to setup a custom TLS reverse proxy for tests.

If someone has an idea to add a reasonable test, I’m all ears. I did test it manually against daml hub which worked fine.

We should probably also add an option to specify a URL because this is bit of a crappy UX for an http API but one step at a time.

changelog_begin

- [Daml Assistant] The `daml ledger` commands now accepts `--tls` in
  combination with `--json-api` to access a JSON API behind a TLS
  reverse proxy.

changelog_end

### Pull Request Checklist

trigger the build.


robin-da (approved):
thanks!

PR #10710, opened 2021-08-30
DPP-368 enable append-only flag in sandbox

rautenrieth-da:
This PR makes sure the `--enable-append-only-schema` flag is respected in sandbox. Sandbox classic and daml-on-sql already respected the flag.


mziolekda (approved):
LGTM

ledger/sandbox/src/main/scala/platform/sandboxnext/Runner.scala
```diff
@@ -206,6 +206,8 @@ class Runner(config: SandboxConfig) extends ResourceOwner[Port] {
                     else IndexerStartupMode.MigrateAndStart,
                   eventsPageSize = config.eventsPageSize,
                   allowExistingSchema = true,
+                  enableAppendOnlySchema = config.enableAppendOnlySchema,
+                  enableCompression = config.enableCompression,
```
SamirTalwar-DA, line 210: See, this is why I don't like default parameters.

Can we remove the default values from these and force callers to make a choice?

PR #10715, opened 2021-08-31
Use soft references for values in the caches [DPP-561]

gerolf-da:
Soft references allow the garbage collector to collect the values referenced
by soft references in case of memory pressure. Since these caches are not the
source of truth for the data, it's better to remove values from the cache
instead of having the process die due to OOM (not enough heap or excessive GC).

Fixes DPP-561.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/caching/src/main/scala/com/daml/caching/SizedCache.scala
+++ b/ledger/caching/src/main/scala/com/daml/caching/SizedCache.scala
@@ -29,6 +29,7 @@ object SizedCache {
       case Configuration(maximumSize) =>
         val builder = caffeine.Caffeine
           .newBuilder()
+          .softValues()
           .maximumSize(maximumSize)
         CaffeineCache[Key, Value](builder, metrics)
     }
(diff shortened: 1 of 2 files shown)


nmarton-da (approved):
Looks sensible.
One thing what makes me wonder, is that documentation sounds like either softValues or maximumSize should be set. (although it looks a safe combination)
It would be nice to test this guy in similar situations which OOM-ed before to have clarity.

nmarton-da:
One more comment: this changes all the clients of the SizedCache. I think it should be alright though.

kamil-da:
@nmarton-da Actually the docs specifies that you can safely combine both `softValues()` and `maximumSize()` so we should be ok here.

gerolf-da:
I don't think only setting `maximumSize` (i.e. number of elements) is a safe thing to do in our case, because the size of the values can vary drastically. One could argue that with `softValues` one doesn't need `maximumSize` at all, but keeping the participant under constant memory pressure is also not a desirable situation (even though one could come up with pathological examples to achieve that with both `maximumSize` and `softValues`).

SamirTalwar-DA:
> the size of the values can vary drastically

If the size can vary drastically, it is probably safer to use a `WeightedCache` instead. I can see that we use `SizedCache` for the LF value translation cache, the contract key state cache, and the contracts state cache.

The first two will always have pretty small values (contract IDs, mostly), but the contracts state cache could probably perform better and avoid exceeding its size if we use `WeightedCache` instead.

PR #10718, opened 2021-08-31
sandbox-classic: Remove default parameters in `SqlLedger.Owner`.

SamirTalwar-DA:
The defaults can be dangerous.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/LedgerResource.scala
+++ b/ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/LedgerResource.scala
@@ -93,6 +93,9 @@ private[sandbox] object LedgerResource {
           metrics = new Metrics(metrics),
           lfValueTranslationCache = LfValueTranslationCache.Cache.none,
           engine = new Engine(),
+          validatePartyAllocation = false,
+          enableAppendOnlySchema = true,
+          enableCompression = false,
         )
       } yield ledger
     )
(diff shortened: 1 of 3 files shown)


kamil-da (approved):

ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/LedgerResource.scala
```diff
@@ -93,6 +93,9 @@ private[sandbox] object LedgerResource {
           metrics = new Metrics(metrics),
           lfValueTranslationCache = LfValueTranslationCache.Cache.none,
           engine = new Engine(),
+          validatePartyAllocation = false,
+          enableAppendOnlySchema = true,
```
rautenrieth-da, line 97: This is changing the behavior, the test will be running append-only schema now. Was this intended?

rautenrieth-da: Changing all tests to use the append-only schema is something that we should do now, but I'd rather have this in a separate PR.

rautenrieth-da (changes requested):

ledger/sandbox-classic/src/test/lib/scala/platform/sandbox/LedgerResource.scala
kamil-da: Good point @rautenrieth-da. 
I thought it's intended to change to the append-only but I agree that it'd be better to do this in a separate PR.

SamirTalwar-DA: Oh, good point. Thanks!

SamirTalwar-DA: Reverted to `false`.

rautenrieth-da (approved):

PR #10722, opened 2021-08-31
sandbox-classic: Reintroduce SqlLedger tests for the mutable index.

SamirTalwar-DA:
I previously changed `SqlLedgerSpec` to use the append-only index, not realizing that we have both test classes. This removes the (now-redundant) `SqlLedgerSpecAppendOnly`, and copies `SqlLedgerSpec` to `SqlLedgerOnMutableIndexSpec`. The only differences are:

- the mutable index tests check for synchronous commits, and
- the append-only index tests check for rejection reason details (which are not present in the mutable index schema).

Previously I had added some tests to `SqlLedgerSpec` but not `SqlLedgerSpecAppendOnly`, so they had diverged a little.

I also think this makes it clearer that where the tests diverge, we want to keep the append-only tests, not the mutable ones.

### Pull Request Checklist

trigger the build.


ledger/sandbox-classic/src/test/resources/logback-test.xml
```diff
@@ -13,28 +13,16 @@
         <appender-ref ref="SqlLedgerSpecAppender" />
     </logger>
 
-    <logger name="com.daml.platform.store.dao.HikariConnection" level="INFO">
```
mziolekda, line 16: Why are we removing this and the following definitions for `SqlLedgerSpecAppender`?

SamirTalwar-DA: Because the test case no longer makes sense for `SqlLedgerSpec`. We do need to verify this but I have no idea how we should go about it, as we don't log anything in `PostgresStorageBackend`, and even if we did, it's only triggered on subscriptions, which is not what we're doing here.

I would hope that as `PostgresStorageBackend` is new code, it's well-tested.

mziolekda (approved):
Looks good

ledger/sandbox-classic/src/test/resources/logback-test.xml
mziolekda: I see it now, thanks, makes sense

PR #10724, opened 2021-08-31
LF: Refactor engine test reinterpret

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
@@ -59,17 +59,6 @@ class EngineTest
 
   import EngineTest._
 
-  private def hash(s: String) = crypto.Hash.hashPrivateKey(s)
-  private def participant = Ref.ParticipantId.assertFromString("participant")
-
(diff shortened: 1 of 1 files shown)


daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -2773,6 +2772,41 @@ object EngineTest {
   private def suffix(tx: Tx.Transaction) =
     data.assertRight(tx.suffixCid(_ => dummySuffix))
 
+  private[this] case class ReinterpretState(
```
cocreature, line 2775: I somewhat wonder if we can and should use the scenario ledger here. This seems like we’re partially reimplementing what that does.
Let’s leave it out of this PR though.

cocreature (approved):
Thanks!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
remyhaemmerle-da: Good point.

PR #10725, opened 2021-08-31
LF: Drop outdated TODOs

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
@@ -151,13 +151,8 @@ object Error {
         with InternalError
 
     final case class DamlException(error: interpretation.Error) extends Error {
-      // TODO https://github.com/digital-asset/daml/issues/9974
-      //  For now we try to preserve the exact same message (for the ledger API)
-      //  Review once all the errors are properly structured
       override def message: String = error match {
         case interpretation.Error.ContractNotFound(cid) =>
-          // TODO https://github.com/digital-asset/daml/issues/9974
-          //   we should probably use ${cid.coid} instead of $cid
           s"Contract could not be found with id $cid"
         case interpretation.Error.ContractKeyNotFound(key) =>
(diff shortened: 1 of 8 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Error.scala
```diff
@@ -151,13 +151,8 @@ object Error {
         with InternalError
 
     final case class DamlException(error: interpretation.Error) extends Error {
-      // TODO https://github.com/digital-asset/daml/issues/9974
-      //  For now we try to preserve the exact same message (for the ledger API)
-      //  Review once all the errors are properly structured
```
remyhaemmerle-da, line 156: KV Team will tackle that in Q4.

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -411,9 +411,6 @@ server_conformance_test(
 
 server_conformance_test(
     name = "conformance-test-exceptions",
```
cocreature, line 413: I think we can just drop the whole suite? Exceptions are enabled by default.

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -427,9 +424,6 @@ server_conformance_test(
 
 server_conformance_test(
     name = "conformance-test-exceptions-append-only",
-    # TODO https://github.com/digital-asset/daml/issues/9914
```
cocreature, line 430: same here

cocreature (approved):
lgtm with a minor comment on the exception tests

ledger/sandbox-classic/BUILD.bazel
remyhaemmerle-da: Good catch.

remyhaemmerle-da: fb4fc905d976543c3b09d5599ae47000c5bb4b12

PR #10706, opened 2021-08-30
[DPP-438] Change open-ended metric names into static ones (by removing partyName part)

pbatko-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Details

Sample metric name before and after this change (obtained by running conformance test tool against ledger-on-memory locally)

BEFORE:
```
daml_commands_AcceptChoiceArgumentInCreateAndExerciseCommand100_alpha_1352983bce9f_party_0_input_buffer_delay
```

AFTER:
```
daml_commands_input_buffer_delay
```

Testing setup details:
```
Build ledger-on-memory
bazel build //ledger/ledger-on-memory:app_deploy.jar

Run ledger-on-memory with Prometheus server built-in:
java -Xmx4g -jar bazel-bin/ledger/ledger-on-memory/app_deploy.jar --participant participant-id=dummy-participant,port=6865 --daml-lf-dev-mode-unsafe --metrics-reporter prometheus://localhost:9494

Configure Prometheus UI (prometheus.yml)
scrape_configs:
  - job_name: "participant-server-local"
    static_configs:
      - targets: ["localhost:9494"]

Run Prometheus UI
./prometheus --config.file=prometheus.yml

Build conformance test tool:
bazel build //ledger/ledger-api-test-tool:ledger-api-test-tool_deploy.jar

Run conformance test tool:
java -jar bazel-bin/ledger/ledger-api-test-tool/ledger-api-test-tool_deploy.jar localhost:6865
```

### Notes from investigating potentially open-ended named metrics:

1. Party names' metrics
```
com.daml.metrics.Metrics.daml.commands#inputBufferLength
def inputBufferLength(firstParty: String): Counter =
  registry.counter(Prefix :+ firstParty :+ "input_buffer_length")
def inputBufferCapacity(firstParty: String): Counter =
  registry.counter(Prefix :+ firstParty :+ "input_buffer_capacity")
def inputBufferDelay(firstParty: String): Timer =
  registry.timer(Prefix :+ firstParty :+ "input_buffer_delay")
def maxInFlightLength(firstParty: String): Counter =
  registry.counter(Prefix :+ firstParty :+ "max_in_flight_length")
def maxInFlightCapacity(firstParty: String): Counter =
  registry.counter(Prefix :+ firstParty :+ "max_in_flight_capacity")
```

2. Committer names' metrics
```
com.daml.metrics.Metrics.daml.kvutils.committer#runTimer
def runTimer(committerName: String): Timer =
[...]

PR #10728, opened 2021-08-31
[DPP-438] Update docs on metrics that no longer use <party_name> in their name

pbatko-da:
This is a follow-up to https://github.com/digital-asset/daml/pull/10706

CHANGELOG_BEGIN
CHANGELOG_END


mziolekda (approved):
Thanks Pawel, well spotted!

PR #10733, opened 2021-09-01
ledger-api-common: Do not mock final classes.

SamirTalwar-DA:
This replaces `URL` with a trait so we can fake it. This means we don't have to use Mockito to stub out `URL`, which is final and therefore requires magic to stub.

This PR brought to you by weird errors in CI along these lines:

```
java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
at org.mockito.internal.configuration.plugins.PluginLoader$1.invoke(PluginLoader.java:84)
...
```

Rather than diagnosing the error, let's just remove the source.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
@@ -19,7 +19,7 @@ object SecretsUrl {
 
   def fromUrl(url: URL): SecretsUrl = new FromUrl(url)
 
-  private final class FromUrl(url: URL) extends SecretsUrl {
+  private final case class FromUrl(url: URL) extends SecretsUrl {
     override def openStream(): InputStream = url.openStream()
   }
 }
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
@@ -13,13 +13,17 @@ trait SecretsUrl {
 }
 
 object SecretsUrl {
-  def FromString(string: String): SecretsUrl = FromUrl(new URL(string))
+  def fromString(string: String): SecretsUrl = new FromUrl(new URL(string))
 
-  final case class FromUrl(url: URL) extends SecretsUrl {
+  def fromUrl(url: URL): SecretsUrl = new FromUrl(url)
+
+  def fromPath(path: Path): SecretsUrl = new FromPath(path)
+
+  private final class FromUrl(url: URL) extends SecretsUrl {
     override def openStream(): InputStream = url.openStream()
   }
 
-  final case class FromPath(path: Path) extends SecretsUrl {
+  private final class FromPath(path: Path) extends SecretsUrl {
     override def openStream(): InputStream = Files.newInputStream(path)
   }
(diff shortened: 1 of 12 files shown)


ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.net.URL
+import java.nio.file.{Files, Path}
+
+// This trait is not sealed so we can replace it with a fake in tests.
+trait SecretsUrl {
+  def openStream(): InputStream
+}
+
+object SecretsUrl {
+  def FromString(string: String): SecretsUrl = FromUrl(new URL(string))
```
miklos-da, line 16: If this is a method, it should be named like:
```suggestion
  def fromString(string: String): SecretsUrl = FromUrl(new URL(string))
```

miklos-da (approved):

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala
SamirTalwar-DA: I'm not convinced; I know it's a method but I want it to look like any other constructor.

If I change this one I'll also make the subclasses private and add wrappers along the lines of `def fromUrl(url: Url) = FromUrl(url)` so they're lowercase too.

kamil-da: I agree with @miklos-da - to me `SecretsUrl.FromString(url)` looks like creation of a class named `FromString`, especially that we have other classes under `SecretsUrl`.
I would either changed the name to `fromString()` or add a constructor in the companion object:
```
  object FromUrl {
    def apply(string: String): FromUrl = FromUrl(new URL(string))
  }
  
  ```

**Review by kamil-da — APPROVED** [review]
Looks good! Only a minor comment left.

File: ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala (hunk 1 above)
Comment (SamirTalwar-DA) [reply ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala]: Made them all methods.

**Comment by pbatko-da** [comment]
Please take a look at my possible fix: https://github.com/digital-asset/daml/pull/10738

File: ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/SecretsUrl.scala [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.ledger.api.tls
+
+import java.io.InputStream
+import java.net.URL
+import java.nio.file.Path
+
+// This trait is not sealed so we can replace it with a fake in tests.
+trait SecretsUrl {
```
fabiotudone-da, line 11: What's about secrets in this class?

SamirTalwar-DA: Nothing, it's just how it's used.

Happy to bikeshed further on this but in a separate changeset, as this is causing failing builds in CI.

PR #10738, opened 2021-09-01
[DPP-578] Fix test flakiness related by adding byte-buddy-agent at jvm startup.

pbatko-da:
..as opposed to relying on dynamic agent attachment which could sometimes fail like this:

```
2021-08-31T09:46:28.7084817Z - should attempt to decrypt private key using by fetching decryption params from an url *** FAILED *** (5 seconds, 452 milliseconds)
2021-08-31T09:46:28.7086294Z   java.lang.IllegalStateException: Could not initialize plugin: interface org.mockito.plugins.MockMaker (alternate: null)
(...)
2021-08-31T09:46:28.8029094Z   Cause: org.mockito.exceptions.base.MockitoInitializationException: Could not initialize inline Byte Buddy mock maker.
2021-08-31T09:46:28.8029669Z 
2021-08-31T09:46:28.8030410Z It appears as if your JDK does not supply a working agent attachment mechanism.
2021-08-31T09:46:28.8031003Z Java               : 1.8
2021-08-31T09:46:28.8031538Z JVM vendor name    : Oracle Corporation
2021-08-31T09:46:28.8032800Z JVM vendor version : 25.272-b10
2021-08-31T09:46:28.8033588Z JVM name           : OpenJDK 64-Bit Server VM
2021-08-31T09:46:28.8034489Z JVM version        : 1.8.0_272-b10
2021-08-31T09:46:28.8035087Z JVM info           : mixed mode
2021-08-31T09:46:28.8035677Z OS name            : Linux
2021-08-31T09:46:28.8036452Z OS version         : 5.11.0-1017-gcp
```

### DIFF
--- a/bazel-java-deps.bzl
+++ b/bazel-java-deps.bzl
@@ -160,6 +160,10 @@ def install_java_deps():
             "org.junit.platform:junit-platform-engine:1.0.0",
             "org.junit.platform:junit-platform-runner:1.0.0",
             "org.mockito:mockito-core:3.6.28",
+            # NOTE: Keep byte-buddy versions in sync with Mockito versions.
+            # We name it explicitly because wa want to sue byte-buddy-agent to work around https://github.com/mockito/mockito/issues/1879
+            "net.bytebuddy:byte-buddy:1.10.8",
+            "net.bytebuddy:byte-buddy-agent:1.10.8",
             "org.mockito:mockito-inline:3.6.28",
             "org.mockito:mockito-scala_{}:1.16.3".format(scala_major_version),
             "org.pcollections:pcollections:2.1.3",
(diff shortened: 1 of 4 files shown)


SamirTalwar-DA:
This is a reasonable solution if we _had_ to mock the `URL` type, but we really don't.

The creators of mocking, as a concept, have long said ["don't mock types you don't own"](https://www.markhneedham.com/blog/2009/12/13/tdd-only-mock-types-you-own/). I highly recommend following this advice; instead of mocking a `final` class, create an abstraction layer.

kamil-da (commented):
I agree with @SamirTalwar - even though the proposed solution looks ok, I would prefer to get rid of the problematic mocking completely.

PR #10739, opened 2021-09-01
[DRAFT] Pipeline wait time

aherrmann-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/build-unix.yml
+++ b/ci/build-unix.yml
@@ -24,6 +24,7 @@ steps:
   - bash: |
       ci/configure-bazel.sh
       env | sort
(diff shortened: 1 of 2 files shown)


aherrmann-da:
The goal was to log the time between the pipeline being queued by a developer opening a PR or otherwise requesting a run and the time a machine becomes available and the pipeline runs.
Azure pipelines does not provide the required variables by itself. To an extend we can take the git committer time of the merged commit as a proxy. Unfortunately, this does not work for re-runs as those use the same commit but will have a potentially much later pipeline start time, hence, yielding to misleading and potentially very large pipeline wait times.

PR #10743, opened 2021-09-01
Command dedup: In the committer set the deduplicate until context value based on the given deduplication period [KVL-1049]

nicu-da:
### Pull Request Checklist

Calculate deduplicate until for committer side deduplication based on the submitter given deduplication period
- if using the deprecated `deduplicateUntil` then just set the given timestamp
- if using duration then calculate the new `deduplicateUntil` by using the formula (`submissionTime` + `deduplicationDuration` + `minSkew`)
- if using offset deduplication then calculate `deduplicateUntil` by using the formula (`submissionTime` + `maxDeduplicationDuration` + `minSkew`)
- if the deduplication period is not set then we don't set `deduplicateUntil`

Update the conformance tests to so that they update the time model before running to account for the usage of `minSkew` when calculating `deduplicateUntil`

Emit completions for `DuplicateCommand` rejections, if not then we would never get a completion for the submitter command.

Tried to disable participant deduplication but there are still 2 failures on the conformance tests:

```
CommandDeduplicationIT
- [CommandDeduplicationIT:CDDeduplicateSubmitterBasic] Commands with identical submitter and command identifier should be deduplicated by the submission client ... Unexpected failure (ExpectedFailureException)
  Expected a failure when submitting a request as Alice for the second time, but got a successful result of: ()
- [CommandDeduplicationIT:CDSimpleDeduplicationBasic] Deduplicate commands within the deduplication time window ... Unexpected failure (ExpectedFailureException)
  Expected a failure when submitting the first request for the second time, but got a successful result of: ()
```

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
@@ -3,23 +3,29 @@
 
 package com.daml.ledger.api.testtool.suites
 
+import java.time.Instant
+
 import com.daml.ledger.api.testtool.infrastructure.Allocation._
 import com.daml.ledger.api.testtool.infrastructure.Assertions._
 import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
 import com.daml.ledger.api.testtool.infrastructure.ProtobufConverters._
+import com.daml.ledger.api.testtool.infrastructure.participant.ParticipantTestContext
+import com.daml.ledger.api.v1.admin.config_management_service.TimeModel
 import com.daml.ledger.api.v1.commands.Commands.DeduplicationPeriod
 import com.daml.ledger.test.model.DA.Types.Tuple2
 import com.daml.ledger.test.model.Test.TextKeyOperations._
 import com.daml.ledger.test.model.Test._
 import com.daml.timer.Delayed
 import io.grpc.Status
+import org.slf4j.LoggerFactory
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.{DurationInt, FiniteDuration}
 import scala.util.{Failure, Success}
 
 final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterval: FiniteDuration)
     extends LedgerTestSuite {
+  private[this] val logger = LoggerFactory.getLogger(getClass.getName)
   private val deduplicationTime = 3.seconds * timeoutScaleFactor match {
     case duration: FiniteDuration => duration
     case _ =>
@@ -31,6 +37,7 @@ final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterva
     "CDSimpleDeduplicationBasic",
     "Deduplicate commands within the deduplication time window",
     allocate(SingleParty),
+    runConcurrently = false,
   )(implicit ec => { case Participants(Participant(ledger, party)) =>
     lazy val requestA1 = ledger
       .submitRequest(party, DummyWithAnnotation(party, "First submission").create.command)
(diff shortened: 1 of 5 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
```diff
@@ -31,6 +37,7 @@ final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterva
     "CDSimpleDeduplicationBasic",
     "Deduplicate commands within the deduplication time window",
     allocate(SingleParty),
+    runConcurrently = false,
```
fabiotudone-da, line 40: Why is this needed, to serialize on the time model update? How much slower does this make the build?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// Copyright
```
fabiotudone-da, line 1: This header seems broken

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -303,31 +303,41 @@ private[kvutils] class TransactionCommitter(
...
-            .add(config.timeModel.minSkew)
+        commandDedupBuilder.setDeduplicatedUntil(transactionEntry.submitterInfo.getDeduplicateUntil)
+      case DamlSubmitterInfo.DeduplicationPeriodCase.DEDUPLICATION_DURATION =>
+        commandDedupBuilder.setDeduplicatedUntil(
+          Conversions.buildTimestamp(
+            transactionEntry.submissionTime
+              .add(
+                Conversions.parseDuration(transactionEntry.submitterInfo.getDeduplicationDuration)
+              )
+              .add(config.timeModel.minSkew)
```
fabiotudone-da, line 322: Why is the minimum skew used for these calculations? I would expect the maximum.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -207,6 +217,91 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
     }
+
+    "setting dedup context" should {
+      val deduplicateUntil = protobuf.Timestamp.newBuilder().setSeconds(30).build()
+      val submissionTime = protobuf.Timestamp.newBuilder().setSeconds(60)
+
+      def buildContextAndTransaction(
+          submitterInfoAugmenter: DamlSubmitterInfo.Builder => DamlSubmitterInfo.Builder
+      ) = {
+
```
fabiotudone-da, line 228:
```suggestion
```

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -207,6 +217,91 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
         }
       }
     }
+
+    "setting dedup context" should {
+      val deduplicateUntil = protobuf.Timestamp.newBuilder().setSeconds(30).build()
+      val submissionTime = protobuf.Timestamp.newBuilder().setSeconds(60)
+
+      def buildContextAndTransaction(
```
fabiotudone-da, line 225: I think having this functions defined here reduces readability; can you move them at the class or companion level?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -207,6 +217,91 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+      def contextSetDeduplicateUntil(
+          context: CommitContext,
+          transactionEntrySummary: DamlTransactionEntrySummary,
+      ) = {
+        context
+          .get(Conversions.commandDedupKey(transactionEntrySummary.submitterInfo))
+          .value
+          .getCommandDedup
+          .getDeduplicatedUntil
+      }
```
fabiotudone-da, line 254: Braces not needed

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
```diff
@@ -207,6 +217,91 @@ class TransactionCommitterSpec extends AnyWordSpec with Matchers with MockitoSug
...
+                  .newBuilder()
+              )
+            )
+            .setSubmissionTime(submissionTime)
+            .build()
+        )
+        context -> transactionEntrySummary
+      }
+
+      def contextSetDeduplicateUntil(
```
fabiotudone-da, line 245:
```suggestion
      def deduplicateUntil(
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
nicu-da: Yes, because we change the time model.
Not sure how much time it will take but we already have other tests doing this and I don't really see any other way around it.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
nicu-da: Per design doc, I assume it;s the min skew we want to account for to be withing boundries

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
nicu-da: While I do agree it makes things harder to read I'm not sure if moving the at the class level would improve anything.
The added benefit of having them here is that we have them scoped to the tests where they're used. So it gives a form of context, while having everything at the class/object level just makes me jump around there to read any definition used in the test, and with time it also starts getting quite cluttered, having tens of variables and functions which make no sense.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
nicu-da: Can we add this to scalafmt? I have no preference in the `with brace`/`without braces` skirmish but I would very much prefer the formatter to handle code styling like this and not have to do it manually.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
nicu-da: I would keep the set in the name because this actually modifies the state of the context so it's a mutable function

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
fabiotudone-da: (It's really a nit so not very important) Idea unfortunately adds those automatically when extracting a method; I think some form of control is available in `scalafmt` according to the [docs](https://scalameta.org/scalafmt/docs/configuration.html), probably around redundant braces.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
fabiotudone-da: Is the mutation of `Context.get` (about accessed inputs) important in this context though? It looks to me like you just want the resulting dedup until, or am I wrong? If that's the case, having a `set` in the name is confusing IMO.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
fabiotudone-da: Scala lets you do things in many different ways and there are many opinions about everything. Mine is that I usually try to understand top-down, i.e. from high-level to details, so anything in the way distracts me and I prefer to have good names than the definition at hand. I agree it's opinable, so this too is just a nit.

fabiotudone-da (approved):

SamirTalwar-DA (approved):

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
nicu-da: Agree it's nicer to have the tests clean to read. Moved the functions to the bottom and kept only the variables which are used by this tests.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitterSpec.scala
nicu-da: Ah yes, I had the thought that this is the one that sets the value in the context. Renamed

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -490,7 +490,12 @@ private[state] object Conversions {
           )
         )
       case DamlTransactionRejectionEntry.ReasonCase.DUPLICATE_COMMAND =>
-        None // No rejection for duplicate commands.
+        Some(
```
miklos-da, line 493: Now you should be able to move `Some(...)` out from all `match` branches.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -212,6 +212,12 @@ class ConversionsSpec extends AnyWordSpec with Matchers with OptionValues {
                     .setPartyNotKnownOnLedger(PartyNotKnownOnLedger.newBuilder()),
                 Code.INVALID_ARGUMENT,
               ),
+              (
+                (builder: DamlTransactionRejectionEntry.Builder) =>
```
miklos-da, line 216: You can make these easier to read by specifying the type on the `Table`:
```scala
            Table[DamlTransactionRejectionEntry.Builder => DamlTransactionRejectionEntry.Builder, Code](
              ("rejection builder", "code"),
              ...
              (
                _.setDuplicateCommand(Duplicate.newBuilder()),
                Code.ALREADY_EXISTS,
              ),
              ...
```

PR #10748, opened 2021-09-02
Add ledger API tests about command deduplication information in completions [KVL-1057]

fabiotudone-da:
### Notes

- the new conformance tests are enabled only for all append-only-schema-supporting ledgers in new append-only schema test suites
- only the fact that a deduplication period is always returned in some form is currently checked; the precise relationship between submitted deduplication and completion deduplication is ledger-specific.
- rejection completions are not tested as the Ledger API does not specify when they can occur and, thus, the cases are ledger-specific

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandSubmissionCompletionIT.scala
```diff
@@ -164,4 +171,157 @@ final class CommandSubmissionCompletionIT extends LedgerTestSuite {
...
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    val request = ledger.submitRequest(party, Dummy(party).create.command)
+    for {
+      _ <- ledger.submit(request)
+      completions <- ledger.firstCompletions(party)
+    } yield {
+      val actualCompletionApplicationId = Option(singleCompletion(completions).applicationId)
+      assert(
+        Try(actualCompletionApplicationId.map(Ref.ApplicationId.assertFromString))
```
SamirTalwar-DA, line 186: Why not use `Ref.ApplicationId.fromString` here? That'll get you an `Either`.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandSubmissionCompletionIT.scala
```diff
@@ -164,4 +171,157 @@ final class CommandSubmissionCompletionIT extends LedgerTestSuite {
...
+    allocate(SingleParty),
+  )(implicit ec => { case Participants(Participant(ledger, party)) =>
+    val request = ledger.submitRequest(party, Dummy(party).create.command)
+    val requestedSubmissionId =
+      Some(Ref.SubmissionId.assertFromString(SubmissionIdGenerator.Random.generate()))
+    for {
+      _ <- ledger.submit(update(request, optSubmissionId = requestedSubmissionId))
+      completions <- ledger.firstCompletions(party)
+    } yield {
+      val actualCompletionSubmissionId = Option(singleCompletion(completions).submissionId)
```
SamirTalwar-DA, line 206: Why `Option`? If it's blank you'll get an empty string here, not `null`.

SamirTalwar-DA:
I think these tests look good but I also think it's worth noting that conformance tests are _slow_, and we run them about a hundred times in different ledger configurations (this may be a slight exaggeration).

Can we merge some of them together? I think we can probably get it down to 1 or 2 tests.

I also think that it would be reasonable to add these tests to `CommandSubmissionCompletionIT`, rather than adding a new suite, but I'm less worried about that.

fabiotudone-da:
> I think these tests look good but I also think it's worth noting that conformance tests are _slow_, and we run them about a hundred times in different ledger configurations (this may be a slight exaggeration).
> 
> Can we merge some of them together? I think we can probably get it down to 1 or 2 tests.

Yes, I'll look into that next week.

> I also think that it would be reasonable to add these tests to `CommandSubmissionCompletionIT`, rather than adding a new suite, but I'm less worried about that.

We need to enable it only for two ledgers at the moment, so that would mean excluding them in many places. Perhaps we can merge them when we enable them for most of the ledgers?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
```diff
@@ -0,0 +1,182 @@
...
+    }
+  })
+}
+
+private[testtool] object CompletionDeduplicationInfoIT {
+  sealed trait Service extends Serializable with Product
+  case object CommandService extends Service
+  case object CommandSubmissionService extends Service
+
+  private def submitRequest(
```
hubert-da, line 72: I think it might be better to split this function into 2 and put each version into the proper service `case object`.

hubert-da: Also, both cases look mostly the same, so it may be worth addressing this.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  private def assertDefined(optCompletion: Option[Completion]): Completion = {
+    assert(optCompletion.isDefined, "No completion has been produced")
+    optCompletion.get
+  }
+
+  private def simpleCreate(party: Primitive.Party): Command = Dummy(party).create.command
+
+  private val aSubmissionId =
```
hubert-da, line 178: Minor: I would put `val`s above `def`s.

hubert-da (approved):
LGTM, just minor comments

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
```diff
@@ -0,0 +1,182 @@
...
+        updateCommandServiceRequest = _.update(_.commands.deduplicationTime := aDeduplicationTime),
+        updateCommandSubmissionServiceRequest =
+          _.update(_.commands.deduplicationTime := aDeduplicationTime),
+      )
+    } yield {
+      assertApplicationIdIsPreserved(ledger.applicationId, optApplicationIdCompletion)
+      assertSubmissionIdIsGenerated(optApplicationIdCompletion)
+      assertDefaultDeduplicationTimeIsReported(config, optApplicationIdCompletion)
+      assertSubmissionIdIsPreserved(aSubmissionId, optSubmissionIdCompletion)
+      assertDeduplicationTimeIsPreserved(aDeduplicationTime, optCompletionDeduplicationTime)
```
andreaslochbihler-da, line 62: Hm. This test won't work for Canton. Ledger implementations are free to convert the provided deduplication period into some other format.

andreaslochbihler-da: In particular, Canton converts everything into a deduplication offset.

andreaslochbihler-da: Moreover, ledger implementations are free to extend the deduplication period as they see fit. So even kvutils is allowed (and after some discusson with @miklos-da probably will) modify the dedup duration.

fabiotudone-da: Thanks for the feedback, then I think we should just remove it, unless the API offers more specific guarantees that can be tested (like which conversions are allowed, which I think it doesn't atm).

fabiotudone-da: Implementation-specific API tests about this may make sense, though.

fabiotudone-da: 51e9b89

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
fabiotudone-da: 3e404be but it's more verbose and structurally more complex. Let me know which one you prefer.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
fabiotudone-da: Why? I like top-bottom code w.r.t. LOD and I see these constants as quite irrelevant details.

hubert-da: If it was a class, it would fail, I think, with a forward reference.  However, it makes sense to keep them at the bottom, ignore.

fabiotudone-da: Even in that case, I'd rather use `lazy`.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
hubert-da: Yeah, more verbose, as there's duplication that we won't be able to tackle easily. At least now the functions are smaller.  I'm fine with both.

fabiotudone-da: Unfortunately the two request types are only related at a `scalapb` level.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
```diff
@@ -0,0 +1,194 @@
...
+  ): Unit = {
+    val completion = assertDefined(optCompletion)
+    assert(completion.status.forall(_.code == Status.Code.OK.value()))
+    val actualDeduplication = completion.deduplicationPeriod
+    assert(actualDeduplication.isDeduplicationTime)
+    assert(
+      actualDeduplication.deduplicationTime == config.maxDeduplicationTime,
+      s"The deduplication $actualDeduplication " +
+        "is not the maximum deduplication time",
+    )
```
andreaslochbihler-da, line 157: There's no requirement on the Ledger API that we report the default deduplication period as a deduplication time. Canton, e.g., will report some offset that it derives from the max deduplication time.

A more meaningful test would look like this:
1. Get a command accepted with a very short dedup time.
2. Then submit a command with the same change id, but without a deduplication time.
3. If we get a completion before the max dedup time has elapsed after the acceptance of the first command, then the completion should be "ALREADY_EXISTS" and the reported deduplication period should start before the accepted command.

fabiotudone-da: > There's no requirement on the Ledger API that we report the default deduplication period as a deduplication time. Canton, e.g., will report some offset that it derives from the max deduplication time.

Yes, I was just thinking about this too.

> 3. If we get a completion before the max dedup time has elapsed after the acceptance of the first command
> [...] the reported deduplication period should start before the accepted command.

KV would report the max deduplication time, other ledgers some other deduplication form, so I think this test would be best implemented as ledger-specific.

At the general API level, I'd perhaps just ensure that some deduplication info is present in the completion no matter what.

fabiotudone-da:
@andreaslochbihler-da The test code is rather minimal and the general utilities in the companion object should be reusable. A KV-specific test will be soon added (in a separate PR). WDYT?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CompletionDeduplicationInfoIT.scala
fabiotudone-da: b60c75c

andreaslochbihler-da: > At the general API level, I'd perhaps just ensure that some deduplication ~info~ period (in some form) is present in the completion no matter what.

Yes. That makes sense too.

andreaslochbihler-da (approved):
Looks good from the perspective of the tested property.

ledger/ledger-on-sql/BUILD.bazel
```diff
@@ -396,6 +396,64 @@ conformance_test(
...
+    test_tool_args = [
+        "--verbose",
+        "--include=" +
+        "AppendOnlyCompletionDeduplicationInfoITCommandService" +
+        ",AppendOnlyCompletionDeduplicationInfoITCommandSubmissionService",
+    ],
+)
+
+conformance_test(
+    name = "conformance-test-append-only-h2-oracle-command-completion-dedup-info",
```
fabiotudone-da, line 419:
```suggestion
    name = "conformance-test-append-only-h2-command-completion-dedup-info",
```

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -437,3 +437,19 @@ server_conformance_test(
         "--include=ContractIdIT:Accept",
     ],
 )
+
+server_conformance_test(
+    name = "conformance-test-command-completion-dedup-info-append-only",
```
fabiotudone-da, line 442:
```suggestion
    name = "conformance-test-append-only-command-completion-dedup-info",
```

PR #10749, opened 2021-09-02
Wrap missing label names in quotes

cocreature:
See
https://discuss.daml.com/t/why-might-i-be-getting-missing-record-label-identity-from-the-ledger-api-when-submitting-a-command/3059/5
this has caused some confusion for users since it’s not obvious that
it is the label name if you have a very generic label name like
`identity`.

I have no strong feelings on whether this should be single or double
quotes so happy to change it.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
@@ -210,7 +210,7 @@ private[engine] final class ValueTranslator(
                       mbLbl.foreach(lbl_ =>
                         if (lbl_ != lbl)
                           typeError(
-                            s"Mismatching record field label $lbl_ (expecting $lbl) for record $tyCon"
+                            s"Mismatching record field label '$lbl_' (expecting '$lbl') for record $tyCon"
                           )
                       )
                       val replacedTyp = AstUtil.substitute(typ, subst)
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
@@ -210,7 +210,7 @@ private[engine] final class ValueTranslator(
                       mbLbl.foreach(lbl_ =>
                         if (lbl_ != lbl)
                           typeError(
-                            s"Mismatching record label $lbl_ (expecting $lbl) for record $tyCon"
+                            s"Mismatching record field label $lbl_ (expecting $lbl) for record $tyCon"
                           )
                       )
                       val replacedTyp = AstUtil.substitute(typ, subst)
@@ -220,7 +220,7 @@ private[engine] final class ValueTranslator(
                     recordFlds.map { case (lbl, typ) =>
                       labeledRecords
                         .get(lbl)
-                        .fold(typeError(s"Missing record label $lbl for record $tyCon")) { v =>
+                        .fold(typeError(s"Missing record field '$lbl' for record $tyCon")) { v =>
                           val replacedTyp = AstUtil.substitute(typ, subst)
                           lbl -> go(replacedTyp, v, newNesting)
                         }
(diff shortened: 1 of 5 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
```diff
@@ -210,7 +210,7 @@ private[engine] final class ValueTranslator(
                       mbLbl.foreach(lbl_ =>
                         if (lbl_ != lbl)
                           typeError(
-                            s"Mismatching record label $lbl_ (expecting $lbl) for record $tyCon"
+                            s"Mismatching record field label $lbl_ (expecting $lbl) for record $tyCon"
```
garyverhaegen-da, line 213:
```suggestion
                            s"Mismatching record field $lbl_ (expecting $lbl) for record $tyCon"
```

cocreature: I disagree with this, this is really about the `name` of the field, i.e., the field label having a mismatch.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandServiceIT.scala
```diff
@@ -280,7 +280,11 @@ final class CommandServiceIT extends LedgerTestSuite {
         .submitAndWait(badRequest)
         .mustFail("submitting a request with a bad parameter label")
     } yield {
-      assertGrpcError(failure, Status.Code.INVALID_ARGUMENT, s"Missing record label")
+      assertGrpcError(
+        failure,
+        Status.Code.INVALID_ARGUMENT,
+        Some(Pattern.compile(s"Missing record (label|field)")),
```
garyverhaegen-da, line 286: I assume the regex here is for backwards compatibility? I'm a little bit confused between this and the new exclusion you add to the compat matrix.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
garyverhaegen-da: Ok.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandServiceIT.scala
cocreature: The exclusion is for old ledger API test tools against new ledgers. This is to make sure we don’t need an exclusion for new ledger API test tools against old ledgers.

garyverhaegen-da: 👍

garyverhaegen-da (approved):
Thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
nickchapman-da, line 213: Everywhere else it is "record field", but here "record field label" -- should we remove "label" from here?
Also add single quotes around `$lbl_` and `$lbl`

nickchapman-da (approved):
We could argue that both "label" and "field"  terminology make sense in different places.
The "label" identifies the "field".
But it seems less confusing to pick just one term, "field", and use it consistently.
At least in the messages.

I'm probably just bike-shedding here. So sorry.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/ValueTranslator.scala
cocreature: quotes got lost during rebasing, added it.
As menitoned in the other comment, I don’t think `record field` here makes sense. We have a field in the right position it just has the wrong name. This is the case where we match fields by order. In the other instance, we match fields by name rather than order and there is no field with the given name so there we really don’t have a field.

PR #10753, opened 2021-09-02
LF: Simplify TransactionBuilder

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ValueEnricherSpec.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ValueEnricherSpec.scala
@@ -5,9 +5,8 @@ package com.daml.lf
 package engine
 
 import com.daml.lf.data._
-import com.daml.lf.language.Ast.{TNat, TTyCon}
+import com.daml.lf.language.Ast.{TNat, TTyCon, Type}
 import com.daml.lf.language.Util._
-import com.daml.lf.testing.parser.Implicits._
 import com.daml.lf.transaction.TransactionVersion
 import com.daml.lf.transaction.test.TransactionBuilder
 import com.daml.lf.value.Value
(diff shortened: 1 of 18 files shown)


daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ValueEnricherSpec.scala
```diff
@@ -16,13 +15,13 @@ import org.scalatest.prop.TableDrivenPropertyChecks
...
-import scala.language.implicitConversions
-
 class ValueEnricherSpec extends AnyWordSpec with Matchers with TableDrivenPropertyChecks {
 
-  import defaultParserParameters.{defaultPackageId => pkgId}
+  import TransactionBuilder.Implicits.{defaultPackageId => _, _}
+  import com.daml.lf.testing.parser.Implicits._
 
-  private implicit def toName(s: String): Ref.Name = Ref.Name.assertFromString(s)
+  private[this] implicit val defaultPackageId: Ref.PackageId =
```
cocreature, line 23: why do we need to change it instead of using the one from the builder?

cocreature (approved):
Thanks!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ValueEnricherSpec.scala
remyhaemmerle-da: Both the builder and the parser need a default Package ID.
We need the both package IDs to have the same value, either we override the one form the builder  either we override the one of the parser.  I have done the former. The latter could be done with something like that:
```suggestion
  private[this] implicit val defaultParserParameters: parser.ParserParameters[this.type] =
    defaultParserParameters.copy(defaultPackageId = defaultPackageId)
```

cocreature: makes sense, so the actual package id doesn’t matter it just matters that they’re the same :+1:

remyhaemmerle-da: :+1:

daml-lf/transaction-test-lib/src/main/scala/lf/transaction/test/TransactionBuilder.scala
```diff
@@ -81,81 +74,62 @@ final class TransactionBuilder(
...
       id: ContractId,
-      template: String,
+      templateId: Ref.Identifier,
       argument: Value,
-      signatories: Seq[String],
-      observers: Seq[String],
-      key: Option[Value],
+      signatories: Set[Ref.Party],
+      observers: Set[Ref.Party],
+      key: Option[Value] = None,
```
oliverse-da, line 91: Hi @remyhaemmerle-da , these looks like good changes and simple simplifications to make. Just wanted to make you aware that these will require me to adapt these canton-side tests (that I am still working on) - just as a heads-up that we are beginning to rely on `TransactionBuilder` more and more: https://github.com/DACH-NY/canton/pull/6969

PR #10754, opened 2021-09-02
Report some oracle_json_perf numbers on slack 

akshayshirahatti-da:
Add `avg, stddev, p90, p99, requests_per_second` numbers to be reported on slack similar to speedy_perf
```
json_oracle/fetchByKey: 208.0 106.0 185.0 470.0 33.333333333333336
json_oracle/fetchByQuery: 442.0 262.0 370.0 950.0 20.0
```
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/daily-compat.yml
+++ b/ci/cron/daily-compat.yml
@@ -216,7 +216,7 @@ jobs:
           }
 
           # dont want to wait forever to test connection , 15m is more than sufficient here.
-          declare -x testConnection
+          declare -xf testConnection
           timeout 15m bash -c 'until testConnection; do echo "Could not connect to Oracle, trying again..." ; sleep 1 ; done'
 
           # Actually run some tests
@@ -270,7 +270,7 @@ jobs:
           RES=""
           for KEY in $READ_PERF_KEYS; do
             # capture the avg, stddev, p90, p99, requests_per_second numbers from gatling summary csv
-            perf=$(cat ${OUT}/${KEY}/*/summary.csv | awk -F"," '{print $11,$12,$6,$8,$13}' | tail -n 1)
+            perf=$(cat ${OUT}/${KEY}/*/summary.csv | awk -F, '{print $11,$12,$6,$8,$13}' | tail -n 1)
             RES="${RES}json_oracle/${KEY}: $perf)\n"
           done
 
--- a/ci/cron/daily-compat.yml
+++ b/ci/cron/daily-compat.yml
@@ -269,11 +269,12 @@ jobs:
           "
           RES=""
           for KEY in $READ_PERF_KEYS; do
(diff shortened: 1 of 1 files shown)


ci/cron/daily-compat.yml
```diff
@@ -263,6 +263,18 @@ jobs:
               --query-store-index=oracle
           done
 
+          READ_PERF_KEYS="\
+           fetchByKey \
+           fetchByQuery\
+          "
+          RES=""
+          for KEY in $READ_PERF_KEYS; do
+            perf=$(cat ${OUT}/${KEY}/*/summary.csv | cut -d, -f8 | tail -n 1)
```
stefanobaghino-da, line 272: I would recommend using `awk` to get the field without relying on a specific length of the line using `cut`. E.g.
```suggestion
            perf=$(cat ${OUT}/${KEY}/*/summary.csv | awk -F, '{print $3}' | tail -n 1)
```
This is just an example and you don't have to merge it as is, but this would mean use `,` as a separator (`-F,`) and print the third column (one-based indexing).

ci/cron/daily-compat.yml
```diff
@@ -457,6 +470,7 @@ jobs:
             MSG="Daily tests passed: $COMMIT_LINK"
             REPORT='```
         speedy_perf: $(speedy_perf)
+        $(oracle_perf_results)
```
stefanobaghino-da, line 473: Should you add a label here to specify that what follows is the performance of the JSON API on Oracle?

stefanobaghino-da (commented):
Looks good, left a couple of suggestions, nothing blocking. 🙂

ci/cron/daily-compat.yml
akshayshirahatti-da: Label should not be necessary here as we are generating descriptive labelled stats from the job itself, we can do a sample run and see how it prints out, @garyverhaegen-da can you kick that off please ?

ci/cron/daily-compat.yml
stefanobaghino-da: Ohhh, yes, you're right sorry, I misread the original command.

ci/cron/daily-compat.yml
```diff
@@ -263,6 +263,19 @@ jobs:
...
           done
 
+          READ_PERF_KEYS="\
+           fetchByKey \
+           fetchByQuery\
+          "
+          RES=""
+          for KEY in $READ_PERF_KEYS; do
+            # capture the avg, stddev, p90, p99, requests_per_second numbers from gatling summary csv
+            perf=$(cat ${OUT}/${KEY}/*/summary.csv | awk -F"," '{print $11,$12,$6,$8,$13}' | tail -n 1)
```
garyverhaegen-da, line 273: I can't think of any reason why you'd need quotes around the comma here. 🤔

PR #10755, opened 2021-09-02
Handle dynamic port in auth middleware client (trigger service)

aherrmann-da:
The auth middleware client, which is used in the trigger service when authentication is enabled, did not handle dynamic port selection (port 0) correctly. The client needs to know the external URI to itself to construct the callback URI for the login flow. If not specified it was previously constructed from the host and port parameters. The port parameter may be 0 leading to an incorrect callback URI.

This PR changes the trigger service to fall back to a URI constructed based on the request URI if the user did not explicitly specify a callback URI.

To this end the auth middleware client is split into two parts. One part that is independent of the callback URI and holds state that needs to persist across requests to the client (e.g. the trigger service). Another part that takes the callback URI as a parameter and does not need to persist across requests.

This change allows us to remove the use of `LockedFreePort` in the trigger service tests and the auth middleware client tests and use port 0 instead.

This moves some code, so is probably easiest reviewed with
```
git diff -w --color-moved --color-moved-ws=allow-indentation-change
```

### Pull Request Checklist

trigger the build.

### DIFF
--- a/triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
+++ b/triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.{Cookie, Location, `Set-Cookie`}
+import akka.http.scaladsl.testkit.ScalatestRouteTest
 import akka.http.scaladsl.unmarshalling.Unmarshal
 import com.daml.auth.middleware.api.{Client, Request, Response}
 import com.daml.auth.middleware.api.Tagged.{AccessToken, RefreshToken}
@@ -21,9 +22,11 @@ import com.daml.ledger.api.testing.utils.SuiteResourceManagementAroundAll
 import com.daml.auth.oauth2.api.{Response => OAuthResponse}
 import org.scalatest.{OptionValues, TryValues}
 import org.scalatest.matchers.should.Matchers
-import org.scalatest.wordspec.AsyncWordSpec
+import org.scalatest.wordspec.{AnyWordSpec, AsyncWordSpec}
 
 import scala.collection.immutable
+import scala.concurrent.duration
+import scala.concurrent.duration.FiniteDuration
 import scala.io.Source
 import scala.util.{Failure, Success}
 
@@ -144,7 +147,7 @@ class TestMiddleware extends AsyncWordSpec with TestFixture with SuiteResourceMa
   "the /login endpoint" should {
     "redirect and set cookie" in {
       val claims = Request.Claims(actAs = List(Party("Alice")))
-      val req = HttpRequest(uri = middlewareClient.loginUri(claims, None))
+      val req = HttpRequest(uri = middlewareClientRoutes.loginUri(claims, None))
       for {
         resp <- Http().singleRequest(req)
         // Redirect to /authorize on authorization server
@@ -172,7 +175,7 @@ class TestMiddleware extends AsyncWordSpec with TestFixture with SuiteResourceMa
     }
     "return OK and set cookie without redirectUri" in {
       val claims = Request.Claims(actAs = List(Party("Alice")))
-      val req = HttpRequest(uri = middlewareClient.loginUri(claims, None, redirect = false))
+      val req = HttpRequest(uri = middlewareClientRoutes.loginUri(claims, None, redirect = false))
       for {
         resp <- Http().singleRequest(req)
         // Redirect to /authorize on authorization server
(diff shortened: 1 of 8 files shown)


triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -560,3 +563,67 @@ class TestMiddlewareClientAutoRedirectToLogin
...
+      authMiddlewareUri = Uri("http://auth.domain"),
+      redirectToLogin = Client.RedirectToLogin.Yes,
+      maxAuthCallbacks = 1000,
+      authCallbackTimeout = FiniteDuration(1, duration.MINUTES),
+      maxHttpEntityUploadSize = 4194304,
+      httpEntityUploadTimeout = FiniteDuration(1, duration.MINUTES),
+    )
+  )
+  "fixed callback URI" should {
+    "must be absolute" in {
```
stefanobaghino-da, line 582:
```suggestion
    "be absolute" in {
```

triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -560,3 +563,67 @@ class TestMiddlewareClientAutoRedirectToLogin
...
+      an[AssertionError] should be thrownBy client.routes(Uri().withPath(Uri.Path./("cb")))
+    }
+    "is used in login URI" in {
```
stefanobaghino-da, line 585:
```suggestion
    "be used in login URI" in {
```

triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -560,3 +563,67 @@ class TestMiddlewareClientAutoRedirectToLogin
...
+      val routes = client.routes(Uri("http://client.domain/cb"))
+      val claims = Request.Claims(actAs = List(Party("Alice")))
+      routes.loginUri(claims = claims) shouldBe
+        Uri(
+          s"http://auth.domain/login?claims=${claims.toQueryString()}&redirect_uri=http://client.domain/cb"
+        )
+    }
+  }
+  "callback URI from request" should {
+    "is used in login URI" in {
```
stefanobaghino-da, line 595:
```suggestion
    "be used in login URI" in {
```

triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -560,3 +563,67 @@ class TestMiddlewareClientAutoRedirectToLogin
...
+      Get("http://client.domain") ~> routes { routes =>
+        complete(routes.loginUri(claims).toString)
+      } ~> check {
+        responseAs[String] shouldEqual
+          s"http://auth.domain/login?claims=${claims.toQueryString()}&redirect_uri=http://client.domain/cb"
+      }
+    }
+  }
+  "automatic callback URI" should {
+    "is fixed when absolute" in {
```
stefanobaghino-da, line 608:
```suggestion
    "be fixed when absolute" in {
```

triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -560,3 +563,67 @@ class TestMiddlewareClientAutoRedirectToLogin
...
+      val routes = client.routesAuto(Uri("http://client.domain/cb"))
+      val claims = Request.Claims(actAs = List(Party("Alice")))
+      import akka.http.scaladsl.server.directives.RouteDirectives._
+      Get() ~> routes { routes => complete(routes.loginUri(claims).toString) } ~> check {
+        responseAs[String] shouldEqual
+          s"http://auth.domain/login?claims=${claims.toQueryString()}&redirect_uri=http://client.domain/cb"
+      }
+    }
+    "is from request when relative" in {
```
stefanobaghino-da, line 617:
```suggestion
    "be from request when relative" in {
```

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Server.scala
```diff
@@ -511,34 +515,34 @@ object Server {
...
               authCallbackTimeout = authCallbackTimeout,
               maxHttpEntityUploadSize = maxHttpEntityUploadSize,
               httpEntityUploadTimeout = httpEntityUploadTimeout,
             )
           )
-        )
+        val routes = client.routesAuto(authCallback.getOrElse(Uri().withPath(Path./("cb"))))
+        Some((client, routes))
     }
+    val authClient = authClientRoutes.map(_._1)
```
stefanobaghino-da, line 535: I may be missing something but from the diff I can't see where this is used.

stefanobaghino-da (commented):
Looks good, mostly nitpicking (`scalatest` prints `should` alongside the rest of the test description) and a minor question.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Server.scala
aherrmann-da: It's used for token refresh [further below](https://github.com/digital-asset/daml/blob/41c31007a546767870bf6706d272687f2f942f01/triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Server.scala#L629).

cocreature (approved):
Thanks!

PR #10756, opened 2021-09-02
[Docs] Propose a guideline on preparing final commit message when a PR is merged

pbatko-da:
What are your thoughts?


hubert-da:
LGTM, maybe I would also mention that the changelog entries should not be lost.

mziolekda (approved):
Sounds good

skisel-da (approved):

mziolekda (approved):

stefanobaghino-da:
This looks very good, but I think it would be way better to identify ways to enforce this.

stefanobaghino-da:
> > This looks very good, but I think it would be way better to identify ways to enforce this.
> 
> I don't think "write good commit messages" is something we can mechanically enforce within reasonable economic constraints.

I agree. I think the point of the guideline is valid but without a way to enforce this I don't think there's too much that will actually follow up from this.

S11001001:
I don't think there's very much benefit. The most important piece of data in the commit is the reference to the PR that produced it; that will always have much more useful context than a commit message could hope for.

Likewise, little harm, though.

PR #10757, opened 2021-09-02
Enhance InMemoryLedger to use the ValueEnricher

nickchapman-da:
Enhance `InMemoryLedger` to use the `ValueEnricher`.
- Allowing `SandboxServer` to run Engine with the default `transactionNormalization=false`.
- Enrichment is *not* done when transactions are published; the ACS stores un-enriched contracts and txs.
- Enrichment is done only when responding to verbose API queries.

This change advances the state of #6665

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
@@ -9,7 +9,7 @@ import com.daml.lf.language.{Ast, LookupError}
 import com.daml.lf.transaction.Node.{GenNode, KeyWithMaintainers}
 import com.daml.lf.transaction.{CommittedTransaction, Node, NodeId, VersionedTransaction}
 import com.daml.lf.value.Value
-import com.daml.lf.value.Value.ContractId
+import com.daml.lf.value.Value.{ContractId, VersionedValue}
 
 // Provide methods to add missing information in values (and value containers):
 // - type constructor in records, variants, and enums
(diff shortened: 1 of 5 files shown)


ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
```diff
@@ -353,12 +367,14 @@ private[sandbox] final class InMemoryLedger(
                 transactionId,
                 transaction,
               )
+          val enrichedCommittedTransaction =
```
cocreature, line 370: i don’t think always enriching is the right approach. We should do this in the translation if the `verbose` flag is set not always.

nickchapman-da: I did initially try the approach of enriching when returning transactions and contracts form the ledger (conditional on `verbose`), but I couldn't get it to work.

I think the problem is that some of the lookup methods might be dependant on having the in memory store the enriched transaction.

cocreature: Which lookups do you mean and which error did you get when trying that? Doing the enrichment conditionally on the `verbose` flag is what the SQL backend does so I don’t understand why it doesn’t work here.

nickchapman-da: discussion moved to slack

nickchapman-da: changed to the suggested approach

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
```diff
@@ -19,13 +19,28 @@ final class ValueEnricher(engine: Engine) {
   def enrichValue(typ: Ast.Type, value: Value[ContractId]): Result[Value[ContractId]] =
     engine.preprocessor.translateValue(typ, value).map(_.toUnnormalizedValue)
 
+  def enrichVersionedValue(
```
cocreature, line 22: looks like this and some of the other methods here could be private.

cocreature (approved):
Thanks!

remyhaemmerle-da (commented):
Nice.

PR #10758, opened 2021-09-02
Retry connection acquisition in PostgresStorageBackend.createDataSource

tudor-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10759, opened 2021-09-02
Fixes participant to do retries on startup as waiting for DB

nmarton-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
@@ -31,6 +31,7 @@ private[platform] class FlywayMigrations(
       .configure()
       .locations((locations(enableAppendOnlySchema, dbType) ++ additionalMigrationPaths): _*)
       .dataSource(dataSource)
+      .connectRetries(10) // Flyway does exponential backoff waiting 1s 2s 4s 8s ...
 
   def validate(): Future[Unit] = run {
     val flyway = configurationBase
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
@@ -36,7 +36,7 @@ import com.daml.platform.store.backend.{
 import javax.sql.DataSource
 import org.postgresql.ds.PGSimpleDataSource
 
-import scala.util.Using
+import scala.util.{Try, Using}
 
 private[backend] object PostgresStorageBackend
     extends StorageBackend[AppendOnlySchema.Batch]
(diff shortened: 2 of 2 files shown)


kamil-da:
Is this related? https://github.com/digital-asset/daml/pull/10758

nmarton-da:
> Is this related? #10758

indeed. see https://github.com/digital-asset/daml/pull/10758#issuecomment-911938140

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
```diff
@@ -31,6 +31,7 @@ private[platform] class FlywayMigrations(
       .configure()
       .locations((locations(enableAppendOnlySchema, dbType) ++ additionalMigrationPaths): _*)
       .dataSource(dataSource)
+      .connectRetries(600)
```
nmarton-da, line 34: deferring waiting for connectivity to Flyway.
this would do exponential backoff waiting 1 2 4 8 16 ... seconds between retries
Flyway is called almost every case at startup: exception sandboxnext reset triggered restart

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresStorageBackend.scala
```diff
@@ -257,7 +257,7 @@ private[backend] object PostgresStorageBackend
     val pgSimpleDataSource = new PGSimpleDataSource()
     pgSimpleDataSource.setUrl(jdbcUrl)
 
-    Using.resource(pgSimpleDataSource.getConnection())(checkCompatibility)
+    Try(Using.resource(pgSimpleDataSource.getConnection())(checkCompatibility))
```
nmarton-da, line 260: as a consequence this won't do much on the first datasource initialisation if noconnectivity, but it will later as ds-s are created for dbdispatchers (at that time we already waited once for connectivity)

rautenrieth-da, line 260: This silently drops any errors, wouldn't it be better to use `com.daml.timer.RetryStrategy.exponentialBackoff` or similar to retry at least a few times?

nmarton-da: I don't think it is a good idea to retry here, see explanation https://github.com/digital-asset/daml/pull/10758#issuecomment-911938140
And yes this will silently drop errors. But still it will run multiple times as we instantiate other datasources later, when connectivity should be already there. Not ideal I know.
The way I see it either we do something quickly if this is pushing (like this) and then we do the proper thing, or we just do the proper thing.

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
rautenrieth-da: [Flyway says](https://flywaydb.org/documentation/configuration/parameters/connectRetries) it always waits 1sec between retries. Waiting for 10min before declaring there is no connectivity to the database seems a bit long, no?

rautenrieth-da (approved):
I don't think silently dropping errors during data source creation is a good idea, but this is a quick fix to unblock others.

A proper fix should be added soon after https://github.com/digital-asset/daml/pull/10705 lands.

ledger/participant-integration-api/src/main/scala/platform/store/FlywayMigrations.scala
nmarton-da: as discussed offline:
* Flyway does exponential backoff now (despite docs) https://github.com/flyway/flyway/commit/37675a0e9c09b301f58142986924ad64cc0834b4
* lowered it to 10 ~ 1023 sec

PR #10745, opened 2021-09-01
Do not drop details when converting between gRPC `Status` classes [KVL-1084]

hubert-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10760, opened 2021-09-02
Populate `definite_answer` in `ApiException` [KVL-1004]

SamirTalwar-DA:
Also:
* Fixes 2 bugs mentioned in comments.
* Adds new conformance-test assertions (by @SamirTalwar-DA).
* As errors from rejections have the default "definite_answer", a proper mapping will be introduced in the next PR.

### Pull Request Checklist

trigger the build.


SamirTalwar-DA:
@hubert-da, I have decided to make this PR your problem. Do whatever you like with it.

hubert-da:
@SamirTalwar-DA JFYI I'm going to rebase it after this one is merged: https://github.com/digital-asset/daml/pull/10745
and then extend this PR by populating the "definite_answer"

ledger-api/rs-grpc-bridge/src/main/java/com/daml/grpc/adapter/server/rs/ServerSubscriber.java
```diff
@@ -114,7 +114,7 @@ public void onError(Throwable throwable) {
     executionSequencer.sequence(
         () -> {
           if (!responseObserver.isCancelled()) {
-            responseObserver.onError(Status.fromThrowable(throwable).asRuntimeException());
```
hubert-da, line 117: This was a bug resulting in dropping the trailers.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -117,20 +117,30 @@ object CompletionResponse {
   }
 
   private def buildException(metadata: Map[String, String], status: StatusJavaProto.Builder) = {
-    val newDetails = status.getDetailsList.asScala.map { detail =>
```
hubert-da, line 120: There was a bug as well, resulting in ignoring the metadata if the details were empty.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/ApiException.scala
```diff
@@ -3,9 +3,11 @@
...
 package com.daml.platform.server.api
 
-import io.grpc.{Status, StatusRuntimeException}
+import io.grpc.StatusRuntimeException
 
 import scala.util.control.NoStackTrace
 
 /** The sole purpose of this class is to give StatusRuntimeException with NoStacktrace a nice name in logs. */
-class ApiException(status: Status) extends StatusRuntimeException(status) with NoStackTrace
+class ApiException(exception: StatusRuntimeException)
```
hubert-da, line 11: Weird code, as I haven't found any other way to convert it while preserving the trailers as well as keeping it `NoStackTrace`.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -4,54 +4,163 @@
...
-      Status.NOT_FOUND.withDescription(
-        s"Ledger ID '${received.unwrap}' not found. Actual Ledger ID is '${expected.unwrap}'."
-      )
+      Status
+        .newBuilder()
+        .setCode(Code.NOT_FOUND.value())
+        .setMessage(
+          s"Ledger ID '${received.unwrap}' not found. Actual Ledger ID is '${expected.unwrap}'."
+        )
+        .addDetails(IndefiniteAnswerInfo)
```
fabiotudone-da, line 39: Why is it indefinite? What's the source of this mapping? If this is going to be the authoritative source, I'd add comments that explain why.

hubert-da: Good question. You reminded me that I should add more reviewers, as there is no source.

hubert-da:
Sorry for failed tests, we'll fix in the background.

hubert-da:
@oliverse-da what would we like to do with failed Canton conformance tests? Shall I exclude them for now?

oliverse-da:
> @oliverse-da what would we like to do with failed Canton conformance tests? Shall I exclude them for now? Or shall we merge the assertions later?

+ @andreaslochbihler-da : Writing here what I mentioned in the #team-deduplication slack channel: Logistically we probably want to produce another daml snapshot release shortly after merging this (with disabled tests), then produce a canton snapshot release, then upgrade the daml repo to use that. Before, it would make sense to use the `daml-sdk-head` approach to ensure that we have identified the canton-side changes to make the tests pass again.

hubert-da:
> Before, it would make sense to use the daml-sdk-head approach to ensure that we have identified the canton-side changes to make the tests pass again.

Thanks! Will do this once I finish sorting one other related failure, and will let you know.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -4,54 +4,163 @@
...
+    )
 
   def outOfRange(description: String): StatusRuntimeException =
-    grpcError(Status.OUT_OF_RANGE.withDescription(description))
+    grpcError(
+      Status
+        .newBuilder()
+        .setCode(Code.OUT_OF_RANGE.value())
+        .setMessage(description)
+        .addDetails(DefiniteAnswerInfo)
```
andreaslochbihler-da, line 79: AFAICT this should be an indefinite answer. Why do you think that we can meet the rank guarantee described in https://github.com/digital-asset/daml/pull/9567/files#diff-ed4ed30e4f593aaf9d013b39ed5272dfc0df08c548cc78f110fd4471eaa6f59cR143-R151 here?

Same question applies to `invalidField`, `invalidArgument`, and `missingField`.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
andreaslochbihler-da: That's an good question. `IndefiniteAnswerInfo` is certainly the safe choice here. We'd have to look at all use cases to understand whether we could make it definite here.

hubert-da:
We decided to provide and assert the `definite answer` just in the context of command deduplication, so I'm going to reshape this in a separate PR. Closing this one.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
hubert-da: @andreaslochbihler-da , thanks for the link. TBH, I've been focusing on what may happen if the application retries based on usages of the "factories":
* `outOfRange` - is being used only by `LedgerOffsetValidator` which on the other hand is being used by `GetTransactionsRequest` and `GrpcCompletionStreamRequest`. What I did not take into account is that there are cases where the ledger end is calculated on the server-side. I assumed that if someone passes invalid offsets multiple times, the answer is always going to be the same. However, it may not be reasonable anymore to provide the `definite_answer` info here, as this doesn't regard command submission/deduplication.
* `invalidField` - is being used for `GrpcCompletionStreamRequest` and **command** validation. I believe if a field is not being able to be parsed, it will always result in a validation error if submitted multiple times without changes.
* `invalidArgument` - used in many places. I assumed that if an argument is invalid, it won't ever be valid, which may be a wrong assumption. Will expand this a bit more below.
* `missingField` - used similarly as `invalidField`. I assumed that if a field is missing, it won't magically appear in a re-submission, so the validation passes.
What I'm concerned about is that these "factories" may be used in more places and different contexts, so perhaps it would be reasonable to pass the `definiteAnswer: Option[Boolean]` as an argument to them. For example, I wouldn't pass it in case of the current usages of `outOfRange`. WDYT?

andreaslochbihler-da: > it would be reasonable to pass the `definiteAnswer: Option[Boolean]` as an argument to them

That definitely sounds reasonable.



>     * if submitted multiple times without changes.

The guarantees about `definite_answer` are independent of whether the application submits the same request or a different request. That's because it's not clear what path such a request has already taken and whether something has augmented the request on the way. The server filling in the ledger end (or maybe ledger begin elsewhere) is one example. So what may look unretryable at a certain level in our stack may actually be due to bad choices having been made during preprocessing. That's why we've decided to deal with retryability separately.

hubert-da: Thanks a lot for the explanation, @andreaslochbihler-da, it's now much clearer to me.

PR #10761, opened 2021-09-02
DPP-577 Use BIGINT instead of TIMESTAMP

rautenrieth-da:
This PR changes all timestamp columns in the index database to use BIGINT (storing microseconds since UNIX epoch) instead of TIMESTAMP values.

This makes the index database independent of any time zone settings. Index databases produced by applications running in a non-UTC time zone can be fixed manually.


ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGField.scala
```diff
@@ -3,33 +3,8 @@
 
 package com.daml.platform.store.backend.postgresql
 
-import java.time.{Instant, ZoneOffset}
-import java.time.format.DateTimeFormatter
-
 import com.daml.platform.store.backend.common.Field
 
-private[postgresql] trait PGTimestampBase[FROM, TO] extends Field[FROM, TO, String] {
```
rautenrieth-da, line 11: Nice side effect of this change: less differences between StorageBackend implementations.

ledger/participant-integration-api/src/main/scala/platform/store/Conversions.scala
```diff
@@ -319,9 +320,17 @@ private[platform] object Conversions {
 
   // Instant
 
-  def instant(name: String): RowParser[Instant] =
+  // TODO append-only: Delete after removing the mutating schema. The append-only schema only uses BIGINT for timestamps.
+  def instantFromTimestamp(name: String): RowParser[Instant] =
     SqlParser.get[Date](name).map(_.toInstant)
 
+  def instantFromMicros(name: String): RowParser[Instant] =
+    SqlParser.get[Long](name).map { micros =>
```
nmarton-da, line 328: ....map(Instant.EPOCH.plus(_, ChronoUnit.MICROS))

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
```diff
@@ -91,11 +92,21 @@ private[backend] case class BooleanOptional[FROM](extract: FROM => Option[Boolea
   override def convert: Option[Boolean] => lang.Boolean = _.map(Boolean.box).orNull
 }
 
+private[backend] object Timestamp {
```
nmarton-da, line 95: can it be that it would be better to just drop the timestamp, and use the bigint field instead?
* Primarily conversions need to happen in one layer above, in UpdateToDDto, because that computation is happening on a different stage, computation done here is happening on Connection holding threads. (although in this particular case it does not sound expensive)
* We anyway convert almost anythign to Instant first, and then here to bigint...would it be easier to roll an implicit decorator to daml Timestamp with .toEpocMicros and change all .toInstant to .toEpochMicros?
* slightly lower footprint in the buffers with holding longs
* smaller surface in supported fields/fieldstrategy in the Schema framework

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
```diff
@@ -91,11 +92,21 @@ private[backend] case class BooleanOptional[FROM](extract: FROM => Option[Boolea
...
+  def instantToMicros(i: Instant): Long =
+    TimeUnit.SECONDS.toMicros(i.getEpochSecond) + TimeUnit.NANOSECONDS.toMicros(i.getNano.toLong)
+}
+
 private[backend] case class Timestamp[FROM](extract: FROM => Instant)
-    extends TrivialField[FROM, Instant]
+    extends Field[FROM, Instant, java.lang.Long] {
```
nmarton-da, line 101: the java type is only needed to force non primitive transport to signall nulls. this could be simply Long I think

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
```diff
@@ -91,11 +92,21 @@ private[backend] case class BooleanOptional[FROM](extract: FROM => Option[Boolea
   override def convert: Option[Boolean] => lang.Boolean = _.map(Boolean.box).orNull
 }
 
+private[backend] object Timestamp {
+  def instantToMicros(i: Instant): Long =
+    TimeUnit.SECONDS.toMicros(i.getEpochSecond) + TimeUnit.NANOSECONDS.toMicros(i.getNano.toLong)
```
nmarton-da, line 97: ChronoUnit.MICROS.between(Instant.EPOCH, i)

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
nmarton-da: also for the Timestamp: it could be an implicit decorator to get out the micros from Instant-s as well for the read side

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -85,16 +86,16 @@ private[backend] object H2StorageBackend
...
       key: String,
       submittedAt: Instant,
       deduplicateUntil: Instant,
   )(connection: Connection): Int =
     SQL(SQL_INSERT_COMMAND)
       .on(
         "deduplicationKey" -> key,
-        "submittedAt" -> submittedAt,
-        "deduplicateUntil" -> deduplicateUntil,
+        "submittedAt" -> Timestamp.instantToMicros(submittedAt),
```
nmarton-da, line 97: and this could be with the impicit decorator
`"submittedAt" -> submittedAt.toEpochMicros,`

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleField.scala
```diff
@@ -3,37 +3,11 @@
...
 package com.daml.platform.store.backend.oracle
 
-import java.sql.{PreparedStatement, Timestamp}
-import java.time.Instant
+import java.sql.PreparedStatement
 import com.daml.platform.store.backend.common.Field
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 
-private[oracle] case class OracleTimestamp[FROM](extract: FROM => Instant)
```
nmarton-da, line 12: yaaay less code

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PGField.scala
nmarton-da: yaaaaaay!

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsTimestamps.scala
```diff
@@ -0,0 +1,83 @@
...
+    } yield {
+      withClue("UTC") { events1.head.ledgerEffectiveTime shouldBe Some(let) }
+      withClue("GMT-1") { events2.head.ledgerEffectiveTime shouldBe Some(let) }
+      withClue("GMT+1") { events3.head.ledgerEffectiveTime shouldBe Some(let) }
+    }
+  }
+
+  // Some JDBC operations depend on the JVM default time zone.
+  // In particular, TIMESTAMP WITHOUT TIME ZONE columns are interpreted in the local time zone of the client.
+  private def withDefaultTimeZone[T](tz: String)(f: Connection => T)(connection: Connection): T = {
```
nmarton-da, line 74: here I would take extra care. 
This test cannot run in the same JVM as other tests in parallel, because it might cause problems with time related code.
How can we ensure that?

nmarton-da: nice play with currying :)

nmarton-da (commented):
looks good!
would be nice to drop timestamps altogether from the ingestion side
the timezone testing needs to be ensure that it does not disturb other things

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
rautenrieth-da: This code is ~copied from~ inspired by `com.daml.lf.data.Time.Timestamp` (which I did not want to depend on).

What would be the advantage of `ChronoUnit`? `ChronoUnit.MICROS.between` executes more code than `TimeUnit.SECONDS.toMicros`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
rautenrieth-da: Great suggestion. The `com.daml.lf.data.Time.Timestamp` type is just a wrapper for a Long value. We can just use `u.recordTime.micros` in UpdateToDto, no need for an implicit decorator. It saves space in DbDto values, and we don't spend time converting to Instant and back to Long.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsTimestamps.scala
rautenrieth-da: I don't think this can cause problems:
- `StorageBackend` tests do not run in parallel (there is code to enforce this)
- each Java/Scala test executable invokes java separately
- `withDefaultTimeZone ` uses try-finally to restore the original time zone in case of test failure

It might mess up things like log timestamps, but that should not make the test flaky.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
rautenrieth-da: I have modified `DbDto`s to store `Long` values. All input timestamps in `Update` use the `com.daml.lf.data.Time.Timestamp` type, so conversion to micros is trivial.

At some point in the future, we should modify the rest of `StorageBackend` to return `com.daml.lf.data.Time.Timestamp` values instead of `Instant` values. Almost everything in Daml (except perhaps deduplication time windows?) uses microsecond resolution, converting to `Instant` and then to `com.daml.lf.data.Time.Timestamp` is just wasting resources.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/Field.scala
nmarton-da: I am not sure what you mean...I had only the client usage in mind, that seems simpler then summing up a tomicros getEpochSecond and a toMicros getNano getLong, or not?

nmarton-da (approved):
minor comments left

PR #10766, opened 2021-09-03
Clean up sandbox-on-x conformance tests.

mziolekda:
Make sure that the unsettled state of the ledger does not spill over to the subsequent test. 
Without that stabilization the subsequent test may pick-up noise and fail for all the wrong reasons.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala
@@ -71,6 +71,10 @@ final class ConfigManagementServiceIT extends LedgerTestSuite {
           newTimeModel = oldTimeModel,
         )
         .mustFail("setting a time model with an expired MRT")
+
+      // Above operation finished on a timeout, but may still succeed asynchronously.
+      // Stabilize the ledger state before leaving the test case.
+      _ <- stabilize(ledger)
     } yield {
       assert(
         response1.configurationGeneration < response2.configurationGeneration,
@@ -240,4 +244,11 @@ final class ConfigManagementServiceIT extends LedgerTestSuite {
         Failure(failure)
 
     }
+
+  // Stabilize the ledger by writing a new element and observing it in the indexDb.
+  // The allocateParty method fits the bill.
+  private def stabilize(ledger: ParticipantTestContext)(implicit
+      executionContext: ExecutionContext
+  ): Future[Unit] =
+    ledger.allocateParty().map(_ => ())
 }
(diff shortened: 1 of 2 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala
```diff
@@ -71,6 +71,10 @@ final class ConfigManagementServiceIT extends LedgerTestSuite {
           newTimeModel = oldTimeModel,
         )
         .mustFail("setting a time model with an expired MRT")
+
+      // Above operation finished on a timeout, but may still succeed asynchronously.
+      // Stabilize the ledger state before leaving the test case.
+      _ <- stabilize(ledger)
```
nmarton-da, line 77: but here we still have to read after stabilizing, and making sure that what we read is not the failed one (as that one will fail anyway as we discussed)
or is this what you wanted to have a separate PR for?

nmarton-da (commented):
LGTM

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/ConfigManagementServiceIT.scala
rautenrieth-da: I also don't understand why we wait for the participant to catch up, but don't read from it afterwards.

miklos-da: @mziolekda Do you want to move this back to draft?

mziolekda: This is making sure that the unsettled state of the ledger does not spill over to the subsequent test. Without stabilizing the subsequent test may pick-up noise and fail for all the wrong reasons.

mziolekda: Oh I see where the confusion comes from, allocateParty is a write+read operation. It waits until party allocation arrives in teh index-db.

rautenrieth-da: > making sure that the unsettled state of the ledger does not spill over to the subsequent test

I see. `setTimeModel` is _also_ a read+write operation, blocking until the change was witnessed in the index db. However, the last call to `setTimeModel` is using an already expired MRT, which means the participant will immediately abort waiting for the confirmation of the change.

You also just swap the "Restore the original time model" and "Try to set a time model with an expired MRT." blocks. The "Restore the original time model" call will serve the same purpose as the `stabilize` call.

rautenrieth-da (approved):
The change makes sense, left a comment about solving the spill over state without adding the `stabilize` method.

PR #10768, opened 2021-09-03
fix claims check in auth middleware

garyverhaegen-da:
The claims check in the auth middleware was switched around: in effect,
it checked that we did not receive _more_ than we asked for, rather than
checking we receive _at least_ what we asked for.

Of course this would still not let anyone run any trigger without the
proper access token, but it would let people list running triggers and
request (or stop) trigger executions.

CHANGELOG_BEGIN
- Fix a bug in the auth middleware where insufficient credentials could
  still give access to list of running triggers.
CHANGELOG_END

### DIFF
--- a/triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
+++ b/triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
@@ -30,7 +30,11 @@ import scala.concurrent.duration.FiniteDuration
 import scala.io.Source
 import scala.util.{Failure, Success}
 
-class TestMiddleware extends AsyncWordSpec with TestFixture with SuiteResourceManagementAroundAll {
+class TestMiddleware
+    extends AsyncWordSpec
+    with TestFixture
+    with SuiteResourceManagementAroundAll
+    with Matchers {
   private def makeToken(
       claims: Request.Claims,
       secret: String = "secret",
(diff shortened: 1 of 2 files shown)


triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
```diff
@@ -96,12 +96,42 @@ class TestMiddleware extends AsyncWordSpec with TestFixture with SuiteResourceMa
...
+        aliceA_bobA <- r("Alice", "Bob")()
+        aliceA_bobR <- r("Alice")("Bob")
+        aliceR_bobA <- r("Bob")("Alice")
+        aliceR_bobR <- r()("Alice", "Bob")
+        bobA <- r("Bob")()
+        bobR <- r()("Bob")
+        bobAR <- r("Bob")("Bob")
       } yield {
-        assert(result == None)
+        assert(aliceA.isEmpty)
```
cocreature, line 126: `aliceA shouldBe None` gives you a better error if it fails iirc.

cocreature (approved):
uff, that’s bad. Thanks for catching & fixing it

triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala
garyverhaegen-da: Done.

aherrmann-da (approved):
Thank you for catching this!

stefanobaghino-da (commented):
👍🏻 🙇🏻

PR #10769, opened 2021-09-03
DPP-577 bad migration

rautenrieth-da:
DO NOT MERGE

This is to show that the compatibility tests would catch a broken database migration.


rautenrieth-da:
This PR served its sole purpose (showing that `PRs (compatibility_linux)` would fail). Closing.

PR #10770, opened 2021-09-03
Fix flaky DBLock tests

nmarton-da:
As it turned out normal connection closeing does not imply released resources on the database side

* Adding synchronous eventually support
* use that for testing connection lost cases
* refactor a case to use proper release instead of connection close
* Increases a timeout to lower risk of flakyness

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
+++ b/ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
@@ -7,13 +7,19 @@ import java.sql.Connection
 
 import com.daml.platform.store.backend.DBLockStorageBackend.{Lock, LockId, LockMode}
 import org.scalatest.Assertion
+import org.scalatest.concurrent.Eventually
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
 import org.scalatest.flatspec.AsyncFlatSpec
 import org.scalatest.matchers.should.Matchers
+import org.scalatest.time.{Seconds, Span}
 
 import scala.concurrent.Future
 import scala.util.Try
 
-private[backend] trait StorageBackendTestsDBLock extends Matchers with StorageBackendSpec {
+private[backend] trait StorageBackendTestsDBLock
+    extends Matchers
+    with StorageBackendSpec
+    with Eventually {
   this: AsyncFlatSpec =>
 
   behavior of "DBLockStorageBackend"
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
```diff
@@ -153,4 +154,20 @@ private[backend] trait StorageBackendTestsDBLock extends Matchers with StorageBa
       Future.successful(1 shouldBe 1)
     }
   }
+
+  def eventually[T](t: => T): T = {
```
nmarton-da, line 158: seems like something we might have somewhere. If we do please come forward

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
```diff
@@ -86,29 +87,29 @@ private[backend] trait StorageBackendTestsDBLock extends Matchers with StorageBa
...
+      backend.release(shared3)(c(3))
       backend.tryAcquire(backend.lock(1), LockMode.Exclusive)(c(4)) should not be empty
   }
 
   it should "lock immediately, or fail immediately" in dbLockTestCase(2) { c =>
     backend.tryAcquire(backend.lock(1), LockMode.Shared)(c(1)) should not be empty
     val start = System.currentTimeMillis()
     backend.tryAcquire(backend.lock(1), LockMode.Exclusive)(c(2)) shouldBe empty
-    (System.currentTimeMillis() - start) should be < 100L
+    (System.currentTimeMillis() - start) should be < 500L
```
nmarton-da, line 112: could be another source of flakyness

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
cocreature, line 158: I’m a bit confused why you need this. Why does scalatest’s `eventually` not do the trick? That should work for both futures and non-future stuff.

cocreature (approved):
Thanks! I don’t quite understand why you need your own `eventually` but no big issue defining it.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsDBLock.scala
nmarton-da: no confusion, I thought that is only for futures, thanks! changed it!

mziolekda (approved):

PR #10774, opened 2021-09-03
FreePort draw from outside ephemeral port range

aherrmann-da:
`LockedFreePort` uses collaborative locking via a file lock on the chosen port. Previously the port was taked out of the ephemeral range by binding on port 0 and then releasing it again right away. However, other parts of the code base bind on port 0 without using `LockedFreePort`. This leaves a potential for collisions on ports.

This changes `FreePort` to draw ports from outside the ephemeral port range.

### Pull Request Checklist

trigger the build.


libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,102 @@
...
+        try {
+          val socket = new ServerSocket(candidate, 0, InetAddress.getLoopbackAddress)
+          socket.close()
+          true
+        } catch {
+          case _: IOException => false
+        }
+      }
+      .map(found => Port(found))
+      .get
```
SamirTalwar-DA, line 31: Consider a more meaningful exception if this fails.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,102 @@
...
+import scala.io.Source
+import scala.sys.process.Process
+import scala.util.{Random, Try}
+
 object FreePort {
 
   def find(): Port = {
-    val socket = new ServerSocket(0, 0, InetAddress.getLoopbackAddress)
+    val maxAttempts = 100
+    val dynamicRange = dynamicPortRange()
```
SamirTalwar-DA, line 17: Can you make these `private val`s of the containing object?

SamirTalwar-DA (approved):
Seems heavy, but probably a good idea.

aherrmann-da (commented):
> Consider a more meaningful exception if this fails.

Done.

There are a couple more things I cleaned up in this PR. I had opened it as a draft early to run CI on Windows and MacOS.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,127 @@
...
+import scala.util.{Failure, Random, Try}
+
 object FreePort {
+  private val maxAttempts = 100
+  private[ports] val dynamicRange = dynamicPortRange()
 
+  /** Find a free port.
+    *
+    * This determines whether a port is free by attempting to bind it and immediately releasing it again.
+    * Note, this does not guarantee that the port is still free after this function returns, see [[LockedFreePort]].
```
aherrmann-da, line 22: I think there is a potential race condition with the `LockedFreePort` approach, as described in the commit message [here](https://github.com/digital-asset/daml/pull/10774/commits/7805c40ae7e56f3b9dbdf2574bb3b0c7210e8846). However, that issue was already present before this PR, so I'd prefer to address it separately.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,127 @@
...
+        }
+      }
+      .map(found => Port(found))
+      .getOrElse(throw NoFreePortException)
+  }
+
+  sealed abstract class FindPortException(message: String, cause: Throwable)
+      extends RuntimeException(message, cause) {
+    def this(message: String) = this(message, null)
+  }
```
stefanobaghino-da, line 47:
```suggestion
  sealed abstract class FindPortException(message: String, cause: Throwable = null)
      extends RuntimeException(message, cause)
```
This should be effectively the same.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,127 @@
...
+    val numAvailablePorts = numLowerPorts + numUpperPorts
+    val gen = new Random()
+    def genPort(): Int = {
+      val n = gen.nextInt(numAvailablePorts)
+      if (n < numLowerPorts) {
+        n + minPort
+      } else {
+        n - numLowerPorts + maxExcl + 1
+      }
+    }
```
stefanobaghino-da, line 69:
```suggestion
    def genPort(): Int = {
      val n = scala.util.Random.nextInt(numAvailablePorts)
      if (n < numLowerPorts) {
        n + minPort
      } else {
        n - numLowerPorts + maxExcl + 1
      }
    }
```
Instantiating a `Random` is useful to improve testability using a fixed seed, otherwise you can just use the companion object which already is a `Random` instance.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,127 @@
...
+    def wrapError[A](a: Try[A]): A = a.recoverWith { case e =>
+      Failure(DynamicPortRangeException(e))
+    }.get
+    sys.props("os.name") match {
+      case os if os.toLowerCase.contains("windows") =>
+        wrapError(windowsDynamicPortRange())
+      case os if os.toLowerCase.contains("mac") =>
+        wrapError(macosDynamicPortRange())
+      case os if os.toLowerCase.contains("linux") =>
+        wrapError(linuxDynamicPortRange())
```
stefanobaghino-da, line 83: This starts to be quite finicky. We only use this for testing, so we control the environment, right?

stefanobaghino-da (commented):
Looks good, thanks!

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
```diff
@@ -3,17 +3,127 @@
...
+  }
+
+  final case object NoFreePortException extends FindPortException("No free port found")
+
+  final case class DynamicPortRangeException(cause: Throwable)
+      extends FindPortException("Failed to determine the dynamic port range", cause)
+
+  private def randomPortGen(dynamicRange: (Int, Int)): () => Int = {
+    val (minPort, maxPort) = (1024, 65536)
+    val minExcl = Math.min(Math.max(minPort, dynamicRange._1), maxPort)
```
cocreature, line 56: maybe a comment on the range you’re trying to find here would be useful.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
cocreature: yes should only be used for testing. Maybe worth considering splitting `libs-scala/ports` into a testing lib and a non-testing lib but let’s leave that for a separate PR.

cocreature (approved):
Thanks!
I admit I’m somewhat scared by the increasing complexity here. I took a look at where we actually use this and afaict it’s only used for two things at this point:
1. toxiproxy
2. postgres

I wonder if we could find solutions there where we bind directly to port 0 and scrape it from log files and kill LockedFreePort completely.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
aherrmann-da: :+1: Was taken from somewhere else in the repo.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
aherrmann-da: From CI:
```
Some internal dependencies are not published to Maven Central!
//daml-script/converter:converter
	- //bazel_tools/runfiles:scala_runfiles
```
We may need to split the testing part out either way.

libs-scala/ports/src/main/scala/com/digitalasset/ports/FreePort.scala
aherrmann-da: done

aherrmann-da:
> Thanks!
> I admit I’m somewhat scared by the increasing complexity here. I took a look at where we actually use this and afaict it’s only used for two things at this point:
> 
>     1. toxiproxy
> 
>     2. postgres
> 
> 
> I wonder if we could find solutions there where we bind directly to port 0 and scrape it from log files and kill LockedFreePort completely.

Yes, unfortunately toxiproxy generates no port file and prints port 0 (not the bound port) [to the log](https://github.com/Shopify/toxiproxy/blob/115744c10c81899b133efca6ff16053f74b2d9d3/api.go#L81). Alternatively, we could patch toxiproxy.

PR #10792, opened 2021-09-06
Fix a race in a test of the RecoveringIndexer

mziolekda:
Make sure the list of openSubscriptions is updated before the resulting future is passed to the test code for further processing.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
+++ b/ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
@@ -367,11 +367,10 @@ object RecoveringIndexerSpec {
             }
           })
         })(handle => {
-          val complete = handle.stop()
-          complete.onComplete { _ =>
+          handle.stop().transform { complete =>
             openSubscriptions -= handle.completed()
+            complete
           }
-          complete
         }).map(_.completed())
       }
     }


ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
```diff
@@ -367,11 +367,10 @@ object RecoveringIndexerSpec {
             }
           })
         })(handle => {
-          val complete = handle.stop()
-          complete.onComplete { _ =>
+          handle.stop().transform { complete =>
```
tudor-da, line 370: What is the difference? (Except for the nicer construction)

kamil-da: I think previously we did:
```
val complete = handle.stop()
complete.onComplete { doStuff() }
complete // return complete immediately without waiting for doStuff()
```

and now it is:
```
handle.stop().transform { complete =>
doStuff() // do stuff when stop() completes
complete // return complete but this time only when doStuff() is completed
}

tudor-da: I see, thanks

rautenrieth-da: I think: Before, the `.onComplete` handler and the `.map` transformation could run in any order. With this change, it is guaranteed that the subscription is removed first.

tudor-da (approved):

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/RecoveringIndexerSpec.scala
mziolekda: In the past the lambda in onComplete was racing against the client code attached to the returned future. Now the lambda is executed strictly before.

PR #10793, opened 2021-09-06
DPP-577 Check compatibility of timestamps

rautenrieth-da:
This PR adds a check to the compatibility tests that verifies that the ledger effective time of existing transactions is not modified by an SDK upgrade.

Adding this test was motivated by the upcoming migration from TIMESTAMP to BIGINT columns to store timestamp values.


cocreature (approved):
Thanks!

PR #10794, opened 2021-09-07
Cleanup DataSource initialization [DPP-583]

nmarton-da:
* Introduce VerifiedDataSource to capture async verification and factoring
* Move compatibility check to VerifiedDataSource
* Wire to FlywayMigrations
* Cleanup HikariDataSourceOwner

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/VerifiedDataSource.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/VerifiedDataSource.scala
@@ -1,0 +1,52 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.backend
+
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.platform.store.DbType
+import com.daml.timer.RetryStrategy
+import javax.sql.DataSource
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.DurationInt
+import scala.util.Using
+
+/** Returns a DataSource that is guaranteed to be connected to a responsive, compatible database. */
+object VerifiedDataSource {
+
+  private val MaxInitialConnectRetryAttempts: Int = 600
+
+  private val logger = ContextualizedLogger.get(this.getClass)
+
+  def apply(jdbcUrl: String)(implicit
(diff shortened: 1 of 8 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/VerifiedDataSource.scala
```diff
@@ -0,0 +1,51 @@
...
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import com.daml.platform.store.DbType
+import com.daml.timer.RetryStrategy
+import javax.sql.DataSource
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration.DurationInt
+import scala.util.Using
+
+object VerifiedDataSource {
```
rautenrieth-da, line 15: There are many classes in `com.daml.platform.store.backend`. At least for me, a short documentation of each class helps me navigating the code base.

```suggestion
/** Returns a DataSource that is guaranteed to be connected to a responsive, compatible database. */
object VerifiedDataSource {
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
```diff
@@ -705,4 +705,7 @@ private[backend] trait CommonStorageBackend[DB_BATCH] extends StorageBackend[DB_
       stmnt.close()
     }
   }
+
+  override def checkDatabaseAvailable(connection: Connection): Unit =
+    assert(SQL"SELECT 1".as(get[Int](1).single)(connection) == 1)
```
rautenrieth-da, line 710: What log output will this produce if the database is not available? The output of the `assert` doesn't look useful. If if even can fail at all - when would `SELECT 1` not return `1`?

rautenrieth-da (approved):
Change looks good, left two minor comments.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/CommonStorageBackend.scala
nmarton-da: log: if this fails the VerifiedDataSource will produce logs, that it attempts to connect to the db
select 1: with this we can make sure a full roundtrip between db and app, like a sanity check: testing all layers. I'd like to keep  it in if you don't mind.

kamil-da (approved):

PR #10795, opened 2021-09-07
fix macOS Bazel cache

garyverhaegen-da:
macOS filesystems have been case-insensitive by default for years, and in particular our laptops are, so if we want the cache to work as expected, CI should be too.

Note: this does not apply to Nix, because the Nix partition is a case-sensitive image per @nycnewman's script on laptops too.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/macos/2-common-box/init.sh
+++ b/infra/macos/2-common-box/init.sh
@@ -94,7 +94,7 @@ reset_cache() {
     fi
 
     rm -f "${file}.sparseimage"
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

aherrmann-da (approved):

garyverhaegen-da:
This is now fully deployed to all nodes.

PR #10800, opened 2021-09-07
get oracle perf test output

garyverhaegen-da:
We can only reference variables set in a named step.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/cron/daily-compat.yml
+++ b/ci/cron/daily-compat.yml
@@ -265,13 +265,13 @@ jobs:
 
           READ_PERF_KEYS="\
            fetchByKey \
-           fetchByQuery\
+           fetchByQuery \
           "
           RES=""
           for KEY in $READ_PERF_KEYS; do
             # capture the avg, stddev, p90, p99, requests_per_second numbers from gatling summary csv
             perf=$(cat ${OUT}/${KEY}/*/summary.csv | awk -F, '{print $11,$12,$6,$8,$13}' | tail -n 1)
-            RES="${RES}json_oracle/${KEY}: $perf)\n"
+            RES="${RES}json_oracle/${KEY}: $perf\n"
           done
 
           setvar oracle_perf_results "$RES"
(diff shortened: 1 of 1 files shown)


ci/cron/daily-compat.yml
```diff
@@ -281,6 +281,7 @@ jobs:
           gcs "$GCRED" cp "$OUT.tgz" "gs://daml-data/perf/http-json-oracle/${REPORT_ID}.tgz"
 
         displayName: http-json-oracle perf
+        name: out
```
garyverhaegen-da, line 284: The variable is referenced as `out.oracle_perf_results` so the job needs to be named `out`.

stefanobaghino-da (approved):
🙇🏻

akshayshirahatti-da (approved):

ci/cron/daily-compat.yml
stefanobaghino-da: Is there some way in which we can catch these kind of mistakes earlier?

garyverhaegen-da: Not an automated one I can think of, no. Running the job when it's a PR and checking the output is the only way I know of. Which I thought we'd done for this one, but I guess it slipped through.

garyverhaegen-da: Looks like the [compat test on the branch failed](https://dev.azure.com/digitalasset/daml/_build/results?buildId=87510&view=results), so the actual print step never ran.

Best lesson I can think of here is if you've added a job _and changed another_, you should ensure _both_ succeed before merging.

akshayshirahatti-da: @garyverhaegen-da I had to merge this PR since I had broken compat-tests with the `declare -x fn` change so this was pushed before we completed the daily-compat tests, which takes hours to run. I was monitoring the channel but it seems this was an unfortunate case of other things breaking adding noise in our ways to test and react quickly.

garyverhaegen-da: Right, so I guess another lesson we could take from this is to push bugfixes in their own separate PRs, rather than mixing them with new features.

PR #9499, opened 2021-04-26
rank-based-deduplication: add draft of kv_utils.proto changes

meiersi-da:
This PR serves as a draft to aid the design discussion around rank-based command deduplication.

FYI: @andreaslochbihler-da 

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10801, opened 2021-09-07
kvutils: Populate `definite_answer` in rejections [KVL-1004]

hubert-da:
### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -445,8 +447,12 @@ private[state] object Conversions {
   def decodeTransactionRejectionEntry(
       entry: DamlTransactionRejectionEntry
   ): Option[FinalReason] = {
-    def buildStatus(code: Code, message: String) = {
-      Status.of(code.value, message, Seq.empty)
+    def buildStatus(code: Code, message: String, metadata: Map[String, String]) = {
```
hubert-da, line 450: @miklos-da, `definite_answer` is not something that we'd like to store, right? So, we can just convert it while decoding.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -456,6 +462,7 @@ private[state] object Conversions {
           buildStatus(
             Code.ABORTED,
             s"Invalid ledger time: ${rejection.getDetails}",
+            Map(GrpcStatuses.DefiniteAnswerKey -> "false"),
```
fabiotudone-da, line 465: Why isn't it definite? Also, what's the source of this mapping?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -464,6 +471,7 @@ private[state] object Conversions {
           buildStatus(
             Code.INVALID_ARGUMENT,
             s"Disputed: ${rejection.getDetails}",
+            Map(GrpcStatuses.DefiniteAnswerKey -> "true"),
```
nicu-da, line 474: Looks like we always just set this key so we can move it into the `buildStatus` as a bool value

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: Similarly, no source. We have to come up with the values, as discussed with Miklos and Samir on Slack.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: I had an impression that in the next story, we may want to add the "structured metadata" here.

nicu-da: I would still split them, so for the metadata just add a new field to the build status so that there's a clear split between them at least in the builder

hubert-da: Yes, I'm going to add it here, so I would leave it as is.

hubert-da: Ok

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: Based on my current understanding, we can only tell that the `DUPLICATE_COMMAND` is a definite answer. @andreaslochbihler-da, is it correct?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
andreaslochbihler-da: I don't fully understand this comment as I don't know where this should not be stored. Whether something is a definite answer should be decided by the committer in kvutils. Accordingly, I'd expect that the participant needs to store whether the committer has marked something as `definite_answer` or not. If this comment is completely out of place, just ignore it.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
andreaslochbihler-da: The rule of thumb is as follows: If the committer has updated the command dedup key with this transaction, then it can be a definite answer. Otherwise it is never a definite answer. AFAIK a `DUPLICATE_COMMAND` result never updates the command dedup key, so it's always indefinite.

hubert-da:
Moving back to a draft, as I'm going to add the `definite_answer` to the proto according to: https://github.com/digital-asset/daml/pull/9499/files.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: Thanks, @andreaslochbihler-da. I needed some additional explanation from @miklos-da and I think I get it now. We think that the only place where we'll be able to set the `definite_answer` to true is the rank-based command deduplication rejection (future use). Do you know any other example?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: @miklos-da recommended adding the `definite_answer` to the proto according to: https://github.com/digital-asset/daml/pull/9499/files#diff-a5535746ffadbcb0c4b32562d134d1de6f87e488eb96aefd72cea1ed1a0748e8R269, as in the future it won't be possible to determine whether a response is definite or not on the participant-side (here). Therefore it is going to be persisted on the ledger and just populated in the `Status` here, so we have everything in place.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
andreaslochbihler-da: No. That's what `definite_answer` was designed for. When we start to add rank-based deduplication logic to other commands (e.g., party migration), we'll probably use the same concept there as well.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -194,6 +194,14 @@ message DamlTransactionRejectionEntry {
   DamlSubmitterInfo submitter_info = 1;
   reserved 5; // was reason.maximum_record_time_exceeded
 
+  // Whether this rejection should be considered a definite answer
+  // as defined in rank-based deduplication; i.e., whether this rejection
+  // signals that no other transaction with the same change-id and equal or lower submission_rank
+  // can be accepted or rejected.
```
miklos-da, line 200: To avoid confusion, I'd suggest removing mentions of rank-based command deduplication here:
```suggestion
  // Whether this rejection should be considered a definite answer; i.e., whether this rejection
  // signals that no other transaction with the same change-id can be accepted or rejected.
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -446,7 +450,17 @@ private[state] object Conversions {
       entry: DamlTransactionRejectionEntry
   ): Option[FinalReason] = {
     def buildStatus(code: Code, message: String) = {
-      Status.of(code.value, message, Seq.empty)
+      Status.of(
+        code.value,
+        message,
+        Seq(
+          AnyProto.pack[ErrorInfo](
```
miklos-da, line 457: :+1:

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
```diff
@@ -338,7 +341,12 @@ object KeyValueConsumption {
               Status.of(
                 Code.ALREADY_EXISTS.value,
                 "Duplicate commands",
-                Seq.empty,
+                Seq(
+                  AnyProto.pack[ErrorInfo](
+                    // the definite answer is false, as the rank-based deduplication is not yet implemented
+                    ErrorInfo(metadata = Map(GrpcStatuses.DefiniteAnswerKey -> "false"))
```
miklos-da, line 347: This must take the value from `transactionRejectionEntry.definiteAnswer`.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
```diff
@@ -371,7 +379,11 @@ object KeyValueConsumption {
               Status.of(
                 Code.ABORTED.value,
                 reason,
-                Seq.empty,
+                Seq(
+                  AnyProto.pack[ErrorInfo](
+                    ErrorInfo(metadata = Map(GrpcStatuses.DefiniteAnswerKey -> "false"))
```
miklos-da, line 384: Same here as above. I.e., `KeyValueConsumption` is called on the read path, hence, it must not override the conclusion made by the committer (and communicated via the rejection entry).

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumptionSpec.scala
```diff
@@ -97,6 +99,11 @@ class KeyValueConsumptionSpec extends AnyWordSpec with Matchers {
           Assertions(update =>
             inside(update) { case Some(CommandRejected(_, _, FinalReason(status))) =>
               status.code shouldBe Code.ALREADY_EXISTS.value
+              status.details shouldBe Seq(
+                AnyProto.pack[ErrorInfo](
+                  ErrorInfo(metadata = Map(GrpcStatuses.DefiniteAnswerKey -> "false"))
+                )
+              )
               ()
```
miklos-da, line 107: It would also make sense to add a test case variant that has `transactionRejectionEntry.definiteAnswer = true`.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -104,6 +104,8 @@ private[kvutils] class TransactionCommitter(
             rejections.reject(
               DamlTransactionRejectionEntry.newBuilder
                 .setSubmitterInfo(transactionEntry.submitterInfo)
+                // the definite answer is false, as the rank-based deduplication is not yet implemented
```
miklos-da, line 107: I think the below may grasp better the reason for returning a non-definite answer:
```suggestion
                // No duplicate rejection is a definite answer as the deduplication entry will eventually expire.
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -367,6 +369,8 @@ private[state] object Conversions {
     val builder = DamlTransactionRejectionEntry.newBuilder
     builder
       .setSubmitterInfo(submitterInfo)
+      // so far it's false for all rejections before the rank-based deduplication is implemented
```
miklos-da, line 372:
```suggestion
```

miklos-da (changes requested):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
hubert-da: You're right, no idea why I missed it.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -194,6 +194,12 @@ message DamlTransactionRejectionEntry {
   DamlSubmitterInfo submitter_info = 1;
   reserved 5; // was reason.maximum_record_time_exceeded
 
+  // Whether this rejection should be considered a definite answer; i.e., whether this rejection
+  // signals that no other transaction with the same change-id can be accepted or rejected.
+  //
+  // False for all rejections before the rank-based deduplication is implemented.
```
miklos-da, line 200: I'd rather not reference future work here.
```suggestion
```

miklos-da (approved):

PR #10804, opened 2021-09-07
Ledger API test tool: ensure that test identifiers are unique

fabiotudone-da:
## Description

Test identifiers should be unique because they are used to generate various IDs (e.g. party names, command IDs) that should not clash.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
+++ b/ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
@@ -26,7 +26,7 @@ class NamesSpec extends AnyWordSpec with Matchers {
       all(allTestIdentifiers) should fullyMatch regex """[A-Za-z][A-Za-z0-9]*""".r
     }
 
-    "not be a prefix of any other name, so that each test can be included independently" in {
+    "not be a prefix of or equal to any other name, so that each test can be included independently" in {
       allTestIdentifiers.zipWithIndex.foreach { case (testIdentifier, i) =>
         all(allTestIdentifiers.patch(i, Nil, 1)) should not startWith testIdentifier
       }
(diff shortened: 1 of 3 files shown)


ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
```diff
@@ -27,8 +27,8 @@ class NamesSpec extends AnyWordSpec with Matchers {
     }
 
     "not be a prefix of any other name, so that each test can be included independently" in {
```
fabiotudone-da, line 29:
```suggestion
    "not be a prefix of or equal to any other name, so that each test can be included independently" in {
```
```suggestion
    "not be a prefix of any other name, so that each test can be included independently" in {
```

ledger/ledger-api-test-tool/src/test/suite/scala/com/daml/ledger/api/testtool/NamesSpec.scala
```diff
@@ -26,9 +26,9 @@ class NamesSpec extends AnyWordSpec with Matchers {
       all(allTestIdentifiers) should fullyMatch regex """[A-Za-z][A-Za-z0-9]*""".r
     }
 
-    "not be a prefix of any other name, so that each test can be included independently" in {
-      allTestIdentifiers.foreach { testIdentifier =>
-        all(allTestIdentifiers.toSet - testIdentifier) should not startWith testIdentifier
+    "not be a prefix of or equal to any other name, so that each test can be included independently" in {
+      allTestIdentifiers.zipWithIndex.foreach { case (testIdentifier, i) =>
+        all(allTestIdentifiers.patch(i, Nil, 1)) should not startWith testIdentifier
       }
```
nicu-da, line 32: Really good catch, I remember reviewing this but really missed the logic 🤦 . I guess that's why tests should always be red first 😆 

I think this is quite difficult to read, especially as patch is not really that familiar. 
Keeping the zip with index, if we keep only the elements from index for `allTestIdentifiers` and then apply the assertios it should be a bit clearer. As it makes no sense to assert again for any element that has an index < currentIndex. So I guess replace patch with a drop

fabiotudone-da: 1216af9

hubert-da (approved):

nicu-da (approved):

Issue #10266, opened 2021-07-14
Explore surrogate template id cache in JSON API


cocreature:
We currently make a lot of queries to map from template ids to the surrogate template ids in the database. While those queries are pretty quick, not doing a database query is usually still faster than doing one.

Maybe worth exploring the costs we pay for the query and consider adding an in-memory cache. Given that we only ever add, the cache should be easy to manage since there is no complex invalidation or mutation.

akshayshirahatti-da:
Preliminary results for this 


- With Gatling multi-user performance tests for oracle with a single template but about 10k queries across 100 parallel users, we some some improvement in throughput and latency in the higher percentiles.


```
No Surrogate template id cache (baseline)
================================================================================
---- SyncFetchByQuery ----------------------------------------------------------
> Number of requests                                 10000 (OK=10000  KO=-     )
> Min. response time                                    23 (OK=23     KO=-     )
> Max. response time                                   701 (OK=701    KO=-     )
> Mean response time                                   125 (OK=125    KO=-     )
> Std. deviation                                        41 (OK=41     KO=-     )
> response time 90th percentile                        125 (OK=125    KO=-     )
> response time 95th percentile                        148 (OK=148    KO=-     )
> response time 99th percentile                        188 (OK=188    KO=-     )
> response time 99.9th percentile                      221 (OK=221    KO=-     )
> Mean requests/second                               78.74 (OK=78.74  KO=-     )
---- Response time distribution ------------------------------------------------
> t < 800 ms                                         10000 (100%)
> 800 ms < t < 1200 ms                                   0 (  -%)
> t > 1200 ms                                            0 (  -%)
> failed                                                 0 (  -%)
================================================================================



With Surrogate template id cache (regression)
================================================================================
---- SyncFetchByQuery ----------------------------------------------------------
> Number of requests                                 10000 (OK=10000  KO=-     )
> Min. response time                                    21 (OK=21     KO=-     )
> Max. response time                                   633 (OK=633    KO=-     )
> Mean response time                                   106 (OK=106    KO=-     )
> Std. deviation                                        34 (OK=34     KO=-     )
> response time 90th percentile                        106 (OK=106    KO=-     )
> response time 95th percentile                        125 (OK=125    KO=-     )
> response time 99th percentile                        156 (OK=156    KO=-     )
> response time 99.9th percentile                      186 (OK=186    KO=-     )
> Mean requests/second                              93.458 (OK=93.458 KO=-     )
[...]

PR #10806, opened 2021-09-07
[JSON-API] surrogate template id cache

akshayshirahatti-da:
Initial changes to add a surrogate_template_id cache to reduce db queries ([#10266](https://github.com/digital-asset/daml/issues/10266))

CHANGELOG_BEGIN
CHANGELOG_END

Preliminary results for this 

- With Gatling multi-user performance tests for oracle with a single template but about 10k queries across 100 parallel users, we see some improvement in throughput and latency in the higher percentiles.

```
No Surrogate template id cache (baseline)
================================================================================
---- SyncFetchByQuery ----------------------------------------------------------
> Number of requests                                 10000 (OK=10000  KO=-     )
> Min. response time                                    23 (OK=23     KO=-     )
> Max. response time                                   701 (OK=701    KO=-     )
> Mean response time                                   125 (OK=125    KO=-     )
> Std. deviation                                        41 (OK=41     KO=-     )
> response time 90th percentile                        125 (OK=125    KO=-     )
> response time 95th percentile                        148 (OK=148    KO=-     )
> response time 99th percentile                        188 (OK=188    KO=-     )
> response time 99.9th percentile                      221 (OK=221    KO=-     )
> Mean requests/second                               78.74 (OK=78.74  KO=-     )
---- Response time distribution ------------------------------------------------
> t < 800 ms                                         10000 (100%)
> 800 ms < t < 1200 ms                                   0 (  -%)
> t > 1200 ms                                            0 (  -%)
> failed                                                 0 (  -%)
================================================================================



With Surrogate template id cache (regression)
================================================================================
---- SyncFetchByQuery ----------------------------------------------------------
> Number of requests                                 10000 (OK=10000  KO=-     )
> Min. response time                                    21 (OK=21     KO=-     )
> Max. response time                                   633 (OK=633    KO=-     )
> Mean response time                                   106 (OK=106    KO=-     )
> Std. deviation                                        34 (OK=34     KO=-     )
> response time 90th percentile                        106 (OK=106    KO=-     )
> response time 95th percentile                        125 (OK=125    KO=-     )
> response time 99th percentile                        156 (OK=156    KO=-     )
> response time 99.9th percentile                      186 (OK=186    KO=-     )
> Mean requests/second                              93.458 (OK=93.458 KO=-     )
---- Response time distribution ------------------------------------------------
> t < 800 ms                                         10000 (100%)
> 800 ms < t < 1200 ms                                   0 (  -%)
> t > 1200 ms                                            0 (  -%)
> failed                                                 0 (  -%)
================================================================================
```

- with JMH `QueryBenchmark` (i.e `ContractDaoBenchmarks`) we didn't see any significant difference, even with increased templates.
```
==> /var/tmp/withCacheBench.log (baseline) <==
Result "com.daml.http.dbbackend.QueryBenchmark.run":
  5.514 ±(99.9%) 0.494 ops/s [Average]
  (min, avg, max) = (4.900, 5.514, 5.859), stdev = 0.327
  CI (99.9%): [5.020, 6.008] (assumes normal distribution)


# Run complete. Total time: 00:03:32

Benchmark           (batchSize)  (extraParties)  (extraTemplates)   Mode  Cnt  Score   Error  Units
QueryBenchmark.run         1000               1                 1  thrpt   10  7.350 ± 0.804  ops/s
QueryBenchmark.run         1000               1                10  thrpt   10  6.478 ± 0.377  ops/s
QueryBenchmark.run         1000               1                20  thrpt   10  5.514 ± 0.494  ops/s

==> /var/tmp/noCacheBench.log (regression)<==
Result "com.daml.http.dbbackend.QueryBenchmark.run":
  5.580 ±(99.9%) 0.296 ops/s [Average]
  (min, avg, max) = (5.217, 5.580, 5.811), stdev = 0.196
  CI (99.9%): [5.284, 5.876] (assumes normal distribution)


# Run complete. Total time: 00:03:28

Benchmark           (batchSize)  (extraParties)  (extraTemplates)   Mode  Cnt  Score   Error  Units
QueryBenchmark.run         1000               1                 1  thrpt   10  7.539 ± 0.326  ops/s
QueryBenchmark.run         1000               1                10  thrpt   10  6.480 ± 0.201  ops/s
QueryBenchmark.run         1000               1                20  thrpt   10  5.580 ± 0.296  ops/s
```

I haven't looked at the QueryBenchmarks in detail and the reason for the insignificant difference but from the multi-user performance tests there seems to be some benefits seen with the cache (also this benefit seems logical)

 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -32,13 +32,15 @@ import com.daml.logging.LoggingContextOf
 import com.daml.metrics.Metrics
 import doobie.free.connection
 
-sealed abstract class Queries(tablePrefix: String)(implicit metrics: Metrics) {
+sealed abstract class Queries(tablePrefix: String, tpIdCacheMaxEntries: Long)(implicit
+    metrics: Metrics
+) {
   import Queries.{Implicits => _, _}, InitDdl._
   import Queries.Implicits._
 
   val schemaVersion = 2
 
-  private[http] val surrogateTpIdCache = new SurrogateTemplateIdCache(metrics)
+  private[http] val surrogateTpIdCache = new SurrogateTemplateIdCache(metrics, tpIdCacheMaxEntries)
 
   protected[this] def dropIfExists(drop: Droppable): Fragment
 
@@ -613,10 +615,10 @@ object Queries {
   }
 }
 
-private final class PostgresQueries(tablePrefix: String)(implicit
+private final class PostgresQueries(tablePrefix: String, tpIdCacheMaxEntries: Long)(implicit
     ipol: Queries.SqlInterpolation.StringArray,
     metrics: Metrics,
-) extends Queries(tablePrefix) {
+) extends Queries(tablePrefix, tpIdCacheMaxEntries) {
   import Queries._, Queries.InitDdl.{Droppable, CreateIndex}
   import Implicits._
 
@@ -746,8 +748,9 @@ import OracleQueries.DisableContractPayloadIndexing
 private final class OracleQueries(
     tablePrefix: String,
     disableContractPayloadIndexing: DisableContractPayloadIndexing,
+    tpIdCacheMaxEntries: Long,
 )(implicit metrics: Metrics)
-    extends Queries(tablePrefix) {
+    extends Queries(tablePrefix, tpIdCacheMaxEntries) {
   import Queries._, InitDdl._
   import Implicits._
 
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -26,14 +26,19 @@ import cats.instances.list._
 import cats.Applicative
 import cats.syntax.applicative._
 import cats.syntax.functor._
+import com.daml.http.util.Logging.InstanceUUID
 import com.daml.lf.crypto.Hash
+import com.daml.logging.LoggingContextOf
+import com.daml.metrics.Metrics
 import doobie.free.connection
 
-sealed abstract class Queries(tablePrefix: String) extends SurrogateTemplateIdCache {
+sealed abstract class Queries(tablePrefix: String)(implicit metrics: Metrics) {
   import Queries.{Implicits => _, _}, InitDdl._
   import Queries.Implicits._
 
   val schemaVersion = 2
+
+  private[http] val surrogateTpIdCache = new SurrogateTemplateIdCache(metrics)
 
   protected[this] def dropIfExists(drop: Droppable): Fragment
 
@@ -161,21 +166,23 @@ sealed abstract class Queries(tablePrefix: String) extends SurrogateTemplateIdCa
   protected[http] def version()(implicit log: LogHandler): ConnectionIO[Option[Int]]
 
   final def surrogateTemplateId(packageId: String, moduleName: String, entityName: String)(implicit
-      log: LogHandler
+      log: LogHandler,
+      lc: LoggingContextOf[InstanceUUID],
   ): ConnectionIO[SurrogateTpId] = {
-    tpIdCachedValue(packageId, moduleName, entityName)
+    surrogateTpIdCache
+      .getCacheValue(packageId, moduleName, entityName)
       .map { tpId =>
         Applicative[ConnectionIO].pure(tpId)
       }
       .getOrElse {
-        surrogateTemplateIdDb(packageId, moduleName, entityName).map { tpId =>
-          setTpIdCacheValue(packageId, moduleName, entityName, tpId)
+        surrogateTemplateIdFromDb(packageId, moduleName, entityName).map { tpId =>
+          surrogateTpIdCache.setCacheValue(packageId, moduleName, entityName, tpId)
           tpId
         }
       }
   }
 
-  private def surrogateTemplateIdDb(packageId: String, moduleName: String, entityName: String)(
+  private def surrogateTemplateIdFromDb(packageId: String, moduleName: String, entityName: String)(
       implicit log: LogHandler
   ): ConnectionIO[SurrogateTpId] =
     sql"""SELECT tpid FROM $templateIdTableName
@@ -607,7 +614,8 @@ object Queries {
 }
 
 private final class PostgresQueries(tablePrefix: String)(implicit
-    ipol: Queries.SqlInterpolation.StringArray
+    ipol: Queries.SqlInterpolation.StringArray,
+    metrics: Metrics,
 ) extends Queries(tablePrefix) {
   import Queries._, Queries.InitDdl.{Droppable, CreateIndex}
   import Implicits._
@@ -738,7 +746,8 @@ import OracleQueries.DisableContractPayloadIndexing
 private final class OracleQueries(
     tablePrefix: String,
     disableContractPayloadIndexing: DisableContractPayloadIndexing,
-) extends Queries(tablePrefix) {
+)(implicit metrics: Metrics)
+    extends Queries(tablePrefix) {
   import Queries._, InitDdl._
   import Implicits._
 
--- a/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
+++ b/ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
@@ -29,7 +29,7 @@ import cats.syntax.functor._
 import com.daml.lf.crypto.Hash
 import doobie.free.connection
 
-sealed abstract class Queries(tablePrefix: String) {
+sealed abstract class Queries(tablePrefix: String) extends SurrogateTemplateIdCache {
   import Queries.{Implicits => _, _}, InitDdl._
   import Queries.Implicits._
 
@@ -162,6 +162,21 @@ sealed abstract class Queries(tablePrefix: String) {
 
   final def surrogateTemplateId(packageId: String, moduleName: String, entityName: String)(implicit
       log: LogHandler
+  ): ConnectionIO[SurrogateTpId] = {
+    tpIdCachedValue(packageId, moduleName, entityName)
+      .map { tpId =>
+        Applicative[ConnectionIO].pure(tpId)
+      }
+      .getOrElse {
+        surrogateTemplateIdDb(packageId, moduleName, entityName).map { tpId =>
+          setTpIdCacheValue(packageId, moduleName, entityName, tpId)
+          tpId
+        }
+      }
+  }
+
+  private def surrogateTemplateIdDb(packageId: String, moduleName: String, entityName: String)(
+      implicit log: LogHandler
   ): ConnectionIO[SurrogateTpId] =
     sql"""SELECT tpid FROM $templateIdTableName
           WHERE (package_id = $packageId AND template_module_name = $moduleName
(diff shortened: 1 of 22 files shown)


ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -162,6 +162,21 @@ sealed abstract class Queries(tablePrefix: String) {
...
+      }
+      .getOrElse {
+        surrogateTemplateIdDb(packageId, moduleName, entityName).map { tpId =>
+          setTpIdCacheValue(packageId, moduleName, entityName, tpId)
+          tpId
+        }
+      }
+  }
+
+  private def surrogateTemplateIdDb(packageId: String, moduleName: String, entityName: String)(
```
akshayshirahatti-da, line 178: need a better name for this function :-/

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+  ) = {
+    val key = s"$packageId-$moduleName-$entityName"
+    logger.trace(s"Set cached value for key ($key) : $tpId")
+    underlying.put(key, SurrogateTpId.unwrap(tpId))
+  }
+
+  //for testing purposes.
+  import metrics.daml.HttpJsonApi.{surrogateTemplateIdCache => cacheStats}
+  final def tpIdCacheHits = cacheStats.hitCount.getCount
+  final def tpIdCacheMiss = cacheStats.missCount.getCount
```
akshayshirahatti-da, line 50: `tpIdCacheMisses`

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import com.codahale.metrics.MetricRegistry
+import com.daml.caching.SizedCache
+import com.daml.http.dbbackend.Queries.SurrogateTpId
+import com.daml.metrics.Metrics
+import org.slf4j.LoggerFactory
+
+object SurrogateTemplateIdCache {
+  private val logger = LoggerFactory.getLogger(classOf[SurrogateTemplateIdCache])
+  private val maxSize = 1000L
```
akshayshirahatti-da, line 14: open to suggestions on what maxSize should be in terms of how many templates we expect

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+}
+
+trait SurrogateTemplateIdCache {
+  import SurrogateTemplateIdCache.{logger, maxSize}
+
+  private val metrics = new Metrics(new MetricRegistry)
+
+  private val underlying = {
+    SizedCache.from[String, java.lang.Long](
```
S11001001, line 23: Why not...?

```suggestion
    SizedCache.from[String, SurrogateTpId](
```

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
stefanobaghino-da: I think this is probably something you want to allow to configure.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+package com.daml.http.dbbackend
+
+import com.codahale.metrics.MetricRegistry
+import com.daml.caching.SizedCache
+import com.daml.http.dbbackend.Queries.SurrogateTpId
+import com.daml.metrics.Metrics
+import org.slf4j.LoggerFactory
+
+object SurrogateTemplateIdCache {
+  private val logger = LoggerFactory.getLogger(classOf[SurrogateTemplateIdCache])
```
stefanobaghino-da, line 13: Any reason not to use `ContextualizedLogger` as in the rest of the JSON API?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+  private val maxSize = 1000L
+}
+
+trait SurrogateTemplateIdCache {
+  import SurrogateTemplateIdCache.{logger, maxSize}
+
+  private val metrics = new Metrics(new MetricRegistry)
```
stefanobaghino-da, line 20: I believe the JSON API already has metrics, we may want to integrate into that.

stefanobaghino-da: As a general suggestion, I would recommend against having `val`s in `trait`s to prevent initialization order issues.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
akshayshirahatti-da: `SizedCache` actually uses Caffeine cache underneath which relies on `AnyRef` :(
```
object CaffeineCache {
  def apply[Key <: AnyRef, Value <: AnyRef](
       builder: caffeine.Caffeine[_ >: Key, _ >: Value],
       metrics: Option[CacheMetrics],
    ): ConcurrentCache[Key, Value] = ???
  }
  
```

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
stefanobaghino-da, line 50: If these are exposed only for testing they should be package private.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.daml.http.dbbackend.Queries.SurrogateTpId
+import com.daml.metrics.Metrics
+import org.slf4j.LoggerFactory
+
+object SurrogateTemplateIdCache {
+  private val logger = LoggerFactory.getLogger(classOf[SurrogateTemplateIdCache])
+  private val maxSize = 1000L
+}
+
+trait SurrogateTemplateIdCache {
```
stefanobaghino-da, line 17: I think this is better served by becoming a class that can be parameterized to accept a `Metrics` object and a `LoggingContext`. Then you can instantiate it in the `Queries` object.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
stefanobaghino-da:
```suggestion
  private def surrogateTemplateIdFromDb(packageId: String, moduleName: String, entityName: String)(
```
I don't know, maybe?

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/Queries.scala
```diff
@@ -162,6 +162,21 @@ sealed abstract class Queries(tablePrefix: String) {
 
   final def surrogateTemplateId(packageId: String, moduleName: String, entityName: String)(implicit
```
stefanobaghino-da, line 163: This is the point where you probably want to inject the `LoggingContext`.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
akshayshirahatti-da: no reason at all just missed it , will change to have `ContextualizedLogger`

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
akshayshirahatti-da: I can make it an env variable (although not sure what general opinion on having env variables sprinkled around in code is) but in either case it would still be good to have a sensible default

stefanobaghino-da: Sure. Regarding the environment variable, at this level I would recommend you just accept a parameter for the class to be instantiated and thread this through the JSON API class that instantiates this. Right now we get all of our configuration parameters from the CLI and we should stick to that. Regarding the sensible default, what is the measurement unit? Bytes, entries, something else?

akshayshirahatti-da: the unit is `entries` , so eviction will kick in after 1000 entries currently which maybe a tad high.

stefanobaghino-da: It depends a lot on the size of the entries and the ledger setup, we can have 1000 very small packages or 10 huge ones, or a mix of both. I would say that an acceptable cautious value to provide as a default could be 10. Do you think it would be feasible to replace the `SizedCache` with a `WeightedCache` ([code](https://github.com/digital-asset/daml/blob/648acf52f8fa7a82cbb1d4523ed4ca384f2e1e32/ledger/caching/src/main/scala/com/daml/caching/WeightedCache.scala)) and give the user control over the maximum size in bytes?

akshayshirahatti-da: ok just read this , I'd like to keep it as `SizedCache` for now I've lowered the num entries to `10` I mean worse case is we are dealing with more evictions and the value should now be configurable now. Happy to re-visit this when we have more information.

ledger-service/db-backend/src/main/scala/com/digitalasset/http/dbbackend/SurrogateTemplateIdCache.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import com.daml.caching.SizedCache
+import com.daml.http.dbbackend.Queries.SurrogateTpId
+import com.daml.http.util.Logging.InstanceUUID
+import com.daml.logging.{ContextualizedLogger, LoggingContextOf}
+import com.daml.metrics.Metrics
+
+object SurrogateTemplateIdCache {
+  private val logger = ContextualizedLogger.get(getClass)
+  final val MAX_ENTRIES = 10L
```
stefanobaghino-da, line 14:
```suggestion
  final val MaxEntries = 10L
```
We usually use this format for constants.

stefanobaghino-da (approved):
LGTM, just one nitpick. Great work, thanks!

PR #10811, opened 2021-09-08
Add daml interfaces to the Haskell LF AST

cocreature:
Just copied from Sofia’s PR with no changes and stubbed all usages of
it that aren’t trivial.

part of #10670

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Base.hs
+++ b/compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Base.hs
@@ -863,7 +863,7 @@ data DefException = DefException
 data DefInterface = DefInterface
   { intLocation :: !(Maybe SourceLoc)
   , intName :: !TypeConName
-  , intChoices :: ![InterfaceChoice]
+  , intChoices :: !(NM.NameMap InterfaceChoice)
   }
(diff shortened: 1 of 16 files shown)


compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Base.hs
```diff
@@ -857,6 +860,22 @@ data DefException = DefException
   }
   deriving (Eq, Data, Generic, NFData, Show)
 
+data DefInterface = DefInterface
+  { intLocation :: !(Maybe SourceLoc)
+  , intName :: !TypeConName
+  , intChoices :: ![InterfaceChoice]
```
robin-da, line 866: I think a `NameMap` would be the better choice here, which would be also in line with the `Template` data type.

robin-da (approved):
Looks good, I'd like to change the type of choices in the interface definition to `NameMap` for ease of use.

compiler/daml-lf-ast/src/DA/Daml/LF/Ast/Base.hs
cocreature: good idea e95de5f0d0ed2d0525b72d784807b14776f4473e

PR #10814, opened 2021-09-08
Stop using `controller … can` syntax in ledger tests

cocreature:
We’re planning to deprecate that in SDK 2.0 to reduce the confusion
around the implicit observer behavior so in preparation for that, this
PR drops the syntax from all ledger tests.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/test-common/src/main/daml/model/Test.daml
+++ b/ledger/test-common/src/main/daml/model/Test.daml
@@ -273,8 +273,9 @@ template BranchingControllers
     ctrlFalse : Party
   where
   signatory giver
-  observer ctrl
   let ctrl = if whichCtrl then ctrlTrue else ctrlFalse
+  observer ctrl
+
   choice Delete : ()
     controller ctrl
     do return ()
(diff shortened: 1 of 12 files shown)


ledger/test-common/src/main/daml/model/Test.daml
```diff
@@ -278,9 +273,11 @@ template BranchingControllers
     ctrlFalse : Party
   where
   signatory giver
-  controller (if whichCtrl then ctrlTrue else ctrlFalse) can
-    Delete : ()
-      do return ()
+  observer ctrl
+  let ctrl = if whichCtrl then ctrlTrue else ctrlFalse
```
nickchapman-da, line 277: Is it syntactically valid to have the `let ctrl =` line above the `observer` line?

nickchapman-da (approved):
Looks great. Again, I can't see any mistakes.

I had one tiny comment on one example.

Also, did you consider using choice observers instead of template observers for the code transformation?

ledger/test-common/src/main/daml/model/Test.daml
cocreature: yes moved it around

nickchapman-da:
> Looks great. Again, I can't see any mistakes.
> 
> I had one tiny comment on one example.
> 
> Also, did you consider using choice observers instead of template observers for the code transformation?

Ignore that. THat of course has a different semantics

PR #10821, opened 2021-09-08
Sandbox on H2 - performance improvements for the append-only schema [DPP-600]

kamil-da:
### Motivation
Performance degradation was observed when running the `sandbox` with H2 and the append-only schema.

### Changes
- added a CLI parameter for configuring the database connection pool size in `sandbox`
- removed reliance on the `participant_events` view and used `participant_events_*` tables directly in queries that proved to have the biggest impact on the performance when H2 with append-only schema is used

### Important
**This PR changes queries for all db types.**

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
@@ -268,7 +268,9 @@ trait EventStorageBackend {
       transactionId: Ref.TransactionId,
       filterParams: FilterParams,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.TreeEvent]]
-  def maxEventSeqIdForOffset(offset: Offset)(connection: Connection): Option[Long]
+
+  /** Max event sequential id of observable (create, consuming and nonconsuming exercise) events. */
+  def maxEventSequentialIdOfAnObservableEvent(offset: Offset)(connection: Connection): Option[Long]
   def rawEvents(startExclusive: Long, endInclusive: Long)(
       connection: Connection
   ): Vector[RawTransactionEvent]
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
@@ -64,26 +64,23 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
         import com.daml.platform.store.Conversions.ContractIdToStatement
         SQL"""
   WITH archival_event AS (
-         SELECT participant_events.*
-           FROM participant_events, parameters
+         SELECT 1
+           FROM participant_events_consuming_exercise, parameters
           WHERE contract_id = $id
-            AND event_kind = 20  -- consuming exercise
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           FETCH NEXT 1 ROW ONLY
        ),
        create_event AS (
          SELECT ledger_effective_time
-           FROM participant_events, parameters
+           FROM participant_events_create, parameters
           WHERE contract_id = $id
-            AND event_kind = 10  -- create
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           FETCH NEXT 1 ROW ONLY -- limit here to guide planner wrt expected number of results
        ),
        divulged_contract AS (
-         SELECT ledger_effective_time
-           FROM participant_events, parameters
+         SELECT null
+           FROM participant_events_divulgence, parameters
           WHERE contract_id = $id
-            AND event_kind = 0 -- divulgence
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           ORDER BY event_sequential_id
             -- prudent engineering: make results more stable by preferring earlier divulgence events
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
@@ -52,6 +52,8 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
       s"The following contracts have not been found: ${missingContractIds.map(_.coid).mkString(", ")}"
     )
 
+  protected val CastNullLedgerEffectiveTime: String = "NULL::BIGINT"
+
   // TODO append-only: revisit this approach when doing cleanup, so we can decide if it is enough or not.
   // TODO append-only: consider pulling up traversal logic to upper layer
   override def maximumLedgerTime(
@@ -78,7 +80,7 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
           FETCH NEXT 1 ROW ONLY -- limit here to guide planner wrt expected number of results
        ),
        divulged_contract AS (
-         SELECT NULL::BIGINT
+         SELECT #$CastNullLedgerEffectiveTime
            FROM participant_events_divulgence, parameters
           WHERE contract_id = $id
             AND event_sequential_id <= parameters.ledger_end_sequential_id
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
@@ -64,26 +64,23 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
         import com.daml.platform.store.Conversions.ContractIdToStatement
         SQL"""
   WITH archival_event AS (
-         SELECT participant_events.*
-           FROM participant_events, parameters
+         SELECT 1
+           FROM participant_events_consuming_exercise, parameters
           WHERE contract_id = $id
-            AND event_kind = 20  -- consuming exercise
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           FETCH NEXT 1 ROW ONLY
        ),
        create_event AS (
          SELECT ledger_effective_time
-           FROM participant_events, parameters
+           FROM participant_events_create, parameters
           WHERE contract_id = $id
-            AND event_kind = 10  -- create
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           FETCH NEXT 1 ROW ONLY -- limit here to guide planner wrt expected number of results
        ),
        divulged_contract AS (
-         SELECT ledger_effective_time
-           FROM participant_events, parameters
+         SELECT NULL::BIGINT
+           FROM participant_events_divulgence, parameters
           WHERE contract_id = $id
-            AND event_kind = 0 -- divulgence
             AND event_sequential_id <= parameters.ledger_end_sequential_id
           ORDER BY event_sequential_id
             -- prudent engineering: make results more stable by preferring earlier divulgence events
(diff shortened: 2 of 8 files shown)


ledger/sandbox/src/main/scala/platform/sandboxnext/Cli.scala
```diff
@@ -36,6 +36,15 @@ private[sandboxnext] object Cli extends SandboxCli {
         s"Deprecated: Use the Daml Driver for PostgreSQL if you need persistence.\nThe JDBC connection URL to a Postgres database containing the username and password as well. If present, $Name will use the database to persist its data."
       )
       .action((url, config) => config.copy(jdbcUrl = Some(url)))
+
+    parser
```
nmarton-da, line 40: do we need this configurable?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -268,7 +268,9 @@ trait EventStorageBackend {
       transactionId: Ref.TransactionId,
       filterParams: FilterParams,
   )(connection: Connection): Vector[EventsTable.Entry[Raw.TreeEvent]]
-  def maxEventSeqIdForOffset(offset: Offset)(connection: Connection): Option[Long]
+
+  /** Max event sequential id of observable (create, consuming and nonconsuming exercise) events. */
+  def maxEventSequentialIdOfAnObservableEvent(offset: Offset)(connection: Connection): Option[Long]
```
nmarton-da, line 273: why this rename, and the restrictive enumeration of event types here?
(more importantly: why we seem to have a matching implementation for H2, but kept the original for other)

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -105,14 +109,24 @@ private[backend] object H2StorageBackend
   override def insertBatch(connection: Connection, batch: AppendOnlySchema.Batch): Unit =
     H2Schema.schema.executeUpdate(batch, connection)
 
-  // TODO FIXME: this is for postgres not for H2
-  def maxEventSeqIdForOffset(offset: Offset)(connection: Connection): Option[Long] = {
+  def maxEventSequentialIdOfAnObservableEvent(
+      offset: Offset
+  )(connection: Connection): Option[Long] = {
     import com.daml.platform.store.Conversions.OffsetToStatement
     // This query could be: "select max(event_sequential_id) from participant_events where event_offset <= ${range.endInclusive}"
```
nmarton-da, line 116: probably this comment is not valid anymore. how about removing?

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -238,4 +252,133 @@ private[backend] object H2StorageBackend
...
+      if (foundLedgerEffectiveTimes.size != ids.size) {
+        val missingIds = queriedIds.collect { case (missingId, None) =>
+          missingId
+        }
+        Failure(notFound(missingIds.toSet))
+      } else Success(foundLedgerEffectiveTimes.max)
+    }
+  }
+
+  override protected def activeContract[T](
```
nmarton-da, line 316: this is a creative approach to override indeed, but I would like to convince you to take another because:
* not exactly DRY, quite some copy paste takes place
* in case of activeContract it overrides an intricate part of the contract implementation. I think this is quite dangereous
* you can think of inheritance here as a side effect, not the goal, it is already on the table to switch to composition in StorageBackend

therefore I would suggest to organize the ContractStorageBackend, that it has template methods defining the sql literals, and override those only in H2, and not overriding maximumLedgerTime and especially activeContract (also keeping private stuff private of course)

happy to elaborate on a call

ledger/sandbox/src/main/scala/platform/sandboxnext/Cli.scala
kamil-da: I am not sure. We have this option configurable in other places and it was considered as an option for performance tuning. On the other hand, I tried changing the pool size for H2 with the append-only schema and the performance didn't improve. I'm open to removing this.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
kamil-da: The rename comes from the fact that this method does not include divulgence events sequential ids when computing the maximum value. In the previous version this was a bit hidden by using `participant_events` view. I think that the name `maxEventSeqIdForOffset()` is misleading in this context.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
kamil-da: I agree, this can definitely be improved. This case is another argument for removing the `participant_events` view at all.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
rautenrieth-da: To explain a bit more:
- Divulgence events have an `event_offset`
- In the view, the `event_offset` of divulgence events is hardcoded to `NULL` (not sure why, looks like a bug to me)
- Divulgence events are not observable over the ledger API. Only write side (command interpretation) queries should look at divulgence events.
- Including divulgence events in the `maxEventSeqIdForOffset` query significantly reduces its performance.

So excluding divulgence events for this query is the correct behavior, and we only get the correct behavior for Postgres/Oracle because of a bug in the view definition.

ledger/sandbox/src/main/scala/platform/sandboxnext/Cli.scala
rautenrieth-da: AFAIK, for the append-only schema, this parameter is not used by the parallel indexer, it's only used by the read side. The description of the property could be improved accordingly.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
```diff
@@ -238,4 +252,133 @@ private[backend] object H2StorageBackend
       pruneAllDivulgedContracts: Boolean,
       connection: Connection,
   ): Unit = ()
+
+  override def maximumLedgerTime(
```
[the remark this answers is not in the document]
kamil-da: I didn't want to touch anything apart from the H2 as I understand this PR as a patch before the actual work but I'll check if changing in `ContractStorageBackendTemplate` is possible.

ledger/sandbox/src/main/scala/platform/sandboxnext/Cli.scala
kamil-da: @rautenrieth-da Do you think that the CLI parameter is useful at all? I'm not sure if I should keep it.

rautenrieth-da: Exposing the parameter could be useful for future performance tests. Maybe change the description to 
```
The number of connections in the database connection pool used for serving ledger API requests
```
because it's only used for the Index but not the Indexer in sandbox.

kamil-da: Fixed.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
kamil-da: Fixed.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
kamil-da: I've changed the approach and have modified queries on the common level instead of modifying them specifically for H2. This removes the issues you mentioned.

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2StorageBackend.scala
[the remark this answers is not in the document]
kamil-da: I've modified queries in the `ContractStorageBackendTemplate` instead of modifying them in `H2StorageBackend`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -52,6 +52,8 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
       s"The following contracts have not been found: ${missingContractIds.map(_.coid).mkString(", ")}"
     )
 
+  protected val CastNullLedgerEffectiveTime: String = "NULL::BIGINT"
```
kamil-da, line 55: @nmarton-da What do you think about defining this and overriding in subclasses?

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
nmarton-da: I see. Can you please make TODO-s for Oracle and Postgres implementation to make this explicit?

kamil-da: Sure, I'll add that.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -64,26 +66,23 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
         import com.daml.platform.store.Conversions.ContractIdToStatement
         SQL"""
   WITH archival_event AS (
-         SELECT participant_events.*
-           FROM participant_events, parameters
+         SELECT 1
+           FROM participant_events_consuming_exercise, parameters
```
nmarton-da, line 70: Wasn't the original motivation to limit changes to H2? This changes for all...

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
nmarton-da: how about just adding to querystrategy with some more general name like nullAsBigint?

ledger/sandbox/src/main/scala/platform/sandboxnext/Cli.scala
nmarton-da: just a side not, I do not care much: but this was not helping at all, I would not add if it is not needed

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
kamil-da: I may have prematurely moved this to the common class following discussions in this PR. After a second thought I would go for having changes only for H2, especially that it's close to a new release.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
kamil-da: Done.

kamil-da:
There's an alternative version of this PR that only modifies H2 queries: #10888

kamil-da:
Closing the PR because #10888 is going to be merged instead to avoid impacting any other db than H2. Changes for other db types should be done in the future.

PR #10825, opened 2021-09-09
make auth middleware aware that actAs => readAs

garyverhaegen-da:
It seems a bit silly that the auth middleware would refuse a token that
has an `actAs` when it needs a `readAs`. (Unless I'm misunderstanding
and the difference between actAs and readAs is actually non-empty?)

CHANGELOG_BEGIN
- The auth middleware will now consider that actAs credentials are valid
  for readAs requests.
CHANGELOG_END


aherrmann-da (approved):
Thanks, that's very sensible. More importantly, it also matches the ledger's scheme described [here](https://docs.daml.com/app-dev/authorization.html#access-tokens-and-claims).

stefanobaghino-da (commented):
🙇🏻

PR #10826, opened 2021-09-09
DPP-463 switch to append only

rautenrieth-da:
This PR changes CI tests of `sandbox` and `sandbox-classic` to run using the append-only schema by default. The mutating schema is only tested using a single test.

Previously, it was the other way round.

This PR is best reviewed commit by commit.


kamil-da (approved):
Looks good!

ledger/sandbox/BUILD.bazel
```diff
@@ -316,47 +319,54 @@ server_conformance_test(
     ],
 )
 
+# This test asserts that the Daml engine is configured correctly for the sandbox executable.
+# It does not depend on the ledger backend, that's why it runs on a single server.
```
tudor-da, line 323: When Sandbox-on-X is ready, we could move the tests that don't depend on the ledger backend there.

tudor-da (approved):
LGTM 👍

ledger/sandbox/BUILD.bazel
rautenrieth-da: Each participant executable needs one of these tests. It checks that the Daml engine configuration (that is hardcoded in the participant application) corresponds to the ledger backend type.

tudor-da: I see. I was assuming that the assertion is static, but indeed I see that `sandbox-classic` does not forbid V0 contract ids. Thanks for clarifying

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -322,23 +322,16 @@ SERVERS = {
...
-    "postgresql": {
-        "binary": ":sandbox-classic-ephemeral-postgresql",
-        "server_args": [
-            "--port=6865",
-            "--eager-package-loading",
-        ],
-    },
+# The in-memory ledger is deprecated, and does not support all operations (e.g., pruning)
+# TODO: remove this after removing the sandbox classic in-memory ledger
+NON_DEPRECATED_SERVERS = {
```
nicu-da, line 327: I think the naming of non-deprecated is quite unusual. I would expect to find DEPRECATED_SERVERS, but the for the other ones I would use a different name. 
I'm a bit out of context but it seems to be specifically connected to the support for append-only schemas, so maybe a name more similiar to `APPEND_ONLY_SUPPORTED_SERVERS`, or `FULLY_SUPPORTED_SERVERS`

rautenrieth-da: :+1:

Context: sandbox has an in-memory ledger that we want to remove as soon as possible. We're not implementing new features (like participant pruning) on it, that's why conformance tests running on the in-memory ledger have to exclude some tests.

PR #10828, opened 2021-09-09
Convert SValue to Value, and normalize, in a single code pass.

nickchapman-da:
Convert `SValue` to `Value`, and _normalize_, in a single code pass

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
@@ -9,7 +9,7 @@ import java.util
 import com.daml.lf.data._
 import com.daml.lf.data.Ref._
 import com.daml.lf.language.Ast._
-import com.daml.lf.transaction.{TransactionVersion, Util}
+import com.daml.lf.transaction.TransactionVersion
 import com.daml.lf.value.Value.ValueArithmeticError
 import com.daml.lf.value.{Value => V}
 import com.daml.nameof.NameOf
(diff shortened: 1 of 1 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -30,13 +30,99 @@ sealed trait SValue {
...
+          V.ValueTextMap(SortedLookupList(entries.map {
+            case (SText(t), v) => t -> go(v, nextMaxNesting)
+            case (_, _) =>
+              throw SError.SErrorCrash(
+                NameOf.qualifiedNameOfCurrentFunc,
+                "SValue.toValue: TextMap with non text key",
+              )
+          }))
+        case SMap(false, entries) =>
+          // TODO: Is it necessary to check GenMap/version here? Can't it just be assumed?
```
cocreature, line 102: afaict it’s only a safe guard. speedy should never produce this for older versions but if we messed up it might catch it. @remyhaemmerle-da cna probably confirm that.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
```diff
@@ -30,13 +30,99 @@ sealed trait SValue {
...
+    toValue(
+      disallowGenMapAtVersion =
+        if (version >= TransactionVersion.minGenMap) None else Some(version),
+      eraseType = version >= TransactionVersion.minTypeErasure,
+    )
+  }
+
+  private def toValue(
+      disallowGenMapAtVersion: Option[TransactionVersion],
+      eraseType: Boolean,
```
cocreature, line 53:
```suggestion
      normalize: Boolean,
```
maybe a bit clearer if we add more normalization steps in the future?

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SValue.scala
remyhaemmerle-da: Yes, I think we can drop the check.

remyhaemmerle-da (approved):
Nice.

PR #10829, opened 2021-09-09
Ledger API Test Tool: support `--additional` tests [KVL-1100]

fabiotudone-da:
## Description

Adds a new, hidden `--additional` command line option to the ledger API test tool. This allows to consolidate the existing conformance test suites and should also improve build time.

The suites simplification commit needs an especially wide review.

### Pull Request Checklist

trigger the build.


nicu-da:
I think this is an amazing addition, really simplifies our life.

The only thing I'm not fully convinced about is the naming of the new options, `add`.
In a perfect world the current behavior of the new flag would actually be `-include` IMO, as it does the opposite of what `--exclude` does. I do believe we should try to have a naming sort of in the same space as the current `--include`, maybe `--include-optional`, `--include-extra`

ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -218,88 +218,38 @@ conformance_test(
...
-conformance_test(
-    name = "deprecated-cli-options",
-    server = "//ledger/ledger-on-memory:app",
-    server_args = [
-        "--contract-id-seeding=testing-weak",
-        "--participant participant-id=ssl-test,port=6865",
-    ],
-    test_tool_args = [
-        "--include=IdentityIT",
-        "--all-tests",
```
mziolekda, line 267: This deprecated flag is no longer tested

fabiotudone-da:
> In a perfect world the current behavior of the new flag would actually be `-include` IMO

Contrary to `exclude`, that can only be interpreted as an exclusion list, `include` is a bit more subject to interpretation IMO, that's why  I am a bit reluctant to use `include` as a prefix for the new parameter, as it could create confusion about the current `include` (now we only support either inclusion or exclusion lists, so it's relatively clear). How about `--additional`?

nicu-da:
> > In a perfect world the current behavior of the new flag would actually be `-include` IMO
> 
> Contrary to `exclude`, that can only be interpreted as an exclusion list, `include` is a bit more subject to interpretation IMO, that's why I am a bit reluctant to use `include` as a prefix for the new parameter, as it could create confusion about the current `include` (now we only support either inclusion or exclusion lists, so it's relatively clear). How about `--additional`?

Agree with this. `additional` sounds good

ledger/ledger-api-test-tool/BUILD.bazel
fabiotudone-da: Nice catch, thanks, fd3b275

ledger/sandbox-classic/BUILD.bazel
```diff
@@ -380,9 +385,6 @@ server_conformance_test(
     servers = ONLY_POSTGRES_SERVER,
     test_tool_args = [
         "--open-world",
-        "--include=ParticipantPruningIT",
```
mziolekda, line 383: These tests have been moved from wall-clock-time to static-time

fabiotudone-da: Right, I should have added them to `conformance-test-wall-clock-append-only` and I should also have removed this suite. Are they supposed to work only with wall clock (I suspect so)?

ledger/sandbox/BUILD.bazel
```diff
@@ -334,29 +336,9 @@ server_conformance_test(
...
-
-server_conformance_test(
-    name = "next-conformance-test-command-deduplication",
-    servers = NEXT_SERVERS,
-    test_tool_args = [
-        "--include=KVCommandDeduplicationIT",
+        "--exclude=ClosedWorldIT",
+        "--exclude=CommandDeduplicationIT",  # It's a KV ledger so it needs the KV variant
+        "--additional=AppendOnlyCompletionDeduplicationInfoITCommandService",
+        "--additional=AppendOnlyCompletionDeduplicationInfoITCommandSubmissionService",
```
mziolekda, line 342: we should add KVCommandDeduplicationIT here

ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -219,6 +210,9 @@ conformance_test(
     test_tool_args = [
         "--verbose",
         "--include=ParticipantPruningIT",
+        # The following tests don't actually care about multi-participant but they do need append-only.
+        "--additional=AppendOnlyCompletionDeduplicationInfoITCommandService",
+        "--additional=AppendOnlyCompletionDeduplicationInfoITCommandSubmissionService",
```
mziolekda, line 215: The test after this one i.e. "conformance-test-multi-participant-command-deduplication" is an append-only test but asserts KVCommandDeduplicationIT

ledger/sandbox/BUILD.bazel
fabiotudone-da: a6d1e7e

ledger/sandbox-classic/BUILD.bazel
fabiotudone-da: 085a15e

ledger/ledger-on-memory/BUILD.bazel
fabiotudone-da: @nicu-da Shouldn't it?

ledger/ledger-on-sql/BUILD.bazel
```diff
@@ -269,7 +246,12 @@ da_scala_test_suite(
             ],
         ),
         conformance_test(
-            name = "conformance-test-participant-pruning-{}".format(db["name"]),
+            name = "conformance-test-{}".format(db["name"]),
+            lf_versions = [
```
mziolekda, line 250: I am assuming that beefing up this test with both more lf versions and test cases was intentional?

fabiotudone-da: This allows unifying pruning-only suite and the conformance test one, that included multiple LF versions.

ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -254,38 +248,9 @@ conformance_test(
...
-    name = "conformance-test-participant-pruning",
-    ports = [6865],
-    server = ":app",
-    server_args = [
-        "--contract-id-seeding=testing-weak",
-        "--participant=participant-id=example,port=6865",
-    ],
-    test_tool_args = [
-        "--verbose",
-        "--include=ParticipantPruningIT",
```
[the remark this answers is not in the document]
fabiotudone-da: 7f8289a46eb7933263c7aafd8aa310a06ed84051

ledger/ledger-on-memory/BUILD.bazel
nicu-da: Yes, we can add it here and remove `conformance-test-multi-participant-command-deduplication`

fabiotudone-da: Removed it for now (also, it was running just that) baa8be4

fabiotudone-da: @nicu-da Can we theoretically run `KVCommandDeduplicationIT` for both append-only and mutable schema? Does it make sense to?

fabiotudone-da: Added, it seems to work 35a274c

mziolekda (approved):
Thanks for all the corrections. I have reviewed it quite thoroughly and am satisfied.
Because this change requires attention to the detail, maybe it would be good to collect one more review before merging.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/LedgerApiTestTool.scala
```diff
@@ -157,8 +157,10 @@ object LedgerApiTestTool {
       if (config.included.isEmpty) defaultCases
       else allCases.filter(matches(config.included))
 
+    val addedTests = allCases.filter(matches(config.additional))
+
     val (excludedTests, testsToRun) =
-      includedTests.partition(matches(config.excluded))
+      (includedTests ++ addedTests).partition(matches(config.excluded))
```
tudor-da, line 163: The behavior might be confusing now. If `--included` is added together with `--additional`, all the `additionals` become practically `--included`s, which defies the intention of this flag. I would disallow using `--included` with `--additional` together. Thus, we would also avoid silent mistakes that would lead to big parts of the test suite disabled.

ledger/ledger-on-memory/BUILD.bazel
```diff
@@ -179,7 +167,9 @@ conformance_test(
         "--daml-lf-min-version-1.14-unsafe",
     ],
     test_tool_args = [
-        "--exclude=CommandDeduplicationIT",  # It's a KV ledger so it needs the KV variant,
+        "--additional=MultiPartySubmissionIT",
```
tudor-da, line 170: Can you please order the test flags? (e.g. first `additional` and then `exclude`).

tudor-da (approved):
Really great improvement, which should also translate into faster CI times 👍 👍 👍

ledger/ledger-on-memory/BUILD.bazel
fabiotudone-da: In this case or in general? I chose to keep (test-wise) related args together even though it's true that exclusions are applied last.

tudor-da: I meant in general. But I see your point as well, so I don't have a strong opinion regarding it.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/LedgerApiTestTool.scala
fabiotudone-da: 3919128

ledger/ledger-on-memory/BUILD.bazel
fabiotudone-da: It was actually possible to do so without separating related tests cd4cf62

pbatko-da (approved):

PR #10832, opened 2021-09-09
Populate `definite_answer` in `ApiException` [KVL-1004]

hubert-da:
Also:
* Fixes bugs mentioned in comments.
* Adds new conformance-test assertions (by @SamirTalwar-DA).
* As errors from rejections have the default "definite_answer", a proper mapping is introduced in a separate PR: https://github.com/digital-asset/daml/pull/10801

### Pull Request Checklist

trigger the build.


ledger-api/rs-grpc-bridge/src/main/java/com/daml/grpc/adapter/server/rs/ServerSubscriber.java
```diff
@@ -114,7 +113,7 @@ public void onError(Throwable throwable) {
     executionSequencer.sequence(
         () -> {
           if (!responseObserver.isCancelled()) {
-            responseObserver.onError(Status.fromThrowable(throwable).asRuntimeException());
```
hubert-da, line 117: This was a bug resulting in dropping the trailers.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -117,20 +117,30 @@ object CompletionResponse {
   }
 
   private def buildException(metadata: Map[String, String], status: StatusJavaProto.Builder) = {
-    val newDetails = status.getDetailsList.asScala.map { detail =>
```
hubert-da, line 120: There was a bug as well, resulting in ignoring the metadata if the details were empty. I added a test to cover the default metadata case.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/ApiException.scala
```diff
@@ -3,9 +3,11 @@
...
 package com.daml.platform.server.api
 
-import io.grpc.{Status, StatusRuntimeException}
+import io.grpc.StatusRuntimeException
 
 import scala.util.control.NoStackTrace
 
 /** The sole purpose of this class is to give StatusRuntimeException with NoStacktrace a nice name in logs. */
-class ApiException(status: Status) extends StatusRuntimeException(status) with NoStackTrace
+class ApiException(exception: StatusRuntimeException)
```
hubert-da, line 11: Weird code, as I haven't found any other way to convert it while preserving the trailers as well as keeping it `NoStackTrace`.

fabiotudone-da (commented):
I couldn't find any `definiteAnswer = Some(true)`, is that intended?

hubert-da:
> I couldn't find any `definiteAnswer = Some(true)`, is that intended?

So far, only the `DuplicateCommandException` has it.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -4,54 +4,204 @@
...
+  lazy val DuplicateCommandException: StatusRuntimeException =
     grpcError(
-      Status.NOT_FOUND.withDescription(
-        s"Ledger ID '${received.unwrap}' not found. Actual Ledger ID is '${expected.unwrap}'."
-      )
+      Status
+        .newBuilder()
+        .setCode(Code.ALREADY_EXISTS.value())
+        .setMessage("Duplicate command")
+        .addDetails(definiteAnswers(true))
```
andreaslochbihler-da, line 26:
```suggestion
        .addDetails(definiteAnswers(false))
```
As explained in https://github.com/digital-asset/daml/pull/10801#discussion_r705956197, we don't persist anything for duplicate commands and therefore that's not a definite answer.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -4,54 +4,204 @@
...
-    grpcError(Status.INVALID_ARGUMENT.withDescription(s"Invalid field $fieldName: $message"))
+  /** @param expected Expected ledger id.
+    * @param received Received ledger id.
+    * @param definiteAnswer A flag that says whether it is a definite answer. Provided only in the context of command deduplication.
+    * @return An exception with the [[Code.NOT_FOUND]] status code.
+    */
+  def ledgerIdMismatch(
+      expected: LedgerId,
+      received: LedgerId,
+      definiteAnswer: Option[Boolean],
```
andreaslochbihler-da, line 47: Minor: I think we can safely assume that a wrong ledger ID can never be a definite answer. So IMO it would make sense to not even offer the possibility to return a `definite_answer = true` field here.

andreaslochbihler-da: Meta-comment: While it makes sense to require a `definite_answer` metadata field on all gRPC status codes other than `OK` for command submissions to facilitate the migration and testing the migration, in the long run, I'd argue that we could say that if there's no definite answer, then that's equivalent to `definite_answer = false`. But that's something for another discussion.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -66,19 +92,34 @@ object Assertions {
...
-      if (exceptionMessageSubString.isEmpty) None
-      else Some(Pattern.compile(Pattern.quote(exceptionMessageSubString))),
-    )
+  private def assertMatches(message: String, pattern: Pattern): Unit = {
+    if (pattern.matcher(message).find()) {
+      ()
+    } else {
+      fail(s"Error message did not contain [$pattern], but was [$message].")
+    }
+  }
```
fabiotudone-da, line 101: Nit: unneeded braces

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -41,21 +44,44 @@ object Assertions {
     }
   }
 
+  /** non-regex overload for assertGrpcError which just does a substring check.
+    */
+  def assertGrpcError(
+      t: Throwable,
+      expectedCode: Status.Code,
+      exceptionMessageSubString: String,
+      checkDefiniteAnswerMetadata: Boolean = false,
```
fabiotudone-da, line 53: How about also allowing to check for the value itself?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -66,19 +92,34 @@ object Assertions {
...
+        val errorInfo = any.unpack(classOf[ErrorInfo])
+        errorInfo.getMetadataMap
+      }
+      .getOrElse {
+        fail(
+          s"The error did not contain a definite answer. Details were: ${details.mkString("[", ", ", "]")}"
+        )
+      }
+    val value = metadata.get("definite_answer")
+    if (value == null) {
```
fabiotudone-da, line 117: Please use `Option()` here

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -41,21 +44,44 @@ object Assertions {
     }
   }
 
+  /** non-regex overload for assertGrpcError which just does a substring check.
+    */
+  def assertGrpcError(
+      t: Throwable,
+      expectedCode: Status.Code,
+      exceptionMessageSubString: String,
```
fabiotudone-da, line 52: I'd rather have it as optional and using `None` in callers when not provided. Also, please name the parameter explicitly in callers.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
hubert-da: It's always `false` so far. However, I wouldn't like to hardcode the assertion to `false`, as Canton may want to set it to `true` somewhere at some point. Do you think it's ok to add `expectedDefiniteAnswerMetadata: Option[Boolean]` and assert it only if it's `Some`?

fabiotudone-da: Yes, that looks like a good solution to me.

fabiotudone-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/CommandDeduplicationBase.scala
```diff
@@ -95,14 +95,14 @@ private[testtool] abstract class CommandDeduplicationBase(
           assertGrpcError(
             failure1,
             Status.Code.ALREADY_EXISTS,
-            "",
-            checkDefiniteAnswerMetadata = true,
+            exceptionMessageSubstring = None,
+            expectedDefiniteAnswerMetadata = Some(false),
```
andreaslochbihler-da, line 99: Whether a completion is a definite answer will vary from ledger to ledger. Canton, e.g., probably sets `true` here. I don't really see how we can reuse these test cases across ledgers if we test for the concrete value.

hubert-da: This was my concern too. Do you want me to revert the last 2 commits?

hubert-da: cc @fabiotudone-da

hubert-da: Could have asked before addressing the review comments :(

hubert-da: Ok, ignore, reverting the assertion changes.

andreaslochbihler-da: Yes, please revert the assertion change that fixes the value of `definite_answer`. Otherwise this test is unusable for Canton.

hubert-da: Done, JFYI enabling the auto-merge.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/ValueValidator.scala
```diff
@@ -47,9 +47,10 @@ object ValueValidator {
     case Sum.ContractId(cId) =>
       ContractId
         .fromString(cId)
-        .bimap(invalidArgument, Lf.ValueContractId(_))
+        .bimap(invalidArgument(definiteAnswer = Some(false)), Lf.ValueContractId(_))
```
nicu-da, line 50: Just for my knowledge, why is this on `false`? Any documentation I would need to read? 
As in my mind this represents a malformed contract id so you can resubmit until the end of ages.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -4,54 +4,205 @@
...
+  )
+
+  private def addDefiniteAnswerDetails(
+      definiteAnswer: Option[Boolean],
+      statusBuilder: Status.Builder,
+  ): Unit = {
+    definiteAnswer.foreach { definiteAnswer =>
+      statusBuilder.addDetails(definiteAnswers(definiteAnswer))
+    }
+  }
```
nicu-da, line 207: I find this a lot more difficult to conceptualize in my mind, I think a simple if inside the foreach would make a lot more sense. This would allow us to also keep the FalseDefinitiveAnswer, TrueDefinitiveAnswer constants.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -86,7 +86,9 @@ conformance_test(
         ",RaceConditionIT:RWArchiveVsLookupByKey,RaceConditionIT:RWArchiveVsNonConsumingChoice,RaceConditionIT:RWArchiveVsFetch,RaceConditionIT:WWDoubleArchive" +
         ",ExceptionsIT,ExceptionRaceConditionIT" +  # need UCK mode - added below
         ",DeeplyNestedValueIT" +  # FIXME: Too deeply nested values flake with a time out (half of the time)
-        ",CommandServiceIT:CSReturnStackTrace",  # FIXME: Ensure canton returns stack trace
+        ",CommandServiceIT:CSReturnStackTrace" +  # FIXME: Ensure canton returns stack trace
+        ",CommandServiceIT:CSsubmitAndWaitForTransactionIdInvalidLedgerId,CommandServiceIT:CSsubmitAndWaitForTransactionInvalidLedgerId,CommandServiceIT:CSsubmitAndWaitForTransactionTreeInvalidLedgerId" +  # FIXME: return definite_answer in gRPC errors
+        ",CommandSubmissionCompletionIT:CSCRefuseBadChoice,CommandSubmissionCompletionIT:CSCSubmitWithInvalidLedgerId,CommandSubmissionCompletionIT:CSCDisallowEmptyTransactionsSubmission",  # FIXME: return definite_answer in gRPC errors
```
nicu-da, line 91: Do we have a Jira issue to track these?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandSubmissionCompletionIT.scala
```diff
@@ -101,14 +101,15 @@ final class CommandSubmissionCompletionIT extends LedgerTestSuite {
...
-      assertGrpcError(
+      assertGrpcErrorRegex(
         failure,
         Status.Code.INVALID_ARGUMENT,
         Some(
           Pattern.compile(
             "(unknown|Couldn't find requested) choice " + badChoice
           )
         ),
+        checkDefiniteAnswerMetadata = true,
```
nicu-da, line 112: Just making sure we are running the full compatibility test suite as well to ensure this won't break for older versions of the sdk.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -41,21 +44,44 @@ object Assertions {
...
+    * the regex matches some part of the message or there is no message and the pattern is
+    * None.
     */
   @tailrec
-  def assertGrpcError(t: Throwable, expectedCode: Status.Code, optPattern: Option[Pattern]): Unit =
+  def assertGrpcErrorRegex(
+      t: Throwable,
+      expectedCode: Status.Code,
+      optPattern: Option[Pattern],
+      checkDefiniteAnswerMetadata: Boolean = false,
```
nicu-da, line 73: I think this naming is a bit obtuse. I was expecting to check something about the definitive answer. 
Maybe renaming to something like `assertDefinitiveAnswerIsReceived`

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandService.scala
```diff
@@ -91,7 +90,7 @@ private[apiserver] final class ApiCommandService private[services] (
         submissionTracker.track(CommandSubmission(commands, timeout))
       } else {
         Future.failed(
-          new ApiException(Status.UNAVAILABLE.withDescription("Service has been shut down."))
+          ErrorFactories.serviceNotRunning(definiteAnswer = Some(false))
```
nicu-da, line 93: Would we ever return this with definitive set to true?

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/TriggerRunnerImpl.scala
```diff
@@ -85,8 +85,8 @@ object TriggerRunnerImpl {
             replyTo ! QueryingACS
             Behaviors.same
           case QueryACSFailed(cause: io.grpc.StatusRuntimeException)
-              if cause.getStatus == io.grpc.Status.UNAUTHENTICATED =>
-            throw new UnauthenticatedException(s"Querying ACS failed: ${cause.toString}")
+              if cause.getStatus.getCode == Code.UNAUTHENTICATED =>
```
nicu-da, line 88: Just double-checking that we have tests for this.

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/ValueValidator.scala
andreaslochbihler-da: The semantics of definite answers is explained in https://github.com/digital-asset/daml/pull/9567/files#diff-ed4ed30e4f593aaf9d013b39ed5272dfc0df08c548cc78f110fd4471eaa6f59cR143-R151. It applies even if the application had previously sent or will later send a different command for the same change ID.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
hubert-da: There are multiple places where the map is being used, so it reduces duplication. I would leave it as is.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
hubert-da: No, as it has to be done in Canton first. So, maybe it would be worth capturing it there (their story). However, at least Oliver and Andreas are aware of it.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandSubmissionCompletionIT.scala
hubert-da: Not sure what you mean by that. Older SDK versions are excluded from compatibility tests, as they do not return the definite answer.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
hubert-da: Cosmetics, I can change if you insist :) This one is shorter :)

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiCommandService.scala
hubert-da: I can imagine that for certain implementations, a service may be turned off and it's not even configurable.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/TriggerRunnerImpl.scala
hubert-da: Not sure, I haven't found one now. There was a test that was failing due to this kind of comparison. However, I think we would need to revisit the test coverage for the whole class.

PR #10836, opened 2021-09-10
[Ledger API error codes] ErrorCode interfaces and generator [DPP-591]

tudor-da:
This PR is the basis for the implementation (as ported and adapted from Canton) of better error codes in our APIs. More specifically:
 * Creates a new `error` library under `//ledger`.
 * Implements the error interfaces.
 * Implements the documentation generator.
 * Implements the common error category definitions.
 * Unit tests of the generator and error code logging.
 
I've tried as much as possible to not change the ported code so the PR contains minimal refactorings.
Trying this PR shorter and easier to review: the following implementation points have been left out (and will be addressed in a subsequent PR):
* Error code logging is simplified.
* Tracing is not bundled with logging.
* Transformation to GRPC errors is not implemented.
* Interface alignment with DAML SDK and Canton will be done in a subsequent PR.
* Refactorings and cleanups.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/error/src/main/scala/com/daml/error/generator/DocItem.scala
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.error.generator
+
+import com.daml.error.{Explanation, Resolution}
+
+/** Contains error presentation data to be used for documentation rendering on the website.
+  *
+  * @param className The error class name (see [[com.daml.error.ErrorCode]]).
```
pbatko-da, line 10: Error code or error class? 
`className` vs `(see [[com.daml.error.ErrorCode]]).`

pbatko-da: Looking at the usages in Canton I now see that this field refers to ~`error.getClass.getName`

pbatko-da: My suggestions:
- rename it to `errorCodeName`.
- add `name` attribute to `ErrorCode` implemented as `this.getClass.getName`

ledger/error/src/main/scala/com/daml/error/ErrorClass.scala
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.error
+
+/** The classes [[ErrorClass]] and [[ErrorGroup]] are used to hierarchically structure error codes (their
+  * hierarchical structure affects how they are displayed on the website)
+  */
+case class ErrorClass(docNames: List[String]) {
```
pbatko-da, line 9: I would name this class e.g `ErrorGroup(Hierarchy)Path`  because:
1) AFAICT This class is encoding a path in a nesting of error groups. 
2) Having error categories, codes, groups and also classes (and what's the distinction) is a bit hard to wrap my head around quickly.
3) We avoid semantic class with JVM's notion of classes (e.g. when writing docs; what 'error class' refers to might be ambiguous)

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
```diff
@@ -0,0 +1,113 @@
...
+package com.daml.error
+
+import com.daml.logging.{ContextualizedLogger, LoggingContext}
+import io.grpc.StatusRuntimeException
+import org.slf4j.event.Level
+
+import scala.annotation.StaticAnnotation
+import scala.util.control.NoStackTrace
+
+/** Error Code Definition
```
pbatko-da, line 13: Would be maybe good to start this doc with an explanation where `ErrorCode` fits within the structure of error related classes. E.g.:
"""
Error codes gathers errors with the same error code.
Error code belongs to an error group.
Error codes determines high level details what it is and how to deal with it. For more information refer to particular subclass of BaseError.
"""

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
```diff
@@ -0,0 +1,120 @@
...
+        sys.error(
+          s"Error code $code is used ${occurrences.size} times but we require each error code to be unique! " +
+            s"Make these error code unique to make this assertion run through"
+        )
+    }
+
+    errorCodes.map(convertToDocItem).sortBy(_.code)
+  }
+
+  private def getErrorCodeInstances =
```
pbatko-da, line 38: Do we care of adding empty parameter list as this theoretically return different results on each call (because a new class has been loaded in the meantime?)

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
```diff
@@ -0,0 +1,120 @@
...
+      .collect {
+        case clazz if !Modifier.isAbstract(clazz.getModifiers) =>
+          clazz.getDeclaredField("MODULE$").get(clazz).asInstanceOf[ErrorCode]
+      }
+      .toSeq
+
+  private def convertToDocItem(error: ErrorCode): DocItem = {
+    val (expl, res) = getErrorNameAndAnnotations(error)
+    DocItem(
+      error.getClass.getName,
```
pbatko-da, line 52: Use named args as many of these fields have the same type?

ledger/error/src/test/scala/com/daml/error/ErrorCodeSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+class ErrorCodeSpec extends AnyFlatSpec with Matchers with BeforeAndAfter {
+  implicit private val testLoggingContext: LoggingContext = LoggingContext.ForTesting
+  private val logger = ContextualizedLogger.get(getClass)
+
+  private val className = classOf[ErrorCode].getSimpleName
+
+  s"$className.log" should "log the error message" in {
+    SeriousError.Error("the error argument").logWithContext(logger)
+
+    val actualLogs = LogCollector.read[this.type, this.type]
```
pbatko-da, line 23: `LogCollector.read[this.type, this.type]` - magic :D

pbatko-da (approved):

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
```diff
@@ -0,0 +1,113 @@
...
+  * }
+  */
+abstract class ErrorCode(val id: String, val category: ErrorCategory)(implicit
+    val parent: ErrorClass
+) {
+
+  // TODO error codes: Implement `asGrpcError`
+
+  require(id.length < 64, s"error-id is too long: $id")
+  require(id.forall(c => c.isUpper || c == '_' || c.isDigit), s"Invalid characters in error-id $id")
```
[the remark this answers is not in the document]
tudor-da: Good observation 👍

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
```diff
@@ -0,0 +1,113 @@
...
+    val loggedAs = s"This error is logged with log-level $logLevel on the server side."
+    val apiLevel = (category.grpcCode, exposedViaApi) match {
+      case (Some(code), true) =>
+        if (category.securitySensitive)
+          s"\nThis error is exposed on the API with grpc-status $code without any details due to security reasons"
+        else
+          s"\nThis error is exposed on the API with grpc-status $code including a detailed error message"
+      case _ => ""
+    }
+    Some(loggedAs ++ apiLevel)
```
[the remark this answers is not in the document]
tudor-da: afaik, use `+` is highly discouraged for strings, since it can silently allow any RHS argument:
```
234 + "this should not be possible" // but it compiles
234 ++ "this is not possible" // doesn't compile
```

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
tudor-da: Good point

ledger/error/src/main/scala/com/daml/error/BaseError.scala
```diff
@@ -0,0 +1,83 @@
...
+    } else None
+  }
+}
+
+object BaseError {
+  private val ignoreFields = Set("cause", "throwable", "loggingContext")
+
+  def extractContext[D](obj: D): Map[String, String] = {
+    obj.getClass.getDeclaredFields
+      .filterNot(x => ignoreFields.contains(x.getName) || x.getName.startsWith("_"))
```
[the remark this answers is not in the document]
tudor-da: Might be, but sometimes internal/mutable fields are prefix with `_` as well

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
```diff
@@ -0,0 +1,113 @@
...
+  * Errors are organised according to ErrorGroups. And we separate the error code definition (using a singleton
+  * by virtue of using objects to express the nested hierarchy) and the actual error.
+  *
+  * Please note that there is some implicit argument passing involved in the example below:
+  *
+  * object SyncServiceErrors extends ParticipantErrorGroup {
+  *   object ConnectionErrors extends ErrorGroup {
+  *     object DomainUnavailable extends ErrorCode(id="DOMAIN_UNAVAILABLE", ..) {
+  *        case class ActualError(someContext: Val) extends BaseError with SyncServiceError
+  *        // this error will actually be referring to the same error code!
```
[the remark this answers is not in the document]
tudor-da: why? The way I understand it is that both errors have the same error code: `ErrorCode(id="DOMAIN_UNAVAILABLE", ..)..`

ledger/error/src/main/scala/com/daml/error/generator/DocItem.scala
tudor-da: That's a good suggestion, but I would refrain from refactoring in this PR. I'll add a TODO, though

ledger/error/src/main/scala/com/daml/error/ErrorClass.scala
tudor-da: For avoiding breaking changes with dependents of this API, I would defer from major refactorings here. This is a good discussion point for a refactoring PR, though

ledger/error/src/main/scala/com/daml/error/ErrorClass.scala
```diff
@@ -0,0 +1,16 @@
...
+  def extend(docName: String = ""): ErrorClass = {
```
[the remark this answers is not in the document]
tudor-da: Right, it isn't. I'll leave a TODO

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
tudor-da: Let's leave this for a refactoring PR as well.

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
tudor-da: I think that `Reflections` scans the entire classpath so the results are (hopefully) deterministic.

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
tudor-da: Done

tudor-da: Done

ledger/error/BUILD.bazel
```diff
@@ -0,0 +1,62 @@
...
+        "//libs-scala/contextualized-logging",
+        "@maven//:io_grpc_grpc_api",
+        "@maven//:org_reflections_reflections",
+        "@maven//:org_slf4j_slf4j_api",
+    ],
+)
+
+da_scala_library(
+    name = "error-test-package",
+    srcs = glob(["src/test/scala/com/daml/error/utils/**/*.scala"]),
```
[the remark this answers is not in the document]
tudor-da: Done, split in `utils` and `test`

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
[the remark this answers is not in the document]
pbatko-da: You're right, I misread it

ledger/error/src/test/utils/scala/com/daml/error/utils/testpackage/SeriousError.scala
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.error.utils.testpackage
```
mziolekda, line 4: Don't we usually call the test files SomethingSpec.scala?

mziolekda (approved):

ledger/error/src/main/scala/com/daml/error/BaseError.scala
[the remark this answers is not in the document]
andreaslochbihler-da: Well, it's a simple way to express that some declared field in an error object should not be included in the error details. Personally, I'd be happier if we could make this more explicit in the field's name, say by testing for a substring like `excludeFromMetadata` in the field name, or (even better) having the option that every subclass of `BaseError` can specify a list of excluded field names.

@daravep knows whether we need this for synthetic fields.

ledger/error/src/main/scala/com/daml/error/BaseError.scala
```diff
@@ -0,0 +1,84 @@
...
+    * If you want to log an exception as part of your error, then use the following example:
+    *
+    * object MyCode extends ErrorCode(id="SUPER_DUPER_ERROR") {
+    *   case class MyError(someString: String, throwable: Throwable) extends SomeInternalError(
+    *     cause = "Something failed with an exception bla",
+    *     throwableO = Some(throwable)
+    *   )
+    * }
+    */
+  def throwableO: Option[Throwable] = None
```
miklos-da, line 33: We don't generally use this naming convention in the Daml SDK (`O` suffix for `Option`s).

Can we have simply `throwable` here, please? We already have e.g. `retryable` defined in this trait which returns an `Option[ErrorCategoryRetry]`.

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
```diff
@@ -0,0 +1,120 @@
...
+case class ErrorCodeDocumentationGenerator(prefix: String = "com.daml") {
+
+  def getDocItems: Seq[DocItem] = {
+    val errorCodes = getErrorCodeInstances
+
+    errorCodes.view.map(_.id).groupBy(identity).collect {
+      case (code, occurrences) if occurrences.size > 1 =>
+        sys.error(
+          s"Error code $code is used ${occurrences.size} times but we require each error code to be unique! " +
+            s"Make these error code unique to make this assertion run through"
```
miklos-da, line 31: Unnecessary string interpolation.

ledger/error/src/main/scala/com/daml/error/ErrorCategory.scala
```diff
@@ -0,0 +1,360 @@
...
+  def asInt: Int
+
+  /** Rank used to order severity (internal only) */
+  private[error] def rank: Int
+
+}
+
+object ErrorCategory {
+
+  val all: Seq[ErrorCategory] =
```
miklos-da, line 41: As this is a constant this should be `All` (see [Scala Style Guide](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables)).

ledger/error/src/main/scala/com/daml/error/ErrorCode.scala
```diff
@@ -0,0 +1,114 @@
...
+  protected def logLevel: Level = category.logLevel
+
+  /** True if this error may appear on the API */
+  protected def exposedViaApi: Boolean = category.grpcCode.nonEmpty
+
+  /** The error conveyance doc string provides a statement about the form this error will be returned to the user */
+  private[error] def errorConveyanceDocString: Option[String] = {
+    val loggedAs = s"This error is logged with log-level $logLevel on the server side."
+    val apiLevel = (category.grpcCode, exposedViaApi) match {
+      case (Some(code), true) =>
```
miklos-da, line 88: To improve readability, I'd suggest branching on `securitySensitive` as part of the pattern match and adding the newline separately, e.g.:
```scala
  val apiLevel = ... match {
    case (Some(code), true) if category.securitySensitive =>
      s"This error is exposed ..."
    case (Some(code), true) if !category.securitySensitive =>
      s"This error is exposed ..."
    case _ => ""
  }
  ...
```

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
```diff
@@ -0,0 +1,120 @@
...
+import java.lang.reflect.Modifier
+import com.daml.error.ErrorCode
+import com.daml.error.generator.ErrorCodeDocumentationGenerator.{
+  explanationTypeName,
+  resolutionTypeName,
+}
+import com.daml.error.{Explanation, Resolution}
+import org.reflections.Reflections
+
+import scala.reflect.runtime.{universe => ru}
```
miklos-da, line 15: Why do we need this import alias? The code would be easier to read if you directly imported `Annotation`.

PR #10841, opened 2021-09-10
LF: Add missing collision check for type synonyms

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/NamedEntity.scala
+++ b/daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/NamedEntity.scala
@@ -3,6 +3,7 @@
 
 package com.daml.lf.validation
 
+import com.daml.lf.data.Ref
 import com.daml.lf.data.Ref.{DottedName, ModuleName, Name}
 import com.daml.lf.language.Ast
 import com.daml.lf.validation.Util._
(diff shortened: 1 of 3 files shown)


daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/NamedEntity.scala
```diff
@@ -29,22 +30,6 @@ object NamedEntity {
     def pretty: String = s"module $name"
   }
 
-  final case class NValDef(
```
cocreature, line 32: did we ever use this for anything?

cocreature (approved):
Thanks!

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/NamedEntity.scala
remyhaemmerle-da: Yes we use to have it when we implement collision, but we quickly drop collision between data types and values as it was annoying for the compiler.

daml-lf/validation/src/test/scala/com/digitalasset/daml/lf/validation/CollisionSpec.scala
```diff
@@ -101,31 +106,44 @@ class CollisionSpec extends AnyWordSpec with Matchers {
...
-        """
+        """,
+        // a package with collision: two constructs have the same fully resoled name "A.B"
+        p"""
+          module A {                    // fully resolved name: "A"
+            synonym B = |Mod:A|;        // fully resolved name: "A.B" (collision)
+          }
+
+          module a.B {                  // fully resolved name: "a.B" (collision)
+
```
sofiafaro-da, line 138: I'm not sure that this is testing collision between module and type synonym, since this also has a collision between module `A` and (virtual) module `a`. Maybe something like this would isolate the synonym/module collision:

```
module A {
   synonym B = |Mod:A|;
   
   module b {
   }
}
```

PR #10843, opened 2021-09-10
interfaces: lf typechecker implementation

robin-da:
This implements the Daml-LF typechecking for interfaces. Currently I don't think there is a good way to test this and I'm also pretty sure that there is more to test. I'm opening this anyways so I can get some feedback.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
+++ b/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
@@ -536,8 +536,7 @@ typeOfCase scrut alts = do
                 DataEnum cons ->  (,,) (length cons) (\k -> CPEnum scrutTCon (cons !! k))
                     <$> typeOfAltsEnum scrutTCon cons alts
                 DataRecord{} -> (,,) 1 (const CPDefault) <$> typeOfAltsOnlyDefault scrutType alts
-                -- TODO https://github.com/digital-asset/daml/issues/10810
-                DataInterface -> error "interfaces are not implemented"
+                DataInterface -> (,,) 1 (const CPDefault) <$> typeOfAltsOnlyDefault scrutType alts
         TUnit -> (,,) 1 (const CPUnit) <$> typeOfAltsUnit alts
         TBool -> (,,) 2 (CPBool . toEnum) <$> typeOfAltsBool alts
         TList elemType -> (,,) 2 ([CPNil, CPCons wildcard wildcard] !!) <$> typeOfAltsList elemType alts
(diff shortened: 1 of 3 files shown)


compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
```diff
@@ -773,9 +772,19 @@ checkDefTypeSyn DefTypeSyn{synParams,synType} = do
...
 
+
+checkIface :: MonadGamma m => DefInterface -> m ()
+checkIface DefInterface{intName, intChoices} = do
+  checkUnique (EDuplicateInterfaceChoiceName intName) $ NM.names intChoices
+  forM_ intChoices checkIfaceChoice
+
+checkIfaceChoice :: MonadGamma m => InterfaceChoice -> m ()
+checkIfaceChoice InterfaceChoice{ifcRetType} =
+  checkType ifcRetType KStar
```
cocreature, line 783: Should we also check the kind of the argument type?

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
```diff
@@ -826,9 +835,27 @@ checkTemplate m t@(Template _loc tpl param precond signatories observers text ch
...
+  -> Qualified TypeConName
+  -> m ()
+checkIfaceImplementation modName tplTcon ifTcon
+  | Qualified PRSelf m _tconName <- ifTcon
+  , m == modName = do
+    DefInterface {intChoices} <- inWorld $ lookupInterface ifTcon
+    forM_ intChoices $ \InterfaceChoice {ifcName, ifcConsuming, ifcArgType, ifcRetType} -> do
+      TemplateChoice {chcConsuming, chcArgBinder, chcReturnType} <- inWorld $ lookupChoice (tplTcon, ifcName)
+      unless (chcConsuming == ifcConsuming) $ throwWithContext $ EBadInterfaceChoiceImplConsuming ifcName ifcConsuming chcConsuming
+      unless (snd chcArgBinder == ifcArgType) $ throwWithContext $ EBadInterfaceChoiceImplArgType ifcName ifcArgType (snd chcArgBinder)
```
cocreature, line 855: We should be using `alphaType` here instead of structural equality. Doesn't matter too much for now but  better to be on the safe side.

cocreature (approved):
Thanks!

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
robin-da: yep, added the check.

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
robin-da: agreed, changed it.

PR #10844, opened 2021-09-10
trigger reach auth on internal network

garyverhaegen-da:
In many network setups, there will be a more direct route for the trigger service to contact the auth middleware than going back through the frontend's public IP address (and possibly thus also through intermediaries like an nginx reverse proxy etc.). In _some_ network setups, it may not even be possible for the trigger service to reach the auth middleware through its externally-visible address.

This PR caters to these cases by allowing the trigger service to use two separate addresses for the auth middleware, an internal one the trigger service uses when it needs to talk to the auth middleware, and an external one used in generating URLs for external clients.

This is backwards-compatible: if the old option is used, we simply use the same value for both.

```
CHANGELOG_BEGIN
- The Trigger Service can now accept separate `--auth-internal` and
  `--auth-external` CLI arguments, where `--auth-internal` is the
  address used by the Trigger Service to reach the Auth Middleware
  directly, and `--auth-external` is the address the Trigger Service uses
  in generated URLs sent back to the client. The `--auth` option remains
  and keeps working as before, setting both internal and external
  addresses to the same given value.
CHANGELOG_END
```

### DIFF
--- a/triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
+++ b/triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
@@ -23,7 +23,9 @@ private[trigger] final case class ServiceConfig(
     httpPort: Int,
     ledgerHost: String,
     ledgerPort: Int,
-    authUri: Option[Uri],
+    authInternalUri: Option[Uri],
+    authExternalUri: Option[Uri],
+    authBothUri: Option[Uri],
     authRedirectToLogin: AuthClient.RedirectToLogin,
     authCallbackUri: Option[Uri],
     maxInboundMessageSize: Int,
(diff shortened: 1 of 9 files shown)


stefanobaghino-da (commented):
LGTM. @aherrmann-da should probably have a second look. 🙂

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
```diff
@@ -90,8 +92,24 @@ private[trigger] object ServiceConfig {
...
+      .action((t, c) => c.copy(authInternalUri = Some(Uri(t))))
+      .text(
+        "Sets the internal auth URIs (used by the trigger service to connect directly to the middleware). Incompatible with --auth."
+      )
+
+    opt[String]("auth-external")
+      .optional()
+      .action((t, c) => c.copy(authExternalUri = Some(Uri(t))))
+      .text(
+        "Sets the external auth URI (the one returned to the browser). Incompatible with --auth."
```
aherrmann-da, line 111: You could use `checkConfig` to ensure that `--auth` is not used in conjuction with `--auth-internal|external` directly in the options parser.

aherrmann-da (approved):
Thank you, looks good to me.
Minor comment about scopt.

I think there are a couple more tests that need updating. 
This might also be worth a test [here](https://github.com/digital-asset/daml/blob/6a9c8a64d32a08fd911aeca25d13844b5050bd2f/triggers/service/auth/src/test/scala/com/daml/auth/middleware/oauth2/TestMiddleware.scala#L602) and [here](https://github.com/digital-asset/daml/blob/6a9c8a64d32a08fd911aeca25d13844b5050bd2f/triggers/service/src/test-suite/scala/com/daml/lf/engine/trigger/ServiceConfigTest.scala#L10).

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
garyverhaegen-da: Added checkConfig, which actually allowed me to add tests in `ServiceConfigTest` (I'd tried before, but with the check happening much later there wasn't anything to test at that level).

I'm not sure what test to _add_ on the middleware side, but I've made the internal and external URLs different so we can at least confirm that only the external one is used in produced URLs.

PR #10850, opened 2021-09-10
interfaces: protobuf encoder haskell side

robin-da:
This is the implementation of the protobuf encoder on the haskell side.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/archive/src/main/protobuf/com/daml/daml_lf_dev/daml_lf_1.proto
+++ b/daml-lf/archive/src/main/protobuf/com/daml/daml_lf_dev/daml_lf_1.proto
@@ -1487,6 +1487,7 @@ message DefDataType {
     Fields record = 3; // Records without fields are explicitly allowed.
     Fields variant = 4; // Variants without constructors are explicitly allowed.
     EnumConstructors enum = 7;
+    Unit interface = 9; // Interfaces are special, they can't be represented as records or variants.
   }
 
   // If true, this data type preserves serializability in the sense that when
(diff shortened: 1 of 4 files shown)


daml-lf/archive/src/main/protobuf/com/daml/daml_lf_dev/daml_lf_1.proto
```diff
@@ -1487,6 +1487,7 @@ message DefDataType {
     Fields record = 3; // Records without fields are explicitly allowed.
     Fields variant = 4; // Variants without constructors are explicitly allowed.
     EnumConstructors enum = 7;
+    Unit interface = 9; // Interfaces are special, they can't be represented as records or variants.
```
cocreature, line 1490: I thought we said we don’t add them here for now since they’re redudant?

cocreature (approved):
lgtm except for the addition of the unit constructor which we should at least agree on, cc @remyhaemmerle-da

daml-lf/archive/src/main/protobuf/com/daml/daml_lf_dev/daml_lf_1.proto
robin-da: yes, I stumbled again over this and found it really weird to add an error case for the encoding of the DataInterface. Where do we have a guarantee that there is nowhere a DataInterface type left when encoding?

cocreature: alright, I’m fine with keeping this as well.

sofiafaro-da (approved):
LGTM, thanks!

PR #10857, opened 2021-09-13
es: drop jobs-* indices

garyverhaegen-da:
We are currently ingesting Bazel events in two forms:

In the `events-*` indices, each Bazel event is recorded as a separate ES object, with the corresponding job name as a field that can serve to aggregate all of the events for a given job.

In the `jobs-*` indices, each job is ingested as a single (composite) ES object, with the individual events as elements in a list-type field.

When I set up the cluster, I wasn't sure which one would be more useful, so I included both. We now have a bit more usage experience and it turns out the `events-*` form is the only one we use, so I think we should stop ingesting everything twice and from now on create only the `events-*` ones.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -714,25 +714,11 @@ emit_build_events() {
   jq -c \
      --slurpfile job_md "$job/job-md.json" \
      --arg cmd "$cmd" \
-     --arg index "$(index "$job" events)" \
+     --arg index "$(index "$job")" \
      --arg job "$job" \
      < "$file" \
      '
      { index: { _index: $index, _id: ($job + "-" + $cmd + "-events-" + (input_line_number | tostring)) } },
-     { job: $job_md[0],
-       command: { name: $cmd },
-       buildEvent: .
-     }
-     '
-  jq -c \
-     --slurpfile job_md "$job/job-md.json" \
-     --arg cmd "$cmd" \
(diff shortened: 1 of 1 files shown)


garyverhaegen-da:
There's no urgency on this — I'd prefer to wait for @aherrmann-da's stamp.

cocreature (approved):
lgtm but waiting for Andreas sounds good

stefanobaghino-da:
> lgtm but waiting for Andreas sounds good

Andreas will not be available for a couple of weeks.

cocreature:
> Andreas will not be available for a couple of weeks.

sure but @garyverhaegen-da said there is no urgency and he prefers to wait

stefanobaghino-da (approved):
LGTM 🙇🏻

stefanobaghino-da:
> > Andreas will not be available for a couple of weeks.
> 
> sure but @garyverhaegen-da said there is no urgency and he prefers to wait

Absolutely, I just wanted to point out that feedback from Andreas will need to wait for a couple of weeks. 🙂

stefanobaghino-da:
> LGTM 🙇🏻

No rush in merging this, feel free to wait for Andreas' opinion on this before merging.

aherrmann-da (approved):
Thank you, LGTM

garyverhaegen-da:
Deployed. All seems to be running fine.

PR #10861, opened 2021-09-13
save kibana exports

garyverhaegen-da:
As explained in #10853, we recently lost our ES cluster. While I'm not planning on trusting Google's "rolling restart" feature ever again, we can't exclude the possibility of future similar outages (without a significant investment in the cluster, which I don't think we want to do).

Losing the cluster is not a huge issue as we can always reingest the data. Worst case we lose visibility for a few days. At least, as far as the bazel logs are concerned.

Losing the Kibana data is a lot more annoying, as that is not derived data and thus cannot be reingested. This PR aims to add a backup mechanism for our Kibana configuration.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/infra/es_cluster.tf
+++ b/infra/es_cluster.tf
@@ -391,6 +391,27 @@ resource "google_project_iam_member" "es-feed" {
   member  = "serviceAccount:${google_service_account.es-feed.email}"
 }
 
+resource "google_project_iam_custom_role" "es-feed-write" {
+  role_id     = "es_feed_write"
+  title       = "es-feed-write"
+  description = "es-feed-write"
+  permissions = [
+    "storage.objects.write"
+  ]
+}
+
+resource "google_project_iam_member" "es-feed-write" {
+  project = local.project
+  role    = google_project_iam_custom_role.es-feed-write.id
+  member  = "serviceAccount:${google_service_account.es-feed-write.email}"
+
+  condition {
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

infra/es_cluster.tf
```diff
@@ -391,6 +391,27 @@ resource "google_project_iam_member" "es-feed" {
   member  = "serviceAccount:${google_service_account.es-feed.email}"
 }
 
+resource "google_project_iam_custom_role" "es-feed-write" {
+  role_id     = "es_feed_write"
```
stefanobaghino-da, line 395: Noob question: how do you decide when to use underscores and when to use dashes?

stefanobaghino-da (approved):
🙇🏻

infra/es_cluster.tf
garyverhaegen-da: GCP has rules. Some fields accept only dashes, some fields accept only underscores, some fields accept both. It's a bit... arcane.

garyverhaegen-da:
Had to add a few tweaks, which I've left as a separate commit for ease of review, but this is now deployed and confirmed to be working (at least in so far as it is writing exports to GCS; I'd rather wait for Andreas before testing the actual restore process).

PR #10862, opened 2021-09-13
Dpp 494 unit testing ha coordinator

nmarton-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
@@ -51,6 +51,9 @@ class PollingChecker(
   def check(): Unit = synchronized {
     logger.debug(s"Checking...")
     Try(checkBody) match {
+      case _ if closed =>
+        throw new Exception("Checker is already closed")
+
       case Success(_) =>
         logger.debug(s"Check successful.")
 
@@ -61,5 +64,10 @@ class PollingChecker(
     }
   }
 
-  def close(): Unit = timer.cancel()
+  private var closed: Boolean = false
+
+  def close(): Unit = synchronized {
+    closed = true
+    timer.cancel()
+  }
 }
--- a/ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
+++ b/ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
@@ -1,0 +1,226 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.indexer.ha
+
+import java.sql.{
+  Blob,
+  CallableStatement,
+  Clob,
+  Connection,
+  DatabaseMetaData,
+  NClob,
+  PreparedStatement,
+  SQLWarning,
+  SQLXML,
+  Savepoint,
+  Statement,
+  Struct,
+}
+import java.util.Properties
+import java.util.concurrent.Executor
+import java.{sql, util}
+
+import com.daml.platform.store.backend.DBLockStorageBackend
+
+class TestDBLockStorageBackend extends DBLockStorageBackend {
+
+  private var locks: Map[DBLockStorageBackend.Lock, Set[Connection]] = Map.empty
+
+  override def tryAcquire(
+      lockId: DBLockStorageBackend.LockId,
+      lockMode: DBLockStorageBackend.LockMode,
+  )(connection: Connection): Option[DBLockStorageBackend.Lock] = synchronized {
+    if (connection.isClosed) throw new Exception("trying to acquire on a closed connection")
+    if (!lockId.isInstanceOf[TestLockId]) throw new Exception("foreign lockId")
+    removeClosedConnectionRelatedLocks()
+    val lock = DBLockStorageBackend.Lock(lockId, lockMode)
+    def doLock(): Option[DBLockStorageBackend.Lock] = {
+      locks = locks + (lock -> (locks.getOrElse(lock, Set.empty) + connection))
+      Some(lock)
+    }
+    lockMode match {
+      case DBLockStorageBackend.LockMode.Exclusive =>
+        (
+          locks.get(lock),
+          locks.get(lock.copy(lockMode = DBLockStorageBackend.LockMode.Shared)),
+        ) match {
+          case (None, None) => doLock()
+          case (Some(connections), None) if connections == Set(connection) => doLock()
+          case _ => None // if any shared lock held, we cannot lock exclusively
+        }
+      case DBLockStorageBackend.LockMode.Shared =>
+        (
+          locks.get(lock),
+          locks.get(lock.copy(lockMode = DBLockStorageBackend.LockMode.Exclusive)),
+        ) match {
+          case (_, None) => doLock()
+          case _ => None // if any exclusive lock held, we cannot lock shared
+        }
+    }
+  }
+
+  override def release(lock: DBLockStorageBackend.Lock)(connection: Connection): Boolean =
+    synchronized {
+      if (connection.isClosed) throw new Exception("trying to release on a closed connection")
+      if (!lock.lockId.isInstanceOf[TestLockId]) throw new Exception("foreign lockId")
+      removeClosedConnectionRelatedLocks()
+      locks.get(lock) match {
+        case None => false
+        case Some(connections) if connections contains connection =>
+          if (connections.size == 1) locks = locks - lock
+          else locks = locks + (lock -> (connections - connection))
+          true
+        case _ => false
+      }
+    }
+
+  private def removeClosedConnectionRelatedLocks(): Unit =
+    locks = locks.collect {
+      case (lock, conns) if conns.exists(!_.isClosed) => (lock, conns.filter(!_.isClosed))
+    }
+
+  override def lock(id: Int): DBLockStorageBackend.LockId =
+    TestLockId(id)
+
+  def cutExclusiveLockHoldingConnection(lockId: Int): Unit = synchronized {
+    val mainExclusiveIndexerLock =
+      DBLockStorageBackend.Lock(TestLockId(lockId), DBLockStorageBackend.LockMode.Exclusive)
+    locks
+      .getOrElse(mainExclusiveIndexerLock, Set.empty)
+      .foreach(_.close())
+    locks = locks - mainExclusiveIndexerLock
+  }
+
+  override def dbLockSupported: Boolean = true
+}
+
+case class TestLockId(id: Int) extends DBLockStorageBackend.LockId
+
+class TestConnection extends Connection {
+  override def createStatement(): Statement = throw new UnsupportedOperationException
+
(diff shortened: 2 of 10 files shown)


ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
```diff
@@ -61,5 +64,10 @@ class PollingChecker(
     }
   }
 
-  def close(): Unit = timer.cancel()
+  private var closed: Boolean = false
+
+  def close(): Unit = synchronized {
+    closed = true
```
mziolekda, line 70: As I understand cancel kills all waiting timer tasks, but not the tasks that are already running. 
There might already be a check function ready to run when this line is executed. This will lead to the Exception being thrown. Am I missing something?

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+    }
+    lockMode match {
+      case DBLockStorageBackend.LockMode.Exclusive =>
+        (
+          locks.get(lock),
+          locks.get(lock.copy(lockMode = DBLockStorageBackend.LockMode.Shared)),
+        ) match {
+          case (None, None) => doLock()
+          case (Some(connections), None) if connections == Set(connection) => doLock()
+          case _ => None // if any shared lock held, we cannot lock exclusively
```
mziolekda, line 50: We should be able to upgrade from shared -> exclusive when we are the only holders, shouldn't we?

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
nmarton-da: critical sections demarcated by synchronized are mutually exclusive, and waiting for each other to enter.
check is also demarcated by synchronized

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
nmarton-da: strictly speaking yes in case of postgres. but I built the solution to avoid dependancy on lock upgrade features, so we do not have to worry about it.
pls note this test-DBLockStorageBackend is tested with the same test suite, then the real ones (Oracle, Postgres)

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
mziolekda: check is demarcated by synchronize but the TimerTask.run() method is not. In fact run() method can be waiting to enter the critical section in Try(check()) as closed=true is being written.
The safest way to cancel a timer is to write a variable then leave inside the check() method first thing.

nmarton-da: not sure what are you suggesting exactly, bringing this offline?

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+}
+import java.util.Properties
+import java.util.concurrent.Executor
+import java.{sql, util}
+
+import com.daml.platform.store.backend.DBLockStorageBackend
+
+class TestDBLockStorageBackend extends DBLockStorageBackend {
+
+  private var locks: Map[DBLockStorageBackend.Lock, Set[Connection]] = Map.empty
```
rautenrieth-da, line 28: This probably means the methods of this class are not safe to use concurrently. Could this be an issue? Could you document this at the class level?

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
mziolekda: We have gone over this offline and the challenge here is that the check function is called from 2 places: from the timer and from HaCoordinator. It is desirable to throw in the latter case, while timer is anyway enclosed in a Try-do-nothing clause.

mziolekda (approved):

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  override def tryAcquire(
```
rautenrieth-da, line 30: This is a fair amount of code. It looks ok, but we should be aware that the locking behavior of real databases can behave differently.

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/PollingChecker.scala
nmarton-da: we leave it as discussed offline

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
nmarton-da: I think it should be thread safe, every place we modify/read this are in synchronized blocks

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
nmarton-da: They can yes, but for what it is important to us that both this and both real databases must confirm to our expectations.
These expectations are encapsulated at StorageBackendTestsDBLock.
And that's why I've added TestDBLockStorageBackendSpec, to have this conformance given for our Ha testing.

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/ha/HaCoordinatorSpec.scala
```diff
@@ -0,0 +1,798 @@
...
+  behavior of "databaseLockBasedHaCoordinator graceful shutdown"
+
+  it should "successfully propagate successful end of execution" in {
+    val protectedSetup = setup()
+    import protectedSetup._
+
+    for {
+      _ <- connectionInitializerFuture
+      _ = {
+        info("As HA initialized")
```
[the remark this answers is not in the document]
nmarton-da: don't mind at all, keep them coming :)
in this case it is not really about the HA setup here. HA setup kicks off the async processing, which has a part when HACoordinator initializes (gets connection, acquires main and worker locks, starts a polling checker), as connectionInitializerFuture is completed, we can start initializing in the protected section.

clarification: all these info commentary relates to thing happened above

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/ha/HaCoordinatorSpec.scala
```diff
@@ -0,0 +1,845 @@
...
+        protectedHandle.completed.isCompleted shouldBe false
+        info("Protected Handle is not completed")
+        executionHandlePromise.success(())
```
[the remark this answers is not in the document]
nmarton-da: sure, I like it!

meiersi-da (commented):
@nmarton-da : I did not manage to completely review the `HaCoordinatorSpec` both because of a lack of knowledge and time. I see though that you are getting good reviews from @mziolekda and @rautenrieth-da.

I thus won't aim to complete this review. Here are the comments I've had so far. All minor, so please feel free to ignore.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
rautenrieth-da: You are right, I have missed the `synchronized`.

rautenrieth-da: Actually, I'm not sure. You can still call `tryAcquire` and `release` at the same time, right? And they both access the `locks` mutable value.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+          if (connections.size == 1) locks = locks - lock
+          else locks = locks + (lock -> (connections - connection))
+          true
+        case _ => false
+      }
+    }
+
+  private def removeClosedConnectionRelatedLocks(): Unit =
+    locks = locks.collect {
+      case (lock, conns) if conns.exists(!_.isClosed) => (lock, conns.filter(!_.isClosed))
```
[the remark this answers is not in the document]
nmarton-da: the higher order function here is a collect, for that collect it is important to filter out locks which have no non-closed connections, since having an entry in the map with empty set is not equivalent to not having that entry

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+      .foreach(_.close())
+    locks = locks - mainExclusiveIndexerLock
+  }
+
+  override def dbLockSupported: Boolean = true
+}
+
+case class TestLockId(id: Int) extends DBLockStorageBackend.LockId
+
+class TestConnection extends Connection {
```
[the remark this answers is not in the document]
nmarton-da: I understand your motivation regarding removing boilerplate here, probably is possible to do it with mockito as well, but I think it is more natural to do it in code. Also the Connection interface is unlikely to change much, so the boilerplate hurts us less here. For me not depending on a mock framework for this simple use-case outweighs the removal of the boilerplate code.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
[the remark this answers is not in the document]
meiersi-da: ah, my Haskell background led me astray. I didn't realize that there is an implicit discarding happening here. Makes sense now.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/TestDBLockStorageBackend.scala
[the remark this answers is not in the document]
meiersi-da: Makes sense. Thanks for the explanation.

ledger/participant-integration-api/src/test/suite/scala/platform/indexer/ha/HaCoordinatorSpec.scala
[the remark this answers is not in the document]
nmarton-da: fixed

PR #10864, opened 2021-09-13
Command deduplication - better support for different deduplication modes in conformance tests [KVL-1099]

nicu-da:
Add deduplication config which defines the following "features"
- participant deduplication, which implies sync duplicate errors
- append-only schema, which allows for extensive completion assertions

Better naming for assertions that support sync and async deduplication

CHANGELOG_BEGIN

CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
```diff
@@ -23,7 +24,9 @@ final class CommandDeduplicationIT(timeoutScaleFactor: Double, ledgerTimeInterva
 
   override def testNamingPrefix: String = "ParticipantCommandDeduplication"
 
-  /** Assertions for append-only schema are important for KV ledgers
-    */
-  override protected def isAppendOnly: Boolean = false
+  override def deduplicationFeatures: CommandDeduplicationBase.DeduplicationFeatures =
```
fabiotudone-da, line 27: How about renaming this test something like `ParticipantOnlyCommandDeduplicationIT`?

fabiotudone-da (approved):

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/CommandDeduplicationIT.scala
nicu-da: I would love to, but this is a default test and I'm not sure about the implications.
No issues about our builds but we would also need to update any external usages (eg canton)

PR #10868, opened 2021-09-13
Make `CommandTracker` distinguish submissions of the same command using `submissionId` [KVL-1104]

hubert-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -4,7 +4,6 @@
 package com.daml.ledger.client.services.commands.tracker
 
 import java.time.{Duration, Instant}
-
 import akka.stream.stage._
 import akka.stream.{Attributes, Inlet, Outlet}
 import com.daml.grpc.{GrpcException, GrpcStatus}
@@ -299,7 +298,7 @@ private[commands] class CommandTracker[Context](
       }
 
       private def getOutputForCompletion(completion: Completion) = {
-        val (commandKey, errorText) = {
+        val (potentialCommandKey, errorText) = {
           completion.status match {
             case Some(StatusProto(code, _, _, _)) if code == Status.Code.OK.value =>
               TrackedCommandKey(
@@ -314,7 +313,25 @@ private[commands] class CommandTracker[Context](
           }
         }
 
-        logger.trace("Handling {} {}", errorText, completion.commandId: Any)
+        val commandKey = if (potentialCommandKey.submissionId.isEmpty) {
+          val potentialKeys =
+            pendingCommands.keys.filter(_.commandId == potentialCommandKey.commandId)
+          if (potentialKeys.size > 1) {
+            throw new IllegalStateException(
+              s"There are multiple pending commands for the id ${potentialCommandKey.commandId}. This can only happen for the mutating schema."
+            ) with NoStackTrace
+          }
+          potentialKeys.head
+        } else {
+          potentialCommandKey
+        }
+
+        logger.trace(
+          "Handling {} {} from submission {}",
+          errorText,
+          completion.commandId,
+          completion.submissionId,
+        )
         pendingCommands.remove(commandKey).map { t =>
           Ctx(t.context, tracker.CompletionResponse(completion))
         }
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -317,7 +317,7 @@ private[commands] class CommandTracker[Context](
           case id if id.nonEmpty => id
         }
         val trackedCommands =
-          pendingCommandKeys(maybeSubmissionId, commandId).flatMap(pendingCommands.remove)
+          pendingCommandKeys(maybeSubmissionId, commandId).flatMap(pendingCommands.remove(_).toList)
 
         if (trackedCommands.size > 1) {
           trackedCommands.map { trackingData =>
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -185,7 +185,7 @@ private[commands] class CommandTracker[Context](
       )
 
       private def pushResultOrPullCommandResultIn(
-          compl: immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]]
+          compl: Seq[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]]
       ): Unit = {
         // The command tracker detects timeouts outside the regular pull/push
         // mechanism of the input/output ports. Basically the timeout
@@ -196,9 +196,8 @@ private[commands] class CommandTracker[Context](
         // handling the signaling properly.
         if (compl.isEmpty && !hasBeenPulled(commandResultIn)) {
           pull(commandResultIn)
-        } else {
-          emitMultiple(resultOut, compl)
         }
+        emitMultiple(resultOut, compl)
       }
 
       private def completeStageIfTerminal(): Unit = {
@@ -211,7 +210,7 @@ private[commands] class CommandTracker[Context](
 
       private def handleSubmitResponse(
           submitResponse: Ctx[(Context, TrackedCommandKey), Try[Empty]]
-      ): immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
+      ): Seq[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
         val Ctx((_, commandKey), value, _) = submitResponse
         value match {
           case Failure(GrpcException(status @ GrpcStatus(code, _), metadata))
@@ -222,12 +221,12 @@ private[commands] class CommandTracker[Context](
               s"Service responded with error for submitting command with context ${submitResponse.context}. Status of command is unknown. watching for completion...",
               throwable,
             )
-            immutable.Iterable.empty
+            Seq.empty
           case Success(_) =>
             logger.trace(
               s"Received confirmation that command ${commandKey.commandId} from submission ${commandKey.submissionId} was accepted."
             )
-            immutable.Iterable.empty
+            Seq.empty
         }
       }
 
@@ -272,33 +271,31 @@ private[commands] class CommandTracker[Context](
 
       private def getOutputForTimeout(instant: Instant) = {
         logger.trace("Checking timeouts at {}", instant)
-        pendingCommands.view
-          .flatMap { case (commandKey, trackingData) =>
-            if (trackingData.commandTimeout.isBefore(instant)) {
-              pendingCommands -= commandKey
-              logger.info(
-                s"Command {} from submission {} (command timeout {}) timed out at checkpoint {}.",
-                commandKey.commandId,
-                commandKey.submissionId,
-                trackingData.commandTimeout,
-                instant,
+        pendingCommands.view.flatMap { case (commandKey, trackingData) =>
+          if (trackingData.commandTimeout.isBefore(instant)) {
+            pendingCommands -= commandKey
+            logger.info(
+              s"Command {} from submission {} (command timeout {}) timed out at checkpoint {}.",
+              commandKey.commandId,
+              commandKey.submissionId,
+              trackingData.commandTimeout,
+              instant,
+            )
+            List(
+              Ctx(
+                trackingData.context,
+                Left(CompletionResponse.TimeoutResponse(commandId = trackingData.commandId)),
               )
-              List(
-                Ctx(
-                  trackingData.context,
-                  Left(CompletionResponse.TimeoutResponse(commandId = trackingData.commandId)),
-                )
-              )
-            } else {
-              Nil
-            }
+            )
+          } else {
+            Nil
           }
-          .to(immutable.Iterable)
+        }.toSeq
       }
 
       private def getOutputForCompletion(
           completion: Completion
-      ): immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
+      ): Seq[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
         val completionDescription = completion.status match {
           case Some(StatusProto(code, _, _, _)) if code == Status.Code.OK.value =>
             "successful completion of command"
@@ -316,8 +313,9 @@ private[commands] class CommandTracker[Context](
         val maybeSubmissionId = Option(completion.submissionId).collect {
           case id if id.nonEmpty => id
         }
+        val value = pendingCommandKeys(maybeSubmissionId, commandId)
         val trackedCommands =
-          pendingCommandKeys(maybeSubmissionId, commandId).flatMap(pendingCommands.remove(_).toList)
+          value.flatMap(pendingCommands.remove(_).toList)
 
         if (trackedCommands.size > 1) {
           trackedCommands.map { trackingData =>
@@ -345,15 +343,15 @@ private[commands] class CommandTracker[Context](
       private def pendingCommandKeys(
           submissionId: Option[String],
           commandId: String,
-      ) =
-        submissionId.map(id => immutable.Iterable(TrackedCommandKey(id, commandId))).getOrElse {
-          pendingCommands.keys.filter(_.commandId == commandId).to(immutable.Iterable)
+      ): Seq[TrackedCommandKey] =
+        submissionId.map(id => Seq(TrackedCommandKey(id, commandId))).getOrElse {
+          pendingCommands.keys.filter(_.commandId == commandId).toList
         }
 
       private def getOutputForTerminalStatusCode(
           commandKey: TrackedCommandKey,
           status: StatusProto,
-      ): immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
+      ): Seq[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
         logger.trace(
           s"Handling failure of command ${commandKey.commandId} from submission ${commandKey.submissionId}."
         )
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -94,7 +94,7 @@ private[commands] class CommandTracker[Context](
         timerKey match {
           case `timeout_detection` =>
             val timeouts = getOutputForTimeout(Instant.now)
-            if (timeouts.nonEmpty) emitMultiple(resultOut, timeouts)
+            if (timeouts.nonEmpty) emitMultiple(resultOut, timeouts.to(immutable.Iterable))
           case _ => // unknown timer, nothing to do
         }
       }
@@ -197,7 +197,7 @@ private[commands] class CommandTracker[Context](
         if (compl.isEmpty && !hasBeenPulled(commandResultIn)) {
           pull(commandResultIn)
         }
-        emitMultiple(resultOut, compl)
+        emitMultiple(resultOut, compl.to(immutable.Iterable))
       }
 
       private def completeStageIfTerminal(): Unit = {
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -3,7 +3,6 @@
 
 package com.daml.ledger.client.services.commands.tracker
 
-import java.time.{Duration, Instant}
 import akka.stream.stage._
 import akka.stream.{Attributes, Inlet, Outlet}
 import com.daml.grpc.{GrpcException, GrpcStatus}
@@ -14,6 +13,7 @@ import com.daml.ledger.client.services.commands.tracker.CommandTracker._
 import com.daml.ledger.client.services.commands.tracker.CompletionResponse.{
   CompletionFailure,
   CompletionSuccess,
+  NotOkResponse,
 }
 import com.daml.ledger.client.services.commands.{
   CommandSubmission,
@@ -26,6 +26,7 @@ import com.google.rpc.status.{Status => StatusProto}
 import io.grpc.Status
 import org.slf4j.LoggerFactory
 
+import java.time.{Duration, Instant}
 import scala.annotation.nowarn
 import scala.collection.compat._
 import scala.collection.{immutable, mutable}
@@ -184,7 +185,7 @@ private[commands] class CommandTracker[Context](
       )
 
       private def pushResultOrPullCommandResultIn(
-          compl: Option[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]]
+          compl: immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]]
       ): Unit = {
         // The command tracker detects timeouts outside the regular pull/push
         // mechanism of the input/output ports. Basically the timeout
@@ -193,7 +194,11 @@ private[commands] class CommandTracker[Context](
         // even though it hasn't been pulled again in the meantime. Using `emit`
         // instead of `push` when a completion arrives makes akka take care of
         // handling the signaling properly.
-        compl.fold(if (!hasBeenPulled(commandResultIn)) pull(commandResultIn))(emit(resultOut, _))
+        if (compl.isEmpty && !hasBeenPulled(commandResultIn)) {
+          pull(commandResultIn)
+        } else {
+          emitMultiple(resultOut, compl)
+        }
       }
 
       private def completeStageIfTerminal(): Unit = {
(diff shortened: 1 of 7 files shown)


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -197,33 +209,40 @@ private[commands] class CommandTracker[Context](
...
         val commandId = commands.commandId
-        logger.trace("Begin tracking of command {}", commandId)
-        if (pendingCommands.contains(commandId)) {
+        logger.trace("Begin tracking of command {} for submission {}", commandId, submissionId)
+        if (pendingCommands.contains(TrackedCommandKey(submissionId, commandId))) {
           // TODO return an error identical to the server side duplicate command error once that's defined.
           throw new IllegalStateException(
-            s"A command with id $commandId is already being tracked. CommandIds submitted to the CommandTracker must be unique."
+            s"A command $commandId from a submission $submissionId is already being tracked. CommandIds submitted to the CommandTracker must be unique."
           ) with NoStackTrace
```
nicu-da, line 246: Do you think we have any gaps in testing though for this? Wondering how come conformance tests were green considering this code. Shouldn't they throw an error for 2 submissions with the same command id currently? 

Also that TODO, should we throw an ApiException for duplicate command from here?

nicu-da (commented):
LGTM! Really nice work! 
Just 2 small questions

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: Thanks for the feedback!

>Do you think we have any gaps in testing though for this? Wondering how come conformance tests were green considering this code. Shouldn't they throw an error for 2 submissions with the same command id currently?

Not yet sure, I was going to think about this tomorrow, as I'm also not sure about some of these changes at all. Partially because I don't know Akka Streams Graph API 😄 I submitted it mostly as an FYI so far.

> Also that TODO, should we throw an ApiException for duplicate command from here?
 
Was also thinking about this. Will keep you posted.

nicu-da: Sounds good, let me know if you need any help with the Graph API, spent a lot of time writing akka graphs.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -116,11 +119,20 @@ private[commands] class CommandTracker[Context](
...
-              "Submitted command {}",
-              submitRequest.value.commands.commandId,
+              "Submitted command {} in submission {}",
+              commandId,
+              submissionId,
+            )
+            push(
+              submitRequestOut,
+              submitRequest.enrich((context, _) =>
+                context -> TrackedCommandKey(submissionId, commandId)
```
andreaslochbihler-da, line 133: IMO we should group the submission IDs by change Id rather than command ID because deduplication happens on the `ChangeId` rather than the command ID alone.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -197,33 +209,40 @@ private[commands] class CommandTracker[Context](
...
+              commandKey.submissionId,
+            )
             None
         }
       }
 
       @nowarn("msg=deprecated")
       private def registerSubmission(submission: Ctx[Context, CommandSubmission]): Unit = {
         val commands = submission.value.commands
+        val submissionId = commands.submissionId
```
andreaslochbihler-da, line 239: The `CommandService`should set the submission ID to a UUID if the application has left it empty in its submission. I haven't seen any such change in this PR. Will this be done separately?

nicu-da: That is already the case, being done here https://github.com/digital-asset/daml/blob/main/ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/validation/CommandsValidator.scala#L188

hubert-da: It is being done in `CommandsValidator.validateCommands` which is being run by `GrpcCommandService` earlier. It was done as part of [KVL-999](https://digitalasset.atlassian.net/browse/KVL-999).

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: Sure. Previously it was the command ID alone, the story mentions adding just the submission ID. The change ID is: the submitting party (+ `act_as`) + application ID + command ID. What about the submission ID?

andreaslochbihler-da (commented):
Thanks for pulling this together so quickly. I think we should add the following changes:

* The `TrackingKey` should be `(submissionId, changeId)`  rather than `(submissionId, commandId)` so that submissions from different submitters or different applications are isolated against each other.
* The `CommandService` should set the submission ID to a UUID if the application hasn't set a submission ID. I haven't seen such a change here and I haven't checked whether the is already happening elsewhere, but I'd be positively surprised if this is already implemented.

AFAICS there are not test cases for the new functionality. Can you make sure we've covered the following cases:
* Several parallel submissions of the same change ID with different submission IDs
* Defaulting of submission IDs in the CommandService (also for parallel submissions) if the application has omitted them.
* Parallel submissions from different applications / submitters (`actAs`) for the same command ID.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
andreaslochbihler-da: The change ID is defined here: https://github.com/digital-asset/daml/blob/main/ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/ChangeId.scala

Sure. We should track the tuple (submissionId, changeId).

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
andreaslochbihler-da: Great. Good to know. Do we have test cases for that?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
gerolf-da: There's no need to use change ID here, because there is a new `CommandTracker` for each (party, applicationId) tuple.

andreaslochbihler-da: Fair enough. I suppose that you mean for each `party union actAs`.

hubert-da: Exactly, I forgot yesterday to double-check how the `Tracking.Key` looks like. It contains everything we need. Thanks!

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: Hmm, sadly, I don't see any. Will have to add them as part of this story.

hubert-da:
> The TrackingKey should be (submissionId, changeId) rather than (submissionId, commandId) so that submissions from different submitters or different applications are isolated against each other.

As discussed in a different comment, the rest of the `changeId` is covered by the `Tracking.Key` which is a different thing. It makes the `Trackers` for different applications (and parties) isolated. Then, the `TrackedCommandKey` makes submissions with the same `commandId` distinguishable within a `Tracker`.

> The CommandService should set the submission ID to a UUID if the application hasn't set a submission ID. I haven't seen such a change here and I haven't checked whether the is already happening elsewhere, but I'd be positively surprised if this is already implemented.

As also discussed in another comment, it's implemented but doesn't have proper test coverage. So, moving smoothly to...

> AFAICS there are not test cases for the new functionality.

The test coverage is something that I have to revisit. Thanks!

hubert-da:
@andreaslochbihler-da , in regard to the test cases:
> Several parallel submissions of the same change ID with different submission IDs

Different change IDs are covered by `TrackerMapSpec`. Different/same submission IDs are now covered by `CommandTrackerFlowTest` and are going to be covered by the conformance test: https://github.com/digital-asset/daml/pull/10869

> Defaulting of submission IDs in the CommandService (also for parallel submissions) if the application has omitted them.

This one is covered by most of the conformance tests and `SubmitRequestValidatorTest` as with this change: https://github.com/digital-asset/daml/pull/10882, the validation requires a submission ID.

> Parallel submissions from different applications / submitters (actAs) for the same command ID.

This one is covered on the level of `TrackerMap` and tested in `TrackerMapSpec`.

Anyway, the test coverage is probably not perfect but if we want to release it tomorrow, this is what we have. I think at least a minimal set of test cases already exists.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: > Also that TODO, should we throw an ApiException for duplicate command from here?

Not sure what to do, as this is a `Unit` method. Maybe let's keep it as is for now and I can do something about this in a separate PR.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: Followed up in: https://github.com/digital-asset/daml/pull/10882

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -279,33 +292,70 @@ private[commands] class CommandTracker[Context](
       }
 
       private def getOutputForCompletion(completion: Completion) = {
-        val (commandId, errorText) = {
+        val (potentialCommandKey, errorText) = {
```
fabiotudone-da, line 295:
```suggestion
        val (potentialCommandKey, tracingText) = {
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -279,33 +292,70 @@ private[commands] class CommandTracker[Context](
...
             case _ =>
-              completion.commandId -> "failed completion of command"
+              TrackedCommandKey(
+                completion.submissionId,
+                completion.commandId,
+              ) -> "failed completion of command"
+          }
+        }
+
+        val commandKey = if (potentialCommandKey.submissionId.isEmpty) {
```
fabiotudone-da, line 310: Why is it OK if we get completions with empty submission IDs? Also, I'd prefer avoiding constructing invalid keys and throwing a little earlier.

fabiotudone-da (approved):

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
fabiotudone-da: A cleaner alternative (i.e. that doesn't mix logging and regular logic) would perhaps to just build the logged string based on the completion status.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: I assume this is for the mutable schema. 
Considering we have this check I would also add some validation in the `registerSubmission` where we validate that the submission id is not empty, so that we don't get up to this point if somehow the submission id was empty from the command.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -279,33 +292,70 @@ private[commands] class CommandTracker[Context](
...
+          val potentialKeys =
+            pendingCommands.keys.filter(_.commandId == potentialCommandKey.commandId)
+          if (potentialKeys.size > 1) {
+            throw new IllegalStateException(
+              s"There are multiple pending commands for the id ${potentialCommandKey.commandId}. This can only happen for the mutating schema."
+            ) with NoStackTrace
```
nicu-da, line 316: A bit worried about the throw though. From what I see it's part of the graph stage, so this will be propagated to the stream, so it might be that we will lose internal state depending on the error handling strategy set on the stream.

nicu-da (approved):
LGMT! 
Nice work, only have a few things that i pointed out.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: The submission id there is not empty, as it is freshly generated and passed in the request. I think that's the only place we can find out whether it's the mutable schema.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: I can return this as a completion failure with an internal error. WDYT?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: Renaming to `completionDescription`, as `tracing` may be confused with telemetry.

hubert-da:
Trigger-related tests are failing and I don't know how to fix them yet, for some reason sometimes there are no `pendingCommands`.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
nicu-da: Sounds good to me.

The problem is how do we complete all of them and remove them from pending? Do we complete just one? Which one in that case? 
Do we remove all of them? What happens when we get a completion that has no pending command in that case

nicu-da: The test failures seem to be related to this actually. 
In the old code we just ignore completions that didn't have a pending command. And that makes sense as not all completions are tracked.
In the new code we assume that there is always a tracked command for a completion that doesn't have a submission id. 

In this case, when not having a submission id we should get all the pending commands for that command id and complete them with an INTERNAL failure, and remove them from the pending commands.

andreaslochbihler-da:
> Different/same submission IDs are now covered by `CommandTrackerFlowTest` and are going to be covered by the conformance test: #10869

AFAICT #10869 tests only the `CommandSubmissionService`. But it doesn't test that the command service doesn't get confused about the incoming command completions that differ only by their submission ID. It would also be great to mix parallel submissions to the `CommandService` and the `CommandSubmissionService` so that the `CommandService` is not aware of concurrent submissions with the same change ID.

> Anyway, the test coverage is probably not perfect but if we want to release it tomorrow, this is what we have. I think at least a minimal set of test cases already exists.

I agree. We should somewhere keep a list of the tests that we want to add so that we don't forget. Is there already a ticket for that?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
andreaslochbihler-da: > I can return this as a completion failure with an `INTERNAL` error. WDYT?

I like that idea better than an exception. `INTERNAL` works for now, but we may want to revisit this later (e.g., change it to `ABORTED` as part of the error code cleanup).


> Which one in that case?

If the application sets the same submission ID for several submissions, then it's the application's fault if it can't distinguish between them. We could also silently drop the request that reuses the submission ID, but that's a worse UX than signalling that something is off.



> What happens when we get a completion that has no pending command in that case

This can happen during regular operations if there are parallel submissions, some of which go through the `CommandService` and some hit the `CommandSubmissionService` directly. Again, if submission IDs are not unique, then the application will observe strange behaviour.

hubert-da: Thanks, done

hubert-da: Thanks, done

hubert-da:
> We should somewhere keep a list of the tests that we want to add so that we don't forget. Is there already a ticket for that?

I'll create a story to improve the test coverage here with a list of suggested test cases.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -197,33 +209,38 @@ private[commands] class CommandTracker[Context](
 
       import CommandTracker.nonTerminalCodes
 
-      private def handleSubmitResponse(submitResponse: Ctx[(Context, String), Try[Empty]]) = {
-        val Ctx((_, commandId), value, _) = submitResponse
+      private def handleSubmitResponse(
+          submitResponse: Ctx[(Context, TrackedCommandKey), Try[Empty]]
+      ): immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
```
fabiotudone-da: Maybe a `type SingleCompletionFlowOutput = Ctx[Context, Either[CompletionFailure, CompletionSuccess]]` and other similar typedefs can improve readability

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -249,19 +266,20 @@ private[commands] class CommandTracker[Context](
           commandTimeout = Instant.now().plus(commandTimeout),
           context = submission.context,
         )
-        pendingCommands += commandId -> trackingData
+        pendingCommands += TrackedCommandKey(submissionId, commandId) -> trackingData
         ()
       }
 
       private def getOutputForTimeout(instant: Instant) = {
```
fabiotudone-da, line 272: Can you specify the return type?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -275,39 +293,91 @@ private[commands] class CommandTracker[Context](
...
 
-      private def getOutputForCompletion(completion: Completion) = {
-        val (commandId, errorText) = {
-          completion.status match {
-            case Some(StatusProto(code, _, _, _)) if code == Status.Code.OK.value =>
-              completion.commandId -> "successful completion of command"
-            case _ =>
-              completion.commandId -> "failed completion of command"
-          }
+      private def getOutputForCompletion(
```
fabiotudone-da, line 296:
```suggestion
      private def getFlowOutputsForCompletion(
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -275,39 +293,91 @@ private[commands] class CommandTracker[Context](
...
+          completion: Completion
+      ): immutable.Iterable[Ctx[Context, Either[CompletionFailure, CompletionSuccess]]] = {
+        val completionDescription = completion.status match {
+          case Some(StatusProto(code, _, _, _)) if code == Status.Code.OK.value =>
+            "successful completion of command"
+          case _ => "failed completion of command"
```
fabiotudone-da, line 302: I'd move it inside `trace` and that should also make it lazy.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -275,39 +293,91 @@ private[commands] class CommandTracker[Context](
...
+          "Handling {} {} from submission {}.",
+          completionDescription,
+          commandId,
+          completion.submissionId,
+        )
+
+        val maybeSubmissionId = Option(completion.submissionId).collect {
+          case id if id.nonEmpty => id
+        }
+        val trackedCommands =
```
fabiotudone-da, line 317:
```suggestion
        val trackedCommandsForSubmissionId =
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -275,39 +293,91 @@ private[commands] class CommandTracker[Context](
...
+        if (trackedCommands.size > 1) {
+          trackedCommands.map { trackingData =>
+            Ctx(
+              trackingData.context,
+              Left(
+                NotOkResponse(
+                  commandId,
+                  StatusProto.of(
+                    Status.Code.INTERNAL.value(),
+                    s"There are multiple pending commands for the id $commandId. This can only happen for the mutating schema.",
```
fabiotudone-da, line 329:
```suggestion
                    s"There are multiple pending commands with ID $commandId for submission ID $maybeSubmissionId. This can only happen for the mutating schema that shouldn't be used anymore as it doesn't fully support command deduplication.",
```

fabiotudone-da: Or something more explanatory anyway.

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -453,27 +475,79 @@ class CommandTrackerFlowTest
       }
     }
 
+    "a completion without submission id arrives" should {
```
fabiotudone-da:
```suggestion
    "an incoming completion without submission id" should {
```

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -453,27 +475,79 @@ class CommandTrackerFlowTest
...
+          unhandled should contain(
+            TrackedCommandKey("submissionId", "commandId") -> submission.context
+          )
+          unhandled should contain(
+            TrackedCommandKey("anotherSubmissionId", "commandId") -> submission.context
+          )
+        }
+      }
+
+      "output the completion" in {
```
fabiotudone-da:
```suggestion
      "include the completion in the flow output" in {
```

fabiotudone-da (approved):
Minor clarity improvement suggestions, still OK to go IMO.

hubert-da:
Thanks @fabiotudone-da for the comments, will address in a separate PR.

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
hubert-da: I think this doesn't work with the current naming.

It would be:
`Command tracking flow when an incoming completion without submission id should...`
vs existing:
`Command tracking flow when a completion without the submission id arrives should...`

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
hubert-da: This name is consistent with other tests, so I think I'll leave it.

hubert-da:
> AFAICT #10869 tests only the CommandSubmissionService. But it doesn't test that the command service doesn't get confused about the incoming command completions that differ only by their submission ID. It would also be great to mix parallel submissions to the CommandService and the CommandSubmissionService so that the CommandService is not aware of concurrent submissions with the same change ID.

@andreaslochbihler-da, `AppendOnlyCommandDeduplicationParallelIT` now supports both `CommandSubmissionService` and `CommandService`. I also added a test case for `command completions that differ only by their submission ID` here: https://github.com/digital-asset/daml/pull/10939. I think it now makes sense to just add a test that mixes both services, so I created a story for that KVL-1106.

hubert-da:
@fabiotudone-da, your comments are addressed by https://github.com/digital-asset/daml/pull/10868, which will be ready to review soon, so I can finally leave this PR.

PR #10875, opened 2021-09-14
Prevent concurrent access to recorded request in mock

stefanobaghino-da:
The code above has been a source of flakes, initially apparently fixed
by the addition of `@volatile` to `submittedRequest` that for some reason
reappeared on MacOS CI nodes as of recently. This is probably not the
best fix for the issue, as these mocks were probably originally
thought to be used just once and have been adapted to be re-used
across multiple test calls, so feel free to suggest a different
approach.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stefanobaghino-da:
Converted to draft as the flake is still present (tested locally with `runs_per_test=80`, failed 30 times).

stefanobaghino-da:
This goes beyond my capabilities at the moment, I have no idea of why this is failing. I suspect the fact that we are binding multiple service through the `global` execution context (which is backed by a `ForkJoinPool`) could be the source of unwanted concurrency but it's just an hypothesis.

PR #10879, opened 2021-09-14
For the client binding propagate the full original completion [KVL-1112]

nicu-da:
Because more fields are added to the completion it gets more difficult to deconstruct it into our own models while keeping the same external API. Because of this we are propagating the full completion as well.

CHANGELOG_BEGIN

CHANGELOG_END

### Pull Request Checklist

trigger the build.


fabiotudone-da (commented):
Why is this needed at this stage? Doesn't this change the bindings API anyway?

nicu-da:
> Why is this needed at this stage? Doesn't this change the bindings API anyway?

The only change to that is the removal of the extractors for the rejections.

I would say this is needed to propagate the completion to clients, as we enriched the completion with multiple fields, but those are not part of our internal models.

stefanobaghino-da (approved):

mziolekda (approved):

ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
```diff
@@ -539,6 +549,8 @@ class CommandTrackerFlowTest
     }
 
   }
+  def failureCompletion(code: Code) =
```
miklos-da, line 552: Nit:
```suggestion
  private def failureCompletion(code: Code): CompletionFailure =
```

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -70,22 +70,18 @@ object CompletionResponse {
     response match {
```
miklos-da, line 70: Who depends on this method? Shouldn't we deprecate this?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -43,25 +47,21 @@ object CompletionResponse {
...
-      commandId: String,
-      transactionId: String,
-      originalStatus: StatusProto,
-  )
+      completion: Completion
+  ) {
 
+    lazy val commandId: String = completion.commandId
+    lazy val transactionId: String = completion.transactionId
+    lazy val originalStatus: StatusProto = completion.getStatus
```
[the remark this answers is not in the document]
S11001001:
```suggestion
    import completion.{commandId, transactionId, getStatus => originalStatus}
```

This should generally be done in all cases where you simply want to alias away stable property lookups.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
nicu-da: Good point. 
Canton is the dependent that actually needs this.

I will open a new PR once this is merged, which deprecates this usage.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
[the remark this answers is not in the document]
nicu-da: @S11001001 This is actually to keep the same external API so I don't think imports would work.
@stefanobaghino-da Done

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CompletionResponse.scala
```diff
@@ -43,25 +47,21 @@ object CompletionResponse {
   private[daml] final case class QueueSubmitFailure(status: Status) extends TrackedCompletionFailure
 
   final case class CompletionSuccess(
-      commandId: String,
-      transactionId: String,
-      originalStatus: StatusProto,
-  )
+      completion: Completion
```
hubert-da, line 50: Nit: can probably be formatted to a one line

hubert-da (approved):
LGTM, just needs a conflict resolution

PR #10822, opened 2021-09-08
ledger-api: Remove unimplemented fields [KVL-1094]

miklos-da:
Summary of changes:
* Removed `Commands.deduplication_offset` and `Completion.deduplication_start` as we don't support deduplication offsets yet.

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10884, opened 2021-09-14
Rebuild the `buf` image as part of the SDK release process [KVL-1088]

fabiotudone-da:
## Description

- Ensures that all protobuf changes released as part of an SDK become stable.
- Avoids the need to manually update the `buf` image to stabilize protobuf changes.
- Still allows manually introducing breaking changes in an auditable way.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/fmt.sh
+++ b/fmt.sh
@@ -11,16 +11,16 @@ cd "$(dirname "$0")"
 # load the dev-env
 eval "$(dev-env/bin/dade-assist)"
 
-# Location of reference image used for buf breaking check.
+# Location of reference image used for `buf`'s protobuf breaking checks, including the Ledger API's.
 #
-# It should be re-built by running './fmt.sh --rebuild-buf-image' and then committed:
+# This `buf` image will be re-built and committed automatically as part of the SDK release process;
+# this means that all protobuf changes in the next SDK development cycle will be checked by `buf`
+# against the new SDK's protobufs, effectively considering them stable.
 #
-# 1. as part of a commit also containing breaking proto changes (that have been agreed to), and/or
-# 2. as part of a subsequent commit (e.g. in the same PR, in a later PR or as part of the SDK release process),
-#    when already committed proto changes become stable.
+# The `buf` image should be re-built manually only when making breaking protobuf changes that have
+# been agreed to and it should be part of the same commit also containing such changes.
 #
-# Proto changes that are part of an SDK release (especially RC and stable ones) should also be included in
-# the buf image.
+# To re-build it manually, run './fmt.sh --rebuild-buf-image'.
 
 buf_image="buf-stable-protos-image.bin"
 
@@ -45,7 +45,7 @@ run() {
   if [[ $is_test = 1 && $ret -gt 0 ]]; then
     log "command failed with return $ret"
     log
-    log "run ./fmt.sh to fix the issue"
+    log "run ./fmt.sh to fix formatting issues or ./fmt.sh --rebuild-buf-image if you intend to make a breaking change"
     exit 1
   fi
   return 0
--- a/release.sh
+++ b/release.sh
@@ -75,10 +75,13 @@ $0 snapshot SHA PREFIX
 
         Any non-ambiguous git commit reference can be given as SHA.
 
+$0 prepare snapshot
+        Prepares a snapshot; currently, this step updates the buf image.
+
 $0 new snapshot
-        Updates LATEST to add current commit as a new snapshot. Figures out
-        prefix version by keeping the same if the first line in LATEST is a
-        snapshot, and incrementing minor if the first line is stable.
+        Prepares a snapshot and updates LATEST to add current commit as a new snapshot.
+        Figures out prefix version by keeping the same if the first
+        line in LATEST is a snapshot, and incrementing minor if the first line is stable.
 
 $0 check
         Checks that each line of the LATEST file is well-formed.
@@ -99,7 +102,13 @@ commit_belongs_to_release_branch() {
       | grep -q -E '^origin/(main$|release/)'
 }
 
+prepare_snapshot() {
+    (./fmt.sh --rebuild-buf-image)
+}
+
 new_snapshot () {
+    prepare_snapshot
+
     local sha latest latest_prefix prefix new_line tmp
     sha=$(git rev-parse HEAD)
     if ! commit_belongs_to_release_branch $sha; then
@@ -143,6 +152,13 @@ case $1 in
                 echo "WARNING: Commit does not belong to a release branch."
             fi
             make_snapshot $(git rev-parse $2) $3
+        else
+            display_help
+        fi
+    ;;
+    prepare)
+        if [ $# -eq 2 ] && [ "$2" == 'snapshot' ]; then
+            prepare_snapshot
         else
             display_help
         fi
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -32,7 +32,7 @@ patches we backport to the 1.0 release branch).
    $ ./release.sh snapshot cc880e2 0.1.2
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR _to ber merged into the `main` branch_ (even if it's for a maintenance release)
+   Then open a PR _to be merged into the `main` branch_ (even if it's for a maintenance release)
    that should contain:
 
      - The files changed by the `./release.sh prepare snapshot` invocation above.
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -24,6 +24,7 @@ patches we backport to the 1.0 release branch).
    If you are manually creating the PR for an out-of-schedule snapshot, start
    _from latest `main`_ and run
    ```
+   ./release.sh prepare snapshot
    ./release.sh snapshot <sha> <prefix>
    ```
    for example:
@@ -31,13 +32,17 @@ patches we backport to the 1.0 release branch).
    $ ./release.sh snapshot cc880e2 0.1.2
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR _to be merged to `main`_ (even if it's for a maintenance release)
-   with the changed `LATEST` file, add the line produced by the `release.sh`
-   invocation in a meaningful position (if you’re not sure, [semver](https://semver.org/) ordering is
-   probably the right thing to do) and add the `Standard-Change` label. It
-   is better to add such a label _before confirming the PR's creation_, else
-   the associated CI check will fail and merging the PR will require you to
-   re-run it after all the other ones have completed successfully.
+   Then open a PR _to ber merged into the `main` branch_ (even if it's for a maintenance release)
+   that should contain:
+
+     - The files changed by the `./release.sh prepare snapshot` invocation above.
+     - The addition to `LATEST` in a meaningful position (if you’re not sure,
+       [semver](https://semver.org/) ordering is probably the right thing to do)
+       of the line produced by the `release.sh snapshot` invocation above.
+
+   Add the `Standard-Change` label _before  confirming the PR's creation_
+   (else the associated CI check will fail and merging the PR will require
+   you to re-run it after all the other ones have completed successfully).
 
 1. Once the PR has built, check that it was considered a release build by our
    CI. If you are working from an automated PR, check that it sent a message to
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -40,7 +40,7 @@ patches we backport to the 1.0 release branch).
        [semver](https://semver.org/) ordering is probably the right thing to do)
        of the line produced by the `release.sh snapshot` invocation above.
 
-   Add the `Standard-Change` label _before  confirming the PR's creation_
+   Add the `Standard-Change` label _before confirming the PR's creation_
    (else the associated CI check will fail and merging the PR will require
    you to re-run it after all the other ones have completed successfully).
 


release/RELEASE.md
```diff
@@ -24,20 +24,25 @@ patches we backport to the 1.0 release branch).
...
    cc880e290b2311d0bf05d58c7d75c50784c0131c 0.1.2-snapshot.20200513.4174.0.cc880e29
    ```
-   Then open a PR _to be merged to `main`_ (even if it's for a maintenance release)
-   with the changed `LATEST` file, add the line produced by the `release.sh`
-   invocation in a meaningful position (if you’re not sure, [semver](https://semver.org/) ordering is
-   probably the right thing to do) and add the `Standard-Change` label. It
-   is better to add such a label _before confirming the PR's creation_, else
-   the associated CI check will fail and merging the PR will require you to
-   re-run it after all the other ones have completed successfully.
+   Then open a PR _to ber merged into the `main` branch_ (even if it's for a maintenance release)
```
fabiotudone-da, line 35:
```suggestion
   Then open a PR _to be merged into the `main` branch_ (even if it's for a maintenance release)
```

release/RELEASE.md
```diff
@@ -24,20 +24,25 @@ patches we backport to the 1.0 release branch).
...
+   that should contain:
+
+     - The files changed by the `./release.sh prepare snapshot` invocation above.
+     - The addition to `LATEST` in a meaningful position (if you’re not sure,
+       [semver](https://semver.org/) ordering is probably the right thing to do)
+       of the line produced by the `release.sh snapshot` invocation above.
+
+   Add the `Standard-Change` label _before  confirming the PR's creation_
```
fabiotudone-da, line 43:
```suggestion
   Add the `Standard-Change` label _before confirming the PR's creation_
```

cocreature (commented):
I think I need more context for this. If I understand it correctly this means that each release checks against the buf image of the latest snapshot. Why is that the correct thing to do, wouldn’t the latest stable release make more sense? And is it a good idea to rebuild the image automatically on each release? If nothing changed it doesn’t seem necessary and if something did change, rebuilding it manually might make more sense than doing it implicitly on the release?

fmt.sh
```diff
@@ -11,16 +11,16 @@ cd "$(dirname "$0")"
...
 #
-# 1. as part of a commit also containing breaking proto changes (that have been agreed to), and/or
-# 2. as part of a subsequent commit (e.g. in the same PR, in a later PR or as part of the SDK release process),
-#    when already committed proto changes become stable.
+# The `buf` image should be re-built manually only when making breaking protobuf changes that have
+# been agreed to and it should be part of the same commit also containing such changes.
 #
-# Proto changes that are part of an SDK release (especially RC and stable ones) should also be included in
-# the buf image.
+# To re-build it manually, run './fmt.sh --rebuild-buf-image'.
```
fabiotudone-da, line 23: This sets up the workflow as follows:

- `buf` will block breaking changes during the SDK development cycle; this can be manually overridden (by rebuilding and committing the `buf` image) and leaves a trace in the history. One caveat is that breaking changes are immediately "stabilized".
- The release process relies on the above and doesn't check for breaking changes, it just stabilizes the new protos. If needed, we can improve that by e.g. checking unless a `BREAKING/PROTO` file is present.

fabiotudone-da:
@cocreature:

> I think I need more context for this. If I understand it correctly this means that each release checks against the buf image of the latest snapshot. Why is that the correct thing to do, wouldn’t the latest stable release make more sense?

I am assuming that the following holds in practice:

- Most of the time we want to avoid breaking protos between snapshots (maybe we should really distinguish between protos, esp. ledger API and other protos though), esp. considering we publish snapshots too.
- Most of the time we won't break compatibility between snapshots but then maintain compatibility with the release. If we do, we probably want to do it (i.e., we are aware of it) and can still manually reset the `buf` image to that of the release whenever we are ready to check compatibility again.

> And is it a good idea to rebuild the image automatically on each release? If nothing changed it doesn’t seem necessary and if something did change, rebuilding it manually might make more sense than doing it implicitly on the release?

If nothing changed the `buf` image won't change. The problem of doing it manually is that you can forget and then assume that everyone is checking against your (compatible) proto changes when they are actually not. This ties the protos stabilization cycle to at least the SDK release one.

That said, the root of the matter is really at the process level and I welcome any feedback because I don't think that use cases are well-defined yet.

cocreature:
Alright, that workflow seems sensible. Should we have some check on CI to make sure that manually created releases also update this buf image?

garyverhaegen-da:
This will not work as is: if the buf image has changed, the PR will not be recognized as a release PR (they have to change *exactly* one line in the `LATEST` file).

I'm not convinced trying to add this to automated tools is the right approach; have you considered just adding steps to the release process?

I know nothing about protobuf or what this "buf image" is, but stabilizing breaking changes sounds like something that should be a deliberate decision, not something that happens automatically at arbitrary points in time.

fabiotudone-da:
@cocreature 

> Should we have some check on CI to make sure that manually created releases also update this buf image?

Good point.

fabiotudone-da:
@garyverhaegen-da:

> I'm not convinced trying to add this to automated tools is the right approach; have you considered just adding steps to the release process?

Yes but I agree with e.g. @SamirTalwar-DA that we should really try to automate anything that can be automated (once understood), especially things that can have unintended bad consequences if not done right or if forgotten: one problem of doing it manually is that you can forget and then assume that everyone is checking against your (compatible) proto changes when they are actually not.

> I know nothing about protobuf or what this "buf image" is

Think of it as an binary AST representation of the protos at a given point in time (a snapshot) that is used to check backward compatibility (w.r.t. that snapshot) of proto changes.

> stabilizing breaking changes sounds like something that should be a deliberate decision, not something that happens automatically at arbitrary points in time.

That's what happens indeed as the check was (and still is) run by `fmt.sh` that is invoked by CI.

garyverhaegen-da:
> @garyverhaegen-da:
> 
> > I'm not convinced trying to add this to automated tools is the right approach; have you considered just adding steps to the release process?
> 
> Yes but I agree with e.g. @SamirTalwar-DA that we should really try to automate anything that can be automated (once understood), especially things that can have unintended bad consequences if not done right or if forgotten: one problem of doing it manually is that you can forget and then assume that everyone is checking against your (compatible) proto changes when they are actually not.
> 
> > I know nothing about protobuf or what this "buf image" is
> 
> Think of it as an binary AST representation of the protos at a given point in time (a snapshot) that is used to check backward compatibility (w.r.t. that snapshot) of proto changes.
> 
> > stabilizing breaking changes sounds like something that should be a deliberate decision, not something that happens automatically at arbitrary points in time.
> 
> That's what happens indeed as the check was (and still is) run by `fmt.sh` that is invoked by CI.

I'm still confused. If you want this automated, why not enforce it on every commit? Why wait for releases?

Or, conversely, if you mean that `fmt.sh` is already enforcing it at every commit, why do we need to do anything special on release?

release.sh
```diff
@@ -99,7 +102,13 @@ commit_belongs_to_release_branch() {
       | grep -q -E '^origin/(main$|release/)'
 }
 
+prepare_snapshot() {
+    (./fmt.sh --rebuild-buf-image)
```
garyverhaegen-da, line 106: Why do we need a subshell here?

garyverhaegen-da:
If I run `./fmt.sh --rebuild-buf-image` right now it changes the `buf-stable-protos-image.bin` file. Is that expected? Is it the intention that that file will now change with every single release, because the process is non-reproducible, or are we currently in a state where the image does not match our source files?

fabiotudone-da:
@garyverhaegen-da:

> I'm still confused. If you want this automated, why not enforce it on every commit? Why wait for releases?

I want to automate stabilizing protos (we do by rebuilding and committing the `buf` image), so that I can leverage compatibility checks against them [without having to remember doing it], whenever they should be stabilized. The definition of this moment is tricky, but there is some consensus that an SDK [snapshot] release could be it.

Conversely, during the SDK development cycle, I want breaking changes to be blocked (by running the `buf` checks), so that I don't break compatibility unintendedly; that is, unless I override.

This is only my view of the use cases as a developer.

cocreature:
I think it’ s useful to spell out what updating the buf image accomplishes:
On each commit we check that changes are compatible with the latest buf image we have checked in. Lets look at the following steps:

1. I make an addition and add a field. That’s compatible. 
2. Now I change that field to have a different type. That’s in general not compatible with the change I made in 1. But if we did not update the image after 1, then it’s compared to the state where the field did not exist and it is compatible.

Updating on each snapshot solves this to some degree but picking the snapshot is somewhat arbitrary given that our snapshots are done at fairly arbitrary points.

It seems like their are two parts here:
1. On each commit check against the buf image from the previous commit. If that fails, it needs manual auditing.
2. After each change, update immediately to the now current state. We do this for both compatible and incompatible (approved) changes.

The main reason for only doing it on releases seems a bit of noise if you’re iterating on changes between snapshots but I’m not sure that’s worth the complexity. Wdyt @fabiotudone-da

release.sh
fabiotudone-da: Because of the way `fmt.sh` is written (can be improved but wouldn't do it here) it `exit`s after the check to avoid also running the formatting logic.

cocreature:
I guess for a reviewer there is an issue in that it’s hard to see just from the diff if we hit 1 or not.

release.sh
garyverhaegen-da: I fail to see the connection. That sounds like a good thing. 🤔

fabiotudone-da: `exit` from `fmt.sh` would fail to execute the rest of this script, like changing `LATEST`.

fabiotudone-da:
@cocreature:

> The main reason for only doing it on releases seems a bit of noise if you’re iterating on changes between snapshots but I’m not sure that’s worth the complexity. Wdyt @fabiotudone-da

My point exactly, especially considering that I don't see esp. bad side effects arising from this simplification.

moritzkiefer-da:
> > 
> 
> My point exactly, especially considering that I don't see esp. bad side effects arising from this simplification.

I think you misunderstood me. I think doing it on releases is extra complexity not doing it on each commit :)

release.sh
garyverhaegen-da: I don't think it would:

```bash
$ echo "exit 1" > exit.sh
$ chmod +x exit.sh 
$ ./exit.sh 
$ 
```

fabiotudone-da:
> I think you misunderstood me. I think doing it on releases is extra complexity not doing it on each commit :)

I see; doing it on non-snapshot releases is much more complex though :) Doing it on every commit is that it increases the noise a lot; I think the snapshot release cadence is a good compromise but the matter is really what's the meaningful unit of change of protobufs that we want to consider.

garyverhaegen-da:
To be clear, my main concern is that as it stands this PR breaks our release process in a non-obvious way, i.e. CI will not produce an error but no release will be produced.

I don't have strong opinions on how to manage protobuf lifecycle as I'm not working on that at all, but I do care about the release process not getting broken. Therefore my preference would be for this to not touch the release process (and thus protobuf lifecycle to not be toed to the release process), because then it's much easier to verify it doesn't break it.

As it stands, and from what I understand so far, I can see at least two ways to move forward:

1. Do the check on each commit. The way I would approach it is to not save the image, but instead have a CI step that builds from the parent (main-branch) commit and check against that. Have a GitHub label to skip that check, and block PRs that break compatibility unless the label is there (like we do for Standard-Change). That way it's easy for reviewers to spot, and it's easy to audit for after the fact.
2. Fix this PR so that it does not break the release process. I think that's an inferior solution, but I'm not going to block this _if the release process keeps working_. For that, you'd need to change the [check that CI does to figure out if a PR is a release PR](https://github.com/digital-asset/daml/blob/main/ci/build.yml#L416) to take this into account, i.e. essentially just check for both LATEST _and_ the buf image to change (and nothing else). You'd also want to add some provisions for manual PRs, for example fail CI if LATEST has changed but not the image.

release.sh
fabiotudone-da: I am no `bash` expert, I'll re-test it and look for an explanation.

garyverhaegen-da:
>Doing it on every commit increases the noise a lot

How often do you guys make breaking changes to the protobuf defintions? 😲

cocreature:
> I see; doing it on non-snapshot releases is much more complex though :) Doing it on every commit is that it increases the noise a lot; I think the snapshot release cadence is a good compromise but the matter is really what's the meaningful unit of change of protobufs that we want to consider.

I’m interested in the "increases the noise" a lot point. The main reason where I see it increasing noise is if we have commit A and B between snapshot S1 and S2 where B breaks a change introduced by A but not the state in S1, i.e., we iterate on a change in a breaking way between releases but the end result is non-breaking. Do we have  concrete examples of that happening?

In terms of complexity, I think managing things that happen on a per-commit basis is simpler to handle and you get notified of changes sooner since the image is updated immediately afterwards.

The main caveat is that it’s hard for reviewers. I don’t have a great solution for that. We could have two buf images with one "lagging" behind I guess but that gets a bit tricky if things get merged in parallel.

fabiotudone-da:
> In terms of complexity, I think managing things that happen on a per-commit basis is simpler to handle and you get notified of changes sooner since the image is updated immediately afterwards.

If you do it on a per-commit basis, then the only way to avoid forgetting would be to automate updating the image every time you make a compatible change; in addition to that, this would also mean that subsequent proto changes incompatible with that change (but compatible with the initial state) would need to override.

release.sh
garyverhaegen-da: Definitely not an expert either, but here's my explanation: any Bash command is calling a subprocess, and waiting for that subprocess to exit. A Bash script that does not have an explicit `exit` call still actually calls `exit` at the end of the file. That's true of any programming language: if you don't explicitly call exit, there's an implicit exit(0) at the end of your program.

There's no special case for when the program Bash starts is also a Bash process; it would be quite hard to use if it inlined every other shell script.

Note that none of that holds if you explicitly inline another Bash file, either through the `source` command or its alias `.`; i.e. `./fmt.sh` is a separate child process we wait on, whereas `. fmt.sh` is running within the current process.

garyverhaegen-da:
> The main caveat is that it’s hard for reviewers.

Wouldn't CI enforcing a label on changes to the buf file solve that?

cocreature:
> Wouldn't CI enforcing a label on changes to the buf file solve that?

I think you want a label for incompatible changes only not on compatible changes but that still seems doable. I guess my worry is that people are not used to paying attention to labels atm.

garyverhaegen-da:
> > Wouldn't CI enforcing a label on changes to the buf file solve that?
> 
> I think you want a label for incompatible changes only not on compatible changes but that still seems doable.

Well, you could enforce the label only on breaking changes by comparing with the parent commit, as I suggested above.

> I guess my worry is that people are not used to paying attention to labels atm.

That is simply absurd. We use labels to mark **Standard Changes**, _of course_ everyone pays attention to them.

cocreature:
> If you do it on a per-commit basis, then the only way to avoid forgetting would be to automate updating the image every time you make a compatible change

Yep, that’s what I’d do and enforce via CI. I think I don’t really see the issue? You run ./fmt.sh before every push already. Whether that updates an extra file or not doesn’t seem more noisy.

> in addition to that, this would also mean that subsequent proto changes incompatible with that change (but compatible with the initial state) would need to override.

Right this gets back at my earlier question of how frequent this actually is. Given that snapshots happen at pretty arbitrary points in time you cannot really rely on being able to get two changes in between snapshots anyway so in theory each change should stand on its own but ofc you might reconsider things here and there.

Maybe let me expand on why I’m hesitant to add things to the release process: Anything tied to a release needs to work across branches. An old release is still made from `main`. So you need to worry about backwards compatibility in any check or logic you have there. It also becomes very tricky to test it short of making a release.
Doing things on a per-commit basis is much simpler from an infrastructure perspective. It never runs on old code, it’s easy to test just via a PR.

fabiotudone-da:
> Whether that updates an extra file or not doesn’t seem more noisy.

The idea here is that changes to the image act as as an audit of breaking changes (or releases) but maybe there is another way.

> Doing things on a per-commit basis is much simpler from an infrastructure perspective.

I'll think it through once more from the start in light of your feedbacks. Thanks!

garyverhaegen-da:
> The idea here is that changes to the image act as as an audit of breaking changes (or releases) but maybe there is another way.

But you'll also get non-breaking changes in there. If instead you enforce a label on breaking changes (or a specific line in commit messages, if you want this queryable locally), then you can trust that you only get breaking changes in that log.

fabiotudone-da:
> But you'll also get non-breaking changes in there.

This would also happen for any commit that doesn't just include the incompatible bit.

Let me rephrase: it would be nice to distinguish points in (source) history where at least one breaking change was introduced (and the possibility to examine what exactly).

> If instead you enforce a label on breaking changes (or a specific line in commit messages, if you want this queryable locally), then you can trust that you only get breaking changes in that log.

Yeah, I'll revisit in that direction. I'd definitely like this be locally queryable and somehow part of source control.

meiersi-da:
@fabiotudone-da : one other point to consider, is whether we need to publish the `buf` image at all. The alternative that @miklos-da proposed is to derive the `buf` image from the relevant .proto files of the latest stable SDK release. Have you considered this approach?

@moritzkiefer-da : do we already have build targets that expose the files from the latest stable release of the SDK that could be used as an example?

cocreature:
> @moritzkiefer-da : do we already have build targets that expose the files from the latest stable release of the SDK that could be used as an example?

The compat workspace does something like this but it gives you access to released artifacts no the source tree. I guess you can build the buf image from the release protobufs or something like that.

fabiotudone-da:
I think that, first of all, we should clarify what exactly we want to achieve in terms of workflow. From my (dev/engineering) point of view:

1. I want backwards-compatible protobuf changes to be incorporated in checks at sensible times (that may well depend on which protobuf we consider), as well as the ability to override; as an alternative, I'd rather have them incorporated all the time and manually revert to a previous state whenever needed.
2. I want to be always prevented from making backwards-incompatible protobuf changes, unless I override.
3. I want all of the above to be auditable (better if locally and without relying only on collab features), at least in the sense that I can figure out easily enough in which commits reference protobuf bumps happened, whether they were automatic or manual and whether they included backwards-incompatible changes (and which ones).
4. I want the mechanisms put in place for the above to be highly reliable, testable, maintainable and low maintenance (bonus: low or at least benefit-proportional effort).

@meiersi-da:

> one other point to consider is whether we need to publish the buf image at all. The alternative that @miklos-da proposed is to derive the buf image from the relevant .proto files of the latest stable SDK release. Have you considered this approach?

Assuming the above use cases make sense, and in particular their manual override portions, we need to keep some state in between releases; `buf` images seemed like a good fit for that to me.

> Yes the kind of infrastructure in the `/compatibility` workspace looks like a good fit

I don't know how it works, I'll have a look.

But first, I propose we agree on goals: if you think the use cases or the workflow should be different, I'd rather take a step back at the drawing board and iterate over a proposal. This PR wasn't expected to be more than an effective way to stimulate discussion on that.

cocreature:
The goals sound sensible to me. If it’s helpful, I’m also happy to do some brainstorming in a zoom call for different options.

meiersi-da:
> I think that, first of all, we should clarify what exactly we want to achieve in terms of workflow. From my (dev/engineering) point of view:
> 
> 1. I want backwards-compatible protobuf changes to be incorporated in checks at sensible times (that may well depend on which protobuf we consider), as well as the ability to override; as an alternative, I'd rather have them incorporated all the time and manually revert to a previous state whenever needed.
> 2. I want to be always prevented from making backwards-incompatible protobuf changes, unless I override.
> 3. I want all of the above to be auditable (better if locally and without relying only on collab features), at least in the sense that I can figure out easily enough in which commits reference protobuf bumps happened, whether they were automatic or manual and whether they included backwards-incompatible changes (and which ones).
> 4. I want the mechanisms put in place for the above to be highly reliable, testable, maintainable and low maintenance (bonus: low or at least benefit-proportional effort).

Thanks for stating these goals. I had the following one requirement in mind:

> I want CI in combination with our release process to ensure for the `main` branch of the `daml` repo that the  gRPC and proto definitions listed below are backwards compatible with all stable SDK releases:
  a. `daml-lf/archive/src/main/protobuf`
  b. `daml-lf/transaction/src/main/protobuf`
  c. `daml/legder-api/grpc-definitions`
  d. `ledger/ledger-configuration/protobuf`
  e. `ledger/participant-integration-api/src/main/protobuf`
  f. `ledger/participant-state/kvutils/src/main/protobuf`  

It seems that the major difference with your goals is that you'd like the ability to override. Why do you need the overriding functionality? 

Given that we must maintain backwards compatibility on these protos for our customers, and have the ability to always introduce messages and gRPC services to the side of the existing ones, I don't understand the need for this overriding functionality.

Also note that while the requirement sounds like having to do n-squared comparisons, I believe that there are better implementation options, including the one of checking only against the latest stable release.

fabiotudone-da:
> Why do you need the overriding functionality?

1. We finish a feature mid-release and we want to make sure nobody breaks its protobuf increment.
2. After having considered everything, we decide nevertheless to introduce somewhere a protobuf breaking change in the upcoming release.
3. Situations we haven't thought of yet.

garyverhaegen-da:
> > Why do you need the overriding functionality?
> 
> 1. We finish a feature mid-release and we want to make sure nobody breaks its protobuf increment.
> 2. After having considered everything, we decide nevertheless to introduce somewhere a protobuf breaking change in the upcoming release.
> 3. Situations we haven't thought of yet.

The repo is currently operating under the assumption that the main branch is always in a releasable state. When designing workflows, I'd much prefer if we kept that assumption and design the workflows from the perspective that releases can be done at any time just by taking the tip of main. I really don't like the idea of making releases special events we have to prepare for.

fabiotudone-da:
> The repo is currently operating under the assumption that the main branch is always in a releasable state. When designing workflows, I'd much prefer if we kept that assumption and design the workflows from the perspective that releases can be done at any time just by taking the tip of main. I really don't like the idea of making releases special events we have to prepare for.

Eh, I think you @garyverhaegen-da are making the point for a further (valid and strong) requirement here, especially if we want to keep/move in the (good) direction of frequent/continuous delivery/ability.

garyverhaegen-da:
> Eh, I think you @garyverhaegen-da are making the point for a **further** (valid and strong) requirement here, especially if we want to keep/**move in the (good) direction of** frequent/continuous delivery/ability.

To clarify, this is an assumption that we have been operating under for the past two years or so and that is built into a lot of the existing processes and CI jobs around this repo. It's not an aspiration. Breaking that assumption would have ramifications I really don't want to think through, so my "additional" requirement here is that you don't change the status quo on that front.

fabiotudone-da:
> To clarify, this is an assumption that we have been operating under for the past two years or so and that is built into a lot of the existing processes and CI jobs around this repo. It's not an aspiration. Breaking that assumption would have ramifications I really don't want to think through, so my "additional" requirement here is that you don't change the status quo on that front.

It's clear, although in practice I don't think we are currently always ready to happily sign on every PR merged to `main` as a potential stable release, especially when long-running epics are in progress, i.e. sometimes there are temporary compromises. I think you made clear that that should be the case, if my understand is correct.

garyverhaegen-da:
> It's clear, although in practice I don't think we are currently always ready to happily sign on every PR merged to `main` as a potential stable release, especially when long-running epics are in progress, i.e. sometimes there are temporary compromises. I think you made clear that that should be the case, if my understanding is correct.

All I have to say on that front is "please stop doing that" 😉

We do very much assume main is generally speaking releasable. For example, today we're making a snapshot release, and we're just picking main at some random time in the day. We're not going around and asking people to please stop merging in-progress work while we do that.

If you really have to split something in multiple PRs (which is a good thing) in such a way that some intermediate state is not fully working (that's less good), you should hide the non-working state behind a compile-time, by-default-off feature flag.

fabiotudone-da:
> All I have to say on that front is "please stop doing that"

Consider that this can also happen after the fact, i.e.: we think that a compatible protobuf change is good (e.g. a field addition addition) and we implement in a PR; after it's merged, we realize that this was not such a good idea and merge PR with a backwards-incompatible protobuf change (see #10822 and another case is upcoming).

fabiotudone-da:
For this reason, if we want to be really strict about observing the "every `main` commit is a [potential] stable release" mantra (and I see the value in that), I see feature branches as the only practical way.

garyverhaegen-da:
> Consider that this can also happen after the fact, i.e.: we think that a compatible protobuf change is good (e.g. a field addition) and we implement in a PR; after it's merged, we realize that this was not such a good idea and merge a new PR with a backwards-incompatible protobuf change (see #10822 and another case is upcoming).

Backwards compatibility is hard, I'm definitely not trying to downplay that. The issue here is you cannot guarantee that you won't have a release in-between these two commits, so the second one really does have to be treated as a breaking change. (That's still true _even after you've merged the second one_, as we do releases from past commits.)

I'm not familiar enough with protobufs, what we do with it, and how we expose it, to suggest specific techniques, but in my experience a combination of feature flags (for code) and _changing names_ (for external APIs) is the way to go to keep backwards compatibility. If you need a different type, use a different name.

> For this reason, if we want to be really strict about observing the "every `main` commit is a [potential] stable release" mantra (and I see the value in that), I see feature branches as the only practical way.

There is nothing I can do to stop you from having long-running branches. In my experience they cause a lot more problems than they solve, and I've yet to see a situation where they are actually needed. I can immediately see all kinds of problems they would cause, and it's not obvious to me _at all_ how they solve the above issue of realizing there's a problem _after_ having merged into main.

I will not, however, approve any changes to the CI system to _support_ feature branches.

fabiotudone-da:
> it's not obvious to me at all how they solve the above issue of realizing there's a problem after having merged into main

I agree they are heavyweight and have lots of problems associated but, in combination with defining the smallest possible product increments, they can mitigate the issue (provided you also integrate very frequently from `main`) in the sense that they would be closed when an epic has really been completed, checked and signed off from a product perspective. Errors may still happen, true, but would hopefully be less frequent because they'd (mostly) mirror issues in defining the product increment.

I agree that the other option would be to treat every protobuf-changing PR as a backward compatibility commitment, I just wonder if that wouldn't end up being an even heavier overhead at our current stage.

This is a complex discussion and partially OT but committing to always being strict, no matter how we do it, definitely affects how we should define the workflow for protobuf checks.

cocreature:
@fabiotudone-da can we close this now?

PR #10895, opened 2021-09-15
Desugar interface implements declarations

cocreature:
This PR adds desugaring for tplImplements. This consists of the
corresponding typeclass instance (which we just ignore in LF for now,
we probably need it once we have pure functions) and a _implements_
top-level value.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
+++ b/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
@@ -532,9 +532,9 @@ convertTypeDef env o@(ATyCon t) = withRange (convNameLoc t) $ if
     , n `elementOfUniqSet` internalTypes
     -> pure []
     | NameIn DA_Internal_Prelude "Optional" <- t -> pure []
-    -- Consumption marker types used to transfer information from template desugaring to LF conversion.
+    -- Types used only for desugaring are dropped during the LF conversion.
     | NameIn DA_Internal_Desugar n <- t
-    , n `elementOfUniqSet` consumingTypes
+    , n `elementOfUniqSet` desugarTypes
     -> pure []
 
     | hasDamlInterfaceCtx t && envLfVersion env `supports` featureInterfaces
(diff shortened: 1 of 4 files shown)


compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -408,12 +409,17 @@ modInstanceInfoFromDetails :: ModDetails -> ModInstanceInfo
 modInstanceInfoFromDetails ModDetails{..} = MS.fromList
     [ (is_dfun, overlapMode is_flag) | ClsInst{..} <- md_insts ]
 
-convertInterfaces :: Env -> [TyThing] -> ConvertM [Definition]
-convertInterfaces env tyThings
+interfaceNames :: Env -> [TyThing] -> ConvertM (S.Set T.Text)
```
sofiafaro-da, line 412: This doesn't need to be in ConvertM, as far as I can tell.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -870,8 +893,8 @@ convertChoice env tbinds (ChoiceData ty expr)
         applyThisAndArg func = func `ETmApp` EVar this `ETmApp` EVar arg
 
 
-convertBind :: Env -> (Var, GHC.Expr Var) -> ConvertM [Definition]
-convertBind env (name, x)
+convertBind :: Env -> S.Set T.Text -> (Var, GHC.Expr Var) -> ConvertM [Definition]
+convertBind env interfaceCons (name, x)
```
sofiafaro-da, line 897: Instead of passing around `interfaceCons`, I think it would be nicer to have it in the `env`.

sofiafaro-da (approved):
Looks great, thank you!

I added some suggestions below.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
cocreature: dropped it

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
cocreature: good point, I orginally needed the `ConvertM` which made this a bit annoying to fit in but now that I don’t, this works nicely.

PR #10897, opened 2021-09-15
Use ValueEnricher in ScenarioRunner

nickchapman-da:
Use `ValueEnricher` in `ScenarioRunner`. 
- Requires `ValueEnricher` works without an `Engine` (which `ScenarioRunner` does not have).
- Also `ScenarioRunner` needs direct access to `ValueTranslator` rather than via the `Preprocessor` wrapper, which only works with `MutableCompiledPackages` (which again, `ScenarioRunner` does not have).

This change advances the state of #6665

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
@@ -4,20 +4,35 @@
 package com.daml.lf
 package engine
 
-import com.daml.lf.data.Ref.{Identifier, Name}
+import com.daml.lf.data.Ref.{Identifier, Name, PackageId}
 import com.daml.lf.language.{Ast, LookupError}
 import com.daml.lf.transaction.Node.{GenNode, KeyWithMaintainers}
 import com.daml.lf.transaction.{CommittedTransaction, Node, NodeId, VersionedTransaction}
 import com.daml.lf.value.Value
 import com.daml.lf.value.Value.VersionedValue
+import com.daml.lf.speedy.SValue
 
 // Provide methods to add missing information in values (and value containers):
 // - type constructor in records, variants, and enums
 // - Records' field names
-final class ValueEnricher(engine: Engine) {
 
-  def enrichValue(typ: Ast.Type, value: Value): Result[Value] =
-    engine.preprocessor.translateValue(typ, value).map(_.toUnnormalizedValue)
+final class ValueEnricher(
+    compiledPackages: CompiledPackages,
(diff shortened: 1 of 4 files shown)


cocreature (approved):
Thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
```diff
@@ -4,20 +4,35 @@
...
+    loadPackage: (PackageId, language.Reference) => Result[Unit],
+) {
+
+  def this(engine: Engine) = {
+    this(
+      engine.compiledPackages(),
+      engine.preprocessor.translateValue,
+      engine.loadPackage,
+    )
+  }
```
remyhaemmerle-da, line 31: I do not really like this java-like overriding of class constructor. 
We are not using it at all in the daml-lf module.
I think this is there in scala for compatibility reason with java, but it is really not very idiomatic.

I would prefer to have an apply method in the companion object.

```
object ValueEnricher{

  def apply(engine: Engine): ValueEnricher =
    new ValueEnricher(
      engine.compiledPackages(),
      engine.preprocessor.translateValue,
      engine.loadPackage,
    )
}
```

remyhaemmerle-da (approved):
Nice.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala
nickchapman-da: Thanks for your comment Remy.

Yeah. I am also not a big fan of these secondary `this` constructors. The syntax is odd and rather constrained. See below.

So I had already tried your suggestion of the `apply` method in a companion object. The problem is that all callers must be updated from `new ValueEnricher(engine)` to `ValueEnricher(engine)` -- which is fine -- but without the `new` at the call site, it seems that `ValueEnricher` looks more like a `case class`, and obscures the fact that we are dealing with a mutable object.

But my opinions are weakly held here... If we think it is acceptable/better without the `new` I can make the change.


(
Aside: I found the `this` syntax is very limited. The following is not allowed syntactically:
```
  def this(engine: Engine) = {
    val compiledPackages = engine.compiledPackages()
    this(
      compiledPackages,
      engine.preprocessor.translateValue,
      engine.loadPackage,
    )
  }
```

-->

```
daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/ValueEnricher.scala:26: error: 'this' expected but 'val' found.
    val compiledPackages = engine.compiledPackages()
    ^
```
er, why? :shrug: ... oh of course, scala :smile: 
)

PR #10898, opened 2021-09-15
[DPP-572] Add ledger API test case for verifying conformance to `--min-tls-version` flag.

pbatko-da:
### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/Cli.scala
```diff
@@ -42,31 +43,15 @@ object Cli {
       case n: Int => (s.slice(0, n), s.slice(n + 1, s.length))
     }
 
-  private val Name = "ledger-api-test-tool"
-
-  private val pemConfig = (path: String, config: Config) =>
```
pbatko-da, line 47: A few incidental small refactorings. Better to split into a separate PR?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/Cli.scala
```diff
@@ -99,20 +86,33 @@ object Cli {
...
 
     opt[String]("cacrt")
       .optional()
       .text("TLS: The crt file to be used as the trusted root CA. Applied to all endpoints.")
-      .action(cacrtConfig)
+      .action { (path: String, config: Config) =>
+        config.withTlsConfig(_.copy(trustCertCollectionFile = Some(new File(path))))
+      }
+
+    opt[TlsVersion]("tls-version")
```
pbatko-da, line 110: TODO: Remove it. Not needed.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/LedgerApiTestTool.scala
```diff
@@ -220,20 +219,25 @@ object LedgerApiTestTool {
...
+    initializeParticipantChannels(
+      config.participantsEndpoints,
+      config.tlsConfig,
+    ).asFuture
+      .map((participants: Vector[(Channel, String, Int)]) =>
         new LedgerTestCasesRunner(
           testCases = cases,
-          participants = participants,
+          participants =
+            participants.map((x: (Channel, String, Int)) => (ChannelEndpoint.apply _).tupled(x)),
```
pbatko-da, line 230: `ChannelEndpoint` is just a more convenient container then Tuple3

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerSession.scala
```diff
@@ -21,7 +23,12 @@ private[infrastructure] final class LedgerSession private (
       else participantSessions
     Future
       .traverse(sessions) { case (endpointId, session) =>
-        session.createTestContext(endpointId, applicationId, identifierSuffix)
+        session.createTestContext(
+          endpointId,
+          applicationId,
+          identifierSuffix,
+          tlsConfiguration,
```
pbatko-da, line 30: TODO: Rename to clientTlsConfiguration

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestSuite.scala
```diff
@@ -32,6 +32,33 @@ private[testtool] abstract class LedgerTestSuite {
...
+  ): Unit = {
+    testGivenAllParticipants(
+      shortIdentifier,
+      description,
+      participants,
+      timeoutScale,
+      runConcurrently,
+      repeated,
+    )((ec: ExecutionContext) =>
+      (testContext: Seq[ParticipantTestContext]) => testCase(ec)(testContext)
```
pbatko-da, line 58: TLS1.3's `testCase` needs `testContext` to inspect it for participant hostname and port which are then used to construct a collection of gRPC channels each exercising different TLS version on the client side.

This seems like an abuse of test tool's framework where `testCase`'s is already effectively a client over some/none TLS.
For the need of our new TLSv1.3 test case we effectively ignore that pre-made client. (We only care for it's `TlsConfiguration` to be able to create our own with only TLS version modified)

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSv1_3IT.scala
```diff
@@ -0,0 +1,174 @@
...
+      { case _ =>
+        assume(testContexts.nonEmpty, "At least one context expected!")
+        val firstTextContext = testContexts.head
+        assume(
+          firstTextContext.clientTlsConfiguration.isDefined,
+          "This test case requires TLS configuration!",
+        )
+        val tlsConfiguration = firstTextContext.clientTlsConfiguration.get
+
+        val concurrency = Concurrency(
```
pbatko-da, line 47: (Thinking aloud..)
Appreciate hints as for what values would be sensible here.
Unless this config could be simplified away completely

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSv1_3IT.scala
```diff
@@ -0,0 +1,174 @@
...
+          port = firstTextContext.ledgerPort,
+        )
+
+        val resources: ResourceOwner[LedgerIdentityServiceBlockingStub] = for {
+          executorService <- threadPoolExecutorOwner(concurrency)
+          channel <- channelOwner(
+            ledger,
+            tlsConfiguration,
+            executorService,
+            clientTlsVersion = clientTls,
```
pbatko-da, line 65: Creating a channel over specified TLS version

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/LedgerTestSuite.scala
mziolekda: Maybe rather than wrapping it in test2, just use testGivenAllParticipants directly in TLSOnePointThreeIT. There is prior art with e.g. KVCommandDeduplicationBase

ledger/ledger-api-common/src/main/scala/com/digitalasset/ledger/api/tls/TlsConfiguration.scala
```diff
@@ -32,9 +33,14 @@ final case class TlsConfiguration(
   def client(enabledProtocols: Seq[TlsVersion] = Seq.empty): Option[SslContext] = {
     if (enabled) {
       val enabledProtocolsNames =
-        if (enabledProtocols.isEmpty)
-          null
-        else
+        if (enabledProtocols.isEmpty) {
```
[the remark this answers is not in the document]
pbatko-da: For sure I need to get rid off one

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/participant/ParticipantSession.scala
```diff
@@ -27,44 +29,57 @@ private[infrastructure] final class ParticipantSession private (
...
+      ledgerPort = ledgerPort,
+      clientTlsConfiguration = clientTlsConfiguration,
     )
 }
 
 object ParticipantSession {
   private val logger = LoggerFactory.getLogger(classOf[ParticipantSession])
 
-  def apply(
+  def createSessions(
```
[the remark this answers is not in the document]
pbatko-da: It confused me a bit in that it was an `apply` that returned a collection of objects rather than a single instance. 
Will revert it

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSOnePointThreeIT.scala
```diff
@@ -0,0 +1,125 @@
...
+import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
+
+/** Verifies that the given participant server correctly handles TLSv1.3 only mode.
+  *
+  * It works by creating and exercising a series of client service stubs, each over different TLS version.
+  * Only TLSv1.3 connection is expected to succeed.
+  * Connections over lower TLS versions are expected to fail.
+  */
+final class TLSOnePointThreeIT extends LedgerTestSuite {
```
meiersi-da, line 30: @mziolekda @pbatko-da : what do you think about also adding a test that ensures that by TLS 1.0 and TLS 1.1 connections are never allowed, which is the current recommendation. See e.g., https://docs.microsoft.com/en-us/lifecycle/announcements/transport-layer-security-1x-disablement

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSOnePointThreeIT.scala
```diff
@@ -0,0 +1,125 @@
...
+      { case _ =>
+        // preconditions
+        assume(testContexts.nonEmpty, "Missing an expected participant test context!")
+        val firstTextContext = testContexts.head
+        assume(
+          firstTextContext.clientTlsConfiguration.isDefined,
+          "Missing required TLS configuration!",
+        )
+        val tlsConfiguration = firstTextContext.clientTlsConfiguration.get
+        assume(tlsConfiguration.enabled, "TLS configuration is disabled but expected be enabled!")
```
meiersi-da, line 58:
```suggestion
        assume(tlsConfiguration.enabled, "TLS configuration is disabled but expected to be enabled!")
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSOnePointThreeIT.scala
pbatko-da: This might be already covered by this test depending what exactly you have in mind.

To recap, this test checks that:
1) clients on TLSv1.3 succeed,
2) clients on TLS 1.2, 1.1 and 1.0 fail. 
(It follows that for this test to pass the server must be started in TLSv1.3 only mode.)

Do you possibly mean to add a test case that only ensures that clients on TLS 1.0 and 1.1 always fail? This would partially duplicate this test case but having just these two cases separate would mean it could be a default test case of Ledger AP Test Tool (as opposed to being optional that needs to be explicitly `--include`d or `--additional`ed) .

mziolekda: Simon, we have decided to create a test that verifies minimum client version of 1.3. We felt that it is the only one that makes sense. If one is serious about security, should set the server to a minimum of 1.3 and that's what this test asserts.

mziolekda (approved):
Nice work @pbatko-da. I like it how you managed to wean the PR from the original heavy changes.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/suites/TLSOnePointThreeIT.scala
meiersi-da: @mziolekda : TLS 1.3 was defined in August 2018, so is very recent in enterprise terms. TLS 1.2 has not been deprecated. I would thus not be surprised that there are users of our software that want to use TLS 1.2 instead of TLS 1.3. For these, it is very important that the actually deprecated TLS 1.0 and 1.1 versions are not accepted. See https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.2 for background.

So, I believe we either decide to not support TLS 1.2 at all, or we do test that if TLS 1.2 is used, the 1.0 and 1.1 versions are not supported.

@mziolekda : how would you like to proceed?

kamil-da (approved):

Issue #7271, opened 2020-08-31
Document HA setup for the jSON API


cocreature:
We should document how users can run an HA setup for the JSON API. If needed, this could also include some implementation work. There are two possible approaches and we should choose one to document (with the possibility of revising this in the future):
- achieving HA for the HTTP-JSON API server as a standalone component, with multiple instances running against a single participant to enable high availability
- achieving HA at the "Connect" level, having one HTTP-JSON API server per participant and having this behave as a single unit of deployment which can be replicated to enable high availability

---

Open question: are offsets are translatable if a fail-over happens and if not, how do users know they are connecting to a different HTTP-JSON API instance?

stefanobaghino-da:
> one json api per participant

What do you mean here exactly? If you just make one JSON API and one participant a unit of deployment and manage that single unit to be deployed in such a way that it's highly available I'm not sure this should be documented as part of the JSON API itself.

cocreature:
Not sure it needs to be documented as part of the JSON API itself but somewhere we need to document how we recommend running an HA connect setup.

stefanobaghino-da:
> Not sure it needs to be documented as part of the JSON API itself but somewhere we need to document how we recommend running an HA connect setup.

Is the participant part of Connect? I thought the participant was part of the driver. It's unclear what you are trying to document in terms of high availability if you only have a single HTTP-JSON API server which is deployed alongside the participant.

cocreature:
In canton participant = connect node so it is part of connect.

stefanobaghino-da:
@akshayshirahatti-da I re-worded the ticket a bit. I think that documenting the first approach (HTTP-JSON API as a single unit of deployment and achieve HA at that level) is good enough of a starting point. We may want to refine this further down the road by including how this interacts with the rest of a Connect deployment but I don't think it's strictly necessary at this point.

stefanobaghino-da:
Closed by #10903

PR #10903, opened 2021-09-15
[JSON-API] Production/HA documentation.

akshayshirahatti-da:
v1 documentation for JSON API production recommendations with notes on scaling and redundancy.
I have moved the metrics section into the new documentation page.

Refers to #7271 

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+Service Configuration
+*********************
+
+.. note:: The full list of configuration flags supported by the JSON API can be seen by running ``daml json-api --help``.
+
+Following are the list of recommended configs for a production deployment ? //TODO
+
+
+Security and privacy
+***************************
```
stefanobaghino-da, line 50:
```suggestion
Security and privacy
********************
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+rest and using a secure communication channel between the *JSON API* server and the PostgreSQL server.
+
+To protect data in transit and over untrusted networks, the JSON API provides
+TLS support, to enable TLS you need to specify the private key for your server and the
+certificate chain via ``daml json-api --pem server.pem --crt server.crt``. You can also
+set a custom root CA certificate used to validate client certificates via ``--cacrt ca.crt``
+
+
+Architecture
+***************************
```
stefanobaghino-da, line 66:
```suggestion
Architecture
************
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+
+In production, we recommend running on a 64-bit x86 architecture in a Linux
+environment. This environment should have a Java SE Runtime Environment such
+as OpenJDK JRE and must be compatible with OpenJDK version 1.8.0_202 or later.
+We recommend using PostgreSQL server as query-store, most of our tests have
+been done with servers running version > 10.
+
+
+Scaling and Redundancy
+***************************
```
stefanobaghino-da, line 91:
```suggestion
Scaling and Redundancy
**********************
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+
+Finally we recommend using app orchestration utilities which monitor the health of the service
+and perform subsequent operations to ensure availability. These utilities can use the
+`healthcheck endpoints <https://docs.daml.com/json-api/index.html#healthcheck-endpoints>`__
+provided by the JSON API. This can also be tied into supporting arbitrary autoscaling implementation
+to ensure minimum number of *JSON API* services on failures or rolling upgrades.
+
+
+Logging
+***************************
```
stefanobaghino-da, line 129:
```suggestion
Logging
*******
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+Logged events should carry information about the request being served by the
+*JSON API* service. When using a traditional logging target (e.g. standard output
+or rotating files) this information will be part of the log description.
+Using a logging target compatible with the Logstash Logback Encoder allows to have rich
+logs with structured information about the event being logged.
+
+
+
+Metrics
+***************************
```
stefanobaghino-da, line 146:
```suggestion
Metrics
*******
```

docs/source/json-api/index.rst
```diff
@@ -92,21 +92,22 @@ using.
 With Query Store
 ------------------
 
-In production setups, you should configure the JSON API to use a
-PostgreSQL backend as a cache. The in-memory backend will call the
+You can configure the JSON API to use a PostgreSQL/Oracle(Enterprise Edition Only)
```
stefanobaghino-da, line 95: The statement about this relevant only for production setups seems relevant, I'd keep it in.

docs/source/json-api/production-setup.rst
stefanobaghino-da: I believe there's a check in our docs build phase that ensures that header markups are the same length as the header text, sprinkled likewise suggestions all over the PR.

docs/source/json-api/index.rst
```diff
@@ -92,21 +92,22 @@ using.
...
+backend as a cache. The in-memory backend will call the
 ledger to fetch the entire active contract set for the templates in
 your query every time so it is generally not recommended to rely on
 this in production. Note that the PostgreSQL backend acts purely as a
 cache. It is safe to reinitialize the database at any time.
 
 To enable the PostgreSQL backend you can use the ``--query-store-jdbc-config`` flag, an example of which is below.
 
-.. note:: When you use the Query Store you'll want your first run to specify ``createSchema=true`` so that all the necessary tables are created. After the first run make sure ``createSchema=false`` so that it doesn't attempt to create the tables again.
+.. note:: For development purposes we suggest to use  ``start-mode=create-if-needed-and-start`` so that all the necessary tables are created if they don't exist.
```
stefanobaghino-da, line 104: The change is relevant since `createSchema=true` is no longer there IIRC, but I wouldn't suggest to use the query store "for development purposes".

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+query performance. Details for enabling a query store are highlighted below.
+
+Query Store
+***********
+
+Query store can be described as a cached search index and is useful for use-cases
+where we need to query large active contract sets. The *JSON API* can be configured
+with PostgreSQL/Oracle(Enterprise Edition only) as the query-store backend.
+
+For e.g to enable the PostgreSQL backend you can use the ``--query-store-jdbc-config`` flag, an example of which is below.
```
stefanobaghino-da, line 27:
```suggestion
For example to enable the PostgreSQL backend you can use the ``--query-store-jdbc-config`` flag, an example of which is below.
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+Components
+----------
+
+A production setup of the *JSON API* service will involve the following components:
+
+- the *JSON API* server
+- the query store backend database server
+- a ledger
+
+*JSON API* exposes a RESTFul API to interact with the Ledger and it uses JDBC to interact
```
stefanobaghino-da, line 77:
```suggestion
*JSON API* exposes an API to interact with the Ledger and it uses JDBC to interact
```
Not really RESTful. 🙂

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,304 @@
...
+
+We advise that the *JSON API server* and *query store* components to have dedicated
+computation and memory resources available to them. This can be achieved via
+containerization or setting them up on independent physical servers. Ensure that the two
+components are **physically co-located** to reduce network latency for
+communication between the *JSON API* and the *database server*
+The scaling and availability aspects heavily rely on the interactions between
+the core components listed above.
+
+With respect to scaling we recommend to follow the general advise in trying to
```
stefanobaghino-da, line 103:
```suggestion
With respect to scaling we recommend to follow the general advice in trying to
```
_Advise_ is the verb, right?

stefanobaghino-da (commented):
Looks good so far. I'd recommend to do a quick round and try to use consistently one formatting (italics or not) and one name for the JSON API. "HTTP-JSON API" is the official name, possibly with "server" or "service" alongside it to specify whether you are referring to the running process or the abstract service it offers.

docs/source/json-api/index.rst
akshayshirahatti-da: ok removed the wording "for development purposes" , my idea was to suggest using `create-if-needed-and-start` here and the relevant `create-only, start-only` modes in production setups where we want separation of credentials.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+For example to enable the PostgreSQL backend you can use the ``--query-store-jdbc-config`` flag, as shown below.
+
+.. code-block:: shell
+
+    daml json-api --ledger-host localhost --ledger-port 6865 --http-port 7575 \
+    --query-store-jdbc-config "driver=org.postgresql.Driver,url=jdbc:postgresql://localhost:5432/test?&ssl=true,user=postgres,password=password,start-mode=start-only"
+
+For using the Query Store you'll want your first run to specify ``start-mode=create-only``
+so that all the necessary tables are created. After the first run make
+sure its set to ``start-mode=start-only`` so that it doesn't attempt to create
```
garyverhaegen-da, line 36:
```suggestion
sure it is set to ``start-mode=start-only`` so that it doesn't attempt to create
```

Would it be worth having a small discussion of the pros and cons of various `start-mode` options? If this is already documented somewhere else, at least a link would be good. I'm personally not a fan of `create-only`, and would be a bit sad if that were our only recommended production setup.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+The *HTTP-JSON API* server is a JVM application that by default uses an in-memory backend.
+This in-memory backend setup is inefficient for larger datasets as for every query it
+ends up fetching the entire active contract set for the templates referenced in that query.
+For this reason for production setups at a minimum we recommend to use a database
+as a query store, this will allow for more efficient caching of the data to improve
+query performance. Details for enabling a query store are highlighted below.
+
+Query Store
+***********
+
```
garyverhaegen-da, line 22:
```suggestion

.. note:: The Community Edition of Daml Connect only supports PostgreSQL backends for the *HTTP JSON API* server, but the Enterprise Edition also supports Oracle backends.

```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+Thus, it is their responsibility to ensure that the data abides by the necessary
+regulations and confidentiality expectations.
+
+It is recommended to use the tools documented by PostgreSQL to protect data at
+rest and using a secure communication channel between the *HTTP-JSON API* server and the PostgreSQL server.
+
+To protect data in transit and over untrusted networks, the *HTTP-JSON API* server provides
+TLS support, to enable TLS you need to specify the private key for your server and the
+certificate chain via ``daml json-api --pem server.pem --crt server.crt``. You can also
+set a custom root CA certificate used to validate client certificates via ``--cacrt ca.crt``
```
garyverhaegen-da, line 58: Should we link to @nycnewman's [ex-secure-daml-infra](https://github.com/digital-asset/ex-secure-daml-infra) repo here for a full example using TLS everywhere?

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+- the query store backend database server
+- the ledger
+
+*HTTP-JSON API* server exposes an API to interact with the Ledger and it uses JDBC to interact
+with its underlying query store for caching and serving data efficiently.
+
+The *HTTP-JSON API* server releases are regularly tested with OpenJDK 8 on a 64-bit x86 architecture,
+with Ubuntu 20.04, macOS 11.5.2 and Windows Server 2016.
+
+In production, we recommend running on a 64-bit x86 architecture in a Linux
```
garyverhaegen-da, line 79:
```suggestion
In production, we recommend running on a x86_64 architecture in a Linux
```

As far as I know `x86` is the designation for the 32-bit version.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+We recommend using PostgreSQL server as query-store, most of our tests have
+been done with servers running version > 10.
+
+
+Scaling and Redundancy
+**********************
+
+.. note:: This section of the document only talks about scaling and redundancy setup for the *HTTP-JSON API* server. In all of the recommendations suggested below we assume that the JSON API always interacts with a single participant on the ledger.
+
+We advise that the *HTTP-JSON API* server and *query store* components to have dedicated
```
garyverhaegen-da, line 91:
```suggestion
We advise that the *HTTP-JSON API* server and *query store* components have dedicated
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+containerization or setting them up on independent physical servers. Ensure that the two
+components are **physically co-located** to reduce network latency for
+communication. The scaling and availability aspects heavily rely on the interactions between
+the core components listed above.
+
+With respect to scaling we recommend to follow the general advice in trying to
+understand the bottlenecks and see if adding additional processing power/memory
+i.e vertical scaling is beneficial.
+
+In general for horizontal scaling purposes , we recommend to treat the
```
garyverhaegen-da, line 102:
```suggestion
In general, for horizontal scaling purposes, we recommend to treat the
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+.. note:: This section of the document only talks about scaling and redundancy setup for the *HTTP-JSON API* server. In all of the recommendations suggested below we assume that the JSON API always interacts with a single participant on the ledger.
+
+We advise that the *HTTP-JSON API* server and *query store* components to have dedicated
+computation and memory resources available to them. This can be achieved via
+containerization or setting them up on independent physical servers. Ensure that the two
+components are **physically co-located** to reduce network latency for
+communication. The scaling and availability aspects heavily rely on the interactions between
+the core components listed above.
+
+With respect to scaling we recommend to follow the general advice in trying to
```
garyverhaegen-da, line 98:
```suggestion
With respect to vertical scaling we recommend to follow the general advice in trying to
```

Otherwise it's a bit weird that we have one paragraph that covers "scaling" and then another one that covers "in general", "horizontal" scaling.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+understand the bottlenecks and see if adding additional processing power/memory
+i.e vertical scaling is beneficial.
+
+In general for horizontal scaling purposes , we recommend to treat the
+*HTTP-JSON API* and the *query store database server* as a single unit for scaling.
+While a setup with multiple *HTTP-JSON API* services running against a single backend
+database server is feasible, it maybe a futile exercise if the database server
+itself is the bottleneck.
```
garyverhaegen-da, line 106: I'm not entirely sure about this paragraph. This seems to strongly imply we expect the database to be the bottleneck most of the time; do we have evidence for that?

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+
+We can run a redundant setup for the *HTTP-JSON API* by using a reverse proxy server with some
+acceptable routing mechanism and then and having multiple *HTTP-JSON API* servers sit behind it
```
garyverhaegen-da, line 109:
```suggestion
acceptable routing mechanism and having multiple *HTTP-JSON API* servers sit behind it
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+each with their own backend database servers and dedicated computation and memory resources.
```
garyverhaegen-da, line 110: This is where I would expect some explanation of exactly how the query store is constructed and from that what an "acceptable routing mechanism" is.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Users may consider running PostgreSQL backend in a `high availability configuration <https://www.postgresql.org/docs/current/high-availability.html>`__.
+The benefits of this are use-case dependent as this may be more expensive for
+smaller active contract datasets, where re-initializing the cache is cheap and fast.
```
garyverhaegen-da, line 114: This is the first time this document implies that it is acceptable to just drop the DB and start from scratch. I think this should be said explicitly, and much earlier, likely around line 23 (i.e. near the start of the "Query store" section).

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+read more about that in the `Logback documentation <http://logback.qos.ch/>`__.
+
+The logging infrastructure leverages structured logging as implemented by the
+`Logstash Logback Encoder <https://github.com/logstash/logstash-logback-encoder/blob/logstash-logback-encoder-6.3/README.md>`__.
+
+Logged events should carry information about the request being served by the
+*HTTP-JSON API* server. When using a traditional logging target (e.g. standard output
+or rotating files) this information will be part of the log description.
+Using a logging target compatible with the Logstash Logback Encoder allows to have rich
+logs with structured information about the event being logged.
```
garyverhaegen-da, line 137: Here I would expect some mention of what fields I can expect in the logs, or a link to an existing page describing that. May also be worth explicitly mentioning that the logs are JSON by default? (Assuming they are; I've lost track of that stream of work.)

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+
+  - ``graphite://<server_host>[:<server_port>]``: sends captured metrics to a Graphite server. If the port
+    is omitted, the default value ``2003`` will be used.
+
+  - ``prometheus://<server_host>[:<server_port>]``: renders captured metrics
+    on a http endpoint in accordance with the prometheus protocol. If the port
+    is omitted, the default value ``55001`` will be used. The metrics will be
+    available under the address ``http://<server_host>:<server_port>/metrics``.
+
+- ``--metrics-reporting-interval``: metrics are pre-aggregated on the sandbox and sent to
```
garyverhaegen-da, line 165:
```suggestion
- ``--metrics-reporting-interval``: metrics are pre-aggregated on the *HTTP JSON API* server and sent to
```
presumably?

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+  the reporter, this option allows the user to set the interval. The formats accepted are based
+  on the ISO-8601 duration format ``PnDTnHnMn.nS`` with days considered to be exactly 24 hours.
```
garyverhaegen-da, line 167:
```suggestion
  on the ISO 8601 duration format ``PnDTnHnMn.nS`` with days considered to be exactly 24 hours.
```
Never seent his written with a dash.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+- ``<metric.qualified.name>.count``: number of registered data points overall
+- ``<metric.qualified.name>.m1_rate``: number of registered data points per minute
+- ``<metric.qualified.name>.m5_rate``: number of registered data points every 5 minutes
+- ``<metric.qualified.name>.m15_rate``: number of registered data points every 15 minutes
+- ``<metric.qualified.name>.mean_rate``: mean number of registered data points
+
+Timers
+------
+
+A timer records all metrics registered by a meter and by an histogram, where
```
garyverhaegen-da, line 196:
```suggestion
A timer records all metrics registered by a meter and by a histogram, where
```

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+the histogram records the time necessary to execute a given operation (unless
+otherwise specified, the precision is nanoseconds and the unit of measurement
+is milliseconds).
+
+List of metrics
+===============
+
+The following is an exhaustive list of selected metrics
```
garyverhaegen-da, line 204: "exhaustive" list of "selected" sounds really strange.

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+that can be particularly important to track.
+
+``daml.http_json_api.command_submission_timing``
+------------------------------------------------
+
+A timer. Meters how long processing of a command submission request takes
```
garyverhaegen-da, line 210: I would remove all of those `Meters` and just say "How long ...". More importantly, I think this should mention the time unit for each metric (even if it is always the same one).

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,300 @@
...
+- ``<metric.qualified.name>.m15_rate``: number of registered data points every 15 minutes
+- ``<metric.qualified.name>.mean_rate``: mean number of registered data points
+
+Timers
+------
+
+A timer records all metrics registered by a meter and by an histogram, where
+the histogram records the time necessary to execute a given operation (unless
+otherwise specified, the precision is nanoseconds and the unit of measurement
+is milliseconds).
```
garyverhaegen-da, line 199: I'm a bit confused by these units. Does that mean we report fractional numbers?

docs/source/json-api/index.rst
garyverhaegen-da: I'm not aware that _we_ want that separation of credentials. We've been asked to support it, but that's not quite the same as presenting it as the only way to run in production.

docs/source/json-api/production-setup.rst
akshayshirahatti-da: @garyverhaegen-da not sure what you mean when you say "how the query store is constructed" ? w.r.t the routing mechanism are you hinting towards suggesting sticky sessions based on parties querying the contract dataset ? I think those decisions can be left to the operators of the deployment.

docs/source/json-api/production-setup.rst
victormueller-da: I think both here is fine

docs/source/json-api/production-setup.rst
victormueller-da: It might make more sense to reword this to rather indicate that scaling should be successful as long as the database (and the ledger) can keep up.

docs/source/json-api/production-setup.rst
garyverhaegen-da: > I think those decisions can be left to the operators of the deployment.

Decisions, sure, but we should make sure they have the information needed. If we don't tell them why they might want to, they can't really make an informed decision.

I'm not familiar enough with the JSON API to be confident any of the following is true, but I would expect something along these lines:

> The Query Store is built by saving the state of the ACS up to the current offset, which allows the JSON API to only request the delta on subsequent queries, making it much faster than having to request the entire ACS every time. Because of the Daml privacy model (JSON API requests are made using the user-provided token), the data stored in a given Query Store will be specific to the set of parties that have made queries through that specific Query Store instance (for a given template). Therefore, if you do run with separate Query Stores, it may be useful to route queries based on requesting party (and possibly queried template), which would minimize the amount of data in each Query Store as well as the overall redundancy of said data.

Which reminds me, it's also not very clear from this document why the operator needs to be careful with data privacy, which this explanation would also somewhat cover.

docs/source/json-api/production-setup.rst
akshayshirahatti-da: @garyverhaegen-da they are not JSON by default , its plaintext encoding, I'll add a line which states that it is the default. Regarding what fields , I can add a note that commands, endpoints and other request/response details are captured in the logs.

docs/source/json-api/production-setup.rst
akshayshirahatti-da: will remove `exhaustive`

docs/source/json-api/production-setup.rst
garyverhaegen-da: Really? I thought we had agreed to make JSON the default across the stack. 🤔

docs/source/json-api/production-setup.rst
akshayshirahatti-da: Regarding time unit for the metric , the Timer description earlier states this
`(unless
otherwise specified, the precision is nanoseconds and the unit of measurement
is milliseconds).`

The wordings for this were from the earlier documentation and I don't personally like either version
with the `Meters` or without it . If we want I can instead add 
`Measures latency for processing a command submission request`

docs/source/json-api/production-setup.rst
akshayshirahatti-da: So AFAIK codahale library measures it in nanoseconds, but then its down to most of the reporters on how they choose to report it.  (ref:  https://metrics.dropwizard.io/4.2.0/getting-started.html#timers)
e.g the console reporter prints this
```
daml.http_json_api.incoming_json_parsing_and_validation_timing
             count = 1
         mean rate = 0.05 calls/second
     1-minute rate = 0.17 calls/second
     5-minute rate = 0.19 calls/second
    15-minute rate = 0.20 calls/second
               min = 225.41 milliseconds
               max = 225.41 milliseconds
              mean = 225.41 milliseconds
            stddev = 0.00 milliseconds
            median = 225.41 milliseconds
              75% <= 225.41 milliseconds
              95% <= 225.41 milliseconds
              98% <= 225.41 milliseconds
              99% <= 225.41 milliseconds
            99.9% <= 225.41 milliseconds
```

docs/source/json-api/production-setup.rst
akshayshirahatti-da: I think I was trying to recommend that clients stick to the server+db as a unit of scaling, unless they know better. I will reword this to 

```suggestion
While a setup with multiple *HTTP-JSON API* services running against a single backend
database server is feasible, we would recommend ensuring that the database server is 
not a bottleneck before pursuing such a setup.
```

garyverhaegen-da: I'd prefer something more neutral, like:
> The *HTTP JSON API* can be scaled independently of its Query Store. You can have any number of *HTTP JSON API* instances talking to the same Query Store (if, for example, your monitoring indicates that the *HTTP JSON API* processing time is the bottleneck), or have each *HTTP JSON API* instance talk to its own independent Query Store.
>
> In the latter case, ... <explanations of how the Query Store is constructed, per my other comment>

I don't think we want to push people towards having many databases here.

docs/source/json-api/production-setup.rst
garyverhaegen-da: Do we care to specify the precision here? Wouldn't

> .. to execute a given operation (in fractional milliseconds).

be good enough here?

docs/source/json-api/production-setup.rst
akshayshirahatti-da: corrected this , based on the discussion I have added a couple of lines about how the `start-mode` can be used , please have a look to check if that makes sense @garyverhaegen-da :)

garyverhaegen-da: 👍

docs/source/json-api/production-setup.rst
akshayshirahatti-da: I have split this into 2 and added parts about how the query store is built to the top `Query Store` section and the other bits here

garyverhaegen-da (approved):
Thanks!

docs/source/json-api/production-setup.rst
```diff
@@ -0,0 +1,318 @@
...
+
+*Query Store* can be described as a cached search index and is useful for use-cases
+where we need to query large active contract sets(ACS). The *HTTP-JSON API* server can be
+configured with PostgreSQL/Oracle(Enterprise Edition only) as the query-store backend.
+
+The *Query Store* is built by saving the state of the ACS up to the current ledger
+offset. This allows the *HTTP-JSON API* to only request the delta on subsequent queries,
+making it much faster than having to request the entire ACS every time.
+
+Given the cache like semantics of the *Query Store* it is safe to drop and re-initialize the store at any point
```
garyverhaegen-da, line 33:
```suggestion
Given the cache like semantics of the *Query Store* it is safe to drop and re-initialize the store at any point, though there may be a temporary performance impact while the *Query Store* is being rebuilt.
```

PR #10911, opened 2021-09-16
Add fetch, exercise implementations for interfaces in speedy.

sofiafaro-da:
This is based on the first PoC, but I tried to improve the cacheing situation a bit. And to do it all without changing any of the existing definitions.

I wasn't sure what the best way to pass "template implements interface" data from the Ast to the engine was. I ended up going with an `ImplementsDefRef` which is supposed to be a function that converts template value to an interface value (an identity function for now), and using its presence to perform the implements checks on fetch and exercise. If you have a better approach, I'm interested.

No tests yet, but it will be easier to test via daml.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
@@ -1021,13 +1021,13 @@ private[lf] final class Compiler(
   ): (SDefinitionRef, SDefinition) =
     topLevelFunction(ChoiceDefRef(ifaceId, choice.name), 2) { case List(cidPos, choiceArgPos, _) =>
       withEnv { _ =>
-        let (
+        let(
           SBUPreFetchInterface(ifaceId)(svar(cidPos))
         ) { tmplArgPos =>
-          SBUChoiceInterface(ifaceId, choice.name) (
+          SBUChoiceInterface(ifaceId, choice.name)(
             svar(cidPos),
             svar(choiceArgPos),
-            svar(tmplArgPos)
+            svar(tmplArgPos),
           )
         }
       }
@@ -1429,12 +1429,12 @@ private[lf] final class Compiler(
   ): (SDefinitionRef, SDefinition) =
     topLevelFunction(FetchDefRef(ifaceId), 2) { case List(cidPos, _) =>
       withEnv { _ =>
-        let (
+        let(
           SBUPreFetchInterface(ifaceId)(svar(cidPos))
         ) { tmplArgPos =>
-          SBUFetchInterface(ifaceId) (
+          SBUFetchInterface(ifaceId)(
             svar(cidPos),
-            svar(tmplArgPos)
+            svar(tmplArgPos),
           )
         }
       }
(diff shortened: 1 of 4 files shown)


remyhaemmerle-da (approved):
Looks pretty good

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1072,6 +1072,115 @@ private[lf] object SBuiltin {
...
+      val coid = getSContractId(args, 0)
+      onLedger.cachedContracts.get(coid) match {
+        case Some(cached) => {
+          machine.returnValue = cached.value
+        }
+        case None => {
+          throw SpeedyHungry(
+            SResultNeedContract(
+              coid,
+              ifaceId, // not actually used, maybe this param should be dropped from SResultNeedContract
```
cocreature, line 1091: hm this is somewhat interesting. We considered at some point to pass this along to ledger implementations so they could use this to make lookups on their side more efficient. I think interfaces don’t fundamentally break this but at least simple table partitioning by template gets more difficult. Anyway, we never actually went down that route so I don’t think this is blocking by any means, just recording my thoughts.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
```diff
@@ -465,6 +465,7 @@ object SExpr {
   final case class KeyDefRef(ref: DefinitionRef) extends SDefinitionRef
   final case class SignatoriesDefRef(ref: DefinitionRef) extends SDefinitionRef
   final case class ObserversDefRef(ref: DefinitionRef) extends SDefinitionRef
+  final case class ImplementsDefRef(ref: DefinitionRef, ifaceId: TypeConName) extends SDefinitionRef
```
cocreature, line 468: maybe worth adding a comment what this points to and what the type is? Arguably helpful for all the others above as well where we don’t have it :disappointed:

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
remyhaemmerle-da: Maybe it will be worthwhile to ask Simon, his option.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SExpr.scala
sofiafaro-da: I added a comment

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -1006,6 +1013,24 @@ private[lf] final class Compiler(
     }
   }
 
+  private[this] def compileChoiceInterface(
+      ifaceId: TypeConName,
+      choice: InterfaceChoice,
+  ): (SDefinitionRef, SDefinition) =
+    topLevelFunction(ChoiceDefRef(ifaceId, choice.name), 2) { case List(cidPos, choiceArgPos, _) =>
```
remyhaemmerle-da, line 1020: suggestion
```topLevelFunction(ChoiceDefRef(ifaceId, choice.name), 2) { case List(cidPos, choiceArgPos) =>```

PR #10912, opened 2021-09-16
Logging delay of submitted commands

oliverse-da:
This would have helped track down the timezone issue more quickly.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
@@ -258,6 +258,9 @@ private[apiserver] final class ApiSubmissionService private[services] (
             if (submissionDelay.isNegative)
               submitTransaction(transactionInfo)
             else {
+              logger.debug(
+                s"Delaying submission of ${transactionInfo.submitterInfo.commandId} by $submissionDelay"
+              )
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
```diff
@@ -258,6 +258,9 @@ private[apiserver] final class ApiSubmissionService private[services] (
             if (submissionDelay.isNegative)
               submitTransaction(transactionInfo)
             else {
+              logger.debug(
```
rautenrieth-da, line 261: Every submission already produces a log line at info level (see `logger.info("Submitting transaction")` above). I think info level output would be good as well.

Did you also check whether the logging context already contains the command ID? If yes (and I think it does), then you could simply write `"Delaying submission by $submissionDelay"`.

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/apiserver/services/ApiSubmissionService.scala
oliverse-da: You are right commandId is already include, so trimming down the text portion and increasing the log level to info. Thanks!

PR #10882, opened 2021-09-14
Do not drop generated `submissionId`s in `GrpcCommandService` [KVL-1104]

hubert-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10919, opened 2021-09-16
ledger-api-client: Throw an exception if the submission id is empty in `CommandTracker` [KVL-1104]

hubert-da:
It turns out that even though we enrich submission requests with the submission id if it's empty (https://github.com/digital-asset/daml/pull/10882), it gets bypassed when using `CommandClient` (don't yet know how), and it breaks the integration with Canton.
Therefore, we throw an exception now if the submission id is empty, to indicate a bug. Later, we may want to change this one and the other exception below to a proper `CompletionFailure` that is returned by the stream.

Follow-up:
We either have to find a place where the submission ID isn't generated in the SDK or generate it in Canton.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -236,6 +236,11 @@ private[commands] class CommandTracker[Context](
         val submissionId = commands.submissionId
         val commandId = commands.commandId
         logger.trace(s"Begin tracking of command $commandId for submission $submissionId.")
+        if (commands.submissionId.isEmpty) {
+          throw new IllegalArgumentException(
+            s"The submission id for the command $commandId is empty. This should not happen."
+          )
+        }
         if (pendingCommands.contains(TrackedCommandKey(submissionId, commandId))) {
           // TODO return an error identical to the server side duplicate command error once that's defined.
           throw new IllegalStateException(
--- a/ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
+++ b/ledger/ledger-api-client/src/test/suite/scala/com/digitalasset/ledger/client/services/commands/CommandTrackerFlowTest.scala
@@ -152,6 +152,20 @@ class CommandTrackerFlowTest
       }
     }
 
+    "a command is submitted without a submission id" should {
+
+      "throw an exception" in {
+        val Handle(submissions, results, _, _) =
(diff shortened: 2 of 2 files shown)


fabiotudone-da (commented):
Is the output helpful when this exception is triggered?

hubert-da:
> Is the output helpful when this exception is triggered?

This is what I get from the test in the IDE:
```
ERROR a.a.RepointableActorRef - Error in stage [com.daml.ledger.client.services.commands.tracker.CommandTracker$$anon$1]: The submission id for the command 21fe332f-6616-4959-a27c-8bdefe61a62c-ping-55053f76-9ce5-4e55-9e16-c2ad0ef28f4f is empty. This should not happen.
java.lang.IllegalArgumentException: The submission id for the command 21fe332f-6616-4959-a27c-8bdefe61a62c-ping-55053f76-9ce5-4e55-9e16-c2ad0ef28f4f is empty. This should not happen.
	at com.daml.ledger.client.services.commands.tracker.CommandTracker$$anon$1.com$daml$ledger$client$services$commands$tracker$CommandTracker$$anon$$registerSubmission(CommandTracker.scala:241)
...
```

So yes, I think so.

fabiotudone-da (approved):
This fail-fast is definitely an improvement, thanks!

fabiotudone-da:
@hubert-da Consider merging it to the 1.17 release branch as well.

hubert-da:
> @hubert-da Consider merging it to the 1.17 release branch as well.

Yes, will do this tomorrow, as it still needs to be fixed either in Canton or here, and this PR is not required for that.

oliverse-da:
> > @hubert-da Consider merging it to the 1.17 release branch as well.
> 
> Yes, will do this tomorrow, as it still needs to be fixed either in Canton or here, and this PR is not required for that.

How about we pick this up together on Monday and ensure we have both daml and canton-side changes ready to go and working? So basically there is no more submitting of commands via the ledger-api without a submission-id? Or is there a layer at which we can generate a submission-id if none was chosen and use that for tracking?

hubert-da:
> How about we pick this up together on Monday and ensure we have both daml and canton-side changes ready to go and working? So basically there is no more submitting of commands via the ledger-api without a submission-id? Or is there a layer at which we can generate a submission-id if none was chosen and use that for tracking?

We generate the submission id in the gRPC services (https://github.com/digital-asset/daml/pull/10882). It turns out that there is one more place where we should generate the submission id. If the `CommandClient` is being used, submissions are getting registered in the `CommandTracker` before enriching them in the services. That's why we should also generate the submission id in `CommandClient.commandUpdaterFlow`. So, I'm going to open a PR (+1 to the release branch) that replaces this one. This way it will be ready for Monday.

hubert-da:
Also, there are new test failures due to failing fast (introduced in this PR), so the change is going to be covered by them.

PR #10926, opened 2021-09-17
ledger-api-client: Generate a submission id if it's empty in the `CommandClient` [KVL-1104]

hubert-da:
Replaces: https://github.com/digital-asset/daml/pull/10919

It turns out that even though we enrich submission requests with the submission id if it's empty (#10882), it gets bypassed when using the `CommandClient`.
Therefore, we generate a submission id if it's empty in `CommandClient.commandUpdaterFlow` and fail fast if the submission id is empty in the `CommandTracker` to indicate a bug.

This change is covered by tests that started failing here: https://dev.azure.com/digitalasset/daml/_build/results?buildId=88924&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=ac0b0b0f-051f-52f7-8fb3-a7e384b0dde9&l=2461

Needs to be backported to the 1.17 release branch.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
+++ b/ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
@@ -236,6 +236,11 @@ private[commands] class CommandTracker[Context](
         val submissionId = commands.submissionId
         val commandId = commands.commandId
         logger.trace(s"Begin tracking of command $commandId for submission $submissionId.")
+        if (commands.submissionId.isEmpty) {
+          throw new IllegalArgumentException(
+            s"The submission id for the command $commandId is empty. This should not happen."
+          )
+        }
         if (pendingCommands.contains(TrackedCommandKey(submissionId, commandId))) {
           // TODO return an error identical to the server side duplicate command error once that's defined.
           throw new IllegalStateException(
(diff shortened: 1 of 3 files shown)


fabiotudone-da (commented):
2 things:

1. Have you considered unifying the code paths so that the submission ID is generated in only one place?
2. Could you add a test for the new behavior in the Command Client?

hubert-da:
@fabiotudone-da,
1. Yes, the one here will have to stay, I believe. I can create a story for the gRPC services. However, it may still require some larger refactoring.
2. There are at least 3 tests that cover this one (as mentioned in the description). However, the command client doesn't have unit tests at all, so I was going to create a story to add them, including the submission id generation. If you'd like, I can also do it in a separate PR that won't have to go to the release branch. I would really like to merge this one at least: https://github.com/digital-asset/daml/pull/10928

fabiotudone-da (approved):

andreaslochbihler-da:
We're now generating submission IDs in many places and I do wonder why that's needed. My preference would have been to identify a single place where this is done, i.e., either in the Ledger API command service or in the command tracker. In particular, it's not clear why we generate one in the ledger API command submission service, as everything below the command service should treat the submission ID as an opaque string without meaning that is just passed on.

This is probably not enough of an issue for the SDK 1.17 release, but something that might be worth cleaning up while we're at it.

hubert-da:
The above issues will be addressed by a newly created KVL-1107.

PR #10929, opened 2021-09-17
Make H2 query improvements common [DPP-600]

kamil-da:
### Motivation
This is a follow-up PR to #10888 . It ports H2-specific changes for all db types.

The aim is to start work on removing `participant_events` view as it proved to have negative impact on performance for H2 and Oracle.

### Testing
I've run the 800tps `sandbox-on-x` benchmark and observed no performance degradation.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
@@ -51,45 +51,47 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
       s"The following contracts have not been found: ${missingContractIds.map(_.coid).mkString(", ")}"
     )
 
-  protected def maximumLedgerTimeSqlLiteral(id: ContractId): SimpleSql[Row] = {
+  /** The data type of the ledger_effective_time column varies between different database types.
+    * For this reason we need to customise the casting method and the result type.
+    */
+  protected val NullAsLedgerEffectiveTimeType: String = queryStrategy.nullAs("BIGINT")
+
+  private def maximumLedgerTimeSqlLiteral(id: ContractId): SimpleSql[Row] = {
     import com.daml.platform.store.Conversions.ContractIdToStatement
     SQL"""
-  WITH archival_event AS (
-         SELECT participant_events.*
-           FROM participant_events, parameters
-          WHERE contract_id = $id
-            AND event_kind = 20  -- consuming exercise
-            AND event_sequential_id <= parameters.ledger_end_sequential_id
-          FETCH NEXT 1 ROW ONLY
-       ),
-       create_event AS (
-         SELECT ledger_effective_time
-           FROM participant_events, parameters
-          WHERE contract_id = $id
-            AND event_kind = 10  -- create
-            AND event_sequential_id <= parameters.ledger_end_sequential_id
-          FETCH NEXT 1 ROW ONLY -- limit here to guide planner wrt expected number of results
-       ),
-       divulged_contract AS (
-         SELECT ledger_effective_time
-           FROM participant_events, parameters
-          WHERE contract_id = $id
-            AND event_kind = 0 -- divulgence
-            AND event_sequential_id <= parameters.ledger_end_sequential_id
-          ORDER BY event_sequential_id
-            -- prudent engineering: make results more stable by preferring earlier divulgence events
-            -- Results might still change due to pruning.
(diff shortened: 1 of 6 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/common/QueryStrategy.scala
```diff
@@ -55,4 +55,6 @@ trait QueryStrategy {
 
   /** Boolean predicate */
   def isTrue(booleanColumnName: String): String
+
+  def nullAs(typeName: String): String = s"NULL::$typeName"
```
rautenrieth-da, line 59: Two comments:
- the double colon operator is not standard, I would not provide a default implementation
- type names are not portable - see for example `BIGINT` vs `NUMBER`

How about adding a `def nullAsNumber: string` instead? Then you wouldn't need the overrideable `NullAsLedgerEffectiveTimeType`.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -100,10 +64,45 @@ trait ContractStorageBackendTemplate extends ContractStorageBackend {
     if (ids.isEmpty) {
       Failure(emptyContractIds)
     } else {
-      def lookup(id: ContractId): Option[Option[Instant]] =
-        maximumLedgerTimeSqlLiteral(id).as(instantFromMicros("ledger_effective_time").?.singleOpt)(
+      def lookup(id: ContractId): Option[Option[Instant]] = {
```
rautenrieth-da, line 67: Why did you inline `maximumLedgerTimeSqlLiteral` here? Could you have just modified `maximumLedgerTimeSqlLiteral` instead?

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
```diff
@@ -116,9 +116,10 @@ private[backend] object OracleStorageBackend
     override def arrayIntersectionNonEmptyClause(
         columnName: String,
         parties: Set[Ref.Party],
-    ): CompositeSql =
-      cSQL"EXISTS (SELECT 1 FROM JSON_TABLE(#$columnName, '$$[*]' columns (value PATH '$$')) WHERE value IN (${parties
-        .map(_.toString)}))"
+    ): CompositeSql = {
+      val inParties = parties.map(p => s"'$p'").mkString(", ")
+      cSQL"EXISTS (SELECT 1 FROM JSON_TABLE(#$columnName, '$$[*]' columns (value PATH '$$')) WHERE value IN (#$inParties))"
```
rautenrieth-da, line 121: Why this change? If I understand correctly, instead of using a query parameter, parties are now hardcoded in the SQL string.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
kamil-da: The `maximumLedgerTimeSqlLiteral` was introduced in my previous PR #10888 in order to create a different implementation for H2. Currently there's a single implementation, so I reverted the `lookup()` method to have an inlined query.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleStorageBackend.scala
kamil-da: This is the fix for the issue with SQL query syntax error on Oracle. I still need to figure out the exact difference in rendering comparing to the proposed version. Without this change the active contract query lookup query fails on Oracle.

meiersi-da: If this value is really inlined then that potentially opens the query up to SQL injection attacks. Let's make sure that we pass the values as a parameter to the query.

kamil-da: I'm on it!

kamil-da:
Removing the `participant_events` view revealed a possible bug in Oracle when using the `JSON_TABLE` function on a `CLOB` column, prepared statements and an index on the same column at the same time.
We're going to postpone removing the view due to some upcoming work that is going to facilitate or possibly cover the task anyway.

PR #10930, opened 2021-09-17
More NonEmpty functions

leonidr-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
+++ b/compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
@@ -13,6 +13,7 @@
 -- ```
 module DA.NonEmpty
   ( NonEmpty(..)
+  , cons
   , append
   , map
   , nonEmpty
@@ -29,7 +30,7 @@ module DA.NonEmpty
   , foldl1A
   ) where
 
-import Prelude hiding (reverse, foldr, foldl, one, map)
+import Prelude hiding (cons, reverse, foldr, foldl, one, map)
 import Prelude qualified as P
 import DA.Foldable qualified as F
 import DA.NonEmpty.Types
@@ -72,6 +73,10 @@ instance F.Foldable NonEmpty where
 
 instance T.Traversable NonEmpty where
   mapA f l = liftA2 NonEmpty (f l.hd) (T.mapA f l.tl)
+
+-- | Prepend an element to a non-empty list.
+cons : a -> NonEmpty a -> NonEmpty a
+cons a ne = NonEmpty a (ne.hd :: ne.tl)
 
 -- | Append or concatenate two non-empty lists.
 append : NonEmpty a -> NonEmpty a -> NonEmpty a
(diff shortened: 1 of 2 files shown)


leonidr-da:
I was also going to add a `fromList` to a `DA.NonEmpty.Total` but per https://discuss.daml.com/t/to-abort-or-to-fail/3012/8, I'm not certain whether we should ActionFail or CanAbort.

compiler/damlc/daml-stdlib-src/DA/NonEmpty.daml
```diff
@@ -99,6 +106,28 @@ toList (NonEmpty head tail) = head :: tail
...
 
+-- | Find an element in a non-empty list.
+find : (a -> Bool) -> NonEmpty a -> Optional a
+find p (NonEmpty head tail)
+  | p head = Some head
+  | otherwise = L.find p tail
+
+-- | The 'deleteBy' function behaves like 'delete', but takes a
+-- user-supplied equality predicate.
+deleteBy : (a -> a -> Bool) -> a -> NonEmpty a -> Optional (NonEmpty a)
```
cocreature, line 117: Have you considered returning `[]` instead? `Optional (NonEmpty a)` doesn’t provide any extra info or type safety.

leonidr-da: Good point. I had not considered it. I think given the  trade-off of unwrapping an `Optional` or recreating the `NonEmpty` I prefer the former. Handling the case where you delete the last element is rarer than deleting an element or deleting none. But, yes, that type does look a bit weird.

leonidr-da: @cocreature Is this ok with you? It would be nice to get this done.

cocreature: sorry missed that, I think I have a preference for the `[a]` case. The type signature here just looks confusing.

leonidr-da: Can you say why it looks confusing? After this `delete` one will have to handle either a `None` or a `[]`, and there are plenty of [examples](https://docs.daml.com/daml/stdlib/DA-List.html#function-da-list-elemindex-66976) where we wrap in `Optional` to deal with the "missing" case.

cocreature: Generally, I opt for the simplest type that expresses the properties I care about. In this case `[a]` is objectively simpler than `Optional (NonEmpty a)` (both at the type and at the value level) but the more complex type provides zero extra information.

While Haskell doesn’t have `delete` on `NonEmpty` it has [filter](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List-NonEmpty.html#v:filter) which is close enough and follows my suggestion here.

leonidr-da: I don't think that simplicity is the only thing one should consider. I do think that  `Optional (NonEmpty a)` is more explicit in  expressing to the user that some operation may not return a `NonEmpty a`.  But I won't belabor this.

cocreature (approved):
Nice, thank you!

PR #10950, opened 2021-09-21
Check protobuf compatibility of `main` and PR commits w.r.t. previous stable release [KVL-1109]

fabiotudone-da:
## Description

- Splits `buf.yaml` in different modules with different checks.
- Adds a new per-`main`-or-PR-commit CI job that checks protobuf compatibility against the most recent stable release.

### Pull Request Checklist

trigger the build.


ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,54 @@
...
+check_against_stable_protos_buf_image() {
+  cd "${SRC_DIR}"
+  buf breaking --against "${TMP_STABLE_PROTOS_DIR}/${BUF_IMAGE}"
+}
+
+set_latest_stable
+echo "Checking protos against '${LATEST_STABLE}' (i.e., the most recent stable tag up to '${CURRENT_BRANCH}')"
+
+checkout_stable_protos
+
```
fabiotudone-da, line 52:
```suggestion
```

azure-pipelines.yml
```diff
@@ -202,6 +202,15 @@ jobs:
     parameters:
       trigger_sha: '$(trigger_sha)'
 
+- job: compat_stable_protobuf_pr
+  pool:
+    name: ubuntu_20_04
+    demands: assignment -equals default
+  steps:
+    - checkout: self
+      persistCredentials: true
```
garyverhaegen-da, line 211:
```suggestion
```
I don't see where you need credentials here, and I'd rather not have them if they're not useful.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+readonly SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+readonly SRC_DIR="$(dirname "${SCRIPT_DIR}")"
```
garyverhaegen-da, line 7: I'd find
```suggestion
readonly PROJECT_ROOT="$SCRIPT_DIR/.."
```
a lot easier to read here.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,53 @@
...
+
+cd "${SRC_DIR}"
+readonly CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
```
garyverhaegen-da, line 10: The `azure-pipelines.yml` file only runs on the `main` branch, so this seems unnecessary.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,53 @@
...
+
+set_latest_stable() {
+  local gitTagSearchSpace
+  if [[ "${CURRENT_BRANCH}" == "main" ]]; then
+    gitTagSearchSpace="--no-merged"
+  else
+    gitTagSearchSpace="--merged"
+  fi
+  LATEST_STABLE="$(git tag ${gitTagSearchSpace} | grep -v "snapshot" | sort -V | tail -1)"
```
garyverhaegen-da, line 19: Bit of a personal preference here, but I'm not too fond of mutating global variables from a function body. You can "return" a string from a function by using `echo` here and then binding it with
```
LATEST_STABLE="$(set_latest_stable)"
```
from the main body.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+cleanup_tmp_files() {
+  rm -rf "${TMP_STABLE_PROTOS_DIR}"
```
garyverhaegen-da, line 23: I'd expect a global variable to be defined at the top-level and before functions start referencing it. Forward references are hard on the reader.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+checkout_stable_protos() {
+  trap cleanup_tmp_files EXIT
+  TMP_STABLE_PROTOS_DIR="$(mktemp -d -t ci-stable-protos-XXXXXXXXXX)"
+  local archive;
+  archive="${TMP_STABLE_PROTOS_DIR}/archive.tar.gz"
+  git archive --output="${archive}" "${LATEST_STABLE}" daml-lf ledger-api ledger
+  tar xzf "${archive}" -C "${TMP_STABLE_PROTOS_DIR}"
```
garyverhaegen-da, line 32: This seems a bit complicated at first glance. How is it better than
```
BUF_IMAGE="$(mktemp -d)/buf.tmp"
git checkout $LATEST_STABLE
buf build -o "$BUF_IMAGE"
git checkout main
buf breaking --against "$BUF_IMAGE"
```
?

garyverhaegen-da (approved):

ci/check-protobuf-against-stable.sh
fabiotudone-da: Oh, I'd like to run this check for fix releases too. How can I do that?

ci/check-protobuf-against-stable.sh
fabiotudone-da: I tried not to touch the main work tree as I don't know much how it is used. I'll do as you suggest.

ci/check-protobuf-against-stable.sh
garyverhaegen-da: At the moment we don't have anything running after merge on release branches. What we've opted for instead is to enforce sequential merges on branches, so PR checks are what you'd want there I suppose. If you implement the next step of the design document (checking against target commit on PRs), I think that should cover it.

Bear in mind that release branches run their own version of the various YAML files, so if you want this to affect existing release branches you'll need to backport at least that bit.

ci/check-protobuf-against-stable.sh
garyverhaegen-da: This resets all its changes so shouldn't be an issue, especially with `BUF_IMAGE` out of the worktree. Even if it crashes, the worst case is it leaves the worktree with the wrong commit checked out. Every single job starts by checking out the commit it wants, so this will not affect other jobs, and this is the last step in this job so shouldn't be an issue on that front either.

fabiotudone-da: Actually, the current code is easier to test locally and (at least locally when using `direnv`) `dade assist` is not invoked again.

azure-pipelines.yml
```diff
@@ -202,6 +202,14 @@ jobs:
     parameters:
       trigger_sha: '$(trigger_sha)'
 
+- job: compat_stable_protobuf_pr
```
cocreature, line 205: I’m confused. Do you not want to run this on PRs? With this, we’ll end up not noticing it on PRs and then we get a red main.

ci/check-protobuf-against-stable.sh
fabiotudone-da: Thanks, I'll do so then. Also, the override mechanism will need to be disabled on release branches (but that makes sense regardless, I think).

azure-pipelines.yml
fabiotudone-da: The current change only ensures that we can't release if we break protobuf, which is a product-level requirement.

The per-PR-commit check, used during development, will be a subsequent change and will check against the target branch's HEAD (and it will replace the current manual check in `fmt.sh`).

ci/check-protobuf-against-stable.sh
fabiotudone-da: Done, 16158c3

ci/check-protobuf-against-stable.sh
fabiotudone-da: Done, ed1ba01

ci/check-protobuf-against-stable.sh
fabiotudone-da: This is also done as part of ed1ba01.

ci/check-protobuf-against-stable.sh
fabiotudone-da: Done, ee20aaf

ci/check-protobuf-against-stable.sh
fabiotudone-da: Done, e3cb71a

azure-pipelines.yml
cocreature: It doesn’t just block releases though, it results in a red main and breaks our CI completely. I’d rather add the per PR checks first, otherwise this will result in a lot of noise and broken CI.

fabiotudone-da: > breaks our CI completely

What does it mean? Consider that we allow making exceptions in per-PR checks, so it will still be possible to merge them.

cocreature: Having to restore compatibility in the rare case where we had a merge conflict is fine. Having to restore compatibility with a check that we didn’t even run on a PR is imho asking for this to be too frequent of an occurence.

garyverhaegen-da: Agree with @cocreature that it'd be better if we merged the per-commit one first, though I'm not _too_ concerned about merging this one first. Yes, there's a risk we'd only catch issues on main, and that would be annoying as then we have red commits on main, but:
1. I'm assuming (hoping?) we don't regularly randomly break protobufs and only rely on CI to tell us about it, so the chance of actually breaking main are not that high.
2. I'm also assuming the PR check is the very next item on your todo list so the window of opportunity is not that large.
3. As this is a check that is only run on main commits, and thus not on PRs, this being red will not block merges.
4. As this is not added to `collect_build_data` dependencies, it will not result in any Slack message.

🤔 

Actually, that last one is kinda bad. What _should_ happen if that test fails? Currently, it will just silently put a red cross in the commit view on GitHub, which is probably not all that useful. Should this perhaps send a Slack message to the team responsible for protobufs?

garyverhaegen-da: > Suppose we use the override in a PR though, because we decide to temporarily allow incompatibility; is it problematic to have the CI failing on `main` on every PR merge until we restore compatibility?

My understanding was that we wanted the option for individual commits to break compatibility _with previous commits_ in-between releases, not the compatibility with the latest release. So I would still expect this to be green on main commits, because on main we only check with latest release. Did I misunderstand something? (Also that does mean this should probably also run on PRs and thus be added to `ci/build.yml` instead of `azure-pipelines.yml`.)

fabiotudone-da: You are right about 1 and 2 and thanks for confirming 3.

As for 4, there is no single team responsible for all protobufs, some are shared (see `CODEOWNERS` for more details) and anyway work across ownership has happened lately and will likely happen again. Given that, what would you consider a useful Slack message? E.g., is sending one to the whole team better than silence?

cocreature: yeah tell-slack-failed was what I was thinking off

fabiotudone-da: > My understanding was that we wanted the option for individual commits to break compatibility with previous commits in-between releases

Yes, you are right. I wonder if there is a case where we would like to temporarily break compatibility with the latest stable but I can't think of any. I'll move the check to the PR and we can always decide to be less strict later on.

fabiotudone-da: I think having the check (also) in the PR means that it should take into account not only `main` as a target but also releases, is that correct? This means that the most recent stable tag is now relative again.

fabiotudone-da: If I understand correctly, this check will be run on:

1. `main` commits and PRs targeting `main`: we want `git tag --no-merged` because release tags happen on release branches (that are not reachable from `main`). Or simply `git tag` because we're considering the most recent anyway.
2. `release/*` commits and PRs targeting `release/*`: we want `git tag --merged` because the most recent release tag for that release branch is reachable from it.

So we need a way to distinguish between these two cases. Figuring out the merge commits is not a problem but what about the PRs'? Do we have the target branch available?

fabiotudone-da: Assuming my reasoning is correct, I think we can probably leverage [`System.PullRequest.TargetBranch`](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables).

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,26 @@
...
+
+eval "$(dev-env/bin/dade assist)"
+
+readonly BUF_IMAGE_TMPDIR="$(mktemp -d)"
+trap 'rm -rf ${BUF_IMAGE_TMPDIR}' EXIT
+
+readonly CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
+
+if [[ "${CURRENT_BRANCH}" == "main" || "${SYSTEM_PULLREQUEST_TARGETBRANCH}" == "main" ]]; then
+  GIT_TAG_SCOPE="--no-merged" # Release tags are not reachable from `main` because release branches predate them
```
garyverhaegen-da, line 14: This comment is false.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,26 @@
...
+else
+  GIT_TAG_SCOPE="--merged"    # The relevant most recent release tag is always reachable from its release branch
```
garyverhaegen-da, line 16: This comment is _mostly_ true. In the current situation, for example, this would return `1.16.0` for the current 1.17.x branch, regardless of whether we ever had a 1.16.1 (right now we don't, but it wouldn't be returned if we did).

azure-pipelines.yml
fabiotudone-da: > So I'd say --merged for PRs to release branches, and nothing for main PRs.
👍🏼 
> [...] we mostly only backport from main into release branches

It's a good idea to make this assumption explicit, I'll add a comment about that.

ci/check-protobuf-against-stable.sh
fabiotudone-da: Let me know if dbc9267 solves this.

ci/check-protobuf-against-stable.sh
fabiotudone-da: Let me know if dbc9267 solves this.

ci/check-protobuf-against-stable.sh
garyverhaegen-da: 👍

ci/build.yml
```diff
@@ -413,6 +413,14 @@ jobs:
     - template: tell-slack-failed.yml
     - template: report-end.yml
 
+- job: compatibility_stable_protobuf
+  pool:
+    name: ubuntu_20_04
+    demands: assignment -equals default
+  steps:
+    - checkout: self
+    - bash: ci/check-protobuf-against-stable.sh
```
garyverhaegen-da, line 422:
```suggestion
    - bash: ci/check-protobuf-against-stable.sh
    - template: tell-slack-failed.yml
```

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,37 @@
...
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+eval "$(dev-env/bin/dade assist)"
+
+readonly BUF_IMAGE_TMPDIR="$(mktemp -d)"
+trap 'rm -rf ${BUF_IMAGE_TMPDIR}' EXIT
+
+readonly CURRENT_BRANCH="$(git rev-parse --abbrev-ref HEAD)"
```
garyverhaegen-da, line 11: This will generally not do what you expect. On PRs, the current branch is a garbage one created by Azure, usually named just `merge` or something like that. If you want to log the source branch (I can't really see a reason to), you can use `SYSTEM_PULLREQUEST_SOURCEBRANCH`.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# For `main` and PRs targeting `main`, we simply check against the most recent stable tag.
+#
+# For release branches and PRs targeting them, we should really check against all the most recent stable tags
+# reachable from either the current branch or from previous release branches (say, against both `1.17.1` and
+# `1.16.2` created after the `release/1.17.x` branch).
+#
+# Instead, we approximate by checking only against the most recent stable tag reachable from the current branch,
+# under the assumption that if a lesser release branch contains a protobuf change, then it will also be present in
+# higher ones either through a shared commit or a back-port from `main`.
```
garyverhaegen-da, line 23:
```suggestion
# For `main` and PRs targeting `main`, we simply check against the most recent
# stable tag.
#
# For release branches and PRs targeting them, we should really check against
# all the most recent stable tags reachable from either the current branch or
# from previous release branches (say, against both `1.17.1` and `1.16.2`
# created after the `release/1.17.x` branch).
#
# Instead, we approximate by checking only against the most recent stable tag
# reachable from the current branch, under the assumption that if a lesser
# release branch contains a protobuf change, then it will also be present in
# higher ones either through a shared commit or a back-port from `main`.
```

Such large lines make it hard to read on GitHub, which for some reason limits the width of the text even when my windows are much larger 🙁

ci/check-protobuf-against-stable.sh
fabiotudone-da: b9005ea

ci/build.yml
fabiotudone-da: Is this enough to enable Slack integration?

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,37 @@
...
+readonly RELEASE_BRANCH_REGEX="^release/.*"
+GIT_TAG_SCOPE=""
+if [[ "${CURRENT_BRANCH}" =~ ${RELEASE_BRANCH_REGEX} || "${SYSTEM_PULLREQUEST_TARGETBRANCH}" =~ ${RELEASE_BRANCH_REGEX} ]]; then
```
garyverhaegen-da, line 29: This will not run on release branches, so I'm not sure what the point of the first condition is here. The only case in which `CURRENT_BRANCH` would match the regex is if we deliberately try to trick it by creating a PR **from** a `release/x` branch, which should never happen.

ci/build.yml
garyverhaegen-da: Yes, assuming no typo on my end of course.

ci/check-protobuf-against-stable.sh
garyverhaegen-da: Nothing runs on release branches because we have enabled sequential merges there, so we believe PR checks are enough.

When I say "create a PR from a release branch" here I mean one where the _source_ branch is the release branch itself. Of course PRs _to_ release branches are made from commits that have some common ancestors with the release branch, but they are made _from_ branches that contain at least one more commit than the release branch itself.

fabiotudone-da: f7ec421

ci/build.yml
```diff
@@ -413,6 +413,15 @@ jobs:
     - template: tell-slack-failed.yml
     - template: report-end.yml
 
+- job: compatibility_stable_protobuf
```
cocreature, line 416: We might want to add this to `collect_build_data` which aggregates results. It doesn’t make a huge difference atm but at least for some slack notifications it makes more sense to wait on this.

cocreature (approved):
Nice work, thanks for addressing all the feedback!

buf.yaml
```diff
@@ -30,3 +30,6 @@ lint:
 breaking:
   use:
     - WIRE_JSON
+    - FIELD_NO_DELETE
+  except:
+    - FIELD_SAME_ONEOF
```
meiersi-da, line 35:
```suggestion
    - FILE
```
@miklos-da @fabiotudone-da @moritzkiefer-da : reading `https://docs.buf.build/breaking/rules/`, I'd expect that we want to use `FILE` level compatibility for:
* daml-lf/archive/src/main/protobuf
* daml-lf/transaction/src/main/protobuf
* ledger-api/grpc-definitions
* ledger/ledger-configuration/protobuf
as they all can be referenced outside of the repo by third parties.

And we want to use at least `WIRE_JSON` with
* ledger/participant-integration-api/src/main/protobuf
* ledger/participant-state/kvutils/src/main/protobuf

Albeit `FILE` will work fine for `ledger/participant-integration-api/src/main/protobuf`

Questions:
*  @miklos-da @fabiotudone-da : would `FILE` work as well for `kvutils`?
*  @moritzkiefer-da : are you OK with `FILE` for the `daml-lf` protos?

cocreature: Looking at this I didn’t realize this was enabled for daml-lf archive protobufs and looking at the current definition I’m not sure it’s correct. You are only checking Daml-LF dev which is explicitly the version that we can and do break so I don’t think the current definition makes sense. I think the check for daml-lf that we want is different, we want to check that dev is backwards compatible to the stable LF releases (those are immutable, we check hashes) not to the dev version included in the latest SDK release. So my suggestion there would be to disable it for daml-lf/archive as part of this PR and then we (doesn’t have to be @fabiotudone-da) can add a check for that separately.

For daml-lf/transaction I’d go with `WIRE` or `WIRE_JSON` (we don’t care about JSON encoding afaik so wire would be sufficient). These are not public APIs, as long as we can read old stuff that’s fine, I don’t see why we should force us to never break generated code if we control all of it.

azure-pipelines.yml
cocreature: all good, thanks!

ci/build.yml
fabiotudone-da: 6d3da77

buf.yaml
meiersi-da: Thanks for the feedback @cocreature . 

@fabiotudone-da : reading https://docs.buf.build/breaking/configuration it looks like the right option for per-proto-group configurations is to place a `buf.yaml` at their roots instead of having a single `buf.yaml` at the root of the repository. Do you understand it the same way?

If yes, could you create the ones for
* `ledger-api/grpc-definitions` using `FILE`
* `ledger/ledger-configuration/protobuf` using `FILE`
* `ledger/participant-integration-api/src/main/protobuf` using `FILE`
* `ledger/participant-state/kvutils/src/main/protobuf` using your team's configuration of choice

fabiotudone-da: @meiersi-da:

> we want to use FILE level compatibility for
> ...
> ledger-api/grpc-definitions

Do we offer source-level compatibility for the gRPC API?

> ledger/ledger-configuration/protobuf

It looks like we persist it but we do not offer API access to it, so why is source-level compatibility important?

> would FILE work as well for kvutils?

If I remember correctly we had trouble with it but @miklos-da may remember better. Are you asking in order to try and consolidate the checks category or do you see a reason why source-level compatibility can be important here?

fabiotudone-da:
@garyverhaegen-da @moritzkiefer-da do you know why the protobuf check waits 20m+ for a Nix cache lock to be released [here](https://dev.azure.com/digitalasset/daml/_build/results?buildId=89393&view=logs&j=00351e57-e4ff-5cd1-1d0e-7baed9c3f219&t=6bfdd20f-1850-50bd-a925-9faae3198299)?

buf.yaml
meiersi-da: @fabiotudone-da : reading https://docs.buf.build/breaking/configuration it looks like the right option for per-proto-group configurations is to place a `buf.yaml` at their roots instead of having a single `buf.yaml` at the root of the repository. Do you understand the docs the same way?

If yes, could you create the ones for
* `daml-lf/transaction` using `WIRE_JSON`
* `ledger-api/grpc-definitions` using `FILE`
* `ledger/ledger-configuration/protobuf` using `FILE` as it is included in the `gRPC` services
* `ledger/participant-integration-api/src/main/protobuf` using `WIRE_JSON`
* `ledger/participant-state/kvutils/src/main/protobuf` using `WIRE_JSON`

This should also take care of @cocreature's request to not check the `daml-lf/archive/dev` versions.

meiersi-da: > @meiersi-da:
> 
> > we want to use FILE level compatibility for
> > ...
> > ledger-api/grpc-definitions

Our docs are not explicit on this, but it's what I would aim for, as we want to make it as easy as possible to build applications against our API.

> Do we offer source-level compatibility for the gRPC API?
> 
> > ledger/ledger-configuration/protobuf
> 
> It looks like we persist it but we do not offer API access to it, so why is source-level compatibility important?

It is referenced from the gRPC API

cocreature:
> @garyverhaegen-da @moritzkiefer-da do you know why the protobuf check waits 20m+ for a Nix cache lock to be released [here](https://dev.azure.com/digitalasset/daml/_build/results?buildId=89393&view=logs&j=00351e57-e4ff-5cd1-1d0e-7baed9c3f219&t=6bfdd20f-1850-50bd-a925-9faae3198299)?

looks like a node issue, let me restart

ci/build.yml
```diff
@@ -413,6 +413,15 @@ jobs:
...
+  pool:
+    name: ubuntu_20_04
+    demands: assignment -equals default
+  steps:
+    - checkout: self
```
cocreature, line 421:
```suggestion
    - checkout: self
    - bash: ci/dev-env-install.sh
      displayName: 'Build/Install the Developer Environment'
```
I think this might be what results in the nix file lock

buf.yaml
fabiotudone-da: > > > ledger/ledger-configuration/protobuf
> > 
> > 
> > It looks like we persist it but we do not offer API access to it, so why is source-level compatibility important?
> 
> It is referenced from the gRPC API

The only import of `ledger_configuration.proto` I could find is in `daml_kvutils.proto`. Am I missing something?

ci/build.yml
garyverhaegen-da: We have lots of jobs that don't do that and work fine. 🤔

cocreature: Do those jobs rely on nix artifacts like this one does? Maybe it’s just very flaky but I reset one node and then it ran into the same issue on another node.

garyverhaegen-da: They do. What seems to be issue here is that we `git checkout` to another version of nixpkgs (Linux CI nodes run the `dev-env-install` on startup, so they should have all packages pre-fetched).

I'm not entirely sure how that breaks, but, do we need to re-run the `eval "$(dev-env/bin/dade-assist)"` after the `git checkout`? When I run `dade-assist` I don't see anything that looks like it would be version-specific, but when we get a new nixpkgs sometimes it helps if I `cd` out and back in.

cocreature: Ah I think you’re right, we might need to rerun it.

fabiotudone-da: Another option would be to revert to the `git archive` solution that doesn't touch the main work tree. Which one do you think is best?

garyverhaegen-da: I can reproduce locally and running `eval` again does not help.

What happens is that we have dev-env activated, and:

- We run `buf`, which activates the dev-env script.
- The dev-env script sees that `buf` is not the right version, so it triggers a nixpkgs fetch.
- Nix then needs to un-tar its nixpkgs definition, and because dev-env is activated, tries to go through dev-env `tar` for that.
- `tar` is a dev-env script, so it tries to build itself from Nix, which is already waiting with a lock in hand.

I'm not quite sure how do work around that. Without direnv I'm not quite sure how we properly _de_activate dev-env within a Bash script. 🤔 

@fabiotudone-da I'd like a solution to the above issue in principle, but yes, for this PR it may be faster to just go back to the git-archive approach. What I don't like about that approach is that we expect the HEAD script to know about the folder structure of "random" old versions of the repo, and that may lead to subtle breakages down the road.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,40 @@
...
+# reachable from the current branch, under the assumption that if a lesser
+# release branch contains a protobuf change, then it will also be present in
+# higher ones either through a shared commit or a back-port from `main`.
+readonly RELEASE_BRANCH_REGEX="^release/.*"
+GIT_TAG_SCOPE=""
+if [[ "${SYSTEM_PULLREQUEST_TARGETBRANCH}" =~ ${RELEASE_BRANCH_REGEX} ]]; then
+  GIT_TAG_SCOPE="--merged"
+fi
+
+readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
```
garyverhaegen-da, line 33:
```suggestion
readonly LATEST_STABLE_TAG="$(eval "$(dev-env/bin/dade assist)"; git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
```

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+eval "$(dev-env/bin/dade assist)"
+
```
garyverhaegen-da, line 7:
```suggestion
```

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,40 @@
...
+if [[ "${SYSTEM_PULLREQUEST_TARGETBRANCH}" =~ ${RELEASE_BRANCH_REGEX} ]]; then
+  GIT_TAG_SCOPE="--merged"
+fi
+
+readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
+echo "Checking protobuf against tag '${LATEST_STABLE_TAG}'"
+git checkout "${LATEST_STABLE_TAG}"
+
+readonly BUF_IMAGE="${BUF_IMAGE_TMPDIR}/buf.bin"
+buf build -o "${BUF_IMAGE}"
```
garyverhaegen-da, line 38:
```suggestion
(eval "$(dev-env/bin/dade assist)"; buf build -o "${BUF_IMAGE}")
```

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,40 @@
...
+git checkout "${CURRENT_BRANCH}"
+buf breaking --against "${BUF_IMAGE}"
```
garyverhaegen-da, line 40:
```suggestion
(eval "$(dev-env/bin/dade assist)"; buf breaking --against "${BUF_IMAGE}")
```

garyverhaegen-da (commented):
Do not activate dev-env "globally" in the script, but only in subshells.

I _think_ this should work.

ci/build.yml
fabiotudone-da: > What I don't like about that approach is that we expect the HEAD script to know about the folder structure of "random" old versions of the repo, and that may lead to subtle breakages down the road.

I'll try to extract everything but not to move in there (or at least to avoid extracting the dev-env portions).

fabiotudone-da: Just noticed your suggestion below, I'll try that first.

fabiotudone-da: Yep (fba0582), this time [it only took half-a-minute](https://dev.azure.com/digitalasset/daml/_build/results?buildId=89412&view=logs&jobId=00351e57-e4ff-5cd1-1d0e-7baed9c3f219&j=00351e57-e4ff-5cd1-1d0e-7baed9c3f219&t=6bfdd20f-1850-50bd-a925-9faae3198299). Thanks!

ci/check-protobuf-against-stable.sh
fabiotudone-da: It was not needed for `git` invocations.

ci/check-protobuf-against-stable.sh
fabiotudone-da: fba0582

ci/check-protobuf-against-stable.sh
fabiotudone-da: fba0582)

ci/check-protobuf-against-stable.sh
fabiotudone-da: fba0582

buf.yaml
meiersi-da: @fabiotudone-da : it looks like I was mistaken wrt its inclusion in the gRPC Ledger API. It is though present in several migrations for the IndexDB, as the `configuration` column in the `configuration_entries` table is encoded using that protobuf. See https://github.com/digital-asset/daml/search?q=ledger_configuration&type=code

meiersi-da: So using `WIRE_JSON` for `ledger/ledger-configuration/protobuf` seems like the right option.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,41 @@
...
+  GIT_TAG_SCOPE="--merged"
+fi
+
+readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
+echo "Checking protobuf against tag '${LATEST_STABLE_TAG}'"
+git checkout "${LATEST_STABLE_TAG}"
+
+readonly BUF_IMAGE="${BUF_IMAGE_TMPDIR}/buf.bin"
+(eval "$(dev-env/bin/dade assist)" ; buf build -o "${BUF_IMAGE}")
+git checkout "${CURRENT_BRANCH}"
```
hubert-da, line 40: Just making sure, we're safe to switch branches here, right?

hubert-da: Thanks!

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,34 @@
...
+# check is enough.
+readonly RELEASE_BRANCH_REGEX="^release/.*"
+GIT_TAG_SCOPE=""
+if [[ "${SYSTEM_PULLREQUEST_TARGETBRANCH}" =~ ${RELEASE_BRANCH_REGEX} ]]; then
+  GIT_TAG_SCOPE="--merged"
+fi
+
+readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
+echo "Checking protobuf against tag '${LATEST_STABLE_TAG}'"
+(eval "$(dev-env/bin/dade assist)" ; buf breaking --against ".git#branch=${LATEST_STABLE_TAG}")
```
hubert-da, line 34: 👍🏻

buf.yaml
SamirTalwar-DA: Why not `WIRE`?

fabiotudone-da: I think it was mostly for consistency, i.e. it can be confusing for the reader to see changing field names. But I don't recall precisely. It was introduced [here](https://github.com/digital-asset/daml/pull/10411/files#diff-1a5ba9cba93e971f532139f694d7da802776bfe578e3f753b9c3f25968dbf42dR27) so @miklos-da may have a better answer.

miklos-da: As we're providing bindings, I'm not sure source-level compatibility is required for the Ledger API (hence, `WIRE` should suffice).
For `ledger/ledger-configuration/protobuf` `WIRE` should be also enough as it is only being used internally by kvutils (hence, we don't care about changing field names).

miklos-da: @meiersi-da Yes, I'm aware that we are importing `com/daml/daml_lf_dev/daml_lf.proto` in `daml_kvutils.proto`, this was required for embedding Daml-LF data and improving compression efficiency (we could replace relevant Daml-LF message references with an `Any` message type for more flexibility). I agree @cocreature's team should take care of putting the right backwards compatibility checks in place that works for their proto files.

fabiotudone-da: > As we're providing bindings, I'm not sure source-level compatibility is required for the Ledger API (hence, WIRE should suffice).

The Ledger gRPC API is part of public documentation, so I think it's more important there to have stability for field names; for this reason I propose to keep `WIRE_JSON`.

We could also try and upgrade to `FILE` if we decide to strive for source-level compatibility, for cases (present or potential, not sure) when an user wants to bypass bindings and still upgrade easily, but I'd prefer we take a conscious and shared decision about this at a product reasoning level (and then also make it explicit in public docs).

fabiotudone-da: abc701 is the best I could do. Notes:

-  Without us publishing `buf` modules at this stage (so that we can declare dependencies between them), we must explicitly list all roots containing referenced `.proto` files and parent paths cannot be referenced, so I had to move `buf-kvutils.yaml` up to the root. For consistency, the other `buf` configuration files are at the root level too, prefixed with `buf-` so at least they show up close to each other.

- Using the `buf` registry to retrieve Google's `status.proto` requires the dependent to also adhere to `buf` module packaging (in particular, dedicated root with a single configuration that must be named `buf.yaml`) and, since we don't do that due to the previous point, I had to commit `status.proto`.

- On the other hand, this structure has the advantage that the check script doesn't have to know anything about the directory structure.

Improvements are possible but I'd leave them for further PRs to avoid scope creep, especially considering that they may end up being possible only after reorganizing the code and/or reviewing dependencies between our protos.

@meiersi-da @miklos-da I'd like especially your feedback (and approval, if you agree) about two main things in this context:

- Whether you see serious issues with this (transitional) `buf` structure for the purpose of this PR.
- Whether we can agree on the modules split and the checks for the purpose of this PR.

fabiotudone-da: Also note that I had to dismantle at this stage the previous manual `fmt.sh`-based protobuf checks.

ci/build.yml
fabiotudone-da: @cocreature I'd resolve this if I did it right

garyverhaegen-da: I'm not quite sure what @cocreature meant here. As it stands, what this will do is make this test part of the individual opt-in CI notification on #team-daml-ci, which as far as I'm aware nobody working on this uses.

If that's all you wanted out of this, hten yes, you did it right.

buf.yaml
miklos-da: @fabiotudone-da This sounds great to me, we can always fine-tune the rules later!

miklos-da (approved):

buf-ledger-api.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+build:
+  roots:
+    - ledger-api/grpc-definitions
+    - ledger-api/3rdparty/grpc-definitions
+
+breaking:
+  use:
+    - WIRE_JSON
+    - FIELD_NO_DELETE
```
meiersi-da, line 14:
```suggestion
    # Set to FILE to ensure that all breakages of the gRPC Ledger API are explicitly
    # discussed with Engineering and Product Leadership.
    - FILE
```

If `FILE` works then, I'd really like to use it; even though we don't guarantee it explicitly to our customers.

My rationale is that backwards-compatibility for the Ledger API matters for the UX of our customers, and is something we want to make a very explicit choice to break. Requiring `FILE` here ensures that the required discussions will happen.

buf-kvutils.yaml
```diff
@@ -21,12 +17,8 @@ build:
...
-lint:
-  use:
-    - DEFAULT
-  except:
-    - PACKAGE_VERSION_SUFFIX
-
 breaking:
   use:
-    - WIRE_JSON
+    - WIRE
```
meiersi-da, line 22: @miklos-da : can you explain your rationale for `WIRE` instead of `WIRE_JSON`. Ideally, we'd add that rationale as a comment here so future developers will be able to build on it.

buf-participant-integration-api.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+version: v1beta1
+
+build:
+  roots:
+    - ledger/participant-integration-api/src/main/protobuf
+
+breaking:
+  use:
+    - WIRE
```
meiersi-da, line 12:
```suggestion
    # Using WIRE_JSON here to ensure we have the option of also using JSON encodings of the
    # .proto values stored in the IndexDB.
    - WIRE_JSON
```

meiersi-da (changes requested):
Thanks @fabiotudone-da for moving this consistently forward.

Please let's resolve the comments I've made in this review before merging.

ci/check-protobuf-against-stable.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
fabiotudone-da: > How can a developer run these checks locally?

I'll add a comment about defining `SYSTEM_PULLREQUEST_TARGETBRANCH` for a local run. I think your issue after that is unrelated and has rather something to do with symlinks and dev-env (I had it as well, then decided to shorten my paths and abandon symlinks b/c I didn't want to spend time troubleshooting that).

> I assume that a developer would want to run these checks locally as well, when working on the protobufs? Or when debugging a ci failure from these checks.

Mostly the latter, as during development we are aiming (in upcoming PRs) for a stricter default, i.e. not breaking w.r.t. the target branch.

buf-ledger-api.yaml
fabiotudone-da: 99df542

ci/check-protobuf-against-stable.sh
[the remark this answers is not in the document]
meiersi-da: Thanks. The issue is unrelated indeed; and your suggestion fixed this.

Can you also include the comment that `If you encounter issues with the dev-env and symlinks, then please run the check from the non-symlinked directory.`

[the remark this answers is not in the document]
fabiotudone-da: 8b06120

buf.yaml
fabiotudone-da: Resolving this thread because further feedback is being addressed in new threads.

buf-kvutils.yaml
fabiotudone-da: Proposed a wording in df65d20.

buf-kvutils.yaml
```diff
@@ -3,30 +3,27 @@
...
 breaking:
   use:
-    - WIRE_JSON
+    # WIRE is enough here as KVUtils data is internal and will be transmitted
+    # and stored in binary protobuf format for the foreseeable future.
+    - WIRE
+  except:
+    # Moving an individual field into a new `one_of` retains binary backwards
+    # compatibility.
+    - FIELD_SAME_ONEOF
```
fabiotudone-da, line 29: Removing this check also allows for incompatible changes but we have to use it at least until 1.17.0 as there are no finer-grained checks for backwards-compatible cases.

ci/check-protobuf-against-stable.sh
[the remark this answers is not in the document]
fabiotudone-da: > Can you also include the comment that If you encounter issues with the dev-env and symlinks, then please run the check from the non-symlinked directory.

I could but this is true in general, not just for this script. I'd rather propose an addition in a more general README in a separate PR.

buf-ledger-api.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - ledger-api/grpc-definitions
+    - ledger-api/3rdparty/grpc-definitions
+
+breaking:
+  use:
+    # Set to FILE to ensure that all breakages of the gRPC Ledger API are explicitly
+    # discussed with Engineering and Product Leadership.
+    - FILE
+  except:
+    - FILE_NO_DELETE # Avoids errors due caused by refactored buf modules.
```
fabiotudone-da, line 17:
```suggestion
    - FILE_NO_DELETE # Avoids errors due to refactored buf modules.
```

buf-ledger-api.yaml
```diff
@@ -0,0 +1,17 @@
...
+    - ledger-api/grpc-definitions
+    - ledger-api/3rdparty/grpc-definitions
+
+breaking:
+  use:
+    # Set to FILE to ensure that all breakages of the gRPC Ledger API are explicitly
+    # discussed with Engineering and Product Leadership.
+    - FILE
+  except:
+    - FILE_NO_DELETE # Avoids errors due to refactored `buf` modules.
```
fabiotudone-da, line 17: We can remove this exception after 1.17.0 as long as we don't change the `buf` modules definitions.

meiersi-da: ack.

buf-kvutils.yaml
meiersi-da: ack.

ci/check-protobuf-against-stable.sh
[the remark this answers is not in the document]
meiersi-da: I've never had problems otherwise, but I've had the problem here. Your call though.

meiersi-da (approved):
Thanks @fabiotudone-da

ci/check-protobuf-against-stable.sh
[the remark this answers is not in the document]
fabiotudone-da: If I remember correctly, `bazel` itself didn't work correctly for me in a devenv from symlinks and so basically the whole environment was unusable.

buf-kvutils.yaml
miklos-da: Saying that we need to be able to read data persisted by previous versions of the code should be enough. I.e., I don't think we need to elaborate, as the previous implies we don't care about field names just the semantics of the data. We should never persist data in JSON format from kvutils.

miklos-da (approved):

buf-kvutils.yaml
```diff
@@ -3,30 +3,27 @@
...
-    # We have to exclude these to avoid duplicate definitions of e.g. Unit within the same package.
+    # We have to exclude these to avoid duplicate definitions of e.g. Unit
+    # within the same package.
     - daml-lf/archive/src/main/protobuf/com/daml/daml_lf_1_11/
     - daml-lf/archive/src/main/protobuf/com/daml/daml_lf_1_12/
     - daml-lf/archive/src/main/protobuf/com/daml/daml_lf_1_13/
     - daml-lf/archive/src/main/protobuf/com/daml/daml_lf_1_14/
     - daml-lf/archive/src/main/protobuf/com/digitalasset/
 
-lint:
```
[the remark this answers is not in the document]
fabiotudone-da: Alright. We may want to automate it at some point but it's out-of-scope, I'll just reintroduce the definition.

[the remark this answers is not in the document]
fabiotudone-da: a63885a09724b176d6736a947691c685941f79cb

buf-kvutils.yaml
fabiotudone-da: 370adc2

ledger-api/3rdparty/grpc-definitions/google/rpc/status.proto
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
fabiotudone-da, line 3:
```suggestion
```

fabiotudone-da: Remove extra (C) header from third-party proto

fabiotudone-da: 5272db9

ledger-api/3rdparty/grpc-definitions/google/rpc/status.proto
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
-// SPDX-License-Identifier: Apache-2.0
-
```
fabiotudone-da: It looks like currently we would have to keep this to make `dade-copyright-headers` happy. @moritzkiefer-da @garyverhaegen-da Is there another option and, if not, do you see a problem with that?

cocreature: You can place a `NO_AUTO_COPYRIGHT` file in the directory to disable it. That said, I’m not sure why you need to copy this file in the first place. We already have the same protobuf as a Bazel dep via `@com_github_googleapis_googleapis//google/rpc:status_proto`. Why do we need it twice?

fabiotudone-da: I'll try and look into how to use the one from the Bazel dep with `buf`.

fabiotudone-da: It looks like bazel deps are cached at the user level but `buf` forbids access to files outside of the configuration root (in the work tree). I think the script would have to copy the file into the work tree. I wouldn't like the check script to know individual protos, so perhaps the `NO_AUTO_COPYRIGHT` solution is better in this case. wdyt?

cocreature: hm that’s a bit sad. I don’t like having to keep the two in sync but I guess status.proto is unlikely to change frequently if at all so NO_AUTO_COPYRIGHT is fine with me.

fabiotudone-da: Moved it to the already existing `/3rdparty`. Thanks!

PR #10954, opened 2021-09-21
Add conversions from/to interfaces in Daml parser

cocreature:
See https://github.com/digital-asset/ghc/pull/71 for the GHC changes

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion/Primitives.hs
```diff
@@ -436,6 +436,33 @@ convertPrim _ "UTryCatch" ((TUnit :-> TUpdate t1) :-> (TBuiltin BTAnyException :
...
+  ETmLam (mkVar "cid", TContractId (TCon iface)) $
+    if iface == tpid
+      then EUpdate $ UPure (TOptional (TContractId (TCon iface))) (ESome (TContractId (TCon iface)) $ EVar $ mkVar "cid")
+      else EUpdate $
+           UBind
+             (Binding (mkVar "iface", TCon iface) (EUpdate $ UFetchInterface iface (EVar $ mkVar "cid"))) $
+             EUpdate $ UPure (TOptional (TContractId (TCon tpid))) $ ECase (EFromInterface iface tpid (EVar $ mkVar "iface"))
+               [ CaseAlternative CPNone (ENone (TContractId (TCon tpid)))
+               , CaseAlternative (CPSome $ mkVar "_") (ESome (TContractId (TCon tpid)) (EBuiltin BECoerceContractId `ETyApp` TCon iface `ETyApp` TCon tpid `ETmApp` (EVar $ mkVar "cid")))
+               ]
```
sofiafaro-da, line 449: I feel like this is very far from a conversion function, since it requires fetch privileges. How do you feel about just having this be another form of `coerceContractId` instead (i.e. `ContractId Token -> ContractId MyToken`, instead of `ContractId Token -> Update (Optional (ContractId MyToken))`)? The failure would be deferred until you actually try to use the contract id.

sofiafaro-da (approved):
Looks great, thanks!

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion/Primitives.hs
cocreature: hm I think I’d prefer to have people call `coerceContractId` in that case explicitly. Having a synonym which provides no extra safety seems somewhat misleading. The other option is to omit it completely but having some form of this seems relativeyl reasonable.

sofiafaro-da: Ok, makes sense!

PR #10956, opened 2021-09-21
[DPP-586] Upgrade to netty 4.1.67.Final and netty-tcnative-boringssl-static 2.0.40.Final

pbatko-da:
1. Upgrading netty to 4.1.67.Final. 
    Release notes: https://netty.io/news/2021/08/16/4-1-67-Final.html
2. Upgrading netty-tcnative-boringssl-static to corresponding version as listed here:
    https://github.com/netty/netty/blob/netty-4.1.67.Final/pom.xml#L511: `<tcnative.version>2.0.40.Final</tcnative.version>`

NOTE: As of this upgrade TLS 1.1 and 1.0 are getting disabled by default  (see https://netty.io/news/2021/05/19/4-1-65-Final.html)

We're not upgrading to Netty 4.1.68 (the newest as of Sep 21) as it seems[1] to require at least Java 8u272 while our Windows CI uses 8u201.

[1] https://github.com/netty/netty/pull/11604/files#diff-d9520f8137242d465d6e625873bfad7b7e27fb10168c144236757d2ec1141b1dL370


mziolekda (approved):
It looks good.
Please add a CHANGELOG entry for one of the commits in this PR stating that the support for old TLS versions is being dropped.
As I understand we have shied away from bumping the required JRE version, if we did that it would have to be communicated as well.

pbatko-da:
> It looks good.
> Please add a CHANGELOG entry for one of the commits in this PR stating that the support for old TLS versions is being dropped.
> As I understand we have shied away from bumping the required JRE version, if we did that it would have to be communicated as well.

@mziolekda Which component shall I mention? https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md#list-of-components-for-changelog-entries

mziolekda:
@pbatko-da These three:
Daml on SQL
Integration Kit
Sandbox

PR #10961, opened 2021-09-21
call npm instead of npm-cli.js in the puppeteer example

richardkapolnai-da:
I suppose this appears here:
https://docs.daml.com/getting-started/testing.html


cocreature (commented):
This breaks CI with `    spawn npm ENOENT`. iirc `npm` is a symlink and this doesn’t work here or something like that.

richardkapolnai-da:
> This breaks CI with ` spawn npm ENOENT`. iirc `npm` is a symlink and this doesn’t work here or something like that.

Oh i see. On my machine there is no `npm-cli.js` (installed npm via Snap), so I thought this is an issue. So the issue is with my npm installation I guess. Closing the PR.

PR #10963, opened 2021-09-21
Drop jdk8 override

cocreature:
The comment here is no longer accurate, Nix upgraded since so we
actually pin an older version for no reason. Perhaps somewhat
ironically while Nix upgraded they left the comment in their
derivation to not upgrade because of jvmci.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


garyverhaegen-da (approved):
Works for me. Thanks for looking into this!

PR #10967, opened 2021-09-21
Switch triggers service time provider default to wall-clock time

### LINKED ISSUE #10957: Make wall clock time default in Trigger service
Trigger service defaults to _static time_ while Sandbox defaults to _wall clock time_. This behavior can cause configuration troubles for a user who intuitively would expect consistent default time.

Repro:
```
$ daml trigger-service --help | grep "static time"
  -w, --wall-clock-time    Use wall clock time (UTC). When not provided, static time is used.
$ daml sandbox --help | grep "static time"
  -s, --static-time        Use static time. When not specified, wall-clock-time is used.
```

stefanobaghino-da:
Fixes #10957

changelog_begin
[Triggers Service] The service now starts by default using wall-clock time instead
of static time. If you want to run using static time, you need to do so explicitly
using the new '-s' or '--static-time' CLI option. If you were already using '-w'
or '--wall-clock-time' the flag has no effect. It's anyway safe to leave it there.
changelog_end

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

PR #10969, opened 2021-09-21
Limit maximum number of concurrent tests

rautenrieth-da:
This PR limits the number of concurrent tests in the ledger API test tool. Otherwise the more CPUs you have, the flakier the test becomes.

Since our CI uses machines with 8 CPUs, this will reduce the number of concurrent tests from 8 to 4. This will increase the CI time slightly for fast ledgers, but should make the test more stable for slow ledgers (see below).

An alternative would be to set the upper bound to 8, and add `--concurrent-test-runs=2` to all sandbox conformance tests.

### Effect on the overall test time: sandbox classic (H2 index+ledger)

`//ledger/sandbox-classic:conformance-test-wall-clock-append-only-postgresql`, test run times on a local machine with 16 CPUs:
- `--concurrent-test-runs=16`: FAILED in 380s (7 tests timed out)
- `--concurrent-test-runs=8`: FAILED in 379s (1 test timed out)
- `--concurrent-test-runs=4`: 401s
- `--concurrent-test-runs=2`: 414s
- `--concurrent-test-runs=1`: 426s

### Effect on the overall test time: sandbox-on-x (H2 index, no-op ledger)

`//ledger/sandbox-on-x:conformance-test`, test run times on a local machine with 16 CPUs:
- `--concurrent-test-runs=16`: 27s
- `--concurrent-test-runs=8`: 32s
- `--concurrent-test-runs=4`: 42s
- `--concurrent-test-runs=2`: 66s

### Effect on the overall test time: ledger-on-sql (H2 index, kvutils Postgres ledger)

`//ledger/ledger-on-sql:conformance-test-append-only-postgres`, test run times on a local machine with 16 CPUs:
- `--concurrent-test-runs=16`: 175s
- `--concurrent-test-runs=8`: 197s
- `--concurrent-test-runs=4`: 208s
- `--concurrent-test-runs=2`: 254s

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
@@ -11,4 +11,8 @@ object Defaults {
 
   val TimeoutScaleFactor: Double = 1.0
 
+  // Neither ledgers nor participants scale perfectly with the number of processors.
+  // We therefore limit the maximum number of concurrent tests, to avoid overwhelming the ledger.
+  val ConcurrentRuns: Int = List(Runtime.getRuntime.availableProcessors(), 4).min
+
 }
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
@@ -13,6 +13,6 @@ object Defaults {
 
   // Neither ledgers nor participants scale perfectly with the number of processors.
   // We therefore limit the maximum number of concurrent tests, to avoid overwhelming the ledger.
-  val ConcurrentRuns: Int = List(Runtime.getRuntime.availableProcessors(), 4).min
+  val ConcurrentRuns: Int = sys.runtime.availableProcessors min 4
 
 }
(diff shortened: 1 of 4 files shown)


stefanobaghino-da (approved):
Wise! Please also edit the CLI description for the `--concurrent-test-runs` option in `com.daml.ledger.api.testtool.Cli`:
```scala
opt[Int](name = "concurrent-test-runs")
  .optional()
  .action((v, c) => c.copy(concurrentTestRuns = v))
  .text("Number of tests to run concurrently. Defaults to the number of available processors")
```
The last line would no longer be true.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/tests/Defaults.scala
```diff
@@ -11,4 +11,8 @@ object Defaults {
 
   val TimeoutScaleFactor: Double = 1.0
 
+  // Neither ledgers nor participants scale perfectly with the number of processors.
+  // We therefore limit the maximum number of concurrent tests, to avoid overwhelming the ledger.
+  val ConcurrentRuns: Int = List(Runtime.getRuntime.availableProcessors(), 4).min
```
SamirTalwar-DA, line 16: You can use `sys.runtime`, and `Int#min`.

```suggestion
  val ConcurrentRuns: Int = sys.runtime.availableProcessors min 4
```

SamirTalwar-DA (approved):
I think this is a good idea. I also think it would be worth running these in parallel (no longer exclusively), but that will be more work.

kamil-da (approved):
Looks good!

Minor: In `ledger/participant-state/kvutils/BUILD.bazel` there's explicitly stated reducing concurrent test runs to 4, which is now default, so that line could be removed.

rautenrieth-da:
> I also think it would be worth running these in parallel (no longer exclusively)

I did not understand this part. We do run most tests in parallel, did you mean getting rid of tests that run exclusively? Some of them need to run exclusively because they touch "global state" (e.g., tests that change the ledger configuration).

mziolekda (approved):

PR #10978, opened 2021-09-22
participant-integration-api: Improve the CommandConfiguration docs.

SamirTalwar-DA:
Explaining the limits in more detail.

There were some discrepancies between the documentation and the code; in particular, if there are too many commands in flight, new ones will just be held in the queue.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
+++ b/ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
@@ -394,7 +394,7 @@ object Config {
             config.copy(commandConfig = config.commandConfig.copy(maxCommandsInFlight = value))
           )
           .text(
-            "Maximum number of submitted commands waiting for completion for each party (only applied when using the CommandService). Overflowing this threshold will cause back-pressure, signaled by a RESOURCE_EXHAUSTED error code. Default is 256."
+            s"Maximum number of submitted commands for which the CommandService is waiting to be completed in parallel, for each distinct set of parties, as specified by the `act_as` property of the command. Reaching this limit will cause new submissions to wait in the queue before being submitted. Default is ${CommandConfiguration.default.maxCommandsInFlight}."
           )
 
         opt[Int]("input-buffer-size")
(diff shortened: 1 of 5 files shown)


ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
```diff
@@ -394,7 +394,7 @@ object Config {
             config.copy(commandConfig = config.commandConfig.copy(maxCommandsInFlight = value))
           )
           .text(
-            "Maximum number of submitted commands waiting for completion for each party (only applied when using the CommandService). Overflowing this threshold will cause back-pressure, signaled by a RESOURCE_EXHAUSTED error code. Default is 256."
+            s"Maximum number of submitted commands for which the CommandService is waiting to be completed in parallel, for each distinct set of parties, as specified by the `act_as` property of the command. Reaching this limit will cause new submissions to wait in the queue before being submitted. Default is ${CommandConfiguration.default.maxCommandsInFlight}."
```
fabiotudone-da, line 397: Since this is user-facing, I'd like to avoid any implementation-specific wording: I propose to either remove "in the queue" or, if we want to make explicit that there is a waiting order, removing just "the" (provided it is correct English).

fabiotudone-da (approved):

oliverse-da (approved):
LGTM

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
SamirTalwar-DA: This queue is what we refer to as the "input buffer" in the next configuration option, so maybe that would be better wording?

```suggestion
            s"Maximum number of submitted commands for which the CommandService is waiting to be completed in parallel, for each distinct set of parties, as specified by the `act_as` property of the command. Reaching this limit will cause new submissions to wait in the input buffer before being submitted. Default is ${CommandConfiguration.default.maxCommandsInFlight}."
```

hubert-da (approved):

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Config.scala
fabiotudone-da: This is really nitpicking on my part, so I think you can safely leave it as it is, but I don't think `the input buffer` is better from this perspective.

SamirTalwar-DA:
Merging. If others want to say anything, their comments can be addressed later.

PR #10983, opened 2021-09-22
[DPP-622] Add conformance tests that verifies TLSv1.0 and TLSv1 are disabled.

pbatko-da:
Because they are deprecated https://docs.microsoft.com/en-us/lifecycle/announcements/transport-layer-security-1x-disablement.

### DIFF
--- a/ledger/ledger-api-test-tool/BUILD.bazel
+++ b/ledger/ledger-api-test-tool/BUILD.bazel
@@ -284,7 +284,6 @@ conformance_test(
         "--cacrt $$(rlocation $$TEST_WORKSPACE/$(rootpath //ledger/test-common/test-certificates:ca.crt))",
         "--pem $$(rlocation $$TEST_WORKSPACE/$(rootpath //ledger/test-common/test-certificates:server.pem.enc))",
         "--tls-secrets-url https://raw.githubusercontent.com/digital-asset/daml/main/ledger/test-common/files/server-pem-decryption-parameters.json",
-        "--min-tls-version 1.2",
     ],
     test_tool_args = [
         "--verbose",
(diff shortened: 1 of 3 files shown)


ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -264,6 +264,37 @@ conformance_test(
     ],
 )
 
+conformance_test(
```
pbatko-da, line 267: FYI @mziolekda @meiersi-da 
Re verifying that TLS 1 and 1.1 are disallowed (https://github.com/digital-asset/daml/pull/10898#pullrequestreview-756348688) this a conformance test for that.

ledger/ledger-api-test-tool/BUILD.bazel
```diff
@@ -264,6 +264,37 @@ conformance_test(
...
+    ports = [6865],
+    server = "//ledger/ledger-on-memory:app",
+    server_args = [
+        "--contract-id-seeding=testing-weak",
+        "--participant=participant-id=example,port=6865",
+        "--crt $$(rlocation $$TEST_WORKSPACE/$(rootpath //ledger/test-common/test-certificates:server.crt))",
+        "--cacrt $$(rlocation $$TEST_WORKSPACE/$(rootpath //ledger/test-common/test-certificates:ca.crt))",
+        "--pem $$(rlocation $$TEST_WORKSPACE/$(rootpath //ledger/test-common/test-certificates:server.pem.enc))",
+        "--tls-secrets-url https://raw.githubusercontent.com/digital-asset/daml/main/ledger/test-common/files/server-pem-decryption-parameters.json",
+        "--min-tls-version 1.2",
```
pbatko-da, line 287: Once Netty upgrade PR is merged this test shall pass without `--min-tls-version 1.2`.

Alternatively we can code defensively and filter out protocols lower than TLS 1.1 before they are passed to initialize server's SSL context. And/or add an assertion to fail fast when server's SSL context has been started with TLS 1.1 or lower enabled.
I haven't added such assertion because I saw value in starting a server with e.g. TLS 1.1 enabled and seeing that can connect with a client on a TLS 1.1. And subsequently seeing that a server on TLS 1.2 fails with the same client.

mziolekda (approved):

PR #10964, opened 2021-09-21
kvutils: Add structured error data to the error metadata [KVL-1032]

hubert-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #10992, opened 2021-09-22
kvutils: Add metadata to `Err` [KVL-1032]

hubert-da:
A follow-up to: 
- https://github.com/digital-asset/daml/pull/10964

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
```diff
@@ -602,6 +602,7 @@ message MissingInputState {
 }
 message InvalidParticipantState {
   string details = 1;
+  map<string, string> metadata = 2;
```
hubert-da, line 605: This is more or less what we discussed with @SamirTalwar-DA today. I think this is the smallest change to propagate more information from the `Err` without introducing more `Rejection`s. @SamirTalwar-DA, @miklos-da WDYT? Is it worth doing?

SamirTalwar-DA: I think it's a good idea to introduce this first, and then break down the subclasses into their own error types in the future.

If you do it the other way around you have to do both at once.

SamirTalwar-DA:
Why is this PR targeting another branch, rather than `main`?

Can we wait until that other branch is merged to merge this one, or are you expecting that we need to merge this one first?

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
miklos-da: Agreed, this looks like a good approach.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
```diff
@@ -567,6 +581,7 @@ private[state] object Conversions {
           buildStatus(
             Code.ABORTED,
             s"Inconsistent: Missing input state for key ${rejection.getKey.toString}",
+            Map("key" -> rejection.getKey.toString),
```
miklos-da, line 584: This will produce an output with newlines. Given that the goal of these pieces of metadata is not to print them out on the console, I'd suggest using [`shortDebugString`](https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/TextFormat.Printer.html#shortDebugString-com.google.protobuf.MessageOrBuilder-) instead, as that doesn't produce newlines and is still human readable.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -104,123 +108,227 @@ class ConversionsSpec extends AnyWordSpec with Matchers with OptionValues {
...
+          finalReason.definiteAnswer shouldBe false
+          val actualDetails = finalReasonToDetails(finalReason)
+          actualDetails should contain allElementsOf (expectedAdditionalDetails)
+        }
+      }
+
+      "produce metadata that can be easily parsed" in {
+        forAll(
+          Table[Rejection, String, String => Any, Any](
+            ("rejection", "metadataKey", "metadataParser", "expectedParsedMetadata"),
```
miklos-da, line 222: Nit: for consistency, you might want to use English expressions instead of camelCasedOnes here as well (e.g., `metadataKey` => `metadata key`).

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Conversions.scala
hubert-da: Sorry, this is something that has been merged in the previous PR. We decided to leave this one, as it's possible to parse it with: `TextFormat.parse`, and this is what is demonstrated by a test. Are you ok with that?

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_kvutils.proto
fabiotudone-da: Do you have an idea about how the proto will evolve?

hubert-da: To me, it makes sense to actually keep the `InvalidParticipantState` error, as this should be a really rare one. So, if you ask me, I don't think we're gonna change the errors soon.

SamirTalwar-DA (approved):

PR #10979, opened 2021-09-22
bump dotnet

garyverhaegen-da:
This bumps dotnet to the version required by the latest azuresigntool, and pins azuresigntool for the future.

As usual for live CI upgrades, this will be rolled out using the blue/green approach. I'll keep each deployed commit in this branch.

CHANGELOG_BEGIN
CHANGELOG_END

PR #10993, opened 2021-09-22
release 1.18.0-snapshot.20210921.7889.1.1b473c2b

garyverhaegen-da:
The `main` branch build for #10971 failed. #10979 should have fixed it, so let's try again.

This needs to be merged _after_ #10979.

CHANGELOG_BEGIN
CHANGELOG_END


stefanobaghino-da (approved):
https://www.youtube.com/watch?v=dQw4w9WgXcQ

cocreature (approved):

garyverhaegen-da:
So this is _also_ a bust, because (of course) the changes in #10979 need to be picked up, so this should target a commit after that rather than try to do the same commit as before. I'll make a new release PR after #10996 is merged (as we can't do both because git conflicts).

garyverhaegen-da:
After discussion with the security team, we have decided that failed releases that went far enough to publish artifacts somewhere (this one published to mvn and npm) still count as standard changes.

PR #11001, opened 2021-09-23
Add MINIMAL pragma for Additive type class

### LINKED ISSUE #11000: Additive instance without `negate` causes GRPC error 'deadline exceeded'
Minimal example:

https://github.com/lucianojoublanc-da/additivebug

There are two issues
* compiler doesn't warn about incomplete definition of `instance Additive`
* using `-` in the example triggers above-said error, when running `daml test`

akrmn:
This ensures that the compiler gives an error if both `(-)` and `negate` are missing in an instance

Fixes #11000

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/tests/daml-test-files/IncompleteAdditiveInstance.daml
+++ b/compiler/damlc/tests/daml-test-files/IncompleteAdditiveInstance.daml
@@ -1,0 +1,17 @@
+-- Copyright (c) 2020, Digital Asset (Switzerland) GmbH and/or its affiliates.
+-- All rights reserved.
+
+-- @ERROR range=14:10-14:23; • No explicit implementation for; either ‘-’ or ‘negate’; • In the instance declaration for ‘Additive Expr’
+
+module IncompleteAdditiveInstance where
+
+data Numero = Zero | Uno | Dos | Tres
+
+data Expr = Add (Expr, Expr)
+          | Neg Expr
+          | Const Numero
+
+instance Additive Expr where
+  (+) = curry Add
+  -- negate = Neg -- uncomment to fix!
+  aunit = Const Zero
--- a/compiler/damlc/daml-prim-src/GHC/Num.daml
+++ b/compiler/damlc/daml-prim-src/GHC/Num.daml
@@ -46,6 +46,7 @@ infixl 7 %
 -- * `negate` gives the additive inverse, ie: `x + negate x` = `aunit`
 --
 class Additive a  where
+    {-# MINIMAL (+), aunit, ((-) | negate) #-}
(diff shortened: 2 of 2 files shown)


akrmn:
Perhaps we should add @lucianojoublanc-da's example from https://github.com/lucianojoublanc-da/additivebug as a negative test.

akrmn:
I added an empty `changelog_{begin,end}` block to the commit message to fix one of the build errors, though I'm not sure if this sort of change should be part of the log. I checked the repo history and it seems like similar changes haven't been added to the changelog, though.

sofiafaro-da (approved):
Looks good, thank you!

I agree, it would be nice ta have a test for this (in `compiler/damlc/tests/daml-test-files`).

cocreature (approved):
Thanks!

cocreature:
> I added an empty `changelog_{begin,end}` block to the commit message to fix one of the build errors, though I'm not sure if this sort of change should be part of the log. I checked the repo history and it seems like similar changes haven't been added to the changelog, though.

That’s exactly the right thing to do. We enforce a changelog to make sure people consider whether it’s changelog entry worthy and if the answer is that it isn’t then adding an empty changelog is the right course of action and we filter those out.

akrmn:
Just added a test case. As suggested on slack, I looked for other classes in the standard library which might need a `MINIMAL` pragma, but I didn't find any.

compiler/damlc/tests/daml-test-files/IncompleteAdditiveInstance.daml
```diff
@@ -0,0 +1,17 @@
+-- Copyright (c) 2020, Digital Asset (Switzerland) GmbH and/or its affiliates.
+-- All rights reserved.
+
+-- @ERROR range=14:10-14:23; • No explicit implementation for; either ‘-’ or ‘negate’; • In the instance declaration for ‘Additive Expr’
```
cocreature, line 4: the unicode might cause issues if it’s rendered differently on Windows (not sure if it is, maybe it’s fine). in that case I’d just start shorten it to `No explicit implementation for; either ‘-’ or ‘negate’`. Together with the range that still tests the right thing.

akrmn: ah good point, I fixed it now. Thanks!

akrmn: ah good point, but it seems to have gone through :)

cocreature: if it goes through even better :)

PR #11011, opened 2021-09-23
Make submission ID optional and generate it in `CommandTracker` if empty [KVL-1107]

hubert-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/CommandUpdaterFlow.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.daml.ledger.api.SubmissionIdGenerator
+import com.daml.ledger.api.domain.LedgerId
+import com.daml.ledger.api.v1.commands.Commands.DeduplicationPeriod
+import com.daml.ledger.client.configuration.CommandClientConfiguration
+import com.daml.util.Ctx
+import com.google.protobuf.duration.Duration
+
+import scala.annotation.nowarn
+
+object CommandUpdaterFlow {
```
hubert-da, line 17: Extracted from `CommandClient`.

andreaslochbihler-da (commented):
The `CommandTracker` tries to make sure that there is ever only a single submission in flight for a given change ID. I don't think that this can work, for the following reasons:
1. With replicated participants like in a VMBC setup, the same change ID may be sent to multiple participants and those participants each see just a single submission, but they get the completion events for both submissions. That should be taken care of by the `CommandService` of each participant picking a unique submission ID and checking in the completion that the submission ID is the expected one.
2. Commands may be sent concurrently via the `CommandService` and the `CommandSubmissionService`, and submission IDs are optional for Ledger API submissions. So if the `CommandSubmissionService` no longer assigns a submission ID, we may actually get back a completion without a submission ID from the submission to the `CommandSubmissionService` that is also processed by the `CommandService`. With the current logic of treating an empty submission ID in a completion as "matches all in-flight submissions", this re-introduces the very confusion that the submission IDs were supposed to eliminate. For example, if the submission via the `CommandSubmissionService` completes first with a rejection, then the `CommandService` will report a rejection even though the still-in-flight submission may end up being accepted.

To avoid the confusion, I'd suggest to introduce two modes to the `CommandTracker`. For legacy ledgers that do not propagate the submission ID, we keep the current behaviour of "empty submission ID matches all in-flight submissions", and for the upgraded ledgers it ignores completions without a submission ID (as it has ensured that the submission ID is set on all commands that it is tracking).

hubert-da:
@andreaslochbihler-da I'm a bit confused about the comment, as this PR reflects what we discussed is the right thing to do.

>So if the CommandSubmissionService no longer assigns a submission ID, we may actually get back a completion without a submission ID from the submission to the CommandSubmissionService that is also processed by the CommandService. With the current logic of treating an empty submission ID in a completion as "matches all in-flight submissions"

This command won't be present in the tracked commands, so it will try to "match all in-flight" submissions **with the same command id**. Do I understand correctly, that due to point 1. this also breaks?

andreaslochbihler-da:
> Do I understand correctly that due to point 1. this also breaks?

In my scenario, all submissions have the same change ID, the same command Id, same application ID and same set of actAs. So yes, this breaks, but independently of point 1.

miklos-da:
> The `CommandTracker` tries to make sure that there is ever only a single submission in flight for a given change ID. I don't think that this can work, for the following reasons:
> 
> 1. With replicated participants like in a VMBC setup, the same change ID may be sent to multiple participants and those participants each see just a single submission, but they get the completion events for both submissions. That should be taken care of by the `CommandService` of each participant picking a unique submission ID and checking in the completion that the submission ID is the expected one.
> 2. Commands may be sent concurrently via the `CommandService` and the `CommandSubmissionService`, and submission IDs are optional for Ledger API submissions. So if the `CommandSubmissionService` no longer assigns a submission ID, we may actually get back a completion without a submission ID from the submission to the `CommandSubmissionService` that is also processed by the `CommandService`. With the current logic of treating an empty submission ID in a completion as "matches all in-flight submissions", this re-introduces the very confusion that the submission IDs were supposed to eliminate. For example, if the submission via the `CommandSubmissionService` completes first with a rejection, then the `CommandService` will report a rejection even though the still-in-flight submission may end up being accepted.
> 
> To avoid the confusion, I'd suggest to introduce two modes to the `CommandTracker`. For legacy ledgers that do not propagate the submission ID, we keep the current behaviour of "empty submission ID matches all in-flight submissions", and for the upgraded ledgers it ignores completions without a submission ID (as it has ensured that the submission ID is set on all commands that it is tracking).

@andreaslochbihler-da What stops us from implementing only the latter, i.e., assuming that we always populate the submission ID?

andreaslochbihler-da:
> What stops us from implementing only the latter, i.e., assuming that we always populate the submission ID?

It's just separations of concerns, i.e., clear responsibilities in our architecture. The command submission service and everything below don't need the submission ID, so why should they bother populating it?

miklos-da:
> > What stops us from implementing only the latter, i.e., assuming that we always populate the submission ID?
> 
> It's just separations of concerns, i.e., clear responsibilities in our architecture. The command submission service and everything below don't need the submission ID, so why should they bother populating it?

Agreed. Assuming we'll shortly implement generating the submission ID (if missing) for the JSON-API, removing generation of a submission ID from the command submission service and ignoring completions without a submission ID should be fine, right?

andreaslochbihler-da:
I'm not sure I understand what the JSON API has to do with this. My point is: The `CommandTracker` implements functionality of tracking commands. It's used in the `CommandService`, in the `CommandClient`, in Canton, and possibly in the JSON API (I don't know about the latter). Submission IDs are meant to track commands. So as long as someone uses the `CommandTracker` with an old ledger that doesn't pass through the submission ID, we'll have a conflict of interest around the `CommandTracker` and where it's used:
* Preserve the old behaviour on old ledgers (completion with empty submission ID completes any tracking of the affected command ID)
* Support the submission IDs on new ledgers (ignore completions with empty submission ID)

So it's not about "when will this be in the JSON API", but about "when will the last user of the `CommandTracker` have moved off a ledger that doesn't propagate submission IDs". I don't know when that will happen, but given the longlivety of sandbox-classic, my hunch is that this will take a while.

miklos-da:
> I'm not sure I understand what the JSON API has to do with this. My point is: The `CommandTracker` implements functionality of tracking commands. It's used in the `CommandService`, in the `CommandClient`, in Canton, and possibly in the JSON API (I don't know about the latter). Submission IDs are meant to track commands. So as long as someone uses the `CommandTracker` with an old ledger that doesn't pass through the submission ID, we'll have a conflict of interest around the `CommandTracker` and where it's used:
> 
> * Preserve the old behaviour on old ledgers (completion with empty submission ID completes any tracking of the affected command ID)
> * Support the submission IDs on new ledgers (ignore completions with empty submission ID)
> 
> So it's not about "when will this be in the JSON API", but about "when will the last user of the `CommandTracker` have moved off a ledger that doesn't propagate submission IDs". I don't know when that will happen, but given the longlivety of sandbox-classic, my hunch is that this will take a while.

Wouldn't it be a good tradeoff then to implement propagation of submission IDs for sandbox-classic so that we don't have to support two modes (completion has/doesn't have submission ID)? @SamirTalwar-DA Have you looked into implementing submission ID propagation for sandbox-classic earlier?

SamirTalwar-DA:
I haven't, but I think it's a good idea to just get that done.

extractor/src/main/scala/com/digitalasset/extractor/Extractor.scala
```diff
@@ -259,7 +259,12 @@ class Extractor[T](config: ExtractorConfig, target: T)(
       LedgerClientConfiguration(
         applicationId = config.appId,
         ledgerIdRequirement = LedgerIdRequirement.none,
-        commandClient = CommandClientConfiguration(1, 1, java.time.Duration.ofSeconds(20L)),
+        commandClient = CommandClientConfiguration(
+          maxCommandsInFlight = 1,
+          maxParallelSubmissions = 1,
+          defaultDeduplicationTime = java.time.Duration.ofSeconds(20L),
+          submissionIdPropagationSupported = false,
```
hubert-da, line 266: Does anyone have an idea if I can do it?

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/configuration/CommandClientConfiguration.scala
```diff
@@ -12,13 +12,21 @@ import java.time.Duration
...
   *                                 a deduplication time set. The deduplication time is also used
   *                                 as the time after which commands time out in the command client.
+  * @param submissionIdPropagationSupported Tells whether submission IDs are being returned in Completions.
+  *                                         Should be false for integrations where the mutating schema is enabled.
   */
 final case class CommandClientConfiguration(
     maxCommandsInFlight: Int,
     maxParallelSubmissions: Int,
     defaultDeduplicationTime: Duration,
+    submissionIdPropagationSupported: Boolean,
```
hubert-da, line 22: I don't have access to the information about the schema on the client-side. I'm open to other ideas.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
```diff
@@ -199,7 +199,7 @@ private[sandbox] final class InMemoryLedger(
                 Some(commandId),
                 transactionId,
                 Some(`appId`),
-                _,
+                submissionId,
```
hubert-da, line 202: I added the submission ID propagation for the in-memory sandbox classic. For the SQL one, it's just about using the append-only schema. Am I missing anything?

hubert-da:
Sorry for making this PR rather huge. The scope was unknown beforehand. I also thought it will be easier to discuss if everything is in one place. Please leave a thumbs up if I should think about splitting it.

@miklos-da @SamirTalwar-DA @andreaslochbihler-da, WDYT about the latest changes? I'm not entirely happy about them, just don't have a better idea.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
```diff
@@ -323,34 +319,14 @@ private[commands] class CommandTracker[Context](
         }
 
         val trackedCommandKeys = pendingCommandKeys(maybeSubmissionId, commandId)
```
andreaslochbihler-da, line 321: Minor: If we're running this `CommandTracker` with `SubmissionIdPropagationMode.Supported`, then we'll do a full scan over the map of tracked commands to compute the `trackedCommandKeys` even though we know the result of that computation will not be used. That's a bit inefficient.

andreaslochbihler-da:
> WDYT about the latest changes?

I can't really comment on whether your choice of tying the submission ID propagation mode to whether the append-only schema is enabled will work. I have too little insight into what combinations of ledgers and ledger API servers are actually used. But if that's the right heuristics, then I think this looks fine.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/configuration/CommandClientConfiguration.scala
hubert-da: A possibly crazy question: Can we just change the mutating schema to use submission IDs easily? @digital-asset/kv-participant?

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
hubert-da: @digital-asset/kv-participant, maybe you have an idea?

SamirTalwar-DA:
Staring at this again, I don't think I'm a fan of making such a visible change. `CommandClientConfiguration` is supported API and I find it somewhat unreasonable to expect a user of our Java API to know whether their ledger supports submission ID propagation.

I would very much prefer if we could just make all ledgers work the same way.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/commands/tracker/CommandTracker.scala
hubert-da: I just realized that it's broken, as we also remove the tracked commands here.

hubert-da: Sorted out

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
miklos-da: > Sorry for making this PR rather huge. The scope was unknown beforehand. I also thought it will be easier to discuss if everything is in one place. Please let me know if I should think about splitting it.
> 
> @miklos-da @SamirTalwar-DA @andreaslochbihler-da, WDYT about the latest changes? I'm not entirely happy about them, just don't have a better idea.

If I understand correctly, after #10826 all tests (except one suite) will be running with the append-only schema and that schema will be supported from the SDK 1.18 release. @rautenrieth-da Can you confirm the latter, please?
The above implies that we should be able to remove the flag for whether submission ID is being supported (i.e., as we won't support the mutating schema anyway).

rautenrieth-da: The append-only schema is considered "production ready" in SDK 1.17, where it was described in the release notes. Hopefully, we'll be able to make it the only existing schema in SDK 1.18 (by deleting the code for the old mutating schema from our codebase).

In the meantime, we want to switch conformance tests to use the append-only schema by default. Some conformance tests will still run on the old (mutating) schema, but it's acceptable if not all optional tests are run on it. #10826 was my first attempt at modifying the conformance tests.

miklos-da: Thanks for the confirmation, @rautenrieth-da.
I'd then suggest to proceed as follows:
* Make submission ID propagation the only option (i.e., this allows removing the flags introduced for choosing between `SubmissionIdPropagation.Supported` vs. `SubmissionIdPropagation.NotSupported`).
* Disable tests (if needed) for the mutating schema.
* Keep refactoring of the `CommandUpdaterFlow`, etc.

extractor/src/main/scala/com/digitalasset/extractor/Extractor.scala
miklos-da: We won't need this change if we assume that submission ID propagation is supported by all ledgers.

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/configuration/CommandClientConfiguration.scala
miklos-da: As pointed out by @rautenrieth-da [here](https://github.com/digital-asset/daml/pull/11011#r719727805), the mutating schema is already deprecated so we can just disable the tests that depend on submission ID propagation.

extractor/src/main/scala/com/digitalasset/extractor/Extractor.scala
hubert-da: Correct.

Issue #10810, opened 2021-09-08
Daml Interfaces PoC


cocreature:
Will add details later for now I just need an issue number to reference in TODOs.

- [x] Protobuf
- [x] Haskell AST
- [x] Haskell typechecker
- [x] Support exercise on interface choice
- [x] Haskell name collision checker
- [x] Haskell protobuf decode & encode
- [x] Scala AST
- [x] Scala typechecker
- [x] Scala protobuf decode
- [x] Daml Parser
- [x] LF conversion
- [x] Speedy
- [x] Speedy compiler
- [x] Engine Preprocessor
- [x] Scala LF parser
- [x] Scala LF encoder
- [x] Add LF primitives for converting between interfaces & instances
- [x] Parse & desugar consuming declarations
- [x] Support interfaces in Daml Script’s `exerciseCmd` (and maybe more)
- [x] Support interfaces in Daml Triggers’s `exerciseCmd` (we probably get that for free via the previous point)
- [x] Consider how we represent interfaces in the typescript bindings
- [x] Consider implications of packages not being known upfront (at least for Daml Engine)
- [x] data-dependencies

cocreature:
I think this issue has served its purpose, closing in favor of more specific ones.

PR #11013, opened 2021-09-23
LF: Add interface support to the Preprocessor

remyhaemmerle-da:
This is part of #10810

We handle only the case of Exercise Command as Fetch Command is 
currenlty used only by Canton.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
@@ -36,17 +36,25 @@ private[lf] final class CommandPreprocessor(
     speedy.Command.Create(templateId, arg)
   }
 
-  @throws[Error.Preprocessing.Error]
   def unsafePreprocessExercise(
-      templateId: Ref.Identifier,
+      identifier: Ref.Identifier,
       contractId: Value.ContractId,
       choiceId: Ref.ChoiceName,
       argument: Value,
   ): speedy.Command.Exercise = {
     val cid = valueTranslator.unsafeTranslateCid(contractId)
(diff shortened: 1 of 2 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
```diff
@@ -119,6 +127,8 @@ private[lf] final class CommandPreprocessor(
           choiceArgument,
         )
       case command.FetchCommand(templateId, coid) =>
+        // TODO https://github.com/digital-asset/daml/issues/10810
```
cocreature, line 130: This only matters for Canton right?

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/PackageInterfaceSpec.scala
```diff
@@ -22,6 +22,9 @@ class PackageInterfaceSpec
   import com.daml.lf.testing.parser.Implicits._
   import com.daml.lf.transaction.test.TransactionBuilder.Implicits.{defaultPackageId => _, _}
 
+  // TODO https://github.com/digital-asset/daml/issues/10810
```
cocreature, line 25: yolo

cocreature (approved):
Thanks!

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/CommandPreprocessor.scala
remyhaemmerle-da: Yes.

PR #11017, opened 2021-09-24
Remove polymorphism from VersionedTransaction

nickchapman-da:
Remove `[Nid]` polymorphism from `VersionedTransaction` which was required precisely nowhere. Replace with concrete `NodeId`.

This change is consistent with the discriminated subtypes `SubmittedTransaction` and `CommittedTransaction` which are already not polymorphic.


remyhaemmerle-da (commented):
This is subsumed by #10921.
fyi, Moritz decided to wait for merging to avoid too limit conflicts in Canton code base.

PR #11019, opened 2021-09-24
Increase es disk size

garyverhaegen-da:
Disks are currently at 75% fullness, so seems like a good idea to bump a bit. #10857 should reduce our needs, too, so this should last us a while.

CHANGELOG_BEGIN
CHANGELOG_END


stefanobaghino-da (approved):

garyverhaegen-da:
Took a bit longer than I expected, but it's all done now.

PR #11031, opened 2021-09-27
Use nix-provided gRPC on Unix-like OSs

akrmn:
This fixes the da-ghcid command on macOS and Linux

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


stack-snapshot.yaml
```diff
@@ -14,9 +14,7 @@ packages:
   - ghc-lib-parser-ex-8.8.5.8
   - bytestring-nums-0.3.6
   - extra-1.7.1
-  - github: awakesecurity/gRPC-haskell
```
cocreature, line 17: we should revert this file. It doesn’t break anything in your PR because you didn’t repin (you could via `bazel run @stackage-unpinned//:pin`) as well.

WORKSPACE
```diff
@@ -894,6 +894,27 @@ load("@com_github_bazelbuild_buildtools//buildifier:deps.bzl", "buildifier_depen
...
+  visibility = ["//visibility:public"],
+  hdrs = [":include"],
+  includes = ["include"],
+)
+    """,
+    nix_file = "//nix:bazel.nix",
+    nix_file_deps = common_nix_file_deps,
+    # Remove once we upgrade to Bazel >=3.0. Until then `nix-build` output
+    # confuses the JAR query in `daml-sdk-head`.
+    quiet = True,
```
cocreature, line 914: we’re on bazel 4. something so you can remove this comment and `quiet=True`.

cocreature (approved):
Thanks!

WORKSPACE
akrmn: there are quite a few of these so I'll remove them in another commit

stack-snapshot.yaml
akrmn: aha! thanks for pointing this out, it's reverted now

PR #11033, opened 2021-09-27
DPP-496 HA indexer integration test

rautenrieth-da:
Additions in this PR:

1. `StorageBackend.verifyIntegrity()`, which (currently) verifies that all event sequential IDs are correct
2.  `EndlessReadService`, a `ReadService` implementation that serves a deterministic infinite stream of artificial state updates
2. `IndexerStabilityTestFixture`, a helper that can start `N` parallel running indexers from the above ReadService
3. `IndexerStabilitySpec`, a test which verifies that N concurrently running indexers correctly cooperate
4. `StorageBackendTestsDebug`, an extension of the `StorageBackend` suite that checks that the new `verifyIntegrity()` method actually catches errors

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
@@ -58,8 +58,8 @@ case class HaConfig(
     workerLockAcquireRetryMillis: Long = 500,
     workerLockAcquireMaxRetry: Long = 1000,
     mainLockCheckerPeriodMillis: Long = 1000,
-    indexerLockId: Int = 0x646d6c00, // note 0x646d6c equals ASCII encoded "dml"
-    indexerWorkerLockId: Int = 0x646d6c01,
+    indexerLockId: Int = 0x646d6c0, // note 0x646d6c equals ASCII encoded "dml"
+    indexerWorkerLockId: Int = 0x646d6c1,
     enable: Boolean = false, // TODO ha: remove as stable
 )
 
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
@@ -50,7 +50,7 @@ trait StorageBackend[DB_BATCH]
     with EventStorageBackend
     with DataSourceStorageBackend
     with DBLockStorageBackend
-    with DebugStorageBackend {
+    with IntegrityStorageBackend {
 
   /** Truncates all storage backend tables, EXCEPT the packages table.
     * Does not touch other tables, like the Flyway history table.
@@ -341,7 +341,7 @@ object DBLockStorageBackend {
   }
 }
 
-trait DebugStorageBackend {
+trait IntegrityStorageBackend {
 
   /** Verifies the integrity of the index database, throwing an exception if any issue is found.
     * This operation is allowed to take some time to finish.
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
@@ -49,7 +49,8 @@ trait StorageBackend[DB_BATCH]
     with ContractStorageBackend
     with EventStorageBackend
     with DataSourceStorageBackend
-    with DBLockStorageBackend {
+    with DBLockStorageBackend
+    with DebugStorageBackend {
 
   /** Truncates all storage backend tables, EXCEPT the packages table.
     * Does not touch other tables, like the Flyway history table.
@@ -338,6 +339,15 @@ object DBLockStorageBackend {
     case object Exclusive extends LockMode
     case object Shared extends LockMode
   }
+}
+
+trait DebugStorageBackend {
+
+  /** Verifies the integrity of the index database, throwing an exception if any issue is found.
+    * This operation is allowed to take some time to finish.
+    * It is not expected that it is used during regular index/indexer operation.
+    */
+  def verifyIntegrity()(connection: Connection): Unit
 }
 
 object StorageBackend {
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/DebugStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/DebugStorageBackendTemplate.scala
@@ -1,0 +1,76 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.backend.common
+
+import anorm.{RowParser, SQL}
+
+import java.sql.Connection
+import anorm.SqlParser.long
+import anorm.~
+import com.daml.platform.store.backend.DebugStorageBackend
+
+private[backend] trait DebugStorageBackendTemplate extends DebugStorageBackend {
+
+  private val allSequentialIds: String =
+    s"""
+      |SELECT event_sequential_id FROM participant_events_divulgence
+      |UNION ALL
+      |SELECT event_sequential_id FROM participant_events_create
+      |UNION ALL
+      |SELECT event_sequential_id FROM participant_events_consuming_exercise
+      |UNION ALL
+      |SELECT event_sequential_id FROM participant_events_non_consuming_exercise
+      |""".stripMargin
+
+  private val SQL_EVENT_SEQUENTIAL_IDS_SUMMARY = SQL(s"""
+      |WITH sequential_ids AS ($allSequentialIds)
+      |SELECT min(event_sequential_id) as min, max(event_sequential_id) as max, count(event_sequential_id) as count
+      |FROM sequential_ids, parameters
+      |WHERE event_sequential_id <= parameters.ledger_end_sequential_id
+      |""".stripMargin)
+
+  // Don't fetch an unbounded number of rows
+  private val maxReportedDuplicates = 100
+
+  private val SQL_DUPLICATE_EVENT_SEQUENTIAL_IDS = SQL(s"""
+       |WITH sequential_ids AS ($allSequentialIds)
+       |SELECT event_sequential_id as id, count(*) as count
+       |FROM sequential_ids, parameters
+       |WHERE event_sequential_id <= parameters.ledger_end_sequential_id
+       |GROUP BY event_sequential_id
+       |HAVING count(*) > 1
+       |FETCH NEXT $maxReportedDuplicates ROWS ONLY
+       |""".stripMargin)
+
+  case class EventSequentialIdsRow(min: Long, max: Long, count: Long)
+
+  private val eventSequantialIdsParser: RowParser[EventSequentialIdsRow] =
+    long("min") ~
+      long("max") ~
+      long("count") map { case min ~ max ~ count =>
+        EventSequentialIdsRow(min, max, count)
+      }
+
+  override def verifyIntegrity()(connection: Connection): Unit = {
+    val duplicates = SQL_DUPLICATE_EVENT_SEQUENTIAL_IDS
+      .as(long("id").*)(connection)
+    val summary = SQL_EVENT_SEQUENTIAL_IDS_SUMMARY
+      .as(eventSequantialIdsParser.single)(connection)
+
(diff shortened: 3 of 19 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
```diff
@@ -340,6 +341,15 @@ object DBLockStorageBackend {
   }
 }
 
+trait DebugStorageBackend {
```
nmarton-da, line 344: not sure about naming here...why debug? How about IntegrityStorageBackend?

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/EndlessReadService.scala
```diff
@@ -0,0 +1,222 @@
...
+    aborted = true
+    killSwitch.abort(cause)
+  }
+
+  def reset(): Unit = synchronized {
+    logger.info(s"EndlessReadService.reset() called")
+    stateUpdatesCalls.set(0)
+    initialConditionCalls.set(0)
+    aborted = false
+    killSwitch = KillSwitches.shared("EndlessReadService")
```
nmarton-da, line 128: sholudn't we need to shutdown the previous killswitch as well?

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/EndlessReadService.scala
```diff
@@ -0,0 +1,222 @@
...
+            optCompletionInfo = Some(completionInfo(i)),
+            transactionMeta = transactionMeta(i),
+            transaction = exerciseTransaction(i),
+            transactionId = transactionId(i),
+            recordTime = recordTime(i),
+            divulgedContracts = List.empty,
+            blindingInfo = None,
+          )
+      }
+      .via(killSwitch.flow)
```
nmarton-da, line 114: shouldn't this method need to be synchronized? it accesses a variable

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilityTestFixture.scala
```diff
@@ -0,0 +1,119 @@
...
+      indexers.filter(x => x.readService.stateUpdatesCalls.get() > 0 && !x.readService.aborted)
+    if (allRunning.length != 1) {
+      val message = s"Expected 1 running indexer, found ${allRunning.length}"
+      logger.info(message)
+      throw new RuntimeException(message)
+    } else {
+      allRunning.head
+    }
+  }
+  def resetAll(): Unit = indexers.foreach(x => x.readService.reset())
```
nmarton-da, line 43:
```suggestion
  def resetAll(): Unit = indexers.foreach(_.readService.reset())
```

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -0,0 +1,120 @@
...
+    info(s"Creating indexers fixture with $indexerCount indexers")
+    IndexerStabilityTestFixture
+      .owner(
+        updatesPerSecond,
+        indexerCount,
+        jdbcUrl,
+        materializer,
+      )
+      .use[Unit] { indexers =>
+        var abortedIndexer: Option[ReadServiceAndIndexer] = None
```
nmarton-da, line 51: minor/idiomatic scala: this does not necessarily have to be a var here, each iteration is passing this info along to the next one, a tailrec function, or a Iterator.iterate(None: Option[ReadServiceAndIndexer])(abortedIndexer => {...}).take(restartIterations) would do the trick

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -0,0 +1,120 @@
...
+          // The indexer should appear "healthy"
+          eventually {
+            assert(
+              activeIndexer.indexing.currentHealth() == HealthStatus.healthy,
+              "Running indexer should be healthy",
+            )
+          }
+          info(s"Indexer ${activeIndexer.readService.name} appears to be healthy")
+
+          // At this point, the indexer that was aborted by the previous iteration can be reset,
```
nmarton-da, line 61: shouldn't we wait here some random duration? without waits, as soon as we detected that the active indexer is healthy we pull the plug already

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -0,0 +1,120 @@
...
+        var abortedIndexer: Option[ReadServiceAndIndexer] = None
+        (1 to restartIterations).foreach(_ => {
+          // Assert that there is exactly one indexer running
+          val activeIndexer = eventuallyAfterRecovery { indexers.runningIndexer }
+          info(s"Indexer ${activeIndexer.readService.name} is running")
+
+          // The indexer should appear "healthy"
+          eventually {
+            assert(
+              activeIndexer.indexing.currentHealth() == HealthStatus.healthy,
```
nmarton-da: I am not sure about this check here (also not sure I understood the intention correctly)
An HA protected indexer will be healthy as waiting for the indexer lock.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -0,0 +1,120 @@
...
+        eventually {
+          indexers.indexers.foreach(_.indexing.currentHealth() == HealthStatus.unhealthy)
+        }
+        info(s"All ReadServices were aborted")
+
+        // Verify the integrity of the index database
+        val storageBackend = StorageBackend.of(DbType.jdbcType(jdbcUrl))
+        val dataSource = storageBackend.createDataSource(jdbcUrl)
+        val connection = dataSource.getConnection()
+        storageBackend.verifyIntegrity()(connection)
```
nmarton-da, line 87: besides verifying the integrity at the end, how about verifying that something actually got indexed for each iteration?

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
nmarton-da, line 1: how about adding a test case which proves that if no Ha protection, the test fails?
(for example with H2 we might be able to implement such, because that has no mutual exclusion on the indexer)
* it would be super nice to see that the integrity check fails
* also it would be nice to see that we have several indexers indexing

nmarton-da (commented):
Nice work, looks great!
Added several questions, and comments

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/DebugStorageBackendTemplate.scala
```diff
@@ -0,0 +1,76 @@
...
+package com.daml.platform.store.backend.common
+
+import anorm.{RowParser, SQL}
+
+import java.sql.Connection
+import anorm.SqlParser.long
+import anorm.~
+import com.daml.platform.store.backend.DebugStorageBackend
+
+private[backend] trait DebugStorageBackendTemplate extends DebugStorageBackend {
```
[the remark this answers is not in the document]
rautenrieth-da: I agree that the naming is strange. I have chosen the name because of consistency: all similar files are called `*StorageBackendTemplate`, even though half of them don't use any strategy or template methods. Let's discuss the file naming separately.

ledger/participant-integration-api/src/main/scala/platform/store/backend/StorageBackend.scala
rautenrieth-da: My intention was to create a trait that would collect all methods that are not to be used in production. I don't have strong opinion here, renamed to `IntegrityStorageBackend`.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/EndlessReadService.scala
rautenrieth-da: Good point. Added `synchronized` to all methods.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/EndlessReadService.scala
rautenrieth-da: The intention is that if you want to abort previously started update streams, you should explicitly call `abort`. `reset` is to make sure all future operations behave as if called on a fresh `EndlessReadService`.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
rautenrieth-da: It takes some time until the abort call is made, the indexer does happen to index a few state updates. I have added a wait to make this more robust.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
rautenrieth-da: Added a H2 test. I have refactored the suite so that it works with H2 as well, and changed the integrity checking part so that the H2 test asserts that the integrity check fails with duplicate event sequential ids.

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
```diff
@@ -96,21 +97,22 @@ trait IndexerStabilitySpec
...
         connection.close()
         Future.successful(())
       }
       .map(_ => succeed)
+  }.transform { result =>
+    if (haModeSupported) {
+      result
+    } else {
+      // If HA mode is not supported, the test must fail, but there are multiple reasons why it can fail.
+      // E.g., duplicate parameter table initialization, or duplicate event sequential ids.
```
nmarton-da, line 112: oh yeah, nice we are catching this one as well :)
but here we are after the duplicate seq id-s... other type of errors might shade some real problem  :thinking: 
how about starting the indexers after each other to prevent the occurrence of the duplicate param table initialization, and assert on the duplicate sequential id errors?

nmarton-da (approved):
solid work
* the error assertion needs some more care
* the test is currently failing on ci
otherwise looks good to me

ledger/participant-integration-api/src/test/lib/scala/platform/indexer/ha/IndexerStabilitySpec.scala
rautenrieth-da: That would be possible, but then we don't test that parallel initialization does work on Postgres/Oracle, which is IMO more important to test.

nmarton-da: can you think of a way to test both?

ledger/participant-integration-api/src/main/scala/platform/indexer/ha/HaCoordinator.scala
```diff
@@ -58,8 +58,8 @@ case class HaConfig(
     workerLockAcquireRetryMillis: Long = 500,
     workerLockAcquireMaxRetry: Long = 1000,
     mainLockCheckerPeriodMillis: Long = 1000,
-    indexerLockId: Int = 0x646d6c00, // note 0x646d6c equals ASCII encoded "dml"
-    indexerWorkerLockId: Int = 0x646d6c01,
+    indexerLockId: Int = 0x646d6c0, // note 0x646d6c equals ASCII encoded "dml"
```
rautenrieth-da, line 61: FYI @meiersi-da (AFAIK you had the original idea for the default lock IDs) - the previous lock ids are too long for Oracle.

meiersi-da: Ack. Thanks for catching this.

PR #11034, opened 2021-09-27
[JSON-API] Migrating tests to use sandbox next

### LINKED ISSUE #6037: port http-json tests to sandbox-next
Tests currently rely on sandbox-classic; they need to be ported to sandbox-next as a prerequisite to removing sandbox-classic. Moreover, given that http-json tests have caught at least one sandbox-classic bug (#5969), it would be useful to have that coverage against the new sandbox.

91d4590c906106e35ff20b675c399f3d77b09ef0 illustrates a trivial port, but this will not work; we can't spin up lots of sandbox-nexts the way we do with sandbox-classic. @gerolf-da suggests

> start sandbox once and run multiple tests against it with different parties

which will require some reworking of the http-json tests, to say the least. I assume that only one sandbox-next can exist in a JVM (classloader context) at a time, so whatever we do must keep that constraint in mind.

akshayshirahatti-da:
Fixes #6037 

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -53,11 +54,11 @@ import com.daml.ledger.api.{domain => LedgerApiDomain}
 import com.daml.ports.Port
 
 object AbstractHttpServiceIntegrationTestFuns {
-  private[http] val dar1 = requiredResource("docs/quickstart-model.dar")
+  private[http] val dar1 = requiredResource(ModelTestDar.path)
```
stefanobaghino-da, line 57: I'm not sure about this change either.

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
```diff
@@ -20,11 +20,14 @@ class HttpServiceWithPostgresIntTest
   override def wsConfig: Option[WebsocketConfig] = None
 
   "query persists all active contracts" in withHttpService { (uri, encoder, _, _) =>
+    val (party, headers) = partyAndAuthHeaders("Ian")
+    val searchDataSet = genSearchDataSet(party)
```
stefanobaghino-da, line 24: Where is `searchDataSet` used?

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -539,21 +547,39 @@ trait AbstractHttpServiceIntegrationTestFuns extends StrictLogging {
...
       payload,
-      headersWithAuth,
+      headers,
     )
   }
 
   protected def initialAccountCreate(
       serviceUri: Uri,
       encoder: DomainJsonEncoder,
+      owner: String = "Alice",
```
stefanobaghino-da, line 578: I would advice strongly against leaving `"Alice"` as a default. If we keep re-using the same ledger I'm fairly sure this would later cause some weird error and we can prevent some wasted time by getting rid of the default.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -651,34 +680,38 @@ abstract class AbstractHttpServiceIntegrationTest
   "multi-party query GET" in withHttpService { (uri, encoder, _, _) =>
     for {
       _ <- postCreateCommand(
-        accountCreateCommand(owner = domain.Party("Alice"), number = "42"),
+        accountCreateCommand(owner = domain.Party("Carol"), number = "42"),
```
stefanobaghino-da, line 683: These are very common names that could lend themselves to conflicts. If we want to re-use the same ledger over and over again we should probably also create a mechanism to create party names on a per-test basis (e.g. adding a prefix or a suffix to each party name) to avoid naming conflicts that could cause some unnecessary head scratch in the process of adding further tests in the future.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -1049,14 +1098,15 @@ abstract class AbstractHttpServiceIntegrationTest
 
   "exercise IOU_Transfer with unknown contractId should return proper error" in withHttpService {
     (uri, encoder, _, _) =>
-      val contractId = lar.ContractId("#NonExistentContractId")
+      val contractIdString = List.fill(66)("0").mkString("")
```
stefanobaghino-da, line 1101:
```suggestion
      val contractIdString = "0" * 66
```
But I'm not sure of what is the purpose of this change, the existing name seems quite telling.

stefanobaghino-da (commented):
1. at this point it's too late to change this, but in the future I would recommend solving one issue at a time and making one PR to transition to `sandbox` and one to switch to re-using the ledger across tests
2. I'm a bit worried about the long-term maintainability of manually giving names to each party in each test -- ideally the test infrastructure should abstract the naming concern away

ledger-service/http-json-testing/src/main/scala/com/daml/http/HttpServiceTestFixture.scala
```diff
@@ -271,8 +275,8 @@ object HttpServiceTestFixture extends LazyLogging with Assertions with Inside {
     }
   }
 
-  private val serverTlsConfig = TlsConfiguration(enabled = true, serverCrt, serverPem, caCrt)
-  private val clientTlsConfig = TlsConfiguration(enabled = true, clientCrt, clientPem, caCrt)
+  final val serverTlsConfig = TlsConfiguration(enabled = true, serverCrt, serverPem, caCrt)
+  final val clientTlsConfig = TlsConfiguration(enabled = true, clientCrt, clientPem, caCrt)
```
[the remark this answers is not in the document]
akshayshirahatti-da: because they get used in the `SandboxNextTestLedger` class which now is responsible for managing the lifecycle of the ledger of tests

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
akshayshirahatti-da: The context for this change is that the `"docs/quickstart-model.dal"` defines a template with module and entity name `Iou` similar to one in `ModelTestDar` and due to that fact that we were keeping the ledger around this was causing the `Template(None, Iou, Iou)` to drop out of the unique map on `TemplateIdMap(below)` 

```scala
  case class TemplateIdMap(
      all: Set[TemplateId.RequiredPkg],
      unique: Map[TemplateId.NoPkg, TemplateId.RequiredPkg],
  )
```  

and thus resulting in a test failures on not being able to resolve template

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
```diff
@@ -53,11 +54,11 @@ import com.daml.ledger.api.{domain => LedgerApiDomain}
 import com.daml.ports.Port
 
 object AbstractHttpServiceIntegrationTestFuns {
-  private[http] val dar1 = requiredResource("docs/quickstart-model.dar")
+  private[http] val dar1 = requiredResource(ModelTestDar.path)
 
   private[http] val dar2 = requiredResource("ledger-service/http-json/Account.dar")
 
-  private[http] val dar3 = requiredResource(ModelTestDar.path)
+  private[http] val dar3 = requiredResource(SemanticTestDar.path)
```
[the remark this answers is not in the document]
akshayshirahatti-da: based on the change for `dar1` this was just to ensure that the test which checks that new packages get loaded continues to function as expected.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
akshayshirahatti-da: Well I did go through the same pain , but it was partly due to the test passing using the jwt headers with `Alice` as default . I am happy to change this, can be done in a followup PR as I had left it out to reduce the footprint of the changes on this PR.

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
akshayshirahatti-da: I changed it as the test started failing complaining about `V0 contract ids being forbidden` on migrating away from ` sandbox-classic`

```scala

"CommandService Error, io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Invalid argument: Command interpretation error in LF-DAMLe: Illegal Contract ID "#NonExistentContractId", V0 Contract IDs are forbidden. Details: N/A." did not include substring "Contract could not be found with id ContractId(#NonExistentContractId)"

```

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -192,35 +192,37 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
     } yield inside(clientMsg) { case warning +: result +: heartbeats =>
       warning should include("\"warnings\":{\"unknownTemplateIds\":[\"Unk")
-      result should include("\"issuer\":\"Alice\"")
+      result should include(s""""issuer":"$party"""")
       Inspectors.forAll(heartbeats)(assertHeartbeat)
     }
   }
 
   "fetch endpoint should warn on unknown template IDs" in withHttpService { (uri, encoder, _, _) =>
+    val (party, headers) = partyAndAuthHeaders("Kevin")
```
S11001001, line 213:
```suggestion
    val (party, headers) = partyAndAuthHeaders("Kevin")
    val partyJson = party.toJson
```

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -192,35 +192,37 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
     for {
-      _ <- initialAccountCreate(uri, encoder)
+      _ <- initialAccountCreate(uri, encoder, party, headers)
 
       clientMsg <- singleClientFetchStream(
-        jwt,
+        jwtForParties(List(party), List(), testId),
         uri,
-        """[{"templateId": "Account:Account", "key": ["Alice", "abc123"]}, {"templateId": "Unknown:Template", "key": ["Alice", "abc123"]}]""",
+        s"""[{"templateId": "Account:Account", "key": ["$party", "abc123"]}, {"templateId": "Unknown:Template", "key": ["$party", "abc123"]}]""",
```
S11001001, line 220:
```suggestion
        s"""[{"templateId": "Account:Account", "key": [$partyJson, "abc123"]}, {"templateId": "Unknown:Template", "key": [$partyJson, "abc123"]}]""",
```

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -192,35 +192,37 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
       ).take(3)
         .runWith(collectResultsAsTextMessage)
     } yield inside(clientMsg) { case warning +: result +: heartbeats =>
       warning should include("""{"warnings":{"unknownTemplateIds":["Unk""")
-      result should include(""""owner":"Alice"""")
+      result should include(s""""owner":"$party"""")
```
S11001001, line 225:
```suggestion
      result should include(s""""owner":$partyJson""")
```

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -620,12 +654,12 @@ abstract class AbstractWebsocketServiceIntegrationTest
       )
     }
     val req =
-      """
-          |[{"templateId": "Account:Account", "key": ["Alice", "abc123"]},
-          | {"templateId": "Account:Account", "key": ["Alice", "def456"]}]
+      s"""
+          |[{"templateId": "Account:Account", "key": ["$party", "abc123"]},
+          | {"templateId": "Account:Account", "key": ["$party", "def456"]}]
```
S11001001, line 659: Likewise with `partyJson`.

ledger-service/http-json/src/itlib/scala/http/AbstractWebsocketServiceIntegrationTest.scala
```diff
@@ -947,23 +988,25 @@ abstract class AbstractWebsocketServiceIntegrationTest
...
       s"""{
            |  "templateId": "Iou:Iou",
            |  "payload": {
            |    "observers": [],
-           |    "issuer": "Alice",
+           |    "issuer": "$party",
            |    "amount": "999.99",
            |    "currency": "$currency",
-           |    "owner": "Alice"
+           |    "owner": "$party"
```
S11001001, line 1001: Likewise with `partyJson`.

S11001001 (approved):

ledger-service/http-json/src/itlib/scala/http/AbstractHttpServiceIntegrationTest.scala
akshayshirahatti-da: That's a valid concern, I'll think about a way to do this but at the same time I am open to any easy solutions/suggestions in this regard.

ledger-service/http-json/src/it/scala/http/HttpServiceWithPostgresIntTest.scala
akshayshirahatti-da: apologies this is a confusing change but its used on line `26` below

akshayshirahatti-da:
> at this point it's too late to change this, but in the future I would recommend solving one issue at a time and making one P
R to transition to sandbox and one to switch to re-using the ledger across tests

A simple migration to use `sandbox` and re-create the ledger for each test led to loads of timeouts i.e (`The server was not able to produce a timely response to your request.`) hence I had to make the big bang change. 

> I'm a bit worried about the long-term maintainability of manually giving names to each party in each test -- ideally the test infrastructure should abstract the naming concern away

Agreed, I'd think at the minimum a utility to generate random party names to be used on the test would be useful and I'd look at reworking the PR in that regard, as mentioned before open for more suggestions or something simpler.

stefanobaghino-da:
> A simple migration to use `sandbox` and re-create the ledger for each test led to loads of timeouts i.e (`The server was not able to produce a timely response to your request.`) hence I had to make the big bang change.

I see, thanks.

Issue #10772, opened 2021-09-03
Support module reexports in data-dependencies


cocreature:
Currently data-dependencies do not support module reexports meaning that if you have a module A that reexports module B and you import it via data-dependencies A is not going to reexport B.

daml-lf does not have a concept of exports (or limiting exports) and the goal here is not to try to add that. but we should provide some limited support for handling reexports.

For now we should only handle full-module reexports (but we might want to extend this later).
For implementing that we need to persist those reexports in daml-lf in some form and then reconstruct them when we have a data-dependency.

PR #11037, opened 2021-09-27
Data dependencies deserialize imports

akrmn:
This reads the value of the `$$imports` value in the previously generated DAML-LF and produces the corresponding imports with `emitModRef`

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
@@ -282,10 +282,21 @@ generateSrcFromLf env = noLoc mod
     exports :: [LIE GhcPs]
     genState :: GenState
     ((exports, decls), genState) = runGen
-        ((,) <$> genExports <*> genDecls)
+        ((,) <$> genExports <*> genDecls <* genOrphanDepImports)
 
     modRefs :: Set ModRef
     modRefs = gsModRefs genState
+
+    genOrphanDepImports :: Gen ()
+    genOrphanDepImports = sequence_ $ do
+        Just LF.DefValue{dvalBinder=(_, ty)} <- [NM.lookup LFC.moduleImportsName . LF.moduleValues $ envMod env]
+        Just quals <- [LFC.decodeModuleImports ty]
+        LF.Qualified { LF.qualModule, LF.qualPackage } <- Set.toList quals
+        pure $ emitModRef ModRef
+            { modRefModule = qualModule
+            , modRefOrigin = importOriginFromPackageRef (envConfig env) qualPackage
+            , modRefImpSpec = EmptyImpSpec
+            }
 
     genExports :: Gen [LIE GhcPs]
     genExports = sequence $ selfReexport : classReexports
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
@@ -177,7 +177,19 @@ data ImportOrigin = FromCurrentSdk UnitId | FromPackage LF.PackageId
 data ModRef = ModRef
     { modRefModule :: LF.ModuleName
     , modRefOrigin :: ImportOrigin
+    , modRefImpSpec :: ModRefImpSpec
     } deriving (Eq, Ord)
+
+data ModRefImpSpec
+    = NoImpSpec
+        -- ^ For open imports, e.g.
+        --
+        -- > import SomeModule
+    | EmptyImpSpec
+        -- ^ For instances-only imports, e.g.
+        --
+        -- > import SomeModule ()
+    deriving (Eq, Ord)
 
 modRefImport :: Config -> ModRef -> LImportDecl GhcPs
 modRefImport Config{..} ModRef{..} = noLoc ImportDecl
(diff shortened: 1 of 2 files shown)


compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
```diff
@@ -199,6 +211,9 @@ modRefImport Config{..} ModRef{..} = noLoc ImportDecl
              | importPkgId == configSelfPkgId -> modRefModule
              -- The module names from the current package are the only ones that are not modified
              | otherwise -> prefixDependencyModule importPkgId modRefModule
+      impSpec = case modRefImpSpec of
+          NoImpSpec -> Nothing
+          EmptyImpSpec -> Just (False, noLoc []) -- False = not 'hiding'
```
cocreature, line 216: I love the GHC API …

compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
```diff
@@ -267,11 +282,22 @@ generateSrcFromLf env = noLoc mod
...
 
     modRefs :: Set ModRef
     modRefs = gsModRefs genState
 
+    genOrphanDepImports :: Gen ()
+    genOrphanDepImports = sequence_ $ do
+        Just LF.DefValue{dvalBinder=(_, ty)} <- [NM.lookup LFC.moduleImportsName . LF.moduleValues $ envMod env]
+        Just quals <- [LFC.decodeModuleImports ty]
+        LF.Qualified { LF.qualModule, LF.qualPackage } <- Set.toList quals
+        pure $ emitModRef ModRef
```
cocreature, line 295: What happens if we have both a reference to a type in a module as well as an orphan instance? It looks like we want a semigroup instance on `ModRefImpSpec` where `NoImpSpec` takes precedence and then change `emitModRef` to make sure that it overwrites.

cocreature (commented):
Nice! I would suggest to merge this after #11038 and add an integration test to https://github.com/digital-asset/daml/blob/main/compiler/damlc/tests/src/DA/Test/DataDependencies.hs in this PR.

compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
akrmn: > Types are amazing! 
> -- GHC devs

> `Bool` 
> -- also GHC devs

compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
akrmn: Hm, I think the only way to test that would involve solving #10772 first, since otherwise there's no way (that I can think of) for `"new-package" Main.daml` to import the instance and a type defined in module `"old-package" X.daml` without importing `X.daml` directly.

akrmn: in any case, I think this would only trigger a redundant import warning, but I'd have to check

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -1205,6 +1205,37 @@ tests Tools{damlc,repl,validate,davlDar,oldProjDar} = testGroup "Data Dependenci
...
+            , "import Type"
+            , "instance Show T where show T = \"T\""
+            ]
+        ,   (,) "Wrapper.daml"
+            [ "module Wrapper where"
+            , "import OrphanInstance ()"
+            ]
+        ]
+        [
+            (,) "Main.daml"
```
akrmn, line 1228: I cherry-picked this test onto `main` and verified that it does fail there :)

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
```diff
@@ -1236,6 +1236,37 @@ tests Tools{damlc,repl,validate,davlDar,oldProjDar} = testGroup "Data Dependenci
...
+            ]
+        ,   (,) "OrphanInstanceAndType.daml"
+            [ "module OrphanInstanceAndType (T(..), S(..)) where"
+            , "import Type (T(..))"
+            , "instance Show T where show T = \"T\""
+            , "data S = S"
+            , "  deriving (Show)"
+            ]
+        ,   (,) "Wrapper.daml"
+            [ "module Wrapper (T(..), S(..)) where"
```
akrmn, line 1256: @cocreature this is my attempt at testing the scenario you mentioned, namely, `Main` indirectly importing both an orphan instance and a datatype from the same module. However, this fails with "`T` and `S` not in scope" since data-dependencies doesn't support reexports yet #10772 (that issue is for _module_ reexports, maybe we need a new issue?)

cocreature: good point, then let’s just drop this test for now from this PR.

cocreature (approved):
Nice work, thanks! I suggest merging the other one first and then retargetting this at `main`.

compiler/damlc/tests/src/DA/Test/DataDependencies.hs
akrmn: cool, will do!

PR #11038, opened 2021-09-27
Data dependencies serialize imports

akrmn:
This generates an `$$imports` value in DAML-LF with the imports needed for orphan instances when the module is included via `data-dependencies`

### Pull Request Checklist

trigger the build.


compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) 2021, Digital Asset (Switzerland) GmbH and/or its affiliates.
+-- All rights reserved.
+
+-- This test checks that we produce an $$imports value including all modules from which this one imports instances.
+
+-- @ERROR range=14:10-14:23; • No explicit implementation for; either ‘-’ or ‘negate’; • In the instance declaration for ‘Additive Expr’
+-- @QUERY-LF .modules[]
```
akrmn, line 7: I'm not sure what's the best way to produce the query I want here. Basically I want to check that `ModuleImports_OrphanInstance` is in the `$$imports` value.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1868,14 +1878,18 @@ rewriteStableQualified env q@(Qualified pkgRef modName obj) =
       Nothing -> q
       Just pkgId -> Qualified (PRImport pkgId) modName obj
 
-convertQualified :: NamedThing a => (T.Text -> t) -> Env -> a -> ConvertM (Qualified t)
-convertQualified toT env x = do
-  pkgRef <- nameToPkgRef env x
-  let modName = convertModuleName $ GHC.moduleName $ nameModule $ getName x
+convertGhcModule :: a -> Env -> GHC.Module -> ConvertM (Qualified a)
```
cocreature, line 1881:
```suggestion
convertModuleName :: a -> Env -> GHC.Module -> ConvertM (Qualified a)
```
Everything in this module converts from GHC to LF so adding `Ghc` doesn’t really add anything.

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
```diff
@@ -0,0 +1,11 @@
+-- Copyright (c) 2021, Digital Asset (Switzerland) GmbH and/or its affiliates.
+-- All rights reserved.
+
+-- This test checks that we produce an $$imports value including all modules from which this one imports instances.
+
+-- @ERROR range=14:10-14:23; • No explicit implementation for; either ‘-’ or ‘negate’; • In the instance declaration for ‘Additive Expr’
```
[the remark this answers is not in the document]
cocreature: I’m confused, can we delete this?

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
cocreature: The queries are produced by JQ. To get the JSON you can change https://github.com/digital-asset/daml/blob/6bf45a344a07568f1c2564ba894860dd9b276129/compiler/damlc/tests/src/DA/Test/DamlcIntegration.hs#L171 to an absolute directory (otherwise it ends up in the bazel sandbox and is annoying to find). That should allow you to run a reasonable query on it it. As an example, maybe look at the one for fundeps https://github.com/digital-asset/daml/blob/6bf45a344a07568f1c2564ba894860dd9b276129/compiler/damlc/tests/daml-test-files/FunctionalDependencies.daml#L8.

If you can’t find the right query, I’m happy to help.

cocreature (commented):
Looks great, minor nits and we should fix the test before merging.

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
akrmn: Thanks! I'll start with this :)

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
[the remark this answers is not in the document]
akrmn: yup, the last commit was just the in-progress test I had, and I only saw this line (copied over form another test) on github

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -686,6 +688,14 @@ convertClassDef env tycon
         -- NOTE (SF): No reason to generate fundep & minimal metadata with old-style typeclasses,
         -- since data-dependencies support for old-style typeclasses is extremely limited.
 
+convertDepOrphanModules :: Env -> [GHC.Module] -> ConvertM [Definition]
+convertDepOrphanModules env availNames = do
```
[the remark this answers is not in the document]
akrmn: ah good catch, thanks

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
akrmn: ah I thought of that too but this module already has another `convertModuleName` function

```haskell
convertModuleName :: GHC.ModuleName -> LF.ModuleName
```

Also, mine converts a `Module`, which is the `ModuleName` plus the `UnitId`. But of course `convertModule` is taken and means something very different.

We could use a different naming scheme, what do you think about `qualifiedFromModule`?

#NamingThingsIsHard.

cocreature: ugh, too many module things. Maybe
1. ` convertModuleName` -- keep as is
2. `convertQualifiedModuleName`  -- what you called `convertGhcModule`?
3. `convertModule` -- keep as is

Alternatively we could also rename `convertModule` to `convertCoreModule` but not sure that helps

compiler/damlc/tests/src/query-lf-interned.jq
```diff
@@ -19,3 +19,28 @@ def get_name(pkg): .name_interned_str | resolve_interned_string(pkg);
 def get_text(pkg): .text_interned_str | resolve_interned_string(pkg);
 
 def norm_ty(pkg): if has("interned") then pkg.interned_types[.interned] else . end;
+
+def norm_imports(pkg):
```
akrmn, line 23: this got a bit out of hand. Please let me know if you know a better way to do this! @cocreature @sofiafaro-da

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
```diff
@@ -0,0 +1,9 @@
+-- Copyright (c) 2021, Digital Asset (Switzerland) GmbH and/or its affiliates.
+-- All rights reserved.
+
+-- This test checks that we produce an $$imports value including all modules from which this one imports instances.
+-- @QUERY-LF .modules[] | select(lf::get_dotted_name($pkg) == ["ModuleImports_Main"]) | .values[].name_with_type | select(lf::get_value_name($pkg) == ["$$$$imports"]) | lf::norm_imports($pkg) | any(. == {package: null, module: ["ModuleImports_OrphanInstance"]})
```
akrmn, line 5: Also this one, I don't think there's a way to split these into multiple lines, right?

akrmn: btw @cocreature, I did need the `select` on `ModuleImports_Main`, since otherwise I got all three `$$imports` definitions.

cocreature: oh right I was talking BS. sorry for the confusion.

And yes there is no way to split it into different lines atm.

compiler/damlc/tests/src/query-lf-interned.jq
cocreature: doesn’t look too bad

cocreature (approved):
Thanks!

compiler/damlc/tests/daml-test-files/ModuleImports_Main.daml
akrmn: no problem, just thought you'd like to know for the future :) 

Got it, guess single line works for now, it's just a bit hard to read

compiler/damlc/tests/src/query-lf-interned.jq
akrmn: thanks!

sofiafaro-da: looks good :)

akrmn: thanks!

akrmn:
not sure what made those tests fail, so I rebased on `main` to see if it helps

cocreature:
I took a look at the failure in the damlc integration test:
It’s failing in `InternedTypes`. The issue is that the `$$imports` declaration also has a type ofc so that gets interned. Probably easiest to just modify the test to the new expected output (meaning more interned types).

akrmn:
oh! that makes sense, thanks for looking into it @cocreature

akrmn:
ah, the `InternedTypes` test also runs on older versions of the compiler. I'll add a check on the presence of `$$imports` and then do either the old or new check (5 interned types vs 20)

cocreature:
> ah, the `InternedTypes` test also runs on older versions of the compiler. I'll add a check on the presence of `$$imports` and then do either the old or new check (5 interned types vs 20)

I don’t think I understand that. It runs for old LF versions but only LF >= 1.11 (that’s what `SINCE-LF 1.11`) does.
As far as I can tell your code should apply regardless of the LF version.
Can you describe the error you’re seeing?

akrmn:
These are the errors I got on CI: 
```
//compiler/damlc/tests:integration-v111                                  FAILED in 169.3s
  /home/vsts/.cache/bazel/_bazel_vsts/704fb9a8c2b8a7269239cfce241f68b9/execroot/com_github_digital_asset_daml/bazel-out/k8-opt/testlogs/compiler/damlc/tests/integration-v111/test.log
//compiler/damlc/tests:integration-v112                                  FAILED in 152.4s
  /home/vsts/.cache/bazel/_bazel_vsts/704fb9a8c2b8a7269239cfce241f68b9/execroot/com_github_digital_asset_daml/bazel-out/k8-opt/testlogs/compiler/damlc/tests/integration-v112/test.log
//compiler/damlc/tests:integration-v113                                  FAILED in 165.7s
  /home/vsts/.cache/bazel/_bazel_vsts/704fb9a8c2b8a7269239cfce241f68b9/execroot/com_github_digital_asset_daml/bazel-out/k8-opt/testlogs/compiler/damlc/tests/integration-v113/test.log
//compiler/damlc/tests:integration-v16                                   FAILED in 131.2s
  /home/vsts/.cache/bazel/_bazel_vsts/704fb9a8c2b8a7269239cfce241f68b9/execroot/com_github_digital_asset_daml/bazel-out/k8-opt/testlogs/compiler/damlc/tests/integration-v16/test.log
//compiler/damlc/tests:repl                                              FAILED in 293.5s
```
(https://dev.azure.com/digitalasset/daml/_build/results?buildId=90136&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=7b720ed4-ea9c-5605-45fe-aca5cec61df4&l=9989)
I then ran `//compiler/damlc/tests:integration-v111` on my machine (since I'm not sure how to fetch the log files from azure), and it failed on the same `InternedTypes` test, since that version didn't produce the `$$imports` value, but with my change the test always expects 20 interned types

akrmn:
oh! the problem is _not_ that older versions don't have an `$$imports` value, they do! _but_ the set of orphan modules in scope is different! in particular, v111 lacks `DA.Internal.Exception`

cocreature:
Oh that makes a lot more sense. exceptions only got added in LF 1.14. You could split it into two tests and limit it with `@since-LF` and `until-lf` maybe.

akrmn:
I guess the easiest change is to weaken the length check to `>= 4`. The important part of this test is that the types are shared for `ap1` and `ap2`, and the second `QUERY-LF` takes care of that

PR #11039, opened 2021-09-27
Draft of Ledger API feature descriptors

meiersi-da:
Intended to aid the discussion about how to provide a streamlined conformance testing approach using the `ledger-api-test-tool` that supports Ledger API evolution, EE-only features, and leder-specific implementation choices for certain specific API features.


ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -26,6 +28,106 @@ message GetLedgerApiVersionRequest {
...
+  // for ledger implementation testing purposes only.
+  //
+  // Daml applications SHOULD not depend on these in production.
+  ExperimentalFeatures experimental = 3;
+}
+
+// See the feature message definitions for descriptions.
+message LedgerFeatures {
+  FeatureOrderedContractIds ordered_contract_ids = 1;
+  FeatureMonotonicRecordTime monotonic_record_time = 2;
```
SamirTalwar-DA, line 68: Perhaps we can explicitly mark each of these as `optional` (and below)?

meiersi-da: Good idea. Will do.

meiersi-da: Actually, it does just add noise doesn't it? `optional` is the default in proto3, and we generally do not mark fields as optional do we?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -26,6 +28,74 @@ message GetLedgerApiVersionRequest {
...
+  // required to run the application.
+  //
+  // See the feature descriptions themselves for the relation between
+  // Ledger API versions and feature presence.
+  FeatureDescriptor features = 2;
+}
+
+message FeatureDescriptor {
+  // Features that are determined by the ledger backing the participant node
+  // that serves the Ledger API endpoint.
```
andreaslochbihler-da, line 48: In DA marketing speech, the ledger is the Daml driver, isn't it?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -26,6 +28,74 @@ message GetLedgerApiVersionRequest {
...
+// to reclaim storage space.  
+//
+// Optional since: v1.10
+message FeatureParticipantPruning {
+
+  // Whether pruning can remove all divulgence events associated to the 
+  // historic transaction log.
+  //
+  // Optional since: v.12
+  bool supports_pruning_all_divulged_events = 2;
```
andreaslochbihler-da, line 100:
```suggestion
  bool supports_pruning_all_divulged_events = 1;
```

ledger-api/grpc-definitions/com/daml/ledger/api/v1/expected_ledger_feature_descriptors.yaml
```diff
@@ -0,0 +1,58 @@
+# Example feature descriptor for our existing ledgers.
+# Written as pseudo YAML using its mapping to JSON and then to `FeatureDescriptor` proto message
```
andreaslochbihler-da, line 2: I'm struggling to understand the translation. I see items without values like `ordered_contract_ids:` and items with values like `supports_pruning_all_divulged_events: False`. What is the difference between those?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -26,6 +28,74 @@ message GetLedgerApiVersionRequest {
...
+// See the feature message definitions for descriptions.
+message LedgerFeatures {
+  FeatureOrderedContractIds ordered_contract_ids = 2;
 }
+
+// See the feature message definitions for descriptions.
+message ParticipantFeatures {
+  FeatureDamlLf daml_lf = 1;
+  FeatureParticipantPruning participant_pruning = 2;
+}
```
andreaslochbihler-da, line 74: Add the FeatureCausalOrdering for whether the participant outputs transactions in causal order.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/expected_ledger_feature_descriptors.yaml
```diff
@@ -0,0 +1,58 @@
...
+      implicit_party_allocation: True 
+    participant_pruning:
+      supports_pruning_all_divulged_events: False
+    ledger_config_management:
+  experimental:
+    rank_based_command_deduplication:
+      completions_include_submission_ids: True
+
+sandbox-next:
+  invocation_matrix:
```
andreaslochbihler-da, line 22: What is the intention behind this invocation matrix? It doesn't show up in the Protobuf definitions and as plain text strings I don't see how to automatically generate invocations for this.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/expected_ledger_feature_descriptors.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - "with and without --enable-append-only-schema"
+    - "wall-clock-time vs. static time"
```
andreaslochbihler-da, line 24: I think static time support should be a ledger feature.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/expected_ledger_feature_descriptors.yaml
```diff
@@ -0,0 +1,58 @@
...
+    ledger_config_management:
+  experimental:
+    rank_based_command_deduplication:
+      completions_include_submission_ids: True
+
+
+canton:
+  invocation_matrix:
+    - "uck vs. non-uck"
+    # TODO: there are likely more.
```
andreaslochbihler-da, line 46: For example:
* Postgres vs. H2 vs. Oracle
* enterprise vs. community (although this influences the available features, so maybe better to have to separate versions)
* HA on/off

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
andreaslochbihler-da: We normally mark optional fields because our convention is that all fields are required.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
meiersi-da: So KV ledgers output transactions in causal order, while Canton can have use-after-archive cases can't it? However, it can't have 'use-before-create' and 'archive-before-create'. So it's not that there is no causal order, but a weaker causal order isn't it?

I'm wondering whether there's a better name like `FeatureParticipantStrongTransactionOrder` that does not make a client equally nervous if it is not present.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
```diff
@@ -26,6 +28,75 @@ message GetLedgerApiVersionRequest {
...
+  // for ledger implementation testing purposes only.
+  //
+  // Daml applications SHOULD not depend on these in production.
+  ExperimentalFeatures experimental = 3;
+}
+
+// See the feature message definitions for descriptions.
+message LedgerFeatures {
+  FeatureOrderedContractIds ordered_contract_ids = 2;
 }
```
meiersi-da, line 68: As pointed out by @andreaslochbihler-da : consider making `static_time` a ledger feature.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/version_service.proto
andreaslochbihler-da: Canton in single-domain mode always achieves causal ordering. In a multi-domain setting, Canton currently doesn't enforce causality, so you can observe both archive-before-create and use-before-create. @phoebe-da is about to add causality enforcement to Canton, but for the time being, we consider this a configuration option because enforcing causality may create liveness problems.

meiersi-da:
Closing as no further design work on this change is expected. I do though expect that we'll make use of the ideas in here, as @tudor-da did some time ago in #11328 .

PR #11041, opened 2021-09-27
remove search index on json fields that harms insert performance

dasormeter:
On latest master
Running 
```
bazel test
//ledger/participant-integration-api:participant-integration-api-tests-oracle_test_suite_src_test_suite_scala_platform_store_dao_JdbcLedgerDaoOracleAppendOnlySpec.scala
--config=oracle
--test_output=streamed
--cache_test_results=no
--test_arg=-z
--test_arg="fall back to limit-based"
```

with json fields search indexes in place
•total time : 3min30s

without (for that slowest fall back to limit-based test that runs 10x)
•1min30s

Therefore, this PR removes these indices due to the high tax on insert time


kamil-da:
Thank you @dasormeter for measuring the performance.
Considering the fact that you've mentioned some negative impact on the read queries after dropping the indices and the upcoming work that may approach the indexing in a more comprehensive way, we should consider postponing work on this issue (this was suggested on a Participant Team meeting).
@meiersi-da @mziolekda could you confirm that what I said is correct?

meiersi-da:
@nmarton-da @dasormeter @mziolekda : deciding on the target state for this PR would fit well into the upcoming Oracle IndexDB hand-over work.

PR #11042, opened 2021-09-27
trigger service: allow non-empty schema

garyverhaegen-da:
Untested as yet.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Server.scala
+++ b/triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/Server.scala
@@ -503,6 +503,7 @@ object Server {
       restartConfig: TriggerRestartConfig,
       initialDars: List[Dar[(PackageId, DamlLf.ArchivePayload)]],
       jdbcConfig: Option[JdbcConfig],
+      allowExistingSchema: Boolean,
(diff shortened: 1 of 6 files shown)


stefanobaghino-da (commented):
No objections on my side, but I wouldn't rush using this to put all our tables in a single schema to solve our problems. Adding some sort of test would also be good.

aherrmann-da (approved):
LGTM

Re testing: Perhaps a test-case around `triggerDao.initialize` with a prior schema setup and `allowExistingSchema = false|true` to test that it fails|succeeds as expected would do.
This really only tests that the flag is forwarded to flyway correctly.
It assumes that flyway works, so we don't need to test it itself.
It also assumes that the trigger service will work as expected once the migration completed, so we don't need to duplicate the entire postgres and oracle test suites for each setting.

garyverhaegen-da:
We now believe that adding table prefixes is a better approach, so there is no need for this anymore.

stefanobaghino-da:
> We now believe that adding table prefixes is a better approach, so there is no need for this anymore.

I'm not entirely sure this rules out this needs to be in, though. Not necessary to open this PR here, but I would test the final solution thoroughly to make sure we don't actually need this in as well. /cc @aherrmann-da

PR #11046, opened 2021-09-28
[Divulgence pruning] Fixes divulgence pruning offset update query

tudor-da:
This PR fixes the query update predicate for updating the `participant_all_divulged_contracts_pruned_up_to_inclusive` after the first attempt (when it is not null anymore) on a pruning call with pruneAllDivulgedContracts enabled. Additionally, a test has been added to `StorageBackendPostgresSpec` for checking this.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
@@ -6,7 +6,6 @@ package com.daml.platform.store.backend.common
 import java.io.InputStream
 import java.sql.Connection
 import java.time.Instant
-
 import anorm.SqlParser.{array, binaryStream, bool, int, long, str}
 import anorm.{Row, RowParser, SimpleSql, ~}
 import com.daml.ledger.offset.Offset
@@ -35,6 +34,11 @@ trait EventStorageBackendTemplate extends EventStorageBackend {
 
   def eventStrategy: EventStrategy
   def queryStrategy: QueryStrategy
+  // TODO Refactoring: This method is needed in pruneEvents, but belongs to [[ParameterStorageBackend]].
+  //                   Remove with the break-out of pruneEvents.
+  def participantAllDivulgedContractsPrunedUpToInclusive(
+      connection: Connection
+  ): Option[Offset]
 
   private val selectColumnsForFlatTransactions =
     Seq(
@@ -402,6 +406,8 @@ trait EventStorageBackendTemplate extends EventStorageBackend {
     )(connection)
   }
 
+  // TODO Refactoring: This method is too complex for StorageBackend.
+  //                   Break the method into its constituents and trigger them from the caller of this method.
   override def pruneEvents(
       pruneUpToInclusive: Offset,
       pruneAllDivulgedContracts: Boolean,
@@ -492,14 +498,6 @@ trait EventStorageBackendTemplate extends EventStorageBackend {
             delete_events.event_offset <= $pruneUpToInclusive"""
     }(connection, loggingContext)
   }
-
-  private def participantAllDivulgedContractsPrunedUpToInclusive(
-      connection: Connection
-  ): Option[Offset] =
-    SQL"select participant_all_divulged_contracts_pruned_up_to_inclusive from parameters"
-      .as(offset("participant_all_divulged_contracts_pruned_up_to_inclusive").?.single)(
-        connection
-      )
 
   private def pruneWithLogging(queryDescription: String)(query: SimpleSql[Row])(
       connection: Connection,
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ParameterStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ParameterStorageBackendTemplate.scala
@@ -145,7 +145,7 @@ private[backend] trait ParameterStorageBackendTemplate extends ParameterStorageB
   private val SQL_UPDATE_MOST_RECENT_PRUNING_INCLUDING_ALL_DIVULGED_CONTRACTS =
     SQL("""
           |update parameters set participant_all_divulged_contracts_pruned_up_to_inclusive={prune_all_divulged_contracts_up_to_inclusive}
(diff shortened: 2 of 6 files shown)


ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsPruning.scala
```diff
@@ -17,15 +17,58 @@ private[backend] trait StorageBackendTestsPruning extends Matchers with StorageB
...
+
+      _ <- executeSql(backend.updatePrunedUptoInclusive(offset_2))
+      updatedPruningOffset_2 <- executeSql(backend.prunedUpToInclusive)
+
+      _ <- executeSql(backend.updatePrunedUptoInclusive(offset_3))
+      updatedPruningOffset_3 <- executeSql(backend.prunedUpToInclusive)
+    } yield {
+      initialPruningOffset shouldBe empty
+      updatedPruningOffset_1 shouldBe Some(offset_1)
+      // The pruning offset is not updated if lower than the existing offset
```
tudor-da, line 38: Enriched the existing test for checking the same corner case as well. CC @rautenrieth-da

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -21,14 +20,15 @@ import com.daml.platform.store.Conversions.{
...
-import com.daml.platform.store.backend.EventStorageBackend
+import com.daml.platform.store.backend.{EventStorageBackend, ParameterStorageBackend}
 import com.daml.platform.store.backend.EventStorageBackend.{FilterParams, RangeParams}
 import com.daml.platform.store.backend.StorageBackend.RawTransactionEvent
 import com.daml.platform.store.backend.common.ComposableQuery.{CompositeSql, SqlStringInterpolation}
 
 import scala.collection.compat.immutable.ArraySeq
 
 trait EventStorageBackendTemplate extends EventStorageBackend {
+  this: ParameterStorageBackend =>
```
tudor-da, line 31: @rautenrieth-da @nmarton-da . Added this self type here since we need to query the `participant_all_divulged_contracts_pruned_up_to_inclusive` in the `pruneEvents` here. Not sold on it, but the alternative would be to execute this query before the call to `pruneEvents` and pass it as a parameter, even though it might not be needed. Wdyt?

kamil-da: I would keep the self type here.

kamil-da (approved):
LGTM

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
nmarton-da: @tudor-da can you please add a template method instead? that is also an ugly inheritance code sharing construct, but at least limits the surface of the dependency.
```
def participantAllDivulgedContractsPrunedUpToInclusive(
      connection: Connection
  ): Option[Offset]
```

The jury is still out regarding code sharing between storage backend's, but some thoughts on that matter:
* It would be nice if we wont have to
* It is on the table to switch to composition instead of inheritance here
* Generally the current direction seem to be that storage backend methods seem to be rather small and dumb, therefore code sharing is generally something we can avoid. In your example the `pruneEvents` is rather big and smart, and maybe that is the cause for this dependency hunger here.
* One possible direction (which we also followed with @rautenrieth-da) is to keep the StorageBackend functions small and dumb, and do composition one layer above (what usually is JdbcLedgerDao, but can be something else as well) -- in your case that would mean to rip apart the `pruneEvents` to smaller functions, and move the conditional logic above, which removes the need to share code on storage-backend layer (I absolutely won't do that in this PR, especially because it would involve move of test code as well)
* One other direction would be to change grouping semantics (create a PruningStorageBackend, and move stuff there from EventStorageBackend and ParameterStorageBackend...and maybe from other places), which removes the need again for code sharing between StorageBackend silos.

@tudor-da all that said, I am ok if you just do the template method, but happy to hear your thoughts on that matter!
cc @rautenrieth-da

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ParameterStorageBackendTemplate.scala
```diff
@@ -145,7 +145,7 @@ private[backend] trait ParameterStorageBackendTemplate extends ParameterStorageB
   private val SQL_UPDATE_MOST_RECENT_PRUNING_INCLUDING_ALL_DIVULGED_CONTRACTS =
     SQL("""
           |update parameters set participant_all_divulged_contracts_pruned_up_to_inclusive={prune_all_divulged_contracts_up_to_inclusive}
-          |where participant_pruned_up_to_inclusive < {prune_all_divulged_contracts_up_to_inclusive} or participant_all_divulged_contracts_pruned_up_to_inclusive is null
+          |where participant_all_divulged_contracts_pruned_up_to_inclusive < {prune_all_divulged_contracts_up_to_inclusive} or participant_all_divulged_contracts_pruned_up_to_inclusive is null
```
nmarton-da, line 148: so this is the only actual change in functionality, right?

nmarton-da (approved):
pls switch to template method instead of the selftype
otherwise looks good

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ParameterStorageBackendTemplate.scala
tudor-da: yep

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
tudor-da: Thanks for the suggestion, @nmarton-da . I'll amend with your suggestion. I agree that `pruneEvents` needs a breakdown. Left 2 TODOs.

PR #11047, opened 2021-09-28
Optional table prefix for trigger service 

aherrmann-da:
Enable the new `--db-table-prefix` flag to add a prefix to all tables used by the trigger service to avoid collisions with other components using the same db-schema.

### Pull Request Checklist

trigger the build.


cocreature (commented):
Does it make sense to have a prefix only for history or should we prefix the history and all other tables with the same flag?

garyverhaegen-da:
> Does it make sense to have a prefix only for history or should we prefix the history and all other tables with the same flag?

Prefixing everything seems safer to me.

aherrmann-da:
> Does it make sense to have a prefix only for history or should we prefix the history and all other tables with the same flag?

Makes sense. I've added a prefix to all the tables.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
```diff
@@ -211,6 +212,17 @@ private[trigger] object ServiceConfig {
           + JdbcConfig.help()
       )
 
+    opt[Boolean]("db-table-prefix")
+      .action { (b, c) =>
+        c.copy(dbTablePrefix = if (b) { "trigger_service_" }
```
garyverhaegen-da, line 217: Would it be worth going all the way and letting the user choose the prefix? (Would it be wise?)

garyverhaegen-da (approved):
Thanks!

As a minor nit: on the JSON API side, we're passing in the prefix as a JDBC option for some reason. Should we align the two? (In either direction; I have a preference for a cli arg, as done here, but I could see an argument for the JSON API approach already being deployed as part of 1.17, so harder to change).

stefanobaghino-da (changes requested):
Not really a blocker from me, but I wanted to raise the issue of consistency with the HTTP JSON API and prevent unwanted merges. Feel free to discard my review once how to deal with this has been agreed on.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
garyverhaegen-da: I'm not sure I like that from a UX perspective. Having two different defaults for the same app seems confusing.

> By default, we do not add a prefix to tables. IF you want, you can turn this flag to add the default prefix of `trigger_service_` to every table. You can also pass a second flag (only if you passed the first) to change the default prefix (which is disabled by default).

Seems a bit clunky to explain, whereas

> Pass this flag to set a prefix on all table names. Example: `--flag trigger_service_`.

seems much simpler and lets people achieve exactly the same thing.

garyverhaegen-da:
@stefanobaghino-da @realvictorprm any opinion on the CLI flag v. JDBC option aspect?

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/ServiceConfig.scala
stefanobaghino-da: Sure, that's fine with me, thanks.

stefanobaghino-da:
> @stefanobaghino-da @realvictorprm any opinion on the CLI flag v. JDBC option aspect?

I would aim at keeping this as close as possible to what we already have in the HTTP JSON API. That means the option would be passed alongside the JDBC configuration. @aherrmann-da thoughts? Would it make sense to move `tablePrefix` from `com.daml.http.dbbackend.JdbcConfig` to `com.daml.dbutils.JdbcConfig`?

stefanobaghino-da:
@aherrmann-da

One important point is that the purpose of this is to make sure that the Daml Driver for PostgreSQL and all the components can be configured so that they can coexist within the same database schema. IIUC Flyway mandates the database to be empty when the schema history table is first written, regardless of how it's called, and requires to explicitly opt-in to starting on a non-empty database by means of establishing a baseline or simply saying that existing schemas are allowed.

This is not something that needs to be addressed within this PR, but as a follow up can you check what the behavior is and see if we need to integrate something like what was initially proposed in https://github.com/digital-asset/daml/pull/11042 is necessary to achieve the desired behavior? The Flyway documentation [describes an alternative based on establishing a baseline](https://flywaydb.org/documentation/learnmore/existing) but I don't believe it's feasible for us to pursue that path, since I don't see a clear way in which we can establish any baseline, since components can be initialized in any order.

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V1__Init.sql.sha256
```diff
@@ -1 +1 @@
-6dc2d7d667ab3052662754853f3aeaaa8323ea6f14184312eef5c3ca56a90f1c
+9f5baba232ea6606dcd0c995207e6ec88bfdf495afd897a2e595f6c96365d763
```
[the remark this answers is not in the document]
aherrmann-da: AFAIK These are only used by [our tests](https://github.com/digital-asset/daml/blob/a8d703b9960d70158493234a6dfcfd1be3abb014/libs-scala/flyway-testing/src/main/scala/flyway/AbstractImmutableMigrationsSpec.scala) to avoid unwanted changes of migration steps. In this case, I don't see a way around changing these. The change is such that as long as the table prefix is empty, the migrations are unchanged.

aherrmann-da:
> Would it make sense to move `tablePrefix` from `com.daml.http.dbbackend.JdbcConfig` to `com.daml.dbutils.JdbcConfig`?

I think that's a good approach. I've changed this PR accordingly.



>  IIUC Flyway mandates the database to be empty when the schema history table is first written, regardless of how it's called, and requires to explicitly opt-in to starting on a non-empty database by means of establishing a baseline or simply saying that existing schemas are allowed.
> 
> This is not something that needs to be addressed within this PR, but as a follow up can you check what the behavior is and see if we need to integrate something like what was initially proposed in #11042 is necessary to achieve the desired behavior? The Flyway documentation [describes an alternative based on establishing a baseline](https://flywaydb.org/documentation/learnmore/existing) but I don't believe it's feasible for us to pursue that path, since I don't see a clear way in which we can establish any baseline, since components can be initialized in any order.

I ran a test, first starting `daml sandbox` on a postgresql database and then starting the trigger service with a table prefix on the same database. Indeed, it complains about the existing schema. So someting like #11042 seems still necessary.

stefanobaghino-da:
> > Would it make sense to move `tablePrefix` from `com.daml.http.dbbackend.JdbcConfig` to `com.daml.dbutils.JdbcConfig`?
> 
> I think that's a good approach. I've changed this PR accordingly.
> 
> > IIUC Flyway mandates the database to be empty when the schema history table is first written, regardless of how it's called, and requires to explicitly opt-in to starting on a non-empty database by means of establishing a baseline or simply saying that existing schemas are allowed.
> > This is not something that needs to be addressed within this PR, but as a follow up can you check what the behavior is and see if we need to integrate something like what was initially proposed in #11042 is necessary to achieve the desired behavior? The Flyway documentation [describes an alternative based on establishing a baseline](https://flywaydb.org/documentation/learnmore/existing) but I don't believe it's feasible for us to pursue that path, since I don't see a clear way in which we can establish any baseline, since components can be initialized in any order.
> 
> I ran a test, first starting `daml sandbox` on a postgresql database and then starting the trigger service with a table prefix on the same database. Indeed, it complains about the existing schema. So someting like #11042 seems still necessary.

Fine with me, feel free to make it part of this PR or have it as a separate one.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
```diff
@@ -74,8 +76,8 @@ abstract class DbTriggerDao protected (
   // `Fragment.const` which will try to use a raw string as a SQL query.
 
   private def insertRunningTrigger(t: RunningTrigger): ConnectionIO[Unit] = {
-    val insert: Fragment = sql"""
-        insert into running_triggers
+    val insert: Fragment =
+      sql"insert into " ++ Fragment.const(s"${tablePrefix}running_triggers") ++ sql"""
```
S11001001, line 80:
```suggestion
      sql"""insert into ${Fragment.const(s"${tablePrefix}running_triggers")}
```

S11001001 (approved):
Please always use fragment interpolation instead of `++` for uses like this; I've included one example but it applies to all occurrences of `++ Fragment.const...` changes in this PR.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
aherrmann-da: That looks much nicer, thank you! I fixed the rest.

PR #11049, opened 2021-09-28
fix blackduck logic

garyverhaegen-da:
We believe the Blakduck logic is currently faulty. We have had a
violation on an NPM dependency, and Blackduck keeps reporting it despite
our having removed the dependency.

We believe that what is happening is that, in the first step of
checking, we udpate the Haskell dependencies, _and then check the
validity of the whole project_, which includes the NPM deps. Because
that fails, we never get to the step where we actually update the NPM
deps, and Blackduck is stuck forever.

The solution is to not fail on violations for the Haskell update steps.
Haskell deps are still checked in the second step, because, again, it is
checking the whole project.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/cron/daily-compat.yml
+++ b/ci/cron/daily-compat.yml
@@ -382,7 +382,6 @@ jobs:
           --detect.tools=BAZEL \
           --detect.bazel.target=//... \
           --detect.bazel.dependency.type=${BAZEL_DEPENDENCY_TYPE} \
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Thanks!

dasormeter (approved):
removal of this line in the haskel blackduck run looks good --```detect.policy.check.fail.on.severities=MAJOR,CRITICAL,BLOCKER \```

PR #11050, opened 2021-09-28
auth middleware: no print secret

garyverhaegen-da:
At the moment, the OAuth2 Middleware prints its config at startup.
Unfortunately that config includes a secret, which presumably we'd
prefer not to show.

```
CHANGELOG_BEGIN
- The OAuth2 Middleware now obfuscates its Clien Secret when logging its
  config.
CHANGELOG_END
```


aherrmann-da (approved):
Thanks!

PR #11055, opened 2021-09-28
daml-ledger: better errors for non-json responses.

### LINKED ISSUE #10985: daml-ledger produces uninformative error message when JSON API reports non-JSON response
If we get a response timeout (timing out in the JSON API not client-side) the JSON API returns a non-JSON response. This then results in us failing with
```
- JSONParseException Request {
  host                 = "localhost"
  port                 = 50900
  secure               = False
  requestHeaders       = [("Accept","application/json"),("Authorization","<REDACTED>")]
  path                 = "/v1/packages"
  queryString          = ""
  method               = "POST"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
}
 (Response {responseStatus = Status {statusCode = 503, statusMessage = "Service Unavailable"}, responseVersion = HTTP/1.1, responseHeaders = [("Server","akka-http/10.2.1"),("Date","Wed, 22 Sep 2021 11:22:55 GMT"),("Content-Type","text/plain; charset=UTF-8"),("Content-Length","105")], responseBody = (), responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}) (ParseError {errorContexts = [], errorMessage = "Failed reading: not a valid json value", errorPosition = 1:1 (0)})
```
which is pretty useless.

I think the issue here is that `httpJSON` is too aggressive, it always tries to parse as JSON. We should look at the content-header and include the response body in the exception if it’s not JSON. And if JSON parsing fails, we probably also want to include the header. `httpJSONEither` might be useful for this.

robin-da:
This fixes #10985.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
+++ b/daml-assistant/daml-helper/src/DA/Daml/Helper/Ledger.hs
@@ -584,8 +584,24 @@ instance A.ToJSON AllocatePartyRequest
 httpJsonRequest :: A.FromJSON a => LedgerArgs -> Method -> Path -> (Request -> Request) -> IO a
 httpJsonRequest args method path modify = do
   req <- makeRequest args method path modify
-  resp <- runHttp httpJSON req
-  pure $ result resp
+  respOrErr <- runHttp httpJSONEither req
+  case respOrErr of
+    Right resp -> pure $ result resp
+    Left err ->
+      case err of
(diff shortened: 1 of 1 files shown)


cocreature (approved):
looks sensible, thanksk

Issue #11053, opened 2021-09-28
Buf compatibility checks vs Daml-LF language, transaction & value protobufs


cocreature:
The recently buf compatibility checks for KV protobufs interact badly withe the way we develop the 3 different Daml-LF protobufs: First a summary of the current state:

1. The data in the Daml-LF protobufs is strictly versioned by the Daml-LF version. This is not the same as the SDK version, we can have multiple SDK releases without a new Daml-LF version.
2. In addition to the stable Daml-LF versions, we have a 1.dev version. This one is explicitly for trying out experiments, no data continuity is provided, ledgers reject it by default so protobuf compatibility here is explicitly not a goal.
3. However, the 1.dev protobufs needs to be wire compatible against older stable LF releases since we always use that for decoding. Enforcing anything more than wire-compatibility here is not useful.
4. The daml-lf en/decoders have extensive hand-crafted backwards compatibility checks and we have integration tests as well for this.
4. The KV protobufs reference currently both the language protobuf as well as values and transactions and they reference the 1.dev versions of those protobufs. The latter is deliberate, we want to be able to try out new features like exceptions on a KV stack.
5. We disabled the check for the language protobuf in https://github.com/digital-asset/daml/commit/a4629a4450de7e525276e04f66ff221b7ded1f13 and #11012 decouples things such that the 1.dev issue no longer arises.

So that leaves us with the buf checks for the transaction and value protobuf. Currently, those prevent us from developing the 1.dev versions in the way we want to because that could break compatibility with the 1.dev version of the latest stable SDK release.

My suggestion is the following:
1. As part of the kv buf checks, exclude `com.daml.lf.transaction` and `com.daml.lf.value`. This gets us into a state where we can develop 1.dev as we want. Compatibility is not ensured via buf but via the existing tests for the en/decoder.
2. To recover the checks from buf as an extra safeguard, we keep a copy of the value & transaction protobuf for the latest stable Daml-LF version in the repository and we setup (wire-compat) buf checks for the 1.dev protobufs against the latest stable one.

Ideally we do 2 before 1 but given the extensive tests we have, if we end up being blocked on the current checks, we can do 1 before. The language team will start working on 2 with the estimate to finish it in the next month (from issue creation).

PR #11064, opened 2021-09-29
LF: check LF transaction protobuf aginst local snapshots

remyhaemmerle-da:
instead of last stable branch.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


daml-lf/transaction/src/stable/protobuf/com/daml/lf/value.proto
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
cocreature, line 1: I was looking at diffs and the dev version of transaction has
```
> // * dev (special staging area for the next version to be released)
```
while the dev version of value doesn’t have that header. maybe worth adding? or maybe drop it from both so stabilizing is just copying the file. Seems easy to forget to drop that header.

daml-lf/transaction/src/stable/protobuf/com/daml/lf/transaction.proto
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
cocreature, line 1: definitely not for this PR but I think it would be great to have a `RELEASE-LF.md` file somewhere which describes the steps to go through for making a stable LF release.

cocreature (approved):
Looks very sensible, thank you! Verified that the protos are really just copies and that the test fails if I drop a field.

fixes #11053

daml-lf/transaction/src/stable/protobuf/com/daml/lf/value.proto
remyhaemmerle-da: In general, we cannot  just "copy" the files. 
Since `dev` features may be dropped without further notice, we have to remove from the file any `dev` feature which is not 
stable at the time of freeze. 

I think it is more clean to make a manual pass over the file and remove all reference to dev before freeze.
This could be describe in the document you asked for below.

cocreature: makes sense, then I think we should also add it to `value.proto`.

remyhaemmerle-da: 210c843c

PR #11067, opened 2021-09-29
Explicit discard in daml-lf/interpreter

nickchapman-da:
Add explicit `discard`s in `daml-lf/interpeter` to allow `NonUnitStatement` wart detector to be switched on.

### DIFF
--- a/daml-lf/interpreter/BUILD.bazel
+++ b/daml-lf/interpreter/BUILD.bazel
@@ -7,12 +7,11 @@ load(
     "da_scala_test",
     "da_scala_test_suite",
     "lf_scalacopts",
+    "lf_scalacopts_stricter",
 )
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repl")
 
-lf_scalacopts_stricter = lf_scalacopts + [
-    "-P:wartremover:traverser:org.wartremover.warts.NonUnitStatements",
-]
+print(lf_scalacopts_stricter)
 
 da_scala_library(
     name = "interpreter",
--- a/daml-lf/interpreter/BUILD.bazel
+++ b/daml-lf/interpreter/BUILD.bazel
@@ -10,6 +10,10 @@ load(
 )
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repl")
 
+lf_scalacopts_stricter = lf_scalacopts + [
+    "-P:wartremover:traverser:org.wartremover.warts.NonUnitStatements",
+]
+
 da_scala_library(
     name = "interpreter",
     srcs = glob(["src/main/**/*.scala"]),
@@ -19,7 +23,7 @@ da_scala_library(
         "@maven//:org_scalaz_scalaz_core",
         "@maven//:org_typelevel_paiges_core",
     ],
-    scalacopts = lf_scalacopts,
+    scalacopts = lf_scalacopts_stricter,
     tags = ["maven_coordinates=com.daml:daml-lf-interpreter:__VERSION__"],
     visibility = [
         "//compiler/repl-service:__subpackages__",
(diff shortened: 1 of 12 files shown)


daml-lf/interpreter/BUILD.bazel
```diff
@@ -10,6 +10,10 @@ load(
 )
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repl")
 
+lf_scalacopts_stricter = lf_scalacopts + [
```
cocreature, line 13: maybe move this next to the definition of `lf_scalacopts`?

cocreature (approved):
Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
```diff
@@ -229,7 +230,7 @@ private[lf] object Speedy {
     private[speedy] def getEnvFree(i: Int): SValue = frame(i)
 
     @inline def pushEnv(v: SValue): Unit = {
-      env.add(v)
+      discard[Boolean](env.add(v))
```
remyhaemmerle-da, line 233: Why do we need this `[Boolean]` ?

remyhaemmerle-da (approved):
LGTM.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
cocreature: you don’t need it but it can be nice to be explicit in what you are discarding both for readability and to ensure that if the type changes it blows up and you need to reconsider whether dropping is still the right thing to do.
I have no strong feelings on whether it makes sense in this particular instance or not, I think it’s a case by case decison, sometimes it helps sometimes it doesn’t.

remyhaemmerle-da: :+1:

daml-lf/interpreter/BUILD.bazel
nickchapman-da: ok, I'll look into doing this

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Speedy.scala
nickchapman-da: I couldn't imagine what was being returned by `ArrayList.add`, so I checked some docs:
https://www.tutorialspoint.com/java/util/arraylist_add.htm
And discovered:
```
Return Value
This method returns true.
```
which made me so cross I thought to document with the type.

nickchapman-da: I wonder if there is an easy way to extend the `ArrayList` class so the `add` method returns `Unit`, instead of having to `discard` at every point :shrug:

daml-lf/interpreter/BUILD.bazel
nickchapman-da: did it

daml-lf/interpreter/BUILD.bazel
```diff
@@ -7,9 +7,12 @@ load(
     "da_scala_test",
     "da_scala_test_suite",
     "lf_scalacopts",
+    "lf_scalacopts_stricter",
 )
 load("@io_bazel_rules_scala//scala:scala.bzl", "scala_repl")
 
+print(lf_scalacopts_stricter)
```
stefanobaghino-da, line 14: @nickchapman-da Did you mean to leave this in? I opened https://github.com/digital-asset/daml/pull/11091, if this is not supposed to be in you only need to approved, otherwise feel free to close it. 🙂

PR #11077, opened 2021-09-29
CI: build PRs not merges

garyverhaegen-da:
On a PR build, Azure builds the result of merging the PR into `main`.
That used to be good, because it meant we had lower chances of ending up
with semantic conflicts between unrelated PRs.

Recently, though, Azure's behaviour has been a bit erratic as to which
commit it merges with. We're not entirely sure how big of an impact this
has on caching, but the only way we can think of to test is to disable
that behaviour and see what happens to our caching times.

This PR changes the YML file for PR builds to trigger on all branches
(except release branches), rather than on PRs.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/ci/prs.yml
+++ b/ci/prs.yml
@@ -1,14 +1,19 @@
 # Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-trigger: none
-
-pr:
-  autoCancel: true # cancel previous builds on push
+#trigger: none
+trigger:
   branches:
-    include:
-      - main
-      - release/*
+    exclude:
+    - main
+    - release/*
+
+pr: none
+# autoCancel: true # cancel previous builds on push
+# branches:
+#   include:
+#     - main
+#     - release/*
 
 jobs:
 - template: build.yml


aherrmann-da:
Does this mean that CI will now also run on branches that don't have a corresponding PR opened, yet?

garyverhaegen-da:
Yes.

ci/prs.yml
```diff
@@ -1,14 +1,19 @@
...
+trigger:
   branches:
-    include:
-      - main
-      - release/*
+    exclude:
+    - main
+    - release/*
+
+pr: none
```
cocreature, line 11: how does this interact with checks on github for PRs? Do they still show up (I expect so, iirc that’s what happened for manually triggered builds for a branch) and are they still blocking (not sure, at least for manual builds I think that was not the case)?

cocreature: If I understand this correctly, this effectively means that we’re giving up on not being able to schedule builds for arbitrary commits without leading some form of audit trail in the form of a PR which has security implications. Not necessarily saying this is a bad idea but probably at least worth thinking about.

garyverhaegen-da: > how does this interact with checks on github for PRs? Do they still show up and are they still blocking?

Yes and yes. The fact that they are blocking has nothing to do with why they started, it's just based on the name of the status report. Manual runs are blocking right now if they're runs of the PR pipeline (which we very rarely do for obvious reasons); manual runs of the compat pipeline are not blocking because the compat pipeline is not blocking.

garyverhaegen-da:
On second thought, there are too many things we do that rely on building the merge commit. I think this would cause more issues than solve.

PR #11078, opened 2021-09-29
[Short] Add empty parameter list to method creating new objects

pbatko-da:


ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -17,7 +17,7 @@ trait ErrorFactories {
 
   import ErrorFactories._
 
-  def duplicateCommandException: StatusRuntimeException =
+  def duplicateCommandException(): StatusRuntimeException =
```
tudor-da, line 20: The convention that I know is that empty-parens are for side-effectful methods. Is this the case in here?

hubert-da: I agree, I wouldn't change that.

pbatko-da: True, there are no side-effects but I'm not sure I can call it referentially transparent and thus pure, because it creates new instances which I, conservatively, assume to be mutable.

Btw, this PR stems from me confusing `duplicateCommandException` to be a val rather than a method because it neither has `()` nor does start with a verb.

🚲 🛖

PR #11081, opened 2021-09-29
LF: move archive snapshots in a separate directory

remyhaemmerle-da:
### Pull Request Checklist

trigger the build.


daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/ProtoTest.scala
```diff
@@ -206,7 +206,7 @@ class ProtoTest extends AnyWordSpec with Matchers with TableDrivenPropertyChecks
     s"daml_lf_1_$version files" should {
 
       val pkg = if (version < 11) "digitalasset" else "daml"
-      val rootDir = s"daml-lf/archive/src/main/protobuf/com/$pkg/daml_lf_1_$version"
+      val rootDir = s"daml-lf/archive/src/snapshot/protobuf/com/$pkg/daml_lf_1_$version"
```
cocreature, line 209: maybe `stable` would be clearer? We have snapshot SDK releases and they are something very different so reusing the name might be confusing.

cocreature (approved):
Thanks!

daml-lf/archive/src/test/scala/com/digitalasset/daml/lf/ProtoTest.scala
remyhaemmerle-da: 19b5cfbc923d46df86c353fa265ebdbdcd2d55b8

PR #11082, opened 2021-09-29
AuthorizationSpec

nickchapman-da:
Unit test the various forms of `FailedAuthorization` which can be returned from `CheckAuthorization`
for the 4 kinds of `GenActionNode`: Create/Fetch/Lookup/Exercise.


remyhaemmerle-da (approved):
Thanks

PR #11084, opened 2021-09-29
[Docs] Bump sdk version to 1.17.0 in manual-download.html.

arturpoor-da:


cocreature (commented):
Thanks for the PR!
I’m not sure we want to keep updating the version, it will just get out of date again. The docs explicitly state that this is only an example.

arturpoor-da:
I know manually setting version number is not the way it should be done. Rephrasing in order to be version-agnostic (e.g. `daml-sdk-x.y.z-macos.tar.gz`) is a long-term low-cost solution.

My motivation behind this PR: The following sentence caught my eyes: _For example, if you are on macOS and want to install release 1.4.0, [...]_ and I was thinking why would anyone want to install 1.4.0? :)

victormueller-da (approved):

victormueller-da:
Bumping from time to time is not too bad :smile:

cocreature:
> My motivation behind this PR: The following sentence caught my eyes: _For example, if you are on macOS and want to install release 1.4.0, [...]_ and I was thinking why would anyone want to install 1.4.0? :)

Can we rephrase this to suggest something like “install the latest release (1.4.0 at the time of writing)” to avoid the confusion and avoid having to bump this in the future.

victormueller-da:
Sounds good to me too, in best case we can link to the latest release then via `https://github.com/digital-asset/daml/releases/latest`

arturpoor-da:
Thank you.

> Can we rephrase this to suggest something like “install the latest release (1.4.0 at the time of writing)” to avoid the confusion and avoid having to bump this in the future.

So if I get it right, the consensus is to go with "1.4.0 at the time of writing". Please advise what to do next. Should I drop this PR and create a new, or simply add a new commit here (basically reverting 5f1a94a) ?

victormueller-da:
You can change the title of the pr and just overwrite the history, after all it is your fork :smile:

arturpoor-da:
I overwrote the commit, and renamed the PR. Please review.

PR #11090, opened 2021-09-30
Split daml kvutils proto [KVL-1090]

nicu-da:
### Pull Request Checklist

- Reorganize the kvutils proto file into multiple easier to manage files
- Binary compatibility is preserved
- Imports for generated code are changed

Open items: 
- Should we split the file further, extracting transactions, log entries, package uploads, and so on into different files?
- To avoid circular imports we had to split the entries related to configuration into 2 files, `daml_configuration.proto` and `daml_configuration_rejection.proto`. I'm not really a fan of this as they really belong together. The other option would be to split the new file `daml_state.proto` into `daml_state_key.proto` and `daml_state_value.proto` but I have no strong opinion as to which split would be better.
 
trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_configuration.proto
+++ b/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_configuration.proto
@@ -1,0 +1,53 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//
+// Protocol buffer messages used by the participant state key-value utilities
+// for implementing a Daml ledger backed by a key-value store.
+//
+// These messages should only be produced and consumed by the methods in
+// `KeyValueCommitting`, `KeyValueConsumption` and `KeyValueSubmission` objects.
+//
+
+syntax = "proto3";
+
+package com.daml.ledger.participant.state.kvutils;
+
+option java_package = "com.daml.ledger.participant.state.kvutils";
+option csharp_namespace = "Com.Daml.Ledger.Participant.State.KVUtils";
+
+import "google/protobuf/timestamp.proto";
+import "com/daml/ledger/configuration/ledger_configuration.proto";
+
+// Configuration change request to change the ledger configuration.
+message DamlConfigurationSubmission {
+  // A unique string scoped to a particular participant for matching the
+  // request with the result.
+  // Implementers are free to select adequate mechanism e.g. UUID or similar.
+  string submission_id = 1;
+
+  // Submitting participant's id.
+  string participant_id = 2;
+
+  // The maximum record time after which the submission will be rejected.
+  // Allows submitter to control when the request times out and to retry.
+  google.protobuf.Timestamp maximum_record_time = 3;
+
+  // The new configuration that replaces the current configuration.
+  com.daml.ledger.configuration.LedgerConfiguration configuration = 4;
+}
+
+// Configuration entry that records a configuration change.
+// Also used in state to look up latest configuration.
+// When a configuration exists, only the participant that
+// submitted previously can change it.
+message DamlConfigurationEntry {
+  // The submission from which this configuration originated.
+  string submission_id = 1;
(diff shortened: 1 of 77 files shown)


fabiotudone-da (commented):
I am a bit puzzled by the naming convention: why do we prefix some files with `daml_` and not others? I think that e.g. a `kv_` prefix for all would be more useful to find them quickly.

As for the split, I propose to go the extra mile because it seems a bit inconsistent to me if we only extract the configuration.

nicu-da:
> I am a bit puzzled by the naming convention: why do we prefix some files with `daml_` and not others? I think that e.g. a `kv_` prefix for all would be more useful to find them quickly.
> 
> As for the split, I propose to go the extra mile because it seems a bit inconsistent to me if we only extract the configuration.

Wdyt about removing the prefix altogether? The problem is that we already have the parent folder `kvutils`, adding another `kv` prefix turns the generated code into some weird repetition eg: `kvutils.RejectionReason` -> `kvutils.KvRejectionReason`

fabiotudone-da:
> Wdyt about removing the prefix altogether?

Also works, as file finding in the IDE seems smart enough to take into account a directory prefix.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/daml_configuration.proto
```diff
@@ -0,0 +1,53 @@
...
+package com.daml.ledger.participant.state.kvutils;
+
+option java_package = "com.daml.ledger.participant.state.kvutils";
+option csharp_namespace = "Com.Daml.Ledger.Participant.State.KVUtils";
+
+import "google/protobuf/timestamp.proto";
+import "com/daml/ledger/configuration/ledger_configuration.proto";
+
+// Configuration change request to change the ledger configuration.
+message DamlConfigurationSubmission {
```
SamirTalwar-DA, line 23: If we move this into _submission.proto_, can we move the other configuration message types into the correct places?

miklos-da: The `wire` package contains messages that we only transmit via the network (and not persist to the ledger), hence, `DamlConfigurationSubmission` definitely belongs there.

nicu-da: @miklos-da good point, the `DamlConfigurationSubmission` definitely belongs there. 
@SamirTalwar Unfortunately not. `DamlConfigurationSubmission` is indeed only reference from `submission.proto`. The problem is `DamlConfigurationEntry` is used in `state.proto`. `state.proto` is used by `rejection_reason.proto` which is in turn used by `configuration_rejection.proto`. This means we can't merge `configuration.proto` and `configuraiton_rejection.proto` because it generates a circular import.

SamirTalwar-DA: Welp, boo.

In the future perhaps we can decouple these.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/Envelope.scala
```diff
@@ -6,6 +6,7 @@ package com.daml.ledger.participant.state.kvutils
 import java.util.zip.{GZIPInputStream, GZIPOutputStream}
 
 import com.daml.ledger.participant.state.kvutils.{DamlKvutils => Proto}
+import com.daml.ledger.participant.state.kvutils.{DamlState => ProtoState}
```
SamirTalwar-DA, line 9: Perhaps you can just stop aliasing these. I think I did this to make things clearer but because it's different to every other file, it's probably not very helpful.

SamirTalwar-DA (approved):
If you want to change the package, that's fine, but I'm approving it as is anyway.

nicu-da:
@miklos-da @SamirTalwar-DA @fabiotudone-da @hubert-da Everyone ok with merging this and proceeding with the split in follow up PRs?

PR #11083, opened 2021-09-29
Fix SimpleDeduplicationBasic when participant deduplication is disabled [KVL-1083]

fabiotudone-da:
The party was not passed down with participant deduplication disabled and would result in an invalid argument error.

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #11095, opened 2021-09-30
KV ledgers: disable participant side command de-duplication [KVL-1083]

fabiotudone-da:
## Description:

- Disables participant side command deduplication in `KeyValueParticipantStateWriter`.
- Informs `KVCommandDeduplicationIT` and `AppendOnlyKVCommandDeduplicationIT` about the above.
- Leaves nevertheless participant side command deduplication enabled for sandbox-next when using static time (as this combination is currently unsupported by KV).
- Fixes an issue in `CommandDeduplicationBase` when triggering async deduplication.
- Adapts `KVDeduplicatioBase` so that it doesn't rely on participant side command deduplication.
- Disable cross-compatibility checks between this ledger API test tool and previous platform versions.

### Pull Request Checklist

trigger the build.


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
```diff
@@ -205,11 +205,11 @@ private[testtool] abstract class CommandDeduplicationBase(
     for {
       // Submit a command as alice
       _ <- submitRequestAndAssertCompletionAccepted(ledger)(aliceRequest, alice)
-      _ <- submitRequestAndAssertDeduplication(ledger)(aliceRequest)
+      _ <- submitRequestAndAssertDeduplication(ledger)(aliceRequest, alice)
 
       // Submit another command that uses same commandId, but is submitted by Bob
       _ <- submitRequestAndAssertCompletionAccepted(ledger)(bobRequest, bob)
-      _ <- submitRequestAndAssertDeduplication(ledger)(bobRequest)
+      _ <- submitRequestAndAssertDeduplication(ledger)(bobRequest, bob)
```
fabiotudone-da, line 212: This is a fix akin to #11083.

compatibility/bazel_tools/testing.bzl
```diff
@@ -435,6 +435,18 @@ excluded_test_tool_tests = [
...
+        "platform_ranges": [
+            {
+                "exclusions": [
+                    "KVCommandDeduplicationIT:KVCommandDeduplicationDeduplicateSubmitterBasic",
+                    "KVCommandDeduplicationIT:KVCommandDeduplicationSimpleDeduplicationBasic",
+                    "KVCommandDeduplicationIT:KVCommandDeduplicationCommitterDeduplication",
+                ],
+            },
+        ],
+    },
```
fabiotudone-da, line 449: What I'd like here is "don't cross-test the ledger API test tool V(snapshot after this PR is merged) with previous platform versions" but not sure how I can achieve this without a further change to this file after the next snapshot is released.

hubert-da: Please make sure you get the above docs (starting at line 12), adding exclusions here is a bit tricky. The next snapshot version can be marked as `1.18.0-snapshot.20210928.7948.1`. So I think you'd like sth similar to what I did at line 396 for the gRPC error assertions.

fabiotudone-da: Yes, that's indeed the point; I think I get now that `1.18.0-snapshot.20210928.7948.1` is used not so much as "the next version" but rather as a fake version between the current and the next one to be used as a "start" without explicitly naming the next one (b/c we don't yet know its name). I'll try and improve the doc a bit.

SamirTalwar-DA: If you feel like changing this stuff so `"start"` is exclusive, not inclusive, it would probably simplify these compatibility checks. I think that's almost always what we want.

SamirTalwar-DA (approved):
Looks good!

hubert-da (approved):

nicu-da (approved):

fabiotudone-da:
How can I avoid checking against `platform-0.0.0`?

garyverhaegen-da:
> How can I avoid checking against `platform-0.0.0`?

As `HEAD` (`0.0.0`) is considered "after" all other versions, if you don't specify an `end` to your exclusion range, it will include `HEAD`. Though given that you currently _only_ have an `end`, that would mean the test is always skipped after if test-tool > `1.18.0-snapshot.20210928.7948.1`, and at that point you might as well just delete the tests, no?

fabiotudone-da:
> As `HEAD` (`0.0.0`) is considered "after" all other versions, if you don't specify an `end` to your exclusion range, it will include `HEAD`.
> Though given that you currently _only_ have an `end`, that would mean the test is always skipped after if test-tool > `1.18.0-snapshot.20210928.7948.1`, and at that point you might as well just delete the tests, no?

I think my understanding of the cross-test exclusions is still flawed somewhere. This change updates KV ledgers in such a way that previous ledger API test tools would fail on them; it also updates the ledger API test tool so that it would fail on previous KV ledgers. Which exclusion should I use in such a case?

garyverhaegen-da:
> > As `HEAD` (`0.0.0`) is considered "after" all other versions, if you don't specify an `end` to your exclusion range, it will include `HEAD`.
> > Though given that you currently _only_ have an `end`, that would mean the test is always skipped after if test-tool > `1.18.0-snapshot.20210928.7948.1`, and at that point you might as well just delete the tests, no?
> 
> I think my understanding of the cross-test exclusions is still flawed somewhere. This change updates KV ledgers in such a way that previous ledger API test tools would fail on them; it also updates the ledger API test tool so that it would fail on previous KV ledgers. Which exclusion should I use in such a case?

The logic (as far as I understand it, at least 😅) is that **if** the api test tool version is in the "external" (JSON-ly speaking) range, **and** the ledger version is in the "internal" range, **then** the exclusions are passed as CLI parameters.

It looks like what you want to say in this case is to add the exclusions if either:
- The test tool is higher than <last current release> (not included) **and** the ledger is before <last current release> (included), or
- The test tool is lower than <last current release> (included) **and** the ledger is higher than <last current release> (excluded).

Does that sound right? If so, you'll need two exclusion blocks to express that.

fabiotudone-da:
> > > As `HEAD` (`0.0.0`) is considered "after" all other versions, if you don't specify an `end` to your exclusion range, it will include `HEAD`.
> > > Though given that you currently _only_ have an `end`, that would mean the test is always skipped after if test-tool > `1.18.0-snapshot.20210928.7948.1`, and at that point you might as well just delete the tests, no?
> > 
> > 
> > I think my understanding of the cross-test exclusions is still flawed somewhere. This change updates KV ledgers in such a way that previous ledger API test tools would fail on them; it also updates the ledger API test tool so that it would fail on previous KV ledgers. Which exclusion should I use in such a case?
> 
> The logic (as far as I understand it, at least sweat_smile) is that **if** the api test tool version is in the "external" (JSON-ly speaking) range, **and** the ledger version is in the "internal" range, **then** the exclusions are passed as CLI parameters.
> 
> It looks like what you want to say in this case is to add the exclusions if either:
> 
> * The test tool is higher than [last current release] (not included) **and** the ledger is before [last current release] (included), or
> * The test tool is lower than [last current release] (included) **and** the ledger is higher than [last current release] (excluded).
> 
> Does that sound right? If so, you'll need two exclusion blocks to express that.

bf5a815bf9d9cd57fba46902e7404782db52070c expresses what I want at a PR level (i.e., HEAD ledger API tool is incompatible with all released platforms and HEAD platform is incompatible with all released ledger API) and it works and it would also work for the nightly checks, but it will need to change into a proper release as soon as a new snapshot is tagged. AFAIU, since HEAD is greater than anything, I can't use e.g. `1.18.0-snapshot.20210928.7948.1` (> last current release) in its stead because HEAD would still be included. Do you see less inconvenient alternatives?

fabiotudone-da:
> [bf5a815](https://github.com/digital-asset/daml/commit/bf5a815bf9d9cd57fba46902e7404782db52070c) expresses what I want at a PR level (i.e., HEAD ledger API tool is incompatible with all released platforms and HEAD platform is incompatible with all released ledger API tools) and it works and it would also work for the nightly checks, but it will need to change into a proper release as soon as a new snapshot is tagged. AFAIU, since HEAD is greater than anything, I can't use e.g. `1.18.0-snapshot.20210928.7948.1` (> last current release) in its stead because HEAD would still be included. Do you see less inconvenient alternatives?

@garyverhaegen-da On second thought, I think your suggestion is right and should work (because HEAD is excluded at the top of the interval).

fabiotudone-da:
Thanks @garyverhaegen-da!

garyverhaegen-da:
LGTM, and I don't think you need to change to "real" releases once we do the next one. I, at least, don't mind if the bounds we use in that file are not real releases.

garyverhaegen-da:
Nitpick, though: if this commit introduces a break in the compat tests, does that mean it introduces a breaking change that our users should be informed of? It seems a bit weird to me to see a PR that needs to add these kinds of exclusions while not having any CHANGELOG entry.

fabiotudone-da:
> I don't think you need to change to "real" releases once we do the next one. I, at least, don't mind if the bounds we use in that file are not real releases.

Yes, your solution has this advantage.

fabiotudone-da:
> Nitpick, though: if this commit introduces a break in the compat tests, does that mean it introduces a breaking change that our users should be informed of? It seems a bit weird to me to see a PR that needs to add these kinds of exclusions while not having any CHANGELOG entry.

It's a good point but this change is under the hood, i.e. it only switches the provider of a ledger API behavior (command deduplication) from the participant to the KV driver. Since the ledger API doesn't specify whether deduplication responses are sync or async, this is invisible to the user [that doesn't rely on unspecified API behavior]. @miklos-da @andreaslochbihler-da do you agree?

andreaslochbihler-da:
The new exclusions refer to kv-specific tests and therefore are actually not conformance tests. So no changelog entry for those needed IMO.

PR #11106, opened 2021-10-01
Add documentation for gRPC error metadata [KVL-1119]

hubert-da:
And a missing test

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/app-dev/services.rst
+++ b/docs/source/app-dev/services.rst
@@ -89,6 +89,7 @@ The command submission service deduplicates submitted commands based on their :r
 
   - The result of the submission (``Empty`` or a gRPC error), if the command was submitted outside of the deduplication period of a previous command with the same :ref`change ID <change-id>` on the same participant.
   - The status error ``ALREADY_EXISTS``, if the command was discarded by the ledger server because it was sent within the deduplication period of a previous command with the same :ref`change ID <change-id>`.
+  - Additional metadata (in case of errors), which should be easily parsable. An example is ``definite_answer`` telling whether a different response may be expected when resubmitting the same command.
 
 - If the ledger provides additional command deduplication across participants, the initial command submission might be successful, but ultimately the command can be rejected if the deduplication check fails on the ledger.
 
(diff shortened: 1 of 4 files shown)


docs/source/app-dev/services.rst
```diff
@@ -89,6 +89,7 @@ The command submission service deduplicates submitted commands based on their :r
 
   - The result of the submission (``Empty`` or a gRPC error), if the command was submitted outside of the deduplication period of a previous command with the same :ref`change ID <change-id>` on the same participant.
   - The status error ``ALREADY_EXISTS``, if the command was discarded by the ledger server because it was sent within the deduplication period of a previous command with the same :ref`change ID <change-id>`.
+  - Additional metadata (in case of errors), which should be easily parsable. An example is ``definite_answer`` telling whether a different response may be expected when resubmitting the same command.
```
hubert-da, line 92: I know this one is rather laconic but we discussed that we do not want to document each and every metadata entry that we return.

meiersi-da: @hubert-da : How do you expect application authors to discover what metadata keys they can rely on?

hubert-da: @meiersi-d, we may want to revisit adding them to the documentation. So far, the clients can find out:
- from the code (and tests)
- by trial and error
- by asking us (at least we would know if someone uses it).
Another idea would be to generate the documentation, like we do as part of the gRPC error code-related work, I think. It may not be easy, though, as different metadata keys belong to different errors.

meiersi-da: I'd expect that all data that Ledger API clients need to rely on is well-documented either in the .rst files or in the .proto files. This way we explicitly choose what data we expose and commit to exposing in the future.

@andreaslochbihler-da @miklos-da could you have a look at what kind of data that is in the context of command deduplication, and how to best document this?

andreaslochbihler-da: I prefer to not mention `definite_answer` here. As long as we don't have submission ranks, the meaning of `definite_answer` is irrelevant.

andreaslochbihler-da: I absolutely agree with @meiersi-da that we should be explicit about what the clients can rely on. IMO clients should only rely on the machine-readable part, as we want to take the freedom to arbitrarily change non-machine-readable parts. The documentation should clearly state what the machine-readable parts of an error are.

The error code proposal by @daravep and @meiersi-da currently says that only the error code in the error info is machine-readable and nothing else. That's also what I would currently document. For command dedup, we probably want to add more details as "machine-readable" and therefore document them. This should only happen after we have decided on the exact scope that we want to implement for now.

hubert-da: Thanks everyone for your help. As this is KV ledger-specific behavior, we decided to create a Confluence page with the description and provide VMware with it.

hubert-da:
Addressed the review comments, so it's even smaller now.

PR #11108, opened 2021-10-04
Port file was not written to... where?

stefanobaghino-da:
Adds the information on where the port file could not be written.
From the current phrasing it appears the intention was to add the
information but ultimately wasn't.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compatibility/bazel_tools/daml_ledger/Sandbox.hs
+++ b/compatibility/bazel_tools/daml_ledger/Sandbox.hs
@@ -143,8 +143,8 @@ nullDevice
     | otherwise =  "/dev/null"
 
 readPortFile :: Int -> String -> IO Int
-readPortFile 0 _file = do
-  T.hPutStrLn stderr "Port file was not written to in time."
+readPortFile 0 file = do
+  T.hPutStrLn stderr ("Port file was not written to '" ++ file ++ "' in time.")
   exitFailure
 readPortFile n file = do
   fileContent <- catchJust (guard . shouldCatch) (readFile file) (const $ pure "")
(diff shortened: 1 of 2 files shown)


compatibility/bazel_tools/daml_ledger/Sandbox.hs
```diff
@@ -143,8 +143,8 @@ nullDevice
     | otherwise =  "/dev/null"
 
 readPortFile :: Int -> String -> IO Int
-readPortFile 0 _file = do
-  T.hPutStrLn stderr "Port file was not written to in time."
+readPortFile 0 file = do
+  T.hPutStrLn stderr ("Port file was not written to '" ++ file ++ "' in time.")
```
stefanobaghino-da, line 147: Is this output correct at all? It looks to me that an attempt is being made at reading a file, not writing it. I'm confused.

stefanobaghino-da: Ok, I guess the idea here is that `readPortFile` happens in parallel with another process writing the file and the intention here is to report that.

aherrmann-da: Correct, the Haskell side will start some service that is expected to write a port file and will wait for this to happen up to a timeout.

aherrmann-da, line 147: This fails to build because `file :: String`, but `T.hPutStrLn` expects `Text`. You could use [`pack`](https://hackage.haskell.org/package/text-1.2.5.0/docs/Data-Text.html#v:pack) in something like this instead:
```
"Port file ..." <> T.pack file <> "..."
```

aherrmann-da (approved):

PR #11114, opened 2021-10-04
Try and improve the explanation of compatibility test exclusions

fabiotudone-da:
### Pull Request Checklist

trigger the build.


SamirTalwar-DA (approved):
Looks good, but I suggest getting @garyverhaegen-da to review too.

fabiotudone-da:
Moved to draft as the example is still not correct as I couldn't yet find a way to skip cross-testing against `platform-0.0.0`.

compatibility/bazel_tools/testing.bzl
```diff
@@ -9,19 +9,37 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
...
+# smaller than the upcoming snapshot version (because either the date or the commits count
+# or both will increase).
+#
+# An example scenario is a change to both a ledger and the Ledger API Test Tool that makes the
+# latter incompatible with previous versions of the ledger.
+# In this case, you'll want to set the platform `start` and `end` to be unbounded as to cover
+# all existing platform versions (including HEAD) and the test tool `start` to be the upcoming
+# snapshot, which you can denote by using the "intermediate version" trick above.
+# After your change is merged, you'll need a further change to set the platform `end` to the
+# latest released snapshot, so that HEAD and subsequent versions are cross-checked again.
```
garyverhaegen-da, line 34: I don't quite understand this. First off, the goal of these tests is to prevent us from breaking things; exclusions should be used primarily for when we fix bugs in the test tool (in which case it makes sense that _old_ test tools cannot run against _new_ platforms), or when we add features (in which case it makes sense that _new_ tests cannot run against _old_ platforms, where the feature did not exist yet).

In your example, presumably HEAD against HEAD works, or else you shouldn't be able to merge. Therefore, I don't quite get your recommended bounds. It seems to me what you want is to add exclusions for api-test-tool starting at current snapshot +1 and platforms ending at current snapshot +1. That still leaves you running HEAD against HEAD, but, again, presumably that works.

Also I don't see a need to go back and update afterwards, as (again, presumably) newer versions of the test tool and the ledger should work against each other.

fabiotudone-da: I'd like to add some common use cases to the doc; you've just enumerated two and the one I refer to is basically their combination (a PR that changes both the ledger API test tool and the platform so that they are compatible with each other but not with previous versions of either).

I realize I am not sure anymore what HEAD is here, though. Is it the result of merging this PR with the target branch at the point the build is started?

garyverhaegen-da: `HEAD` is "the current commit", i.e. the one the build is being run for. We run the compat tests daily, and for those runs HEAD is "the tip of main at 6AM UTC".

We also run a subset of the compat tests on each PR (IIRC just HEAD against latest release), and in those cases HEAD is the commit being tested by Azure, which is indeed the result of merging the branch with the latest commit on main (at the time of starting that one job, so — at the moment — potentially a different commit per platform).

garyverhaegen-da (approved):
Thanks!

PR #11116, opened 2021-10-04
releases: check component status page

garyverhaegen-da:
The component status page has been neglected as of late. Craig
suggested adding a step to the release checklist.

CHANGELOG_BEGIN
CHANGELOG_END

### DIFF
--- a/release/RELEASE.md
+++ b/release/RELEASE.md
@@ -17,6 +17,10 @@ patches we backport to the 1.0 release branch).
    that changes the `LATEST` file to remove the `-snapshot` suffix on the
    corresponding snapshot, and add the `Standard-Change` label.
 
+1. **[RC]** Before blessing a snapshot as a release candidate, make sure that
+   the [Component Status](../docs/source/support/component-statuses.rst) page
+   is up to date.
+
 1. **[SNAPSHOT]** For most snapshot releases, the PR is created automatically.
    Double-check the snapshot version: it may need incrementing. Ask on Slack
    (`#team-daml`) if you're not sure.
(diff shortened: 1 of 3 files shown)


release/RELEASE.md
```diff
@@ -17,6 +17,10 @@ patches we backport to the 1.0 release branch).
    that changes the `LATEST` file to remove the `-snapshot` suffix on the
    corresponding snapshot, and add the `Standard-Change` label.
 
+1. **[RC]** Before blessing a snapshot as a release candidate, make sure that
+   the [Component Status](../docs/source/support/component-statuses.rst) page
+   is up to date.
```
stefanobaghino-da, line 22: I guess this step is only for "people in the know", but otherwise maybe it could make sense to document "how".

stefanobaghino-da (dismissed):

release/RELEASE.md
garyverhaegen-da: I actually have no idea about the "how" here, so I'm very open to suggestions.

Ideally, changing the component status page should _be_ the process by which we change the status of components, but that doesn't seem to currently be the case. Do you know what the process currently is?

stefanobaghino-da: I'm not sure about changing the status of components, I was wondering more of how one would make sure that the component status page is up to date at release time.

garyverhaegen-da: > I'm not sure about changing the status of components, I was wondering more of how one would make sure that the component status page is up to date at release time.

Well, how to check for component status really depends on what the process is for changing component status. If we're not using this page as a source of truth, I would assume we're using some spreadsheet somewhere? Is that information currently just stored in @bame-da's head?

If there is no way for the person in charge of the release to check for this, we should not add it as a step.

PR #11119, opened 2021-10-04
Add deprecation output for sandbox classic

rautenrieth-da:
This PR adds a log line that reminds users that sandbox-classic in persistence mode is deprecated. The [CLI](https://github.com/digital-asset/daml/blob/main/ledger/sandbox-classic/src/main/scala/platform/sandbox/Cli.scala#L39) already describes persistence mode as being deprecated and points users to Daml Driver for Postgres. 

# Context

- Sandbox classic on H2 with append-only schema enabled is very slow. User experience would be bad in this mode.
- Conformance tests on H2 often time out, or take a very long time to finish. We want to improve CI speed/stability by only running a minimal set of tests on H2.
- We would be still reasonably sure sandbox-classic works on H2, but we want to discourage users from using it. And since sandbox-classic on Postgres is deprecated as well(?), we point all users to Daml Driver for Postgres if they try to use sandbox-classic with a database backend.

### DIFF
--- a/ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxMain.scala
+++ b/ledger/sandbox-classic/src/main/scala/platform/sandbox/SandboxMain.scala
@@ -5,15 +5,28 @@ package com.daml.platform.sandbox
 
 import com.daml.ledger.resources.ResourceContext
 import com.daml.cliopts.GlobalLogLevel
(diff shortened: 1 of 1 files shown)


meiersi-da (approved):
Good solution @rautenrieth-da ! Let's merge it.

PR #11122, opened 2021-10-05
Fix speedy performance benchmark

nickchapman-da:
Reinstate use of `cachedCommit` in the `CollectAuthority` benchmark, so only speedy execution gets measured (as intended!)

- This has the effect of speeding up the benchmark run by a factor of about x25
- so prior to this change, only about 4% of the measured time was actually in speedy code


nickchapman-da:
On reflection, I think this is worse.

nickchapman-da:
With this change we only measure the speedy performance of the single off-ledger machine
And ignore the 100 on-ledger machines.

PR #11123, opened 2021-10-05
Check proto compatiblity with target head for PRs [KVL-1110]

nicu-da:
- Add extra job that is run during PRs which checks the protobufs, using buf, against the target branch set by the PR

### Pull Request Checklist

trigger the build.


ci/prs.yml
```diff
@@ -159,6 +169,6 @@ jobs:
 
         user=$(user_slack_handle $(branch_sha))
         if [ "$user" != "" ]; then
-            tell_slack "<@${user}> <https://dev.azure.com/digitalasset/daml/_build/results?buildId=$(Build.BuildId)|Build $(Build.BuildId)> for <https://github.com/digital-asset/daml/pull/$(pr.num)|PR $(pr.num)> has completed with status $(build_status) (changelog: $(changelog_status))." \
+            tell_slack "<@${user}> <https://dev.azure.com/digitalasset/daml/_build/results?buildId=$(Build.BuildId)|Build $(Build.BuildId)> for <https://github.com/digital-asset/daml/pull/$(pr.num)|PR $(pr.num)> has completed with status $(build_status) (changelog: $(changelog_status), proto compat: $(check_compat_proto))." \
```
garyverhaegen-da, line 172: I'm not sure we want this message to get longer. What is the value you expect to get from adding this here?

ci/check-protobuf-stability.sh
```diff
@@ -3,6 +3,17 @@
...
+# The `SYSTEM_PULLREQUEST_TARGETBRANCH` environment variable is defined by
+# Azure Pipelines; in order to run this script locally, define it beforehand
+# as the branch being targeted. For example:
+#
+#   SYSTEM_PULLREQUEST_TARGETBRANCH=main bash -x ci/check-protobuf-stability.sh
+#
+TARGET="${SYSTEM_PULLREQUEST_TARGETBRANCH:-main}"
+echo "The target branch is '${TARGET}'."
+
+BUF_TAG_TO_CHECK=""
```
garyverhaegen-da, line 15: I don't think this line does anything: the value is overridden through all code paths anyway, unless I'm missing something.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +45,52 @@ function check_non_lf_protos() {
...
-
-}
-
+  BUF_TAG_TO_CHECK=".git#tag=${LATEST_STABLE_TAG}"
+  ;;
+--head)
+  # Check against the head of the target branch.
+  #
+  # This check ensures that backwards compatibility is never broken on the target branch,
+  # which is stricter than guaranteeing compatibility between release tags.
```
garyverhaegen-da, line 86: My understanding, from my discussions with @fabiotudone-da, was that we specifically wanted to be able to break compatibility in-between releases (which I think is a bad idea, but not my call). With this PR, that would no longer be possible. If we're checking that PRs never break compatibility with their parent, and we assume not-breaking-backwards-compatibility is a transitive property (which the comment here strongly implies), then what do we still gain by ever comparing to tags at all?

If we're willing to have this stricter policy, then we can simplify this a bit and just always check against all parents, on both PR builds and main builds. (This will do what you want because Azure builds merge commits, so the target branch is one of the parents.)

You can list the parents of the current commit with
```
git log --pretty=%P -1
```

ci/prs.yml
```diff
@@ -73,6 +73,14 @@ jobs:
             exit 1
         fi
 
+- job: check_compatibility_protobuf
```
garyverhaegen-da, line 76: Note that the job is already running on PRs, because `build.yml` is common to both PR builds and main builds. Therefore, as it stands, this would run it twice on PRs.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +45,52 @@ function check_non_lf_protos() {
...
+  echo "Using head target"
+  BUF_TAG_TO_CHECK="https://github.com/digital-asset/daml.git#branch=${TARGET}"
```
garyverhaegen-da, line 88: Not supe important but I'd prefer if the name of the variable (and the message when we print it) did not claim it always contain tags, when in fact it contains buf references to either tags or branches.

ci/prs.yml
nicu-da: Indeed, not really helpful. Will remove it.

ci/check-protobuf-stability.sh
fabiotudone-da: > My understanding, from my discussions with @fabiotudone-da, was that we specifically wanted to be able to break compatibility in-between releases (which I think is a bad idea, but not my call). With this PR, that would no longer be possible. If we're checking that PRs never break compatibility with their parent, and we assume not-breaking-backwards-compatibility is a transitive property (which the comment here strongly implies), then what do we still gain by ever comparing to tags at all?

An upcoming change will introduce a check override mechanism via commit message trailers.

ci/check-protobuf-stability.sh
```diff
@@ -3,6 +3,21 @@
...
+TARGET="${SYSTEM_PULLREQUEST_TARGETBRANCH:-main}"
+echo "The target branch is '${TARGET}'."
+
+BUF_TAG_TO_CHECK=""
+# The `buf.yml` file was split into multiple files after v1.17
+# This flag indicates if the against target we use for buf is before or after the split
+# Will be removed once we have a stable tag released after the split point
+BUF_CONFIG_UPDATED=true
+
 # LF protobufs are checked against local snapshosts.
```
fabiotudone-da, line 21:
```suggestion
# LF protobufs are checked against local snapshots.
```

ci/check-protobuf-stability.sh
nicu-da: The idea is to add an exception to this check, working on that as soon as I'm done with this story (KVL-1111).
So we would have a way of going around this check.
AFAIK there is no way of skipping the stable check, so while this check is stricter and it would basically invalidate the stable check, we still want to run the stable check to close the loophole where we explicitly skip this check.

ci/check-protobuf-stability.sh
nicu-da: Yes, it is overridden but it's declared here just for visibility. Do you prefer we don't declare the vars beforehand?

ci/check-protobuf-stability.sh
nicu-da: Good point, renaming this one

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
-  # stable tag.
-  #
-  # For PRs targeting release branches, we should really check against
-  # all the most recent stable tags reachable from either the current branch or
-  # from previous release branches (say, against both `1.17.1` and `1.16.2`
-  # created after the `release/1.17.x` branch).
-  # Instead, we approximate by checking only against the most recent stable tag
-  # reachable from the current branch, under the assumption that if a lesser
-  # release branch contains a protobuf change, then it will also be present in
-  # higher ones either through a shared commit or a back-port from `main`.
```
fabiotudone-da, line 57: Please don't remove this explanation.

ci/check-protobuf-stability.sh
garyverhaegen-da: I usually don't; the whole point of `set -u` is to detect if we forget to set a var before using it, and this seems like it foils that: we'll happily run with an empty string instead of properly failing.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
-
   readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
-  echo "Checking protobufs against tag '${LATEST_STABLE_TAG}'"
-  for buf_module in "${BUF_MODULES_AGAINST_STABLE[@]}"; do
-    (eval "$(dev-env/bin/dade assist)" ; buf breaking --config "${buf_module}" --against ".git#tag=${LATEST_STABLE_TAG}")
-  done
-
-}
-
+  # For the latest stable snapshot the config is still the default "buf.yaml"
```
fabiotudone-da, line 92: Do you mean with "stable release"? Will this require a further change after 1.18? Isn't it better to check for the specific release so that we just have to remove legacy code, rather than fixing the build?

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
+
+Options:
+  -h, --help: shows this help
+  --stable:     check against the latest stable tag (default)
+  --head:       check against the head commit
+USAGE
+  exit
+  ;;
+--stable)
+  echo "Using stable target"
```
fabiotudone-da, line 80:
```suggestion
  echo "Checking protobuf against the latest stable tag"
```

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
+
+case "${1:---stable}" in
+-h | --help)
+  cat <<USAGE
+Usage: ./check-protobuf-stability.sh [options]
+
+Options:
+  -h, --help: shows this help
+  --stable:     check against the latest stable tag (default)
+  --head:       check against the head commit
```
fabiotudone-da, line 75:
```suggestion
  --target:       check against the tip of the target branch
```

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
+  # For the latest stable snapshot the config is still the default "buf.yaml"
+  BUF_CONFIG_UPDATED=false
+  BUF_TAG_TO_CHECK=".git#tag=${LATEST_STABLE_TAG}"
+  ;;
+--head)
+  # Check against the head of the target branch.
+  #
+  # This check ensures that backwards compatibility is never broken on the target branch,
+  # which is stricter than guaranteeing compatibility between release tags.
+  echo "Using head target"
```
fabiotudone-da, line 101:
```suggestion
  echo "Checking protobuf against the tip of the '${TARGET}' branch"
```

ci/check-protobuf-stability.sh
fabiotudone-da: But I suggest to print it later and include the actual tag.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +49,65 @@ function check_non_lf_protos() {
...
-    (eval "$(dev-env/bin/dade assist)" ; buf breaking --config "${buf_module}" --against ".git#tag=${LATEST_STABLE_TAG}")
-  done
-
-}
-
+  # For the latest stable snapshot the config is still the default "buf.yaml"
+  BUF_CONFIG_UPDATED=false
+  BUF_TAG_TO_CHECK=".git#tag=${LATEST_STABLE_TAG}"
+  ;;
+--head)
```
fabiotudone-da, line 96:
```suggestion
--target)
```

fabiotudone-da (changes requested):

ci/prs.yml
```diff
@@ -141,6 +149,7 @@ jobs:
     - collect_build_data
     - check_for_release
     - check_changelog_entry
+    - check_compatibility_protobuf
```
garyverhaegen-da, line 152:
```suggestion
    - compatibility_head_protobuf
```

Azure will not start the build if the name here doesn't match the job name.

ci/check-protobuf-stability.sh
nicu-da: I think this adds little value. 

>    For `main` and PRs targeting `main`, we simply check against the most recent stable tag.

This actually describes the ci jobs that we have, so I would much rather add a comment referencing the job definition directly, and add any extra explanations to the job definitions themselves. If not, this will be out of sync sooner or later, as the jobs can change without anyone updating this comment.

>    For PRs targeting release branches, we should really check against
   all the most recent stable tags reachable from either the current branch or
   from previous release branches (say, against both `1.17.1` and `1.16.2`
   created after the `release/1.17.x` branch).
   
This sounds like a TODO which we either track in a ticket and reference the ticket number if we want to do that change, or not add this comment as we will not be doing the change so it makes no sense in mentioning it. Also, my understanding is that checking against older versions than the latest stable doesn't really make sense anyway, as it's transitively compatible. 

>    Instead, we approximate by checking only against the most recent stable tag
   reachable from the current branch, under the assumption that if a lesser
   release branch contains a protobuf change, then it will also be present in
   higher ones either through a shared commit or a back-port from `main`.
   
I would simplify this section and keep it as an explanation for the stable check:

`Check against the most recent stable (non-snapshot) tag, under the assumption that any protobuf changes which are subsequently added to prior releases are also included in the latest stable release tag.`

ci/check-protobuf-stability.sh
nicu-da: No, I misworded this one. It should've been actually `latest stable tag`

ci/check-protobuf-stability.sh
fabiotudone-da: I would like the help to make understandable what this script does to everyone, without assuming context or knowledge of its implementation. My understanding of what `--stable` does matches its current description but `--head` is not clear to me as `HEAD` in `git` references the current commit, while instead it checks against the tip of the target branch with that option. Am I missing/misunderstanding something?

ci/check-protobuf-stability.sh
nicu-da: @fabiotudone-da I now understand why that is not really a TODO. But for me, it sounds as though it tries to convey the same meaning as the last message (specifically the second part of the message, which clarifies our assumption that any protobuf changes will be ported to the latest stable tag). 

```
Cherck against the most recent stable (non-snapshot) tag reachable from
the current branch, under the assumption that if a lesser release branch
contains a protobuf change, then it will also be present in higher ones.
```

This sounds good to me, the only thing I would like to double-check is the wording using "lesser release branch" and "higher ones" (I haven't really encountered these quantifiers before when mentioning releases). I assume we mean "previous release branches" and "following releases". Do you think the lesser/higher notation is clearer?

fabiotudone-da: I avoided using "previous" and "following" because they tend to imply time rather than version ordering but I guess it's understandable anyway.

ci/check-protobuf-stability.sh
nicu-da: Good point. I will add a story to remove the check after the 1.18 release, and will update the code to check the release version

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +51,76 @@ function check_non_lf_protos() {
...
+  # This check does not need to run on release branch commits because
   # they are built sequentially, so no conflicts are possible and the per-PR
   # check is enough.
+  readonly RELEASE_BRANCH_REGEX="^release/.*"
   GIT_TAG_SCOPE=""
+  # For PRs targeting release branches, we should really check against
+  # all the most recent stable tags reachable from either the current branch or
+  # from previous release branches (say, against both 1.17.1 and 1.16.2
+  # created after the release/1.17.x branch).
+  #
```
fabiotudone-da, line 92: If you decide to leave this, then the next paragraph's start should be adjusted to something like "We rather check".

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +51,76 @@ function check_non_lf_protos() {
...
-
   readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
-  echo "Checking protobufs against tag '${LATEST_STABLE_TAG}'"
-  for buf_module in "${BUF_MODULES_AGAINST_STABLE[@]}"; do
-    (eval "$(dev-env/bin/dade assist)" ; buf breaking --config "${buf_module}" --against ".git#tag=${LATEST_STABLE_TAG}")
-  done
-
-}
+  # The current stable release is v1.17 and it includes the buf config file with the default name `buf.yml`.
+  # Once we have the v1.18 release we can remove this check for the updated config (KVL-1131)
```
fabiotudone-da, line 101:
```suggestion
  # v1.17 includes the buf config file with the default name `buf.yml`.
```
Time references in code become stale very fast.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +51,76 @@ function check_non_lf_protos() {
...
-  #
-  #   SYSTEM_PULLREQUEST_TARGETBRANCH=main bash -x ci/check-protobuf-stability.sh
-  #
-  TARGET="${SYSTEM_PULLREQUEST_TARGETBRANCH:-main}"
-  echo "The target branch is '${TARGET}'."
+  echo "Checking protobufs against git target '${BUF_GIT_TARGET_TO_CHECK}'"
+  for buf_module in "${BUF_MODULES_AGAINST_STABLE[@]}"; do
+    # Starting with version 1.17 we split the default `buf.yaml` file into multiple config files
+    # This in turns requires that we pass the `--against-config` flag for any check that is run on versions > 1.17
+    if [[ $BUF_CONFIG_UPDATED = true ]]
```
fabiotudone-da, line 58: I think `= true` doesn't add to understandability.

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +51,76 @@ function check_non_lf_protos() {
...
+  ;;
+--target)
+  # Check against the tip of the target branch.
+  #
+  # This check ensures that backwards compatibility is never broken on the target branch,
+  # which is stricter than guaranteeing compatibility between release tags.
+  #
+  # The files are always split for versions > 1.17, and there is no way of opening a PR against a target <= 1.17 which includes this check
+  BUF_CONFIG_UPDATED=true
+  BUF_GIT_TARGET_TO_CHECK=".git#branch=origin/${TARGET}"
```
fabiotudone-da, line 117: Why is it important to refer to `origin`?

nicu-da: On CI we don't have any local branch, so we need to include the remote as well for buf to understand what we mean.

ci/check-protobuf-stability.sh
nicu-da: How do you mean by time references? Do you refer to the version? 
In this case, the version is actually absolute, I don't see a way how this comment can become stale, as we would only be removing it when we advance to the 1.18 release.

nicu-da: Updated the comment, I hope it's clearer now. 👍 for your points

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +50,75 @@ function check_non_lf_protos() {
...
+  readonly RELEASE_BRANCH_REGEX="^release/.*"
   GIT_TAG_SCOPE=""
+  # For PRs targeting release branches, we should really check against
+  # all the most recent stable tags reachable from either the current branch or
+  # from previous release branches (say, against both 1.17.1 and 1.16.2
+  # created after the release/1.17.x branch).
+  #
+  # We rather check against the most recent stable (non-snapshot) tag reachable from the current branch,
+  # under the assumption that if a previous release branch contains a protobuf change,
+  # then it will also be present in former ones (previous/former with regards to versioning).
```
fabiotudone-da, line 93:
```suggestion
  # then it will also be present in later ones (previous/later with regards to versioning).
```

ci/check-protobuf-stability.sh
```diff
@@ -35,42 +50,75 @@ function check_non_lf_protos() {
...
-
   readonly LATEST_STABLE_TAG="$(git tag ${GIT_TAG_SCOPE} | grep -v "snapshot" | sort -V | tail -1)"
-  echo "Checking protobufs against tag '${LATEST_STABLE_TAG}'"
-  for buf_module in "${BUF_MODULES_AGAINST_STABLE[@]}"; do
-    (eval "$(dev-env/bin/dade assist)" ; buf breaking --config "${buf_module}" --against ".git#tag=${LATEST_STABLE_TAG}")
-  done
-
-}
+  # The v1.17 stable release includes the buf config file with the default name `buf.yml`.
+  # Starting with the v1.18 release we have multiple buf config files and this check can be removed. (KVL-1131)
```
fabiotudone-da, line 99: Nit (this TODO is tracked):

```suggestion
  # Starting with v1.18 we have multiple buf config files. (KVL-1131)
```

ci/check-protobuf-stability.sh
fabiotudone-da: This seems not having been addressed yet, @nicu-da do you agree with @garyverhaegen-da (I do)?

nicu-da: Done

nicu-da:
> LGTM, thanks for pushing this forward!

@fabiotudone-da thank you for the fast and thorough reviews!

ci/check-protobuf-stability.sh
```diff
@@ -3,7 +3,18 @@
 # SPDX-License-Identifier: Apache-2.0
 set -euo pipefail
 
-# LF protobufs are checked against local snapshosts.
+eval "$(dev-env/bin/dade assist)"
```
remyhaemmerle-da, line 6: Since you added this globally, you may drop the one at line 20/31, 
meaning replace 
```
  (eval "$(dev-env/bin/dade assist)"; buf breaking --config "buf-lf-transaction.yaml" --against ${stable_snapshot_dir})
```
by 
```
  buf breaking --config "buf-lf-transaction.yaml" --against ${stable_snapshot_dir}
```

fabiotudone-da: Good point, I can remember we had trouble with enabling this globally, I'll try to find the relevant discussion.

fabiotudone-da: https://github.com/digital-asset/daml/pull/10950#pullrequestreview-760652273

garyverhaegen-da: You should not enable it globally if you plan to checkout old commits, in which dev-env might have changed.

fabiotudone-da (changes requested):

PR #11147, opened 2021-10-06
Data dependencies support for reexports

akrmn:
### Pull Request Checklist

trigger the build.


compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -488,7 +488,14 @@ convertModule lfVersion pkgMap stablePackages isGenerated file x depOrphanModule
     templates <- convertTemplateDefs env
     exceptions <- convertExceptionDefs env
     interfaces <- convertInterfaces env (eltsUFM (cm_types x))
-    pure (LF.moduleFromDefinitions lfModName (Just $ fromNormalizedFilePath file) flags (types ++ templates ++ exceptions ++ definitions ++ interfaces ++ depOrphanModules))
+    let defs =
+            types
```
akrmn, line 492: I got a bit tired of the super long line, so I factored this out. Please let me know if this is the style you'd prefer for splitting the expression into lines, or if you prefer operators at the end or something different.

akrmn:
That silenced some errors (and some warnings), but I'm still getting quite a few errors such as this one:
```
File:
  .daml/package-database/1.dev/daml-stdlib-26def009ea16788210e60d5b5927068ffe8a492b441937798898a4f317bfbb5b-26def009ea16788210e60d5b5927068ffe8a492b441937798898a4f317bfbb5b/DA/Internal/RebindableSyntax.daml
Hidden:   no
Range:    37:8-37:85
Source:   not found
Severity: DsError
Message:
  Could not find module
  ‘Pkg_312f93eed15148f6b5d47ac144b47c510af030d99135277f7a890156ad068540.GHC.Real’
  Perhaps you meant
  Pkg_312f93eed15148f6b5d47ac144b47c510af030d99135277f7a890156ad068540.GHC.Base (defined via package
  flags to be GHC.Base)
  Pkg_312f93eed15148f6b5d47ac144b47c510af030d99135277f7a890156ad068540.GHC.Enum (defined via package
  flags to be GHC.Enum)
  Pkg_312f93eed15148f6b5d47ac144b47c510af030d99135277f7a890156ad068540.GHC.Err (defined via package
  flags to be GHC.Err)
Failed to compile interface for data-dependency: daml-stdlib-26def009ea16788210e60d5b5927068ffe8a492b441937798898a4f317bfbb5b
```

akrmn:
> > A few tests still fail, and they mention missing values from modules such as `DA.List` or `DA.Action.State.Class`, which really don't feel like they belong in the list above.
> 
> which test specifically is failing?
These two:
* Cross DAML-LF version: 1.6 -> 1.7
* Cross DAML-LF version: 1.7 -> 1.8

cocreature:
> EDIT: which might make sense if the missing functions from `DA.List` weren't available before DAML-LF 1.8 ?

Thanks will take a look later today or early tomorrow. Your idea sounds like it might be on the right track but I haven’t thought about it too deeply.

cocreature:
I’ve looked into it and I think I understand the failure: data-dependencies for LF < 1.8 are very limited. Specifically, typeclasses and typeclass-constrained methods are not available here (they depend on LF type synonyms which only got introduced in 1.8).

So I think there are two options:
1. Don’t emit those export markers for LF < 1.8.
2. For LF < 1.8 filter out the things that rely on typeclasses.

2 seems harder to implement and at this point nobody should be on LF < 1.8 (it’s more than 2 years old iirc) so I’d suggest to go with 1.

akrmn:
thanks for looking into it @cocreature! I'll go with #1

akrmn:
- [ ] Rebase on `main` once #11260 and #11263 are merged

akrmn:
> [An] option to avoid large expressions from blowing up could be to emit multiple export expressions at the extreme end one per export but we could also do some form of batching like 5 exports max per item or whatever a sensible number is.
>
> A large number of expressions isn’t as problematic as a single very large expression because the latter will be handled relatively poorly by some of our code.

@cocreature, how would I find a sensible number for the size of these batches? If that's too much trouble, I could go ahead with one expression per `GHC.AvailInfo`. That leads me to a design question, should these expressions simply be named `$exports_<i>` or do you have something else in mind?

PR #11152, opened 2021-10-06
LF: clean daml-lf/language module

remyhaemmerle-da:
- enable NonUnitStatements, and find a useless statement.
- add missing `final`s,
- replace `{ ... }` by `( ... )` for one statement anonymous function
- drop useless `new` for case class.
- replace `sealed class X { ... object Y extend X` by
  `class X private[Z] { ... val Y = new X` to avoid unecessary class
  alocation.
- move invariant check inside the body of some case class instead of
  the companion object.
- other cosmetic changes.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
@@ -12,7 +12,7 @@ object Ast {
   //
 
   /** Fully applied type constructor. */
-  case class TypeConApp(tycon: TypeConName, args: ImmArray[Type]) {
+  final case class TypeConApp(tycon: TypeConName, args: ImmArray[Type]) {
     def pretty: String =
       args.foldLeft(TTyCon(tycon): Type) { case (arg, acc) => TApp(acc, arg) }.pretty
   }
@@ -33,7 +33,7 @@ object Ast {
   type EnumConName = Name
 
   /* Binding in a let/update/scenario block. */
-  case class Binding(binder: Option[ExprVarName], typ: Type, bound: Expr)
+  final case class Binding(binder: Option[ExprVarName], typ: Type, bound: Expr)
 
   //
   // Expressions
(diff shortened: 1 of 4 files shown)


daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
```diff
@@ -686,9 +681,9 @@ object Ast {
       observers: E, // Observers of the contract.
       key: Option[GenTemplateKey[E]],
       implements: Map[TypeConName, GenTemplateImplements[E]],
-  ) extends NoCopy
```
remyhaemmerle-da, line 689: There is no special invariant for this case class, so no reason to have a private constructor and prevent copy method.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
```diff
@@ -926,14 +920,19 @@ object Ast {
...
       exceptions: Map[DottedName, GenDefException[E]],
       interfaces: Map[DottedName, DefInterface],
       featureFlags: FeatureFlags,
-  ) extends NoCopy
+  ) {
+    templates.keysIterator.foreach(name =>
+      if (exceptions.keySet.contains(name))
+        throw PackageError(s"Collision between exception and template name ${name.toString}")
+    )
+  }
```
remyhaemmerle-da, line 935: Once putting the invariant check inside `GenModule`, we do not need the constructor to be private

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
```diff
@@ -612,7 +610,6 @@ object Ast {
   final case class DataVariant(variants: ImmArray[(VariantConName, Type)]) extends DataCons {
     lazy val constructorInfo: Map[VariantConName, (Type, Int)] =
       variants.iterator.zipWithIndex.map { case ((cons, typ), rank) => (cons, (typ, rank)) }.toMap
-    variants.iterator.zipWithIndex.map { case ((cons, typ), rank) => (cons, (typ, rank)) }.toMap
```
remyhaemmerle-da, line 615: Thanks to NonUnitStatements,

PR #11166, opened 2021-10-07
Explore testing of Daml evaluation order

nickchapman-da:
As a precursor to improving the test coverage for Daml's evaluation order:
- Change speedy's function argument evaluation order. Was: left->right. Now: right->left. 
- What tests fail?
- Think just 1 test fails.  Adapt it here.


paulbrauner-da:
Implemented in https://github.com/digital-asset/daml/pull/17253

PR #11168, opened 2021-10-07
timestamp-conversions: A set of libraries for converting timestamps and durations.

SamirTalwar-DA:
I got fed up of writing converters between various different timestamp and duration types (and not knowing if I was inadvertently introducing bugs in doing so), so I decided to make libraries for all of them.

These provide implicit converters between:

  - _java-scala_: Java durations and Scala durations
  - _java-lf_: Java instants and Daml-LF timestamps
  - _java-protobuf_:
    - Java instants and Java Protocol Buffers timestamps
    - Java durations and Java Protocol Buffers durations
  - _java-scalaprotobuf_:
    - Java instants and Scala Protocol Buffers timestamps
    - Java durations and Scala Protocol Buffers durations
  - _scala-protobuf_: Scala durations and Scala Protocol Buffers durations
  - _scala-javaprotobuf_: Scala durations and Java Protocol Buffers durations
  - _lf-javaprotobuf_: Daml-LF timestamps and Java Protocol Buffers timestamps
  - _lf-scalaprotobuf_: Daml-LF timestamps and Scala Protocol Buffers timestamps

There are many small libraries so you can depend on what you need without bringing in extra dependencies. For example, if you don't use ScalaPB, you don't want to include it inadvertently.

They are currently not used, but I have plans to replace existing conversion functions with these.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/libs-scala/timestamp-conversions/java-lf/BUILD.bazel
+++ b/libs-scala/timestamp-conversions/java-lf/BUILD.bazel
@@ -1,0 +1,43 @@
+# Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+load(
+    "//bazel_tools:scala.bzl",
+    "da_scala_library",
+    "da_scala_test_suite",
+    "lf_scalacopts_stricter",
+)
+
+scalacopts = lf_scalacopts_stricter
+
+da_scala_library(
+    name = "java-lf",
+    srcs = glob(["src/main/scala/**/*.scala"]),
+    scalacopts = scalacopts,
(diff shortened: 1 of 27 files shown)


S11001001:
There's something I didn't notice before that I think we should really avoid here: we generally are very explicit about our existing conversions being total or not, hence the `assert` names and their well-typed counterparts. By hiding that, this library would also be hiding that distinction.

Following the `TConv` design idea we can solve that problem by declaring it as `TConv[-From, To] {type OF[A]}` and returning `OF[To]` in the various methods. Regardless, the existing scheme of merely flattening this distinction makes these operations look _way_ more innocuous than they actually are.

nicu-da:
@SamirTalwar-DA do we want to proceed with this PR?

SamirTalwar-DA:
I lost interest. Let's close it and re-open it if anyone wants to pick it up.

PR #11178, opened 2021-10-08
Update scalafmt to 3.0.6

nicu-da:
This will allow us to introduce https://scalameta.org/scalafmt/docs/configuration.html#newlinestoplevelstatementblanklines

CHANGELOG_BEGIN

CHANGELOG_END

### Pull Request Checklist

trigger the build.


stefanobaghino-da:
> > This will allow us to introduce https://scalameta.org/scalafmt/docs/configuration.html#newlinestoplevelstatementblanklines
> 
> Please keep in mind that introducing a new formatting change has the potential to introduce conflicts on _all_ open PRs, potentially causing puzzling moments for the people working on the codebase.
> 
> My advice is to _not_ do this, unless there is a _very_ good reason why we need to adopt this.

If you decide to go for it anyway, please let other contributors know about this and collect their feedback.

stefanobaghino-da:
> The value would be in not wasting time fixing minor code formatting styles through manual review/changes. I would expect that our code style is consistently enforced and does not depend on "magic" settings. Eg: this PR [#11174 (comment)](https://github.com/digital-asset/daml/pull/11174#discussion_r724876488), where there are multiple occurrences where IJ decided to drop the newlines when reformating imports, but scalafmt does not enforce that.

It's a lot of noise, I agree. Is there some way in which we can somehow disable IJ's import rearrangement with a committed config file? This way this would stop automatically for all IJ users on the codebase without any interruption.

garyverhaegen-da:
> Ah ok, was not aware of that. From the wording, it leads me to think that it represents just a normal code change and not some "special" change.

The naming is a bit weird here, but code changes are considered "Routine Changes" whereas changes to production systems (in the case of this repo, mostly making releases and changing the CI infrastructure) are considered "Standard" or "Normal" changes. "Standard" changes are the ones that require following the change management process, including tracking and monthly review; "Routine" changes are the ones for which we can have a more informal process where it's enough for two people to agree (in general, PR opener + 1 reviewer).

Search for "Change Management" on Confluence for details.

garyverhaegen-da:
(They're "Normal" changes to production system as opposed to the third type of change, "Emergency Change", usually done in response to an incident and possibly bypassing the normal review process — but nevertheless tracked for after-the-fact review.)

PR #11370, opened 2021-10-25
[Trigger-Service] Add support for read as in the trigger service

### LINKED ISSUE #11309: Support multi-party readAs in trigger service
trigger-service part of #7640

realvictorprm:
Fixes #11309

### Pull Request Checklist

trigger the build.


stefanobaghino-da:
This has been left open without any further progress. We should try to close it as soon as possible.

stefanobaghino-da:
@akshayshirahatti-da Can yo review this, please?

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V5__Add_read_as.sql
```diff
@@ -0,0 +1,5 @@
+-- Copyright (c) 2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Add application_id to running trigger table defaulting to trigger_instance
+alter table ${table.prefix}running_triggers add (read_as nvarchar2(2000));
```
S11001001, line 5: one of these things is not like the other

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/postgres/V5__Add_read_as.sql
```diff
@@ -0,0 +1,5 @@
+-- Copyright (c) 2020 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Add application_id to running trigger table defaulting to trigger_instance
+alter table ${table.prefix}running_triggers add column read_as text;
```
S11001001, line 5: likewise

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
```diff
@@ -38,6 +39,15 @@ abstract class DbTriggerDao protected (
   protected implicit def uuidPut: Put[UUID]
   protected implicit def uuidGet: Get[UUID]
 
+  implicit val readAsPut: Put[Set[Party]] =
+    implicitly[Put[String]].contramap { (readAs: Set[Party]) =>
+      readAs.map(Tag.unwrap).mkString(",")
+    }
+
+  implicit val readAsGet: Get[Set[Party]] = implicitly[Get[String]].map {
+    _.split(",").map(_.trim).toSet.filter(_.nonEmpty).map(Party.apply)
```
S11001001, line 48: Using `%` instead of `,` might be safer as it is explicitly reserved from being an identifier character by the LF spec.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
```diff
@@ -38,6 +39,15 @@ abstract class DbTriggerDao protected (
   protected implicit def uuidPut: Put[UUID]
   protected implicit def uuidGet: Get[UUID]
 
+  implicit val readAsPut: Put[Set[Party]] =
+    implicitly[Put[String]].contramap { (readAs: Set[Party]) =>
+      readAs.map(Tag.unwrap).mkString(",")
+    }
```
S11001001, line 45:
```suggestion
  implicit val readAsPut: Put[Set[Party]] = {
    type F[A] = Put[Set[A]]
    Party.subst[F, String](implicitly[Put[String]].contramap {
      _.mkString(",")
    })
  }
```

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
```diff
@@ -38,6 +39,15 @@ abstract class DbTriggerDao protected (
...
+
+  implicit val readAsGet: Get[Set[Party]] = implicitly[Get[String]].map {
+    _.split(",").map(_.trim).toSet.filter(_.nonEmpty).map(Party.apply)
+  }
```
S11001001, line 49:
```suggestion
  implicit val readAsGet: Get[Set[Party]] = {
    type F[A] = Get[Set[A]]
    Party.subst[F, String](implicitly[Get[String]].map {
      _.split(",").map(_.trim).toSet.filter(_.nonEmpty)
    }
  }
```

triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceTest.scala
```diff
@@ -94,14 +94,19 @@ trait AbstractTriggerServiceTest
       triggerName: String,
       party: Party,
       applicationId: Option[ApplicationId] = None,
+      readAs: Set[Party] = Set(),
   ): Future[HttpResponse] = {
+    import Request.PartyFormat
+    val readAsContent =
+      if (readAs.isEmpty) "null"
+      else spray.json.DefaultJsonProtocol.listFormat.write(readAs.toList).toString
```
S11001001, line 102:
```suggestion
      else {
        import spray.json.DefaultJsonProtocol._
        readAs.toJson.compactPrint
      }
```

S11001001 (approved):
Nice and clean.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
realvictorprm: didn't know about that, thanks!

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V5__Add_read_as.sql
stefanobaghino-da: I thought I left a comment here but apparently I haven't: what happens if there is a running trigger and the trigger service undergoes an upgrade on the following restart? Will this additional column be nullable, thus meaning that new rows will have a `null` value for `read_as`? (Which I assume works fine, as `null` seems to be used for empty `read_as` from what I can tell reading at the code)

stefanobaghino-da (commented):
In principle LGTM but please have a second look at my migration question.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
S11001001: Another note, `PartyIdString` allows spaces. So `trim` is not OK here.

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/oracle/V5__Add_read_as.sql
S11001001: `null = ''` on Oracle; `not null` rejects empty strings, which is not desirable here. In other words the default is correct for Oracle.

triggers/service/src/main/resources/com/daml/lf/engine/trigger/db/migration/postgres/V5__Add_read_as.sql
S11001001: However, _this_ one should be `not null` with a default `''`'.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
realvictorprm: Is there a specific reason why this code is better than the other?

stefanobaghino-da:
@realvictorprm You already have an approval, merge it.

triggers/service/src/main/scala/com/digitalasset/daml/lf/engine/trigger/dao/DbTriggerDao.scala
S11001001: This does subst once for the whole program, the original code does subst once for each party that is read every time it is read.

realvictorprm: Nice, thanks for the info!

PR #11392, opened 2021-10-26
json indexes with less regular syncronisation

dasormeter:
during the performance tuning of canton on oracle we did several experiments with json indexes on the json array fields within the append-only schema.  

the broad conclusion was 
1) adding the index did not lead to an appreciable throughput gain 
2) enabling the indexes massively slowed down the time to prune tables and 
3) enabling the indexes increases insert time for new rows

we therefore opted not to enable the json indexes

this draft pr enables the indexes with a tweak to their update frequency, so that they are not updated upon every row insert but rather at a periodic interval, which should help insert time.  however this won’t solve the problem with very low prune time due to the cleanup required when rows referenced by the indexes are removed.

An additional permission is required to the oracle db service account in use for the schema
```stmt.execute(s"""GRANT CREATE JOB TO $name""")```

This permission is required to be able to alter the timing as the json index update becomes a scheduled job rather than the index updating on each row insert, otherwise you will receive the following error

```
  Cause: oracle.jdbc.OracleDatabaseException: ORA-29855: error occurred in the execution of ODCIINDEXCREATE routine
ORA-20000: Oracle Text error:
DRG-50857: oracle error in drvddl.submit_job
ORA-27486: insufficient privileges
ORA-06512: at "CTXSYS.DRUE", line 186
ORA-06512: at "CTXSYS.TEXTINDEXMETHODS", line 320
```


dasormeter:
during the performance tuning of canton on oracle we did several experiments with json indexes on the json array fields within the append-only schema.  

the broad conclusion was 
1) adding the index did not lead to an appreciable throughput gain 
2) enabling the indexes massively slowed down the time to prune tables and 
3) enabling the indexes increases insert time for new rows

we therefore opted not to enable the json indexes

this draft pr enables the indexes with a tweak to their update frequency, so that they are not updated upon every row insert but rather at a periodic interval, which should help insert time.  however this won’t solve the problem with very low prune time due to the cleanup required when rows referenced by the indexes are removed.

dasormeter:
Haven't touched this in over 1 year so closing it

PR #11411, opened 2021-10-26
Explicitly import Java runtime from Nix/dadew

aherrmann-da:
Defines the Java runtime in a more hermetic way, by generating `java_runtime` definitions with Nix/dadew instead of pulling the path out of `JAVA_HOME`. Ideally, this would also allow us to drop `JAVA_HOME` in the Nix Bazel wrapper. However, there are still parts of the code-base that rely on that env-var.
Configures the Java runtime using the [`--javabase`](https://docs.bazel.build/versions/main/user-manual.html#flag--javabase) and [`--host_javabase`](https://docs.bazel.build/versions/main/user-manual.html#flag--host_javabase) flags.
This also fixes an issue with `rules_scala`'s launcher on Windows, which failed when using a `--javabase` with an absolute path. See https://github.com/bazelbuild/rules_scala/pull/1313.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -61,9 +61,17 @@ build --copt -DGRPC_BAZEL_BUILD
 
 # Bazel defaults to JDK9. This makes sure we run Nix provided JDK, and
 # avoid dreaded "Unrecognized VM option 'CompactStrings'" error
-build --host_javabase=@java_home//:javabase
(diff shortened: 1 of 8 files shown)


cocreature (approved):
Very nice, thanks!

PR #11435, opened 2021-10-27
DbDispatcher handles SQL exceptions with self-service error codes [DPP-690]

tudor-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
@@ -1,0 +1,109 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao
+
+import com.daml.error.ContextualizedErrorLogger
+import com.daml.error.definitions.IndexErrors
+import io.grpc.StatusRuntimeException
+import org.h2.jdbc.JdbcSQLIntegrityConstraintViolationException
+import org.postgresql.util.PSQLException
+
+import java.sql._
+import scala.annotation.tailrec
+
+/** Wraps SQLExceptions into transient and non-transient errors.
+  * Transience classification is done as follows:
+  *
+  * * Problems that are likely to be resolved with retries are transient errors. For example, network outages or db access
+  * serialization problems.
+  *
+  * * Problems that cannot be recovered from are non-transient. For example, an illegal argument exception inside a
+  * database transaction or a unique constraint violation.
+  *
+  * TODO error codes: Move and handle the error specialization per storage backend
+  *                   in [[com.daml.platform.store.backend.StorageBackend]].
+  */
+object DatabaseSelfServiceError {
+  @tailrec
+  def apply(
+      exception: Throwable
+  )(implicit contextualizedErrorLogger: ContextualizedErrorLogger): Throwable = exception match {
+    // This frequently occurs when running with H2, because H2 does not properly implement the serializable
+    // isolation level. This causes unexpected constraint violation exceptions when running with H2 in the presence
+    // of contention. For now, we retry on these exceptions.
+    // See https://github.com/h2database/h2database/issues/2167
+    case ex: JdbcSQLIntegrityConstraintViolationException => retryable(ex)
+    case ex: SQLRecoverableException => retryable(ex)
+    case ex: SQLTransientException => retryable(ex)
+    case ex: SQLNonTransientException => nonRetryable(ex)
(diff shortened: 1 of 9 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
```diff
@@ -0,0 +1,109 @@
...
+      case "25006" => true
+      // Retry on operator intervention errors, but not on `query_canceled` and `database_dropped`
+      case state if state.startsWith("57P") && state != "57014" && state != "57P04" => true
+      case _ => false
+    }
+
+  // Oracle exceptions are handled based on the error codes
+  // https://docs.oracle.com/en/database/oracle/oracle-database/19/errmg/ORA-00000.html#GUID-27437B7F-F0C3-4F1F-9C6E-6780706FB0F6
+  private def isRetryableOracleException(ex: SQLException): Boolean =
+    ex.getErrorCode match {
```
mziolekda, line 76: Are the values returned from `getErrorCode` also visible as part of `getMessage`? Asking because I am asking myself if this valuable information will be available to us when troubleshooting.

tudor-da: I would expect so.. But the ojdbc driver doesn't seem to be public and I have to navigate through decompiled code to understand.

tudor-da: Hard to confirm from the decompiled code. But I remember seeing error codes in stack traces. Furthermore, the messages are unique and should be able to trace back to the codes (in case they are missing).

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
```diff
@@ -0,0 +1,109 @@
...
+  )(implicit contextualizedErrorLogger: ContextualizedErrorLogger): Throwable = exception match {
+    // This frequently occurs when running with H2, because H2 does not properly implement the serializable
+    // isolation level. This causes unexpected constraint violation exceptions when running with H2 in the presence
+    // of contention. For now, we retry on these exceptions.
+    // See https://github.com/h2database/h2database/issues/2167
+    case ex: JdbcSQLIntegrityConstraintViolationException => retryable(ex)
+    case ex: SQLRecoverableException => retryable(ex)
+    case ex: SQLTransientException => retryable(ex)
+    case ex: SQLNonTransientException => nonRetryable(ex)
+    case ex: PSQLException => if (isRetryablePsqlException(ex)) retryable(ex) else nonRetryable(ex)
```
nmarton-da, line 40: Normally I would suggest to push that down under StorageBackend. It was created to hide the specifics of used databases. (for example there is no isRetryableH2Exception)
Also I understand this is best effort try to solve the problem, right? I mean even in this PR there are inconsistencies like SQLNonTransientExceptions are handled as non-retry here, but retry in the initialization. Which makes sense to me: precisely it could be decided only on a case by case fashion. (needless to say: which would lead to an enormous explosion...multiplied by supported DB backends)

No need to change as it is best effort

nmarton-da (commented):
Looks a decent best effort to me.
I do not like the fact we have DB specific code outside StorageBackend. We created storage backend to hide these underneath.

Since best effort will have wholes (not sure how big) where this is incorrect, it makes me wonder what will happen on LAPI clients when they see a non-retryable error which is retryable...or would that so bad to retry a non-retryable error? :thinking:

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
mziolekda: :+1:

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/DatabaseSelfServiceError.scala
tudor-da: Thanks. Specializing exception handling behavior per StorageBackend implementation is a good point. (Already in a TODO in this file) 👍

rautenrieth-da:
> what will happen on LAPI clients when they see a non-retryable error which is retryable...or would that so bad to retry a non-retryable error

The self-service error codes design defines exactly how clients should react to individual gRPC error codes. For example, `UNAVAILABLE` is a retriable error where the request can be quickly and transparently retried in a load balancer, while `INTERNAL` should be resolved by contacting the ledger operator. 
Returning `INTERNAL` instead of `UNAVAILABLE` and vice versa wouldn't be catastrophic, but it would lead to suboptimal application behavior and more support requests. Still better than returning `INTERNAL` for everything.

PR #11437, opened 2021-10-28
[Self-service error codes] Adapt all grpc assertions to additionally take self-service error codes [DPP-680]

tudor-da:
Adapt self-service error codes in conformance tests
 * `Assertions.assertGrpcError` and `assertGrpcErrorRegex` assert self-service error codes based on feature descriptor
 * Adapted all GRPC error assertions to expect self-service error codes as well
 * Added **TODO** and exclusions list to conformance-test-self-service-error-codes in anticipation of some missing self-service error codes implementation. (To be amended in a subsequent PR)

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
@@ -53,35 +53,13 @@ object Assertions {
       expectedCode: Status.Code,
       selfServiceErrorCode: ErrorCode,
       exceptionMessageSubstring: Option[String],
-      checkDefiniteAnswerMetadata: Boolean,
-  ): Unit =
-    if (participant.features.selfServiceErrorCodes)
-      t match {
-        case statusRuntimeException: StatusRuntimeException =>
-          assertSelfServiceErrorCode(statusRuntimeException, selfServiceErrorCode)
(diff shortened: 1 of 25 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/Assertions.scala
```diff
@@ -176,19 +160,23 @@ object Assertions {
     val actualRetryabilitySeconds = actualErrorDetails
       .collectFirst { case err: ErrorDetails.RetryInfoDetail => err.retryDelayInSeconds }
 
+    if (!actualErrorId.contains(expectedErrorId))
```
mziolekda, line 163: Hmm, this was a problem already before, but what happens when expected error code is a substring of the actual?

tudor-da: That was initially an `Option`.. When changed to String, it still typechecked. Thanks, will change

mziolekda: It makes sense. Thanks

mziolekda (approved):
Looks good, thanks!

PR #11470, opened 2021-10-29
LF: drop com.daml.lf.transaction.Transaction.Transaction type alias.

remyhaemmerle-da:
After dropping type parameter from GenTransaction we do not
com.daml.lf.transaction.Transaction.Transaction type alias.  We use
instead directly com.daml.lf.transaction.VersionedTransaction, wich is
anyway more informative.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
@@ -792,14 +792,9 @@ object Transaction {
   type ActionNode = Node.GenActionNode
   type LeafNode = Node.LeafOnlyActionNode
 
-  /** (Complete) transactions, which are the result of interpreting a
-    * ledger-update. These transactions are consumed by either the
-    * scenario-interpreter or the Daml-engine code. Both of these
-    * code-paths share the computations for segregating the
-    * transaction into party-specific ledgers and for computing
-    * divulgence of contracts.
-    */
+  @deprecated("use com.daml.transaction.VersionedTransaction", since = "1.8.0")
   type Transaction = VersionedTransaction
+  @deprecated("use com.daml.transaction.VersionedTransaction", since = "1.8.0")
   val Transaction: VersionedTransaction.type = VersionedTransaction
 
   /** Transaction meta data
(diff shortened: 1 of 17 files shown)


daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/Blinding.scala.orig
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
cocreature, line 1: looks like you messed up a merge or something

cocreature: same thing in a few other places

remyhaemmerle-da: Fixed.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -792,14 +792,9 @@ object Transaction {
...
   type LeafNode = Node.LeafOnlyActionNode
 
-  /** (Complete) transactions, which are the result of interpreting a
-    * ledger-update. These transactions are consumed by either the
-    * scenario-interpreter or the Daml-engine code. Both of these
-    * code-paths share the computations for segregating the
-    * transaction into party-specific ledgers and for computing
-    * divulgence of contracts.
-    */
+  @deprecated("use com.daml.transaction.VersionedTransaction", since = "1.8.0")
```
cocreature, line 795: do you mean 1.18? Retroactively deprecating since 1.8 seems a bit weird.

cocreature (approved):
Thanks!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: Good catch. There is the same in issue in `remy-lf-nodes` branch. 
I will fix that.

PR #11475, opened 2021-10-31
Add string-interning cache primitives [DPP-702]

nmarton-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/interning/RawStringInterning.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/interning/RawStringInterning.scala
@@ -1,0 +1,40 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.interning
+
+private[interning] case class RawStringInterning(
+    map: Map[String, Int],
+    idMap: Map[Int, String],
+    lastId: Int,
+)
+
+private[interning] object RawStringInterning {
+
+  def from(
+      entries: Iterable[(Int, String)],
+      rawStringInterning: RawStringInterning = RawStringInterning(Map.empty, Map.empty, 0),
+  ): RawStringInterning =
+    if (entries.isEmpty) rawStringInterning
+    else
+      RawStringInterning(
+        map = rawStringInterning.map ++ entries.view.map(_.swap),
+        idMap = rawStringInterning.idMap ++ entries,
+        lastId = entries.view.map(_._1).max,
+      )
+
+  def newEntries(
+      strings: Iterator[String],
+      rawStringInterning: RawStringInterning,
+  ): Vector[(Int, String)] =
+    strings
+      .filterNot(rawStringInterning.map.contains)
+      .toVector
+      .distinct // TODO Iterators do not have .distinct in Scala 2.12
+      .view
+      .zipWithIndex
+      .map { case (string, index) =>
+        (index + 1 + rawStringInterning.lastId, string)
+      }
+      .toVector
+}
--- a/ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterning.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterning.scala
@@ -1,0 +1,87 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.interning
+
+import com.daml.lf.data.Ref
+
+/** The facade for all supported string-interning domains
+  */
+trait StringInterning {
+  def templateId: StringInterningDomain[Ref.Identifier]
+  def party: StringInterningDomain[Ref.Party]
+}
+
+/** Composes a StringInterningAccessor for the domain-string type and an unsafe StringInterningAccessor for raw strings
+  *
+  * @tparam T is the type of the string-related domain object which is interned
+  */
+trait StringInterningDomain[T] extends StringInterningAccessor[T] {
+  def unsafe: StringInterningAccessor[String]
+}
+
+object StringInterningDomain {
+  private[interning] def prefixing[T](
+      prefix: String,
+      prefixedAccessor: StringInterningAccessor[String],
+      to: String => T,
+      from: T => String,
+  ): StringInterningDomain[T] =
+    new StringInterningDomain[T] {
+      override val unsafe: StringInterningAccessor[String] = new StringInterningAccessor[String] {
+        override def internalize(t: String): Int = prefixedAccessor.internalize(prefix + t)
+
+        override def tryInternalize(t: String): Option[Int] =
+          prefixedAccessor.tryInternalize(prefix + t)
+
+        override def externalize(id: Int): String =
+          prefixedAccessor.externalize(id).substring(prefix.length)
+
+        override def tryExternalize(id: Int): Option[String] =
+          prefixedAccessor.tryExternalize(id).map(_.substring(prefix.length))
+      }
+
+      override def internalize(t: T): Int = unsafe.internalize(from(t))
+
+      override def tryInternalize(t: T): Option[Int] = unsafe.tryInternalize(from(t))
+
+      override def externalize(id: Int): T = to(unsafe.externalize(id))
+
+      override def tryExternalize(id: Int): Option[T] = unsafe.tryExternalize(id).map(to)
+    }
+}
+
+/** The main interface for using string-interning.
+  * Client code can use this to map between interned id-s and string-domain objects back and forth
+  *
+  * @tparam T is the type of the string-related domain object which is interned
+  */
+trait StringInterningAccessor[T] {
+
+  /** Get the interned id
+    *
+    * @param t the value
+    * @return the integer id, throws exception if id not found
+    */
+  def internalize(t: T): Int
+
+  /** Optionally get the interned id
+    * @param t the value
+    * @return some integer id, or none if not found
+    */
+  def tryInternalize(t: T): Option[Int]
+
+  /** Get the value for an id
+    *
+    * @param id integer id
+    * @return the value, throws exception if no value found
+    */
+  def externalize(id: Int): T
+
+  /** Optionally get the value for an id
+    *
+    * @param id integer id
+    * @return some value, or none if not found
+    */
+  def tryExternalize(id: Int): Option[T]
+}
--- a/ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningView.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningView.scala
@@ -1,0 +1,114 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.interning
+
+import com.daml.dec.DirectExecutionContext
+import com.daml.lf.data.Ref
+import com.daml.logging.LoggingContext
+
+import scala.concurrent.Future
+
+class DomainStringIterators(
+    val parties: Iterator[String],
+    val templateIds: Iterator[String],
+)
+
+trait InternizingStringInterningView {
+
+  /** Internize strings of different domains. The new entries are returend as prefixed entries for persistent storage.
+    *
+    * @param domainStringIterators iterators of the new entires
+    * @return If some of the entries were not part of the view: they will be added, and these will be returned as a interned-id and raw, prefixed string pairs.
+    */
+  def internize(domainStringIterators: DomainStringIterators): Iterable[(Int, String)]
+}
+
+trait UpdatingStringInterningView {
+
+  /** Update the StringInterningView from persistence
+    *
+    * @param lastStringInterningId this is the "version" of the persistent view, which from the StringInterningView can see if it is behind
+    * @return a completion Future: if the view is behind it will load the missing entries from persistence, and update the view state
+    */
+  def update(lastStringInterningId: Int)(implicit loggingContext: LoggingContext): Future[Unit]
+}
+
+/** Encapsulate the dependency to load a range of string-interning-entries from persistence
+  */
+trait LoadStringInterningEntries {
+  def apply(
+      fromExclusive: Int,
+      toInclusive: Int,
+  ): LoggingContext => Future[Iterable[(Int, String)]]
+}
+
+/** This uses the prefixed raw representation internally similar to the persistence layer.
+  * Concurrent view usage is optimized for reading:
+  * - The single, volatile reference enables non-synchronized access from all threads, accessing persistent-immutable datastructure
+  * - On the writing side it synchronizes (this usage is anyway expected) and maintains the immutable internal datastructure
+  */
+class StringInterningView(loadPrefixedEntries: LoadStringInterningEntries)
+    extends StringInterning
+    with InternizingStringInterningView
+    with UpdatingStringInterningView {
+  @volatile private var raw: RawStringInterning = RawStringInterning.from(Nil)
+
+  private def rawAccessor: StringInterningAccessor[String] = new StringInterningAccessor[String] {
+    override def internalize(t: String): Int = raw.map(t)
+    override def tryInternalize(t: String): Option[Int] = raw.map.get(t)
+    override def externalize(id: Int): String = raw.idMap(id)
+    override def tryExternalize(id: Int): Option[String] = raw.idMap.get(id)
+  }
+
+  private val TemplatePrefix = "t|"
+  private val PartyPrefix = "p|"
+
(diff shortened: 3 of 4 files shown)


meiersi-da (commented):
@nmarton-da :

I get the impression that we are mixing two concerns in this PR that can be factored better. The two concerns are:

1. Mapping between the Daml-LF specific string-like types (Party and TemplateId) and a uniform type of internable strings `(prefix, actualString)`. This mapping does not require any state and can be implemented as pure functions.
2. Maintaining the mapping from internable strings to int identifiers. That is what the interning cache and table maintain. The functions constituting that part of the code should not contain any reference to the types for `Party` or `Template`.

With this factoring of the code the caller takes care of the appropriate composition; and interning other kinds of string-like identifiers becomes a localized change.

nmarton-da:
> @nmarton-da :
> 
> I get the impression that we are mixing two concerns in this PR; and that these concerns can be factored into independent changes. The two concerns are:
> 
>     1. Mapping between the Daml-LF specific string-like types (Party and TemplateId) and a uniform type of internable strings `(prefix, actualString)`. This mapping does not require any state and can be implemented as pure functions.
> 
>     2. Maintaining the mapping from internable strings to int identifiers. That is what the interning cache and table maintain. The functions constituting that part of the code should not contain any reference to the types for `Party` or `Template`.
> 
> 
> With this factoring of the code the caller takes care of the appropriate composition; and interning other kinds of string-like identifiers becomes a localized change.
> 
> I'd expect the concrete implementation to be such that the interning cache defines the type of `InternableString`, and the `Party` and `Template` mapping then build on that.

thanks, @meiersi-da it is a interesting idea, which I would like to check out very much (although I am not entirely sure it would be better whatever that means ;) )
As discussed I'd like to move on with this approach for now, but created a task to check out this change.

ledger/participant-integration-api/src/main/scala/platform/store/interning/RawStringInterning.scala
```diff
@@ -0,0 +1,40 @@
...
+      )
+
+  def newEntries(
+      strings: Iterator[String],
+      rawStringInterning: RawStringInterning,
+  ): Vector[(Int, String)] =
+    strings
+      .filterNot(rawStringInterning.map.contains)
+      .toVector
+      .distinct // TODO Iterators do not have .distinct in Scala 2.12
```
tudor-da, line 33: Maybe `import scala.collection.compat._` works

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+      partyShouldPresent(testee, "22:same:name", 3)
+      partyShouldNotPresent(testee, "unknown")
+      templateShouldPresent(testee, "22:t:a", 4)
+      templateShouldPresent(testee, "22:t:b", 5)
+      templateShouldPresent(testee, "22:same:name", 6)
+      templateShouldNotPresent(testee, "22:unk:nown")
+    }
+  }
+
+  private def partyShouldPresent(cache: StringInterning, party: String, id: Int) = {
```
mziolekda, line 177:
```suggestion
  private def partyShouldBePresent(cache: StringInterning, party: String, id: Int) = {
```

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    cache.party.tryInternalize(typedParty) shouldBe Some(id)
+    cache.party.externalize(id) shouldBe typedParty
+    cache.party.tryExternalize(id) shouldBe Some(typedParty)
+    cache.party.unsafe.internalize(party) shouldBe id
+    cache.party.unsafe.tryInternalize(party) shouldBe Some(id)
+    cache.party.unsafe.externalize(id) shouldBe party
+    cache.party.unsafe.tryExternalize(id) shouldBe Some(party)
+  }
+
+  private def partyShouldNotPresent(cache: StringInterning, party: String) = {
```
mziolekda, line 189:
```suggestion
  private def partyShouldNotBePresent(cache: StringInterning, party: String) = {
```

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    val typedParty = Ref.Party.assertFromString(party)
+    Try(cache.party.internalize(typedParty)).isFailure shouldBe true
+    cache.party.tryInternalize(typedParty) shouldBe None
+    Try(cache.party.unsafe.internalize(party)).isFailure shouldBe true
+    cache.party.unsafe.tryInternalize(party) shouldBe None
+  }
+
+  private def templateShouldPresent(cache: StringInterning, template: String, id: Int) = {
```
mziolekda, line 197:
```suggestion
  private def templateShouldBePresent(cache: StringInterning, template: String, id: Int) = {
```

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    cache.templateId.tryInternalize(typedTemplate) shouldBe Some(id)
+    cache.templateId.externalize(id) shouldBe typedTemplate
+    cache.templateId.tryExternalize(id) shouldBe Some(typedTemplate)
+    cache.templateId.unsafe.internalize(template) shouldBe id
+    cache.templateId.unsafe.tryInternalize(template) shouldBe Some(id)
+    cache.templateId.unsafe.externalize(id) shouldBe template
+    cache.templateId.unsafe.tryExternalize(id) shouldBe Some(template)
+  }
+
+  private def templateShouldNotPresent(cache: StringInterning, template: String) = {
```
mziolekda, line 209:
```suggestion
  private def templateShouldNotBePresent(cache: StringInterning, template: String) = {
```

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
```diff
@@ -0,0 +1,111 @@
...
+    * @return If some of the entries were not part of the cache: they will be added, and these will be returned as a interned-id and raw, prefixed string pairs.
+    */
+  def add(domainStringIterators: DomainStringIterators): Iterable[(Int, String)]
+
+  /** Update the StringInterning Cache from persistence
+    *
+    * @param lastStringInterningId this is the "version" of the persistent cache, which from the StringInterningCache can see if it is behind
+    * @return a completion Future: if the cache is behind it will load the missing entries from persistence, and update the cache state
+    */
+  def updateStringInterningCache(lastStringInterningId: Int)(implicit
```
mziolekda, line 31: Should it not be verboten to call `updateStringInterningCache` after calls to `add`? This is inviting trouble.

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
```diff
@@ -0,0 +1,111 @@
...
+import com.daml.logging.LoggingContext
+
+import scala.concurrent.Future
+
+class DomainStringIterators(
+    val parties: Iterator[String],
+    val templateIds: Iterator[String],
+)
+
+trait StringInterningCache extends StringInterning {
```
tudor-da, line 17: As discussed offline, `View` would better suit the name of this structure, since it's a projection of the index.

ledger/participant-integration-api/src/main/scala/platform/store/interning/RawStringInterning.scala
```diff
@@ -0,0 +1,40 @@
...
+      .view
+      .zipWithIndex
+      .map { case (string, index) =>
+        (index + 1 + rawStringInterning.lastId, string)
+      }
+      .toVector
```
tudor-da, line 39:
```suggestion
    strings
      .foldLeft((Vector.newBuilder[(Int, String)], map.keySet, lastId)) {
        case ((result, extra, nextId), newString) if !extra(newString) =>
          (result += nextId -> newString, extra + newString, nextId + 1)
        case (state, _) => state
      }
      ._1
      .result()
```
Shorter and with less traversals and memory footprint, at the expense of readability though 🤔

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
tudor-da, line 31: This cache is very smart. For better separation of concerns, I'd like a dumber version of it: 
* Only has one method that updates the string interning data.
* An external loader takes care of updating from persistence.
* Have a `StringInterningLayer` that wires them up (for means of comprehensive testing - if we want to).
wdyt? In this manner, testing would also be greatly simplified

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
```diff
@@ -0,0 +1,111 @@
...
+package com.daml.platform.store.interning
+
+import com.daml.dec.DirectExecutionContext
+import com.daml.lf.data.Ref
+import com.daml.logging.LoggingContext
+
+import scala.concurrent.Future
+
+class DomainStringIterators(
+    val parties: Iterator[String],
```
tudor-da, line 13: Any specific reason why these are `Iterators` ? When would we want to have a lazy struct passed in here? Reminds me of the issue with the byte transport issue that we had in the Jdbc layer and it seems that this type is inviting to it.

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    partyShouldPresent(testee, "22:same:name", 3)
+    partyShouldNotPresent(testee, "unknown")
+    templateShouldPresent(testee, "22:t:a", 4)
+    templateShouldPresent(testee, "22:t:b", 5)
+    templateShouldPresent(testee, "22:same:name", 6)
+    templateShouldNotPresent(testee, "22:unkno:wn")
+    succeed
+  }
+
+  it should "extend working cache correctly" in {
```
tudor-da, line 51: Isn't this one a superset of the previous one?

tudor-da (approved):

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+      partyShouldPresent(testee, "22:same:name", 3)
+      partyShouldNotPresent(testee, "unknown")
+      templateShouldPresent(testee, "22:t:a", 4)
+      templateShouldPresent(testee, "22:t:b", 5)
+      templateShouldPresent(testee, "22:same:name", 6)
+      templateShouldNotPresent(testee, "22:unk:nown")
+    }
+  }
+
+  it should "be able to update working cache correctly" in {
```
tudor-da, line 132: Nitpicking here: these tests can be more concise.
e.g. why is this one asserting synchronously inserting 2 elements and updating another 4? Why not 1 and 1?

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+    )
+    partyShouldPresent(testee, "p1", 1)
+    partyShouldPresent(testee, "p2", 2)
+    partyShouldPresent(testee, "22:same:name", 3)
+    partyShouldNotPresent(testee, "unknown")
+    templateShouldPresent(testee, "22:t:a", 4)
+    templateShouldPresent(testee, "22:t:b", 5)
+    templateShouldPresent(testee, "22:same:name", 6)
+    templateShouldNotPresent(testee, "22:unkno:wn")
+    succeed
```
kamil-da, line 48:
```suggestion
```

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
```diff
@@ -0,0 +1,216 @@
...
+      3 -> "p|22:same:name",
+      4 -> "t|22:t:a",
+      5 -> "t|22:t:b",
+      6 -> "t|22:same:name",
+    )
+    partyShouldPresent(testee, "p1", 1)
+    partyShouldPresent(testee, "p2", 2)
+    partyShouldPresent(testee, "22:same:name", 3)
+    partyShouldNotPresent(testee, "unknown")
+    templateShouldPresent(testee, "22:t:a", 4)
```
kamil-da, line 44: Could you rename this asserting functions not to be so similar when skimming through tests? E.g.:
```
partyShouldPresent -> assertPartyPresent
partyShouldNotPresent -> assertPartyAbsent
```

or even:

```
partyShouldPresent -> partyPresent
partyShouldNotPresent ->partyAbsent
```

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+  override def updateStringInterningCache(
+      lastStringInterningId: Int
+  )(implicit loggingContext: LoggingContext): Future[Unit] =
+    if (lastStringInterningId <= raw.lastId) {
+      Future.unit
+    } else {
+      loadPrefixedEntries(raw.lastId, lastStringInterningId)(loggingContext)
+        .map(updateCache)(DirectExecutionContext)
```
kamil-da, line 100: I would prefer having synchronous update method with already fetched data instead of passing a Future loading it from persistence.

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
nmarton-da: Good point, how about segregating this interface, and 
* for the read-side only allow `updateStringInterningCache` 
* for the write side's initialization only `updateStringInterningCache`
* and for the write side updating mechanism only `add`

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
nmarton-da: Yeah I think so too. @meiersi-da are you ok with this naming? (in the design doc we use the term Covering Cache)

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
nmarton-da: I am not so sure about that: for example If you expose one method to update the string interning data, then also you have to expose the last id, and one of the maps in the Raw datastructure...and this is just the beginning ;)
I have the feeling that this trait represents perfectly the `StringInterningLayer` you suggest...
can we have a ppsession together where we explore your suggestion?

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
nmarton-da: not to construct intermediate datastructures

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
nmarton-da: respectively: 
* to test both domains are updated, going not from scratch
* to test update is hitting both domains with colliding strings

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
nmarton-da: good idea

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
kamil-da: Seems like a sensible choice. I'm ok with injecting a Future 👍

ledger/participant-integration-api/src/test/suite/scala/platform/store/interning/StringInterningCacheSpec.scala
nmarton-da: actually the first one does one the first iteration with broader set of values, I'd like to keep it if that is fine with you

ledger/participant-integration-api/src/main/scala/platform/store/interning/StringInterningCache.scala
nmarton-da: As discussed offline we get back to this later

Issue #11401, opened 2021-10-26
Extend test evidence


cocreature:
1. More fine-grained tests for authorization. Something along the lines of "create with correct signatories gets accepted", "create with missing signatories gets rejected".
2. Input validation of daml-lf values in the engine: Accept correctly typed values, reject invalidly typed values.
3. Input validaton of commands: positive & negative cases for each command
4. Blinding spec
5. Contract keys (move from EngineTest)
6. Merge authorization tests & do them at the speedy level

nickchapman-da:
Item 5 done

mziolekda:
Closing, obsolete issue

PR #11479, opened 2021-11-01
Evidence for contract-key testing

nickchapman-da:
Split contract-key tests from engine tests, and mark for evidence.
- Addresses item 5 of https://github.com/digital-asset/daml/issues/11401


daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractKeySpec.scala
```diff
@@ -0,0 +1,430 @@
...
+    new preprocessing.Preprocessor(
+      ConcurrentCompiledPackages(suffixLenientEngine.config.getCompilerConfig)
+    )
+
+  "contract key" should {
+    val now = Time.Timestamp.now()
+    val submissionSeed = crypto.Hash.hashPrivateKey("contract key")
+    val txSeed = crypto.Hash.deriveTransactionSeed(submissionSeed, participant, now)
+
+    // TEST_EVIDENCE: Semantics: contract keys should be evaluated only when executing create
```
cocreature, line 146:
```suggestion
    // TEST_EVIDENCE: Semantics: contract keys should be evaluated after ensure clause
```
maybe a bit clearer and matches what this test actually tests afaict.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractKeySpec.scala
```diff
@@ -0,0 +1,430 @@
...
+          seeding = InitialSeeding.TransactionSeed(txSeed),
+        )
+        .consume(_ => None, lookupPackage, lookupKey)
+      result shouldBe a[Right[_, _]]
+    }
+
+    // TEST_EVIDENCE: Semantics: contract keys should be evaluated after ensure clause
+    "be evaluated after ensure clause" in {
+      val templateId =
+        Identifier(basicTestsPkgId, "BasicTests:ComputeContractKeyAfterEnsureClause")
```
cocreature, line 188: I’m confused. What is this test testing that the one above doesn’t test?
`ComputeContractKeyWhenExecutingCreate` just calls `ComputeContractKeyAfterEnsureClause`.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractKeySpec.scala
```diff
@@ -0,0 +1,430 @@
...
+        err.message should include(
+          "Update failed due to a contract key with an empty sey of maintainers"
+        )
+      }
+    }
+
+    // Note that we provide no stability for multi key semantics so
+    // these tests serve only as an indication of the current behavior
+    // but can be changed freely.
+    "multi keys" should {
```
cocreature, line 255: I think these deserve some markers. Not for the multi-keys behavior (although no harm in already including that) but for the uck mode (we run them in both).

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -1819,6 +1819,7 @@ class EngineTest
     }
   }
 
+  // TEST_EVIDENCE: Semantics: Contract key behaviour
   "contract key" should {
```
cocreature, line 1823: did you intend to drop these here? It looks like they are duplicated now.

cocreature (approved):
Nice, thanks!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
nickchapman-da: I totally did mean to drop these! Thanks for catching this. Dropped now.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/ContractKeySpec.scala
nickchapman-da: It looks like the first test shows the behaviour when the `create` is not actually executed:
```
        let _ignore = create ComputeContractKeyAfterEnsureClause with owner
```
I'm not sure how useful it is to test that here, as that's just standard behaviour for a action which is not executed,

Issue #11137, opened 2021-10-06
Interface Fixed Choices


sofiafaro-da:
- [x] daml-lf protobuf
- [x] haskell ast
- [x] haskell typechecker
   - [x] add check that interface fixed choice doesn't collide with template choice
- [x] haskell conversion
- [x] ghc parser
- [x] scala ast
- [x] scala validation
   - [x] add check that interface fixed choice doesn't collide with template choice
- [x] speedy

PR #11481, opened 2021-11-01
Check fixed choices in name collision checker

sofiafaro-da:
This is a follow up to #11364. This PR checks that the list of inherited
choice names is correct in the type checker, and moves the "fixed choice name
collision" check into the name collision checker, since it's now a local check.

Haskell side only for now. Part of #11137.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
+++ b/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
@@ -920,7 +920,6 @@ checkIfaceImplementation tplTcon TemplateImplements{..} = do
       Just InterfaceMethod{ifmType} ->
         checkExpr tpiMethodExpr (TCon tplTcon :-> ifmType)
 
-
 _checkFeature :: MonadGamma m => Feature -> m ()
 _checkFeature feature = do
     version <- getLfVersion
--- a/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
+++ b/compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
@@ -885,27 +885,18 @@ checkTemplate m t@(Template _loc tpl param precond signatories observers text ch
   whenJust mbKey $ checkTemplateKey param tcon
   forM_ implements $ checkIfaceImplementation tcon
 
-  -- Check template choice and interface fixed choice name collisions.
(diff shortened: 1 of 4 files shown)


compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
```diff
@@ -885,27 +885,18 @@ checkTemplate m t@(Template _loc tpl param precond signatories observers text ch
...
-
 checkIfaceImplementation :: MonadGamma m => Qualified TypeConName -> TemplateImplements -> m ()
 checkIfaceImplementation tplTcon TemplateImplements{..} = do
   let tplName = qualObject tplTcon
-  DefInterface {intVirtualChoices, intMethods} <- inWorld $ lookupInterface tpiInterface
+  DefInterface {intFixedChoices, intVirtualChoices, intMethods} <- inWorld $ lookupInterface tpiInterface
+
+  -- check fixed choices
+  let inheritedChoices = S.fromList (NM.names intFixedChoices)
+  unless (inheritedChoices == tpiInheritedChoiceNames) $
```
cocreature, line 898: is it possible to trigger this in Daml? If so, a test might be nice.

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
```diff
@@ -929,6 +920,7 @@ checkIfaceImplementation tplTcon TemplateImplements{..} = do
       Just InterfaceMethod{ifmType} ->
         checkExpr tpiMethodExpr (TCon tplTcon :-> ifmType)
 
+
```
cocreature, line 923:
```suggestion
```
unintentional? if not ignore

cocreature (approved):
Thanks!

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
sofiafaro-da: I don't think it's possible. ATM this only serves as a sanity check against the LF completer.

compiler/daml-lf-tools/src/DA/Daml/LF/TypeChecker/Check.hs
sofiafaro-da: unintentional!

Issue #11372, opened 2021-10-25
Drop non-fixed choices from interfaces


cocreature:
If we want them, we can add them as syntactic sugar later.

PR #11482, opened 2021-11-01
Remove virtual choices

akrmn:
* Remove choices without a body in 'interface' definition
* Remove choices in 'template ... implements' section

part of #11372

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


remyhaemmerle-da (commented):
Scala/proto changes looks good.

akrmn:
thanks @remyhaemmerle-da!

build.ps1
```diff
@@ -49,34 +49,3 @@ function bazel() {
...
-# ScalaCInvoker, a Bazel worker, created by rules_scala opens some of the bazel execroot's files,
-# which later causes issues on Bazel init (source forest creation) on Windows. A shutdown closes workers,
-# which is a workaround for this problem.
-bazel shutdown
-
-# Prefetch nodejs_dev_env to avoid permission denied errors on external/nodejs_dev_env/nodejs_dev_env/node.exe
-# It isn’t clear where exactly those errors are coming from.
-bazel fetch @nodejs_dev_env//...
-
-bazel build //... `
```
garyverhaegen-da, line 62: I'm all for dropping Windows, but... is this intentional?

garyverhaegen-da (changes requested):

build.ps1
akrmn: heh, no, I thought I only checked in the other file, thanks for spotting it.

cocreature (approved):
Thanks, looks good!

akrmn:
I think I'm missing something, `bazel test //language-support/ts/codegen/tests:build-and-lint-test` fails with
```
  ● interfaces

    TypeError: Cannot read property 'template' of undefined

      at Ledger.<anonymous> (../../../../language-support/ts/daml-ledger/index.ts:878:26)
      at step (../node_modules/@daml/ledger/index.js:76:23)
      at Object.next (../node_modules/@daml/ledger/index.js:57:53)
      at ../node_modules/@daml/ledger/index.js:51:71
      at Object.<anonymous>.__awaiter (../node_modules/@daml/ledger/index.js:47:12)
      at Ledger.Object.<anonymous>.Ledger.exercise (../node_modules/@daml/ledger/index.js:760:16)
      at src/__tests__/test.ts:348:55
      at step (src/__tests__/test.ts:46:23)
      at Object.next (src/__tests__/test.ts:27:53)
```

Similarly, `bazel test //daml-script/test:test_test_suite_src_com_digitalasset_daml_lf_engine_script_test_FuncWallClockIT.scala` fails with

```
 - Received request for ledger identity: GetLedgerIdentityRequest()
  - should succeed *** FAILED *** (544 milliseconds)
    com.daml.lf.engine.script.ScriptF$FailedCmd: Command submit failed: INVALID_ARGUMENT: Disputed: definition ChoiceDefRef(db848fc5342300b19eca7af68a02ace4f041603fbb20d16dbef30e5abe2d578c:TestInterfaces:Asset,Noop) not found even after caller provided new set of packages
```

I'm not familiar with this part of the codebase, so perhaps someone else could help me? It seems like @robin-da has worked on the typescript test recently, and @cocreature on the scala one.

cocreature:
The thing the two have in common is that they both try to exercise a choice via the ledger API. I suspect https://github.com/digital-asset/daml/pull/11013/files might not work for fixed choices and only for virtual choices at this point. @remyhaemmerle-da does that sound right?

remyhaemmerle-da:
> The thing the two have in common is that they both try to exercise a choice via the ledger API. I suspect https://github.com/digital-asset/daml/pull/11013/files might not work for fixed choices and only for virtual choices at this point. @remyhaemmerle-da does that sound right?

For now, fixed choices can be called only through interface id. The more I think about, the more I think we should keep it like that. We should discuss about that in the next "interface" meeting.

cocreature:
> For now, fixed choices can be called only through interface id. The more I think about, the more I think we should keep it like that. We should discuss about that in the next "interface" meeting.

are you sure even that case works? Looking at the daml script testcase https://github.com/digital-asset/daml/blob/5e93df6b521fe669baa2d7f859cc81f6bb60f6f6/daml-script/test/daml/TestInterfaces.daml#L66 we do call it through the interface id afaict and it still fails.

remyhaemmerle-da:
> are you sure even that case works? Looking at the daml script testcase

theoretically it should work. But I probably messed up somewhere. let's me take a look.

cocreature:
The error is
```
    com.daml.lf.engine.script.ScriptF$FailedCmd: Command submit failed: INVALID_ARGUMENT: Disputed: definition ChoiceDefRef(db848fc5342300b19eca7af68a02ace4f041603fbb20d16dbef30e5abe2d578c:TestInterfaces:Asset,Noop) not found even after caller provided new set of packages
```
So it looks like it is looking at the concrete template rather than the interface and it looks like we might be blowing up on a lookup on the compiled packages rather than in the preprocessor.

akrmn:
> fixed choices can be called only through interface id

> Looking at the daml script testcase https://github.com/digital-asset/daml/blob/5e93df6b521fe669baa2d7f859cc81f6bb60f6f6/daml-script/test/daml/TestInterfaces.daml#L66 we do call it through the interface id afaict and it still fails.

Note that `toContractId` just coerces `ContractId Implements -> ContractId Interface`. I'm not sure if @remyhaemmerle-da's comment meant that they are different ids.

cocreature:
> > 
> 
> Note that `toContractId` just coerces `ContractId Implements -> ContractId Interface`. I'm not sure if @remyhaemmerle-da's comment meant that they are different ids.

What @remyhaemmerle-da is referring to is what we pass when sending the command on the ledger API: We need to fill in the `templateId` field in the command which can either be the template id or the interface id. Daml Scirpt makes the choice based on the type so the coercion here does accomplish what we want.

akrmn:
Gotcha! I wasn't sure since e.g. in Typescript it seems we mostly (only?) use structural typing.

cocreature:
> Gotcha! I wasn't sure since e.g. in Typescript it seems we mostly (only?) use structural typing.

haven’t quite checked that example yet, daml script is easier to debug (for me at least) so I was looking into getting that working first and hopefully that fixes both issues :)

remyhaemmerle-da:
I figured out the problem with the script:
KV calls the engine twice, once for interpreting and once for validation the transaction. 
The first succeeds, but the second does not, because we never consider it. 

Should we discuss about that in today meeting.

remyhaemmerle-da:
@moisesackerman-da for now I think you should disable those failing test, and adding a TODO like 
```
// TODO https://github.com/digital-asset/daml/issues/10810
// Renable test 
``

**Comment by cocreature** [comment]
I agree with @remyhaemmerle-da, let’s merge this and disable the tests for now.

File: compiler/damlc/tests/daml-test-files/Interface.daml [hunk 2]
```diff
@@ -46,30 +59,18 @@ template Asset
...
-          cid <- create this with owner = newOwner
-          pure (toTokenContractId cid)
-
-      nonconsuming choice Noop : ()
-        with
-          nothing : ()
-        controller owner
-        do
-          pure ()
+      let splitImpl = \splitAmount -> do
```
robin-da, line 62: So this separation of implementation and choice body is something we always have to do now, right?

robin-da (approved):

compiler/damlc/tests/daml-test-files/Interface.daml
akrmn: exactly! As @cocreature pointed out in [the issue](https://github.com/digital-asset/daml/issues/11372) this PR should fix, we can later add virtual choices again as syntactic sugar, but internally they would still be fixed choices with interface methods as the implementation.

akrmn:
not sure what made the mac tests fail, they run fine on my machine. Not sure about the windows failure either, `//daml-script/test:test_6`. I'll start another build 🤷

cocreature:
> /azp run

note that `/azp run` reruns everything whereas `rerun failed` in the `checks` tab on GH only reruns failed jobs. Usually the latter is preferrable.

akrmn:
oh! thanks for the tip @cocreature, still getting used to github+azure coming from jira+jenkins

PR #11485, opened 2021-11-01
Pass `damlc build` logs through daml logger

cocreature:
This gives us some timings which can be useful in some
circumstances (I want timings around data-dependencies next).

I also changed the default log level to INFO. The `damlc build` output
doesn’t get any noisier from that.

Example output:

```
2021-11-01 15:25:17.16 [INFO]  [build]  []
Compiling foobar to a DAR.

2021-11-01 15:25:17.80 [INFO]  [build]  []
Created .daml/dist/foobar-0.0.1.dar
```

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/lib/DA/Cli/Damlc.hs
+++ b/compiler/damlc/lib/DA/Cli/Damlc.hs
@@ -453,7 +453,7 @@ execIde telemetry (Debug debug) enableScenarioService options =
                   then Logger.Debug
                   -- info is used pretty extensively for debug messages in our code base so
                   -- I've set the no debug threshold at warning
-                  else Logger.Warning
+                  else Logger.Info
           loggerH <- Logger.IO.newIOLogger
             stderr
             (Just 5000)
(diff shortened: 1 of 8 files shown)


compiler/damlc/lib/DA/Cli/Damlc.hs
```diff
@@ -453,7 +453,7 @@ execIde telemetry (Debug debug) enableScenarioService options =
                   then Logger.Debug
                   -- info is used pretty extensively for debug messages in our code base so
                   -- I've set the no debug threshold at warning
```
sofiafaro-da, line 455: This comment should be updated or removed.

sofiafaro-da (approved):
That sounds very useful, thanks!

compiler/damlc/lib/DA/Cli/Damlc.hs
cocreature: good catch, just dropped it. I think INFO is the usual default so doesn’t really need an explanation.

PR #11491, opened 2021-11-01
ledger-offset: Add a conversion to ByteString.

SamirTalwar-DA:
I seem to be doing this a lot.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-offset/src/main/scala/com/daml/ledger/offset/Offset.scala
+++ b/ledger/ledger-offset/src/main/scala/com/daml/ledger/offset/Offset.scala
@@ -36,7 +36,9 @@ final case class Offset(bytes: Bytes) extends Ordered[Offset] {
 }
 
 object Offset {
-  val beforeBegin: Offset = Offset.fromByteArray(Array.empty[Byte])
+  val beforeBegin: Offset = new Offset(Bytes.Empty)
+
+  def fromByteString(bytes: ByteString) = new Offset(Bytes.fromByteString(bytes))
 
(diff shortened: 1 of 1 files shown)


rautenrieth-da (commented):
Where do you need this? I can't find a single place in this repository where we would convert an Offset to a ByteString.

fabiotudone-da (commented):
I wonder, since this class is immutable, how about using lazy vals?

SamirTalwar-DA:
@rautenrieth-da: I find myself using it a lot in another repository when converting offsets back and forth into protobuf values.

SamirTalwar-DA:
@fabiotudone-da: I am not sure. In the case of `toByteString`, this just accesses a property on `bytes`, and I'm not sure using a `lazy val` to capture the result of this property access would improve performance. However, this is due to my naive understanding of how well the JVM needs help like this. What do you think?

For `toByteArray`, `toByteBuffer`, and `toInputStream`, the return value is mutable, and so we cannot keep it around.

fabiotudone-da:
> @fabiotudone-da: I am not sure. In the case of `toByteString`, this just accesses a property on `bytes`, and I'm not sure using a `lazy val` to capture the result of this property access would improve performance. However, this is due to my naive understanding of how well the JVM needs help like this. What do you think?
> 
> For `toByteArray`, `toByteBuffer`, and `toInputStream`, the return value is mutable, and so we cannot keep it around.

True, definitely not possible/worth.

fabiotudone-da (approved):

hubert-da (approved):

PR #11498, opened 2021-11-02
Enable experimental nix command

cocreature:
Nix 2.4 got released which confusingly marked the nix command as
experimental despite it being actively worked on and being more stable
than before.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ci/dev-env-push.py
+++ b/ci/dev-env-push.py
@@ -104,7 +104,7 @@ def main():
         shutil.rmtree(nix_cache_dir)
 
     # copy to nix cache
-    cmd = ["nix", "copy", "--to", store_url, "-f", "./nix", "tools", "ci-cached"]
+    cmd = ["nix", "--extra-experimental-features", "nix-command", "copy", "--to", store_url, "-f", "./nix", "tools", "ci-cached"]
     log_cmd(cmd)
     proc = subprocess.run(
             cmd,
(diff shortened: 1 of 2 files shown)


ci/dev-env-push.py
```diff
@@ -104,7 +104,7 @@ def main():
         shutil.rmtree(nix_cache_dir)
 
     # copy to nix cache
-    cmd = ["nix", "copy", "--to", store_url, "-f", "./nix", "tools", "ci-cached"]
+    cmd = ["nix", "--extra-experimental-features", "nix-command", "copy", "--to", store_url, "-f", "./nix", "tools", "ci-cached"]
```
cocreature, line 107: I’ve gone for the CLI flag since that doesn’t result in a warning on nix 2.3 for people locally since this script is only run by CI and afaict this is our only usage of the nix command (which probably could be replaced but then again it worked fine forever so no real reason to).

hubert-da (approved):

ci/dev-env-push.py
cocreature: apparently on macos we still have an older version of nix because our setup is a mess so went back to the config file. While it produces a warning it works across both versions at least.

aherrmann-da (approved):

PR #11500, opened 2021-11-02
Add priority-queue based worker Source [DPP-717]

nmarton-da:
* Add ACSReader trait, and primitives
* Add pullWorkerSource implementation + unit tests

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
@@ -1,0 +1,138 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.appendonlydao.events
+
+import akka.NotUsed
+import akka.stream.{BoundedSourceQueue, Materializer, QueueOfferResult}
+import akka.stream.scaladsl.Source
+import com.daml.dec.DirectExecutionContext
+import com.daml.ledger.api.v1.active_contracts_service.GetActiveContractsResponse
+import com.daml.ledger.offset.Offset
+import com.daml.lf.data.Ref
+import com.daml.logging.LoggingContext
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+trait ACSReader {
+  def acsStream(
+      filter: FilterRelation,
+      activeAt: (Offset, Long),
+      verbose: Boolean,
+  )(implicit loggingContext: LoggingContext): Source[GetActiveContractsResponse, NotUsed]
+}
+
+class FilterTableACSReader extends ACSReader {
+
+  override def acsStream(
+      filter: FilterRelation,
+      activeAt: (Offset, Long),
+      verbose: Boolean,
+  )(implicit loggingContext: LoggingContext): Source[GetActiveContractsResponse, NotUsed] =
+    throw new NotImplementedError() // FIXME will be implemented in a later increment
+
+}
+
+private[events] object FilterTableACSReader {
+
+  case class Filter(party: Party, templateId: Option[Ref.Identifier])
+
+  case class QueryTask(fromExclusiveEventSeqId: Long, filter: Filter)
+
+  object QueryTask {
+    implicit val ordering: Ordering[QueryTask] =
+      Ordering.by[QueryTask, Long](_.fromExclusiveEventSeqId)
+  }
+
+  /** This Source implementation solves the following problem:
+    *  - let us have n TASKs, which are ordered
+    *  - let us define some work over these tasks, which gives us a RESULT and a possible continuation of the TASK
+    *  - let us have configurable parallelism to work on these TASKs
+    *  This implementation ensures that all the time the smallest available TASK will be picked for work.
+    *
+    * Please note:
+    *  - If workerParallelism is one, this should result in monotonously increasing execution sequence
+    *    (regardless of the demand downstream)
+    *  - If workerParallelism is equal or bigger than the number of initial n tasks, and downstream is faster,
+    *    then prioritization has no time to kick in (the backing priority queue will have mostly one element), so
+    *    execution order will be similar to simple parallel execution of sequences of tasks
+    *
+    * @param workerParallelism defines the maximum parallelism of unordered processing.
+    *                          Naturally capped by size of initialTasks
+    * @param work The worker function, asynchronous computation should return a RESULT,
+    *             and the next TASK, or no TASK if TASK processing is finished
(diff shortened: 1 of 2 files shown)


meiersi-da (commented):
@nmarton-da : LGTM, I'm not fluent enough with Akka and Scala to review this one in detail. Please check with one of your team mates.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -0,0 +1,138 @@
...
+  case class QueryTask(fromExclusiveEventSeqId: Long, filter: Filter)
+
+  object QueryTask {
+    implicit val ordering: Ordering[QueryTask] =
+      Ordering.by[QueryTask, Long](_.fromExclusiveEventSeqId)
+  }
+
+  /** This Source implementation solves the following problem:
+    *  - let us have n TASKs, which are ordered
+    *  - let us have an work defined over these tasks, which gives us a RESULT and a possible continuation of the TASK
```
mziolekda, line 50:
```suggestion
    *  - let us define some work over these tasks, which gives us a RESULT and a possible continuation of the TASK
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -0,0 +1,138 @@
...
+    *  - let us have configurable parallelism to work on these TASKs
+    *  This implementation ensures that all the time the "smallest" available TASK will be picked for work.
```
mziolekda, line 52: I think "smallest" doesn't need to be quoted. You have introduced the concept of ordering in the first bullet point.
What happens with a task that was smallest and finished with continuation. Does it go back to the front of the queue?

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -0,0 +1,138 @@
...
+    *
+    * Please note:
+    *  - If workerParallelism is one, this should result in monotonously increasing execution sequence
+    *    (regardless of the demand downstream)
+    *  - If workerParallelism is equal or bigger than the number of initial n tasks, and downstream is faster,
+    *    then prioritization has on time to kick in (the backing priority queue will have mostly one element), so
```
mziolekda, line 58:
```suggestion
    *    then prioritization has no time to kick in (the backing priority queue will have mostly one element), so
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -0,0 +1,138 @@
...
+    *    execution order will be similar to simple parallel execution of sequences of tasks
+    *
+    * @param workerParallelism defines the maximum parallelism of unordered processing.
+    *                          Naturally capped by size of initialTasks
+    * @param work The worker function, asynchronous computation should return a RESULT,
+    *             and the next TASK, or no TASK if TASK processing is finished
+    * @param initialTasks The collection of initial TASKS for execution
+    * @tparam TASK type of TASKs, needs to have an Ordering defined.
+    *              Always the "smallest" task will be selected for execution
+    * @tparam RESULT The ype of the RESULT
```
mziolekda, line 68:
```suggestion
    * @tparam RESULT The type of the RESULT
```

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
nmarton-da: > I think "smallest" doesn't need to be quoted. 
yes, good point, will do

> What happens with a task that was smallest and finished with continuation. Does it go back to the front of the queue?
If a task finished with a continuation it will be put back to the queue. The queue is a priority queue, so it will be put in the right place based on the ordering. As picking the next task, the smallest task will be chosen.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -0,0 +1,138 @@
...
+    initialTasks.foreach(addTask)
+
+    def startTask: TASK = synchronized {
+      runningTasks += 1
+      priorityQueue.dequeue()
+    }
+
+    def finishTask(nextTask: Option[TASK]): Unit = synchronized {
+      nextTask match {
+        case None if priorityQueue.isEmpty && runningTasks == 1 =>
```
mziolekda, line 118: I think there is a problem with the runningTasks counter. You have not run into it yet, because unit tests only use parallelism of 1.
Imagine there are 2 tasks without continuation and the parallelism is 2. They both will go through `case None` without decrement of runningTasks. As a result the signalQueue will never complete.

ledger/participant-integration-api/src/test/suite/scala/platform/store/appendonlydao/events/ACSReaderSpec.scala
```diff
@@ -0,0 +1,114 @@
...
+  private implicit val ec: ExecutionContext = actorSystem.dispatcher
+
+  override def afterAll(): Unit = {
+    Await.result(actorSystem.terminate(), Duration(10, "seconds"))
+    ()
+  }
+
+  behavior of "pullWorkerSource"
+
+  it should "give an empty source if initialTasks are empty" in {
```
mziolekda, line 29: Please add one test with parallelism 2. There will be new class of issues when the code is used concurently.

mziolekda (commented):
The PR looks ok, I would like to see a test for parallelism>1 though

ledger/participant-integration-api/src/test/suite/scala/platform/store/appendonlydao/events/ACSReaderSpec.scala
nmarton-da: oh boy, I saw that coming :(
actually doing parallel tests with expectations is not something which is easy (we need to hand-orchestrate completion from test suit. time based expectations just not work in our CI).
I will figure something out, since this is a workhorse of the 2 phase retrieval :+1:

kamil-da (approved):
LGTM!

ledger/participant-integration-api/src/test/suite/scala/platform/store/appendonlydao/events/ACSReaderSpec.scala
nmarton-da: Added some more elaborate tests for increased parallelism, even a puppetmaster test where asynchronous behavior is analyzed in great detail

mziolekda (approved):

PR #11501, opened 2021-11-02
[Trigger-Service] Migrate tests to sandbox

akshayshirahatti-da:
Migrate trigger service tests to sandbox from sandbox-classic.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceFixture.scala
```diff
@@ -260,26 +259,27 @@ trait SandboxFixture extends BeforeAndAfterAll with AbstractAuthFixture with Akk
...
     LedgerIdMode.Static(LedgerId(this.getClass.getSimpleName))
-  private def sandboxConfig: SandboxConfig = sandbox.DefaultConfig.copy(
+  private def sandboxConfig: SandboxConfig = SandboxConfig.defaultConfig.copy(
     port = Port.Dynamic,
     ledgerIdMode = ledgerIdMode,
     damlPackages = damlPackages,
     timeProviderType = Some(TimeProviderType.Static),
     delayBeforeSubmittingLedgerConfiguration = JDuration.ZERO,
     authService = authService,
+    engineMode = SandboxConfig.EngineMode.Dev,
```
stefanobaghino-da, line 269: What does this do?

stefanobaghino-da (commented):
LGTM, thanks. Ask for someone else from the team to review as well. 🙂

triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceFixture.scala
akshayshirahatti-da: It seems to influence language versions, but my main reason for adding it is because the `AbstractSandboxFixture` sets it to that https://github.com/digital-asset/daml/blob/main/ledger/sandbox-common/src/test/lib/scala/platform/sandbox/AbstractSandboxFixture.scala#L71

not that well thought TBH 🤷

cocreature: what breaks if we don’t do this? I’m fine with this but I’d like to understand why we need this.

cocreature (approved):
Thanks!

aherrmann-da (approved):
Thank you!

triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceFixture.scala
akshayshirahatti-da: > what breaks if we don’t do this? I’m fine with this but I’d like to understand why we need this.

did a quick check and nothing broke, default is `engineMode = EngineMode.Stable`. As I said `SandboxNextFixture` which extends `AbstractSandboxFixture` sets the mode to `Dev` and it would be worth keeping that constant across all our tests, so I am leaning towards leaving it there, unless there is a strong reason to remove it 🤔

cocreature: I’d rather change the value in `SandboxNextFixture` tbh. `daml sandbox` defaults to rejecting 1.dev, having the fixture behave differently just seems confusing.

S11001001 (approved):

triggers/service/src/test/scala/com/digitalasset/daml/lf/engine/trigger/TriggerServiceFixture.scala
akshayshirahatti-da: alright removed

PR #11502, opened 2021-11-02
LF: Deprecate  ensureNoCid and assertNoCid from CidContainer

remyhaemmerle-da:
Before #10827 wich drops type parameters in Value, `CidContainer`'s,
`ensureNoCid` and `assertNoCid` were useful to convert `Value[ContractId]`
 to `Value[Nothing]`. They are now useless as converter, so we deprecate 
them.  

To check a value does not contain Contract Ids one should use
`foreachCid` or `cids` methods.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
@@ -1747,16 +1747,11 @@ private[lf] object SBuiltin {
     v match {
       case SStruct(_, vals) =>
         val key = onLedger.ptx.normValue(templateId, vals.get(keyIdx))
-        key.ensureNoCid match {
-          case Right(keyVal) =>
-            Node.KeyWithMaintainers(
-              key = keyVal,
-              maintainers =
-                extractParties(NameOf.qualifiedNameOfCurrentFunc, vals.get(maintainerIdx)),
-            )
-          case Left(_) =>
-            throw SErrorDamlException(IE.ContractIdInContractKey(key))
-        }
+        key.foreachCid(_ => throw SErrorDamlException(IE.ContractIdInContractKey(key)))
+        Node.KeyWithMaintainers(
+          key = key,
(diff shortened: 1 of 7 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1742,16 +1742,11 @@ private[lf] object SBuiltin {
...
-          case Right(keyVal) =>
-            Node.KeyWithMaintainers(
-              key = keyVal,
-              maintainers =
-                extractParties(NameOf.qualifiedNameOfCurrentFunc, vals.get(maintainerIdx)),
-            )
-          case Left(_) =>
-            throw SErrorDamlException(IE.ContractIdInContractKey(key))
-        }
+        key.foreachCid(_ => throw SErrorDamlException(IE.ContractIdInContractKey(key)))
```
remyhaemmerle-da, line 1745: Here we need to check for contract Ids.

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
```diff
@@ -159,9 +159,7 @@ private[platform] class ActiveLedgerStateManager[ALS <: ActiveLedgerState[ALS]](
                 witnesses = disclosure(nodeId),
                 // A contract starts its life without being divulged at all.
                 divulgences = Map.empty,
-                key = nc.versionedKey.map(
```
cocreature, line 162: I assume those checks don’t matter because we do them in the engine already?

cocreature (approved):

ledger/participant-integration-api/src/main/scala/platform/store/ActiveLedgerStateManager.scala
remyhaemmerle-da: Yes the `assertNoCid` was used to change the type from `Value[ContractId]` to `Value[Nothing]`. Here I assume that the engine already check the absence of contract IDs.

PR #11507, opened 2021-11-02
Fixed ledger-api-bench-tool random distribution in submission phase.

kamil-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/Distribution.scala
+++ b/ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/Distribution.scala
@@ -3,14 +3,15 @@
 
 package com.daml.ledger.api.benchtool.submission
 
+/** Allows to pseudo-randomly pick an index out of a set of indices according to their weights. */
 class Distribution(weights: List[Int]) {
   assert(weights.nonEmpty, "Weights list must not be empty.")
   assert(!weights.exists(_ < 1), "Weights must be strictly positive.")
 
   def index(randomDouble: Double): Int = {
-    assert(randomDouble <= 1.0, "Given random double must be <= 1.0.")
+    assert(randomDouble < 1.0, "Given random double must be < 1.0.")
     // Consider changing implementation to use binary search when using on large lists.
-    distribution.indexWhere(_ >= randomDouble)
+    distribution.indexWhere(_ > randomDouble)
   }
 
   private lazy val totalWeight: Long = weights.map(_.toLong).sum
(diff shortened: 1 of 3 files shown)


ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/Distribution.scala
```diff
@@ -3,14 +3,15 @@
...
 class Distribution(weights: List[Int]) {
   assert(weights.nonEmpty, "Weights list must not be empty.")
   assert(!weights.exists(_ < 1), "Weights must be strictly positive.")
 
   def index(randomDouble: Double): Int = {
-    assert(randomDouble <= 1.0, "Given random double must be <= 1.0.")
+    assert(randomDouble < 1.0, "Given random double must be < 1.0.")
     // Consider changing implementation to use binary search when using on large lists.
-    distribution.indexWhere(_ >= randomDouble)
+    distribution.indexWhere(_ > randomDouble)
```
rautenrieth-da, line 14: We should probably test the case where you call `.index(0)` and `.index(1)` on a non-trivial distribution. This should return the first and last index, respectively.

rautenrieth-da (approved):

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/Distribution.scala
kamil-da: I could add a test for `0.0` and `(Int.MaxValue - 1)/Int.MaxValue` but AFAIK the ScalaCheck tests corner cases quite extensively when using generators.
We already have this:
```
      val cases = for {
        double <- zeroToOneDouble
        weights <- listOfWeights
      } yield (weights, double)
```
which should cover test cases you mentioned.
The only gain would be that we would have explicit test case. Do you think it's worth doing?

rautenrieth-da: If ScalaCheck tests corner cases, then we are good. I was mainly wondering why the change in this line (which AFAICT changes the behavior for `.indexWhere(0)`) wasn't caught by tests.

kamil-da: The reason was that limits in tests were incorrect. Now they are fixed.

PR #11508, opened 2021-11-02
Migrate language-support codegen tests to sandbox

akshayshirahatti-da:
Migrate language-support codegen tests to sandbox.
 `//language-support/java/codegen:ledger-tests` tests were timing out 😞 so had to modify tests to re-use the sandbox which meant some shenanigans around unique parties per test.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
```diff
@@ -19,55 +18,63 @@ import com.daml.ledger.javaapi.data
...
+
+  val LedgerID = "ledger-test"
 
   // unfortunately this is needed to help with passing functions to rxjava methods like Flowable#map
   implicit def func2rxfunc[A, B](f: A => B): io.reactivex.functions.Function[A, B] = f(_)
   private def randomId = UUID.randomUUID().toString
 
+  def uniqueId(): String = UUID.randomUUID.toString
+
+  protected[daml] def getUniqueParty(name: String): String = s"${name}_${uniqueId()}"
```
S11001001, line 68:
```suggestion
  private[daml] def getUniqueParty(name: String): String = s"${name}_${uniqueId()}"
```

S11001001 (approved):

language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
cocreature: rather than trying to make up unique parties can you allocate them via the party admin service on the ledger API? On Canton that will be required anyway so this seems like a step in the right direction.

cocreature (approved):
Thanks!

language-support/java/codegen/src/ledger-tests/scala/com/digitalasset/TestUtil.scala
akshayshirahatti-da: > rather than trying to make up unique parties can you allocate them via the party admin service on the ledger API

done 👍

cocreature: good stuff, thank you!

PR #11509, opened 2021-11-02
Save the submission time in the command deduplication value [KVL-1173]

nicu-da:
This will allow us to change the deduplication logic to use the submission time instead of deduplicate_until to determine if a command is a duplicate or not

CHANGELOG_BEGIN

CHANGELOG_END

### Pull Request Checklist

- [ x Set a descriptive title and thorough description

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
+++ b/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
@@ -54,12 +54,11 @@ message DamlCommandDedupKey {
 
 message DamlCommandDedupValue {
   reserved 1; // was record_time
-  // the time until when future commands with the same
-  // deduplication key will be rejected due to a duplicate submission
+  // The time until when future commands with the same
+  // deduplication key will be rejected due to a duplicate submission.
   google.protobuf.Timestamp deduplicated_until = 2;
-  //The submission time of the initial command
-  // We use submission time because record time is not available during pre-execution
-  google.protobuf.Timestamp submission_time = 3;
+  // The submission time of the initial command.
+  // We use submission time because record time is not available during pre-execution.
 }
 
 message DamlSubmissionDedupKey {
--- a/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
+++ b/ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
@@ -57,6 +57,9 @@ message DamlCommandDedupValue {
   // the time until when future commands with the same
   // deduplication key will be rejected due to a duplicate submission
   google.protobuf.Timestamp deduplicated_until = 2;
+  //The submission time of the initial command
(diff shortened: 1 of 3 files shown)


ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
```diff
@@ -57,6 +57,9 @@ message DamlCommandDedupValue {
   // the time until when future commands with the same
   // deduplication key will be rejected due to a duplicate submission
   google.protobuf.Timestamp deduplicated_until = 2;
+  //The submission time of the initial command
+  // We use submission time because record time is not available during pre-execution
+  google.protobuf.Timestamp submission_time = 3;
```
fabiotudone-da, line 62: Nit (I don't think we use this protos to genarate docs btw though):

```suggestion
  // Tthe time until when future commands with the same
  // deduplication key will be rejected due to a duplicate submission.
  google.protobuf.Timestamp deduplicated_until = 2;
  // The submission time of the initial command.
  // We use submission time because record time is not available during pre-execution.
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
```diff
@@ -328,13 +328,14 @@ private[kvutils] class TransactionCommitter(
...
-    val commandDedupBuilder = DamlCommandDedupValue.newBuilder.setDeduplicatedUntil(
-      Conversions.buildTimestamp(
-        transactionEntry.submissionTime
-          .add(parseDuration(transactionEntry.submitterInfo.getDeduplicationDuration))
-          .add(config.timeModel.minSkew)
-      )
+    val deduplicateUntil = Conversions.buildTimestamp(
+      transactionEntry.submissionTime
+        .add(parseDuration(transactionEntry.submitterInfo.getDeduplicationDuration))
+        .add(config.timeModel.minSkew)
```
fabiotudone-da, line 334: Can you add a comment about why we're using the min skew here?

fabiotudone-da (approved):

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
hubert-da: `Tthe` -> `the`

hubert-da (approved):
Same comments as Fabio. Other than that 👍🏻

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/TransactionCommitter.scala
nicu-da: This is taken from the design doc. I could add my understanding of it but I would prefer to leave it as is for now, and add a proper comment in the follow up PR which changes this logic

miklos-da (approved):

PR #11512, opened 2021-11-02
Update to Java 11

### LINKED ISSUE #11065: Migrate to Java 11

aherrmann-da:
Update the Nix imported Java runtime to Java 11.
Update the dadew provided JDK to Java 11.
Update the Java toolchain to Java 11.
Fix deprecation warnings and build errors due to the update to Java 11.

Closes #11065

### Pull Request Checklist

trigger the build.


compatibility/sandbox-migration/src/com/daml/Application.scala
```diff
@@ -58,7 +58,9 @@ object Application {
       TransactionResult(
         transaction.transactionId,
         transaction.events.map(toEventResult),
-        transaction.effectiveAt.map(t => t.seconds * 1000000L + t.nanos.toLong / 1000L).get,
+        transaction.effectiveAt
+          .map(t => t.seconds * 1000000L + t.nanos.toLong / 1000000L * 1000L)
+          .get,
```
aherrmann-da, line 63: The migration tests compare transactions [represented as ](https://github.com/digital-asset/daml/blob/6a637c501e88db5e86ebe7aa05517aac9ac4ddfe/compatibility/sandbox-migration/runner/Migration/Types.hs#L92-L96)
```
data Transaction = Transaction
  { transactionId :: T.Text
  , events :: [Event]
  , letMicros :: !I.Int64    -- :: Ledger effective time, in microseconds since unix epoch
  } deriving (Generic, Eq, Show)
```
The `transactionId` and `events` seem reasonable to compare for equality. But, the ledger effective time is surprising. Should we really expect timestamps to be exactly the same?

It turns out that they no longer are the same when we update to Java 11. The reason seems to be that Java increased the time resolution in some contexts on the update from version 8 to 9, see commit message of https://github.com/digital-asset/daml/pull/11512/commits/af77e1447fb520e7655c2421254ec9011c53f409 and [JDK-8068730](https://bugs.openjdk.java.net/browse/JDK-8068730). The [migration tests fail](https://dev.azure.com/digitalasset/daml/_build/results?buildId=93822&view=logs&j=870bb40c-6da0-5bff-67ed-547f10fa97f2&t=deecee86-545a-596e-8b0d-fb7d606fe9f2&l=114444) due to a difference in precision in these timestamps: In one case we only have mill second precision but in the other microsecond precision:
```
    -      letMicros = 1636122007333866
    +      letMicros = 1636122007333000
```

This change is a simple hack to round down to millisecond precision again. But, the question is, what guarantees about timestamp stability do we really want to give and subsequently test in these migration tests?

cc @cocreature

cocreature: I think I don’t fully understand what is failing. So for this setup, we run everything with JDK11 afaict so this seems to be a build-time change. and if I understand it correctly the precision increased rather than decreased in newer JDK versions.
So shouldn’t we be able to preserve the precision on upgrades?

cocreature: for now, I don’t think this is blocking.

cocreature (approved):
Nice work, thanks! As discussed, only to be merged after SDK 1.18.

ledger/ledger-api-test-tool-on-canton/BUILD.bazel
```diff
@@ -58,7 +58,7 @@ conformance_test(
         "@curl_nix//:bin/curl",
         "@grpcurl_nix//:bin/grpcurl",
         "@jq_dev_env//:jq",
-        "@jdk11_nix//:bin/java",
+        "@bazel_tools//tools/jdk",
```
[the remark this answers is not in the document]
aherrmann-da: Yes, it could perhaps even be simplified to importing canton as a `java_binary` since that would use the correct JDK version as well. Could avoid some of the dance around `$JAVA ...`. There are other instances of this pattern in the code base where that may also be possible.

compatibility/sandbox-migration/src/com/daml/Application.scala
cocreature: After looking into this again it turns out that we’re dropping support for migrating from SDK versions 1.6 in SDK 2.0. So my suggestion would actually be slightly different: Change the migration tests to start at 1.6 and drop the change to adjust precisions.

rautenrieth-da: Sorry I missed the ping. How exactly is the migration test failing? Does the ledger effective time differ when we read the same transaction (via the transaction service) using two different SDK versions?

cocreature: Yes when migrating from SDK 1.0.1 to SDK 1.1.0 (both running under Java 11) the ledger effective time changes. Starting from SDK 1.6.0 the migration succeed without changes to ledger effective time so unless you think differently, I’d suggest we brush this under the carpet with general pre-1.6 migrations.

rautenrieth-da: Are these actual numbers that appear in the migration test? The ledger effective time (LET) should have microsecond precision (like all times in Daml). Returning a LET with nanosecond precision sounds like a bug. We do have old (pre 1.6) java migrations that touch the timestamp column, but I don't see anything that would truncate the timestamps.

```
    -      letMicros = 1636122007333866
    +      letMicros = 1636122007333000
```

I'm fine with brushing this under the carpet:
- There are no Java migrations after SDK 1.6, so changing the JDK version should not affect data in the database.
- As of SDK 1.17, all timestamps are serialized and processed as integers (microseconds since epoch) and we depend much less on `Instant` (and other timestamp handling related bugs in database drivers). 

> Should we really expect timestamps to be exactly the same?

Yes. The ledger effective time timestamp is a constant property of a transaction that is assigned when evaluating the Daml command. It should be as immutable as the contract argument.

cocreature: I believe those are the actual numbers but they are the wrong way around. The numbers seem to have increased in precision. You can find the full transactions that cause the failure at https://gist.github.com/cocreature/054fb8fa171b51f59a14b0dd6644b20d.

So it looks like we stored with higher precision in both cases but now we only emit the lower precision after the upgrade.

rautenrieth-da: Any sub-microsecond precision should be erased with [the migration](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V111__timestamp_to_bigint.sql#L41) to SDK 1.17. It's impossible to store nanosecond-precision timestamps in the 1.17 index database.

If the transactions returned by SDK 1.1.0 contain nanosecond-precision LETs, then I'm surprised the migration to SDK 1.17 did not fail.

cocreature: So I pushed a commit to debug this https://github.com/digital-asset/daml/commit/624a60739ce716bed818906aff9ae5be6e5a9099 and I still get the mismatch even after 1.17. So it doesn’t seem like the migrations on 1.17 blow up nor does it seem like we only emit microsecond precision at least for old transactions (haven’t checked for new ones).

cocreature: I don’t have time to debug this further and I don’t understand the indexer schema but happy to help you run those tests to reproduce, just ping me.

cocreature: Just to avoid confusion: I had an off by 3 orders of magnitude above:
On SDK 1.0.0 we get _mili_second precision, afterwards we get _micro_second precision. We never get _nano_second precision.

We expect to get microsecond precision and it looks like that is always stored in the database. There seems to be a bug in SDK 1.0.0 where things get truncated to milisecond precision on the ledger API.

Given that this doesn’t affect later versions, we agreed with @rautenrieth-da to continue with what I suggested above and bump the starting version in the compat tests to 1.6.

aherrmann-da:
I've added logic to skip versions before 1.6.0 in the migration tests and removed the lte rounding patch.

PR #11515, opened 2021-11-02
DPP-664 Create contracts in batches

rautenrieth-da:
This PR adds batched contract creation - each command submission contains many commands.


ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandDefinition.scala
```diff
@@ -0,0 +1,63 @@
...
+package com.daml.ledger.api.benchtool.submission
+
+import com.daml.ledger.api.v1.commands.Command
+import com.daml.ledger.client.binding.Primitive
+import com.daml.ledger.test.model.Foo.{ContractDefinition, Factory, Foo1, Foo2, Foo3}
+
+import scala.util.Try
+
+// Stores all information required to create a contract using a template defined in Foo.daml
+case class CommandDefinition(
```
rautenrieth-da, line 13: Note: Codegen related code has moved to this class.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandSubmitter.scala
```diff
@@ -163,9 +208,15 @@ case class CommandSubmitter(services: LedgerApiServices) {
 object CommandSubmitter {
   case class CommandSubmitterError(msg: String) extends RuntimeException(msg)
 
-  class ProgressMeter(totalItems: Int, startTimeMillis: Long) {
+  class ProgressMeter(totalItems: Int) {
+    var startTimeMillis: Long = System.currentTimeMillis()
+
+    def start(): Unit = {
```
rautenrieth-da, line 214: Note: explicit start method to move the start of measurement closer to the start of contract submission. There is still a ~0.2sec delay between the start call and the first submission, which leads to overestimated remaining times for the first few progress reports.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandGenerator.scala
```diff
@@ -24,65 +24,70 @@ final class CommandGenerator(
       .toMap
   private val observersWithIndices: List[(Primitive.Party, Int)] = observers.zipWithIndex
 
-  def next(): Try[Primitive.Party => Command] =
+  def nextSingle(): Try[Command] =
```
kamil-da, line 27: This isn't used anywhere. Are we going to support both batched and single modes in the future?

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandDefinition.scala
```diff
@@ -0,0 +1,63 @@
...
+    val observers: List[Primitive.Party] = observerNamesWithIndices
+      .filter { case (_, index) => observerUsed(index) }
+      .map(_._1)
+    templateName match {
+      case "Foo1" => Foo1(_, observers, payload).create.command
+      case "Foo2" => Foo2(_, observers, payload).create.command
+      case "Foo3" => Foo3(_, observers, payload).create.command
+      case invalid => throw new RuntimeException(s"Invalid template: $invalid")
+    }
+  }
```
kamil-da, line 32: Do we use this method of command submission anymore?

ledger/test-common/src/main/daml/model/Foo.daml
```diff
@@ -29,3 +32,52 @@ template Foo3
...
+takeIf (False :: as) (b :: bs) = takeIf as bs
+takeIf [] [] = []
+takeIf _ _ = error("Unbalanced input")
+
+createContract : Party -> [Party] -> ContractDefinition -> Update ()
+createContract signatory observers definition =
+  do
+    case definition.templateId of
+      1 -> do
+        cid <- create Foo1 with signatory=signatory, observers=(takeIf definition.observerUsed observers), payload=definition.payload
```
mziolekda, line 58: I think create+archive done in DAML code will result in a transient contract, but I would live to be proven wrong.

mziolekda: I thought you would create batches on the proto level:
https://github.com/digital-asset/daml/blob/da76e2825b4b1c78766d8db0a4a68359e95c681e/ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto#L58

mziolekda: Then it would be possible to add a CreateAndExerciseCommand that likely would not result in a transient contract.

rautenrieth-da: What is a transient contract? When the archive flag is set, this Daml code creates 1 create and 1 archive event in the index database (just verified on Postgres).

I'll try to rewrite the code using batching at proto level, that sounds much simpler than what I did here.

rautenrieth-da: I have pushed a change that rewrites the code to use batching at proto level. This is slightly slower on my machine, but much easier to maintain.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandSubmitter.scala
```diff
@@ -91,65 +99,90 @@ case class CommandSubmitter(services: LedgerApiServices) {
...
-                ).map(_ => index)
-              case Failure(ex) =>
-                Future.failed {
-                  logger.error(s"Command generation failed. Details: ${ex.getLocalizedMessage}", ex)
-                  CommandSubmitter.CommandSubmitterError(ex.getLocalizedMessage)
+        for {
+          _ <- Source
+            .fromIterator(() => (1 to descriptor.numberOfInstances).iterator)
+            .wireTap(i => if (i == 1) progressMeter.start())
+            .mapAsync(32)(index =>
```
mziolekda, line 158: That's quite an aggresive number, but I guess it doesn't matter that much as this stage will be frequently backpressured by the buffer with maxInFlight limit.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandSubmitter.scala
```diff
@@ -91,65 +99,90 @@ case class CommandSubmitter(services: LedgerApiServices) {
...
+              Future.fromTry(
+                generator.next().map(cmd => index -> cmd)
+              )
+            )
+            .groupedWithin(submissionBatchSize, 1.minute)
+            .map(cmds => cmds.head._1 -> cmds.map(e => e._2).toList)
```
mziolekda, line 164: nit
```suggestion
            .map(cmds => cmds.head._1 -> cmds.map(_._2).toList)
```

mziolekda (approved):
This looks really succinct now that you use proto-level batching!
Thank you

ledger/test-common/src/main/daml/model/Foo.daml
mziolekda: > What is a transient contract? When the archive flag is set, this Daml code creates 1 create and 1 archive event in the index database (just verified on Postgres).

If you were able to see it in the db, it is not a transient contract after all. My worry was unsubstantiated.

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandSubmitter.scala
kamil-da: And possibly add a type description to make reasoning easier:
```suggestion
            .map((cmds: Seq[(Int, Command)]) => cmds.head._1 -> cmds.map(_._2).toList)
```

kamil-da (approved):
👏

ledger/ledger-api-bench-tool/src/main/scala/com/daml/ledger/api/benchtool/submission/CommandSubmitter.scala
rautenrieth-da: Good point, I have changed this to 8.

PR #11539, opened 2021-11-03
Self-service errors: generate error deprecation docs [KVL-1187]

fabiotudone-da:
A new "Deprecated" section is rendered for errors that have `@deprecated` with a message.

[Here's the preview](https://htmlpreview.github.io/?https://raw.githubusercontent.com/fabiotudone-da/kvl-1187-html-preview/main/html/error-codes/self-service/index.html).

### Pull Request Checklist

trigger the build.


fabiotudone-da:
The build fails due to the Scala Reflection API being different in Scala 2.12 and 2.13, looking into that.

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
```diff
@@ -104,38 +106,52 @@ class ErrorCodeDocumentationGenerator(prefixes: Array[String] = Array("com.daml"
...
+          sys.error(
+            s"Unexpected annotation detected (${annotations.map(annotationTypeName)} but the only supported ones are $acceptedTypeNames)."
+          )
     }
 
-    ErrorDocumentationAnnotations(doc.explanation, doc.resolution)
+    ErrorDocumentationAnnotations(doc.deprecation, doc.explanation, doc.resolution)
   }
 
+  private def parseDeprecatedAnnotationValue(tree: ru.Tree): String =
```
hubert-da, line 142: wow

hubert-da (approved):
LGTM, however, you may want to wait for other reviewers

tudor-da (approved):

ledger/error/src/main/scala/com/daml/error/generator/ErrorCodeDocumentationGenerator.scala
fabiotudone-da: Yeah, Java reflection would be easier but it doesn't seem to help with Scala's `@deprecated` (the properties are not accessible).

cocreature (approved):
Scala version guards look fine, we build & push the docs from a 2.13 build so this shouldn’t break anything.

cocreature (approved):
Scala version guards look fine, we build & push the docs from a 2.13 build so this shouldn’t break anything.

fabiotudone-da:
There is still a remaining problem in that we build the generator also as part of the `/ledger/error` library (and libraries are cross-compiled).

cocreature:
> There is still a remaining problem in that we build the generator also as part of the /ledger/error library (and libraries are cross-compiled).

fabiotudone-da:
> If it does need to be in 1.18 happy to have a chat on possible solutions. We can exclude some stuff from being cross-compiled & published in pricniple.

Yes it should be in 1.18, I'm splitting off the generator and that should solve the issue; if not, I'll leave it to you @tudor-da :)

PR #11548, opened 2021-11-04
fix hoogle by pinning nixpkgs

garyverhaegen-da:
Hoogle has been down for at least 24h accoridng to user reports. What
seems to be happening is that our nixpkgs pinning is not taking effect,
and the nixpkgs version of Hoogle already includes the patch we are
trying to add. This confuses nix, which fails, and thus the boot
sequence is broken.

I've applied the minimal possible patch here (i.e. enforce the pin),
which gets things running again. I've already deployed this change.

We may want to look at bumping the nixpkgs snapshot.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
Thanks!

aherrmann-da (approved):

PR #11551, opened 2021-11-04
interfaces: Improve fixed choice lookup

### LINKED ISSUE #11346: Improve efficiently of interface lookup of interfaces

sofiafaro-da:
This PR adds a lazy map to match fixed choice names to their
interface, so we're not doing a linear search over every interface.

Part of #10810
Fixes #11346

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
+++ b/daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
@@ -712,7 +712,12 @@ object Ast {
       observers: E, // Observers of the contract.
       key: Option[GenTemplateKey[E]],
       implements: Map[TypeConName, GenTemplateImplements[E]],
-  )
+  ) {
+    lazy val inheritedChoices: Map[ChoiceName, TypeConName] =
+      implements.iterator.flatMap { case (iface, impl) =>
+        impl.inheritedChoices.iterator.map ( chName => (chName, iface) )
+      }.toMap
+  }
 
   final class GenTemplateCompanion[E] private[Ast] {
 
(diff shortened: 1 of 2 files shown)


daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
```diff
@@ -712,7 +712,12 @@ object Ast {
       observers: E, // Observers of the contract.
       key: Option[GenTemplateKey[E]],
       implements: Map[TypeConName, GenTemplateImplements[E]],
-  )
+  ) {
+    lazy val inheritedChoices: Map[ChoiceName, TypeConName] =
+      implements.iterator.flatMap { case (iface, impl) =>
+        impl.inheritedChoices.iterator.map ( chName => (chName, iface) )
+      }.toMap
```
remyhaemmerle-da: In this case, the first `.iterator` together with the `.toMap` are not useful because there is not intermediary `Map` that is built. 
In general,  when available, prefer `.view` over `iterator` as the later is pure. 

```suggestion
    lazy val inheritedChoices: Map[ChoiceName, TypeConName] =
      implements.flatMap { case (iface, impl) =>
        impl.inheritedChoices.view.map ( chName => (chName, iface) )
      }
```

remyhaemmerle-da (approved):
Nice.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
sofiafaro-da: Thanks!

PR #11552, opened 2021-11-04
Restart the submission interpretation in case of a race [DPP-737]

gerolf-da:
1) Daml interpretation
2) Determine a suitable ledger effective time

interpretation.
now archived.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
@@ -47,7 +47,19 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
   )(implicit
       executionContext: ExecutionContext,
       loggingContext: LoggingContext,
-  ): Future[Either[ErrorCause, CommandExecutionResult]] =
+  ): Future[Either[ErrorCause, CommandExecutionResult]] = {
+    def retryOrError(newCommands: Commands) = {
+      if (retriesLeft > 0) {
+        metrics.daml.execution.retry.mark()
+        logger.info(s"Restarting the computation")
+        loop(newCommands, submissionSeed, ledgerConfiguration, retriesLeft - 1)
+      } else {
+        logger.info(
+          s"Did not find a suitable ledger time after ${maxRetries - retriesLeft} retries. Giving up."
+        )
+        Future.successful(Left(ErrorCause.LedgerTime(maxRetries)))
+      }
+    }
     delegate
       .execute(commands, submissionSeed, ledgerConfiguration)
       .flatMap {
@@ -66,26 +78,31 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
           else
             contractStore
               .lookupMaximumLedgerTime(usedContractIds)
-              .flatMap { maxUsedTime =>
-                if (maxUsedTime.forall(_ <= commands.commands.ledgerEffectiveTime)) {
-                  Future.successful(Right(cer))
-                } else if (!cer.dependsOnLedgerTime) {
-                  logger.debug(
-                    s"Advancing ledger effective time for the output from ${commands.commands.ledgerEffectiveTime} to $maxUsedTime"
+              .flatMap {
+                case Left(missingContracts) =>
+                  logger.info(
+                    s"Did not find the ledger time for some contracts. This can happen if there is contention on contracts used by the transaction. Missing contracts: ${missingContracts
+                      .mkString(", ")}."
                   )
-                  Future.successful(Right(advanceOutputTime(cer, maxUsedTime)))
-                } else if (retriesLeft > 0) {
-                  metrics.daml.execution.retry.mark()
-                  logger.debug(
-                    s"Restarting the computation with new ledger effective time $maxUsedTime"
-                  )
-                  val advancedCommands = advanceInputTime(commands, maxUsedTime)
-                  loop(advancedCommands, submissionSeed, ledgerConfiguration, retriesLeft - 1)
-                } else {
-                  Future.successful(Left(ErrorCause.LedgerTime(maxRetries)))
-                }
+                  retryOrError(commands)
+
+                case Right(maxUsedTime) =>
+                  if (maxUsedTime.forall(_ <= commands.commands.ledgerEffectiveTime)) {
+                    Future.successful(Right(cer))
+                  } else if (!cer.dependsOnLedgerTime) {
+                    logger.debug(
+                      s"Advancing ledger effective time for the output from ${commands.commands.ledgerEffectiveTime} to $maxUsedTime"
+                    )
+                    Future.successful(Right(advanceOutputTime(cer, maxUsedTime)))
+                  } else {
+                    logger.info(
+                      s"Ledger time was used in the Daml code, but a different ledger effective time $maxUsedTime was determined afterwards. The transaction needs to be rerun with the new ledger effective time."
+                    )
+                    val advancedCommands = advanceInputTime(commands, maxUsedTime)
+                    retryOrError(advancedCommands)
+                  }
               }
-              .recover {
+              .recoverWith {
                 // An error while looking up the maximum ledger time for the used contracts
                 // most likely means that one of the contracts is already not active anymore,
                 // which can happen under contention.
@@ -93,12 +110,13 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
                 // Direct references to archived contracts will result in the same error.
                 case error =>
                   logger.info(
-                    s"Lookup of maximum ledger time failed. This can happen if there is contention on contracts used by the transaction. Used contracts: ${usedContractIds
+                    s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
                       .mkString(", ")}. Details: $error"
                   )
-                  Left(ErrorCause.LedgerTime(maxRetries - retriesLeft))
+                  Future.successful(Left(ErrorCause.LedgerTime(maxRetries - retriesLeft)))
               }
       }
+  }
 
   // Does nothing if `newTime` is empty. This happens if the transaction only regarded divulged contracts.
   private[this] def advanceOutputTime(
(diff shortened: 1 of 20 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -47,7 +47,19 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
   )(implicit
       executionContext: ExecutionContext,
       loggingContext: LoggingContext,
-  ): Future[Either[ErrorCause, CommandExecutionResult]] =
+  ): Future[Either[ErrorCause, CommandExecutionResult]] = {
+    def retryOrError(newCommands: Commands) = {
+      if (retriesLeft > 0) {
```
mziolekda, line 52: So there is no reason to assume that the two restart mechanisms will never interleave on the same request.

mziolekda: That doesn't seem to be a problem does it?

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -66,39 +78,42 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
...
+                  if (maxUsedTime.forall(_ <= commands.commands.ledgerEffectiveTime)) {
+                    Future.successful(Right(cer))
+                  } else if (!cer.dependsOnLedgerTime) {
+                    logger.debug(
+                      s"Advancing ledger effective time for the output from ${commands.commands.ledgerEffectiveTime} to $maxUsedTime"
+                    )
+                    Future.successful(Right(advanceOutputTime(cer, maxUsedTime)))
+                  } else {
+                    logger.info(
+                      s"Ledger time was used in the Daml code, but a different ledger effective time $maxUsedTime was determined afterwards. The transaction needs to be rerun with the new ledger effective time."
```
mziolekda, line 99: If we keep on failing on the time-advancement path, we will eventually print the line above i.e. the intent to `rerun with the new ledger effective time`. But then we might decide not to do it inside of `retryOrError` due to exhausted number of trials.

ledger/participant-integration-api/src/main/scala/platform/store/backend/common/ContractStorageBackendTemplate.scala
```diff
@@ -94,30 +82,27 @@ class ContractStorageBackendTemplate(queryStrategy: QueryStrategy) extends Contr
   // TODO append-only: consider pulling up traversal logic to upper layer
   override def maximumLedgerTime(
       ids: Set[ContractId]
-  )(connection: Connection): Try[Option[Timestamp]] = {
-    if (ids.isEmpty) {
-      Failure(emptyContractIds)
```
mziolekda, line 99: Are you saying that an empty list will not happen anymore?

gerolf-da: that's right. not in the types though :(

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
gerolf-da: Yeah, I'll delete the last sentence. The rerun logic will then log the next step anyway.

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
gerolf-da: What do you mean?

mziolekda: I mean a "contract spent in the meantime" and having to "advance the time" can alternate on the same request.
I was just thinking aloud.

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -66,39 +78,42 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
...
+                  logger.info(
+                    s"Did not find the ledger time for some contracts. This can happen if there is contention on contracts used by the transaction. Missing contracts: ${missingContracts
+                      .mkString(", ")}."
                   )
-                  val advancedCommands = advanceInputTime(commands, maxUsedTime)
-                  loop(advancedCommands, submissionSeed, ledgerConfiguration, retriesLeft - 1)
-                } else {
-                  Future.successful(Left(ErrorCause.LedgerTime(maxRetries)))
-                }
+                  retryOrError(commands)
```
tudor-da, line 87: Why are we retrying here? Previously these contracts were available, since interpretation worked. If they some are missing now, how can this not be permanent? 🤔

cocreature: they could come from a contract key assignment which has changed since. on reinterpretation we get the new assignment and things might succeed now

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
```diff
@@ -338,28 +338,18 @@ private[sandbox] final class InMemoryLedger(
...
-    }
+  ): Future[Either[Set[ContractId], Option[Timestamp]]] =
+    Future.fromTry(Try(this.synchronized {
+      contractIds
+        .foldLeft[Either[Set[ContractId], Option[Instant]]](Right(Some(Instant.MIN)))((acc, id) => {
+          val letE = acs.activeContracts.get(id).map(c => Right(c.let)).getOrElse(Left(Set(id)))
+          for {
+            let <- letE
+            acc <- acc
+          } yield acc.map(acc => if (let.isAfter(acc)) let else acc)
```
mziolekda, line 349: Aren't we supposed to collect all the missing ids into the accumulator's left?

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
tudor-da: I see, thanks

ledger/participant-integration-api/src/test/suite/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutorSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+            // in case the usage of getTime with a different LET would result in a different transaction
+            foundEpochPlus5,
+            // The second lookup finds the same ledger time again
+            foundEpochPlus5,
+          ),
+          finalExecutionResult = Right(epochPlus5),
+        )
+      }
+
+      "retry if the contract's LET is in the future and then retry if the contract is missing" in {
```
mziolekda, line 223: Nice, this is precisely what I was looking for

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
tudor-da: Ah, I see that we are retrying on the command interpretation as well. Missed that

ledger/participant-integration-api/src/main/scala/platform/store/cache/MutableCacheBackedContractStore.scala
```diff
@@ -72,36 +73,38 @@ private[platform] class MutableCacheBackedContractStore(
...
-          Success(timestamps + active.createLedgerEffectiveTime)
-        case (Success(_), Some(Archived(_))) => Failure(ContractNotFound(ids))
-        case (Success(_), Some(NotFound)) => Failure(ContractNotFound(ids))
-        case (Success(timestamps), None) => Success(timestamps)
-        case (failure, _) => failure
+      .foldLeft[Either[Set[ContractId], Set[Timestamp]]](Right(Set.empty[Timestamp])) {
+        // successful lookups
+        case (Right(timestamps), (_, Some(active: Active))) =>
+          Right(timestamps + active.createLedgerEffectiveTime)
+        case (Right(timestamps), (_, None)) => Right(timestamps)
```
tudor-da, line 101: Hmm, this one is not right, is it? It was like this before although not sure why. Here we should update the missing as well. We can have `None` in two cases:
* The id pertains to a divulged contract
* The entry has been evicted (very unlikely, since it was just looked up in interpretation - but possible)
wdyt?

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
```diff
@@ -338,28 +338,18 @@ private[sandbox] final class InMemoryLedger(
...
-            acc.map(acc => if (let.isAfter(acc)) let else acc)
-          })
-          .map(Timestamp.assertFromInstant)
-      }))
-    }
+  ): Future[Either[Set[ContractId], Option[Timestamp]]] =
+    Future.fromTry(Try(this.synchronized {
+      contractIds
+        .foldLeft[Either[Set[ContractId], Option[Instant]]](Right(Some(Instant.MIN)))((acc, id) => {
+          val letE = acs.activeContracts.get(id).map(c => Right(c.let)).getOrElse(Left(Set(id)))
```
tudor-da, line 345: If `contractIds` is empty, this function returns `Right(Some(Instant.MIN))` - shouldn't it be `Right(None)`?

tudor-da: I see this is how it was before as well. Fine not to change then.

gerolf-da:
Setting it back to draft. We'd rather go with the exception-based implementation in #11579.

Issue #84, opened 2019-03-14
have `trace` output to be visible in sandbox


francesco-da:

gerolf-da:
This is now fixed. DAML `trace` messages are logged at `DEBUG` level.

PR #11555, opened 2021-11-04
Restructure interface desugaring

akrmn:
DA GHC PR [#84](https://github.com/digital-asset/ghc/pull/84)

### Pull Request Checklist

trigger the build.


compiler/damlc/daml-stdlib-src/DA/Internal/Desugar.daml
```diff
@@ -35,4 +35,19 @@ data NonConsuming t = NonConsuming {}
...
+  fromInterface : i -> Optional t
+  fromInterface = GHC.Types.primitive @"EFromInterface"
+
+  toInterfaceContractId : ContractId t -> ContractId i
+  toInterfaceContractId = GHC.Types.primitive @"EToInterfaceContractId"
+
+  fromInterfaceContractId : ContractId i -> Update (Optional (ContractId t))
+  fromInterfaceContractId = GHC.Types.primitive @"UFromInterfaceContractId"
+
+  interfaceTypeRep : forall p. p i -> t -> TypeRep
```
akrmn, line 52: I added the `p i` argument since otherwise the type of the method would be ambiguous. I first thought of using `Proxy` but it seems that's not a thing in Daml. Alternatively, we could just enable `AllowAmbiguousTypes` here, but I thought I'd rather avoid the extension here.

compiler/damlc/daml-stdlib-src/DA/Internal/Desugar.daml
```diff
@@ -35,4 +35,19 @@ data NonConsuming t = NonConsuming {}
 data PreConsuming t = PreConsuming {}
 data Consuming t = Consuming {}
 data PostConsuming t = PostConsuming {}
-data Implements t i = Implements
+
+class Implements t i where
+  toInterface : t -> i
+  toInterface = GHC.Types.primitive @"EToInterface"
```
akrmn, line 41: Note that I simply added the implementations here since they're always the same set of primitives. This way, the generator only needs to make empty instances. However, please let me know if there's a good reason not to do it this way

akrmn: ah, it seems that some of the primitives only work with concrete types, so I'll go back to the class methods being defined by the instances.

https://github.com/digital-asset/daml/blob/21c97453bd5370bfaeec69afece2ec56659ef304/compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion/Primitives.hs#L458-L487

cocreature: yeah one of the reasons for going via the class dance is to be able to use the concrete types in the instances.

compiler/damlc/daml-stdlib-src/DA/Internal/Desugar.daml
cocreature: We use AllowAmbiguousTypes in a bunch of places so I wouldn’t really worry about that. As long as users don’t have to enable it, you should be good.

akrmn: Fair! I just put the old type back now

compiler/damlc/tests/daml-test-files/InterfaceDesugared.daml
```diff
@@ -7,84 +7,204 @@
 
 -- | The "desugared" counterpart to Interface.daml
 -- Please keep in sync with Interface.daml
+-- You can compare this file with the output of
+-- > bazel run damlc -- compile $REPO/compiler/damlc/tests/daml-test-files/Interface.daml --target=1.dev --ghc-option=-ddump-parsed > desugared.out
```
cocreature, line 11: Definitely not for this PR:
Somehow this seems to be getting out of hand. I don’t have a fantastic idea for what to do about it though. At least at this stage it seems very useful to be able to iterate on the desugaring quickly so this is helpful but keeping it up2date is super annoying and it’s becoming so large that it’s hard to maintain.
Maybe we can just use the simplest interface we can possibly come up with?

wdyt @sofiafaro-da @akrmn?

cocreature (approved):
Great work, thank you so much!

compiler/damlc/daml-stdlib-src/DA/Internal/Desugar.daml
```diff
@@ -28,11 +31,24 @@ import DA.Internal.Exception ()
...
+class Implements t i where
+  toInterface : t -> i
+  fromInterface : i -> Optional t
+  toInterfaceContractId : ContractId t -> ContractId i
+  fromInterfaceContractId : ContractId i -> Update (Optional (ContractId t))
+  interfaceTypeRep : t -> TypeRep
+
+class HasMethod i (m : Symbol) r | i m -> r
+
+newtype Method t i (m : Symbol) r = Method { unMethod : t -> r }
```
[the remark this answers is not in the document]
akrmn: will do!

compiler/damlc/daml-stdlib-src/Prelude.daml
```diff
@@ -25,3 +25,4 @@ import DA.Internal.Template.Functions as X
 #endif
 import DA.Internal.Compatible as X
 import DA.Internal.Assert as X
+import DA.Internal.Desugar as X (Implements (..))
```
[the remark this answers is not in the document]
akrmn: perhaps even a new `DA.Internal.Interface`?

compiler/damlc/tests/daml-test-files/Interface.daml
```diff
@@ -57,19 +57,19 @@ template Asset
     implements Token where
       let getOwner = owner
       let getAmount = amount
-      let setAmount = \x -> toToken (this with amount = x)
+      let setAmount = \x -> toInterface @_ @Token (this with amount = x)
```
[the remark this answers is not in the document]
akrmn: tbh I just search-and-replace'd `toToken` and similar, but I agree that type inference should do the right thing here, though I'd still keep the type app for the reader's sake. But yeah we should definitely reorder the tyvars

compiler/damlc/daml-stdlib-src/Prelude.daml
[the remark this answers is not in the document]
sofiafaro-da: I like the idea of moving the `Implements` typeclass to a new `DA.Internal.Interface`. This will be necessary for it to show up in the stdlib docs.

The other functions/types (`mkMethod`, `Method`, `HasMethod`) should probably stay in `DA.Internal.Desugar`, they're only needed for desugaring.

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -1138,7 +1162,7 @@ internalTypes = mkUniqSet
     ]
 
 desugarTypes :: UniqSet FastString
-desugarTypes = mkUniqSet ["Consuming", "PreConsuming", "PostConsuming", "NonConsuming", "Implements"]
+desugarTypes = mkUniqSet ["Consuming", "PreConsuming", "PostConsuming", "NonConsuming"]
```
sofiafaro-da, line 1165: I think you can add `"Method"` in here. We don't really want `DA.Internal.Desugar.Method` ending up in LF.

sofiafaro-da: Oh, and I think `"HasMethod"` too.

akrmn: ah well spotted, thanks!

compiler/damlc/daml-lf-conversion/src/DA/Daml/LFConversion.hs
```diff
@@ -525,20 +535,59 @@ convertModule lfVersion pkgMap stablePackages isGenerated file x depOrphanModule
...
+          , Just mod <- [nameModule_maybe (getName iface)]
+          , Var (NameIn DA_Internal_Desugar "mkMethod")
+            `App` Type _tpl
+            `App` Type _iface
+            `App` Type _methodName
+            `App` Type _methodType
+            `App` _implDict
+            `App` _hasMethodDict
+            `App` body
+              <- [untick val]
```
akrmn, line 567: If you noticed me frowning during the meeting, it was because of this! 

The problem was that in the `InterfaceDesugared` test, the rhss of the `_metod_<...>` bindings got an extra `Tick` node, so this list comprehension was returning the empty list

sofiafaro-da: Typical ghc mischief!

Issue #10915, opened 2021-09-16
Consider preserving "fetched by interface" info in Fetch and Exercise nodes


cocreature:

sofiafaro-da:
If we only have fixed choices, the choice name implies already whether something was exercised by interface. (If it's a fixed choice, it's always exercised by interface. Otherwise, it is a template-specific choice.)

cocreature:
As discussed, while you’re completely right preserving the information means we can make the distiniction locally rather than having to look in the package which generally fits in well with how we handle other things like LF versions so we should follow that approach here.

sofiafaro-da:
We can also add `by_interface` in create nodes. I'm not sure that it's required for validation (even taking projections into account) but it might be good to keep track of it anyway.

cocreature:
I’m in favor of storing it there as well. I don’t think we need it for validation but better store more than less.

Issue #10917, opened 2021-09-16
Test coverage for Daml interfaces


cocreature:
- [ ] #11698
- [x] Scala Daml-LF decoder unit tests
- [x] Scala Daml-LF parser implement & add test
- [x] Scala Daml-LF encoder implement & add test
- [ ] Maybe tests at the speedy level
- [ ] #13332
- [x] Tests for engine preprocessor
- [x] Tests for `by_interface` in the transaction coder.
- [ ] #13333
- [ ] #13334
- [ ] #13335
- [x] Test Typing of requires fields (in particular for check errors `EMissingRequiredInterface`, `EWrongInterfaceRequirement`, `ENotClosedInterfaceRequires`).

sofiafaro-da:
Closing this issue in favor of all the smaller ones.

PR #11576, opened 2021-11-08
Add by_interface field in Create, Fetch, & Exercise transaction nodes.

sofiafaro-da:
This PR adds an optional `by_interface` field in the Create, Fetch,
and Exercise nodes, as part of #10915. It also updates TransactionCoder
to support these fields.

Setting these fields (in the LF interpreter) is left to a separate PR.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
```diff
@@ -515,6 +524,12 @@ object TransactionCoder {
             nodeVersion,
             protoCreate.getKeyWithMaintainers,
           )
+          byInterface <-
+            if (protoCreate.hasByInterface) {
```
cocreature, line 528: should we check the version here?

cocreature: same question for the ones below.

cocreature (approved):
Nice! We should add some tests for that both for the transacton decoder/encoder and maybe some more integration-type test that calls speedy and checks the transaction. If you don’t want to do that in that PR please add it to https://github.com/digital-asset/daml/issues/10917

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/TransactionCoder.scala
sofiafaro-da: Added!

PR #11579, opened 2021-11-08
Restart the submission interpretation in case of a race [DPP-737]

gerolf-da:
This is a smaller-surface-area change with the same goal as #11552.

1. Daml interpretation
2. Determine a suitable ledger effective time

interpretation.
now archived.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
@@ -98,7 +98,7 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
                   } else {
                     logger.info(
                       s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
-                        .mkString(", ")}."
+                        .mkString("[", ", ", "]")}."
                     )
                     Future.successful(Left(ErrorCause.LedgerTime(maxRetries)))
                   }
@@ -108,7 +108,7 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
                 case error =>
                   logger.info(
                     s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
-                      .mkString(", ")}. Details: $error"
+                      .mkString("[", ", ", "]")}. Details: $error"
                   )
                   Future.successful(Left(ErrorCause.LedgerTime(maxRetries - retriesLeft)))
               }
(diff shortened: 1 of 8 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -85,18 +85,32 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
...
+                    metrics.daml.execution.retry.mark()
+                    logger.info(
+                      s"Some input contracts could not be found. Restarting the computation. Missing contracts: ${contracts
+                        .mkString("[", ", ", "]")}"
+                    )
+                    loop(commands, submissionSeed, ledgerConfiguration, retriesLeft - 1)
+                  } else {
+                    logger.info(
+                      s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
+                        .mkString(", ")}."
```
mziolekda, line 101: Could you unify the way that the contract lists are printed in the different log-lines produced by this method?

gerolf-da: Good point. Sorry about that. I like the `[a, b, c]` notation, because it clearly shows empty lists. I'll update the other occurrences.

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
```diff
@@ -339,27 +340,19 @@ private[sandbox] final class InMemoryLedger(
...
-          .map(Timestamp.assertFromInstant)
-      }))
-    }
+    Future.fromTry(Try(this.synchronized {
+      contractIds
+        .foldLeft[Option[Instant]](Some(Instant.MIN))((acc, id) => {
+          val let = acs.activeContracts
+            .getOrElse(
+              id,
+              throw MissingContracts(Set(id)),
```
mziolekda, line 349: This exception will only ever contain a single contract id even if multiple are missing. 
I am ok with leaving it this way, as the in-memory ledger will disappear this year and it is not worth fixing it.

mziolekda (changes requested):
Requested one small correction

ledger/sandbox-classic/src/main/scala/platform/sandbox/stores/ledger/inmemory/InMemoryLedger.scala
gerolf-da: Yes, this retains the same behavior as before. We only use the contractIds for logging. For the sake of the behavior, it's enough that the exception is thrown.

mziolekda (approved):

tudor-da (approved):

PR #11588, opened 2021-11-08
[DPP-739][Self-service error codes] Adopt QueueBasedTracker

pbatko-da:
1. `QueueBackedTracker#track` is creating (via error factories) `StatusRuntimeException`s rather than creating `io.grpc.Status`es.
2. `ApiCommandService` uses those `StatusRuntimeException`s.
3. Switch between legacy (V1) ans self-service (V2) error codes as usual inside error factories.
4. Created some new error codes.
5. Created new error category for RESOURCE_EXHAUSTED.

### DIFF
--- a/daml-script/test/daml/MultiTest.daml
+++ b/daml-script/test/daml/MultiTest.daml
@@ -6,6 +6,7 @@
 module MultiTest where
 
 import DA.List
+import DA.Time
 import Daml.Script
 
 template T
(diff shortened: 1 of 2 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/services/tracking/QueueBackedTracker.scala
```diff
@@ -39,49 +40,42 @@ private[services] final class QueueBackedTracker(
...
         case QueueOfferResult.Enqueued =>
           trackedPromise.future.map(
-            _.left.map(completionFailure => QueueCompletionFailure(completionFailure))
+            _.left.map(completionFailure =>
+              QueueCompletionFailure(completionFailure.statusRuntimeException)
+            )
           )
         case QueueOfferResult.Failure(t) =>
-          failedQueueSubmission(
-            GrpcStatus.ABORTED
```
pbatko-da, line 53: Internal??

PR #10353, opened 2021-07-21
Support deletion of a large number of contracts

cocreature:
fixes #10339

There are two orthogonal issues here:

1. scalaz’s toVector from the Foldable[Set] instance
   stackoverflows. I’ve just avoided using that altogether.
2. Oracle doesn’t like more than 1k items in the IN clause. I chunked
   the queries into chunks of size 1k to fix this.

changelog_begin

- [JSON API] Fix an error where transactions that delete a large
  number of contracts resulted in stackoverflows with the PostgreSQL
  backend and database errors with Oracle.

changelog_end

### Pull Request Checklist

- [ ] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/main/CONTRIBUTING.md)
- [ ] Include appropriate tests
- [ ] Set a descriptive title and thorough description
- [ ] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [ ] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #3936, opened 2020-01-01
websocket variant of query endpoint

S11001001:
Fixes #3880, fixes #3878. Obsoletes #3912 (by incorporation of most of its parts) for the purposes of the JSON API itself.

Major changes versus #3912 :

1. Input interface matches `/contracts/search` exactly. That means no empty queries are allowed, and the filtering part is actually used.
2. Endpoint renamed `/contracts/searchForever` to emphasize that this is not a stream of transactions, as discussed in #3878.
3. Output objects are of the format `{errors, created, archived}`, where all members are lists, rather than containing one contract per output object.

```rst
CHANGELOG_BEGIN

- [JSON API - Experimental] Websocket contract search at ``/contracts/searchForever``.
  See `issue #3936 <https://github.com/digital-asset/daml/pull/3936>`_.

CHANGELOG_END
```

* [x] detach WebsocketServiceIntegrationTest from tests added by AbstractHttpServiceIntegrationTest
* [x] ensure that nontermination is happening
* [x] longer test: create, check, exercise, check adds/removes
* [x] include output in documentation

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Include changelog additions in one or more commit message bodies between the `CHANGELOG_BEGIN` and `CHANGELOG_END` tags, if appropriate
- [ ] Normal production system change, include purpose of change in description

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #3192, opened 2019-10-15
json-api: store contracts similar to Extractor's single-table

S11001001:
Fixes #2781. This PR only _stores_ the contracts in database, it does not _search_ that database.

### TODOs

* [x] command-line options for PG connection (see notes below)
* [x] init DB if absent (`initDatabase`)
* [x] catch up to new offset with creates/archives
* [x] update release notes (workflowId field removed from JSON API)
* [ ] split issue for handling rollback (key conflicts, offset overlap)
* [ ] split issue for compiling search

### Request flow:

1. Accept party auth, set of template IDs with request
1. For each template ID:
    1. retrieve that (party, ID)’s last ending offset (`surrogateTemplateId`, `lastOffset`)
    1. retrieve all contracts since last offset for that (party, ID) (`insertContracts`)
    1. save new offset from end of retrieved contracts (`updateOffset`)
    1. run query against database
1. Transcode all results to proper output format (e.g. string numbers)

#### command-line options for PG connection
```
bazel run //ledger-service/http-json:http-json-binary  -- --ledger-host=localhost --ledger-port=6865 \
--http-port=7575 --package-reload-interval 15s \
--query-store-jdbc-config=driver=org.postgresql.Driver,url=jdbc:postgresql://localhost:5432/postgres?ssl=false,user=postgres,password=mysecretpassword,createSchema=true
```

### Pull Request Checklist

- [x] Read and understand the [contribution guidelines](https://github.com/digital-asset/daml/blob/master/CONTRIBUTING.md)
- [x] Include appropriate tests
- [x] Set a descriptive title and thorough description
- [x] Add a reference to the [issue this PR will solve](https://github.com/digital-asset/daml/issues), if appropriate
- [x] Add a line to the [release notes](https://github.com/digital-asset/daml/blob/master/unreleased.rst), if appropriate

NOTE: CI is not automatically run on non-members pull-requests for security
reasons. The reviewer will have to comment with `/AzurePipelines run` to
trigger the build.

PR #11589, opened 2021-11-08
limit contract insertion/deletion batching on backpressure

S11001001:
#3936 uses a limit on batching websocket output under the assumption that clients would rather download their deltas in bite-sized chunks than very large ones.  It was inspired by #3192, which also batches transactions to be added to the query store, albeit with no limit.

While #10353 updates #3192 to support arbitrarily large batches as input to the database, we might as well defensively limit ourselves to the largest batch likely to yield any meaningful RDBMS roundtrip reduction time, using a similar technique to #3936.  This will also feed more backpressure to the participant server, possibly reducing its own load in high-load scenarios.

```rst
CHANGELOG_BEGIN
- [JSON API] While updating the contract table for a query, if the DB appears to be slow,
  JSON API will slow down its own inserts and deletes at some point rather than construct
  ever-larger INSERT and DELETE batch commands.
  See `issue #11589 <https://github.com/digital-asset/daml/pull/11589>`__.
CHANGELOG_END
```

### Pull Request Checklist

trigger the build.


ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
```diff
@@ -406,6 +407,16 @@ private[http] object ContractsFetch {
     }.void
   }
 
+  private def conflation[D, C: InsertDeleteStep.Cid]
+      : Flow[InsertDeleteStep[D, C], InsertDeleteStep[D, C], NotUsed] = {
+    // when considering this cost, keep in mind that each deleteContracts
+    // may entail a table scan.  Backpressure indicates that DB operations
+    // are slow, the idea here is to set the DB up for success
+    val maxCost = 250L
```
stefanobaghino-da, line 415: Should this be configurable, rather than hard-coded?

stefanobaghino-da (approved):
@S11001001 Should this be considered for back-porting to prevent issues for users running older versions?

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
S11001001: No, I don't think so. The benefit comes from setting a [reasonably large] limit, more than what that limit is.

S11001001:
> Should this be considered for back-porting to prevent issues for users running older versions?

I wrote it with that possibility in mind (else, the conflation function would be in `fetch-contracts`).

akshayshirahatti-da:
looks good 👍

ledger-service/http-json/src/main/scala/com/digitalasset/http/ContractsFetch.scala
akshayshirahatti-da: @S11001001 how did we come up with this `maxCost` value , also can you remind me of the semantics of delete and why it may entail a table scan ?

S11001001: We came up with it by me taking the number 200 from the client output and adding 50 to it. It's a buffer size; it doesn’t really matter what the number is so much as that there is a number and it is reasonably "large" so that the buffer doesn't grow without bound but still gives you good-sized "chunks".

As for the deletes, it depends on how the DB implements `delete from contracts where contract_id in (id1, id2, ...)`.

PR #11606, opened 2021-11-09
Set symlink_node_modules = False

aherrmann-da:
By default `yarn_install` will install `node_modules` in the repository.
The advantage is that editor tooling and the like can use this
`node_modules` tree without needing to duplicate it between Bazel and
tooling. The downside is that Bazel has less strict control over this
tree. In particular, on refetch it can happen that old files remain in
the tree and confuse the build going forward.

For example, we have occasionally observed errors of the form
```
warning Cannot find a suitable global folder. Tried these: "/usr/local, /does-not-exist/.yarn"
 ERROR  The following files have the same case insensitive path, which isn't supported by the VSIX format:
  - extension/node_modules/form-data/Readme.md
  - extension/node_modules/form-data/README.md
```

Setting this flag to `False` means that `node_modules` will instead be
installed under Bazel's output base and Bazel can make sure to clear the
tree before a refetch.

See [docs](https://github.com/bazelbuild/rules_nodejs/blob/9f8a18505f839f1c5c9bb7e69040d695aac677d5/internal/npm_install/npm_install.bzl#L330)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,11 +1,5 @@
 workspace(
     name = "com_github_digital_asset_daml",
-    managed_directories = {
-        "@npm": ["node_modules"],
-        "@daml_extension_deps": ["compiler/daml-extension/node_modules"],
-        "@navigator_frontend_deps": ["navigator/frontend/node_modules"],
-        "@language_support_ts_deps": ["language-support/ts/packages/node_modules"],
-    },
 )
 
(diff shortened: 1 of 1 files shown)


remyhaemmerle-da (approved):
:crossed_fingers:

PR #11607, opened 2021-11-09
Provide byInterface fields for fetch, exercise in speedy.

sofiafaro-da:
Part of #10915

Create is left for a separate PR, since it requires adding a new LF primitive.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
@@ -1030,7 +1030,13 @@ private[lf] final class Compiler(
         _env.bindExprVar(tmpl.param, tmplArgPos).bindExprVar(choice.argBinder._1, choiceArgPos)
       let(
         env,
-        SBUBeginExercise(tmplId, choice.name, choice.consuming, byKey = mbKey.isDefined)(
+        SBUBeginExercise(
+          tmplId,
+          choice.name,
+          choice.consuming,
+          byKey = mbKey.isDefined,
+          byInterface = None,
+        )(
           env.toSEVar(choiceArgPos),
           env.toSEVar(cidPos),
           compile(env, choice.controllers),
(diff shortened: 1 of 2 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -1063,7 +1069,7 @@ private[lf] final class Compiler(
       val env = _env.bindExprVar(param, payloadPos).bindExprVar(choice.argBinder._1, choiceArgPos)
       let(
         env,
-        SBResolveSBUBeginExercise(choice.name, choice.consuming, byKey = false)(
+        SBResolveSBUBeginExercise(choice.name, choice.consuming, byKey = false, ifaceId = ifaceId)(
```
cocreature, line 1072: is this sufficient for ledger API commands with an interface id rather than just the LF exercise by interface? If not, we should either address this in this PR or in a follow-up.

cocreature (approved):
Looks good, thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
sofiafaro-da: This is just for speedy. This doesn't concern ledger API commands at all?

sofiafaro-da: Fixed/inherited choices should always have `byInterface` set, even if initially invoked at template id.
I think this is all working correctly wrt remy's PR to support inherited choice exercise by template in the ledger API.

remyhaemmerle-da (approved):

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
cocreature: the speedy compiler is also used to compile ledger API commands via `compileCommands`. I _think_ it works correctly but haven’t checked. probably worth tackling at least some minimal tests for this sooner rather than later.

remyhaemmerle-da: My PR is still not merge, so a bit tricky to test. ;)

remyhaemmerle-da: It is actually just merged

PR #11609, opened 2021-11-09
kvutils: Use consistent argument name casing in KVErrors

miklos-da:
Summary of changes:
* Use camelCased argument names consistent with the Scala style guide.

### Pull Request Checklist

trigger the build.


nicu-da (commented):
This all looks good to me. 
I haven't checked though how does this interacts with the reflection code that we have. Isn't it going to delete some of the current logic that we have? Or include the field twice? Old name and new name

miklos-da:
> This all looks good to me. I haven't checked though how does this interacts with the reflection code that we have. Isn't it going to delete some of the current logic that we have? Or include the field twice? Old name and new name

It's going to override it, i.e., we won't use reflection to generate the default context.

stefanobaghino-da (approved):

nicu-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/errors/KVErrors.scala
```diff
@@ -62,13 +62,18 @@ object KVErrors extends LedgerApiErrorGroup {
...
+          minimumRecordTime: Instant,
+          maximumRecordTime: Instant,
       )(implicit loggingContext: ContextualizedErrorLogger)
           extends KVLoggingTransactionErrorImpl(
             cause =
-              s"Invalid ledger time: Record time is outside of valid range [$minimum_record_time, $maximum_record_time]"
-          )
+              s"Invalid ledger time: Record time is outside of valid range [$minimumRecordTime, $maximumRecordTime]"
+          ) {
+        override def context: Map[String, String] = Map(
```
SamirTalwar-DA, line 72: It really concerns me that we could easily forget to override `context` for a particular error. I honestly think that having the logic in `BaseError` is dangerous.

To mitigate this, I suggest adding `override def context = kvContext` in `KvLoggingTransactionErrorImpl` and making `kvContext` an abstract method that _must_ be overridden by subclasses.

hubert-da (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/errors/KVErrors.scala
miklos-da: I share your concern, especially that the logic in `BaseError` is not transparent.
I have made changes similar to as you suggested, with the slight change that `KvLoggingTransactionErrorImpl` now explicitly overrides `context` with an empty `Map` so subclasses must override it to produce any error metadata.
Does this work for you?

SamirTalwar-DA: Yes, sounds good!

SamirTalwar-DA (approved):

PR #11610, opened 2021-11-09
Draft of Ledger API v2 services

meiersi-da:
Intended as
- a basis for discussion across teams
- a means to double-check that we could avoid Daml-LF-version-bound limits on transaction subtree sizes
- see the `v2/DESIGN_NOTES.md` file for details

### Pull Request Checklist

trigger the build.


ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+import "com/daml/ledger/api/v1/transaction_filter.proto";
+
+
+option java_outer_classname = "EventsServiceServiceOuterClass";
+option java_package = "com.daml.ledger.api.v1";
+option csharp_namespace = "Com.Daml.Ledger.Api.V1";
+
+
+service EventsService {
+  // Get the address of a recent transaction as of which an event source can be provided.
```
andreaslochbihler-da, line 45: Does it have to be a transaction? Is "event" not good enough? I imagine the case where there are a lot of non-transaction events and we may want to resubscribe from in between.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+
+    // TODO: what kind of type do we need to support multi-domain transactions?
+    // Could the below work?
+    TransactionGraphOffset transaction_graph_offset = 2;
+  }
+}
+
+message TransactionGraphOffset {
+  // Party-id to record_time vector clock.
+  map<string, uint64> vector_clock = 1;
```
andreaslochbihler-da, line 71: A vector clock itself is a map from some index set (in Canton domain IDs) to a linearly ordered set (64bit integers in Canton). So this should be `map<string, map<string, uint64>>` at the very least.

Unrelatedly, I tend to avoid `map` in Protobuf and instead use a `repeated PartyToVectorClock` field instead with a custom tuple type `PartyToVectorClock`, for two reasons:
1. It is clear from the field names what the `string` is, namely a party.
2. The code to generate the protobuf message has more control over how the "map" is serialized.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+    string transaction_log_offset = 1;
+
+    // TODO: what kind of type do we need to support multi-domain transactions?
+    // Could the below work?
+    TransactionGraphOffset transaction_graph_offset = 2;
+  }
+}
+
+message TransactionGraphOffset {
+  // Party-id to record_time vector clock.
```
andreaslochbihler-da, line 70: :clap: I think there is some benefit of exposing vector clocks by party id, but only if the application wants to deal with causality itself. Otherwise, we should also offer a causally consistent mode for the parties in the filter of the stream subscription.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  map<string, uint64> vector_clock = 1;
+}
+
+// A participant-wide address of an event and the state it may depend on.
+// Comparisons are done in lexicographic order of the fields.
```
andreaslochbihler-da, line 75: This only makes sense if `transaction_offset` is a `transaction_log_offset`. You can't sensibly compare vector clocks (at least not if you want a total order, which is the usual expectation).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+message EventOffset {
+  TransactionOffset transaction_offset = 1;
+  LocalEventOffset local_event_offset = 2;
+}
+
+// A transaction-local address of an event.
+message LocalEventOffset {
+  // Equivalent to v1 node-id within a transaction.
+  // TODO: do we need something more powerful?
```
andreaslochbihler-da, line 84: How stable do you intend these local offsets to be across parties? Suppose there's a transaction `tx` with nodes 1, 2, 3 where nodes 1 and 3 are visible to `Alice` and `1, 2` are visible to Bob. Ideally, Alice shouldn't learn that there is exactly one node `2` that is not visible to her. So on Alice's event stream, we should have the local offsets 1 and 2, and on the stream for Alice and Bob together we have three local offsets 1, 2, 3.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // The offset of this event, which is strictly monotonically increasing
+  // for consecutive events wrt the order on `EventOffset`s.
+  //
+  // Thus all events from the same transaction will be served consecutively
+  // and be assigned the same `TransactionOffset`.
+  EventOffset event_offset = 1;
+
+  // The event at this offset provided the event matches the filter.
+  Event event = 2;
+}
```
andreaslochbihler-da, line 138: I'd also add the delta on the vector clocks that the event is about. It's not yet clear to me though what to do about vector clock updates of skipped events (i.e., those removed by the event projection).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+
+// If no projection is set for a specific kind of event, then none of these events will be served.
+//
+// Use the default `EventProjection` to subscribe to just the changes of the ledger end.
+//
+// Note that the projections generally provide an opinionated set of of fields only.
+message EventProjection {
+  TransactionCommittedEventProjection transaction_committed_event_projection = 1;
+  ContractCreatedEventProjection contract_created_event_projection = 2;
+  ChoiceExercisedEventProjection exercised_event_projection = 4;
```
andreaslochbihler-da, line 151: What about `fetch`/`lookupByKey` nodes? What about other kinds of events like party allocation, package upload and vetting?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Optional
+  string workflow_id = 3;
+
+  // Ledger effective time.
+  // Required
+  google.protobuf.Timestamp effective_at = 4;
+
+  // TODO: potentially missing
+  // string application_id = 6;
+  // repeated string submitters = 7;
```
andreaslochbihler-da, line 219: Is `TransactionCommittedEvent` supposed to subsume the accepting completion info? If so, then we should include all the Completion fields, i.e., application id, submitters, deduplication period, submission id, etc.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Required
+  google.protobuf.Timestamp effective_at = 4;
+
+  // TODO: potentially missing
+  // string application_id = 6;
+  // repeated string submitters = 7; 
+}
+
+
+// Records that a contract has been created, and choices may now be exercised on it.
```
andreaslochbihler-da, line 223: If we want to output events in causal order, we should think about how to separate the announcement of "contract X is now active and can be exercised" from "this is the transaction that created this contract".

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  repeated string witness_parties = 5;
+
+  // The signatories for this contract as specified by the template.
+  // Required
+  repeated string signatories = 8;
+
+  // The observers for this contract as specified explicitly by the template or implicitly as choice controllers.
+  // This field never contains parties that are signatories.
+  // Required
+  repeated string observers = 9;
```
andreaslochbihler-da, line 264: Given that you're designing a new ledger API, I'd take the liberty to clean this up so that we don't have to repeat the same party in several of these fields.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // We use StringValue to properly reflect optionality on the wire for backwards compatibility.
+  // This is necessary since the empty string is an acceptable (and in fact the default) agreement
+  // text, but also the default string in protobuf.
+  // This means a newer client works with an older sandbox seamlessly.
+  // Optional
+  google.protobuf.StringValue agreement_text = 6;
+
+  // The time as of which this contract is present on the ledger.
+  google.protobuf.Timestamp ledger_effective_time = 10;
+}
```
andreaslochbihler-da, line 276: I'd also add another field for further contract-specific data that is not part of the Daml model (e.g., the nonce for opening the commitment in an authenticating contract ID).

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // If true, the target contract may no longer be exercised.
+  // Required
+  bool consuming = 8;
+
+  reserved 9; // removed field
+
+  // The parties that are notified of this event. The witnesses of an exercise
+  // node will depend on whether the exercise was consuming or not.
+  // If consuming, the witnesses are the union of the stakeholders and
+  // the actors.
```
andreaslochbihler-da, line 320: What about choice observers?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // If not consuming, the witnesses are the union of the signatories and
+  // the actors. Note that the actors might not necessarily be observers
+  // and thus signatories. This is the case when the controllers of a
```
andreaslochbihler-da, line 323: I find the comment `Note that the actors might not necessarily be observers
  // and thus signatories.` confusing. Observers are in general not signatories anyway.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // choice are specified using "flexible controllers", using the
+  // ``choice ... controller`` syntax, and said controllers are not
+  // explicitly marked as observers.
+  // Each element must be a valid PartyIdString (as described in ``value.proto``).
+  // Optional for ARCHIVALS_ONLY
+  repeated string witness_parties = 10;
+
+  reserved 11; // was repeated string child_event_ids = 11;
+
+  // An upper bound that is larger or equal to the largest `LocalEventOffset` of any descendant event.
```
andreaslochbihler-da, line 333: For privacy reasons, I'd prefer to put "equal to" here. See the above comment on event offsets with Alice and Bob

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  // Optional for ARCHIVALS_ONLY
+  LocalEventOffset descendant_event_offset_upper_bound = 14;
+
+  // The result of exercising the choice
+  // Optional for ARCHIVALS_ONLY
+  Value exercise_result = 12;
+
+  // The time as of which this choice is effective on the ledger.
+  // Required.
+  google.protobuf.Timestamp ledger_effective_time = 13;
```
andreaslochbihler-da, line 343: I think we have renamed "ledger-effective time" into "ledger time" a long time ago in our docs. Now would be a good time to pull this change through in the protobuf definitions as well.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
...
+  TransactionCommittedEventProjection transaction_committed_event_projection = 1;
+  ContractCreatedEventProjection contract_created_event_projection = 2;
+  ChoiceExercisedEventProjection exercised_event_projection = 4;
+}
+
+// No fine-grained projection available for now.
+message TransactionCommittedEventProjection {}
+
+// No fine-grained projection available for now. 
+message ContractCreatedEventProjection {
```
andreaslochbihler-da, line 158: Just an idea: How about being able to specify what fields in the events should be filled in. If an application isn't interested in `ExerciseResult`, then there's little point in reading it from the database and shipping it to the application over the network.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
```diff
@@ -0,0 +1,364 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+
+/*
+Satisfied Design goals:
+- can co-exist with existing API services for single-domain participants
+- can be served with the 1.17 IndexDB
```
andreaslochbihler-da, line 8: I don't think this should be a design goal for a major version of the ledger API. With v2, we should get to the point of explicit information (https://docs.google.com/document/d/1c1svk_JD1-KTjNjla-xVduJKQJAb37qnsyTzr-7_dek/edit#heading=h.yowjto3ekggr)

ledger-api/grpc-definitions/com/daml/ledger/api/v1/events_service.proto
meiersi-da: Discussion outcome: cleanup language around transaction, action, node, event.

meiersi-da:
Closing, as no longer required.

PR #11612, opened 2021-11-09
Cache type synonym expansion in safeToReexport

cocreature:
Not really significant but seems better to be on the safe side and
keep this consistent with isDuplicate than try to do something
different here.

Needs rebasing on #11611 once that is merged.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
+++ b/compiler/damlc/daml-compiler/src/DA/Daml/Compiler/DataDependencies.hs
@@ -121,20 +121,21 @@ envLfVersion :: Env -> LF.Version
 envLfVersion = worldLfVersion . envWorld
 
 -- | Type classes coming from dependencies. This maps a (module, synonym)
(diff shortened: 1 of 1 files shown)


sofiafaro-da (approved):
Looks good, thank you!

I was very confused by this PR. I didn't realize we were doing this sort of comparison for both synonyms and instances. I thought it was only for synonyms. I see why we're doing it in both case, though I'm not sure this is the correct approach for instances. But this PR isn't changing the approach, just making it faster, so it's definitely worth approving.

cocreature:
> I was very confused by this PR. I didn't realize we were doing this sort of comparison for both synonyms and instances. I thought it was only for synonyms. I see why we're doing it in both case, though I'm not sure this is the correct approach for instances. But this PR isn't changing the approach, just making it faster, so it's definitely worth approving.

yeah I agree it’s a bit iffy. For now, I’m only trying to make things faster but I’m very open to change the approach as well.

Issue #11608, opened 2021-11-09
Fixed choices are not generated properly in the typescript codegen


cocreature:
In https://github.com/digital-asset/daml/blob/f722cf13cda26fe1bb573004416ac944cf334789/language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts#L351 `buildAndLint.Main.Asset.Transfer` is `undefined`. It looks like somewhere we’re just not generating the code for the choice at all so the selector returns `undefined`

robin-da:
I added support for fixed choices in the TS codegen. However, the tests are still failing because the JSON API does not like interface template IDs.

cocreature:
Closing in favor of https://github.com/digital-asset/daml/issues/12689 so we have more descriptive issues.

PR #11630, opened 2021-11-10
iface: support for fixed choices in TS codegen

robin-da:
This adds support for fixed choices to the TS codegen. However,
currently this fails because the JSON API does not accept interface
template ID's when exercising a choice.

Part of #11608.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
+++ b/language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
@@ -337,12 +337,12 @@ test('create + fetch & exercise', async () => {
 
 });
 
+
 // TODO https://github.com/digital-asset/daml/issues/10810
-// Reenable test
-/*
+// Reenable full test when JSON API can handle interface contract IDs.
 test("interfaces", async () => {
   const aliceLedger = new Ledger({token: ALICE_TOKEN, httpBaseUrl: httpBaseUrl()});
-  const bobLedger = new Ledger({token: BOB_TOKEN, httpBaseUrl: httpBaseUrl()});
+  // const bobLedger = new Ledger({token: BOB_TOKEN, httpBaseUrl: httpBaseUrl()});
 
   const assetPayload = {
     issuer: ALICE_PARTY,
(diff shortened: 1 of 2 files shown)


stefanobaghino-da:
> However, currently this fails because the JSON API does not accept interface template ID's when exercising a choice.

Should this PR be a draft then? Or do changes need to land on the JSON API as a prerequisite to this? If yes, who is in charge of making such changes?

cocreature:
> Should this PR be a draft then? Or do changes need to land on the JSON API as a prerequisite to this? If yes, who is in charge of making such changes?

I think it’s fine to merge this first and then fix the JSON API later.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
```diff
@@ -348,15 +348,15 @@ test("interfaces", async () => {
   }
   const ifaceContract = await aliceLedger.create(buildAndLint.Main.Asset, assetPayload);
   expect(ifaceContract.payload).toEqual(assetPayload);
-  const [ifaceContract1, events1] = await aliceLedger.exercise(buildAndLint.Main.Asset.Transfer, ifaceContract.contractId, {newOwner: BOB_PARTY});
+  const [ifaceContract1, events1] = await aliceLedger.exercise(buildAndLint.Main.Token.Transfer, ifaceContract.contractId, {newOwner: BOB_PARTY});
```
cocreature, line 351: can we change the test to send the request and check that we get a failure to make sure the client side can send a request now and so that it blows up once the error from the JSON API changes?

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
```diff
@@ -348,15 +348,15 @@ test("interfaces", async () => {
...
   expect(events1).toMatchObject(
```
cocreature, line 352: we should also extend this test to exercise the choice both via the interface and via the template. Not quite sure if that passes the JSON api. probably not so assert on the failure.

cocreature (approved):
Nice, thanks! I’d suggest trying to enable the tests and assert on the JSON API failures in this PR.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
robin-da: One could try to cast the contract id to template/interface contract id. but the fixed choice is now defined in the interface and not part of the template anymore. We could try to add it back to the template, but that means we would need to traverse all implementations of a template and then generate those choices as part of the template. I'd prefer the choice only defined as part of the interface.

cocreature: Going back to the “archive as an interface” example, that behavior seems somewhat confusing to me: I wouldn’t expect to have to cast to an interface to call that and we deliberately decided that the ledger API should not require this. So having the typescript bindings support this seems like the only consistent option, otherwise we should revisit the choice to allow it on the ledger API.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
robin-da: done.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
robin-da: I managed to find a way so that both, the interface object, as well as the template object, have the interface fixed choices available. So this now works both ways.

PR #11633, opened 2021-11-10
Republish error codes generator lib

tudor-da:
Republish error codes generator libraries
* Additionally, add kvutils to classpath for generating KVErrors docs

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


cocreature (approved):
Thanks!

release/artifacts.yaml
```diff
@@ -237,3 +237,5 @@
   type: jar-scala
 - target: //ledger/error:error
   type: jar-scala
+- target: //ledger/error/generator:lib
```
cocreature, line 240: the artifact here seems to be off, CI gets unhappy about it

tudor-da: Hmm, that's because we can't generate 2.12 with it. Fabio specialized it to 2.13 only. Can we specify that we only publish for 2.13?

cocreature: not atm, it’s not super hard to fix though.

oliverse-da:
Thank you, @tudor-da !

PR #11634, opened 2021-11-10
kvutils - Backwards looking command deduplication [KVL-1174]

nicu-da:
Points to watch:
- pre-execution uses previous command max record time and current command to calculate the period between commands
- normal execution uses previous command record time and current command record time to calculate the period between commands
- the duplicate rejection is produced as the 'successfulWriteSet' in the committer, and is no longer part of the 'outoftimeboundsWriteSet'

This points are open for discussion if anyone has different thoughts about any of them

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -492,6 +492,7 @@ object CommandDeduplicationBase {
       */
     protected def delayBy(duration: Duration): Future[Unit]
   }
+
   class TimeDelayMechanism(val deduplicationDuration: Duration, val extraWait: Duration)(implicit
       ec: ExecutionContext
   ) extends DelayMechanism {
@@ -508,9 +509,9 @@ object CommandDeduplicationBase {
     override protected def delayBy(duration: Duration): Future[Unit] =
       ledger
         .time()
-        .flatMap(currentTime => {
+        .flatMap { currentTime =>
           ledger.setTime(currentTime, currentTime.plusMillis(duration.toMillis))
-        })
+        }
   }
 
   /** @param participantDeduplication If participant deduplication is enabled then we will receive synchronous rejections
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -64,6 +64,8 @@ private[testtool] abstract class CommandDeduplicationBase(
             DeduplicationPeriod.DeduplicationTime(deduplicationDuration.asProtobuf)
         )
       runWithDeduplicationDelay(configuredParticipants) { delay =>
+        val deltaForTransactionLatency =
+          300.millis // account for 3 requests, first accepted and the following 2 rejected
         for {
           // Submit command (first deduplication window)
           // Note: the second submit() in this block is deduplicated and thus rejected by the ledger API server,
@@ -71,10 +73,10 @@ private[testtool] abstract class CommandDeduplicationBase(
           completion1 <- submitRequestAndAssertCompletionAccepted(ledger)(request, party)
           _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
           // Delay close to the end of the minimum guaranteed deduplication
-          _ <- delay.delayToEndOfMinimumDeduplicationPeriod(100.millis)
+          _ <- delay.delayToEndOfMinimumDeduplicationPeriod(deltaForTransactionLatency)
           _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
           // Wait until the end of first deduplication window
-          _ <- delay.delayFromEndOfMinimumDeduplicationPeriod(100.millis)
+          _ <- delay.delayFromEndOfMinimumDeduplicationPeriod(deltaForTransactionLatency)
 
           // Submit command (second deduplication window)
           // Note: the deduplication window is guaranteed to have passed on both
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -473,23 +473,20 @@ object CommandDeduplicationBase {
 
     /** Delay by the guaranteed full deduplication period. After calling this method any duplicate calls should succeed
       */
-    def delayForEntireDeduplicationPeriod(): Future[Unit] = delayBy(
-      deduplicationDuration + extraWait
-    )
+    def delayForEntireDeduplicationPeriod(): Future[Unit] =
+      delayBy(deduplicationDuration + extraWait)
 
     /** Delay by [[deduplicationDuration]] - [[delta]]. Delta should be a value big enough to account for the latency required to process a transaction.
       * By using this delay we can validate that the minimum deduplication period [[deduplicationDuration]] is respected
       */
-    def delayToEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] = delayBy(
-      deduplicationDuration - delta
-    )
+    def delayToEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] =
+      delayBy(deduplicationDuration - delta)
 
     /** Delay by [[delta]] + [[extraWait]]. By using [[delayToEndOfMinimumDeduplicationPeriod]] and [[delayFromEndOfMinimumDeduplicationPeriod]]
       * with the same [[delta]] value, it mimics the behavior of [[delayForEntireDeduplicationPeriod]]
       */
-    def delayFromEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] = delayBy(
-      delta + extraWait
-    )
+    def delayFromEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] =
+      delayBy(delta + extraWait)
 
     /** Delay with [[duration]]
       */
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -64,19 +64,14 @@ private[testtool] abstract class CommandDeduplicationBase(
             DeduplicationPeriod.DeduplicationTime(deduplicationDuration.asProtobuf)
         )
       runWithDeduplicationDelay(configuredParticipants) { delay =>
-        val deltaForTransactionLatency =
-          300.millis // account for 3 requests, first accepted and the following 2 rejected
         for {
           // Submit command (first deduplication window)
           // Note: the second submit() in this block is deduplicated and thus rejected by the ledger API server,
           // only one submission is therefore sent to the ledger.
           completion1 <- submitRequestAndAssertCompletionAccepted(ledger)(request, party)
           _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
-          // Delay close to the end of the minimum guaranteed deduplication
-          _ <- delay.delayToEndOfMinimumDeduplicationPeriod(deltaForTransactionLatency)
-          _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
           // Wait until the end of first deduplication window
-          _ <- delay.delayFromEndOfMinimumDeduplicationPeriod(deltaForTransactionLatency)
+          _ <- delay.delayForEntireDeduplicationPeriod()
 
           // Submit command (second deduplication window)
           // Note: the deduplication window is guaranteed to have passed on both
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -476,18 +476,6 @@ object CommandDeduplicationBase {
     def delayForEntireDeduplicationPeriod(): Future[Unit] =
       delayBy(deduplicationDuration + extraWait)
 
-    /** Delay by [[deduplicationDuration]] - [[delta]]. Delta should be a value big enough to account for the latency required to process a transaction.
-      * By using this delay we can validate that the minimum deduplication period [[deduplicationDuration]] is respected
-      */
-    def delayToEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] =
-      delayBy(deduplicationDuration - delta)
-
-    /** Delay by [[delta]] + [[extraWait]]. By using [[delayToEndOfMinimumDeduplicationPeriod]] and [[delayFromEndOfMinimumDeduplicationPeriod]]
-      * with the same [[delta]] value, it mimics the behavior of [[delayForEntireDeduplicationPeriod]]
-      */
-    def delayFromEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] =
-      delayBy(delta + extraWait)
-
     /** Delay with [[duration]]
       */
     protected def delayBy(duration: Duration): Future[Unit]
--- a/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
+++ b/ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
@@ -5,8 +5,8 @@ package com.daml.ledger.api.testtool.infrastructure.deduplication
 
 import com.daml.error.ErrorCode
 import com.daml.error.definitions.LedgerApiErrors
-
 import java.util.UUID
+
 import com.daml.ledger.api.testtool.infrastructure.Allocation._
 import com.daml.ledger.api.testtool.infrastructure.Assertions._
 import com.daml.ledger.api.testtool.infrastructure.LedgerTestSuite
@@ -21,6 +21,7 @@ import com.daml.ledger.api.v1.ledger_offset.LedgerOffset
 import com.daml.ledger.client.binding.Primitive.Party
 import com.daml.ledger.test.model.DA.Types.Tuple2
 import com.daml.ledger.test.model.Test.{Dummy, DummyWithAnnotation, TextKey, TextKeyOperations}
+import com.daml.timer.Delayed
 import io.grpc.Status.Code
 
 import scala.annotation.nowarn
@@ -69,8 +70,11 @@ private[testtool] abstract class CommandDeduplicationBase(
           // only one submission is therefore sent to the ledger.
           completion1 <- submitRequestAndAssertCompletionAccepted(ledger)(request, party)
           _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
+          // Delay close to the end of the minimum guaranteed deduplication
+          _ <- delay.delayToEndOfMinimumDeduplicationPeriod(100.millis)
+          _ <- submitRequestAndAssertDeduplication(ledger)(request, party)
           // Wait until the end of first deduplication window
-          _ <- delay()
+          _ <- delay.delayFromEndOfMinimumDeduplicationPeriod(100.millis)
 
           // Submit command (second deduplication window)
           // Note: the deduplication window is guaranteed to have passed on both
@@ -184,7 +188,7 @@ private[testtool] abstract class CommandDeduplicationBase(
           _ <- submitAndWaitRequestAndAssertDeduplication(ledger)(request)
 
           // Wait until the end of first deduplication window
-          _ <- delay()
+          _ <- delay.delayForEntireDeduplicationPeriod()
 
           // Submit command (second deduplication window)
           _ <- ledger.submitAndWait(request)
@@ -253,7 +257,7 @@ private[testtool] abstract class CommandDeduplicationBase(
                     _ <- submitAndAssertDeduplicated(secondCall)
 
                     // Wait until the end of first deduplication window
-                    _ <- delay()
+                    _ <- delay.delayForEntireDeduplicationPeriod()
 
                     // Submit command (second deduplication window)
                     _ <- submitAndAssertAccepted(thirdCall)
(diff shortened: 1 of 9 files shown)


ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
```diff
@@ -469,7 +470,52 @@ private[testtool] abstract class CommandDeduplicationBase(
...
   trait DelayMechanism {
-    def apply(): Future[Unit]
+    val deduplicationDuration: Duration
+    val extraWait: Duration
+
+    /** Delay by the guaranteed full deduplication period. After calling this method any duplicate calls should succeed
+      */
+    def delayForEntireDeduplicationPeriod(): Future[Unit] = delayBy(
+      deduplicationDuration + extraWait
+    )
```
SamirTalwar-DA, line 480: Tiny nitpick: I find one-liners with long declarations easier to read if you put the line break after the `=`.

This is unfortunately something scalafmt can't do, so I am not so fussed that I would block anything; just bringing it up.

```suggestion
    def delayForEntireDeduplicationPeriod(): Future[Unit] =
      delayBy(deduplicationDuration + extraWait)
```

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
```diff
@@ -469,7 +470,52 @@ private[testtool] abstract class CommandDeduplicationBase(
...
+    protected def delayBy(duration: Duration): Future[Unit]
+  }
+  class TimeDelayMechanism(val deduplicationDuration: Duration, val extraWait: Duration)(implicit
+      ec: ExecutionContext
+  ) extends DelayMechanism {
+
+    override protected def delayBy(duration: Duration): Future[Unit] = Delayed.by(duration)(())
+  }
+
+  class StaticTimeDelayMechanism(
```
SamirTalwar-DA, line 507: This is pretty smart.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,140 @@
...
+          case Some(recordTime) =>
+            commandDedupBuilder.setRecordTime(recordTime)
+          case None =>
+            commandDedupBuilder.setMaxRecordTime(
+              commitContext.maximumRecordTime
+                .map(Conversions.buildTimestamp)
+                .getOrElse(
+                  throw Err.InternalError("Maximum record time is not set for pre-execution")
+                )
+            )
```
SamirTalwar-DA, line 137: I'd find this easier to read if you extracted a variable.

```suggestion
            val maxRecordTime = commitContext.maximumRecordTime.getOrElse(
              throw Err.InternalError("Maximum record time is not set for pre-execution")
            )
            commandDedupBuilder.setMaxRecordTime(
              Conversions.buildTimestamp(maximumRecordTime)
            )
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,140 @@
...
+          case None =>
+            // We select minimum record time for pre-execution
+            // During pre-execution in the deduplication state value we stored the maximum record time
+            // To guarantee the deduplication duration, we basically compare the maximum record time of the previous transaction
+            // with the minimum record time of the current transaction. This gives us the smallest possible interval between two transactions.
+            commitContext.minimumRecordTime match {
+              case Some(minimumRecordTime) => minimumRecordTime.toInstant
+              case None =>
+                throw Err.InternalError("Minimum record time is not set for pre-execution")
+            }
```
SamirTalwar-DA, line 61:
```suggestion
            commitContext.minimumRecordTime
              .getOrElse(throw Err.InternalError("Minimum record time is not set for pre-execution"))
              .toInstant
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,140 @@
...
+                  parseTimestamp(commandDeduplication.getMaxRecordTime).toInstant
+                Some(
+                  maxRecordTime
+                    .plus(commandDeduplicationDuration)
+                )
+              case TimeCase.TIME_NOT_SET =>
+                None
+            }
+          )
+          .forall(deduplicatedUntil => recordTimeOrMinimumRecordTime.isAfter(deduplicatedUntil))
```
SamirTalwar-DA, line 89: Perhaps convert `.toInstant` here and keep things as `Timestamp` until then?

SamirTalwar-DA (approved):
Just some nitpicks; nothing that would stop me from approving. I do have some broader questions which I will ask directly on the PR conversation.

SamirTalwar-DA:
Can you explain where this happens? I can't find the change.

> the duplicate rejection is produced as the 'successfulWriteSet' in the committer, and is no longer part of the 'outoftimeboundsWriteSet'

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
nicu-da: Fully agree with you, reorganized the methods.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
```diff
@@ -469,7 +470,49 @@ private[testtool] abstract class CommandDeduplicationBase(
 
 object CommandDeduplicationBase {
   trait DelayMechanism {
-    def apply(): Future[Unit]
+    val deduplicationDuration: Duration
+    val extraWait: Duration
+
+    /** Delay by the guaranteed full deduplication period. After calling this method any duplicate calls should succeed
```
andreaslochbihler-da, line 476: This is not guaranteed. A ledger may extend the deduplication period arbitrarily, even beyond the max deduplication period in the time model (and IIUC this can actually happen with this change in kvutils).

nicu-da:
> Can you explain where this happens? I can't find the change.
> 
> > the duplicate rejection is produced as the 'successfulWriteSet' in the committer, and is no longer part of the 'outoftimeboundsWriteSet'

The producing of the rejection as part of the deduplication step happens here: https://github.com/digital-asset/daml/pull/11634/files#diff-4326b69214fbfea2cbcd6f8b70767426a5dd3946ec5b24718b8523086708d40aR29 - the change we did is running the deduplication logic for pre-execution as well, compared to the previous behavior which was skipping this step during pre-execution.

And the change for the decoupling from the outOfTimeBoundWriteSet happens implicitly here https://github.com/digital-asset/daml/pull/11634/files#diff-4326b69214fbfea2cbcd6f8b70767426a5dd3946ec5b24718b8523086708d40aR128 as we no longer populate the deduplicate until in the state value. This is going to be made more explicit in a follow up PR that deprecates the deduplication_until values and removes it from the CommitContext

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
nicu-da: That is correct. I Will remove this comment but still leave the assumption in the test that this delay brings us "past" the deduplication period. Once we get the current changes in, and we re-evaluate the testing for command deduplication, we should either add metadata which helps us understand when the deduplication period ends/ provide a way to indicate to the test tool how it should calculate when the deduplication period ends/make the assertion for end of deduplication period optional

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,141 @@
...
+            commitContext.minimumRecordTime
+              .getOrElse(
+                throw Err.InternalError("Minimum record time is not set for pre-execution")
+              )
+              .toInstant
+        }
+        val isDuplicate = maybeDedupValue
+          .flatMap(commandDeduplication =>
+            commandDeduplication.getTimeCase match {
+              // Backward-compatibility, will not  be set for new entries
```
andreaslochbihler-da, line 66: I'm surprised by this comment. IIRC `DEDUPLICATE_UNTIL` was used to determine when the corresponding key can be deleted by pruning. How do you determine prunability nowadays?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,141 @@
...
-        .getOrElse(StepContinue(transactionEntry))
+      }
+
+      private def duplicateRejection(
+          commitContext: CommitContext,
+          transactionEntry: DamlTransactionEntrySummary,
+      )(implicit loggingContext: LoggingContext) = {
+        rejections.reject(
+          DamlTransactionRejectionEntry.newBuilder
+            .setSubmitterInfo(transactionEntry.submitterInfo)
```
andreaslochbihler-da, line 109: I think you want to adjust the deduplication duration in the submitter info for the rejection for the skews. We don't know whether the earlier submission was actually committed at the maxRecordTime stored in command dedup entry; it could have been commited with any record time `t1` within the minSkew + maxSkew before the stored maxRecordTime. So if this rejection gets recorded at say `t2 = commitConntext.maxRecordTime`, we may have that `t2 - t1` > `deduplicationDuration` in `submitterInfo`. However, if you set the `deduplicationDuration` to `max(commitContext.maxRecordTime - maybeDedupValue.minRecordTime, submitterInfo.deduplicationDuration)`, then you're guaranteed to specify a duration during which a conflicting completion was committed. For that, you not only have to store the `max_record_time` in `DamlCommandDedupValue`, but also `min_record_time` because we don't know any more what the skew parameters were when the `DamlCommandDedupValue` was written.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
nicu-da: Yes, the plan for this is to use the max deduplication duration and the record time to calculate the key expiration.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,141 @@
...
+          )
+          .forall(deduplicatedUntil =>
+            recordTimeOrMinimumRecordTime.isAfter(deduplicatedUntil.toInstant)
+          )
+        if (isDuplicate) {
+          StepContinue(transactionEntry)
+        } else {
+          if (commitContext.preExecute) {
+            // The out of time bounds entry is required in the committer, so we set it to the default value as we stop the steps here with the duplicate rejection
+            commitContext.outOfTimeBoundsLogEntry = Some(DamlLogEntry.getDefaultInstance)
```
miklos-da, line 97: This is not going to get consumed on the read path because we won't be setting record time bounds in this case, right?

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
```diff
@@ -469,7 +470,49 @@ private[testtool] abstract class CommandDeduplicationBase(
...
+      */
+    def delayForEntireDeduplicationPeriod(): Future[Unit] =
+      delayBy(deduplicationDuration + extraWait)
+
+    /** Delay by [[deduplicationDuration]] - [[delta]]. Delta should be a value big enough to account for the latency required to process a transaction.
+      * By using this delay we can validate that the minimum deduplication period [[deduplicationDuration]] is respected
+      */
+    def delayToEndOfMinimumDeduplicationPeriod(delta: Duration): Future[Unit] =
```
miklos-da, line 484: Where do we need these `delayTo*` and `delayFrom*`  methods?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
nicu-da: Done as port of [`4dd6e78` (#11634)](https://github.com/digital-asset/daml/pull/11634/commits/4dd6e784464ea56b12b53004a276dca12a768e30)

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,141 @@
...
+            // During pre-execution in the deduplication state value we stored the maximum record time
+            // To guarantee the deduplication duration, we basically compare the maximum record time of the previous transaction
+            // with the minimum record time of the current transaction. This gives us the smallest possible interval between two transactions.
+            commitContext.minimumRecordTime
+              .getOrElse(
+                throw Err.InternalError("Minimum record time is not set for pre-execution")
+              )
+              .toInstant
+        }
+        val isDuplicate = maybeDedupValue
```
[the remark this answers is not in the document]
nicu-da: Broke this down into multiple methods

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,141 @@
...
+                    .add(commandDeduplicationDuration)
+                )
+              case TimeCase.TIME_NOT_SET =>
+                None
+            }
+          )
+          .forall(deduplicatedUntil =>
+            recordTimeOrMinimumRecordTime.isAfter(deduplicatedUntil.toInstant)
+          )
+        if (isDuplicate) {
```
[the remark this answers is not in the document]
nicu-da: Yes, the naming was wrong, I renamed it

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,205 @@
...
+    new Step {
+      def apply(
+          commitContext: CommitContext,
+          transactionEntry: DamlTransactionEntrySummary,
+      )(implicit loggingContext: LoggingContext): StepResult[DamlTransactionEntrySummary] = {
+        val commandDeduplicationDuration =
+          if (transactionEntry.submitterInfo.hasDeduplicationDuration)
+            parseDuration(transactionEntry.submitterInfo.getDeduplicationDuration)
+          else
+            throw Err.InternalError(
```
hubert-da, line 45: What if nothing is specified?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
```diff
@@ -3,90 +3,205 @@
...
+            .setSubmitterInfo(submitterInfo)
+            // No duplicate rejection is a definite answer as the deduplication entry will eventually expire.
+            .setDefiniteAnswer(false)
+            .setDuplicateCommand(Duplicate.newBuilder.setDetails("")),
+          "the command is a duplicate",
+          commitContext.recordTime,
+        )
+      }
+
+      private def isTheCommandNotADuplicate(
```
hubert-da, line 125: Nit: this private function is the first one being used, so I would move it a bit higher.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
```diff
@@ -44,165 +52,312 @@ class CommandDeduplicationSpec
   private val rejections = new Rejections(metrics)
   private val deduplicateCommandStep = CommandDeduplication.deduplicateCommandStep(rejections)
   private val setDeduplicationEntryStep =
-    CommandDeduplication.setDeduplicationEntryStep(theDefaultConfig)
+    CommandDeduplication.setDeduplicationEntryStep()
 
+  private val timestamp: Timestamp = Timestamp.now()
```
hubert-da, line 57: Nit:
```suggestion
  private val timestamp: Timestamp = Timestamp.now()
  
```

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
```diff
@@ -54,12 +54,14 @@ message DamlCommandDedupKey {
 
 message DamlCommandDedupValue {
   reserved 1; // was record_time
-  // The time until when future commands with the same
-  // deduplication key will be rejected due to a duplicate submission.
-  google.protobuf.Timestamp deduplicated_until = 2;
-  // The submission time of the initial command.
-  // We use submission time because record time is not available during pre-execution.
-  google.protobuf.Timestamp submission_time = 3;
```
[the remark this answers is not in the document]
nicu-da: Not it was only one of them. But it was never really used as it was introduced in a previous PR

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
nicu-da: That is right, as we write the successful write set in this case, therefore this is just for the purpose of passing committer validation.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
nicu-da: Not currently used, I had previously added a test case that was using them but decided to leave that test case for a separate different PR where we would use these methods as well.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
```diff
@@ -44,165 +52,312 @@ class CommandDeduplicationSpec
...
+                  builder.setRecordTime(buildTimestamp(timestamp))
+              ),
+              "max record time" -> ((timestamp: Timestamp) =>
+                (builder: DamlCommandDedupValue.Builder) =>
+                  builder.setRecordTimeBounds(
+                    buildPreExecutionDeduplicationBounds(timestamp, timestamp)
+                  )
+              ),
+            )
+          )(
```
hubert-da, line 137: There are a lot of parentheses/brackets here and below. I think you should be able to improve readability with `{` instead of `(` here and using `case` for example. And I don't think you need to repeat the types.

hubert-da (commented):
Mostly LGTM, will have to do another run

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
nicu-da: That's not supported as of yet. It must always be specified as a duration

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
nicu-da: Good catch. Moved it. 
Would be amazing to have this organized automatically

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
nicu-da: Did a try on this, let me know what you think. I do agree that it was a mess to read.

hubert-da: LGTM, thanks!

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
miklos-da: Thanks for the confirmation. It may be worth adding a note that we assume the read path is going to handle an empty `DamlLogEntry`.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
miklos-da: Can we introduce these methods in the PR which is going to use them then, please? This should help in reducing the scope of this PR as well.

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplication.scala
andreaslochbihler-da: I don't think we have plans for supporting anything else than a duration for now. If the ledger client specifies an offset, then it is up to the participant to convert it into a suitable duration.

ledger/ledger-api-test-tool/src/main/scala/com/daml/ledger/api/testtool/infrastructure/deduplication/CommandDeduplicationBase.scala
andreaslochbihler-da: Fine for now.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
```diff
@@ -44,165 +52,304 @@ class CommandDeduplicationSpec
...
-    }
 
-    "produce rejection log entry in case record time is on or before deduplication time" in {
-      for (
-        (recordTime, deduplicationTime) <- Iterable(
-          (aRecordTime, aRecordTime),
-          (aRecordTime, aRecordTime.addMicros(1)),
+      "the deduplication duration is the delta between records when this exceeds the command deduplication rejection" in {
+        val dedupValue = newDedupValue(
+          _.setRecordTimeBounds(buildPreExecutionDeduplicationBounds(timestamp, timestamp))
```
andreaslochbihler-da, line 218: When you use the same timestamp for min and max, you can't test in the happy case that the code doesn't inadvertently pick the min record time. How about using two different timestamps here?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
```diff
@@ -44,165 +52,304 @@ class CommandDeduplicationSpec
...
+        )
+        val commitContext = createCommitContext(None, Map(aDedupKey -> Some(dedupValue)))
+        val deltaBetweenRecords = deduplicationDuration.plusMillis(1)
+        commitContext.minimumRecordTime = Some(timestamp)
+        commitContext.maximumRecordTime = Some(timestamp.add(deltaBetweenRecords))
```
andreaslochbihler-da, line 223: Here, too, I'd recommend to use a different timestamp than what we store in the commit context, so that we detect when the code mixes up the timestamps.

andreaslochbihler-da (approved):
The logic for computing the deduplication duration looks good. Thanks for working this out.

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/committer/transaction/CommandDeduplicationSpec.scala
nicu-da: Good point. I changed the test cases to use different values.

hubert-da (approved):
Good stuff!

PR #11638, opened 2021-11-10
String interning read integration [DPP-706]

nmarton-da:
* Wire to read pipelines (initialization and update)
* Patch StringInterning through the appropriate StorageBackend-s
* Creating a StringInterning for tests (which interns everything it sees)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
@@ -25,6 +25,7 @@ import com.daml.platform.store.backend.EventStorageBackend.{FilterParams, RangeP
 import com.daml.platform.store.backend.EventStorageBackend.RawTransactionEvent
 import com.daml.platform.store.backend.common.ComposableQuery.{CompositeSql, SqlStringInterpolation}
 import com.daml.platform.store.cache.LedgerEndCache
+import com.daml.platform.store.interning.StringInterning
 
 import scala.collection.compat.immutable.ArraySeq
 
@@ -32,12 +33,13 @@ abstract class EventStorageBackendTemplate(
     eventStrategy: EventStrategy,
     queryStrategy: QueryStrategy,
     ledgerEndCache: LedgerEndCache,
+    stringInterning: StringInterning,
     // TODO Refactoring: This method is needed in pruneEvents, but belongs to [[ParameterStorageBackend]].
     //                   Remove with the break-out of pruneEvents.
     participantAllDivulgedContractsPrunedUpToInclusive: Connection => Option[Offset],
 ) extends EventStorageBackend {
   import com.daml.platform.store.Conversions.ArrayColumnToStringArray.arrayColumnToStringArray
-
+  assert(stringInterning != null) // TODO remove
   private val logger: ContextualizedLogger = ContextualizedLogger.get(this.getClass)
 
   private val selectColumnsForFlatTransactions =
--- a/ledger/participant-integration-api/src/test/lib/scala/platform/store/interning/InterningStringInterning.scala
+++ b/ledger/participant-integration-api/src/test/lib/scala/platform/store/interning/InterningStringInterning.scala
@@ -1,0 +1,64 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.interning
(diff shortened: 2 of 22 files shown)


ledger/participant-integration-api/src/main/scala/platform/store/backend/common/EventStorageBackendTemplate.scala
```diff
@@ -25,19 +25,21 @@ import com.daml.platform.store.backend.EventStorageBackend.{FilterParams, RangeP
...
     queryStrategy: QueryStrategy,
     ledgerEndCache: LedgerEndCache,
+    stringInterning: StringInterning,
     // TODO Refactoring: This method is needed in pruneEvents, but belongs to [[ParameterStorageBackend]].
     //                   Remove with the break-out of pruneEvents.
     participantAllDivulgedContractsPrunedUpToInclusive: Connection => Option[Offset],
 ) extends EventStorageBackend {
   import com.daml.platform.store.Conversions.ArrayColumnToStringArray.arrayColumnToStringArray
-
+  assert(stringInterning != null) // TODO remove
```
nmarton-da, line 42: These will be removed in the next PRs to come

ledger/participant-integration-api/src/test/lib/scala/platform/store/interning/InterningStringInterning.scala
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.platform.store.interning
+
+import com.daml.lf.data.Ref
+import com.daml.lf.data.Ref.Party
+
+class InterningStringInterning extends StringInterning {
```
mziolekda, line 9: Can we call it `MockStringInterningView` or some such. It took me a while to understand where to place this class in the string interning space.

mziolekda (approved):
Nothing controversial here. Picked one naming nit.

ledger/participant-integration-api/src/test/lib/scala/platform/store/interning/InterningStringInterning.scala
rautenrieth-da: Why do we need this class? It looks like a simple implementation of the `StringInterning` trait. Can't we use `StringInterningView` instead? That one also implements `StringInterning` and lives in the same package.

RecoveringIndexerIntegrationSpec:
```
val stringInterning = new StringInterningView((_, _) => _ => Future.successful(Nil)) // not used
```

StorageBackendProvider:
```
val stringInterning = new InterningStringInterning
```

rautenrieth-da (approved):

ledger/participant-integration-api/src/test/lib/scala/platform/store/interning/InterningStringInterning.scala
nmarton-da: I was going with MockStringInterning :ok_hand:

nmarton-da: @rautenrieth-da this is a step forward: this guy will be the workhorse for StorageBackend tests not caring about interning

rautenrieth-da: Why couldn't you use `StringInterningView` for those tests? It doesn't perform any IO, and is implemented in the same package.

rautenrieth-da: Unless I misunderstand the code, this is also not a mock - it's a full implementation of the `StringInterning` interface.

nmarton-da: > Why couldn't you use StringInterningView for those tests? It doesn't perform any IO, and is implemented in the same package.

We could, but I think we are better off if we don't because if we use the real thing, then in those tests we also will test the StringInterningView implementation, and also we have to wire to the test runner and to the tests that we populate correct entries there. (please note this MockStringInterning does not need maintenance: it is transparent from the test point of view)

> Unless I misunderstand the code, this is also not a mock - it's a full implementation of the StringInterning interface.

Correct it is not a mock in a strict sense, but I am ok with the naming because it implies test usage. It is a trivial, crude lightweight implementation of the real thing.
Have no strong opinions on the name, if you would have a better one, happy to take it :)

PR #11639, opened 2021-11-10
interfaces: Preserve/provide by_interface data for create actions.

sofiafaro-da:
Part of #10915

This was a lot more involved than fetch or exercise. The first issue is
that we need to preserve the interface id into speedy, so it needs a
separate primitive ("experimental" won't cut it). Second, because
speedy's create requires the template definition, and now the interface
id as well, we basically need to compile a separate version of "create"
for each interface that a template implements, hence the separate
`CreateByInterfaceDefRef(templateId, ifaceId)`.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/compiler/damlc/daml-visual/src/DA/Daml/Visual.hs
+++ b/compiler/damlc/daml-visual/src/DA/Daml/Visual.hs
@@ -136,6 +136,8 @@ startFromUpdate seen world update = case update of
     LF.UGetTime -> Set.empty
     LF.UEmbedExpr _ upEx -> startFromExpr seen world upEx
     LF.UCreate tpl _ -> Set.singleton (ACreate tpl)
+    LF.UCreateInterface{} ->
+      error "Interfaces are not supported"
     LF.UExercise tpl choice _ _ -> Set.singleton (AExercise tpl choice)
     LF.UExerciseInterface{} ->
       -- TODO https://github.com/digital-asset/daml/issues/10810
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
@@ -1553,15 +1553,12 @@ private[lf] final class Compiler(
     ref -> SDefinition(withLabelT(ref, unsafeCompile(method.value)))
   }
 
-  @nowarn("msg=parameter value tokenPos in method compileCreateBody is never used")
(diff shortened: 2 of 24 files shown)


compiler/damlc/daml-visual/src/DA/Daml/Visual.hs
```diff
@@ -136,6 +136,8 @@ startFromUpdate seen world update = case update of
     LF.UGetTime -> Set.empty
     LF.UEmbedExpr _ upEx -> startFromExpr seen world upEx
     LF.UCreate tpl _ -> Set.singleton (ACreate tpl)
+    LF.UCreateInterface{} ->
```
cocreature, line 139: kind of tempted to try killing the visualizer in 2.0 but there is no chance I’ll convince people of that and as long as we don’t have to support new features it also does not get in the way all that much :shrug:

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
```diff
@@ -1534,33 +1538,61 @@ private[lf] final class Compiler(
     ref -> SDefinition(withLabel(ref, unsafeCompile(method.value)))
   }
 
-  private[this] def compileCreate(
+  @nowarn("msg=parameter value tokenPos in method compileCreateBody is never used")
```
cocreature, line 1541: probably a silly question but why do we pass it in if we don’t need it?

cocreature: I guess so that `topLevelFunction2` works without having to wrap it? I think I’d take the lambda wrapper that drops the argument but happy to leave it up to you.

cocreature (approved):
Nice, thank you!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Compiler.scala
sofiafaro-da: That and tradition mostly. There's a `compileFetchBody` that is very similar. But I'll try to remove the tokenPos entirely.

cocreature: if `compileFetchBody` is similar then it’s perfectly fine to keep it like this in this PR.

remyhaemmerle-da (approved):

PR #11640, opened 2021-11-10
Draft: scripts to extract size limits from existing IndexDBs

meiersi-da:
Not intended for merging right now, as we don't have a good place to put such one off scripts.

Using this PR thought to sync with @davidrichards-da who's asked me to build these scripts.

### Pull Request Checklist

trigger the build.


meiersi-da:
TODO as a next round of improvements, if required: apply this feedback from @lucb-da 

* --csv is not available on my psql, version 11.5. Using -A -F ',' seems to generate csv files.
* DBNANE and PSQL should be modifiable without having to edit the script. I ran the script from the db container, and it has limited tooling. I had to install an editor to be able to run the code. This can be done using bash default value for env variable, it will allow user to define the values before running the script.
* you should change the first line to `#!/usr/bin/env bash`. It is more portable.
* I didn't test it, but the script likely doesn't work if the user is not in the right folder.
* The error from the failing queries should not be visible. I guess the best would be to test if a query is likely to work first, and then run the query.
* support for a `-h` option, with usage would be great.
* you should run ShellCheck, it is a pretty good linter for bash. It will likely point to a few details which can be problems. (edited)

meiersi-da:
Closing, as no longer required.

meiersi-da:
Closing, as no longer required.

PR #11656, opened 2021-11-11
Move closureConvert and validate compiler phases into separate files.

nickchapman-da:
This PR just moves code from `Compiler.scala` into two new files.

The following step will be to introduce a new expression type `SExpr1` to distinguish expressions before and after closure conversion. And then we should be in a good place to rewrite the `closureConvert` phase so it is stack safe.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
@@ -1,0 +1,194 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package com.daml.lf.speedy
+
+/**  Closure Conversion (Phase of the speedy compiler pipeline)
+  *
+  *  This compilation phase transforms from SExpr0 to SExpr0.
+  *    SExpr0 contains expression forms which exist during the speedy compilation pipeline.
+  *
+  * TODO: introduces new expression type (SExpr1) for the result of this phase, and input to the
+  * following ANF transformation phase.
+  */
+
+import com.daml.lf.speedy.{SExpr0 => s}
+
+private[speedy] object ClosureConversion {
+
+  case class CompilationError(error: String) extends RuntimeException(error, null, true, false)
+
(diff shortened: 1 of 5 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ValidateCompilation.scala
```diff
@@ -0,0 +1,119 @@
...
+
+package com.daml.lf.speedy
+
+/**  ValidateCompilation (Phase of the speedy compiler pipeline)
+  *
+  *  This phases validates the final compilation result -- SExpr
+  */
+
+import com.daml.lf.speedy.SValue._
+import com.daml.lf.speedy.{SExpr => t}
```
remyhaemmerle-da, line 12: same remark here.

remyhaemmerle-da (approved):
Looks good to me.

cocreature (approved):
THanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
```diff
@@ -0,0 +1,194 @@
...
+/**  Closure Conversion (Phase of the speedy compiler pipeline)
+  *
+  *  This compilation phase transforms from SExpr0 to SExpr0.
+  *    SExpr0 contains expression forms which exist during the speedy compilation pipeline.
+  *
+  * TODO: introduces new expression type (SExpr1) for the result of this phase, and input to the
+  * following ANF transformation phase.
+  */
+
+import com.daml.lf.speedy.{SExpr0 => s}
```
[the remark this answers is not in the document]
nickchapman-da: AH yes, sorry.
I started this PR before the comments on the last.

Actually in the next PR, closure conversion will move between two different expression types
so I will probably use the `source.` / `target.` convention like  `Anf.scala`

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ValidateCompilation.scala
nickchapman-da: Here I will avoid the prefix entirely, since validate will only operate on a single kind of expression: `SExpr`.

PR #11662, opened 2021-11-11
Use by_interface field when fetching transaction dependencies.

### LINKED ISSUE #10915: Consider preserving "fetched by interface" info in Fetch and Exercise nodes

sofiafaro-da:
Fixes #10915

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
@@ -184,7 +184,8 @@ private[engine] final class Preprocessor(
       tx: SubmittedTransaction
   ): Result[ImmArray[speedy.Command]] =
     safelyRun(
-      getDependencies(List.empty, tx.rootNodes.toList.map(_.templateId))
+      getDependencies(List.empty, tx.rootNodes.toList.map(_.templateId)
+        ++ tx.byInterfaceNodes.toList.map(_.templateId))
     ) {
       transactionPreprocessor.unsafeTranslateTransactionRoots(tx)
     }
--- a/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
+++ b/daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
@@ -184,8 +184,11 @@ private[engine] final class Preprocessor(
       tx: SubmittedTransaction
   ): Result[ImmArray[speedy.Command]] =
     safelyRun(
-      getDependencies(List.empty, tx.rootNodes.toList.map(_.templateId)
-        ++ tx.byInterfaceNodes.toList.map(_.templateId))
+      getDependencies(
+        List.empty,
+        tx.rootNodes.toList.map(_.templateId)
+          ++ tx.byInterfaceNodes.toList.map(_.templateId),
+      )
     ) {
       transactionPreprocessor.unsafeTranslateTransactionRoots(tx)
     }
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Looks sensible thanks! We should add it to the testing issue.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -285,6 +285,19 @@ sealed abstract class HasTxNodes {
...
+        case action: Node.Action =>
+          if (action.byInterface.isDefined)
+            Iterator(action)
+          else
+            Iterator()
+        case _: Node.Rollback =>
+          Iterator()
+      }
+    )
+
```
remyhaemmerle-da, line 300: Not a very big deal, by I would prefer the node order to be preserved.
```suggestion
 def byInterfaceNodes: Iterable[Node.Action] = {
    fold(List.newBuilder[Node.Action]){
      case (acc, (_, node: Node.Action)) if node.byInterface.isDefined =>
        acc += node
      case (acc, _) => acc
    }
```

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
```diff
@@ -184,7 +184,11 @@ private[engine] final class Preprocessor(
       tx: SubmittedTransaction
   ): Result[ImmArray[speedy.Command]] =
     safelyRun(
-      getDependencies(List.empty, tx.rootNodes.toList.map(_.templateId))
+      getDependencies(
+        List.empty,
+        tx.rootNodes.toList.map(_.templateId)
+          ++ tx.byInterfaceNodes.toList.map(_.templateId),
+      )
```
remyhaemmerle-da, line 191: Since `tx.byInterfaceNodes` is traversing all the nodes, it is problably simpler to just get all the template Id of all the nodes.
```suggestion
  getDependencies(
        List.empty,
        tx.nodes.values.collect { case node: Node.Action => node.templateId }.toList)
  )
```

remyhaemmerle-da (commented):
Thanks.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: In fact If you make this public, it is probably a good idea to keep the `NodeId`.
Maybe simpler to make it `lf` private.

daml-lf/engine/src/main/scala/com/digitalasset/daml/lf/engine/preprocessing/Preprocessor.scala
sofiafaro-da: Won't this slow things down for transactions that don't use interfaces?

sofiafaro-da: I do like the simplicity of just fetching all templateIds, though.

remyhaemmerle-da: > Won't this slow things down for transactions that don't use interfaces?

After looking at `getDependencies`, you may be right. 
However I do not understand why I made it so complicate.

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
sofiafaro-da: I'll make `byInterfaceNodes` private and clean it up, based on your suggestion.

sofiafaro-da: Here: #11672

PR #11663, opened 2021-11-11
Activate party interning [DPP-712]

nmarton-da:
* Add DB schema changes
* Data migration scripts (PostgreSQL only)
* Add StringInterning to Query strategies
* Adapt queries, result-parsers
* Adapt ingestion
* Fix JdbcLedgerDaoBackend StringInterning handling

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V114__activate_string_interning_for_parties.sql
```diff
@@ -0,0 +1,317 @@
...
+   result INTEGER[];
+BEGIN
+   EXECUTE
+   'SELECT CASE' ||
+   '  WHEN $2 IS NULL THEN NULL' ||
+   '  WHEN cardinality($2) = 0 THEN ARRAY[]::INTEGER[]' ||
+   '  ELSE (' ||
+   '    SELECT array_agg(string_interning.internal_id)' ||
+   '      FROM' ||
+   '        unnest($2) AS original(string_elem),' ||
```
mziolekda, line 28: comma is equivalent to a CROSS JOIN that is only later whittled down with a WHERE statement on the result table. I would suggest replacing with a INNER JOIN.
I don't now to what extent Postgres is smart enough to optimize in this manner behind the scenes anyway

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
```diff
@@ -82,6 +82,7 @@ CREATE TABLE party_entries (
     typ VARCHAR NOT NULL,
     rejection_reason VARCHAR,
     is_local BOOLEAN,
+    party_id INTEGER,
```
tudor-da, line 85: is it on purpose to keep the existing `party` column?

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V114__activate_string_interning_for_parties.sql
nmarton-da: We use cross joins everywhere, and in all cases it could be transformed to an inner join (so we never need a true cross join).
My working assumption is that it does not really matter which you choose from planner point of view (at least for PostgreSQL)
I checked the plans for the cross vs inner join in that particular case, and it is identical:
```
       explain SELECT array_agg(string_interning.internal_id)
         FROM
           unnest(array['a', 'b', 'c', 'ACSinvalidLedgerId-alpha-b894853580a6-party-0', 'ACSemptyResponse-alpha-b894853580a6-party-0']) AS original(string_elem)
           inner join string_interning on ('p|' || original.string_elem = string_interning.external_string);
```
```
                                          QUERY PLAN                                          
----------------------------------------------------------------------------------------------
 Aggregate  (cost=23.03..23.04 rows=1 width=32)
   ->  Hash Join  (cost=0.12..23.02 rows=5 width=4)
         Hash Cond: (string_interning.external_string = ('p|'::text || original.string_elem))
         ->  Seq Scan on string_interning  (cost=0.00..18.90 rows=790 width=68)
         ->  Hash  (cost=0.05..0.05 rows=5 width=32)
               ->  Function Scan on unnest original  (cost=0.00..0.05 rows=5 width=32)
```
I think we can leave it as it is.

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
nmarton-da: yes, although I was also a bit uncertain. My arguments where:
* It does not matter much (relatively small table)
* Better for manual DB checking
* In this case it is actually the source of truth which we would intern, and in other cases it is usage
* This we only need for pruning, so if we would not need this for pruning for some reason, it would be easier to track back

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V114__activate_string_interning_for_parties.sql
mziolekda: :+1: as comma join also happens to be more redable

ledger/participant-integration-api/src/main/resources/db/migration/h2database-appendonly/V1__Append_only_schema.sql
mziolekda: I second the decision to keep the source of truth table complete.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V114__activate_string_interning_for_parties.sql
```diff
@@ -0,0 +1,317 @@
...
+
+DROP INDEX string_interning_external_string_temp_idx;
+DROP FUNCTION internalize_string_array(TEXT, TEXT[]);
+
+CREATE VIEW participant_events
+AS
+SELECT
+    0::smallint as event_kind,
+    event_sequential_id,
+    NULL::text as event_offset,
```
tudor-da, line 186: any reason to keep this one null?

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresEventStrategy.scala
```diff
@@ -7,45 +7,72 @@ import com.daml.platform.store.appendonlydao.events.Party
...
       case wildCardParties if wildCardParties.isEmpty =>
         Nil
 
       case wildCardParties =>
-        val partiesArray = wildCardParties.view.map(_.toString).toArray
-        cSQL"(#$witnessesColumnName::text[] && $partiesArray::text[])" :: Nil
+        val partiesArray: Array[java.lang.Integer] = wildCardParties.view
+          .flatMap(party => stringInterning.party.tryInternalize(party).map(Int.box).toList)
+          .toArray
+        if (partiesArray.isEmpty)
```
mziolekda, line 51: Why can `partiesArray` come out as `isEmpty` for `witnessesWhereClause` but not for `filteredEventWitnessesClause`

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V114__activate_string_interning_for_parties.sql
nmarton-da: I don't know, but I think that is not part of this PR (I would prefer not to do unrelated changes).
To remove the view entirely is on the horizon, if we get there, I guess that would resolve this question as well.

ledger/participant-integration-api/src/main/scala/platform/store/Conversions.scala
```diff
@@ -142,7 +144,56 @@ private[platform] object Conversions {
...
+        case Right(value) => Right(value)
+        case Left(_) =>
+          val MetaDataItem(qualified, _, _) = meta
+          value match {
+            case someArray: Array[_] =>
+              Try(
+                someArray.view.map {
+                  case i: Int => i
+                  case null =>
+                    throw new Exception(s"Cannot convert object array element null to Int")
```
tudor-da, line 163: Let's make sure they are handled properly with the new error codes.
```suggestion
                    throw new SQLNonTransientException(s"Cannot convert object array element null to Int")
```

ledger/participant-integration-api/src/main/scala/platform/store/Conversions.scala
```diff
@@ -142,7 +144,56 @@ private[platform] object Conversions {
...
+                  case invalid =>
+                    throw new Exception(
```
tudor-da, line 165:
```suggestion
                    throw new SQLNonTransientException(
```

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresEventStrategy.scala
nmarton-da: sharp eye!
It could based on the type.
That would boil down passing empty array down.
This can only happen, if we have never seen any of those parties which was the argument:
* This type of queries are quite futile because they will never get back any result: the visibility filter is controlled by the same set of parties (= if that is empty -> no result)
* But they still need to be valid SQLs
* Front-running next question from this: for the OracleEventStrategy there is a empty handling there because the parameter is not passed as an array but a anorm-multivalue IN clause, which if empty produces invalid SQL
* Front-running possible suggestion for improvement: it is possible in the aforementioned circumstance to not do the query at all! But it would require a slight re-wamping of the functions and interfaces here. Also the improvement is not obviously huge, clients probably not banging the LAPI service with futile queries, or not that much that it is worth optimizing for (because of that reason I was not even creating an improvement task for this neither)

tudor-da (approved):
LGTM

mziolekda (approved):

ledger/participant-integration-api/src/main/scala/platform/store/backend/postgresql/PostgresEventStrategy.scala
```diff
@@ -7,45 +7,72 @@ import com.daml.platform.store.appendonlydao.events.Party
...
-    (wildCardClause ::: partiesTemplatesClauses).mkComposite("(", " OR ", ")")
+      filterParams.partiesAndTemplates.iterator
+        .map { case (parties, templateIds) =>
+          (
+            parties.flatMap(s => stringInterning.party.tryInternalize(s).toList),
+            templateIds,
+          )
+        }
+        .filterNot(_._1.isEmpty)
+        .filterNot(_._2.isEmpty)
```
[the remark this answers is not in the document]
nmarton-da: Yes: everything is possible since on the API real strings (party, template id) arrive, and during interning those we might end up with empty lists...ahh I just realized you are referring to the template_ids which are one PR away from here: then this will make sense, right now not.

ledger/participant-integration-api/src/main/scala/platform/store/backend/oracle/OracleEventStrategy.scala
```diff
@@ -7,48 +7,79 @@ import com.daml.lf.data.Ref
...
+          (
+            parties.flatMap(s => stringInterning.party.tryInternalize(s).toList),
+            templateIds,
           )
-        cSQL"( ($clause) AND (template_id IN (${templateIds.map(_.toString)})) )"
-      }.toList
-    (wildCardClause ::: partiesTemplatesClauses).mkComposite("(", " OR ", ")")
+        }
+        .filterNot(_._1.isEmpty)
+        .filterNot(_._2.isEmpty)
```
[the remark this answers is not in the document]
nmarton-da: Similarly: no. But in the next PR it will.

PR #11666, opened 2021-11-12
Limit gRPC wire transport size legacy error codes size

tudor-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
@@ -509,7 +509,7 @@ class ErrorFactories private (errorCodesVersionSwitcher: ErrorCodesVersionSwitch
   }
 
   private def truncated(message: String): String = {
-    val maxMessageLength = 1536
+    val maxMessageLength = 1536 // An arbitrary limit that doesn't break netty serialization while being useful to human operator.
     if (message.length > maxMessageLength) message.take(maxMessageLength) + "..." else message
   }
 
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
@@ -509,7 +509,8 @@ class ErrorFactories private (errorCodesVersionSwitcher: ErrorCodesVersionSwitch
   }
 
   private def truncated(message: String): String = {
-    val maxMessageLength = 1536 // An arbitrary limit that doesn't break netty serialization while being useful to human operator.
+    val maxMessageLength =
+      1536 // An arbitrary limit that doesn't break netty serialization while being useful to human operator.
     if (message.length > maxMessageLength) message.take(maxMessageLength) + "..." else message
   }
 
--- a/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
+++ b/ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
@@ -3,6 +3,7 @@
 
 package com.daml.platform.server.api.validation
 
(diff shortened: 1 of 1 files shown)


ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -476,9 +486,15 @@ class ErrorFactories private (errorCodesVersionSwitcher: ErrorCodesVersionSwitch
     * @param status A Protobuf [[Status]] object.
     * @return An exception without a stack trace.
     */
-  def grpcError(status: Status): StatusRuntimeException = new NoStackTraceApiException(
-    StatusProto.toStatusRuntimeException(status)
-  )
+  def grpcError(status: Status): StatusRuntimeException = {
+    val maxMessageLength = 1536
```
mziolekda, line 490: This value owes an explanation in the comment.

nicu-da (approved):
LGTM!

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
tudor-da: Added a comment detailing why we need to truncate. Re the actual numerical limit justification, I would defer to @daravep (maybe you have an heuristic from Canton where this problem occurs more frequently). The upper limit of what we can use there is 8K (default Netty header size), but that can include other KV pairs.

daravep: I'd say the following: the message is human centric. Any message beyond 1.5k is not a message anymore, but an encyclopedia. Though the status runtime exception call this field really message. Hence, let's limit it.

I don't have a precise heuristic on how to compute the allowed size. I even think it is not possible, because we are not talking just about the 8k size of the exception. There is other stuff transported as part of the header which we can't influence.

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -491,6 +483,31 @@ class ErrorFactories private (errorCodesVersionSwitcher: ErrorCodesVersionSwitch
...
+      Status
+        .newBuilder(status)
+        .setMessage(truncated(status.getMessage))
+    new NoStackTraceApiException(
+      StatusProto.toStatusRuntimeException(newStatus.build)
+    )
+  }
+
+  private def truncated(message: String): String = {
+    val maxMessageLength = 1536
```
mziolekda, line 508:
```suggestion
    val maxMessageLength = 1536 // An arbitrary limit that doesn't break netty serialization while being useful to human operator.
```

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
pbatko-da: Is this length in bytes?

mziolekda (approved):

ledger/ledger-api-common/src/main/scala/com/digitalasset/platform/server/api/validation/ErrorFactories.scala
```diff
@@ -496,6 +488,32 @@ class ErrorFactories private (errorCodesVersionSwitcher: ErrorCodesVersionSwitch
...
+      Status
+        .newBuilder(status)
+        .setMessage(truncated(status.getMessage))
+    new NoStackTraceApiException(
+      StatusProto.toStatusRuntimeException(newStatus.build)
+    )
+  }
+
+  private def truncated(message: String): String = {
+    val maxMessageLength =
```
miklos-da, line 512: Can you make this arbitrary limit more explicit, please? E.g., by moving its definition to above `grpcError` and referencing it from the latter method's ScalaDoc. The same should be done for `internalAuthenticationError` to make this logic transparent.

PR #11668, opened 2021-11-12
Remove SEDamlException & SEImportValue from SExpr{0,1}

nickchapman-da:
Remove unnecessary constructors: `SEDamlException`, `SEImportValue`, from `SExpr{0,1}`

Actually they probably shouldn't even be a part of `SExpr` either. Instead we should implement the functionality using new continuation variants. But that's a refactoring step for another day.

Part of: #11669


remyhaemmerle-da (approved):
Thanks

PR #11672, opened 2021-11-12
Improve byInterfaceNodes based on suggestions

sofiafaro-da:
From the last PR #11662

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
+++ b/daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
@@ -285,18 +285,16 @@ sealed abstract class HasTxNodes {
       }
     )
 
-  def byInterfaceNodes: Iterable[Node.Action] =
-    nodes.values.flatMap(node =>
-      node match {
-        case action: Node.Action =>
-          if (action.byInterface.isDefined)
-            Iterator(action)
-          else
-            Iterator()
-        case _: Node.Rollback =>
-          Iterator()
-      }
-    )
+  private[lf] def byInterfaceNodes: List[Node.Action] = {
+    val builder = List.newBuilder[Node.Action]
+    nodes.values.foreach {
+      case action: Node.Action if action.byInterface.isDefined =>
+        (builder += action)
+      case _ =>
+        ()
+    }
+    builder.result()
+  }
 
   /** This function traverses the transaction tree in pre-order traversal (i.e. exercise node are traversed before their children).
     *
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Thanks!

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
```diff
@@ -285,18 +285,16 @@ sealed abstract class HasTxNodes {
...
-    )
+  private[lf] def byInterfaceNodes: List[Node.Action] = {
+    val builder = List.newBuilder[Node.Action]
+    nodes.values.foreach {
+      case action: Node.Action if action.byInterface.isDefined =>
+        (builder += action)
+      case _ =>
+        ()
+    }
+    builder.result()
```
remyhaemmerle-da, line 296: My point was to list the nodes by traversal order, for that you can use Transaction `fold` or `foreach` method.
However, if you make it private I am completely fine you keep the iterator version.

remyhaemmerle-da (approved):

daml-lf/transaction/src/main/scala/com/digitalasset/daml/lf/transaction/Transaction.scala
remyhaemmerle-da: Since nodes is an `HashMap`, the resulting `nodes.values` will be in an unspecified order (different at each jvm restart)

cocreature: oh sorry somehow I misread this. We should definitely use `fold` or `foreach` on the transaction.

remyhaemmerle-da: > However, if you make it private I am completely fine you keep the original iterator version.

In fact, we should get the node by traversal order, other error message can be non deterministic.

remyhaemmerle-da:
```suggestion
    foreach {
      case (_, action: Node.Action) if action.byInterface.isDefined =>
        (builder += action)
      case _ =>
        ()
    }
    builder.result()
```

remyhaemmerle-da: better to use a discard for `builder += action`

sofiafaro-da: Oh, I see! I definitely didn't understand the first time around.

Issue #11669, opened 2021-11-12
Refactoring Speedy


nickchapman-da:
This issue collects various _refactorings_ for speedy compiler and runtime engine.

- Some refactoring are purely to improve clarity.
- Some are intended as enablers for more substantive changes, such as the stack-safety work.
- Some refactoring are be steps towards improving efficiency.

### Planned:

- Distinguish expression types for different phases of the speedy compilation pipeline & expressions which run on the speedy machine. Remove those expression forms which are not required/allowed at each phase. Specifically we want:

    - `SExpr0` : as first converted from LF
    - `SExpr1` : after closure conversion
    - `SExpr`  : after ANF, and running on speedy machine

- Split compilation phases into separate files

- Move code which is logically part of the phase1 compilation (_LF_ `Expr` -> `SExpr0`) out of `SExpr` classes and into Compiler.scala

- Stop manufacturing expressions at runtime. Instead use new continuation forms (`Kont`) to implement the functionality.

- Avoid construction of `SEApp` / `SEAppGeneral` application nodes, to allow this expression form to be removed from `SExpr`. (runtime expressions).

- Other refactorings necessary to enable all speedy compiler phases to be recoded for stack-safety. #11561

- Simplify `SEBuiltinRecursiveDefinition` to its one case `EqualList`

### Streach:

- Fix/re-order speedy compiler phases. Specifically:

    - Avoid early conversion to de-Bruijn indexes (done in phase 1) which causes subsequent compiler phases to be way more complicated that otherwise.

    - Re-order compilation phases: _closure-conversion_ and _ANF-transformation_ so that _closure-conversion_ is done later. See notes below.

Closure-conversion should come later because it's essential purpose is to map from _variables_ to _locations_. i.e. (In what location will the value bound to each variable be found at runtime? The answer: on the _stack_, in the _args_ array, or in the surrounding closure's _free_ array). Before closure-conversion, the compiler phases can work with the much simpler concept of variables. Currently the ANF transformation is done before closure-conversion and consequently has to jump through complicated hoops to track a mapping between locations before/after the transformation step.

nickchapman-da:
29/3/22
Added tick boxes to the issue description to reflect status

PR #11676, opened 2021-11-12
Simplify handling for EqualList builtin.

nickchapman-da:
Simplify handling for `EqualList`
- Remove special handling via expression form `SEBuiltinRecursiveDefinition`
- Instead support simply as just another `SBuiltin` --  `SEBuiltinEqualList`
- Retain/relocate the implementation code which expands an application of the builtin  into its body on each iteration.

The payoff is we have one less expression form to worry about in the speedy compiler pipeline.
 
Part of: #11669


cocreature (approved):
Looks good, nice simplification! I’d like a review from @remyhaemmerle-da before merging as well.

remyhaemmerle-da (approved):
Nice.

PR #11680, opened 2021-11-12
Activate template_id interning [DPP-713]

nmarton-da:
* Add DB schema changes
* Data migration scripts (PostgreSQL only)
* Adapt queries, result-parsers
* Adapt ingestion

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V6__activate_template_id_interning.sql
```diff
@@ -0,0 +1,19 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+DROP INDEX participant_events_divulgence_template_id_idx;
+DROP INDEX participant_events_create_template_id_idx;
+DROP INDEX participant_events_consuming_exercise_template_id_idx;
+DROP INDEX participant_events_non_consuming_exercise_template_id_idx;
+
+ALTER TABLE participant_events_divulgence
+  MODIFY template_id NUMBER;
```
rautenrieth-da, line 10: This will likely fail if the table is not empty (conversion from string to integer). Which is fine because we don't support data continuity for Oracle yet.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V115__activate_string_interning_for_template_ids.sql
```diff
@@ -0,0 +1,221 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+SELECT 'Template Id interning data migration - Preparation...';
```
rautenrieth-da, line 4: Nice way to add debug output.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V115__activate_string_interning_for_template_ids.sql
```diff
@@ -0,0 +1,221 @@
...
+
+
+
+SELECT 'Template Id interning data migration - Migrating participant_events_create...';
+
+DROP INDEX participant_events_create_template_id_idx;
+
+ALTER TABLE participant_events_create
+  ALTER COLUMN template_id TYPE INTEGER
+  USING internalize_string('t|', template_id);
```
rautenrieth-da, line 56: Did you measure the performance of this?

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2EventStrategy.scala
```diff
@@ -45,15 +45,21 @@ object H2EventStrategy extends EventStrategy {
...
             stringInterning,
           )
-        val templateIdsArray = templateIds.view.map(_.toString).toArray
-        cSQL"( ($clause) AND (template_id = ANY($templateIdsArray)) )"
+        val templateIdsArray: Array[java.lang.Integer] =
+          templateIds.view
+            .map(stringInterning.templateId.tryInternalize)
+            .flatMap(_.toList)
+            .map(Int.box)
+            .toArray
```
rautenrieth-da, line 60:
```suggestion
            .toArray // anorm does not like primitive arrays, so we need to box it
```
The same comment was added elsewhere, will help us remember why we call `.map(Int.box)`

ledger/participant-integration-api/src/main/scala/platform/store/backend/h2/H2EventStrategy.scala
```diff
@@ -45,15 +45,21 @@ object H2EventStrategy extends EventStrategy {
...
+        if (templateIdsArray.isEmpty) Iterator.empty
```
rautenrieth-da, line 61:
```suggestion
        if (templateIdsArray.isEmpty) Iterator(cSQL"FALSE")
```
This would be an alternative (I think) to matching for an empty list below.

rautenrieth-da (commented):
The changes look good. 

We should add tests that try to filter for unknown templates, as the logic around that changed.

ledger/participant-integration-api/src/main/resources/db/migration/postgres-appendonly/V115__activate_string_interning_for_template_ids.sql
nmarton-da: Not yet, it is planned to measure this together with the party interning and the filter table conversion together

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/resources/db/migration/oracle-appendonly/V6__activate_template_id_interning.sql
```diff
@@ -0,0 +1,19 @@
+--  Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+--  SPDX-License-Identifier: Apache-2.0
+
+DROP INDEX participant_events_divulgence_template_id_idx;
```
nmarton-da, line 4: @meiersi-da next to dropping all the GIN indexes, we dropping the template_id single column indexes as well, since I could not find any query usage which would make use of it. What do you think?

meiersi-da: Yes this makes sense.

PR #11681, opened 2021-11-12
[TS-Bindings] Changes to re-enable ws multiplexing

akshayshirahatti-da:
There were 2 issues here 
 - one was that on reconnect we created a new websocket connection but the `onMessage` handler still referred to the old ws object. Now the `onMessage` handler takes a websocket as an input argument.
 - the second issue was that we didn't update `wsLiveSince` parameter on the first message with offset , which indicated that the stream is live. 

CHANGELOG_BEGIN
[TS-BINDINGS] Renable ws multiplexing for stream queries after resolving the reconnect connection close bug.
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
+++ b/language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
@@ -28,6 +28,8 @@ const ALICE_PARTY = 'Alice';
 const ALICE_TOKEN = computeToken(ALICE_PARTY);
 const BOB_PARTY = 'Bob';
 const BOB_TOKEN = computeToken(BOB_PARTY);
+const CHARLIE_PARTY = 'Charlie'
+const CHARLIE_TOKEN = computeToken(CHARLIE_PARTY)
 
 let sandboxPort: number | undefined = undefined;
 const SANDBOX_PORT_FILE = 'sandbox.port';
@@ -77,7 +79,7 @@ beforeAll(async () => {
     getEnv('JSON_API'),
     ['--ledger-host', 'localhost', '--ledger-port', `${sandboxPort}`,
      '--port-file', JSON_API_PORT_FILE, '--http-port', "0",
-     '--allow-insecure-tokens', '--websocket-config', '--heartBeatPer=1', '--log-level=INFO'],
+     '--allow-insecure-tokens', '--websocket-config=maxDuration=1,heartBeatPer=1', '--log-level=INFO'],
     ['-Dakka.http.server.request-timeout=60s'],
   )
   await waitOn({resources: [`file:${JSON_API_PORT_FILE}`]})
@@ -216,7 +218,7 @@ test('create + fetch & exercise', async () => {
   const personRawStream = aliceLedger.streamQuery(buildAndLint.Main.Person);
   const personStream = promisifyStream(personRawStream);
   const personStreamLive = pEvent(personRawStream, 'live');
-  expect(await personStream.next()).toEqual([[alice6Contract], [{created: alice6Contract, matchedQueries:[0]}]]);
+  expect(await personStream.next()).toEqual([[alice6Contract], [{created: alice6Contract, matchedQueries:[1]}]]);
 
   // end of non-live data, first offset
   expect(await personStreamLive).toEqual([alice6Contract]);
@@ -226,7 +228,7 @@ test('create + fetch & exercise', async () => {
   const bob4Contract = await bobLedger.create(buildAndLint.Main.Person, bob4);
   expect(bob4Contract.payload).toEqual(bob4);
   expect(bob4Contract.key).toEqual(bob4Key);
-  expect(await personStream.next()).toEqual([[alice6Contract, bob4Contract], [{created: bob4Contract, matchedQueries:[0]}]]);
+  expect(await personStream.next()).toEqual([[alice6Contract, bob4Contract], [{created: bob4Contract, matchedQueries:[1]}]]);
 
 
   // Alice changes her name.
@@ -243,7 +245,7 @@ test('create + fetch & exercise', async () => {
   expect(cooper6Contract.key).toEqual(alice6Key);
   expect(await aliceStream.next()).toEqual([[cooper6Contract], [{archived: alice6Archived}, {created: cooper6Contract, matchedQueries:[0]}]]);
   expect(await alice6KeyStream.next()).toEqual([cooper6Contract, [{archived: alice6Archived}, {created: cooper6Contract}]]);
-  expect(await personStream.next()).toEqual([[bob4Contract, cooper6Contract], [{archived: alice6Archived}, {created: cooper6Contract, matchedQueries:[0]}]]);
+  expect(await personStream.next()).toEqual([[bob4Contract, cooper6Contract], [{archived: alice6Archived}, {created: cooper6Contract, matchedQueries:[1]}]]);
 
   personContracts = await aliceLedger.query(buildAndLint.Main.Person);
   expect(personContracts).toEqual([bob4Contract, cooper6Contract]);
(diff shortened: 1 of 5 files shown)


language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
```diff
@@ -611,3 +613,38 @@ test('package API', async () => {
   const downSuc = await ledger.getPackage(buildAndLint.packageId);
   expect(downSuc.byteLength > 0).toBe(true);
 });
+
+test('reconnect on timeout, when multiplexing is enabled', async () => {
```
akshayshirahatti-da, line 617: not quite happy with the test as it does a wait to trigger a timeout and re-connect , can someone suggest a cleaner/better way to do this ? want to test reconnect logic in an integration test.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
```diff
@@ -611,3 +613,38 @@ test('package API', async () => {
...
+    party: CHARLIE_PARTY,
+    age: '10',
+    friends: [],
+  };
+
+  const charlieContract1 = await charlieLedger.create(buildAndLint.Main.Person, charlieRecord1);
+  expect(await charlieStream.next()).toEqual([[charlieContract1], [{created: charlieContract1, matchedQueries: [0]}]]);
+
+  // wait 70s to trigger a disconnect on json-api which is configured to close conn after 1 minute.
+  await new Promise(resolve => setTimeout(resolve, 70000));
```
garyverhaegen-da, line 635: Can we reduce that a bit? IIRC the timeout on WS connection is configurable, it would be great if this could be set to like 10s for these tests here.

akshayshirahatti-da: IIUC `maxDuration` is specified in minutes, so 1 minute is the min we can do without changing the cli opt definition(might have to deprecate `maxDuration` in favour of a new flag `maxDurationSeconds`) or adding a hacky backdoor maybe via env variables to override this.
```scala 
protected[this] override def create(implicit readCtx: DummyImplicit) =
    for {
      md <- optionalLongField("maxDuration")
      hbp <- optionalLongField("heartBeatPer")
    } yield Config.DefaultWsConfig
      .copy(
        maxDuration = md
          .map(t => FiniteDuration(t, TimeUnit.MINUTES))
          .getOrElse(Config.DefaultWsConfig.maxDuration),
        heartBeatPer = hbp
          .map(t => FiniteDuration(t, TimeUnit.SECONDS))
          .getOrElse(Config.DefaultWsConfig.heartBeatPer),
      )
```

language-support/ts/daml-ledger/index.ts
```diff
@@ -493,7 +493,7 @@ class QueryStreamsManager {
 
             const ws = new WebSocket(manager.url, manager.protocols);
```
garyverhaegen-da, line 494: If I'm counting that right, this block of code has 4 different variables called `ws` which need to be distinguished. I'm not surprised that resulted in a bug here.

There seems to be:
- Two `const ws` with nested scopes, one shadowing the other.
- One `this.ws` at the class level.
- One `manager.ws` which is "a stable alias for `this.ws` in callbacks" but is also sometimes used outside of callbacks. Though not always.

Since both `const ws` are constructed specifically in order to ultimately assign them to `manager.ws`, could we just always work with `manager.ws` directly? Could we also make sure we always refer to that property with `manager.ws`, rather than a mix of `this.ws` and `manager.ws`?

Since the `onWs*` functions are created (and presumably called?) as standalone functions, not as object properties, do we know that they end up being called in a context that rebinds `this`? If not, can we skip the `manager` alias and just use `this.ws` everywhere?

What do we gain by making QueryStreamManager a class? It seems we get some amount of confusion and complexity that obviously led to bugs, but I'm unclear on what we actually gain compared to having a `queryStreamManager` function that returns a simple JS object with the properties we need (and thus no `this` anywhere, aliased or not). That function could easily use a single `let ws`, like the preexisting `streamSubmit` did, avoiding opportunities for confusion.

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
garyverhaegen-da: Ugh. All right. 😿

language-support/ts/codegen/tests/ts/build-and-lint-test/src/__tests__/test.ts
garyverhaegen-da: I don't really have a better idea for an integration test, no. If you could mock the server, I suppose you could fake a timeout faster, but our current fake server (in react-ledger tests) is not exactly great for that IIRC.

language-support/ts/daml-ledger/index.ts
stefanobaghino-da: > Since the onWs* functions are created (and presumably called?) as standalone functions, not as object properties, do we know that they end up being called in a context that rebinds this? If not, can we skip the manager alias and just use this.ws everywhere?

It could be, I originally wrote the code with those being object properties, which is why I introduced the aliasing.

> Two const ws with nested scopes, one shadowing the other.

Avoiding the shadowing is probably good, thanks.

> One manager.ws which is "a stable alias for this.ws in callbacks" but is also sometimes used outside of callbacks. Though not always.

It should not be used outside of the callbacks.

> It seems we get some amount of confusion and complexity that obviously led to bugs, but I'm unclear on what we actually gain compared to having a queryStreamManager function that returns a simple JS object with the properties we need (and thus no this anywhere, aliased or not).

The usage of `this` does not require classes. Classes in JavaScript are syntactic sugar around functions, so I'm not sure that you can implement this logic without referring to the object's own properties. Can you make a short example of how you would get rid of the usages of `this`?

garyverhaegen-da: > The usage of `this` does not require classes. Classes in JavaScript are syntactic sugar around functions, so I'm not sure that you can implement this logic without referring to the object's own properties. Can you make a short example of how you would get rid of the usages of `this`?

I'm aware that you _can_ use `this` without classes, but my point is that if you use classes you have to use it, whereas if you don't use classes you can pretend it doesn't exist, and that's usually my preferred approach to JavaScript. Specifically (ignoring TypeScript annotations):

```javascript
class QueryStreamsManager {
  private ws = undefined;
  private resetAllState() {
    // If I want to reset `ws` here, I _have to_ use `this`, and thus worry
    // about all the complexity it brings.
    this.ws = undefined;
  }
}
```

vs

```javascript
function queryStreamsManager() {
  let ws = undefined;
  let resetAllState = () => {
    // no `this` to be seen anywhere, so no need to worry about it being
    // dynamically rebound. I get lexical scope, which is better imo than
    // dynamic.
    ws = undefined;
  };
  return {resetAllState};
}
```

garyverhaegen-da: To be clear, though, I'm not saying I want this PR to be blocked on this (non-trivial, given the size of this class) refactoring, even if I do manage to convince you it's worth doing.

garyverhaegen-da (approved):

language-support/ts/daml-ledger/index.ts
akshayshirahatti-da: right i'll cleanup to remove `this` alias and instead use `manager` in all stream multiplexing related code to reduce confusion.

`QueryStreamsManager` class is useful to encapsulate/isolate all stream management related information into a class away from polluting the ledger code e.g mainly the stream queries registered so far, some state information (e.g  when stream was live) , plus additional data structures for lookup efficiency purposes.  The handle on the `ws` also allows for us to cleanup ws connections on queries change

garyverhaegen-da: > `QueryStreamsManager` class is useful to encapsulate/isolate all stream management related information into a class away from polluting the ledger code e.g mainly the stream queries registered so far, some state information (e.g when stream was live) , plus additional data structures for lookup efficiency purposes. The handle on the `ws` also allows for us to cleanup ws connections on queries change

Which parts of that isolation would not work just as well as local, lexically-scoped variables within a function?

(These are not rhetorical questions, I'm genuinely curious to understand what I'm missing here.)

akshayshirahatti-da: so two main things that I think would not work(at least is not obvious to me) are
 - `streamSubmit` adds/registers/submits a new query/request and we store the state(total set) of queries to be used for creating new websocket connections(and reconnects) and reference it via `manager.queries` which IMO seems cleaner as the underlying impl for handling stream queries should not be a concern of the `Ledger` class.
 - on every new `streamSubmit` we also have to close the existing websocket and create a new one, dont think this works with lexically scoped variables but even if it did overall that be one gigantic method. Am not sure what standard javascript design practices are and if this is the norm.

PR #11688, opened 2021-11-14
Dpp 720 integrate acs reader

nmarton-da:
### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
@@ -47,8 +47,10 @@ import com.daml.platform.store.entries.{
   PartyLedgerEntry,
 }
 import com.daml.platform.store.interning.StringInterning
-
 import java.sql.Connection
+
+import akka.stream.Materializer
+
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
@@ -69,6 +71,7 @@ private class JdbcLedgerDao(
     deduplicationStorageBackend: DeduplicationStorageBackend,
     resetStorageBackend: ResetStorageBackend,
     errorFactories: ErrorFactories,
+    materializer: Materializer,
 ) extends LedgerDao {
 
   import JdbcLedgerDao._
@@ -654,6 +657,17 @@ private class JdbcLedgerDao(
       eventProcessingParallelism = eventsProcessingParallelism,
       metrics = metrics,
       lfValueTranslation = translation,
+      acsReader = new FilterTableACSReader(
+        dispatcher = dbDispatcher,
+        queryNonPruned = queryNonPruned,
+        eventStorageBackend = readStorageBackend.eventStorageBackend,
+        pageSize = eventsPageSize,
+        idPageSize = eventsPageSize * 20,
+        idFetchingParallelism = 2,
+        acsFetchingparallelism = 2,
+        metrics = metrics,
+        materializer = materializer,
+      ),
     )(
       servicesExecutionContext
     )
@@ -768,6 +782,7 @@ private[platform] object JdbcLedgerDao {
       storageBackendFactory: StorageBackendFactory,
       ledgerEndCache: LedgerEndCache,
       stringInterning: StringInterning,
+      materializer: Materializer,
   ): LedgerReadDao =
     new MeteredLedgerReadDao(
       new JdbcLedgerDao(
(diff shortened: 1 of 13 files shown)


nmarton-da:
DO NOT MERGE !!! WIP !!!

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
```diff
@@ -366,35 +365,19 @@ private[appendonlydao] final class TransactionsReader(
...
-            query,
-            nextPageRange[Event](requestedRange.endInclusive),
-          )(requestedRange)
-        })
-        .mapMaterializedValue(_ => NotUsed)
-
-    groupContiguous(events)(by = _.transactionId)
+      .mapAsync(eventProcessingParallelism) { rawResult =>
+        Timed.future(
+          future = Future.traverse(rawResult)(
```
rautenrieth-da, line 375: The parallelism of the surrounding `mapAsync` is bounded by `eventProcessingParallelism`, but the parallelism of `Future.traverse` is (I assume) unbounded. Did you check whether the `mapAsync` has any measurable effect on performance?

You could also move the call to `EventsTable.Entry.toGetActiveContractsResponse` to the `.mapAsync` block, and use `.mapConcat(identity)` below, for _slightly_ less sequential processing.

rautenrieth-da (approved):

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -654,6 +657,17 @@ private class JdbcLedgerDao(
...
       metrics = metrics,
       lfValueTranslation = translation,
+      acsReader = new FilterTableACSReader(
+        dispatcher = dbDispatcher,
+        queryNonPruned = queryNonPruned,
+        eventStorageBackend = readStorageBackend.eventStorageBackend,
+        pageSize = eventsPageSize,
+        idPageSize = eventsPageSize * 20,
+        idFetchingParallelism = 2,
+        acsFetchingparallelism = 2,
```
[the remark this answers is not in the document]
nmarton-da: idPageSize: id retrival is so fast, returns so little data, with 20 it is much faster than the batch retrieval (second phase) - so empirical.
acsFetchingparallelism: scales very well, makes able to amortize call times and sequential processing overhead quite nicely already, it is fairly small not to endanger the server
idFetchingParallelism: this could be probably one in most cases, but as the stream starts, there is a spike in id fetching, keeping this also higher enables a lower lag

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -20,18 +22,93 @@ trait ACSReader {
...
+    pullWorkerSource[QueryTask, Vector[Long]](
+      workerParallelism = idFetchingParallelism,
+      materializer = materializer,
+    )(
+      query => {
+        logger.debug(
+          s"getActiveContractIds query($query)"
+        ) // TODO ACS more meaningful logging here, TODO more metrics???
+        dispatcher
+          .executeSql(metrics.daml.index.db.getActiveContracts)( // FIXME metrics
```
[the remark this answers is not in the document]
nmarton-da: doing it

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/TransactionsReader.scala
nmarton-da: hm it seems to me the deserializeEntry is doing
* all deserialization
* all enriching/Engine execution
BEFORE the future is created, so these things not only not running in parallel, but not even in a future(see traverse code: it executes the future creating function sequentially), which practically eliminates any good effect of the mapAsync.
It looks like that only the package loading is running in a real future. So we seem to have unbounded parallelism for package loading here :thinking: 
If that analyses is true, we need to wrap the creation of the future into a future to at least have parallelism of 2 here.

Hm it is actually quite nasty because it does all this deserializing computation on a thread which should just move around stuff in akka-streams...I am wondering how malicious this is to stream performance....oh my, we might have found the throughput cap I was talking about before :metal:

rautenrieth-da: Good point, most of the work of `deserializeEntry` is executed before the first Future is created.

PR #11728, opened 2021-11-16
[DPP-417][Self-service error codes][Docs] Mention that correlation id…

pbatko-da:
… is usually submission id.

### DIFF
--- a/docs/source/error-codes/self-service/index.rst
+++ b/docs/source/error-codes/self-service/index.rst
@@ -58,9 +58,9 @@ Error
 
         - an `error category`,
 
-        - a `correlation id`.
+        - a `correlation id`,
 
-        - a human readable message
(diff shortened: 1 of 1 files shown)


docs/source/error-codes/self-service/index.rst
```diff
@@ -80,7 +80,9 @@ Error category
                  you can act on particular error codes.
 
 Correlation id
-                  a value whose purpose is to allow a user to clearly identify the request, such that the operator can lookup any log information associated with this error.
+                  A value whose purpose is to allow a user to clearly identify the request,
```
mziolekda, line 83:
```suggestion
                  A value whose purpose is to allow the user to clearly identify the request,
```

mziolekda (approved):
one little nit, otherwise ok

PR #11738, opened 2021-11-17
Fix docs regarding deduplication periods [KVL-1194]

miklos-da:
Summary of changes:
* Use *effective deduplication period* to refer to the actual deduplication period used.
* Added separate section on ledgers supporting command deduplication across participants.

### Pull Request Checklist

trigger the build.

### DIFF
--- a/docs/source/app-dev/app-arch.rst
+++ b/docs/source/app-dev/app-arch.rst
@@ -139,14 +139,18 @@ If you want to make sure that a command is not executed twice, your application 
 Daml ledgers provide a mechanism for :ref:`command deduplication <command-submission-service-deduplication>` to help deal with this problem.
 
 For each command the application provides a command ID and an optional parameter that specifies the deduplication period. If the latter parameter is not specified in the command submission itself, the ledger will fall back to using the configured maximum deduplication period.
-The ledger will then guarantee that commands with the same :ref:`change ID <change-id>` will be ignored within the deduplication period.
+The ledger will then guarantee that commands with the same :ref:`change ID <change-id>` will generate a rejection within the effective deduplication period.
 
 To use command deduplication, you should:
 
-- Use generous values for the deduplication duration. It should be large enough such that you can assume the command was permanently lost if the deduplication period has passed and you still don’t observe any effect of the command on the ledger (i.e. you don't see a transaction with the command ID via the :ref:`transaction service <transaction-service>`).
+- Use generous values for the deduplication duration. It should be large enough such that you can assume the command was permanently lost if the effective deduplication period has passed and you still don’t observe any effect of the command on the ledger (i.e. you don't see a transaction with the command ID via the :ref:`transaction service <transaction-service>`).
 - Make sure you set command IDs deterministically, that is to say: the "same" command must use the same command ID. This is useful for the recovery procedure after an application crash/restart, in which the application inspects the state of the ledger (e.g. via the :ref:`Active contracts service <active-contract-service>`) and sends commands to the ledger. When using deterministic command IDs, any commands that had been sent before the application restart will be discarded by the ledger to avoid duplicate submissions.
-- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the deduplication period, the duplicate submission will safely be ignored. If the deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
+- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the effective deduplication period, the duplicate submission will safely be ignored. If the effective deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
 
+.. note:: Effective deduplication period
+
+  In case of ledgers with command deduplication across participants, the effective deduplication period may be longer than the one specified by the client.
+  To learn more, see the :ref:`Ledger API Services <command-submission-service-deduplication>` documentation.
 
 For more details on command deduplication, see the :ref:`Ledger API Services <command-submission-service-deduplication>` documentation.
 
--- a/docs/source/app-dev/app-arch.rst
+++ b/docs/source/app-dev/app-arch.rst
@@ -149,7 +149,7 @@ To use command deduplication, you should:
 
 .. note:: Effective deduplication period
 
-  In case of ledgers with command deduplication across participants, the effective deduplication period may be longer than the one specified by the client.
+  In case of ledgers with command deduplication across participants, the effective deduplication period may be longer than the one specified in the command.
   To learn more, see the :ref:`Ledger API Services <command-submission-service-deduplication>` documentation.
 
 For more details on command deduplication, see the :ref:`Ledger API Services <command-submission-service-deduplication>` documentation.
--- a/docs/source/app-dev/services.rst
+++ b/docs/source/app-dev/services.rst
@@ -83,7 +83,7 @@ The command submission service deduplicates submitted commands based on their :r
 
 - Applications can provide a :ref:`deduplication duration <com.daml.ledger.api.v1.Commands.deduplication_duration>` for each command. If this parameter is not set, the default maximum deduplication period is used.
 - A command submission is considered a duplicate submission if the ledger API server is aware of another command within the deduplication period and with the same :ref:`change ID <change-id>`.
-- Duplicate command submissions will be ignored until either the deduplication period of the original command has passed or the original submission was rejected (i.e. the command failed and resulted in a rejected transaction), whichever comes first.
+- Duplicate command submissions will generate a rejection until either the effective deduplication period of the original command has passed or the original submission was rejected (i.e. the command failed and resulted in a rejected transaction), whichever comes first.
 - Command deduplication is only *guaranteed* to work if all commands are submitted to the same participant. Ledgers are free to perform additional command deduplication across participants. Consult the respective ledger's manual for more details.
 - A command submission will return:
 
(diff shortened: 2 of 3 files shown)


docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,17 @@ The command submission service deduplicates submitted commands based on their :r
 
 For details on how to use command deduplication, see the :ref:`Application Architecture Guide <command-deduplication>`.
 
+.. note:: Ledgers with command deduplication across participants
+
+  - The ledger server may extend the deduplication period specified in the request arbitrarily, up until the maximum deduplication duration specified in the :ref:`ledger configuration <ledger-configuration-service>`.
```
gerolf-da, line 99:
```suggestion
  - The ledger may extend the deduplication period specified in the request arbitrarily, up until the maximum deduplication duration specified in the :ref:`ledger configuration <ledger-configuration-service>`.
```

docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,17 @@ The command submission service deduplicates submitted commands based on their :r
...
+    The deduplication period chosen by the ledger server is the *effective deduplication period*.
```
gerolf-da, line 100:
```suggestion
    The deduplication period chosen by the ledger is the *effective deduplication period*.
```

docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,17 @@ The command submission service deduplicates submitted commands based on their :r
...
+
+  - Regardless, the deduplication period specified in the request is always checked against the configured maximum deduplication duration.
+
+  - A command submission is considered a duplicate submission if the ledger server is aware of another command within the *effective* deduplication period and with the same :ref:`change ID <change-id>`.
```
gerolf-da, line 104:
```suggestion
  - A command submission is considered a duplicate submission if the ledger is aware of another command within the *effective* deduplication period and with the same :ref:`change ID <change-id>`.
```

docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,17 @@ The command submission service deduplicates submitted commands based on their :r
...
+
+  - Command deduplication is *guaranteed* to work even if commands are submitted to different participants.
```
gerolf-da, line 106: Is this true for VMBC and Canton? Maybe we can just mention both in this sentence.

miklos-da: Sounds good. This is certainly true for VMBC, @andreaslochbihler-da can comment on Canton.

docs/source/app-dev/services.rst
nicu-da: Do we want to capture that it also accounts for time skews? Currently we basically account for `min_record_time` by adding the `min_skew` to calculate the deduplicate until

docs/source/app-dev/services.rst
miklos-da: @gerolf-da Opted for a slightly different presentation (and left out Canton till I get a confirmation) -- let me know what you think.

docs/source/app-dev/app-arch.rst
```diff
@@ -139,14 +139,18 @@ If you want to make sure that a command is not executed twice, your application
...
-- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the deduplication period, the duplicate submission will safely be ignored. If the deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
+- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the effective deduplication period, the duplicate submission will safely be ignored. If the effective deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
 
+.. note:: Effective deduplication period
+
+  In case of ledgers with command deduplication across participants, the effective deduplication period may be longer than the one specified in the command.
```
andreaslochbihler-da, line 152: The effective deduplication period may always be longer than the one specified in the command. I don't think this is specific to VMBC ledgers.
```suggestion
  The effective deduplication period may be longer than the one specified in the command, but it never is shorter.
```

docs/source/app-dev/services.rst
```diff
@@ -105,6 +105,10 @@ For details on how to use command deduplication, see the :ref:`Application Archi
 
   - Command deduplication is *guaranteed* to work even if commands are submitted to different participants.
 
+  The following ledger integrations support command deduplication across participants:
+    - `Daml Driver for VMware Blockchain <https://www.digitalasset.com/daml-for-vmware-blockchain/>`__
+    - :ref:`Daml Sandbox <sandbox-manual>`
```
gerolf-da, line 110: There really is only one single participant. So while the statement technically is true, it's not very helpful.

docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,21 @@ The command submission service deduplicates submitted commands based on their :r
 
 For details on how to use command deduplication, see the :ref:`Application Architecture Guide <command-deduplication>`.
 
+.. note:: Ledgers with command deduplication across participants
```
andreaslochbihler-da, line 97:
```suggestion
.. note::
```

docs/source/app-dev/services.rst
```diff
@@ -94,6 +94,21 @@ The command submission service deduplicates submitted commands based on their :r
...
+
+  - The ledger may extend the deduplication period specified in the request arbitrarily, up until the maximum deduplication duration specified in the :ref:`ledger configuration <ledger-configuration-service>`.
```
andreaslochbihler-da, line 99:
```suggestion
  - The ledger may extend the deduplication period specified in the request arbitrarily, even beyond the maximum deduplication duration specified in the :ref:`ledger configuration <ledger-configuration-service>`.
```

docs/source/app-dev/services.rst
andreaslochbihler-da: Command deduplication is formally defined via the completion event stream. Since Canton produces a completion event only on the participant that made the submission, Canton doesn't deduplicate commands across participants.

docs/source/app-dev/services.rst
```diff
@@ -83,7 +83,7 @@ The command submission service deduplicates submitted commands based on their :r
 
 - Applications can provide a :ref:`deduplication duration <com.daml.ledger.api.v1.Commands.deduplication_duration>` for each command. If this parameter is not set, the default maximum deduplication period is used.
 - A command submission is considered a duplicate submission if the ledger API server is aware of another command within the deduplication period and with the same :ref:`change ID <change-id>`.
-- Duplicate command submissions will be ignored until either the deduplication period of the original command has passed or the original submission was rejected (i.e. the command failed and resulted in a rejected transaction), whichever comes first.
+- Duplicate command submissions will generate a rejection until either the effective deduplication period of the original command has passed or the original submission was rejected (i.e. the command failed and resulted in a rejected transaction), whichever comes first.
```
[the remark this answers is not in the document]
nicu-da: The backwards looking will land in the 2.0 release. The PR is ready, but we need a follow up PR and decided not to rush for this release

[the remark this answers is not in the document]
miklos-da: Good point, applied.

docs/source/app-dev/services.rst
miklos-da: Are you suggesting 'even beyond' because of accounting for time skew or for a different reason?

docs/source/app-dev/services.rst
miklos-da: You mean for Daml Sandbox, right? I've removed that and put this note above where we mention 'command deduplication across participants'.

docs/source/app-dev/services.rst
miklos-da: OK, removed it.

docs/source/app-dev/services.rst
andreaslochbihler-da: Time skews are one thing where it can already happen for kv-based ledgers.
The other thing is that Canton has to report a deduplication offset rather than a duration; but if the submission had specified a duration, then there may not be a suitable completion offset around the time where the deduplication duration starts. So if the deduplication duration already is max dedup time, we cannot really return an offset that was published within the max dedup duration without incurring the risk that we're shortening the effective deduplication duration. So instead, Canton doesn't even bother to find a completion offset close the the beginning of the deduplication duration. Instead, we return the first known offset on the ledger or the offset after the last completion for the change ID. So for an unpruned ledger and a fresh change ID, the effective deduplication period is unbounded. Given that longer durations provide a stronger statement to the application, the application should actually be happy about this.

andreaslochbihler-da: If there is already a completion for the same change ID, then Canton typically does not extend the deduplication period before the completion if the earlier completion wouldn't normally fall into the deduplication period. But as always, with clock drift, this may appear to the application as if some extension happened.

docs/source/app-dev/app-arch.rst
```diff
@@ -139,14 +139,18 @@ If you want to make sure that a command is not executed twice, your application
...
-- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the deduplication period, the duplicate submission will safely be ignored. If the deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
+- If you are not sure whether a command was submitted successfully, just resubmit it with the same deduplication duration. If the new command was submitted within the effective deduplication period, the duplicate submission will safely be ignored. If the effective deduplication period has passed, you can assume the command was lost or rejected and a new submission is justified.
```
nicu-da, line 148: Should we correct the `will safely be ignored` here as well?

miklos-da: Good point, corrected for consistency.

andreaslochbihler-da (approved):
LGTM. Thanks for updating the docs.

PR #11740, opened 2021-11-17
release 1.18.0-snapshot.20211117.8399.0.a05a40ae

realvictorprm:
commit log:
```

- [Daml Script] When run over the JSON API, Daml Script can now use a
  subset of the claims in the token. E.g., `submit p` works even if you
  have a token with `actAs = [p, p2]`.

- [Integration Kit] - ledger-api-bench-tool - All stream configuration parameters can be defined through the YAML configuration file.
[TS-BINDINGS] Re-enable ws multiplexing for stream queries after resolving the reconnect connection close bug associated with ws state and liveness.
- [JSON API] ``actAs`` and ``readAs`` may be specified for create, exercise,
  create-and-exercise, non-WS fetch, and non-WS query.
  See `issue #11454 <https://github.com/digital-asset/daml/pull/11454>`__.
[TS-Bindings] Ws multiplexing for stream queries is disabled by default as we investigate issues
of intermittent websocket closures.

* There is now an Early Access Helm chart to start all of the Connect components on a Kubernetes cluster.

[JSON-API] Make certain Hikari cp connection pool properties configurable via jdbc conf string, the properties are listed below
poolSize -- specifies the max pool size for the database connection pool
minIdle -- specifies the min idle connections for database connection pool
connectionTimeout -- long value, specifies the connection timeout for database connection pool
idleTimeout -- long value, specifies the idle timeout for the database connection pool
- [Integration Kit] - ledger-api-bench-tool can run a workflow consisting of the command submission phase and stream benchmark phase.
- [JSON API] While updating the contract table for a query, if the DB appears to be slow,
  JSON API will slow down its own inserts and deletes at some point rather than construct
  ever-larger INSERT and DELETE batch commands.
  See `issue #11589 <https://github.com/digital-asset/daml/pull/11589>`__.
[ledger-api] - Return max_deduplication_duration as part of the metadata sent with the gRPC status for commands rejected because of invalid deduplication periods

- [Integration Kit] - ledger-api-bench-tool - added flexible party and template filters for transactions, transaction trees and active contract streams



[ledger-api] - Return FAILED_PRECONDITION gRPC status code, instead of INVALID_ARGUMENT, for an invalid command deduplication duration

- [@daml/react] Our React bindings now request the `react` library as a
  peer dependency, and will accept React 17 in addition to the current
  React 16. For new projects, the `create-daml-app` template now uses
  React 17.

- [Integration Kit] Document KV-specific self-service errors

[TS bindings] When using daml-react and daml-ledger, all streaming request to the query endpoint will be multiplexed through a single web socket.
[daml LF] - Add LoggingValue instance for Time.Timestamp
- [Integration Kit] - ledger-api-bench-tool can generate synthetic contracts with multiple observers
[api-test-tool] - remove AppendOnly prefix from test suites as the append-only schema is the only one left
[JSON-API] fixes a bug related to the matchedQueries value returned for websocket multiqueries,
this only happens for patterns where the multiqueries contain a mixture of queries with and without
offsets.
- [Integration Kit] - ledger-api-bench-tool can generate test contracts with configurable payload size.
- [Integration Kit] - Added multi-template support for command submission in the ledger-api-bench-tool
- [Sandbox] The ``--eager-package-loading`` flag has been deprecated. It
  hasn't actually done anything for many releases; packages are always
  loaded eagerly. This does not affect Daml on SQL, which does support
  lazy package loading.

- [Daml Studio] Fix a bug where script results in Daml Studio
  sometimes do not show up.


- [Integration Kit] - The ledger-api-bench-tool is now capable of generating test contracts for testing purposes.
- [JSON API] Fixed a rare error that manifested as
  ‘violates foreign key constraint "contract_tpid_fkey"
   Detail: Key (tpid)=(...) is not present in table’
  when attempting to run queries and goes away on JSON API restart.
  See `issue #11330 <https://github.com/digital-asset/daml/pull/11330>`__.
- [Participant] All participants now use the new append-only schema. Existing databases will
  automatically upgrade to the new schema the first time a participant/ledger is started.

- [Daml Triggers] Triggers now support readAs parties. They can be
  specified via `--ledger-readas a,b,c`. As part of this change
  testRule gained an extra argument to specify the `readAs`
  parties. If you previously used

  
  testRule trigger party acsBuilder commandsInFlight s
  

  you now need to use

  
  testRule trigger party [] acsBuilder commandsInFlight s
  

- [JSON API] Several kinds of gRPC server errors are now reported with
  associated HTTP statuses; for example, a Daml-LF interpreter error now
  returns a 400 instead of a 500, and an exercise on an archived contract
  returns a 409 Conflict instead of a 500.  Errors internal to JSON API
  (e.g. internal assertion failures) are no longer detailed in the HTTP
  response; their details are only logged.
  See `issue #11184 <https://github.com/digital-asset/daml/pull/11184>`__.

- [Daml Stdlib] Add `debugRaw` as a convenience wrapper around
  `traceRaw` when used inside a do-block. `debugRaw` compares to
  `debug` like `traceRaw` compares to `trace` meaning it expects a
  `Text` instead of calling `show` on an expression.


sandbox-next - Disable participant command deduplication

kvutils - Command deduplication uses the configured time model (static/wall) and not running always using wall-clock
[JSON API] Solving a bug that could cause the JSON API to return
correct result if a contract with the same key is observed twice
required a schema change. The JSON API data needs to be dropped
and the query store needs to reset. If you are migrating from a
previous version, either reset your database manually or start
the HTTP JSON API with one of the options that regenerate the
schema (`create-only`, `create-if-needed-and-start`, `create-and-start`).

Participant HA feature added: Indexer component is guarded by database locks, and executes in a critical section (prevents overlapping indexer activity). Supported by Oracle and Postgresql database backends, and not supported by H2.



CI - add check to PR for protobuf breaking changes against the target branch head
- [Daml Studio] Scripts no longer fail when allocating a party with an
  empty display name. See https://github.com/digital-asset/daml/issues/11099.




- [JSON-API] make key_hash indexes non-unique, this fixes a bug where a duplicate key conflict was raised on the query store when the same contract was being witnessed twice by two separate parties

* [Trigger Service] Enable the new `tablePrefix` setting in the `--jdbc`
  flag to add a prefix to all tables used by the trigger service to
  avoid collisions with other components using the same db-schema.
* [Trigger Service] Enable the new ``--allow-existing-schema`` flag to
  initialize the trigger service on a database with a pre-existing
  schema.



- [Daml Assistant] Add support for `--max-inbound-message-size` flag to the Ledger Export tool.

Add cons, find, delete and deleteBy functions to NonEmpty module.
- [JSON API] Under rare conditions, a multi-template query backed by database
  could have an ACS portion that doesn't match its transaction stream, if
  updated concurrently.  This conditions is now checked and accounted for.
  See `issue #10617 <https://github.com/digital-asset/daml/pull/10617>`__.
- The OAuth2 Middleware now obfuscates its Client Secret when logging
  its config.
- [Integration Kit] We have added ``loggingContext`` as an implicit
  parameter to more _kvutils_ trait methods. Implementors may need to do
  the same in their trait implementations. This can make it easier to
  log with the appropriate context.
java-client-bindings - the original full completion is included in the `CompletionResponse` when available
Daml on SQL, Integration Kit, Sandbox: Drop support for TLS 1.0 and 1.1 in Ledger API.
[Triggers Service] The service now starts by default using wall-clock time instead
of static time. If you want to run using static time, you need to do so explicitly
using the new '-s' or '--static-time' CLI option. If you were already using '-w'
or '--wall-clock-time' the flag has no effect. It's anyway safe to leave it there.
Sandbox: Add CLI flag `--min-tls-version` to select minimum enabled TLS version for participant server.
```


stefanobaghino-da (approved):

cocreature (approved):
Thanks!

realvictorprm:
Yaaay

realvictorprm:
Manual tests passed on Windows.

realvictorprm:
Manual tests passed on Linux.

PR #11744, opened 2021-11-17
 LF: Simplify transaction versionning for interface

remyhaemmerle-da:
We revert #11626, and just change the way transaction version is
computed:
    
- As before, Node version is calculated from the
  package of the template ID action.
    
- Transaction version is the max of the version of all the nodes,
   instead of the root nodes.
    
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/VersionRange.scala
+++ b/daml-lf/data/src/main/scala/com/digitalasset/daml/lf/VersionRange.scala
@@ -21,13 +21,7 @@ final case class VersionRange[V](
 
   require(min <= max)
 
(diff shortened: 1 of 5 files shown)


cocreature (commented):
Looks good codewise, let’s wait until our meeting later before deciding on an approach.

One downside of this is that to get the version of a projection you have to retraverse the whole subtransaction. In the other case you can only look at the versions of the roots. I don’t think that’s a huge issue but worth keeping in mind.

cocreature (approved):
Thanks!

PR #11758, opened 2021-11-18
Drop Scala 2.12 compatibility layer from //daml-lf/data

cocreature:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/InsertOrdSet.scala
```diff
@@ -40,16 +46,35 @@ final class InsertOrdSet[T] private (_items: Queue[T], _hashSet: HashSet[T])
...
-  // Here only for 2.12 (harmless in 2.13); placed in InsertOrdSetCompanion the
-  // implicit gets in an unwinnable fight with Set's version
-  implicit override def canBuildFrom[A]: Factory[A] =
-    super.canBuildFrom
-
+  private final class InsertOrdSetBuilder[T] extends ReusableBuilder[T, InsertOrdSet[T]] {
+    var m: InsertOrdSet[T] = empty;
+    override def clear(): Unit = {
+      m = empty;
+    }
```
remyhaemmerle-da, line 73: Those `;` are a bit strange.

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/ImmArray.scala
```diff
@@ -11,7 +11,10 @@ import scalaz.syntax.applicative._
 import scalaz.{Applicative, Equal, Foldable, Order, Traverse}
 
 import scala.annotation.tailrec
+import scala.collection.StrictOptimizedSeqFactory
 import scala.collection.compat.immutable.ArraySeq
```
remyhaemmerle-da, line 15:
```suggestion
import scala.collection.immutable.ArraySeq
```

remyhaemmerle-da (commented):
Thanks. 

I think we could get rid of `@maven//:org_scala_lang_modules_scala_collection_compat`

daml-lf/data/src/main/scala/com/digitalasset/daml/lf/data/InsertOrdSet.scala
cocreature: just copied the code around not sure why they were there before. Killed them now

cocreature:
> I think we could get rid of `@maven//:org_scala_lang_modules_scala_collection_compat`

dropped all dependencies on the compat lib

Issue #11703, opened 2021-11-15
Check types on exercise/fetch by template for interface choices


cocreature:
Both in Daml and over the ledger API you can exercise a choice from an interface by template id,e.g.,
```
exercise (cid : ContractId InterfaceImpl) InterfaceChoice
```
However, currently we only check that `cid` corresponds to a template that implements the interface, we do not check that it is `InterfaceImpl`. So someone can pass a cid of a different template that implements the interface and things will succeed.

This is at best confusing and at worse a potential attack vector so we need to fix this.

This is slightly tricky since we desugar such a choice as
```
instance IsToken t => HasExercise t GetRich (ContractId Token) where
  exercise cid = GHC.Types.primitive @"UExerciseInterface" (toTokenContractId cid)
```
This means that we generate the same call regardless of whether we call by interface or by template. We also don't have the ability to create a monomorphic `HasExercise` instance for each template, since the desugarer doesn't know which fixed choices are available (and they may not even be in scope).

To fix this, we should be able to change the primitive for `UExerciseInterface` to accept an optional typerep and validate against that.

Another option would be to accept a boolean predicate on the template payload that we evaluate between the fetch & the exercise. However, if we want to make failures of that catchable we have to record a fetch in the case of failure to track transaction inputs which is somewhat tricky.

so for now, the typerep solution seems like the better option.

sofiafaro-da:
This is done! Thanks to @akrmn for merging the final PR into ghc :)

sofiafaro-da:
There's even a test: https://github.com/digital-asset/daml/blob/main/compiler/damlc/tests/daml-test-files/InterfaceTypeRepCheck.daml

Issue #11558, opened 2021-11-04
Consistently reject exercise on interface choices via template id in LF


cocreature:

remyhaemmerle-da:
Fixed by #11584.

cocreature:
probably deserves a bullet point in #10917

remyhaemmerle-da:
create  #11698 and add a coment there.

cocreature:
@remyhaemmerle-da given that you created #11698 to cover testing, should we close this for now?

cocreature:
Closing, @remyhaemmerle-da please reopen if there’s something left in here.

PR #11773, opened 2021-11-18
Add engine tests for interfaces.

sofiafaro-da:
Part of #10917 

Adds engine tests using the new commands. This uncovered a few issues:

- getDependencies doesn't work when given an interfaceId as a templateId
  (such as when exercising an interface via the ledger api). I patched
  it up so it deals with interface ids. We've already discussed a way to
  simplify getDependencies that will also avoid this in the future, but I'll
  leave that to the future.

- issue #11703
  is confirmed via some tests that don't pass (and are currently disabled)

- PackageInterface.lookupTemplateChoice returns inherited choices, when it
  should only return own choices. At least, the typechecker assumes it doesn't --
  this affects #11558
  -- as does the command preprocessor. I'll leave that to a separate PR.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/EngineTest.scala
```diff
@@ -1925,6 +1925,245 @@ class EngineTest
     }
   }
 
+  "interfaces" should {
```
remyhaemmerle-da, line 1928: Please put this in a separate file.
`EngineTest` is already a giant mess we do not wan to grow.

sofiafaro-da: Moved!

remyhaemmerle-da (approved):
LGTM

PR #11775, opened 2021-11-18
Move ghc-lib{,-parser} to bazel-haskell-deps

akrmn:
### Pull Request Checklist

trigger the build.


akrmn:
Ignore the fact that this currently targets the wrong branch :)

bazel-haskell-deps.bzl
```diff
@@ -17,6 +17,12 @@ load("@os_info//:os_info.bzl", "is_linux", "is_windows")
 load("@dadew//:dadew.bzl", "dadew_tool_home")
 load("@rules_haskell//haskell:cabal.bzl", "stack_snapshot")
 
+GHC_LIB_REV = "0f0fb98205a8df24d0a4c4b869345c59"
+GHC_LIB_SHA256 = "ff32aee85a7c399814912ac5d3f3a7060860d3159f06fbb7ed8b96f326cc2c86"
+GHC_LIB_VERSION = "8.8.1.20211118"
```
akrmn, line 22: Hm, the date at the end of this version is going to be a bit painful to work with. We could print it from `ci/da-ghc-lib/compile.yml`, but the problem is that it also ends up in `stackage_snapshot.json` and `stackage_snapshot_windows.json`, so we would have to run `bazel run @stackage-unpinned//:pin` for those anyway (they have a big `__GENERATED_FILE_DO_NOT_MODIFY_MANUALLY` at the top), defeating the purpose of the PR.

I tried setting the `version` arg in the call to `haskell_cabal_library` to a constant `"8.8.1"` (no trailing date) but it doesn't seem to change the resulting `stackage_snapshot.json`

akrmn: I ended up patching `ghc-lib`'s `CI.hs` script to take a version suffix, which, if provided, replaces the date at the end. This way, the version is just `8.8.1`, so `stackage_snapshot.json` doesn't need to be updated

cocreature (approved):
Very nice, thanks a lot!

PR #11776, opened 2021-11-18
Write exception reports into the log-file in case of SSL misconfiguration [DPP-759]

mziolekda:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/LedgerApiServer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/LedgerApiServer.scala
@@ -4,7 +4,6 @@
 package com.daml.platform.apiserver
 
 import java.util.concurrent.Executor
-
 import com.daml.ledger.api.tls.TlsConfiguration
 import com.daml.ledger.resources.{Resource, ResourceContext, ResourceOwner}
 import com.daml.logging.{ContextualizedLogger, LoggingContext}
@@ -13,6 +12,7 @@ import com.daml.ports.Port
 import io.grpc.ServerInterceptor
 
 import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success}
 
 private[daml] final class LedgerApiServer(
     apiServicesOwner: ResourceOwner[ApiServices],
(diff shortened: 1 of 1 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/LedgerApiServer.scala
```diff
@@ -65,6 +65,11 @@ private[daml] final class LedgerApiServer(
         override def servicesClosed(): Future[Unit] =
           servicesClosedPromise.future
       }
+    }).transformWith {
+      case Failure(ex) =>
```
tudor-da, line 69: Why do we need to log explicitly here? Isn't the Resource going to bubble up the failure all the way through the chain?

tudor-da (approved):
LGTM if you deem it to bring value

ledger/participant-integration-api/src/main/scala/platform/apiserver/LedgerApiServer.scala
tudor-da: So here is how they bubble up. Unless the exception is a `SuppressedStartupException`, I don't see how it would skip logging.

mziolekda: I think it skips because of timing, the log is protracted until the very end of the shutdown and is not printed when something else pops up in between

mziolekda: This is not purely theoretical, I have been able to reproduce it many times

tudor-da: >I think it skips because of timing, the log is protracted until the very end of the shutdown and is not printed when something else pops up in between

I see. The possibility of skipping failure logs on application startup is not ideal so we should revisit this part at some point since we perform a lot of startup operation in these resources. Nice findings. LGTM 👍

PR #11780, opened 2021-11-18
kvutils - Remove unused protobuf field to prevent breaking changes 

nicu-da:
- the `submission_time` field was added as part of the 1.18 work but was later on replaced after the release branch was created
- to prevent backwards-incompatible changes and not store this field that is not used, easiest solution is to just remove it for the 1.18 release

### Pull Request Checklist

trigger the build.


SamirTalwar-DA (approved):
Seems sensible to me.

ledger/participant-state/kvutils/src/main/protobuf/com/daml/ledger/participant/state/kvutils/store/state.proto
```diff
@@ -54,12 +54,10 @@ message DamlCommandDedupKey {
 
 message DamlCommandDedupValue {
   reserved 1; // was record_time
+  reserved 3; // was submission_time
```
gerolf-da, line 57: If we introduced the `submission_time` field after SDK 1.17.x (that is, within the timeframe of 1.18), then there's no need to put it as "reserved" field, but you can just remove it.

gerolf-da: Within a release, there's no need to preserve backwards compatibility.

nicu-da: The field was introduced after 1.17.

If this is considered safe, and it's safe to break compatibility between 1.18 RC1 and 1.18 RC2 then there's actually no need for https://github.com/digital-asset/daml/pull/11781

Issue #11755, opened 2021-11-17
make Scala 2.12-suppressed optimizations


S11001001:
With #11315 done, we can do some small optimizations:

* NonEmpty
  * [ ] if it makes a difference, redefine `pour` to abstract over `IterableFactory` and `MapFactory`, if those implementations do optimization for same-type conversions
  * [ ] redesign `pour`'s API to put the factory type first, no more tricky currying
  * [ ] just use `toSeq` for `toSeq`
  * [ ] define `collect`, `map`, and `flatMap` as proper generalized wrappers (needs typeclass)
* libs-scala
  * [ ] remove DirectExecutionContextInternal as commented
* ledger-service
  * [ ] remove `Mark0`
  * [ ] remove SkipScala212
  * [ ] remove NullPointerException branch in TestUtil.readFile
* participant-integration-api
  * [ ] EventsBuffer: return to 5108eefc8f4371d8591f04092ad4a9849ebcaeb1's `searchableView` from #9775

PR #11783, opened 2021-11-18
Use ExecutionContext.parasitic instead of DirectExecutionContext.

SamirTalwar-DA:
We no longer need the latter, as of Scala 2.13.

Part of #11755.

### Pull Request Checklist

trigger the build.


nicu-da (approved):

stefanobaghino-da:
Seems to be rather part of https://github.com/digital-asset/daml/issues/11315

ledger/ledger-api-client/src/main/scala/com/digitalasset/ledger/client/services/testing/time/StaticTime.scala
```diff
@@ -90,11 +89,11 @@ object StaticTime {
...
-          // format: OFF
-          instantSource ~> killSwitch ~> updateClock ~> broadcastTimes.in
-                                                        broadcastTimes.out(0) ~> sinkHead
-                                                        broadcastTimes.out(1) ~> ignore
-          // format: ON
+            // format: OFF
+            instantSource ~> killSwitch ~> updateClock ~> broadcastTimes.in
+            broadcastTimes.out(0) ~> sinkHead
+            broadcastTimes.out(1) ~> ignore
+            // format: ON
```
stefanobaghino-da, line 96: Please leave this unchanged and keep the `broadcastTime` aligned.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/PaginatingAsyncStream.scala
```diff
@@ -73,9 +72,7 @@ private[platform] object PaginatingAsyncStream {
           query(offset).map { result =>
             val nextPageOffset: Option[Off] = result.lastOption.map(getOffset)
             Some((nextPageOffset, result))
-          }(
-            DirectExecutionContext
-          ) // run in the same thread as the query, avoid context switch for a cheap operation
```
stefanobaghino-da, line 78: > run in the same thread as the query

I would challenge this claim. I would also challenge any usage of `parasitic` as well, given that it's based on this assumption, which happens to be true most of the times, but not deterministically so. But that's definitely for another time.

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityChecker.scala
```diff
@@ -327,7 +321,7 @@ object IntegrityChecker {
           case Failure(exception) =>
             exception.printStackTrace()
             sys.exit(1)
-        }(DirectExecutionContext)
+        }
```
stefanobaghino-da, line 324: Should you have used `parasitic` here?

stefanobaghino-da (approved):

mziolekda (changes requested):
@SamirTalwar-DA 
I would like you to hold off from merging this change for a couple of days. We are in the mode where a number of changes need to be backported to 1.18. A change like this that cuts across so many classes can be very disruptive to this effort. I will accept the PR when we are out of the woods.

Thanks for your understanding

SamirTalwar-DA:
@mziolekda: Absolutely!

@stefanobaghino-da: Thanks for catching those. I will address them when we're good to go.

ledger/participant-state/kvutils/tools/src/main/scala/ledger/participant/state/kvutils/tools/integritycheck/IntegrityChecker.scala
SamirTalwar-DA: I don't see why it ever used `DirectExecutionContext` when there's a perfectly good thread pool available.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/PaginatingAsyncStream.scala
SamirTalwar-DA: Agreed, which is why I deleted it. Someone else's problem, I am afraid.

SamirTalwar-DA:
@mziolekda, can we unblock this yet?

mziolekda (approved):
I have confirmed that the performance of the application is the same after the change.
Nice work, thank you Samir!

SamirTalwar-DA:
Thanks!

PR #11811, opened 2021-11-22
Backport: Restart the submission interpretation in case of a race [DPP-737]

gerolf-da:
1. Daml interpretation
2. Determine a suitable ledger effective time

interpretation.
now archived.

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
@@ -71,13 +71,13 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
                 if (maxUsedTime.forall(_ <= commands.commands.ledgerEffectiveTime)) {
                   Future.successful(Right(cer))
                 } else if (!cer.dependsOnLedgerTime) {
-                  logger.debug(
+                  logger.info(
                     s"Advancing ledger effective time for the output from ${commands.commands.ledgerEffectiveTime} to $maxUsedTime"
                   )
                   Future.successful(Right(advanceOutputTime(cer, maxUsedTime)))
                 } else if (retriesLeft > 0) {
                   metrics.daml.execution.retry.mark()
-                  logger.debug(
+                  logger.info(
                     s"Restarting the computation with new ledger effective time $maxUsedTime"
                   )
                   val advancedCommands = advanceInputTime(commands, maxUsedTime)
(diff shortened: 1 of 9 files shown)


tudor-da (approved):

gerolf-da:
thank you, @cocreature, for merging in the latest changes from the target branch.

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -86,18 +86,32 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
...
+                  if (retriesLeft > 0) {
+                    metrics.daml.execution.retry.mark()
+                    logger.info(
+                      s"Some input contracts could not be found. Restarting the computation. Missing contracts: ${contracts
+                        .mkString("[", ", ", "]")}"
+                    )
+                    loop(commands, submissionSeed, ledgerConfiguration, retriesLeft - 1)
+                  } else {
+                    logger.info(
+                      s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
```
pbatko-da, line 101: Not logging the missing contracts after the last attempt?

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
```diff
@@ -86,18 +86,32 @@ private[apiserver] final class LedgerTimeAwareCommandExecutor(
...
+                    Future.successful(Left(ErrorCause.LedgerTime(maxRetries)))
+                  }
+
+                // An error while looking up the maximum ledger time for the used contracts. The nature of this error is not known.
+                // Not retrying automatically. All other automatically retryable cases are covered by the logic above.
                 case error =>
                   logger.info(
-                    s"Lookup of maximum ledger time failed. This can happen if there is contention on contracts used by the transaction. Used contracts: ${usedContractIds
-                      .mkString(", ")}. Details: $error"
+                    s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft}. Used contracts: ${usedContractIds
```
pbatko-da, line 111:
```suggestion
                    s"Lookup of maximum ledger time failed after ${maxRetries - retriesLeft} retries. Used contracts: ${usedContractIds
```

ledger/participant-integration-api/src/main/scala/platform/apiserver/execution/LedgerTimeAwareCommandExecutor.scala
gerolf-da: You're right. Thanks.

gerolf-da: I also set the other `debug` logs to `info`, as they likely won't happen too often (i.e. there's no logspam), but it's useful to know that it happens.

Issue #10283, opened 2021-07-15
Large numbers of parties cause the dreaded `RESOURCE_EXHAUSTED` error


da-tanabe:
A large number of parties onboarded onto any ledger will eventually cause clients that list the universe of parties to receive this lovely error (with the max size of course highly dependent on the number of parties):

```
io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304: 4870993
	at io.grpc.Status.asRuntimeException(Status.java:535)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:533)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:553)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:68)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:739)
```

The right solution is to probably change the `ListKnownParties` call to return a `stream` of `ListKnownPartiesResponse` instead of a single object, and to chunk up the responses:

https://github.com/digital-asset/daml/blob/11e5dd37e730156b97748d67914b5e3f14a48625/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/party_management_service.proto#L38-L44

With the current API design, this endpoint will only ever return more data over time, and raising this limit globally, forever doesn't seem like a good idea.

cocreature:
Definitely in favor of making this a streaming endpoint/adding a streaming endpoint to preserve backwards compatibility.

meiersi-da:
Adding a streaming endpoint is definitely an option, and would match the event-streaming nature of our system. An alternative to consider is to expose views onto the state of the system using the resource oriented design of Google's Cloud APIs. Listing collections is paged there as per these design guidelines: https://google.aip.dev/132

@da-tanabe: thanks for reporting this. What is the priority on a fix for this? 

(FYI: @gerolf-da @mziolekda )

cocreature:
We already do “pagination” on the ACS endpoint by using streaming there. Using something else for the parties endpoints doesn’t seem great from a user pov.

da-tanabe:
@meiersi-da it's a simple thing to work around (we just keep cranking up the limit), but it's definitely not ideal.

[...]

PR #11818, opened 2021-11-22
Add gRPC definitions for participant user management service

meiersi-da:
This PR introduces the gRPC definitions for an experimental user management service to manage users and their rights for interacting with the Ledger API served by a participant node. The PR does not change the set of actually exposed services of a Ledger API server. 

I expect to implement the user management service in a later PR behind a feature flag such that we can start a PoC phase to evaluate whether all the Daml SDK tools can be made to work well (or hopefully better) building on this user management service. I expect that we'll iterate further on the service's gRPC definitions as part of this PoC phase.

@bame-da @cocreature @stefanobaghino-da could I ask one of you to approve the PR so we can start the PoC implementation work?

FYI: @da-tanabe @adriaanm-da @gerolf-da and others, feel free to review the code directly on this PR and add comments here. I will then take care of addressing the comments on a new PR in case this one is already merged.

### Pull Request Checklist

- [x] If you mean to change the status of a component, please make sure you keep [the Component Status page](https://github.com/digital-asset/daml/blob/main/docs/source/support/component-statuses.rst) up to date.

trigger the build.

### DIFF
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
@@ -1,0 +1,161 @@
+// Copyright (c) 2021 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+package com.daml.ledger.api.v1.admin;
+
+option java_outer_classname = "UserManagementServiceOuterClass";
+option java_package = "com.daml.ledger.api.v1.admin";
+option csharp_namespace = "Com.Daml.Ledger.Api.V1.Admin";
+
+import "google/protobuf/empty.proto";
+
+
+// Manage users and their rights for interacting with the Ledger API served by a participant node.
+//
+// The API roughly follows the Google API style guidelines for resource-oriented design
+// (https://cloud.google.com/apis/design/resources) with the following simplifications to
+// ease its implementation:
+//
+// 1. List methods do not (yet) support pagination,
+//    as we expect to have fewer than 10k users and 1k rights per user.
+//
+// 2. Resources are not named as per the Google API style guide,
+//    as this would be in contrast with the other Ledger API services.
+//
+// Authorization rules for the services RPCs are specified on the `RpcNameRequest`
+// messages as boolean expressions over these facts:
+// - ``HasRight(r)``: true iff the authenticated user has right `r`
+// - ``IsAuthenticatedUser(u)``: true iff ``u`` is equal to the name of the authenticated user
+//
+service UserManagementService {
+
+    // Create a new user, failing if it already exists.
+    rpc CreateUser (CreateUserRequest) returns (User);
+
+    // Get the user data of a specific user or the authenticated user.
+    rpc GetUser (GetUserRequest) returns (User);
+
+    // Delete an existing user and all its rights.
+    rpc DeleteUser (DeleteUserRequest) returns (google.protobuf.Empty);
+
+    // List the all existing users.
+    rpc ListUsers (ListUsersRequest) returns (ListUsersResponse);
+
+    // Grant a right to a user. Idempotent if the user already has the right.
+    rpc CreateUserRight (CreateUserRightRequest) returns (Right);
+
+    // Revoke a right from a user. Fails if the user does not have the right.
+    rpc DeleteUserRight (DeleteUserRightRequest) returns (google.protobuf.Empty);
+
+    // List the set of all rights granted to a user.
+    rpc ListUserRights (ListUserRightsRequest) returns (ListUserRightsResponse);
+}
+
+
+// Users and rights
+///////////////////
+
+// Users are are used to manage the rights given to authenticated Ledger API clients.
+// They are stored and managed per participant node.
+//
+// TLS client certificates authenticate the user specified by the certificates common name.
+// JWT tokens authenticate the user specified by the JWT tokens 'sub' (subject) claim.
+message User {
+    // The user identifier, which must be a non-empty string of at most 64
+    // characters matching the regexp [a-z0-9-_.], i.e. lower-case characters, digits,
+    // dashes or dots. Double-dots, double-dashes, and double-underscores are not allowed.
+    string id = 1;
+
+    // The primary party as which this user reads and acts by default on the ledger
+    // provided it has the corresponding ``CanReadAs(primary_party)`` or 
+    // ``CanActAs(primary_party)`` rights.
+    string primary_party = 2;
+}
+
+// A right granted to a user.
+message Right {
+    // The user is allowed to administrate the participant node.
+    message ParticipantAdmin {}
+
+    // The user can authorize commands for the given party.
+    message CanActAs {
+        string party = 1;
+    }
+
+    // The user can read ledger data visible to the given party.
+    message CanReadAs {
+        string party = 1;
+    }
+
+    oneof kind {
+        ParticipantAdmin participant_admin = 1;
+        CanActAs can_act_as = 2;
+        CanReadAs can_read_as = 3;
+    }
+}
+
+
+// RPC requests and responses
+/////////////////////////////
+
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message CreateUserRequest {
+    // The user to create, which is granted exactly
+    // the ``CanActAs(user.primary_party)`` right upon creation.
+    User user = 1;
+}
+
+// Required authorization: ``HasRight(ParticipantAdmin) OR IsAuthenticatedUser(user_id)``
+message GetUserRequest {
+    // The user whose data we want to retrieve.
+    // If set to empty string (the default), then the data for the authenticated user will be retrieved.
+    string user_id = 1;
+}
+
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message DeleteUserRequest {
+    string user_id = 1;
+}
+
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message ListUsersRequest {
+    // TODO: add pagination, cf. https://cloud.google.com/apis/design/design_patterns#list_pagination
+}
+
+message ListUsersResponse {
+    repeated User users = 1;
+}
+
+// Add the right to the set of rights given to the user.
+// Idempotent if the user already has the right.
+//
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message CreateUserRightRequest {
+    string user_id = 1;
+    Right right = 2;
+}
+
+// Remove the right from the set of rights granted to the user.
+// Fails if the user does not have the right.
+//
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message DeleteUserRightRequest {
+    string user_id = 1;
+    Right right = 2;
+}
+
+// Required authorization: ``HasRight(ParticipantAdmin) OR IsAuthenticatedUser(user_id)``
+message ListUserRightsRequest {
+    // The user for which to list the rights.
+    // If set to empty string (the default), then the rights for the authenticated user will be shown.
+    string user_id = 1;
+
+    // TODO: add pagination, cf. https://cloud.google.com/apis/design/design_patterns#list_pagination
+}
+
+message ListUserRightsResponse {
+    repeated Right rights = 1;
+}
+
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
@@ -12,7 +12,8 @@ option csharp_namespace = "Com.Daml.Ledger.Api.V1.Admin";
 import "google/protobuf/empty.proto";
 
 
-// Manage users and their rights for interacting with the Ledger API served by a participant node.
+// Experimental API to manage users and their rights for interacting with the Ledger API
+// served by a participant node.
 //
 // The API roughly follows the Google API style guidelines for resource-oriented design
 // (https://cloud.google.com/apis/design/resources) with the following simplifications to
@@ -43,11 +44,11 @@ service UserManagementService {
     // List the all existing users.
     rpc ListUsers (ListUsersRequest) returns (ListUsersResponse);
 
-    // Grant a right to a user. Idempotent if the user already has the right.
-    rpc CreateUserRight (CreateUserRightRequest) returns (Right);
+    // Grant rights to a user.
+    rpc GrantUserRights (GrantUserRightsRequest) returns (GrantUserRightsResponse);
 
-    // Revoke a right from a user. Fails if the user does not have the right.
-    rpc DeleteUserRight (DeleteUserRightRequest) returns (google.protobuf.Empty);
+    // Revoke rights from a user.
+    rpc RevokeUserRights (RevokeUserRightsRequest) returns (RevokeUserRightsResponse);
 
     // List the set of all rights granted to a user.
     rpc ListUserRights (ListUserRightsRequest) returns (ListUserRightsResponse);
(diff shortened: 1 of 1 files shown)


ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,140 @@
...
+
+
+// Manage users and their rights for interacting with the Ledger API served by participant node.
+//
+// The API roughly follows the Google API style guidelines for resource-oriented design
+// (https://cloud.google.com/apis/design/resources) with the following simplifications to 
+// ease its implementation:
+// 
+// 1. List methods do not (yet) support pagination, 
+//    as we expect to have fewer than 10k users and 1k rights per user.
```
da-tanabe, line 25: I largely agree with these numbers, but given https://github.com/digital-asset/daml/issues/10283 this could be a tough API to change later…

meiersi-da: My aim would be to add pagination support as per https://cloud.google.com/apis/design/design_patterns#list_pagination

This can be done in a backwards compatible way, by assuming a default limit in case none is specified.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,140 @@
...
+// 
+// 2. Resources are not named as per the Google API style guide, 
+//    as this would be in contrast with the other Ledger API services.
+//
+service UserManagementService {
```
cocreature, line 30: I think adding the required auth claims here would be quite helpful.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,140 @@
...
+
+// Remove the right from the set of rights granted to the user.
+// Fails if the user does not have the right.
+message DeleteUserRightRequest {
+    string user_id = 1;
+    Right right = 2;
+}
+
+message ListUserRightsRequest {
+    string user_id = 1;
```
cocreature, line 132: If we mandate the presence of a token here and read the user from the token we probably make it easier for client applications to actually treat the token opaquely. Otherwise we’ll repeat the pattern we have in the JSON API where the token is decoded in the client to figure out the parameters to send to the ledger which makes it increasingly hard to change the token format.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,140 @@
...
+// 2. Resources are not named as per the Google API style guide, 
+//    as this would be in contrast with the other Ledger API services.
+//
+service UserManagementService {
+
+    // Create a new user, failing if it already exists.
+    rpc CreateUser (CreateUserRequest) returns (User);
+
+    // Delete an existing user and all its rights.
+    rpc DeleteUser (DeleteUserRequest) returns (google.protobuf.Empty);
```
bame-da, line 36: What happens to "User Parties"? Do we orphan them?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,140 @@
...
+// 1. List methods do not (yet) support pagination, 
+//    as we expect to have fewer than 10k users and 1k rights per user.
+// 
+// 2. Resources are not named as per the Google API style guide, 
+//    as this would be in contrast with the other Ledger API services.
+//
+service UserManagementService {
+
+    // Create a new user, failing if it already exists.
+    rpc CreateUser (CreateUserRequest) returns (User);
```
bame-da, line 33: Does this also allocate a party?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
bame-da: Would this API be needed at all if we did the user management on-ledger?

bame-da (commented):
All makes sense to me. My main question is why you are designing this as extra services rather than just keeping users and rights on-ledger.

```
template User
  with
    participant : Party
    userParty : Party
    userName : Text
  where
    ensure (valid userName)
    signatory participant
    observer userParty

data Right
  = Admin
  | ReadAs Party
  | ActAs Party

template UserRight
  with
    participant : Party
    user : Party
    right : Right
  where
    signatory participant
    observer user
```

Now the API is given by the transaction, acs, and command APIs, and the authorization is dealt with in the Daml Model.

bame-da:
If we do this in Daml there's a question of causality and transactionality. Ie I could imagine two models:

1. The user rights are resolved once at the beginning of command processing. So the following sequence of events is valid
  A. Admin submits TX1: Delete User Alice
  B. Alice Submits TX2
  C. TX1 completes
  D. TX2 completes
2. The user rights are resolved as part of validation and can change during a transaction. So the above is invalid, but the following is valid, allowing for "transient" user rights:
  A. Alice submits a transaction with events
     1. Exercise delegated choice to add ReadAs right for user Alice on Party Bob
     2. Fetch contract known only to Bob
     3. Exercise delegated choice to remove ReadAs right for user Alice on Party Bob

When I talk about modelling users in Daml, I have model 1 in mind for now.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: This API does not have a concept of "User Parties". That can be added as an increment.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: No it does not.

meiersi-da:
> All makes sense to me. My main question is why you are designing this as extra services rather than just keeping users and rights on-ledger.

@bame-da: I agree that we ultimately want to end up with using Daml for managing the participant-local state as well. The main motivation for introducing this API is to manage technical risk. The API proposed here is simple and boring, and thus has a low chance of unforeseen implementation problems.

When building user management directly in Daml, we for example have to consider how to upgrade the `User` contracts in the future to add new features without breaking all clients interacting with the current version. Likewise, there is a potential bootstrapping problem for creating the first user contract that allocates the 'admin' user with admin rights. I'believe we can and should discover and solve these problems, but not on a tight deadline.

Instead, I propose that we introduce a simple gRPC API that 
1. covers the needs on which we have a tight deadline, and 
2. can be implemented both with direct DB access and also using on-ledger Daml contracts.
This way we can first deliver the API using direct DB access, and then iterate on it in a backwards-compatible way using on-ledger Daml contracts as the backing storage.

What do you think?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: added the auth claims in https://github.com/digital-asset/daml/pull/11818/commits/b9cb2c6914380cabd286320a4956bd34c5ee8062; and answered to @bame-da 's comment in https://github.com/digital-asset/daml/pull/11818#issuecomment-976702106

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: Good point. Addressed in https://github.com/digital-asset/daml/pull/11818/commits/8880bd081d46649833257196a6a8018e9862ef9b

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,170 @@
...
+    // Get the user data of a specific user or the authenticated user.
+    rpc GetUser (GetUserRequest) returns (User);
+
+    // Delete an existing user and all its rights.
+    rpc DeleteUser (DeleteUserRequest) returns (google.protobuf.Empty);
+
+    // List the all existing users.
+    rpc ListUsers (ListUsersRequest) returns (ListUsersResponse);
+
+    // Grant a right to a user. Idempotent if the user already has the right.
```
cocreature, line 50: at the risk of bikeshedding names: Maybe `GrantUserRight` would be a better name? I wouldn’t expect a create operation to be idempotent.

bame-da:
> What do you think?

Makes sense, you are almost certainly right that this is the lower-risk route.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: It still does not, but the service now tracks a primary_party per user.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: Yes, they continue existing, and can be reallocated to a user if desired.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: Good point. I've changed the API accordingly in cb72cd35191166e2fc84323ff6e101bff40f7e50

cocreature (approved):
Thank you!

The API looks very reasonable. So far, I think I’m fairly confident we can implement this for 2.0. 

My main concern is the amount of breakage we might cause for our users and our examples if we try to change e.g., Daml Script to work with users by default. But that definitely doesn’t need to block this PR and I also haven’t thought too deeply about it so far.

stefanobaghino-da:
I'll make sure to find time to review the design and this PR by the end of the day.

gerolf-da:
cc @mziolekda

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,175 @@
...
+service UserManagementService {
+
+    // Create a new user, failing if it already exists.
+    rpc CreateUser (CreateUserRequest) returns (User);
+
+    // Get the user data of a specific user or the authenticated user.
+    rpc GetUser (GetUserRequest) returns (User);
+
+    // Delete an existing user and all its rights.
+    rpc DeleteUser (DeleteUserRequest) returns (google.protobuf.Empty);
```
gerolf-da, line 42: I would recommend against using `Empty` as the response type. A specific message that doesn't have any fields is as "expensive" as `Empty`, but allows you to add fields later without breaking the API.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,175 @@
...
+// - ``HasRight(r)``: true iff the authenticated user has right `r`
+// - ``IsAuthenticatedUser(u)``: true iff ``u`` is equal to the name of the authenticated user
+//
+service UserManagementService {
+
+    // Create a new user, failing if it already exists.
+    rpc CreateUser (CreateUserRequest) returns (User);
+
+    // Get the user data of a specific user or the authenticated user.
+    rpc GetUser (GetUserRequest) returns (User);
```
gerolf-da, line 39: Is there a chance that we want to transmit other metadata than what is contained in User?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,175 @@
...
+// Add the rights to the set of rights granted to the user.
+//
+// Required authorization: ``HasRight(ParticipantAdmin)``
+message GrantUserRightsRequest {
+    string user_id = 1;
+
+    repeated Right rights = 2;
+}
+
+message GrantUserRightsResponse {
```
gerolf-da, line 144: Maybe it's useful to also include all rights (or just the set of already existing rights) in the response. If I think about the usage of this in a GUI, I'd probably want to refresh the user's rights with a definitive authoritative answer about all rights the user has.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
cocreature: So this means the expected use is that I first allocate and then create a user? Is there a reason why we don’t allow to create the primary party as part of the createuser request? Having looked at the Daml Script API, having to create the party first seems a bit odd.

stefanobaghino-da:
Sorry, couldn't review today, I'll need to do it tomorrow.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
```diff
@@ -0,0 +1,175 @@
...
+message ListUserRightsRequest {
+    // The user for which to list the rights.
+    // If set to empty string (the default), then the rights for the authenticated user will be listed.
+    string user_id = 1;
+
+    // TODO: add pagination, cf. https://cloud.google.com/apis/design/design_patterns#list_pagination
+}
+
+message ListUserRightsResponse {
+    repeated Right rights = 1;
```
cocreature, line 173: In Daml script, the natural thing to do seems to be to flip this around immediately and turn it into:
```
data Rights = Rights
  { actAs : [Party]
  , readAs : [Party]
   , admin : Bool
   }
```
I could see other clients wanting to do the same thing (the JSON API for example). Have you considered emitting the result in something closer to that format?

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: @gerolf-da : I'm following the Google API Style guidelines here: https://cloud.google.com/apis/design/standard_methods

If we really require more info to be transmitted, then we should introduce a custom method: https://cloud.google.com/apis/design/custom_methods

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: I can do that. It's against the style guide here: https://cloud.google.com/apis/design/standard_methods, but it is a light violation, and can come in handy in future evoluations.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: The reason why these rights are not included is that this won't scale to large sets of rights.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: I have tried it, and it was not easy to present that way. The flat set of rights is a simple, well-defined structure that lends itself well to extensions.

ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
meiersi-da: Addressed in https://github.com/digital-asset/daml/pull/11908

PR #11820, opened 2021-11-22
LF: Drop Archive Snapshot for LF < 1.14

remyhaemmerle-da:
CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


cocreature:
I think I’m lacking context here. Why can we drop this now? I see two points here:
1. To drop the Java libs we need to at least discuss with product. 2.0 seems like a good point to do that so happy to talk to them but let’s hold off on dropping them until that.
2. It seems like we lose the tests from `ProtoTest` that older versions can decode newer versions. Are we testing this somewhere else?

daml-lf/language/daml-lf.bzl
```diff
@@ -99,11 +86,11 @@ def _cmp(a, b):
         return -1
 
 def _cmp_lf_version(a, b):
-    if a == "1.dev" and b == "1.dev":
+    if (a == "1.dev" or a == "dev") and (b == "1.dev" or b == "dev"):
```
cocreature, line 89: this seems  weird, can we instead make the arguments we pass here consistently be `1.dev` instead of trying to accept both?

cocreature (approved):
Looks good provided we add buf  checks for forwards compatibility.

daml-lf/language/daml-lf.bzl
remyhaemmerle-da: Will do that in another PR.

PR #11828, opened 2021-11-23
[Self-service error codes] Do not return error code id and definite_answer in metadata for security sensitive errors

tudor-da:
This PR fixes the current situation where the error code id and definite_answer are returned in the gRPC metadata even for security sensitive error categories. A specific test is written for this in `ErrorCodeSpec`

Additionally, this PR enforces the test assertions (by asserting the metadata in `ErrorFactoriesSpec`) and removes some redundant information for the error contexts

- Hint: see fields renames with `_` prefix 

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/error/src/main/scala/com/daml/error/definitions/IndexErrors.scala
+++ b/ledger/error/src/main/scala/com/daml/error/definitions/IndexErrors.scala
@@ -3,10 +3,9 @@
 
 package com.daml.error.definitions
 
+import com.daml.error.ErrorCode.LoggingApiException
 import com.daml.error._
 import com.daml.error.definitions.ErrorGroups.ParticipantErrorGroup.IndexErrorGroup
-import com.daml.error.utils.ErrorDetails
-import io.grpc.StatusRuntimeException
 
 @Explanation("Errors raised by the Participant Index persistence layer.")
 object IndexErrors extends IndexErrorGroup {
@@ -19,11 +18,10 @@ object IndexErrors extends IndexErrorGroup {
         extends ErrorCode(
           id = "INDEX_DB_SQL_TRANSIENT_ERROR",
           ErrorCategory.TransientServerFailure,
-        )
-        with HasUnapply {
+        ) {
       case class Reject(throwable: Throwable)(implicit
           val loggingContext: ContextualizedErrorLogger
-      ) extends LoggingTransactionErrorImpl(
+      ) extends DbError(
             cause =
               s"Processing the request failed due to a transient database error: ${throwable.getMessage}",
             throwableO = Some(throwable),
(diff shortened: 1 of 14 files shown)


fabiotudone-da (approved):

pbatko-da (approved):

ledger/error/src/main/scala/com/daml/error/definitions/IndexErrors.scala
```diff
@@ -56,21 +53,34 @@ object IndexErrors extends IndexErrorGroup {
...
+      override val throwableO: Option[Throwable] = None,
+  )(implicit
+      code: ErrorCode,
+      loggingContext: ContextualizedErrorLogger,
+  ) extends LoggingTransactionErrorImpl(cause, throwableO) {
+    override def asGrpcErrorFromContext(implicit
+        contextualizedErrorLogger: ContextualizedErrorLogger
+    ): IndexDbException = {
+      val err = super.asGrpcErrorFromContext(contextualizedErrorLogger)
+      IndexDbException(err.getStatus, err.getTrailers)
```
pbatko-da, line 80: Out of interest, is `IndexDbException` name leaked to the user of the API? (I.e. can users tell a difference between getting `StatusRuntimeException` and `IndexDbException`?)

tudor-da: I don't think so. What gets over the wire is the status and metadata. JVM type information should be ignored I believe

fabiotudone-da: Possibly in logs but that shouldn't be an issue I think

ledger/error/src/main/scala/com/daml/error/definitions/LedgerApiErrors.scala
```diff
@@ -808,19 +819,17 @@ object LedgerApiErrors extends LedgerApiErrorGroup {
...
-          message: String,
+          override val cause: String,
           ledger_time: Instant,
           ledger_time_lower_bound: Instant,
           ledger_time_upper_bound: Instant,
       )(implicit loggingContext: ContextualizedErrorLogger)
-          extends LoggingTransactionErrorImpl(
-            cause = s"Invalid ledger time: $message"
-          )
+          extends LoggingTransactionErrorImpl(cause = cause)
```
mziolekda, line 827: Why do you need this assignment? Doesn't `LoggingTransactionErrorImpl(cause)` suffice?

mziolekda (approved):

ledger/error/src/main/scala/com/daml/error/definitions/LedgerApiErrors.scala
tudor-da: usually I add named arguments when the args list has default values for some arguments.

PR #11836, opened 2021-11-23
Add a guard when exercising by interface.

sofiafaro-da:
This fixes part of #11703, when exercising an inherited choice by
interface and you know the template id, via the command preprocessor.

It does this by inserting a "guard" in between the interface fetch and
the exercise body. The guard is a function Interface -> Bool, which
is general enough to check the template id, without complicating too
much in speedy. And can be generalized in the future to check more,
like signatories, etc.

I added the guard as an optional argument to UExerciseByInterface.
This isn't hooked up to the protobuf AST yet (or Haskell side for
that matter) -- but I'll do it in the next PR! For now you can invoke
the guarded exercise via the command preprocessor, so I can enable the
approprate engine tests. (There's still some failing fetch tests left,
but I decided to leave this for later. Fetch can be a lot simpler than
guarded choices, since you always add a fetch node. No need for fancy
continuations.)

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
```diff
@@ -504,6 +504,12 @@ object Ast {
       choice: ChoiceName,
       cidE: Expr,
       argE: Expr,
+      guardE: Option[Expr],
+      // `guardE` is an optional expression of type Interface -> Bool which is evaluated
+      // after fetching the contract but before running the exercise body. If the guard
+      // returns false, the transaction is aborted. If the guard raises an exception,
```
cocreature, line 510: maybe worth stating here that the fetch nodes are not yet inserted.

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
```diff
@@ -906,10 +906,12 @@ private[validation] object Typing {
         chName: ChoiceName,
         cid: Expr,
         arg: Expr,
+        guard: Option[Expr],
     ): Type = {
       checkExpr(cid, TContractId(TTyCon(tpl)))
       val choice = handleLookup(ctx, interface.lookupInterfaceChoice(tpl, chName))
       checkExpr(arg, choice.argBinder._2)
+      guard.foreach(guardExpr => checkExpr(guardExpr, TFun(TTyCon(tpl), TBool)))
```
cocreature, line 914: we need to make sure we cover those in typechecking tests (doesn’t have to be in this PR)

cocreature (approved):
Nice, thank you!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
```diff
@@ -166,18 +166,14 @@ class InterfacesTest
...
-      "be unable to exercise T1 (disguised as T2) by interface I1" in {
-        val command = ExerciseCommand(idT2, cid1, "C1", ValueRecord(None, ImmArray.empty))
-        run(command) shouldBe a[Left[_, _]]
-      }
-      "be unable to exercise T2 (disguised as T1) by interface I1" in {
-        val command = ExerciseCommand(idT1, cid2, "C1", ValueRecord(None, ImmArray.empty))
-        run(command) shouldBe a[Left[_, _]]
-      }
-     */
+    "be unable to exercise T1 (disguised as T2) by interface I1" in {
```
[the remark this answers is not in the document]
sofiafaro-da: You can only exercise an inherited choice by interface, so we are definitely exercising by interface. The fact that we expect T2 and get T1 instead, is what's important for the test. Maybe there's a better way to word it?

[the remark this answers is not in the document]
cocreature: I think the confusion comes from what you do at the LF level vs what you do in the command. In the command you specify the template id so I’d say from that pov you are execising by template. From the LF pov you are exercising by interface. Not sure what a better phrasing is.

daml-lf/validation/src/main/scala/com/digitalasset/daml/lf/validation/Typing.scala
sofiafaro-da: added a TODO

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
sofiafaro-da: added a TODO

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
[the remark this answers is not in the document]
sofiafaro-da: Well, the command is called `ExerciseByInterface`, so...  :-) 

Maybe we should leave this for now and fix it later if it comes up again.

remyhaemmerle-da (commented):
I would rather made the type of the guard `Option (Interface -> Unit)` instead of `Option (Interface -> Bool)`. 

1. We were thinking to do the same for precondition, (we did not because of backward compatibility)
1. We need to handle the case where the guard throw an exception (and insert transaction node). I believe it is already there. 
2. It slightly simpler.
3. I do not see any reason why failing guard should not be catchable, in this case the  `ChoiceGuardFailed` should be in a stable package like we have done ArithmeticError.

remyhaemmerle-da (commented):
After discussing with Moritz, I am now inclined in the total opposite direction: 
I will let the type as it is : `Option (Interface -> Boolean)`, however make failing guard non catchable. Indeed, we definitively do not want to insert a fetch node in case of failure, as we will not be able to replay the failing guard from the only information contains in the node. 

One big caveat, is that we have to ensure that guards do not throw exception. So we have to catch any exception thrown by a guard and globally abort the transaction.

sofiafaro-da:
> After discussing with Moritz, I am now inclined in the total opposite direction: I will let the type as it is : `Option (Interface -> Boolean)`, however make failing guard non catchable. Indeed, we definitively do not want to insert a fetch node in case of failure, as we will not be able to replay the failing guard from the only information contains in the node.
> 
> One big caveat, is that we have to ensure that guards do not throw exception. So we have to catch any exception thrown by a guard and globally abort the transaction.

Nice, ok. This PR should already have the right behavior then. I'll update the comments because they were talking about propagating exceptions.

daml-lf/language/src/main/scala/com/digitalasset/daml/lf/language/Ast.scala
sofiafaro-da: removed the TODO and updated the comment, since we're going to abort in every case

PR #11838, opened 2021-11-23
participant-state - Add an implicit logging context to the write service [kvl-1072]

nicu-da:
propagate the logging context through the write service
In a subsequent PR the context is used to log errors from deduplication period conversions and validations.

changelog_begin
participant-state - WriteService submit transaction now requires an implicit LoggingContext
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
+++ b/ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
@@ -30,8 +30,8 @@ final class TimedWriteService(delegate: WriteService, metrics: Metrics) extends 
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
   )(implicit
-      telemetryContext: TelemetryContext,
       loggingContext: LoggingContext,
+      telemetryContext: TelemetryContext,
   ): CompletionStage[SubmissionResult] =
     Timed.timedAndTrackedCompletionStage(
       metrics.daml.services.write.submitTransaction,
--- a/ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
+++ b/ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
@@ -48,7 +48,10 @@ final class TimedWriteService(delegate: WriteService, metrics: Metrics) extends 
       submissionId: Ref.SubmissionId,
       archives: List[DamlLf.Archive],
       sourceDescription: Option[String],
-  )(implicit telemetryContext: TelemetryContext): CompletionStage[SubmissionResult] =
+  )(implicit
+      loggingContext: LoggingContext,
+      telemetryContext: TelemetryContext,
+  ): CompletionStage[SubmissionResult] =
     Timed.completionStage(
       metrics.daml.services.write.uploadPackages,
       delegate.uploadPackages(submissionId, archives, sourceDescription),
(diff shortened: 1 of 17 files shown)


fabiotudone-da (approved):
Pre-approving but could you add some more context about why we're doing this?

nicu-da:
> Pre-approving but could you add some more context about why we're doing this?

Added the usage which will be introduced.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
```diff
@@ -95,13 +96,17 @@ trait WriteService
     * @param estimatedInterpretationCost Estimated cost of interpretation that may be used for
     *                                    handling submitted transactions differently.
     * @param telemetryContext            Implicit context for tracing.
+    * @param loggingContext            Implicit context for logging.
```
hubert-da, line 99: Nit: format

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
```diff
@@ -95,13 +96,17 @@ trait WriteService
...
     */
   def submitTransaction(
       submitterInfo: SubmitterInfo,
       transactionMeta: TransactionMeta,
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
-  )(implicit telemetryContext: TelemetryContext): CompletionStage[SubmissionResult]
+  )(implicit
+      telemetryContext: TelemetryContext,
+      loggingContext: LoggingContext,
```
hubert-da, line 108: @miklos-da, just to make sure, can we do it like this without deprecating the method (like we did for the telemetry context)?

hubert-da (approved):

ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
```diff
@@ -28,7 +29,10 @@ final class TimedWriteService(delegate: WriteService, metrics: Metrics) extends
       transactionMeta: TransactionMeta,
       transaction: SubmittedTransaction,
       estimatedInterpretationCost: Long,
-  )(implicit telemetryContext: TelemetryContext): CompletionStage[SubmissionResult] =
+  )(implicit
+      telemetryContext: TelemetryContext,
+      loggingContext: LoggingContext,
```
SamirTalwar-DA, line 34: Everywhere else, it goes "executionContext, loggingContext, telemetryContext". Can we stick to this order if possible?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/api/KeyValueParticipantStateWriter.scala
```diff
@@ -14,8 +15,8 @@ import com.daml.ledger.participant.state.kvutils.{Envelope, KeyValueSubmission}
 import com.daml.ledger.participant.state.v2._
 import com.daml.lf.data.{Ref, Time}
 import com.daml.lf.transaction.SubmittedTransaction
-import com.daml.logging.ContextualizedLogger
 import com.daml.logging.LoggingContext.newLoggingContextWith
```
SamirTalwar-DA, line 18: We probably want to enrich the logging context now, instead of creating a new one.

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
SamirTalwar-DA: I think I would be OK with just deleting the Scaladoc for the implicits. They don't tell us anything.

ledger/participant-state-metrics/src/main/scala/com/daml/ledger/participant/state/v2/metrics/TimedWriteService.scala
nicu-da: Sure, will change them. I do find this to be close to impossible to guarantee in the long run though.

SamirTalwar-DA (approved):

ledger/participant-state/src/main/scala/com/daml/ledger/participant/state/v2/WriteService.scala
miklos-da: Sounds good to me.

nicu-da: Will do so, even if not used. 
The reason I did not do it from the first is that we have the `submitTransaction` as part of the WriteService definition, but all of the other methods have their own trait. 
I will add the logging context in this PR, and I'm also thinking of extracting `submitTransaction` into its own trait as well, to be consistent in usages.

PR #11842, opened 2021-11-23
Refactor StandaloneApiServer factory

nmarton-da:
To enable direct IndexServer building (and enable two step ApiServer build process)
This will be needed for KV WriteService proxy: a wrapper is being build to serve deduplication time conversion, which needs access to completion service.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
@@ -10,106 +10,84 @@ import com.daml.buildinfo.BuildInfo
 import com.daml.error.ErrorCodesVersionSwitcher
 import com.daml.ledger.api.auth.interceptor.AuthorizationInterceptor
 import com.daml.ledger.api.auth.{AuthService, Authorizer}
-import com.daml.ledger.api.domain
 import com.daml.ledger.api.health.HealthChecks
 import com.daml.ledger.configuration.LedgerId
 import com.daml.ledger.participant.state.{v2 => state}
-import com.daml.ledger.resources.{Resource, ResourceContext, ResourceOwner}
+import com.daml.ledger.resources.ResourceOwner
 import com.daml.lf.data.Ref
-import com.daml.lf.data.Time.Timestamp
-import com.daml.lf.engine.{Engine, ValueEnricher}
+import com.daml.lf.engine.Engine
 import com.daml.logging.{ContextualizedLogger, LoggingContext}
 import com.daml.metrics.Metrics
 import com.daml.platform.configuration.{
   CommandConfiguration,
   PartyConfiguration,
-  ServerRole,
   SubmissionConfiguration,
 }
-import com.daml.platform.index.JdbcIndex
-import com.daml.platform.packages.InMemoryPackageStore
 import com.daml.platform.services.time.TimeProviderType
-import com.daml.platform.store.LfValueTranslationCache
 import com.daml.ports.{Port, PortFiles}
 import io.grpc.{BindableService, ServerInterceptor}
 import scalaz.{-\/, \/-}
-import java.io.File
 import java.time.Clock
 
+import com.daml.ledger.participant.state.index.v2.IndexService
 import com.daml.telemetry.TelemetryContext
 
 import scala.collection.immutable
 import scala.concurrent.ExecutionContextExecutor
 import scala.util.{Failure, Success, Try}
 
-final class StandaloneApiServer(
-    ledgerId: LedgerId,
-    config: ApiServerConfig,
-    commandConfig: CommandConfiguration,
-    partyConfig: PartyConfiguration,
-    submissionConfig: SubmissionConfiguration,
-    optWriteService: Option[state.WriteService],
-    authService: AuthService,
-    healthChecks: HealthChecks,
-    metrics: Metrics,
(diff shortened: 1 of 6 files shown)


ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
```diff
@@ -36,45 +36,118 @@ import scalaz.{-\/, \/-}
...
+      lfValueTranslationCache: LfValueTranslationCache.Cache,
+      checkOverloaded: TelemetryContext => Option[state.SubmissionResult] =
+        _ => None, // Used for Canton rate-limiting
+  )(implicit
+      actorSystem: ActorSystem,
+      materializer: Materializer,
+      loggingContext: LoggingContext,
+  ): ResourceOwner[ApiServer] = {
+    for {
+      indexService <- this.indexService(
```
rautenrieth-da, line 73:
```suggestion
      indexService <- indexService(
```

rautenrieth-da (commented):
The main point of this refactoring seems to be to allow users to start the ledger api server in two steps: first the indexer (`def indexService`), then the ledger API server (`def from`). Why is that? Do we have use cases where some of the parameters of `from` depend on the index?

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
nmarton-da: ahh, right, because left hand side lives in a nested scope already :man_facepalming:

nmarton-da:
> The main point of this refactoring seems to be to allow users to start the ledger api server in two steps: first the indexer (def indexService), then the ledger API server (def from). Why is that? Do we have use cases where some of the parameters of from depend on the index?

We will have, see PR description. The factory sequence will change
* from: writeservice=... -> StandAloneApiServer.apply(...Some(writeservice)...)
* to: writeservice=... -> indexservice=StandAloneApiServer.indexService -> wrappedwriteservice=???(...indexservice...) -> StandAloneApiServer.from(indexservice...Some(wrappedwriteservice)...)

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
```diff
@@ -145,7 +260,7 @@ final class StandaloneApiServer(
         config.tlsConfig,
         AuthorizationInterceptor(
           authService,
-          executionContext,
+          materializer.executionContext,
```
nicu-da, line 263: We should protect the akka internal execution contexts and avoid any direct usages.

nicu-da (commented):
Just one small thing from me. 
Thank you for taking care of this :)

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
```diff
@@ -36,45 +36,118 @@ import scalaz.{-\/, \/-}
...
+        otherServices = otherServices,
+        otherInterceptors = otherInterceptors,
+        engine = engine,
+        servicesExecutionContext = servicesExecutionContext,
+        checkOverloaded = checkOverloaded,
+      )
+    } yield apiServer
+  }
 
+  def indexService(
```
SamirTalwar-DA, line 102: Does this need to be public?

If it does, can we factor it into its own file?

rautenrieth-da: I think this being public is the whole point of this refactoring (see [comment](https://github.com/digital-asset/daml/pull/11842#issuecomment-977593551)).

rautenrieth-da (approved):
I'm not entirely sure about the naming in the new API (`StandaloneApiServer.indexService` + `StandaloneApiServer.from`), but the refactoring looks good.

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
SamirTalwar-DA: Then why not put it in its own object?

ledger/participant-integration-api/src/main/scala/platform/apiserver/StandaloneApiServer.scala
nmarton-da: yeah about that...(@SamirTalwar-DA don't kill me please) I not too fond of rolling a class all the time with the acquire, I think is nicer just composing the ResourceOwners.
Aaand thanks for pointing out because I was doing some digging, and it seems to me that this is really not a place for the Resoure's EC, since that is ticking in some runtime code for the AuthorizationInterceptor, so I switched it to the serviceExecutionContext. (@SamirTalwar-DA I saw you were operating in that area, can you check maybe whether my conclusion holds?)

nmarton-da:
> I'm not entirely sure about the naming in the new API (`StandaloneApiServer.indexService` + `StandaloneApiServer.from`), but the refactoring looks good.

pulled it apart, in different files how you suggested :+1:

kamil-da (approved):

nicu-da (approved):
Awesome, thank you!

SamirTalwar-DA (approved):
I like this split a lot.

nmarton-da:
Thank you, vigilant reviewers!

PR #11848, opened 2021-11-24
kvutils - For duplicate command rejections, add the submission id as metadata [KVL-1175]

nicu-da:
- include the metadata key existing_submission_id, which is populated with the value of the previously accepted command for duplicate commands

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
+++ b/ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
@@ -363,6 +363,7 @@ object KeyValueConsumption {
             recordTime,
             rejectionEntry,
             errorVersionSwitch,
+            None, // Not available for historical entries
           )(contextualizedErrorLogger(loggingContext, rejectionEntry))
         )
 
--- a/ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
+++ b/ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
@@ -445,9 +445,9 @@ class ConversionsSpec extends AnyWordSpec with Matchers with OptionValues {
                 Map.empty,
               ),
               (
-                _.setDuplicateCommand(Duplicate.newBuilder()),
+                _.setDuplicateCommand(Duplicate.newBuilder().setSubmissionId("not_used")),
                 Code.ALREADY_EXISTS,
-                Map.empty,
+                Map(),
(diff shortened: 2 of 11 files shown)


ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
```diff
@@ -363,6 +363,7 @@ object KeyValueConsumption {
             recordTime,
             rejectionEntry,
             errorVersionSwitch,
+            None, // Not available for historical entries
```
hubert-da, line 366: What about the future entries? Can we pass an `Option` here?

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
```diff
@@ -445,9 +445,9 @@ class ConversionsSpec extends AnyWordSpec with Matchers with OptionValues {
                 Map.empty,
               ),
               (
-                _.setDuplicateCommand(Duplicate.newBuilder()),
+                _.setDuplicateCommand(Duplicate.newBuilder().setSubmissionId("not_used")),
                 Code.ALREADY_EXISTS,
-                Map.empty,
+                Map(),
```
hubert-da, line 450: Nit: why?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
nicu-da: So this is not valid anymore because this flow will not have new entries. 
This is for the out of time bounds, and we no longer mix deduplication there

ledger/participant-state/kvutils/src/test/suite/scala/com/daml/ledger/participant/state/kvutils/ConversionsSpec.scala
nicu-da: Actually only because I initially tried to pass the metadata here as well. I would revert it but would not run another build just for this change

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/KeyValueConsumption.scala
hubert-da: Got it, thanks!

hubert-da (approved):

SamirTalwar-DA (approved):

PR #11850, opened 2021-11-24
All packages must be valid

rautenrieth-da:
The package management service already [validates all uploaded packages](https://github.com/digital-asset/daml/blob/main/ledger/participant-integration-api/src/main/scala/platform/apiserver/services/admin/ApiPackageManagementService.scala#L101), so let's add it to the `WriteService` contract.


andreaslochbihler-da (approved):
Thanks for making this explicit.

PR #11862, opened 2021-11-24
Error when fetching the wrong template id (via fetch by interface).

sofiafaro-da:
When doing a "fetch by interface" command with a known template id,
error out with a WronglyTypedContract if the fetched contract has
a different template id. This doesn't affect daml, only affects
replays, so it's rather minor. I also enabled the engine test that
caught this.

Part of #11703, follow up to #11836.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
@@ -274,18 +274,14 @@ class InterfacesTest
       val command = FetchByInterfaceCommand(idI2, idT2, cid1)
       run(command) shouldBe a[Left[_, _]]
     }
-    // TODO https://github.com/digital-asset/daml/issues/11703
-    //   Enable these tests.
-    /*
-      "be unable to fetch T1 (disguised as T2) via interface I1" in {
-        val command = FetchByInterfaceCommand(idI1, idT2, cid1)
-        run(command) shouldBe a[Left[_, _]]
-      }
-      "be unable to fetch T2 (disguised as T1) via interface I1" in {
-        val command = FetchByInterfaceCommand(idI1, idT1, cid2)
-        run(command) shouldBe a[Left[_, _]]
-      }
-     */
+    "be unable to fetch T1 (disguised as T2) via interface I1" in {
+      val command = FetchByInterfaceCommand(idI1, idT2, cid1)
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Thank you!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
```diff
@@ -274,18 +274,14 @@ class InterfacesTest
...
-        run(command) shouldBe a[Left[_, _]]
-      }
-      "be unable to fetch T2 (disguised as T1) via interface I1" in {
-        val command = FetchByInterfaceCommand(idI1, idT1, cid2)
-        run(command) shouldBe a[Left[_, _]]
-      }
-     */
+    "be unable to fetch T1 (disguised as T2) via interface I1" in {
+      val command = FetchByInterfaceCommand(idI1, idT2, cid1)
+      run(command) shouldBe a[Left[_, _]]
```
remyhaemmerle-da, line 279: I would like that we check the test is setup correctly.
I personally wrote so many buggy test, and sometime broke others without realized it.
Ideally we would first check the fetch with T1 success, checking the 2 template Ids in the error should be OK as well. 

```suggestion
      run(FetchByInterfaceCommand(idI1, idT1, cid1) shouldBe a[Right[_, _]]
      val command = FetchByInterfaceCommand(idI1, idT2, cid1)
      run(command) shouldBe a[Left[_, _]]
```

remyhaemmerle-da (approved):
Nice. Thanks.

> This doesn't affect daml, only affects replays, so it's rather minor. I also enabled the engine test that caught this.

This does not really matter as interface are not frozen.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
sofiafaro-da: Fetch with T1 succeeds, it is already tested above.

remyhaemmerle-da: My bad. Thanks.

sofiafaro-da: I definitely agree with strengthening the check on the test output, so I did that.

PR #11863, opened 2021-11-24
DPP-723 Add global parallelism limit for ACS queries

rautenrieth-da:
This PR limits the number of ACS contract id fetching queries running in parallel (across all ACS streams).

# Performance impact

I have measured the performance impact using the following method:
- Ingest 1M contracts, where 50% contracts are of template `Foo1` and 50% of those contracts are immediately archived
- Start 16 concurrent ACS streams for `Foo1`
- Measure average items/sec per stream

The results don't show any improvement of limiting the parallelism. Variance between different runs with the same settings is quite high (up to 20% difference).

```
Parallelism   Items/second
1             1538,68
2             2683,97
4             4169,93
8             4323,48
16            4277,05
32            4062,43
64            4238,05
```

The same for 64 concurrent streams:
```
2    608,61
4    944,82
8    1047,07
1024 968,88
```


ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,69 @@
...
+    while (running < parallelism && waiting.nonEmpty) {
+      val head = waiting.dequeue()
+      startTask(
+        head.task().andThen { case result => head.promise.tryComplete(result) }(executionContext)
+      )
+    }
+  }
+}
+
+class ThreadPoolBasedConcurrencyLimiter[T](
```
nmarton-da, line 58: Remove?

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,69 @@
...
+  override def execute(task: () => Future[T]): Future[T] = task()
+}
+
+class QueueBasedConcurrencyLimiter[T](
+    parallelism: Int,
+    executionContext: ExecutionContext,
+) extends ConcurrencyLimiter[T] {
+  assert(parallelism > 0)
+
+  case class WaitingTask(
```
nmarton-da, line 25: You do not really need so much data to track. We could as good as do line 44 and line 52 as we compose the future generating function at line 34. You do not need to have a future generating function neither, just something to create the future and being done with it. So the queue-s signature can be () => Unit.

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,69 @@
...
+      task: () => Future[T],
+      promise: Promise[T] = Promise[T](),
+  )
+  private val waiting = mutable.Queue[WaitingTask]()
+  private var running: Int = 0
+
+  override def execute(task: () => Future[T]): Future[T] = synchronized {
```
nmarton-da, line 32: If we do not care about T in the queue, then we could easily move the generic parameter T from the class to the method. This makes it more generic, and limiting would not relate to what we are limiting.

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,69 @@
...
+    if (running >= parallelism) {
+      val waitingTask = WaitingTask(task)
+      waiting.enqueue(waitingTask)
+      waitingTask.promise.future
+    } else {
+      startTask(task())
```
nmarton-da, line 38: I would consider not doing this optimization with the enqueue/start, and just having one path with enqueueing, which would make the code lighter

nmarton-da (commented):
Looks good!
As discussed offline I would not go with a complete threadpool for this purpose, the performance difference is minimal.
Nice test suite.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/JdbcLedgerDao.scala
```diff
@@ -670,6 +671,8 @@ private class JdbcLedgerDao(
         acsFetchingparallelism = acsContractFetchingParallelism,
         metrics = metrics,
         materializer = materializer,
+        // TODO: properly parametrize the parallelism. Currently using 8, because the default connection pool size is 16.
```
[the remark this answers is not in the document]
rautenrieth-da: Added a `--acs-global-parallelism-limit` CLI option for this, defaulting to 10.

ledger/participant-integration-api/src/main/scala/platform/store/appendonlydao/events/ACSReader.scala
```diff
@@ -69,15 +71,18 @@ class FilterTableACSReader(
...
-        dispatcher
-          .executeSql(metrics.daml.index.db.getActiveContractIds)(
-            eventStorageBackend.activeContractEventIds(
-              partyFilter = query.filter.party,
-              templateIdFilter = query.filter.templateId,
-              startExclusive = query.fromExclusiveEventSeqId,
-              endInclusive = activeAt._2,
-              limit = idPageSize,
-            )
+        idQuerylimiter
```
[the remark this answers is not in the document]
rautenrieth-da: Moved the limiting to `eventStorageBackend.activeContractEventBatch`

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
rautenrieth-da: Refactored the code to make the class non-parametric.

kamil-da (approved):
LGTM!

nmarton-da:
> This PR limits the number of ACS contract id fetching queries running in parallel (across all ACS streams).
> 
> I have tried to measure the performance impact using the following method:
> 
>     * Ingest 1M contracts, where 50% contracts are of template `Foo1` and 50% of those contracts are immediately archived
> 
>     * Start 16 concurrent ACS streams for `Foo1`
> 
>     * Measure average items/sec per stream
> 
> 
> However, the differences between different parallelism limits were the same order of magnitude as the variance between runs with the same settings.

does these still apply after fixing the configuration flag?

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,61 @@
...
+    result
+  }
+
+  private def enqueueTask[T](task: () => Future[T]): Future[T] = synchronized {
+    val promise = Promise[T]()
+
+    // Note: tasks are started sequentially (see startTasks), but they can finish concurrently.
+    // Only the block withing `andThen` needs to be synchronized.
+    val waitingTask = () => {
+      running = running + 1
```
nmarton-da, line 39: would you consider moving this to startTasks? Looks so dangerous this state modification in a lambda without synchronized ;)

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
```diff
@@ -0,0 +1,61 @@
...
+    parallelism: Int,
+    executionContext: ExecutionContext,
+) extends ConcurrencyLimiter {
+  assert(parallelism > 0)
+
+  type Task = () => Unit
+  private val waiting = mutable.Queue[Task]()
+  private var running: Int = 0
+
+  override def execute[T](task: () => Future[T]): Future[T] = synchronized {
```
nmarton-da, line 27: enqueueTask with a startTask() as the before last statement would give this method. maybe just inline?

ledger/participant-integration-api/src/main/scala/platform/store/utils/ConcurrencyLimiter.scala
nmarton-da: (also would be more natural to make the incrementing next to the while)

nmarton-da (approved):
LGTM!
Left some minor comments on the logic.
Would be nice to see how this behaves, and fine tuning that magic 10 number, but that can be a different PR cc @kamil-da

nmarton-da:
> This PR limits the number of ACS contract id fetching queries running in parallel (across all ACS streams).
> # Performance impact
> 
> I have measured the performance impact using the following method:
> 
>     * Ingest 1M contracts, where 50% contracts are of template `Foo1` and 50% of those contracts are immediately archived
> 
>     * Start 16 concurrent ACS streams for `Foo1`
> 
>     * Measure average items/sec per stream
> 
> 
> The results don't show any improvement of limiting the parallelism. Variance between different runs with the same settings is quite high (up to 20% difference).
> 
> ```
> Parallelism   Items/second
> 1             1538,68
> 2             2683,97
> 4             4169,93
> 8             4323,48
> 16            4277,05
> 32            4062,43
> 64            4238,05
> ```
> 
> The same for 64 concurrent streams:
> 
> ```
> 2    608,61
> 4    944,82
> 8    1047,07
> 1024 968,88
> ```

Thanks for testing, it is clearly visible it works :+1: 
I wouldn't worry too much about the impact as you go higher: it seem to me from your experiments, that something else is the bottle-neck in those cases (probably CPU? 1000 * 64 = 4000 * 16). Also test scenario gives ideal situation for postgresql to cache, and this limiting is primarily for limiting impact on disk.
Feel free to merge, LGTM

PR #11866, opened 2021-11-24
remove mergify

garyverhaegen-da:
It's been working pretty well for us, but our usage is now completely
covered by GitHub's auto-merge feature, which has the advantage of not
requiring us to duplicate our branch protection rules.

This will also reduce our security audit surface as that's one fewer
provider to scan (not that we've found anything wrong with them so far,
but still, less work in the future).

CHANGELOG_BEGIN
CHANGELOG_END


garyverhaegen-da:
Once this is approved I'll also take care of disabling the integration in GitHub settings.

stefanobaghino-da (approved):

cocreature (approved):

Issue #11830, opened 2021-11-23
Issues for `Env` management in `ClosureConversion`


nickchapman-da:
Issues for `Env` management in `ClosureConversion:
- (A) Quadratic complexity for `Env.shift`
- (B) Quadratic complexity of `freeVars` calculation in nested sub-expressions.
- (C) `freeVars` computation is not stack safe.

#### `Env.shift` discussion
The quadratic complexity is caused by the work of `shift` being proportional to the env-size, each time it is extended with `n` new bindings. There are two aspects to this work:
- (1) The map keys are _shifted_ by `n`. 
- (2) Map values which are stack-locations (`SELocS`) are _shifted_ by `n`.

(1) is because the map is keyed by the number of binders a variable is found under, and that is because `source.SEVar` is a deBruijn index. This shift can be avoided by keying the map with an _absolute_ quantity - the static depth of the variable binding. We will then track the depth so that `SEVar`s can be converted from DB-index to the absolute-index.

(2) is because the `SELocS` in the `target.SExpr` is a stack-relative index. Meaning that is relative to the top of the stack.
There is no good reason for `SExpr1` to use relative-stack indexing. In fact it is very awkward!. The `Anf` pass in effect reconstructs what the _absolute_ depth would be, so that it has a stable map key! The solution here is to use absolute stack indexes in `SExpr1`. This will simplify the code in both the producer `ClosureConversion` and the consumer `Anf`.

nickchapman-da:
#11877 fixes Aspect (1) of `EnvShift` issue
#11912 fixes Aspect (2) of `EnvShift` issue

nickchapman-da:
#11946 solves issue (C)

nickchapman-da:
We decided that Issue B needn't be solved, because:
- The very _essence_ of the closure conversion transformation causes a runtime quadratic complexity, in terms of the number of syntactically nested abstractions, which is deemed acceptable, because N of 10 would be considered large, Hence, the fact that we have a quadratic compile time complexity is rather unimportant.
- We have stack-safe testcases for nested abstractions of depth 10,000, which despite the quadratic time complexity still only take on the order of 1 second to process.

PR #11877, opened 2021-11-25
Use absolute stack locations in SExpr1

nickchapman-da:
Use _absolute_ stack locations in `SExpr1`
- More convenient for the consuming phase `Anf`
- More convenient for the producing phase `ClosureConversion`

But the real payoff is that it allows us to remove the call to `shiftLoc` in `ClosureConversion` which is the major blocker to avoiding the quadratic time complexity in the `Env` management. See:  #11830 for details on the complexity issue. The change here relates to issue _(2)_ described in the issue.

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Anf.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Anf.scala
@@ -162,8 +162,13 @@ private[lf] object Anf {
     */
   private[this] case class AbsBinding(abs: DepthA)
 
-  private[this] def makeAbsoluteB(env: Env, rel: Int): AbsBinding = {
+  private[this] def makeAbsoluteB(env: Env, rel: Int, abs: Int): AbsBinding = {
     val oldAbs = env.oldDepth.incr(-rel)
+    if (DepthE(abs) != oldAbs) {
+      sys.error(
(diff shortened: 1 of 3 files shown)


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Anf.scala
```diff
@@ -178,7 +176,7 @@ private[lf] object Anf {
 
   private def convertLoc(x: source.SELoc): target.SELoc = {
     x match {
-      case source.SELocS(x) => target.SELocS(x)
+      case source.SELocAbsoluteS(_) => sys.error("Anf.convertLoc/SELocAbsoluteS, unreachable code")
```
cocreature, line 179: should we use different types for the locations in different phases to avoid this?

cocreature (approved):
Nice, thank you!

remyhaemmerle-da (approved):
Nice

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/Anf.scala
nickchapman-da: No, this is just an impossible code path. It was impossible before this change.
(_impossible_ in the sense than an earlier code branch has already dealt with this variant)

I improved the `Anf` code here (#11901) to avoid this impossible branch.

PR #11880, opened 2021-11-25
Update protobuf docs plugin

### LINKED ISSUE #11761: oneof types are not properly rendered with protoc-gen-doc
The current version of protoc-gen-doc we're using (`1.1`) doesn't support rendering `oneof` types. This was fixed in protoc-gen-doc `1.4` or later.

The `oneof` in [commands.proto](https://github.com/digital-asset/daml/blob/bd2a6852eedca32d20ef4518c1727a4ca7946e22/ledger-api/grpc-definitions/com/daml/ledger/api/v1/commands.proto#L60-L90) is currently rendered like this: 
image

aherrmann-da:
* Updates `protoc-gen-doc` - installing it via `rules_go` and Gazelle's support for Go modules
* Updates Gazelle and googleapis
  * Patches googleapis to expose `status.proto` needed for gRPC Haskell bindings
  * Uses a more recent version of Gazelle than the latest release to fix an issue with `embedsrcs` on Windows
* Documents how to add Go dependencies

To address https://github.com/digital-asset/daml/issues/11761 we'll also need to update the documentation templates.

### Pull Request Checklist

trigger the build.


compatibility/deps.bzl
```diff
@@ -30,10 +30,14 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file"
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load(
     "@daml//:deps.bzl",
+    "bazel_gazelle_sha256",
```
cocreature, line 33: why do we need this in the compat workspace now?

cocreature (approved):
Nice work, thanks a lot for fixing this!

compatibility/deps.bzl
aherrmann-da: The main `BUILD` file of `@daml` pulls in Gazelle, and the compatibility workspace depends on some targets defined in the top-level package, so Bazel reads that build file and fails with file not found if we don't pull it in.

PR #11884, opened 2021-11-25
Add ContractDoesNotImplementInterface error.

sofiafaro-da:
For when fetching or exercising by interface generically, and the fetched contract doesn't implement the interface. Right now this takes precedence over type errors, but it shouldn't.

Part of #10810 and #11703.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
+++ b/daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
@@ -17,6 +17,7 @@ import com.daml.lf.command._
 import com.daml.lf.transaction.test.TransactionBuilder.assertAsVersionedContract
 import org.scalatest.prop.TableDrivenPropertyChecks
 import org.scalatest.EitherValues
+import org.scalatest.Inside._
 import org.scalatest.wordspec.AnyWordSpec
 import org.scalatest.matchers.should.Matchers
 import interpretation.{Error => IE}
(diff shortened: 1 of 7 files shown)


daml-lf/transaction/src/main/scala/com/daml/lf/Error.scala
```diff
@@ -89,6 +89,15 @@ object Error {
       actual: TypeConName,
   ) extends Error
 
+  /** We tried to fetch / exercise a contract by interface, but
+    * the contract does not implement this interface.
+    */
+  final case class ContractDoesntImplementInterface(
```
cocreature, line 95: bikeshed: Should it be `ContractDoesNotImplementInterface`?

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
```diff
@@ -255,7 +281,16 @@ class InterfacesTest
     "be unable to exercise T1 (disguised as T2) by interface I2 via 'exercise by interface'" in {
       val command =
         ExerciseByInterfaceCommand(idI2, idT2, cid1, "C2", ValueRecord(None, ImmArray.empty))
-      run(command) shouldBe a[Left[_, _]]
+      run(command) match {
+        case Left(Error.Interpretation(err, _)) =>
```
cocreature, line 285: you might want to use `inside` to make the assertion a bit nicer.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
```diff
@@ -283,7 +318,16 @@ class InterfacesTest
     }
     "be unable to fetch T1 (disguised as T2) via interface I2" in {
       val command = FetchByInterfaceCommand(idI2, idT2, cid1)
-      run(command) shouldBe a[Left[_, _]]
+      run(command) match {
```
cocreature, line 321: same here.

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
cocreature: and for most of the other cases as well

cocreature (approved):
Thanks!

remyhaemmerle-da (approved):

daml-lf/transaction/src/main/scala/com/daml/lf/Error.scala
sofiafaro-da: done!

daml-lf/engine/src/test/scala/com/digitalasset/daml/lf/engine/InterfacesTest.scala
sofiafaro-da: thanks for the suggestion!

PR #11886, opened 2021-11-25
Change daml script’s sleep to sleep for a minimum amount of time

cocreature:
We’ve seen a few flaky test failures where we slept for less than the
expected amount of time which isn’t what we want. We definitely cannot
guarantee an exact sleep time but at least a minimum.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ScriptF.scala
+++ b/daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ScriptF.scala
@@ -394,10 +394,22 @@ object ScriptF {
         mat: Materializer,
         esf: ExecutionSequencerFactory,
     ): Future[SExpr] = Future {
-      val sleepMillis = micros / 1000
-      val sleepNanos = (micros % 1000) * 1000
-      Thread.sleep(sleepMillis, sleepNanos.toInt)
+      sleepAtLeast(micros * 1000)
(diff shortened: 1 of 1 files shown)


daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ScriptF.scala
```diff
@@ -394,11 +394,23 @@ object ScriptF {
...
+
+    private def sleepAtLeast(totalNanos: Long) = {
+      // Thread.sleep can wake up earlier so we loop it to guarantee a minimum
+      // sleep time
+      val t0 = System.nanoTime
+      var nanosLeft = totalNanos
+      while (nanosLeft > 0) {
+        val milis = nanosLeft / 1000000
+        val nanos = (nanosLeft % 1000000).toInt
+        Thread.sleep(milis, nanos);
```
remyhaemmerle-da, line 409:
```suggestion
        java.util.concurrent.TimeUnit.NANOSECONDS.sleep(nanosLeft)
```

remyhaemmerle-da (approved):

daml-script/runner/src/main/scala/com/digitalasset/daml/lf/engine/script/ScriptF.scala
cocreature: oh nice! didn’t know about that

PR #11888, opened 2021-11-25
LF check stable proto with buf and md5sum.

### LINKED ISSUE #11443: daml_lf_ProtoTest.scala times out
I’ve seen this test time out twice on CI now suggesting that we should do something about it.

Under load you can get it up to 30s or so locally, so it’s not surprising that occasionally it takes longer than 60s on CI.

I’m fine with just bumping the timeout given that this test is often cached reasonably well and this seems rare.

Alternatively we could also look into speeding it up or splitting it up.

remyhaemmerle-da:
The resulting tests should be faster and more reliable

fixes #11443

CHANGELOG_BEGIN
CHANGELOG_END

### Pull Request Checklist

trigger the build.


remyhaemmerle-da:
I did not write test for 1.6, 1.7, and 1.8 as all versions from 1.6 to 1.13 will be dropped by #11820.

WORKSPACE
```diff
@@ -908,3 +908,24 @@ dev_env_tool(
...
+nixpkgs_package(
+    name = "buf",
+    attribute_path = "buf",
+    fail_not_supported = False,
+    nix_file = "//nix:bazel.nix",
+    nix_file_deps = common_nix_file_deps,
+    repositories = dev_env_nix_repos,
+)
+
+dev_env_tool(
```
cocreature, line 921: this doesn’t actually work does it? we don’t get buf on Windows. I’d just drop this and work with the `nixpkgs_package` directly.

cocreature (approved):
Looks sensible minus the `dev_env_tool` stuff, thanks!

PR #11896, opened 2021-11-26
Draft: User management service PoC implementation

### LINKED ISSUE #12014: ledger api server: build PoC of user management service
Scope:
1. maintain the mapping from users to their primary party and their access rights as part of the state of a participant node
     1. expose the mapping as this gRPC a user management service in the Ledger API
     2. bootstrap the mapping by inserting a `participant_admin` user with participant admin rights when creating a fresh participant node
     3. for PoC: maintain the mapping as an in-memory store
2. change JWT authorization to authenticate a user instead of a set of claims; concretely
    1. change JWT authorization to also support standard JWT tokens as follows
       1. the JWT ‘sub’ claim specifies the authenticated user and is required
       2. the AuthorizationInterceptor computes the ClaimSet before request execution as 
        1. claims = claims retrieved from storage for the for authenticated user
        2. ledgerId = None, as the ledger-id concept is dropped in Daml 2.0
        4. participantId = JWT ‘aud’ claim, if present
        5. applicationId = JWT ‘sub’ claim
        6. expiration = JWT ‘exp’ claim, if present
    2. existing JWT tokens are parsed as they are now, and work as they do now
      1. their application-id is considered the authenticated user
      2. no lookup of rights for that user is made
3. include tests for all of the above to ensure client tooling can rely on the API

meiersi-da:
Fixes #12014 

This PR aims to bring together all changes such that user management is implemented in all sandboxes with an in-memory backend and default JWT tokens are supported alongside the JWT tokens with the custom Daml claims.

Note that this service is not hidden behind a feature flag, as we expect to land it for the next release; and in the worst case can rip it out quite cleanly by deleting the gRPC definitions and following the thread from there.

### TODO's before merge

Functionality:
- [x] listing users
- [x] use self-service errors to simplify debugging user management failures
- [x] hook-up standard JWT tokens to auth service
- [x] authorization for UserManagement API

Testing work moved to separate tickets:
- #12057
- #12059
- #12062

Preparation for review:
- [x] create proper type for standard JWT tokens
- [x] handle FIXME's and TODO's and/or convert them to issues
- [x] split change into a logical sequence of commits

### Pull Request Checklist

trigger the build.


ledger/ledger-api-domain/src/main/scala/com/digitalasset/ledger/api/UserManagement.scala
```diff
@@ -0,0 +1,21 @@
...
+
+package com.daml.ledger.api
+
+import com.daml.lf.data.Ref
+
+object UserManagement {
+
+  case class User(
+      id: String,
+      primaryParty: Option[Ref.Party],
```
cocreature, line 12: do you expect to keep this optional or is this just because you haven’t solved bootstrapping so far? My understanding was that the primary party is required.

meiersi-da: I believe for an `admin` user there might be no primary party. At least on VMBC, it doesn't yet make sense as there is no participant party, yet. Likewise, for a user that serves to export the whole ledger, a primary party might not make sense.

So, I'm likely going to update the commentary on the gRPC services to allow for not setting the primary party.

FYI: @bame-da

cocreature: In that setting, do you think tools like Daml Script, triggers, … that largely expect to act as a party can fail if you pass a user with no-primary party?

meiersi-da: I'd consider the `primary_party` a convenience mechanism that is not required for these tools to function, but makes them easier to use for the default case where a user has an on-ledger role.

For special users like one with `CanActAsAnyParty` rights (see https://github.com/digital-asset/daml/pull/11908), I'd expect that these tool should still be usable, but might require extra parameters or using lower-level functions; e.g., `submitMulti` instead of `submitUser` in Daml script.

I'm being slightly vague here, as I don't know the tools well enough yet. I defer to you and @bame-da to make the call.

ledger/participant-integration-api/src/main/scala/platform/index/InMemoryUserManagementService.scala
```diff
@@ -0,0 +1,99 @@
...
+      case None => Left(UserNotFound(id))
+    }
+  }
+
+  // Underlying mutable map to keep track of UserInfo state.
+  // Structured so we can use a ConcurrentHashMap (to more closely mimic a real implementation, where performance is key).
+  // We synchronize on a private object (the mutable map), not the service (which could cause deadlocks).
+  // (No need to mark state as volatile -- rely on synchronized to establish the JMM's happens-before relation.)
+  private val state: mutable.Map[String, UserInfo] = mutable.Map(AdminUser.toStateEntry)
+  private def lookup(id: String) = state.synchronized { state.get(id) }
```
meiersi-da, line 73: @adriaanm-da : if you use a `ConcurrentHashMap` why do you need the `synchronized`? That means all reads and writes get sequentially ordered doesn't it?

adriaanm-da: Yes, if we switched to a concurrent map, the `synchronized` usages should be removed (and the `val` storing the concurrent map should be final so that it's published properly). I meant the logic was structured so that switching to a lock-free implementation like CHM would be straightforward (using operations like `putIfAbsent` and `replace`). I don't know if the additional complexity is warranted, though -- I'd start with a simple solution like this, using `synchronize` on a mutable map, and see if it scales well enough.

language-support/java/bindings-rxjava/src/test/scala/com/daml/ledger/rxjava/grpc/helpers/LedgerServices.scala
```diff
@@ -43,7 +43,7 @@ final class LedgerServices(val ledgerId: String) {
 
   import LedgerServices._
 
-  val executionContext: ExecutionContext = global
+  val executionContext: ExecutionContext = global // TODO: is global the right execution context? Usually it's passed implicitly -- why is this val not implicit?
```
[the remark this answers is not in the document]
gerolf-da: Implicitly passing values around sounds nice, until you run into the situation where you're trying to figure out where the value comes from.
IMHO, implicit parameters are acceptable for typeclass-"lookup", but anything else should really be passed explicitly. It's more verbose, but also much clearer what's going on.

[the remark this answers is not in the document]
adriaanm-da: Agreed, implicits can be overused. I do think they can make sense beyond the "type class" use case.  I'd agree without hesitation they should be explicit when you're juggling multiple execution contexts (which seems rare?). Here, I added that comment because the code is using the `global` execution context (and only that one), and this really seems like a "background" decision.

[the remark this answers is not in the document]
meiersi-da: Thanks for the input. I've removed the implicit passing of the `ExecutionContext` in this PR.

meiersi-da:
Closing in favor of https://github.com/digital-asset/daml/pull/12063

PR #11902, opened 2021-11-26
self-service compat: set branch name to not main

garyverhaegen-da:
The "tell slack about failures" step is conditional on the branch name being main; for whatever reason if we don't supply a branch name Azure sets it to `main`, regardless of the fact that the commit is not in that branch at all.

CHANGELOG_BEGIN
CHANGELOG_END


cocreature (approved):
fun

garyverhaegen-da:
Screenshot 2021-11-26 at 17 20
Good enough for me.

PR #11903, opened 2021-11-27
wip

nmarton-da:
changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -69,6 +69,7 @@ start_postgresql
 
 # Run the tests.
 bazel test //... \
+  --cache_test_results=no \
   --build_tag_filters "$tag_filter" \
   --test_tag_filters "$tag_filter" \
   --test_env "POSTGRESQL_HOST=${POSTGRESQL_HOST}" \
--- a/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
+++ b/ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
@@ -43,10 +43,19 @@ object DbType {
         supportsAsynchronousCommits = false,
       )
 
+  object M
+      extends DbType(
+        "M",
+        "N/A",
+        supportsParallelWrites = false,
+        supportsAsynchronousCommits = false,
+      )
+
   def jdbcType(jdbcUrl: String): DbType = jdbcUrl match {
-    case h2 if h2.startsWith("jdbc:h2:") => H2Database
+    case h2 if h2.startsWith("jdbc:h2:") => M
     case pg if pg.startsWith("jdbc:postgresql:") => Postgres
     case oracle if oracle.startsWith("jdbc:oracle:") => Oracle
+    case m if m.startsWith("jdbc:m:") => M
     case _ =>
       sys.error(s"JDBC URL doesn't match any supported databases (h2, pg, oracle)")
   }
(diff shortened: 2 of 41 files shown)


build.sh
```diff
@@ -69,6 +69,7 @@ start_postgresql
 
 # Run the tests.
 bazel test //... \
+  --cache_test_results=no \
```
mziolekda, line 72: This needs to be changed back before it is unwipped

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsPruning.scala
```diff
@@ -88,7 +88,7 @@ private[backend] trait StorageBackendTestsPruning extends Matchers with StorageB
       contractId = "#1",
       signatory = someParty,
     )
-    val createFilter1 = DbDto.CreateFilter(1L, someTemplateId.toString, "signatory")
+    val createFilter1 = DbDto.CreateFilter(1L, someTemplateId.toString, "party")
```
meiersi-da, line 91: Why is this rename required?

meiersi-da (commented):
Impressive changes @nmarton-da . Looks like a good direction.

Makes me wonder whether we couldn't implement in-memory fan-out more simply using an aggressively pruned in-memory storage backend for the events that wraps a DB storage backend. For calls that fall within the event-range of the in-memory backend, we'd serve them from memory; the other ones get served from disk.

build.sh
nmarton-da: I think we will have other PRs to roll this out, but if no, for sure :)
I added this to trigger full test builds for measurement purposes.

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsPruning.scala
nmarton-da: This implementation has no filter table, and is getting the filtering info directly from events. So if the test needed to be changed to align the fixtures: same information in create-filters and create-events.
There is a TODO though to add improvement of filter tables (that would be nested maps as indexes in the in memory implementation), because that would be more performant for the in memory as well, and also it would be more conformant as well. (not sure it is needed though)

nmarton-da:
> Impressive changes @nmarton-da . Looks like a good direction.
> 
> Makes me wonder whether we couldn't implement in-memory fan-out more simply using an aggressively pruned in-memory storage backend for the events that wraps a DB storage backend. For calls that fall within the event-range of the in-memory backend, we'd serve them from memory; the other ones get served from disk.

Interesting idea!

ledger/participant-integration-api/src/test/lib/scala/platform/store/backend/StorageBackendTestsPruning.scala
meiersi-da: Ah OK. Thanks.

ledger/participant-integration-api/src/main/scala/platform/store/DbType.scala
```diff
@@ -43,10 +43,19 @@ object DbType {
         supportsAsynchronousCommits = false,
       )
 
+  object M
```
mziolekda, line 46: M as in Marton or as in M-theory?

```
In the absence of an understanding of the true meaning and structure of M-theory, Witten has suggested that the M should stand for "magic", "mystery", or "membrane" according to taste, and the true meaning of the title should be decided when a more fundamental formulation of the theory is known.
```

nmarton-da:
superseded by associated project

Issue #11371, opened 2021-10-25
Respect upper version bound for version check of stable packages in Daml Engine


cocreature:
A daml engine with an upper version bound of LF X should not allow stable packages with version Y > X. We only want to ignore the lower bound.

PR #11905, opened 2021-11-29
Limit supported input versions in damlc to >= LF 1.8

cocreature:
1.8 was the version that introduced type synonyms, we really don’t
gain much by dropping more since data-dependencies mainly depends on
that. and this provides for a very natural upgrade path for users
where pretty much everyone should be able to upgrade directly to SDK
2.0 without having to go through intermediate versions.

changelog_begin

- [Daml Compiler] The supported input LF versions for
  data-dependencies are now limited to LF 1.8 and newer.

changelog_end

### Pull Request Checklist

trigger the build.


cocreature:
> Nice code generation :) I'm tempted to say we should do something similar for the list of stable packages we have in scala (for a similar version check).

we definitely should but I’d like to tackle https://github.com/digital-asset/daml/issues/11371 as part of that.

language-support/ts/codegen/tests/src/DA/Test/Daml2js.hs
```diff
@@ -14,31 +14,21 @@ import qualified DA.Daml.LF.Ast.Version as LF
 import DA.Test.Daml2jsUtils
 import Data.List.Extra
 import qualified Data.Text.Extended as T
-import qualified Data.ByteString.Lazy as BSL
```
cocreature, line 17: @stefanobaghino-da can you please take a look if that looks reasonable to you? Factored it into a separate commit for easier review 4ec55da95bcf47e844d3b709be090d6d78e49e8a. I just dropped everything that depends on LF < 1.8.

robin-da (approved):
Looks good, thanks!

language-support/ts/codegen/tests/src/DA/Test/Daml2js.hs
stefanobaghino-da: I tried to go through it but my Haskell and knowledge of `daml2js` is insufficient to give you a meaningful answer. Who owned this previously? All I can say is that I would trust whomever actively maintained this previously, as to the best of my understanding there hasn't been any handover for this. If the application runtime need to own it, we'll probably need to rewrite it at some point as I don't think we should spread ourselves too thin into maintaining more and more stuff far away from the core skills of the team.

cocreature: Martin & Robin wrote it originally. since then I did some of the work and Gary made some changes to support genmaps.

PR #11906, opened 2021-11-29
Split daml-lf encode/decode Haskell libraries

cocreature:
I was working on limiting the supported input versions in damlc. This
requires the list of stable package ids so we can allow those to be
decoded even if they are older.

However that leads to a cyclical dependency:
To get the package ids of the stable packages, I need to encode
them. The version check is in the decoder. If those live in the same
package, I now end up with a cyclical dependency and
Bazel (understandably) gets very sad.

This PR splits decoding/encoding into two libraries with the
generation of stable packages only depending on encoding.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


sofiafaro-da (approved):
Nice, thank you!

robin-da (approved):
Looks good, thanks!

PR #11908, opened 2021-11-29
Add additional rights and clarifications to user management API

meiersi-da:
Addresses: learning from design doc and review comments from https://github.com/digital-asset/daml/pull/11818/ that were added after merging; concretely
- add `CanActAsAnyParty` and `CanReadAsAnyParty` rights
- make `primary_party` optional for special users
- add `DeleteUserResponse`

FYI: @da-tanabe @bame-da @cocreature @gerolf-da @stefanobaghino-da @adriaanm-da @nmarton-da 

### Pull Request Checklist

- [x] If you mean to change the status of a component, please make sure you keep [the Component Status page](https://github.com/digital-asset/daml/blob/main/docs/source/support/component-statuses.rst) up to date.

trigger the build.

### DIFF
--- a/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
+++ b/ledger-api/grpc-definitions/com/daml/ledger/api/v1/admin/user_management_service.proto
@@ -87,12 +87,6 @@ message Right {
     // The user is allowed to administrate the participant node.
     message ParticipantAdmin {}
 
-    // The user can authorize commands for any party hosted on this participant node.
-    message CanActAsAnyParty {}
-
-    // The user can all read ledger data visible to some party on this participant node.
-    message CanReadAsAnyParty {}
-
     // The user can authorize commands for the given party.
     message CanActAs {
         string party = 1;
@@ -105,9 +99,8 @@ message Right {
 
     oneof kind {
         ParticipantAdmin participant_admin = 1;
-        CanActAsAnyParty can_act_as_any_party = 2;
-        CanActAs can_act_as = 3;
-        CanReadAs can_read_as = 4;
+        CanActAs can_act_as = 2;
+        CanReadAs can_read_as = 3;
     }
 }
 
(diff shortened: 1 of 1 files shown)


cocreature (approved):
Looks sensible, thank you!

cocreature (commented):
@meiersi-da given that the extra rights are out of date now iiuc should this be updated? (and merged so we have the same base to work against)

meiersi-da:
Agreed @moritzkiefer-da . I've just found that we already have `ClaimActAsAnyParty`, so I've only removed the `ReadAsAnyParty` right. Will merge as soon as the checks pass.

cocreature:
> Agreed @moritzkiefer-da . I've just found that we already have `ClaimActAsAnyParty`, so I've only removed the `ReadAsAnyParty` right.

Note that while we have this internally, we currently do not expose this in any form to our users. So we may still want to be cautious in making this part of our public APi.

meiersi-da:
> Note that while we have this internally, we currently do not expose this in any form to our users. So we may still want to be cautious in making this part of our public APi.

I'm inclined to expose it as it is
1. well-defined
2. solves use-cases like using Navigator in an authenticated setup

That said, I might be overlooking things. So what actions do you think we need to take before exposing this?

meiersi-da:
Hmm... now having written that I get the rubber-duck effect: it seems that `CanActAsAnyParty` invites all kinds of hacks, instead of forcing us to solve the actual rights managements problems which would exist anyways in production. So yeah, I'm more inclined to remove it for now.

@bame-da what is your opinion?

cocreature:
I’m not super strongly opposed to exposing it but it doesn’t seem necessary for 2.0 and I have concerns similar to the ones you mentioned above (encouraging various antipatterns). Adding it later seems fully backwards compatible so we can add it once we’re confident it’s a good idea.

meiersi-da:
Thanks @cocreature for the valuable push-back. I've removed the xxxAnyRights

cocreature:
> Thanks @cocreature for the valuable push-back. I've removed the xxxAnyRights

thanks! and just to be clear I’m happy to revisit this outside of the 2.0 scope

PR #11910, opened 2021-11-29
Add type rep argument for interface exercises.

sofiafaro-da:
(Still WIP.)

Part of #11703. Fixes the order in which errors are raised ("wrong type"
takes priority over "does not implement interface"). This PR also simplifies
ExerciseInterface by making the guard mandatory, otherwise there's too
many variations. We can revisit that later if we want.

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1097,58 +1097,72 @@ private[lf] object SBuiltin {
...
-  final case class SBUFetchInterface(ifaceId: TypeConName) extends OnLedgerBuiltin(1) {
+  final case class SBUFetchInterface(ifaceId: TypeConName) extends OnLedgerBuiltin(2) {
     override protected def execute(
         args: util.ArrayList[SValue],
         machine: Machine,
         onLedger: OnLedger,
     ): Unit = {
       val coid = getSContractId(args, 0)
+      val expectedTemplateIdOpt = getSOptional(args, 1).map(typeRep =>
+        typeRep match {
```
cocreature, line 1108: Maybe worth defining `getTypeRep` or something like that similar to how we have `getSOptional`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
```diff
@@ -1175,19 +1189,12 @@ private[lf] object SBuiltin {
...
-        args: util.ArrayList[SValue],
-        machine: Machine,
-    ): Unit = {
-      val coid = getSContractId(args, 0)
-      val record = getSRecord(args, 1)
-      if (record.id != templateId)
-        machine.ctrl = SEDamlException(IE.WronglyTypedContract(coid, templateId, record.id))
-      else
-        machine.returnValue = SBool(true)
+  final case object SBGuardConstTrue extends SBuiltinPure(1) {
```
cocreature, line 1192: why do we need this as a builtin?

cocreature (approved):
Nice, thank you!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
sofiafaro-da: Wouldn't help here, since it's a typeRep inside an optional 🙃

cocreature: why not? Can’t you do `getSOptional(args,1).map(getSTypeRep(_))`?

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
sofiafaro-da: We don't, it just seemed much more direct than using `SEAbs(1, SEValue(SBool(true)))` as the default for now.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/SBuiltin.scala
sofiafaro-da: What is `getSTypeRep(_)` analogous to?
`getSOptional` and such are applied to args directly.

Also, what is the point in extracting `getSTypeRep` when it's only used here, and still wouldn't be enough to validate the type rep has the format we expect?

cocreature: fair, just seemed like it might be slightly cleaner to write this differently but you’re probably right that it doesn’t matter much.

remyhaemmerle-da (approved):
LGTM

Issue #11497, opened 2021-11-01
limit recursion when generating values for Scala codegen tests


S11001001:
Until [size-limited nesting](https://github.com/digital-asset/daml/blob/d678a4046d5a61bf4059cec77f669cc7c62b4be9/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala#L212-L240) was added to the LF transaction library tests, the generated values quite frequently "ran away with recursion", so to speak.

The somewhat older [Scala codegen value generators](https://github.com/digital-asset/daml/blob/22ce9409545202ef134e20d2cfb126dd6bb19149/language-support/scala/codegen-testing/src/test/scala/com/digitalasset/ledger/client/binding/ValueGen.scala#L88-L123) have proven far less likely to run away, possibly because they rely on more built-in `Arbitrary` instances, which tend to have some size checks incorporated, where the `Gen` combinators do not. However, it would probably still be wise to use `Gen.sized` in the older `ValueGen.valueChecks` to put a hard limit on recursion.

It would be best to

1. check how each built-in implicit `Arbitrary` instance used there uses `size`,
2. take special care to apply a `size` shrink on those recursive code paths in `valueChecks` that do not already shrink `size` themselves, and finally
3. exclude all recursive cases from the produced `frequency` call when `size` is small enough.

The extra `nesting` parameter will not be needed; `sized` and `resize` are sufficient.

_This is in response to a CI timeout in `//language-support/scala/codegen-testing:tests` `ValueSpec`. Increasing the CI timeout is not likely to be helpful at all with an issue like the one described above._

PR #11911, opened 2021-11-29
Remove the use of nesting with `size` approach as suggested by scalacheck.

skisel:
This is an attempt to solve #11497 

On my local workstation the test in the main branch was running for 6 seconds and with the new approach it takes 20 sec. It can be improved by redistributing flat generators and recursive ones with `Gen.frequency`. Please suggest what would you prefer. 

### Pull Request Checklist

trigger the build.

### DIFF
--- a/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
+++ b/daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
@@ -84,7 +84,8 @@ object ValueGenerators {
       firstChars ++ "1234567890"
     for {
       h <- Gen.oneOf(firstChars)
-      t <- Gen.listOf(Gen.oneOf(mainChars))
+      suffixLength <- Gen.choose(0, 999)
+      t <- Gen.listOfN(suffixLength, Gen.oneOf(mainChars)) //Name is max 999 characters
     } yield Name.assertFromString((h :: t).mkString)
   }
 
@@ -101,7 +102,7 @@ object ValueGenerators {
       .map(Time.Timestamp.assertFromLong)
 
   // generate a variant with arbitrary value
-  private def variantGen(nesting: Int): Gen[ValueVariant] =
+  def variantGen: Gen[ValueVariant] =
     for {
       id <- idGen
       variantName <- nameGen
@@ -111,12 +112,10 @@ object ValueGenerators {
           if (withoutLabels) (_: Identifier) => None
           else (variantId: Identifier) => Some(variantId)
         )
-      value <- Gen.lzy(valueGen(nesting))
+      value <- valueGen
     } yield ValueVariant(toOption(id), variantName, value)
 
-  def variantGen: Gen[ValueVariant] = variantGen(0)
-
-  private def recordGen(nesting: Int): Gen[ValueRecord] =
+  def recordGen: Gen[ValueRecord] =
     for {
       id <- idGen
       toOption <- Gen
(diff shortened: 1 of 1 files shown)


daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
```diff
@@ -84,7 +84,7 @@ object ValueGenerators {
       firstChars ++ "1234567890"
     for {
       h <- Gen.oneOf(firstChars)
-      t <- Gen.listOf(Gen.oneOf(mainChars))
+      t <- Gen.listOfN(999, Gen.oneOf(mainChars)) //Name is max 999 characters
```
skisel, line 87: This has to be limited to 999 as otherwise generator generates invalid name:
`Message: Name is too long (max: 1000)`

This can be reproduced by playing around with `minSize` of `PropertyCheckConfiguration` of tests using these generators.

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
```diff
@@ -201,34 +189,35 @@ object ValueGenerators {
...
+    party.map(ValueParty),
+    Gen.oneOf(ValueTrue, ValueFalse)
+  )
+
+  def valueGen: Gen[Value] = Gen.lzy {
+    Gen.sized { size =>
+      for {
+        s <- Gen.choose(0, size)
+        value <- if (s < 1) flatGen
+        else Gen.oneOf(flatGen, Gen.resize(size / (s + 1), nestedGen))
```
skisel, line 218: To reduce the use of recursive generators, it can be changed as follows
```
        else Gen.frequency(
          5 -> flatGen,
          1 -> Gen.resize(size / (s + 1), nestedGen)
        )
```

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
```diff
@@ -201,34 +189,35 @@ object ValueGenerators {
...
+    valueOptionalGen,
+    valueMapGen
+  )
+
+  private def flatGen = Gen.oneOf(
+    dateGen.map(ValueDate),
+    Gen.alphaStr.map(ValueText),
+    unscaledNumGen.map(ValueNumeric),
+    numGen(Decimal.scale).map(ValueNumeric),
+    Arbitrary.arbLong.arbitrary.map(ValueInt64),
```
S11001001, line 205:
```suggestion
    arbitrary[Long].map(ValueInt64),
```

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
S11001001: We definitely want to reduce that usage; that's why the original code uses `frequency` after all.

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
S11001001: I don't think we want exactly 1000 characters in every name, though, as `listOfN` will deliver; perhaps we could `choose` a size (with a bias towards small names, perhaps; we don't get big value out of big names in most tests I'd wager).

S11001001 (commented):
I'm very glad to remove `nesting` and replace it with `size`; overall these changes look good to me. Thanks for your contribution!

However, this is separate from #11497, which regards changing the generators in `language-support/scala/codegen-testing`. So we should merge this (with changes mentioned here) but leave #11497 open.

stefanobaghino-da:
I see that #11497's being mentioned will not be automatically picked by Github to close it, so that should be fine. @S11001001 if you are satisfied please feel free to approve. @skisel once CI is green and @S11001001 approves, our tradition is that whomever opens the PR clicks "squash and merge". 😉

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
skisel: Thanks for pointing it out - I missed it somehow. Not sure yet how to make it biased towards small names, suggestions? probability distribution?

stefanobaghino-da:
@skisel The PR needs formatting. You can run `./fmt.sh --diff` from the project root to fix it and commit the change. I generally recommend having `./fmt.sh --test --diff && ./ci/check-changelog.sh` in a pre-push `git` hook. It usually saves a round-trip to CI.

skisel:
Thanks @stefanobaghino-da, will do that now and squash the changes

stefanobaghino-da:
TIL you can use raw pointers in Go. Not sure of what happened on CI, I hope it's some kind of flake. @cocreature why do we have Go in there? It seems like it's part of the documentation build, do we use a static site generator like Hugo?

cocreature:
> TIL you can use raw pointers in Go. Not sure of what happened on CI, I hope it's some kind of flake. @cocreature why do we have Go in there? It seems like it's part of the documentation build, do we use a static site generator like Hugo?

I’ve seen it flake before so very likely. We use it for the protobuf plugin that generates docs out of our protobufs.

stefanobaghino-da:
> I’ve seen it flake before so very likely. We use it for the protobuf plugin that generates docs out of our protobufs.

Uh-uh, thanks.

stefanobaghino-da:
@S11001001 if you think this can be merged I'd say you can approve it so that @skisel can merge it tomorrow. Thanks in advance.

daml-lf/transaction-test-lib/src/main/scala/lf/value/test/ValueGenerators.scala
S11001001: I'd just use `frequency` to pick the second (upper-bound) parameter to `choose`. So you could have 90% small names, 8% small-to-medium names, 2% any size.

S11001001 (approved):

PR #11912, opened 2021-11-29
Use Absolute-indexes as keys for the Env-mapping during closure-conversion

nickchapman-da:
Done
- Use Absolute-indexes as keys for the `Env.mapping` during closure-conversion.
- [Do runtime check to confirm behaviour matches the existing Relative-indexes, on full repo. See 1st commit]
- remove Relative-index keys for `Env.mapping`
- remove quadratic `shiftLoc`
- improve `ClosureConversion` stack-safety tests

This addresses aspect (2) of the quadratic complexity of `shiftLoc`, described in issue #11830.

#### Benchmarking

shows a small, 3%, performance improvement in speedy compilation with this change (Actually, multiply that percentage by 3, because the speedy compilation benchmark is dominated by type checking which takes twice as long as the compilation -- fixed in coming PR: #11927).

base:
```
  SpeedyCompilation.bench             avgt  200  0.197 ±  0.001   s/op
```
change:
```
  SpeedyCompilation.bench             avgt  200  0.191 ± 0.001   s/op
```

### DIFF
--- a/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
+++ b/daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
@@ -34,25 +34,25 @@ private[speedy] object ClosureConversion {
 
       def extend(n: Int): Env = {
         // Create mappings for `n` new stack items, and combine with the (unshifted!) existing mapping.
-        val m2 = (1 to n).view.map { i =>
-          val abs = Abs(this.sourceDepth + (n - i))
-          (abs, target.SELocAbsoluteS(this.targetDepth + n - i))
+        val m2 = (0 to n - 1).view.map { i =>
+          val abs = Abs(sourceDepth + i)
+          (abs, target.SELocAbsoluteS(targetDepth + i))
         }
-        Env(this.sourceDepth + n, mapping ++ m2, this.targetDepth + n)
+        Env(sourceDepth + n, mapping ++ m2, targetDepth + n)
       }
 
       def absBody(arity: Int, fvs: List[Abs]): Env = {
         val newRemapsF: Map[Abs, target.SELoc] = fvs.view.zipWithIndex.map { case (abs, i) =>
           abs -> target.SELocF(i)
         }.toMap
-        val newRemapsA = (1 to arity).view.map { case i =>
-          val abs = Abs(arity - i + sourceDepth)
-          abs -> target.SELocA(arity - i)
+        val newRemapsA = (0 to arity - 1).view.map { case i =>
+          val abs = Abs(sourceDepth + i)
+          abs -> target.SELocA(i)
         }
         // The keys in newRemapsF and newRemapsA are disjoint
         val m1 = newRemapsF ++ newRemapsA
         // Only targetDepth is reset to 0 in an abstraction body
-        Env(this.sourceDepth + arity, m1, 0)
+        Env(sourceDepth + arity, m1, 0)
       }
     }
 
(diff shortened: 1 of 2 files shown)


cocreature (approved):
Nice, thank you!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
```diff
@@ -20,42 +20,45 @@ private[speedy] object ClosureConversion {
...
-        // We just update the keys of the map (the relative-indexes from the original SEVar)
-        val m1 = mapping.map { case (k, loc) => (n + k, loc) }
-        // And create mappings for the `n` new stack items
-        val m2 = (1 to n).view.map { rel =>
-          val abs = this.depth + n - rel
-          (rel, target.SELocAbsoluteS(abs))
+      def extend(n: Int): Env = {
+        // Create mappings for `n` new stack items, and combine with the (unshifted!) existing mapping.
+        val m2 = (1 to n).view.map { i =>
+          val abs = Abs(this.sourceDepth + (n - i))
```
remyhaemmerle-da, line 38:
```suggestion
          val abs = Abs(this.sourceDepth + n - i)
```

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
```diff
@@ -20,42 +20,45 @@ private[speedy] object ClosureConversion {
...
+          (abs, target.SELocAbsoluteS(this.targetDepth + n - i))
```
remyhaemmerle-da, line 39: curious to know why we subtract `i` from `n`, and not use something like :
```suggestion
        val m2 = (0 until n).view.map { i =>
          val abs = Abs(this.sourceDepth + i)
          (abs, target.SELocAbsoluteS(this.targetDepth + i))
```

remyhaemmerle-da (approved):
LGTM.

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
nickchapman-da: Yeah. This was a hang-over from the old _relative_ indexing which started from `1`.
The new _absolute_ indexing starts from `0`

I simplified the code along the lines you suggest in both the `extend` and `absBody` methods.
Much simpler now. Thanks!

daml-lf/interpreter/src/main/scala/com/digitalasset/daml/lf/speedy/ClosureConversion.scala
```diff
@@ -20,42 +20,45 @@ private[speedy] object ClosureConversion {
...
-      def shift(n: Int): Env = {
-        // We just update the keys of the map (the relative-indexes from the original SEVar)
-        val m1 = mapping.map { case (k, loc) => (n + k, loc) }
-        // And create mappings for the `n` new stack items
-        val m2 = (1 to n).view.map { rel =>
-          val abs = this.depth + n - rel
-          (rel, target.SELocAbsoluteS(abs))
+      def extend(n: Int): Env = {
+        // Create mappings for `n` new stack items, and combine with the (unshifted!) existing mapping.
+        val m2 = (0 to n - 1).view.map { i =>
```
remyhaemmerle-da, line 37:
```suggestion
        val m2 = (0 until n).view.map { i =>
```

nickchapman-da: ok!

PR #11916, opened 2021-11-29
[kvutils] - Integrate the deduplication offset support into kvutils [kvl-1172]

nicu-da:
- this adds full support for offsets as deduplication periods in kv ledgers
- in a follow up PR we will expose the offset through the API and also add conformance tests for it

changelog_begin
kvutils - Add support for offsets as deduplication periods. This is done by reading the completion at the specified offset and converting the deduplication offset into a deduplication duration using the interval between the record time of the read completion and the submission time of the current submission
changelog_end

### Pull Request Checklist

trigger the build.


ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
```diff
@@ -148,21 +162,36 @@ trait LedgerFactory[+RWS <: ReadWriteService, ExtraConfig]
       config: Config[ExtraConfig],
       participantConfig: ParticipantConfig,
       engine: Engine,
-  )(implicit materializer: Materializer, loggingContext: LoggingContext): ResourceOwner[RWS] =
-    readWriteServiceOwner(config, participantConfig, engine)
+      completionService: IndexCompletionsService,
```
SamirTalwar-DA, line 165: So, this parameter (and the engine) are not actually used by the read service, only the write service. We therefore don't even need to construct these objects in "indexer" mode.

I think this is starting to tell us that it's no longer sensible to couple the reader and writer.

SamirTalwar-DA (commented):
Looks sensible; I have some comments.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala
```diff
@@ -122,8 +122,31 @@ final class Runner[T <: ReadWriteService, Extra](
...
+                  metrics = metrics,
+                  engine = sharedEngine,
+                  servicesExecutionContext = servicesExecutionContext,
+                  lfValueTranslationCache = lfValueTranslationCache,
+                ).acquire()
                 ledger <- factory
-                  .readWriteServiceOwner(config, participantConfig, sharedEngine)
+                  .readWriteServiceOwner(config, participantConfig, sharedEngine, indexService)(
+                    materializer,
+                    servicesExecutionContext,
```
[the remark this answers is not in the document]
nicu-da: The issue here it's not that we have too many implicits in scope. We have to be explicit to avoid having to refactor this method so that we can make `servicesExecutionContext` implicit.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
nicu-da: Yes. As a matter of fact, I've initially started refactoring this but realized that it's not really that straightforward. 
The problem here is that the design is inherently wrong, and the composition flow is actually reversed, instead of creating a read-write service from the individual read service and write service, we create the individual read service and write service from the always composed read-write service. 
I will try to tackle this in a new PR

hubert-da (approved):
LGTM

hubert-da:
I forgot to mention that I'm a huge fan of not doing it in the gRPC layer and using the `IndexCompletionService`, great stuff!

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/Runner.scala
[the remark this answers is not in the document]
SamirTalwar-DA: I understand; I'm asking why it's not an explicit parameter of the `readWriteServiceOwner` method (and other methods lower down).

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
SamirTalwar-DA: Totally agree that it's out of scope here.

SamirTalwar-DA (approved):

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/api/KeyValueParticipantState.scala
```diff
@@ -24,33 +29,47 @@ import com.daml.lf.data.{Ref, Time}
...
   * Will report [[com.daml.ledger.api.health.Healthy]] as health status only if both
   * `reader` and `writer` are healthy.
   *
-  * @param reader       [[LedgerReader]] instance to adapt
-  * @param writer       [[LedgerWriter]] instance to adapt
-  * @param metrics      used to record timing metrics for [[LedgerWriter]] calls
+  * @param reader                       [[LedgerReader]] instance to adapt
+  * @param writer                       [[LedgerWriter]] instance to adapt
+  * @param metrics                      used to record timing metrics for [[LedgerWriter]] calls
+  * @param deduplicationPeriodConverter used to add support for [[com.daml.ledger.api.DeduplicationPeriod.DeduplicationOffset]]
```
fabiotudone-da, line 47:
```suggestion
  * @param deduplicationPeriodConverter supports [[com.daml.ledger.api.DeduplicationPeriod.DeduplicationOffset]] by converting it into a duration
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/deduplication/CompletionBasedDeduplicationPeriodConverter.scala
```diff
@@ -0,0 +1,69 @@
...
+  private def completionAtOffset(
+      request: CompletionRequest
+  )(implicit
+      mat: Materializer,
+      loggingContext: LoggingContext,
+  ): Future[Option[CompletionStreamResponse]] = {
+    completionService
+      .getCompletions(request.offset, request.offset, request.applicationId, request.parties)
+      .runWith(Sink.headOption)
+  }
```
fabiotudone-da, line 68: Nit:

```suggestion
  ): Future[Option[CompletionStreamResponse]] =
    completionService
      .getCompletions(request.offset, request.offset, request.applicationId, request.parties)
      .runWith(Sink.headOption)
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/deduplication/CompletionBasedDeduplicationPeriodConverter.scala
```diff
@@ -0,0 +1,69 @@
...
+            case None => Left(DeduplicationConversionFailure.CompletionRecordTimeNotAvailable)
+          }
+        } else {
+          Left(DeduplicationConversionFailure.CompletionOffsetNotMatching)
+        }
+      case Some(CompletionStreamResponse(None, _)) =>
+        Left(DeduplicationConversionFailure.CompletionCheckpointNotAvailable)
+      case None => Left(DeduplicationConversionFailure.CompletionAtOffsetNotFound)
+    }
+  }
```
fabiotudone-da, line 57: Nit:

```suggestion
  ): Future[Either[DeduplicationConversionFailure, Duration]] =
    completionAtOffset(
      CompletionRequest(
        ledgerId,
        applicationId,
        actAs,
        LedgerOffset.Absolute(offset),
      )
    ).map {
      case Some(CompletionStreamResponse(Some(checkpoint), _)) =>
        if (checkpoint.offset.flatMap(_.value.absolute).contains(offset)) {
          checkpoint.recordTime match {
            case Some(recordTime) =>
              val duration = Duration.between(recordTime.asJavaInstant, submittedAt)
              Right(duration)
            case None => Left(DeduplicationConversionFailure.CompletionRecordTimeNotAvailable)
          }
        } else {
          Left(DeduplicationConversionFailure.CompletionOffsetNotMatching)
        }
      case Some(CompletionStreamResponse(None, _)) =>
        Left(DeduplicationConversionFailure.CompletionCheckpointNotAvailable)
      case None => Left(DeduplicationConversionFailure.CompletionAtOffsetNotFound)
    }
```

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/deduplication/NotSupportedDeduplicationPeriodConverter.scala
```diff
@@ -0,0 +1,30 @@
...
+      applicationId: ApplicationId,
+      actAs: Set[Party],
+      submittedAt: Instant,
+  )(implicit
+      mat: Materializer,
+      ec: ExecutionContext,
+      loggingContext: LoggingContext,
+  ): Future[Either[DeduplicationConversionFailure, Duration]] = throw new IllegalArgumentException(
+    "Offsets are not supported as deduplication periods"
+  )
```
fabiotudone-da, line 28: What failure mode does this result into? Shouldn't a failed future be used?

ledger/ledger-on-memory/src/test/suite/scala/com/daml/ledger/on/memory/InMemoryLedgerReaderWriterIntegrationSpec.scala
```diff
@@ -61,5 +62,6 @@ class InMemoryLedgerReaderWriterIntegrationSpec
       readerWriter,
       metrics,
       enableSelfServiceErrorCodes = true,
+      NotSupportedDeduplicationPeriodConverter,
```
fabiotudone-da, line 65: Why can't we support it here?

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/deduplication/NotSupportedDeduplicationPeriodConverter.scala
nicu-da: Good catch, a failed future would make a lot more sense.

ledger/ledger-on-memory/src/test/suite/scala/com/daml/ledger/on/memory/InMemoryLedgerReaderWriterIntegrationSpec.scala
nicu-da: We basically test one layer too low at this point, therefore we don't have a way of reading a completion at this level.

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
```diff
@@ -66,3 +74,30 @@ class KeyValueReadWriteFactory(
...
+    delegate: ReadWriteServiceFactory,
+    config: Config[_],
+    completionsService: IndexCompletionsService,
+)(implicit materializer: Materializer, ec: ExecutionContext)
+    extends ReadWriteServiceFactory {
+  override def readService(): ReadService = delegate.readService()
+
+  override def writePackagesService(): WritePackagesService = delegate.writePackagesService()
+
+  override def writeService(): WriteService = {
```
nicu-da, line 88: There were 3 alternatives as to where do this:
- in the Runner, not chosen as to not any more logic in the runner, it should ideally just connect bigger pieces and not be aware of such a low level detail
- as part of the existing ReadWriteServiceFactory, adding the IndexCompletionService as an argument to the `writeService` method - not chosen as this would expose the `IndexCompletionService` in multiple modules where the factory implementation should really not know about it
- as a new decorator for the `ReadWriteServiceFactory` - the implementation, not ideal as it still forces us to wrap the initial factory in this new layer, but feels cleaner than the other options

ledger/participant-state/kvutils/src/main/scala/com/daml/ledger/participant/state/kvutils/deduplication/NotSupportedDeduplicationPeriodConverter.scala
fabiotudone-da: What will happen here from the point of view of the application? I am wondering about that because we may actually need a self-service error here.

hubert-da (approved):

ledger/participant-state/kvutils/app/src/main/scala/com/daml/ledger/participant/state/kvutils/app/LedgerFactory.scala
```diff
@@ -35,7 +43,7 @@ trait ReadWriteServiceFactory {
 
   def readService(): ReadService
 
-  def writePackageService(): WritePackagesService
+  def writePackagesService(): WritePackagesService
```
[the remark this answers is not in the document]
nicu-da: Me too but I decided that it's best if we keep in sync with the type :(

PR #11922, opened 2021-11-30
Drop support for Daml-LF party literals from the Scala side

### LINKED ISSUE #11581: Enforce `forbidPartyLiterals = True` for Daml 2.0
The compiler has always emitted this since SDK 2.0 so there is zero customer impact here unless someone has handwritten daml-lf (not the case afaik).

cocreature:
This PR enforces that forbidPartyLiterals is always `true` and drops
the corresponding literals from the AST. Haskell side is in #11930

fixes #11581

changelog_begin
changelog_end

### Pull Request Checklist

trigger the build.


daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ComparisonSBuiltinTest.scala
```diff
@@ -114,7 +114,9 @@ class ComparisonSBuiltinTest extends AnyWordSpec with Matchers with TableDrivenP
             e"1970-01-01T00:00:00.000000Z",
             e"2020-02-02T20:20:02.020000Z",
           ),
-          t"Party" -> List(e"'alice'", e"'bob'", e"'carol'"),
+          // Parties cannot be built from expressions.
+          // We map at runtime the `party1`, `party2` and, `party3` two 3 party IDS in increasing order.
+          t"Party" -> List(e"party1", e"party2", e"party3"),
```
remyhaemmerle-da, line 119:
```suggestion
           t"Party" -> List(
            e"""TEXT_TO_PARTY "alice"""",
            e"""TEXT_TO_PARTY "bob"""",
            e"""TEXT_TO_PARTY "carol"""",
          ),
```

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/SBuiltinTest.scala
```diff
@@ -731,8 +731,10 @@ class SBuiltinTest extends AnyFreeSpec with Matchers with TableDrivenPropertyChe
 
     "EQUAL @ContractId" - {
       "works as expected" in {
-        eval(e"EQUAL @(ContractId Mod:T) 'contract1' 'contract1'") shouldBe Right(SBool(true))
-        eval(e"EQUAL @(ContractId Mod:T) 'contract1' 'contract2'") shouldBe Right(SBool(false))
+        val cid1 = SContractId(Value.ContractId.assertFromString("#contract1"))
+        val cid2 = SContractId(Value.ContractId.assertFromString("#contract2"))
+        evalApp(e"EQUAL @(ContractId Mod:T)", Array(cid1, cid1)) shouldBe Right(SBool(true))
+        evalApp(e"EQUAL @(ContractId Mod:T)", Array(cid1, cid2)) shouldBe Right(SBool(false))
```
remyhaemmerle-da, line 737: Thanks to fix that.

remyhaemmerle-da (approved):
LGTM.

I most of the test that use the LF parser, I would have use `TEXT_TO_PARTY` instead abstraction. This would problably help a bit reducing the diff.

cocreature:
> I would have use `TEXT_TO_PARTY` instead abstraction.

I disagree with that, imho `TEXT_TO_PARTY` is a big antipattern and our tests should roughly test what we want users to write not something different. We also don’t have a `FROM_SOME` so dealing with the optional is very annoying.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ComparisonSBuiltinTest.scala
```diff
@@ -114,7 +114,13 @@ class ComparisonSBuiltinTest extends AnyWordSpec with Matchers with TableDrivenP
...
             e"2020-02-02T20:20:02.020000Z",
           ),
-          t"Party" -> List(e"'alice'", e"'bob'", e"'carol'"),
+          // Parties cannot be built from expressions.
+          // We map at runtime the `party1`, `party2` and, `party3` two 3 party IDS in increasing order.
+           t"Party" -> List(
+            e"""TEXT_TO_PARTY "alice"""",
+            e"""TEXT_TO_PARTY "bob"""",
+            e"""TEXT_TO_PARTY "carol"""",
+          ),
```
remyhaemmerle-da, line 123: If you use my suggestion, you can revert all the other changes to the file.

daml-lf/interpreter/src/test/scala/com/digitalasset/daml/lf/speedy/ComparisonSBuiltinTest.scala
cocreature: I’ve reverted that again, The fact that it returns an `Optional` (which works but is against the spirit of this test imho) and that I’d like to eventually kill that primitive completely makes me dislike that approach. Happy to discuss if you disagree.

remyhaemmerle-da:
> > I would have use `TEXT_TO_PARTY` instead abstraction.
> 
> I disagree with that, imho `TEXT_TO_PARTY` is a big antipattern and our tests should roughly test what we want users to write not something different.

:trollface: 
I think it is an antipattern, to consider the DAML antipattern when designing low-level tests for LF.

>> I most of the test that use the LF parser, I would have use TEXT_TO_PARTY instead abstraction. This would probably help a bit reducing the diff.

I will soften a bit my position, I think it is totally fine to `TEXT_TO_PARTY` in `daml-lf/validation` and `daml-lf/interpretation` (exception for testing the  for text <-> party conversion builtins).


Note: 
- in some cases you can use an intermediate definition like
```
val toParty: Text -> a = 
   \(t: Text) -> case TEXT_TO_PARTY t of  Some x -> x | None -> ERROR @a "empty option";
```
- in other cases you can simply avoid the party literals.

cocreature:
> I think it is an antipattern, to consider the DAML antipattern when designing low-level tests for LF. :trollface:

that’s fair but ideally I’d like to kill that primitive from LF eventually as well and at that point having tests rely on it becomes really annoying.

daml-lf/archive/src/main/scala/com/digitalasset/daml/lf/archive/DecodeV1.scala
```diff
@@ -472,15 +470,17 @@ private[archive] class DecodeV1(minor: LV.Minor) {
         "EnumConstructors.constructors",
       )
 
-    private[this] def decodeDefValue(lfValue: PLF.DefValue): DValue = {
+    private[lf] def decodeDefValue(lfValue: PLF.DefValue): DValue = {
+      if (!lfValue.getNoPartyLiterals) {
+        throw Error.Parsing("DefValue must have no_party_literals set to true")
```
oliverse-da, line 475: @cocreature @remyhaemmerle-da : throwing this parsing error changes a handful of canton test based on the "illformed.dar" to no longer generate a "standard"/"self-service" package validation error anymore according to canton-ci. e.g.:

https://app.circleci.com/pipelines/github/DACH-NY/canton/29790/workflows/9b403b83-fe5b-4078-9a56-0cc86bc6aef4/jobs/595170

Shouldn't this be a package validation error, too?

cc - @andreaslochbihler-da

cocreature: @remyhaemmerle-da correct me if I’m wrong, my understanding is that we throw different errors for decoding than we throw for validation (e.g. typechecking). So this changes it to throw a decoding error instead of a package validation error. If we don’t have self-service errors for decoding failures we should fix that.

cocreature: In particular even without this change you can have decoding failures so we need to handle them regardless.

oliverse-da: Makes sense in general, but I think with these tests in particularly we were looking for a dar (provided by Remy?) that triggered a package validation problem, so we could verify that we properly handle validation errors in canton.

andreaslochbihler-da: Yes, we wanted to get a type checking error with `illformed.dar`. Probably it suffices to regenerate the dar such that the `getNoPartyLiterals` is set. @remyhaemmerle-da Can you regenerate such a dar?

cocreature: sent the DAR on slack

remyhaemmerle-da:
>  my understanding is that we throw different errors for decoding than we throw for validation (e.g. typechecking). So this changes it to throw a decoding error instead of a package validation error. If we don’t have self-service errors for decoding failures we should fix that.
:+1:

remyhaemmerle-da:
What do you mean by "self-service" ? 
`com.daml.lf.archive.Error.Parsing` are standard error for package decoding. I think using a string to describe the error in this case is good enough, we do not need an ADT for this kind of errors.

